Issue #1103, opened 2018-09-07
Tests for the console


mrzzzrm:
These should at the very least include a number of console commands in a console script that shouldn't crash the console. 
Full output checks are probably hard to realize (e.g., checking generated image data), but checking that the text output is correct is probably doable?

What do you think?

mrks:
Agreed. Maybe this is something that we want to test from the outside, e.g., using `expect`?

mrzzzrm:
Okay, but we'll just test text output, right? Testing the visualizations is out of the question, amirite?

mrks:
If anything, we can just make sure that *something* is visualized (i.e., a png is created)

mrks:
d'oh

mrzzzrm:
We need this, and we need it for the benchmarks, too: https://github.com/hyrise/hyrise/pull/1356#discussion_r244670617

mrzzzrm:
Have you used `expect` before, @mrks?

mrzzzrm:
Where did you find docs about the `default {` part? The man page only vaguely mentions it and I do not want to risk missing a crucial source of documentation.

mrks:
<pre>
<strong>expect</strong> [[-opts] pat1 body1] ... [-opts] patn [bodyn]
    waits until one of the patterns matches the output of a spawned process, a specified time period has passed, or an end-of-file is seen. If the final body is empty, it may be omitted. 
Patterns from the most recent
    expect_before command are implicitly used before any other patterns. Patterns from the most recent expect_after command are implicitly used after any other patterns. 
If the arguments to the entire
    expect statement require more than one line, all the arguments may be "braced" into one so as to avoid terminating each line with a backslash. In this one case, the usual Tcl substitutions will occur despite the braces. 
If a pattern is the keyword
    eof, the corresponding body is executed upon end-of-file. If a pattern is the keyword timeout, the corresponding body is executed upon timeout. If no timeout keyword is used, an implicit null action is executed upon timeout. The default timeout period is 10 seconds but may be set, for example to 30, by the command "set timeout 30". An infinite timeout may be designated by the value -1. <strong>If a pattern is the keyword default, the corresponding body is executed upon either timeout or end-of-file. </strong>
</pre>

I haven't found a way to say "this and nothing else", only "wait until the program prints ...".

mrzzzrm:
I tinkered with this a bit and found the tcl-expect a pain. Maybe this: https://pypi.org/project/pyexpect/?

[...]

PR #1399, opened 2019-01-12
Update join_order_benchmark.cpp

mrzzzrm:
I am not convinced by the copy-and-paste approach...


mrzzzrm:
(also: format not being happy with the master)

mrks:
We really need #1103 :/

PR #1400, opened 2019-01-13
Add encodings to Console and Print

### LINKED ISSUE #1363: Console: load command should accept encoding options
Whenever we "want to try something real quick", we can only do so with unencoded data. At least the `load` command for csv/tbl files should allow setting an encoding option.

mrks:
fixes #1363

## Adds encoding options to `load` command in console

```
(debug)> load ../resources/test_data/csv/float_int.csv t Dictionary
Loading ../resources/test_data/csv/float_int.csv into table "t"
Encoding "t" using Dictionary
```

## Removes the need for specifying a table name

```
(debug)> load ../resources/test_data/csv/float_int.csv
Loading ../resources/test_data/csv/float_int.csv into table "float_int"
Encoding "float_int" using Unencoded
```

## Enables printing of segment encoding information
```
(debug)> print t
=== Columns
|       b|       a||        MVCC        |
|   float|     int||_BEGIN|_END  |_TID  |
|not null|not null||      |      |      |
=== Chunk 0 ===
|<Dic:1B>|<Dic:1B>||
|   458.7|   12345||     0|      |      |
```

### DIFF
--- a/src/benchmarklib/benchmark_runner.cpp
+++ b/src/benchmarklib/benchmark_runner.cpp
@@ -1,5 +1,6 @@
 #include <json.hpp>
 
+#include <boost/range/adaptors.hpp>
 #include <random>
 
 #include "cxxopts.hpp"
@@ -432,23 +433,10 @@ void BenchmarkRunner::_create_report(std::ostream& stream) const {
 cxxopts::Options BenchmarkRunner::get_basic_cli_options(const std::string& benchmark_name) {
   cxxopts::Options cli_options{benchmark_name};
 
-  // Make sure all current encoding types are shown
-  std::vector<std::string> encoding_strings;
-  encoding_strings.reserve(encoding_type_to_string.right.size());
-  for (const auto& encoding : encoding_type_to_string.right) {
-    encoding_strings.emplace_back(encoding.first);
-  }
-
-  const auto encoding_strings_option = boost::algorithm::join(encoding_strings, ", ");
-
-  // Make sure all current compression types are shown
-  std::vector<std::string> compression_strings;
-  compression_strings.reserve(vector_compression_type_to_string.right.size());
-  for (const auto& vector_compression : vector_compression_type_to_string.right) {
-    compression_strings.emplace_back(vector_compression.first);
-  }
-
-  const auto compression_strings_option = boost::algorithm::join(compression_strings, ", ");
+  const auto get_first = boost::adaptors::transformed([](auto it) { return it.first; });
+  const auto encoding_strings_option = boost::algorithm::join(encoding_type_to_string.right | get_first, ", ");
+  const auto compression_strings_option =
+      boost::algorithm::join(vector_compression_type_to_string.right | get_first, ", ");
 
   // If you add a new option here, make sure to edit CLIConfigParser::basic_cli_options_to_json() so it contains the
   // newest options. Sadly, there is no way to to get all option keys to do this automatically.
--- a/src/benchmarklib/benchmark_runner.cpp
+++ b/src/benchmarklib/benchmark_runner.cpp
@@ -433,6 +433,7 @@ void BenchmarkRunner::_create_report(std::ostream& stream) const {
 cxxopts::Options BenchmarkRunner::get_basic_cli_options(const std::string& benchmark_name) {
   cxxopts::Options cli_options{benchmark_name};
 
+  // Create a comma separated strings with the encoding and compression options
   const auto get_first = boost::adaptors::transformed([](auto it) { return it.first; });
   const auto encoding_strings_option = boost::algorithm::join(encoding_type_to_string.right | get_first, ", ");
   const auto compression_strings_option =
--- a/src/bin/console/console.cpp
+++ b/src/bin/console/console.cpp
@@ -474,7 +474,7 @@ int Console::_generate_tpch(const std::string& args) {
 int Console::_load_table(const std::string& args) {
   std::vector<std::string> arguments = trim_and_split(args);
 
-  if (arguments.size() < 1 || arguments.size() > 3) {
+  if (arguments.empty() || arguments.size() > 3) {
     out("Usage:\n");
     out("  load FILEPATH [TABLENAME [ENCODING]]\n");
     return ReturnCode::Error;
--- a/src/bin/console/console.cpp
+++ b/src/bin/console/console.cpp
@@ -359,11 +359,13 @@ int Console::_help(const std::string&) {
   auto encoding_options = std::string{"                                               Encoding options: "};
   encoding_options += boost::algorithm::join(
       encoding_type_to_string.right | boost::adaptors::transformed([](auto it) { return it.first; }), ", ");
-  // Split the encoding options in lines of 120 and add padding
+  // Split the encoding options in lines of 120 and add padding. For each input line, it takes up to 120 characters
+  // and replaces the following space(s) with a new line. `(?: +|$)` is a non-capturing group that matches either
+  // a non-zero number of spaces or the end of the line.
   auto line_wrap = std::regex{"(.{1,120})(?: +|$)"};
   encoding_options =
       regex_replace(encoding_options, line_wrap, "$1\n                                                 ");
-  // Remove line added at the end
+  // Remove the 49 spaces and the new line added at the end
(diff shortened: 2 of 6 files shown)


src/benchmarklib/benchmark_runner.cpp
```diff
@@ -432,23 +433,10 @@ void BenchmarkRunner::_create_report(std::ostream& stream) const {
...
-
-  // Make sure all current compression types are shown
-  std::vector<std::string> compression_strings;
-  compression_strings.reserve(vector_compression_type_to_string.right.size());
-  for (const auto& vector_compression : vector_compression_type_to_string.right) {
-    compression_strings.emplace_back(vector_compression.first);
-  }
-
-  const auto compression_strings_option = boost::algorithm::join(compression_strings, ", ");
+  const auto get_first = boost::adaptors::transformed([](auto it) { return it.first; });
```
mrzzzrm: (comment nazi) Can you top this block with "// Create a comma separated strings with the encoding and compression options"

src/bin/console/console.cpp
```diff
@@ -351,14 +356,26 @@ void Console::out(const std::shared_ptr<const Table>& table, uint32_t flags) {
 int Console::_exit(const std::string&) { return Console::ReturnCode::Quit; }
 
 int Console::_help(const std::string&) {
+  auto encoding_options = std::string{"                                               Encoding options: "};
+  encoding_options += boost::algorithm::join(
+      encoding_type_to_string.right | boost::adaptors::transformed([](auto it) { return it.first; }), ", ");
+  // Split the encoding options in lines of 120 and add padding
+  auto line_wrap = std::regex{"(.{1,120})(?: +|$)"};
```
mrzzzrm: I can not read this. Can a stupid bloody function with a loop do this? What the hell is "?: +|$"? "-50", where does that number come from?

src/bin/console/console.cpp
```diff
@@ -455,57 +472,72 @@ int Console::_generate_tpch(const std::string& args) {
...
+        encoding_type_to_string.right | boost::adaptors::transformed([](auto it) { return it.first; }), ", ");
+    out("Error: Invalid encoding type: '" + encoding + "', try one of these: " + encoding_options + "\n");
+    return ReturnCode::Error;
+  }
+
+  out("Encoding \"" + tablename + "\" using " + encoding + "\n");
+  try {
+    ChunkEncoder::encode_all_chunks(StorageManager::get().get_table(tablename), type->second);
+  } catch (const std::exception& exception) {
+    out("Exception while encoding - table left partially unencoded:\n  " + std::string(exception.what()) + "\n");
```
mrzzzrm: I thought we agreed that `Assert` failure means "Hyrise is borked, shutting down"?

Can you check whether the encoding options are valid for this table - and not do anything if they are not. I know this is cumbersome, but I really do not want to legitimize the catching of Exceptions from Asserts.

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -1,5 +1,6 @@
 #include <json.hpp>
 
+#include <boost/range/adaptors.hpp>
 #include <random>
```
mrzzzrm: You left vector compression out deliberately, right?

mrks: I felt that `load <filename> [<tablename> [<encoding> [<compression>]]]` would be too complicated if we don't use it. If we need it in the future, we should revisit how the load command takes its arguments.

src/bin/console/console.cpp
mrzzzrm: (Should use the csv parser directly) + Make the CsvParser use AssertInput.

mrzzzrm: Had my look, beheld greatness.

```
  try {
    gt->execute();
  } catch (const std::exception& exception) {
    out("Exception thrown while loading table:\n  " + std::string(exception.what()) + "\n");
    return ReturnCode::Error;
  }
```

mrzzzrm: The trycatchs in console.cpp range from two catching InvalidInputs (which is okay) over some catching std::exception (which is not okay) that do *probably* not touch global data (hopefully) over those catching std::exception while happily accessing global data (StorageManager::add_table generates stats, e.g.) and potentially turning hyrise into Frankensteins monster. 
This is not okay.

src/benchmarklib/benchmark_runner.cpp
mrks: Yepp, good point.

src/bin/console/console.cpp
mrzzzrm: \*sobs profoundly*

src/lib/operators/print.cpp
```diff
@@ -152,4 +164,64 @@ std::string Print::_truncate_cell(const AllTypeVariant& cell, uint16_t max_width
...
+          segment_type += "1B";
+          break;
+        }
+        case CompressedVectorType::SimdBp128: {
+          segment_type += "BP";
+          break;
+        }
+      }
+    } else {
+      segment_type += "--";
```
Bouncner: That would print `RLE:--` for RLE-encoded columns, right? Why not simply `RLE` and only append for encoding types that have the ability to further compress?

src/test/operators/print_test.cpp
```diff
@@ -307,4 +310,26 @@ TEST_F(OperatorsPrintTest, NullableColumnPrinting) {
...
+
+  Print::print(table, 2, output);
+
+  auto expected_output =
+      "=== Columns\n"
+      "|       a|       b||        MVCC        |\n"
+      "|     int|   float||_BEGIN|_END  |_TID  |\n"
+      "|not null|not null||      |      |      |\n"
+      "=== Chunk 0 ===\n"
+      "|<Dic:1B>|<Dic:1B>||\n"
```
Bouncner: Maybe test one chunk encoding that does not have an vector compression?

src/lib/operators/print.cpp
```diff
@@ -83,6 +86,15 @@ std::shared_ptr<const Table> Print::_on_execute() {
       continue;
     }
 
+    // print the encoding information
+    for (ColumnID column_id{0}; column_id < chunk->column_count(); ++column_id) {
+      const auto column_width = widths[column_id];
+      const auto& segment = chunk->get_segment(column_id);
+      _out << "|" << std::setw(column_width - 2) << _segment_type(segment) << std::setw(0);
```
Bouncner: Why `column_width - 2`?

src/bin/console/console.cpp
mrks: I'll grudgingly replace it, but tbh, I think that regular expressions are something that programmers should know, love, and use. It's not like this is an [e-mail validation regex](https://emailregex.com/) or uses non-regular stuff like lookarounds or non-greedy behavior.

src/lib/operators/print.cpp
mrks: Good catch. That was from a time where `<>` were already included.

src/bin/console/console.cpp
mrks: Changed it to `encoding_supports_data_type`. Let's have the rest of that discussion in the appropriate issue.

src/bin/console/console.cpp
mrks: After looking into the options with a loop, I found that the stackoverflow answers mistreat new lines in the input data. Fixing this and adding support for padding would blow this up. I strongly consider the regex version to be better, but agree to add an explanatory comment.

mrzzzrm: Is making this a nice little utility function harder than this:

```
words = boost::split_this_by_newline_or_whitespace(str)
lines = [str()]
while !words.empty():
  if !lines.back().empty() or len(lines.back() + words.front()) > 120:
    lines.append(str())
  lines.back.append(words.pop_front())

return boost::join_this_stuff_with_newlines(lines)
```
?

But in the end, of course, it's up to you.

mrks: If I am not mistaken, implementation would suffer from the same problems as the Stack Overflow solutions (mistreating multi-line input). To solve that, you would first have to split the input by lines, then by words.

mrks:
This should be good to go now

PR #1403, opened 2019-01-14
Benchmarks: Remove --verbose, remove JSON from output, shorten --help

### LINKED ISSUE #1392: Benchmarks: get rid of --verbose, don't print JSON to console
Does anyone have a need for the non-verbose mode? In the last months, I have *always* used `-v` to see the progress and I can't come up with a reason not to make that the only option. That would also make it easier for new users who might not know how helpful `-v` is.

mrks:
fixes #1392

### DIFF
--- a/src/lib/operators/table_scan.cpp
+++ b/src/lib/operators/table_scan.cpp
@@ -146,8 +146,6 @@ std::shared_ptr<const Table> TableScan::_on_execute() {
             }
           }
 
-          if (pos_list_in->references_single_chunk()) filtered_pos_list->guarantee_single_chunk();
-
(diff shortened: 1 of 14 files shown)


src/lib/operators/table_scan.cpp
```diff
@@ -146,8 +146,6 @@ std::shared_ptr<const Table> TableScan::_on_execute() {
             }
           }
 
-          if (pos_list_in->references_single_chunk()) filtered_pos_list->guarantee_single_chunk();
```
mrzzzrm: Intended for this PR?

mrks: Should have left a review comment for that. Yes, it is repeated from a few lines above.

PR #1408, opened 2019-01-14
Faster type erasure for iterators

mrks:
This reduces the compile time for the debug build as typed iterators are created once by `any_segment_iterable.cpp`, not every time `create_iterable_from_segment.hpp` is used.

You won't see any effect just with this, but it is blocking some bigger compile time reductions that I am working on.

(Also, it removes the pesky cqf warnings)

### DIFF
--- a/src/lib/storage/create_iterable_from_segment.hpp
+++ b/src/lib/storage/create_iterable_from_segment.hpp
@@ -27,45 +27,45 @@ class ReferenceSegmentIterable;
  * @{
  */
 
-template <typename T>
+template <typename T, bool EraseType = HYRISE_DEBUG>
 auto create_iterable_from_segment(const ValueSegment<T>& segment) {
-  if constexpr (HYRISE_DEBUG) {
+  if constexpr (EraseType) {
     return create_any_segment_iterable<T>(segment);
   } else {
     return ValueSegmentIterable<T>{segment};
   }
 }
 
-template <typename T>
+template <typename T, bool EraseType = HYRISE_DEBUG>
 auto create_iterable_from_segment(const DictionarySegment<T>& segment) {
-  if constexpr (HYRISE_DEBUG) {
+  if constexpr (EraseType) {
     return create_any_segment_iterable<T>(segment);
   } else {
     return DictionarySegmentIterable<T, pmr_vector<T>>{segment};
   }
 }
 
-template <typename T>
+template <typename T, bool EraseType = HYRISE_DEBUG>
 auto create_iterable_from_segment(const RunLengthSegment<T>& segment) {
-  if constexpr (HYRISE_DEBUG) {
+  if constexpr (EraseType) {
     return create_any_segment_iterable<T>(segment);
   } else {
     return RunLengthSegmentIterable<T>{segment};
   }
 }
 
-template <typename T>
+template <typename T, bool EraseType = HYRISE_DEBUG>
 auto create_iterable_from_segment(const FixedStringDictionarySegment<T>& segment) {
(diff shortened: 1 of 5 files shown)


src/lib/storage/segment_iterables/any_segment_iterable.cpp
```diff
@@ -11,10 +11,12 @@ AnySegmentIterable<T> CreateAnySegmentIterable<T>::create(const BaseSegment& bas
   auto any_segment_iterable = std::optional<AnySegmentIterable<T>>{};
 
   resolve_segment_type<T>(base_segment, [&](const auto& segment) {
-    const auto actual_iterable = create_iterable_from_segment<T>(segment);
+    const auto actual_iterable = create_iterable_from_segment<T, false>(segment);
     any_segment_iterable.emplace(erase_type_from_iterable(actual_iterable));
   });
 
+  DebugAssert(any_segment_iterable, "Could not create iterable");
```
mrzzzrm: I think we rely on resolve_segment_type calling the functor exactly once all over the code base and don't check whether it did anywhere else. I'd remove this assert

src/lib/storage/segment_iterables/any_segment_iterable.cpp
```diff
@@ -11,10 +11,12 @@ AnySegmentIterable<T> CreateAnySegmentIterable<T>::create(const BaseSegment& bas
   auto any_segment_iterable = std::optional<AnySegmentIterable<T>>{};
 
   resolve_segment_type<T>(base_segment, [&](const auto& segment) {
-    const auto actual_iterable = create_iterable_from_segment<T>(segment);
+    const auto actual_iterable = create_iterable_from_segment<T, false>(segment);
```
mrzzzrm: I'm being annoying if I point out that literal boolean parameters are not readable on the calling side?

src/lib/storage/create_iterable_from_segment.hpp
```diff
@@ -30,36 +27,56 @@ class ReferenceSegmentIterable;
...
-  return erase_type_from_iterable_if_debug(RunLengthSegmentIterable<T>{segment});
+  if constexpr (EraseType) {
+    return create_any_segment_iterable<T>(segment);
+  } else {
+    return RunLengthSegmentIterable<T>{segment};
+  }
 }
 
-template <typename T>
+template <typename T, bool EraseType = HYRISE_DEBUG>
```
mrzzzrm: There are two "types" being handled here, Data**Type** and Segment**Type**. As it is the latter we're erasing, I'd call this parameter "EraseSegmentType"

src/lib/storage/segment_iterables/any_segment_iterable.cpp
mrks: This is literally the only caller of that method that sets the second template parameter. That is unlikely to change, so adding another type just for that seems like overkill. Also it would complicate things on the receiver side, where instead of simply writing

```
template <typename T, bool EraseType = HYRISE_DEBUG>
auto create_iterable_from_segment(const ValueSegment<T>& segment) {
  if constexpr (EraseType) {
```

you would write (untested) 

```
template <typename T, EraseType = EraseType::Yes>
auto create_iterable_from_segment(const ValueSegment<T>& segment) {
  if constexpr (EraseType == EraseType::Yes && HYRISE_DEBUG) {
```

While I started to appreciate seeing `UseMvcc::Yes` instead of `true`, I think we should decide on a case-to-case basis if this pattern really improves readability.

mrzzzrm: kk

PR #1410, opened 2019-01-15
Add transaction context assert to forbid using the transaction context twice.

engelfa:


src/test/operators/delete_test.cpp
```diff
@@ -268,4 +268,26 @@ TEST_F(OperatorsDeleteTest, DeleteOwnInsert) {
...
+
+  auto validate1 = std::make_shared<Validate>(_gt);
+  validate1->set_transaction_context(t1_context);
+
+  validate1->execute();
+
+  auto delete_op = std::make_shared<Delete>(_table_name, validate1);
+  delete_op->set_transaction_context(t1_context);
+
+  delete_op->execute();
```
mrks: When using empty lines, please try to group things that belong together. This has a surprisingly big impact on readability The `execute` here belongs to the two lines above, while the `commit` is rightfully its own block

PR #1413, opened 2019-01-15
Replace default memory allocator with jemalloc

lanice:
First things first: I am not entirely sure if this is the best way to do it. I added jemalloc as submodule, and added a custom build for it in the third_party/CMakeLists.txt. This takes care of building jemalloc and providing the library target, it also timestamps it properly, so that we don't need to build it again every time.

The other option would have been to just require jemalloc to be installed on the system and link against it. I think the submodule way is more convenient, and allows for more control over updates etc., but I'm open to other thoughts.

Performance difference on Nemea with 224 cores (100 clients, sf 1):
```
+-----------+------------------+-------+------------------+-------+--------+---------------------------------+
| Benchmark | prev. iter/s     | runs  | new iter/s       | runs  | change | p-value (significant if <0.001) |
+-----------+------------------+-------+------------------+-------+--------+---------------------------------+
| TPC-H 1   | 12.4830493927002 | 749   | 13.1818952560425 | 791   | +6%    |                          0.0001 |
| TPC-H 2   | 5.94975709915161 | 357   | 5.71649599075317 | 343   | -4%    |                          0.4339 |
| TPC-H 3   | 68.2477645874023 | 4095  | 191.329956054688 | 10001 | +180%  |                          0.0000 |
| TPC-H 4   | 51.0626945495605 | 3064  | 92.7384872436523 | 5565  | +82%   |                          0.0000 |
| TPC-H 5   | 64.3960494995117 | 3864  | 151.011932373047 | 9062  | +135%  |                          0.0000 |
| TPC-H 6   | 449.550750732422 | 10002 | 465.972076416016 | 10003 | +4%    |                          0.0000 |
| TPC-H 7   | 42.7109870910645 | 2563  | 50.697566986084  | 3042  | +19%   |                          0.0000 |
| TPC-H 8   | 17.5886325836182 | 1056  | 18.9117641448975 | 1135  | +8%    |                          0.0001 |
| TPC-H 9   | 37.1128311157227 | 2227  | 67.3653793334961 | 4042  | +82%   |                          0.0000 |
| TPC-H 10  | 78.2251052856445 | 4694  | 94.1749114990234 | 5651  | +20%   |                          0.0000 |
| TPC-H 11  | 896.33203125     | 10000 | 898.320373535156 | 10010 | +0%    |                          0.0085 |
| TPC-H 12  | 299.461212158203 | 10000 | 301.11376953125  | 10002 | +1%    |                          0.4997 |
| TPC-H 13  | 22.6824207305908 | 1361  | 17.6833076477051 | 1061  | -22%   |                          0.0000 |
| TPC-H 14  | 555.671569824219 | 10000 | 560.26904296875  | 10003 | +1%    |                          0.0023 |
| TPC-H 16  | 44.9596176147461 | 2698  | 36.9782638549805 | 2219  | -18%   |                          0.0000 |
| TPC-H 18  | 34.1649436950684 | 2050  | 76.0229873657227 | 4562  | +123%  |                          0.0000 |
| TPC-H 19  | 360.028015136719 | 10000 | 318.787994384766 | 10000 | -11%   |                          0.0000 |
| TPC-H 22  | 407.951080322266 | 10003 | 532.314025878906 | 10000 | +30%   |                          0.0000 |
| average   |                  |       |                  |       | +35%   |                                 |
+-----------+------------------+-------+------------------+-------+--------+---------------------------------+
```

Singlethreaded performance difference:
```
+-----------+--------------------+------+--------------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s       | runs | new iter/s         | runs | change | p-value (significant if <0.001) |
+-----------+--------------------+------+--------------------+------+--------+---------------------------------+
| TPC-H 1   | 0.438022792339325  | 27   | 0.421905279159546  | 26   | -4%    |                          0.0003 |
| TPC-H 2   | 6.57991933822632   | 395  | 6.6384539604187    | 399  | +1%    |                          0.0296 |
| TPC-H 3   | 4.81768369674683   | 290  | 5.96744680404663   | 359  | +24%   |                          0.0000 |
| TPC-H 4   | 1.76354730129242   | 106  | 2.1845805644989    | 132  | +24%   |                          0.0000 |
| TPC-H 5   | 3.71122431755066   | 223  | 4.35764026641846   | 262  | +17%   |                          0.0000 |
| TPC-H 6   | 14.7273111343384   | 884  | 15.1507396697998   | 910  | +3%    |                          0.0000 |
| TPC-H 7   | 1.4710601568222    | 89   | 1.47819912433624   | 89   | +0%    |                          0.1544 |
| TPC-H 8   | 4.43694257736206   | 267  | 5.38061428070068   | 323  | +21%   |                          0.0000 |
| TPC-H 9   | 1.47725760936737   | 89   | 1.63468766212463   | 99   | +11%   |                          0.0000 |
| TPC-H 10  | 1.99470174312592   | 120  | 2.21957635879517   | 134  | +11%   |     (run time too short) 0.0000 |
| TPC-H 11  | 30.1635894775391   | 1810 | 29.6797027587891   | 1781 | -2%    |     (run time too short) 0.0000 |
| TPC-H 12  | 6.50658178329468   | 391  | 6.56643962860107   | 394  | +1%    |     (run time too short) 0.0843 |
| TPC-H 13  | 2.11807823181152   | 128  | 2.29881453514099   | 139  | +9%    |                          0.0000 |
| TPC-H 14  | 12.669361114502    | 761  | 14.2372140884399   | 855  | +12%   |                          0.0000 |
| TPC-H 15  | 9.9648380279541    | 598  | 8.46727657318115   | 509  | -15%   |     (run time too short) 0.0000 |
| TPC-H 16  | 6.04949331283569   | 363  | 7.20750379562378   | 433  | +19%   |     (run time too short) 0.0000 |
| TPC-H 17  | 0.0949677005410194 | 6    | 0.0906565114855766 | 6    | -5%    |        (not enough runs) 0.4396 |
| TPC-H 18  | 1.58406460285187   | 96   | 1.94895660877228   | 117  | +23%   |                          0.0000 |
| TPC-H 19  | 5.628098487854     | 338  | 5.7858681678772    | 348  | +3%    |                          0.0000 |
| TPC-H 20  | 0.0550514124333858 | 4    | 0.0544252768158913 | 4    | -1%    |        (not enough runs) 0.4724 |
| TPC-H 21  | 0.0980852991342545 | 6    | 0.106777161359787  | 7    | +9%    |        (not enough runs) 0.0552 |
| TPC-H 22  | 11.9763822555542   | 719  | 16.5382442474365   | 993  | +38%   |                          0.0000 |
| average   |                    |      |                    |      | +9%    |                                 |
+-----------+--------------------+------+--------------------+------+--------+---------------------------------+
```


mrks:
@Bensk1 - There was something with `autoconf` lately, which seems to be missing here. What was up with that?

Bensk1:
I believe we didn’t need it anymore because we removed a dependency or changed the way we are building it a while ago. It might be time to add autoconf again.

mrks:
Wasn't there something with it installing a second clang version?

lanice:
Another option could be to take the packaged release version, adding just that version instead of the subrepository. With that, they ship a configure script with no need for autoconf (see here: https://github.com/jemalloc/jemalloc/blob/dev/INSTALL.md)

lanice:
Replaced the submodule with jemalloc-5.1.0 release version. Question: The configure step generates a lot of output (see below), do we want to pipe that away somehow (logfile, /dev/null, ...), or are we okay with having that in the console?

```
checking for xsltproc... false
checking for gcc... no
checking for cc... cc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether cc accepts -g... yes
checking for cc option to accept ISO C89... none needed
checking whether compiler is cray... no
checking whether compiler supports -std=gnu11... yes
checking whether compiler supports -Wall... yes
checking whether compiler supports -Wshorten-64-to-32... no
checking whether compiler supports -Wsign-compare... yes
checking whether compiler supports -Wundef... yes
checking whether compiler supports -Wno-format-zero-length... yes
checking whether compiler supports -pipe... yes
checking whether compiler supports -g3... yes
checking how to run the C preprocessor... cc -E
checking for g++... g++
checking whether we are using the GNU C++ compiler... yes
checking whether g++ accepts -g... yes
checking whether g++ supports C++14 features by default... yes
checking whether compiler supports -Wall... yes
checking whether compiler supports -g3... yes
checking whether libstdc++ linkage is compilable... yes
checking for grep that handles long lines and -e... /bin/grep
checking for egrep... /bin/grep -E
checking for ANSI C header files... yes
checking for sys/types.h... yes
checking for sys/stat.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for memory.h... yes
checking for strings.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for unistd.h... yes
checking whether byte ordering is bigendian... no
checking size of void *... 8
checking size of int... 4
checking size of long... 8
checking size of long long... 8
checking size of intmax_t... 8
checking build system type... x86_64-pc-linux-gnu
checking host system type... x86_64-pc-linux-gnu
checking whether pause instruction is compilable... yes
checking number of significant virtual address bits... 48
checking for ar... ar
checking for nm... nm
checking for gawk... no
checking for mawk... mawk
checking malloc.h usability... yes
checking malloc.h presence... yes
checking for malloc.h... yes
checking whether malloc_usable_size definition can use const argument... no
checking for library containing log... -lm
checking whether __attribute__ syntax is compilable... yes
checking whether compiler supports -fvisibility=hidden... yes
checking whether compiler supports -fvisibility=hidden... yes
checking whether compiler supports -Werror... yes
checking whether compiler supports -herror_on_warning... no
checking whether tls_model attribute is compilable... yes
checking whether compiler supports -Werror... yes
checking whether compiler supports -herror_on_warning... no
checking whether alloc_size attribute is compilable... yes
checking whether compiler supports -Werror... yes
checking whether compiler supports -herror_on_warning... no
checking whether format(gnu_printf, ...) attribute is compilable... yes
checking whether compiler supports -Werror... yes
checking whether compiler supports -herror_on_warning... no
checking whether format(printf, ...) attribute is compilable... yes
checking for a BSD-compatible install... /usr/bin/install -c
checking for ranlib... ranlib
checking for ld... /usr/bin/ld
checking for autoconf... /usr/bin/autoconf
checking for memalign... yes
checking for valloc... yes
checking for __libc_calloc... yes
checking for __libc_free... yes
checking for __libc_malloc... yes
checking for __libc_memalign... yes
checking for __libc_realloc... yes
checking for __libc_valloc... yes
checking for __posix_memalign... no
checking whether compiler supports -O3... yes
checking whether compiler supports -O3... yes
checking whether compiler supports -funroll-loops... yes
checking configured backtracing method... N/A
checking for sbrk... yes
checking whether utrace(2) is compilable... no
checking whether a program using __builtin_unreachable is compilable... yes
checking whether a program using __builtin_ffsl is compilable... yes
checking LG_PAGE... 12
checking pthread.h usability... yes
checking pthread.h presence... yes
checking for pthread.h... yes
checking for pthread_create in -lpthread... yes
checking dlfcn.h usability... yes
checking dlfcn.h presence... yes
checking for dlfcn.h... yes
checking for dlsym... no
checking for dlsym in -ldl... yes
checking whether pthread_atfork(3) is compilable... yes
checking whether pthread_setname_np(3) is compilable... yes
checking for library containing clock_gettime... none required
checking whether clock_gettime(CLOCK_MONOTONIC_COARSE, ...) is compilable... yes
checking whether clock_gettime(CLOCK_MONOTONIC, ...) is compilable... yes
checking whether mach_absolute_time() is compilable... no
checking whether compiler supports -Werror... yes
checking whether syscall(2) is compilable... yes
checking for secure_getenv... yes
checking for sched_getcpu... yes
checking for sched_setaffinity... yes
checking for issetugid... no
checking for _malloc_thread_cleanup... no
checking for _pthread_mutex_init_calloc_cb... no
checking for TLS... yes
checking whether C11 atomics is compilable... yes
checking whether GCC __atomic atomics is compilable... yes
checking whether GCC __sync atomics is compilable... yes
checking whether Darwin OSAtomic*() is compilable... no
checking whether madvise(2) is compilable... yes
checking whether madvise(..., MADV_FREE) is compilable... yes
checking whether madvise(..., MADV_DONTNEED) is compilable... yes
checking whether madvise(..., MADV_DO[NT]DUMP) is compilable... yes
checking whether madvise(..., MADV_[NO]HUGEPAGE) is compilable... yes
checking whether to force 32-bit __sync_{add,sub}_and_fetch()... no
checking whether to force 64-bit __sync_{add,sub}_and_fetch()... no
checking for __builtin_clz... yes
checking whether Darwin os_unfair_lock_*() is compilable... no
checking whether Darwin OSSpin*() is compilable... no
checking whether glibc malloc hook is compilable... yes
checking whether glibc memalign hook is compilable... yes
checking whether pthreads adaptive mutexes is compilable... yes
checking whether compiler supports -D_GNU_SOURCE... yes
checking whether compiler supports -Werror... yes
checking whether compiler supports -herror_on_warning... no
checking whether strerror_r returns char with gnu source is compilable... yes
checking for stdbool.h that conforms to C99... yes
checking for _Bool... yes
configure: creating ./config.status
config.status: creating Makefile
config.status: creating jemalloc.pc
config.status: creating doc/html.xsl
config.status: creating doc/manpages.xsl
config.status: creating doc/jemalloc.xml
config.status: creating include/jemalloc/jemalloc_macros.h
config.status: creating include/jemalloc/jemalloc_protos.h
config.status: creating include/jemalloc/jemalloc_typedefs.h
config.status: creating include/jemalloc/internal/jemalloc_preamble.h
config.status: creating test/test.sh
config.status: creating test/include/test/jemalloc_test.h
config.status: creating config.stamp
config.status: creating bin/jemalloc-config
config.status: creating bin/jemalloc.sh
config.status: creating bin/jeprof
config.status: creating include/jemalloc/jemalloc_defs.h
config.status: include/jemalloc/jemalloc_defs.h is unchanged
config.status: creating include/jemalloc/internal/jemalloc_internal_defs.h
config.status: include/jemalloc/internal/jemalloc_internal_defs.h is unchanged
config.status: creating test/include/test/jemalloc_test_defs.h
config.status: test/include/test/jemalloc_test_defs.h is unchanged
config.status: executing include/jemalloc/internal/public_symbols.txt commands
config.status: executing include/jemalloc/internal/private_symbols.awk commands
config.status: executing include/jemalloc/internal/private_symbols_jet.awk commands
config.status: executing include/jemalloc/internal/public_namespace.h commands
config.status: executing include/jemalloc/internal/public_unnamespace.h commands
config.status: executing include/jemalloc/internal/size_classes.h commands
config.status: executing include/jemalloc/jemalloc_protos_jet.h commands
config.status: executing include/jemalloc/jemalloc_rename.h commands
config.status: executing include/jemalloc/jemalloc_mangle.h commands
config.status: executing include/jemalloc/jemalloc_mangle_jet.h commands
config.status: executing include/jemalloc/jemalloc.h commands
===============================================================================
jemalloc version   : 5.1.0-0-g61efbda7098de6fe64c362d309824864308c36d4
library revision   : 2

CONFIG             : 
CC                 : cc
CONFIGURE_CFLAGS   : -std=gnu11 -Wall -Wsign-compare -Wundef -Wno-format-zero-length -pipe -g3 -fvisibility=hidden -O3 -funroll-loops
SPECIFIED_CFLAGS   : 
EXTRA_CFLAGS       : 
CPPFLAGS           : -D_GNU_SOURCE -D_REENTRANT
CXX                : g++
CONFIGURE_CXXFLAGS : -Wall -g3 -fvisibility=hidden -O3
SPECIFIED_CXXFLAGS : 
EXTRA_CXXFLAGS     : 
LDFLAGS            : 
EXTRA_LDFLAGS      : 
DSO_LDFLAGS        : -shared -Wl,-soname,$(@F)
LIBS               : -lm -lstdc++ -lpthread -ldl
RPATH_EXTRA        : 

XSLTPROC           : false
XSLROOT            : 

PREFIX             : /usr/local
BINDIR             : /usr/local/bin
DATADIR            : /usr/local/share
INCLUDEDIR         : /usr/local/include
LIBDIR             : /usr/local/lib
MANDIR             : /usr/local/share/man

srcroot            : 
abs_srcroot        : /home/Leander.Neiss/hyrise/third_party/jemalloc-5.1.0/
objroot            : 
abs_objroot        : /home/Leander.Neiss/hyrise/third_party/jemalloc-5.1.0/

JEMALLOC_PREFIX    : 
JEMALLOC_PRIVATE_NAMESPACE
                   : je_
install_suffix     : 
malloc_conf        : 
autogen            : 0
debug              : 0
stats              : 1
prof               : 0
prof-libunwind     : 0
prof-libgcc        : 0
prof-gcc           : 0
fill               : 1
utrace             : 0
xmalloc            : 0
log                : 0
lazy_lock          : 0
cache-oblivious    : 1
cxx                : 1
===============================================================================
```

mrks:
https://unix.stackexchange.com/questions/41381/suppress-output-unless-non-zero-exit-code ?

lanice:
> https://unix.stackexchange.com/questions/41381/suppress-output-unless-non-zero-exit-code ?

I'll look into that.

There is still something missing here when adding jemalloc to our project, marking this pull request WIP for now.

lanice:
Added the suppressed output stuff. Replaced `echo` with `printf`, because `echo` is a bit dorky and printed everything with no linebreaks, and [apparently](https://stackoverflow.com/a/11193533) printf is more suitable for stuff like that anyway.

Also, I changed the cmake `externalproject_add` command to download the jemalloc package instead of putting it into the repo.

Now the build server complains about not finding a C compiler...:
```
[ 15%] Performing configure step for 'libjemalloc'
checking for xsltproc... false
checking for gcc... gcc
checking whether the C compiler works... no
sed: can't read conftest.c: No such file or directory
configure: error: in `/var/lib/jenkins/workspace/hyrise_hyrise_PR-1413/third_party/jemalloc/src/libjemalloc':
configure: error: C compiler cannot create executables
```

mrzzzrm:
Can we just ship with jemalloc unpacked into third_party/ (as we do with json, e.g.)? Or is it a sizable project? I'd rather avoid the download during configure if we can.

lanice:
> Can we just ship with jemalloc unpacked into third_party/ (as we do with json, e.g.)? 

We could, but only if we modify it / force add individual files. It comes with a `.gitignore`, which excludes some files that are needed for the config step. I don't really liked that, because that would mean if we ever update it to a new version we would have to tinker with the `.gitignore` again, etc.
The next best option if we really don't want to download is to just ship the packaged thing, and then extract it only instead of download+extract. I went with the download step because it looks like that is a quite common approach with [externalproject_add](https://cmake.org/cmake/help/latest/module/ExternalProject.html).

lanice:
Unfortunately it is not only that. If we delete the gitignore, then yes, the config files are gonna be added. But in turn, after building, hundreds of other generated files float around in the directory and polluted the unstaged git area.

third_party/CMakeLists.txt
```diff
@@ -150,4 +150,20 @@ target_link_libraries(pqxx PUBLIC ${PQ_LIBRARY})
...
+    URL https://github.com/jemalloc/jemalloc/releases/download/5.1.0/jemalloc-5.1.0.tar.bz2
+    # Executing the configure script with suppressed output, only printing it in case of non-zero return
+    CONFIGURE_COMMAND bash -c "cd <SOURCE_DIR> && output=`./configure 2>&1` || printf \"$output\\n\""
+    BUILD_COMMAND ${CMAKE_BUILD_TOOL}
+    BUILD_IN_SOURCE 1
+    STEP_TARGETS build
+)
+add_library(jemalloc SHARED IMPORTED GLOBAL)
+add_dependencies(jemalloc libjemalloc-build)
+set_target_properties(jemalloc PROPERTIES IMPORTED_LOCATION ${CMAKE_CURRENT_SOURCE_DIR}/jemalloc/src/libjemalloc/lib/libjemalloc.so)
```
mrzzzrm: Just so I understand this correctly: Building jemalloc as a static library and linking it into hyrise statically would not work, because the weak symbol replacement stuff only works with dynamic libs? (@mrks?).

If that's not the case would it be trivially possible to build jemalloc as a static lib, for consistency with the other third party libs?

mrzzzrm:
Another option (sorry!): Include an unpacked jemalloc in the master and write a CMake target for it, as we do for pqxx, cqf, sqlparser, benchmark. jemalloc is just a handful of C files, after all.

third_party/CMakeLists.txt
mrks: Not even sure if it generally does not work, but we might lose compatibility with other tools that expect malloc to be weak. Think about profilers, valgrind, ...

lanice:
Maybe...? Based on the configure and Makefile, and on the [attempt](https://github.com/jemalloc/jemalloc/issues/303) of porting jemalloc to CMake (only for Windows, discontinued), it looks like more than that though, at least to me. If it really is that simple, and you know how to do it, feel free to go for it.

Otherwise, these are the options we have so far:

1. Integrate jemalloc as submodule. This would require autoconf, which was the original reason we switched to the packaged release version.
2. Use the packaged release version, let CMake handle the download, extration, configure, and build part, which seems to be the standard way with the [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) target.
3. Use the packaged release version, but add it to our repository ourselves, manually extract, and then proceed as in (2).

I think (1) or (2) would be the best. I don't see why we would want to add 500 KB to the repository with no real reason, but if the majority hates the download part of (2), I can live with that.

mrks:
@Bensk1 This questions looks unanswered:

> Wasn't there something with it installing a second clang version?

Bensk1:
Seems like it.

lanice:
Alright, then we're back where we started, and need to install autoconf on our CI server.

mrks:
If you update install.sh, that should happen automatically. Please also add it to `DEPENDENCIES.md` and the `Dockerfile`

mrks:
```
configure: error: cannot compute suffix of executables: cannot compile and link

See `config.log' for more details

Error 0 in ./configure

```

mrks:
Looks like some dependencies are not explicitly set. Did you run the docker image locally? It might make a difference on bigger machines where some steps complete earlier and expect results that have not yet been produced.

mrks:
`./autogen.sh: 13: ./autogen.sh: ./configure: Text file busy`

For CI, we configure (cmake) and build (make) multiple configurations of Hyrise at the same time. Any chance that all of these fight to be the first to build jemalloc?

lanice:
> Any chance that all of these fight to be the first to build jemalloc?

Yeah, that could be the reason, especially the new error message (`Text file busy`) looks like something like that. Could be that the configure file can't be used in parallel.

I'm wondering, for our other third-party libs, where are those built? In their own directory, or in the according build directory (clang-debug, clang-release, etc...)?

lanice:
> Did you run the docker image locally?

Both locally and on pella. But I always only ran one build at a time, not multiple in parallel, which (as suggested above) could be the issue here.

mrks:
> I'm wondering, for our other third-party libs, where are those built? In their own directory, or in the according build directory (clang-debug, clang-release, etc...)?

In `./build/third-party/`. Is this different here?

lanice:
I see. Yes, jemalloc is built within its directory, `third-party/jemalloc`. I remember there was a reason why I set the option `BUILD_IN_SOURCE 1`, but I think this could be the issue. I will look into that.

lanice:
The actual target is called `libjemalloc-build`, if you want to `make` just that in the first step, instead of a full build. (Is that necessary? I thought I covered it with the check)

mrks:
Two builds that perform that "file exists" check at the same time might still decide that they both need to build jemalloc.

mrks:
Since you're still awake - do those changes make sense to you? If yes, I'll merge once this passes.

PR #1414, opened 2019-01-15
Enable forced colored output

Bouncner:
Flag to enable "forced colored output". Not necessary for make, but for Ninja.
More information can be found here: https://medium.com/@alasher/colored-c-compiler-output-with-ninja-clang-gcc-10bfe7f2b949

Usage:
`cmake -DFORCE_COLORED_OUTPUT=ON -G Ninja ..`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,10 +66,8 @@ if(NOT CMAKE_BUILD_TYPE IN_LIST BUILD_TYPES)
     message(FATAL_ERROR "Unknown Build Type: ${CMAKE_BUILD_TYPE}")
 endif()
 
-# Flag to enforce colored output. When compilers recognize output to terminals, recent version automatically
-# use colored output. However, when using Ninja, the output is not directly sent to the terminal. With this
-# flag enabled, also Ninja builds have colored outputs.
-if (${FORCE_COLORED_OUTPUT})
+# Enforce colored output when generator is Ninja.
+if ("${CMAKE_GENERATOR}" STREQUAL "Ninja")
     if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
        add_compile_options(-fdiagnostics-color=always)
     elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,17 @@ if(NOT CMAKE_BUILD_TYPE IN_LIST BUILD_TYPES)
     message(FATAL_ERROR "Unknown Build Type: ${CMAKE_BUILD_TYPE}")
 endif()
 
+# Flag to enforce colored output. When compiler recognize output to terminals, recent version automatically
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -66,6 +66,17 @@ if(NOT CMAKE_BUILD_TYPE IN_LIST BUILD_TYPES)
     message(FATAL_ERROR "Unknown Build Type: ${CMAKE_BUILD_TYPE}")
 endif()
 
+# Flag to enforce colored output. When compiler recognize output to terminals, recent version automatically
+# set colored output. However, when using Ninja, the output is not directly the terminal. With this flag enabled,
```
Bensk1: < the output is not directly the terminal

This is easily misunderstood. Maybe something like "directly sent to the terminal"?

CMakeLists.txt
```diff
@@ -66,6 +66,17 @@ if(NOT CMAKE_BUILD_TYPE IN_LIST BUILD_TYPES)
...
+# also Ninja build will have colored outputs.
```
Bensk1: Ninja buildS?

CMakeLists.txt
```diff
@@ -66,6 +66,17 @@ if(NOT CMAKE_BUILD_TYPE IN_LIST BUILD_TYPES)
...
+if (${FORCE_COLORED_OUTPUT})
```
Bensk1: Most (all?) other flags start with `ENABLE`. But `FORCE` makes still more sense here, right?

CMakeLists.txt
```diff
@@ -66,6 +66,17 @@ if(NOT CMAKE_BUILD_TYPE IN_LIST BUILD_TYPES)
...
+    if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+       add_compile_options (-fdiagnostics-color=always)
```
Bensk1: There is no space between `add_compile_options` and `(` in all other cases.

Bensk1:
Could we also detect the generator specified in the `cmake` command and add the compile options accordingly?

Bouncner:
@Bensk1 : I am not sure if there are any problems with having colored output enabled per default (e.g., I don't know what the compile script does etc.). 

@mrks : In the long run, probably everybody will be using Ninja (at least I like it very much so far). Having to set this flag each and every time is cumbersome and guaranteed to be forgotten. At least I think, it would be of great help. And with @Bensk1 idea of checking for the generator (`CMAKE_GENERATOR`) this could be enabled per default and just ease everybody's life. :)

CMakeLists.txt
Bouncner: With your idea of checking of the generator, we might get rid of the flag altogether.

Bouncner:
Maybe something to discuss tomorrow in the fortnightly?

Bensk1:
Let’s do that.

Bouncner:
We had two options:
- use a parameter to enforce coloured outputs, but that's not really much easier than using `CXXFLAGS=-fcolor-diagnostics`
- enable colors when Ninja is used,  but polutes the output when the output is piped to an output file

PR #1417, opened 2019-01-16
Reduce compile time of ExpressionEvaluator

mrks:
This gets rid of a couple of template instantiations.

Previously (my laptop, debug): 0m45.742s
Now: 0m28.267s

Previous (nemea, release): 1m35.498s
Now: 0m45.300s

### DIFF
--- a/src/lib/expression/evaluation/expression_evaluator.cpp
+++ b/src/lib/expression/evaluation/expression_evaluator.cpp
@@ -54,8 +54,10 @@ void resolve_binary_predicate_evaluator(const PredicateCondition predicate_condi
     case PredicateCondition::NotEquals:         functor(boost::hana::type<NotEqualsEvaluator>{});         break;
     case PredicateCondition::LessThan:          functor(boost::hana::type<LessThanEvaluator >{});         break;
     case PredicateCondition::LessThanEquals:    functor(boost::hana::type<LessThanEqualsEvaluator>{});    break;
-    case PredicateCondition::GreaterThan:       functor(boost::hana::type<GreaterThanEvaluator>{});       break;
-    case PredicateCondition::GreaterThanEquals: functor(boost::hana::type<GreaterThanEqualsEvaluator>{}); break;
+    case PredicateCondition::GreaterThan:       
+    case PredicateCondition::GreaterThanEquals: 
+      Fail("PredicateCondition should have been flipped");
+      break;
 
     default:
       Fail("PredicateCondition should be handled in different function");
@@ -220,6 +222,7 @@ std::shared_ptr<ExpressionResult<Result>> ExpressionEvaluator::_evaluate_arithme
   // clang-format on
   Fail("GCC thinks this is reachable");
 }
+
 template <>
 std::shared_ptr<ExpressionResult<ExpressionEvaluator::Bool>>
 ExpressionEvaluator::_evaluate_binary_predicate_expression<ExpressionEvaluator::Bool>(
@@ -500,11 +503,11 @@ ExpressionEvaluator::_evaluate_in_expression<ExpressionEvaluator::Bool>(const In
     });
 
   } else {
-    /**
+    *
      * `<expression> IN <anything_but_list_or_select>` is not legal SQL, but on expression level we have to support
      * it, since `<anything_but_list_or_select>` might be a column holding the result of a subselect.
      * To accomplish this, we simply rewrite the expression to `<expression> IN LIST(<anything_but_list_or_select>)`.
-     */
+     
 
     return _evaluate_in_expression<ExpressionEvaluator::Bool>(*std::make_shared<InExpression>(
         in_expression.predicate_condition, in_expression.value(), list_(in_expression.set())));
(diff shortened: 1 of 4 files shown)


src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -54,8 +54,10 @@ void resolve_binary_predicate_evaluator(const PredicateCondition predicate_condi
     case PredicateCondition::NotEquals:         functor(boost::hana::type<NotEqualsEvaluator>{});         break;
     case PredicateCondition::LessThan:          functor(boost::hana::type<LessThanEvaluator >{});         break;
     case PredicateCondition::LessThanEquals:    functor(boost::hana::type<LessThanEqualsEvaluator>{});    break;
-    case PredicateCondition::GreaterThan:       functor(boost::hana::type<GreaterThanEvaluator>{});       break;
-    case PredicateCondition::GreaterThanEquals: functor(boost::hana::type<GreaterThanEqualsEvaluator>{}); break;
```
mrzzzrm: You should be able to remove GreaterThanEqualsEvaluator and GreaterThanEvaluator then, right?

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -583,38 +591,37 @@ std::shared_ptr<ExpressionResult<Result>> ExpressionEvaluator::_evaluate_case_ex
...
-            nulls[chunk_offset] = then_result.is_null(chunk_offset);
-          } else {
-            values[chunk_offset] = to_value<Result>(else_result.value(chunk_offset));
-            nulls[chunk_offset] = else_result.is_null(chunk_offset);
+        if constexpr (CaseEvaluator::supports_v<Result, ThenResultType, ElseResultType>) {
+          const auto result_size = _result_size(when->size(), then_result.size(), else_result.size());
+          values.resize(result_size);
+          nulls.resize(result_size);
+
+          for (auto chunk_offset = ChunkOffset{0};
```
mrzzzrm: unintended linebreak?

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -1152,18 +1159,19 @@ std::shared_ptr<ExpressionResult<Result>> ExpressionEvaluator::_evaluate_logical
 template <typename Result, typename Functor>
 std::shared_ptr<ExpressionResult<Result>> ExpressionEvaluator::_evaluate_binary_with_default_null_logic(
     const AbstractExpression& left_expression, const AbstractExpression& right_expression) {
-  auto result = std::shared_ptr<ExpressionResult<Result>>{};
+  std::vector<Result> values;
+  std::vector<bool> nulls;
```
mrzzzrm: Just curious: Moving this out of the lambda really makes a difference, ct-wise?

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -1177,46 +1185,70 @@ std::shared_ptr<ExpressionResult<Result>> ExpressionEvaluator::_evaluate_binary_
...
     if constexpr (Functor::template supports<Result, LeftDataType, RightDataType>::value) {
-      const auto result_row_count = _result_size(left.size(), right.size());
-
-      std::vector<bool> nulls(result_row_count);
-      std::vector<Result> values(result_row_count);
+      const auto result_size = _result_size(left.size(), right.size());
+      values.resize(result_size);
+      nulls.resize(result_size);
+
+      auto get_null = [](const auto& result, ChunkOffset row_idx) -> bool {
```
mrzzzrm: How big is the perf-impact, compared to the compile time saving? I think I instantiated all Literal/Segment/Nullable combinations here, intentionally.

src/lib/expression/evaluation/expression_evaluator.cpp
mrks: Yes. Wasn't aware of those.

src/lib/expression/evaluation/expression_evaluator.cpp
mrks: There is no need to turn of the formatter there any more, anyway.

mrks: There is no need to turn of the formatter there any more, anyway.

src/lib/expression/evaluation/expression_evaluator.cpp
mrks: I didn't measure this specifically. Generally, when trying to keep the size of the lambda short in terms of instructions, I think it helps to also reduce the number of source code lines.

src/lib/expression/evaluation/expression_evaluator.cpp
mrks: I'll measure it once nemea is idle, but my understanding is that we still initiate every combination here. It's just that we don't work our ways through layers and layers of templates to reach that.

mrzzzrm: > but my understanding is that we still initiate every combination here

you're now checking for the nullability of an operand for every row, instead of instantiating this for every combination of nullable/non-nullable. TPC-H will not show you whether this had an impact, as this particular path is only taken for OR and ANDs and TPC-H is pretty light on those.

mrzzzrm: > It's just that we don't work our ways through layers and layers of templates to reach that.

There is one layer that you removed here: That were the ExpressionResults are resolved to Views.

My issue is that this is more verbose now, and potentially slower, though likely faster to compile. But this is a path only taken for AND and OR, so I am not sure how big the compile time win really is to justify the verbosity and deviation from the intended way operations should be evaluated in the ExpressionEvaluator (which is: Resolve operands to views and implement you operation not caring about whether an operand is a Literal, a Series, Nullable or Not Nullable.)

mrks: The performance difference is given in the PR description.

Not sure if I can follow though. How is this different from `_evaluate_binary_with_default_null_logic`?

mrks: Aaah, ok, got it.

mrks: Ok, can't come up with a good solution for now. Let's just look at the other things.

PR #1418, opened 2019-01-16
Reduce compile time of ColumnVsColumnTableScanImpl

mrks:
If the left and the right segment and/or type are not the same, we erase the types even for the release build. For example, ValueSegment<int> == ValueSegment<float> will be erased. So will ValueSegment<int> == DictionarySegment<int>. ReferenceSegments do not need to be handled differently because we expect a table to either have only ReferenceSegments or non-ReferenceSegments.

We use type erasure here because we currently do not compare, e.g., a ValueSegment with a DictionarySegment, and we don't want the compiler to spend time instantiating unused templates. Whenever the types of the iterators is removed, we also erase the comparator lambda by wrapping it into an std::function. All of this brought the compile time down by a factor of 5. This is only relevant for the release build - in the debug build, iterators are erased anyway.

Also, compiling that file with GCC now only takes 3,3 GB of memory instead of 9,0 GB. Isn't that great?

Old (nemea, release):
	Elapsed (wall clock) time (h:mm:ss or m:ss): 10:21.41
	Maximum resident set size (kbytes): 9000508

New:
	Elapsed (wall clock) time (h:mm:ss or m:ss): 2:04.83
	Maximum resident set size (kbytes): 3307204

When running the TPC-H, no performance warning is printed, so there should be no performance impact.

### DIFF
--- a/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
@@ -42,7 +42,8 @@ std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::scan_chunk(ChunkID chunk_i
   // We use type erasure here because we currently do not compare, e.g., a ValueSegment with a DictionarySegment,
   // and we don't want the compiler to spend time instantiating unused templates. Whenever the types of the iterators
   // is removed, we also erase the comparator lambda by wrapping it into an std::function. All of this brought the
-  // compile time down by a factor of 5.
+  // compile time down by a factor of 5. This is only relevant for the release build - in the debug build, iterators
+  // are erased anyway.
 
   resolve_data_and_segment_type(left_segment, [&](auto left_type, auto& left_typed_segment) {
     resolve_data_and_segment_type(right_segment, [&](auto right_type, auto& right_typed_segment) {
--- a/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
@@ -32,81 +32,95 @@ std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::scan_chunk(ChunkID chunk_i
   const auto& left_segment = *chunk->get_segment(_left_column_id);
   const auto& right_segment = *chunk->get_segment(_right_column_id);
 
+  std::shared_ptr<PosList> result;
+
   // If the left and the right segment and/or type are not the same, we erase the types even for the release build.
-  // This because we have not worked with those combinations and we don't want the templates to be instantiated.
-  const bool either_is_reference_segment =
-      dynamic_cast<const ReferenceSegment*>(&left_segment) || dynamic_cast<const ReferenceSegment*>(&right_segment);
-  // if (typeid(left_segment) == typeid(right_segment) && either_is_reference_segment) {
-  //   return _typed_scan_chunk<SegmentIterationTypeErasure::OnlyInDebug>(chunk_id);
-  // } else {
-    return _typed_scan_chunk<SegmentIterationTypeErasure::Always>(chunk_id);
-    if (!HYRISE_DEBUG)
-      PerformanceWarning("Using non-specialized (i.e., type-erased) version of ColumnVsColumnTableScan");
-  // }
+  // For example, ValueSegment<int> == ValueSegment<float> will be erased. So will ValueSegment<int> ==
+  // DictionarySegment<int>. ReferenceSegments do not need to be handled differently because we expect a table to
+  // either have only ReferenceSegments or non-ReferenceSegments.
+  //
+  // We use type erasure here because we currently do not compare, e.g., a ValueSegment with a DictionarySegment,
+  // and we don't want the compiler to spend time instantiating unused templates. Whenever the types of the iterators
+  // is removed, we also erase the comparator lambda by wrapping it into an std::function. All of this brought the
+  // compile time down by a factor of 5.
+
+  resolve_data_and_segment_type(left_segment, [&](auto left_type, auto& left_typed_segment) {
+    resolve_data_and_segment_type(right_segment, [&](auto right_type, auto& right_typed_segment) {
+      using LeftType = typename decltype(left_type)::type;
+      using RightType = typename decltype(right_type)::type;
+
+      if constexpr (!HYRISE_DEBUG && std::is_same_v<decltype(left_typed_segment), decltype(right_typed_segment)>) {
+        // Same segment types - do not erase types
+        result = _typed_scan_chunk<SegmentIterationTypeErasure::OnlyInDebug>(
+            chunk_id, create_iterable_from_segment<LeftType>(left_typed_segment),
+            create_iterable_from_segment<RightType>(right_typed_segment));
+      } else {
+        PerformanceWarning("ColumnVsColumnTableScan using type-erased iterators");
+        result = _typed_scan_chunk<SegmentIterationTypeErasure::Always>(
+            chunk_id, create_any_segment_iterable<LeftType>(left_segment),
+            create_any_segment_iterable<RightType>(right_segment));
+      }
+    });
+  });
+
+  return result;
 }
 
-template <SegmentIterationTypeErasure type_erasure>
-std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::_typed_scan_chunk(ChunkID chunk_id) const {
+template <SegmentIterationTypeErasure type_erasure, typename LeftIterable, typename RightIterable>
+std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::_typed_scan_chunk(ChunkID chunk_id,
+                                                                        const LeftIterable& left_iterable,
+                                                                        const RightIterable& right_iterable) const {
   const auto chunk = _in_table->get_chunk(chunk_id);
-
-  const auto left_segment = chunk->get_segment(_left_column_id);
-  const auto right_segment = chunk->get_segment(_right_column_id);
 
   auto matches_out = std::make_shared<PosList>();
 
-  auto erase_comparator_type = [](auto comparator, const auto& left_it, const auto& right_it) {
-    if constexpr (type_erasure == SegmentIterationTypeErasure::Always) {
-      return comparator;
-    } else {
-      return std::function<bool(const AbstractSegmentPosition<std::decay_t<decltype(left_it->value())>>&,
-                                const AbstractSegmentPosition<std::decay_t<decltype(right_it->value())>>&)>{comparator};
+  using LeftType = typename LeftIterable::ValueType;
+  using RightType = typename RightIterable::ValueType;
+
+  // C++ cannot compare strings and non-strings out of the box:
+  if constexpr (std::is_same_v<LeftType, std::string> == std::is_same_v<RightType, std::string>) {
+    bool flipped = false;
+    auto condition = _predicate_condition;
+    if (condition == PredicateCondition::GreaterThan || condition == PredicateCondition::GreaterThanEquals) {
+      condition = flip_predicate_condition(condition);
+      flipped = true;
     }
-  };
 
-  segment_with_iterators<ResolveDataTypeTag, type_erasure>(*left_segment, [&](auto left_it,
-                                                                              [[maybe_unused]] const auto left_end) {
-    segment_with_iterators<ResolveDataTypeTag, type_erasure>(
-        *right_segment, [&](auto right_it, [[maybe_unused]] const auto right_end) {
-          using LeftType = typename decltype(left_it)::ValueType;
-          using RightType = typename decltype(right_it)::ValueType;
+    auto erase_comparator_type = [](auto comparator, const auto& it1, const auto& it2) {
+      if constexpr (type_erasure == SegmentIterationTypeErasure::Always) {
+        return comparator;
+      } else {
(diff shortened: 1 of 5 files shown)


Bensk1:
Does this mean that these changes do not have any performance implications right now because there are no scans in the TPC-H where we compare either different data types or different encodings? If we would have such a comparison there would be performance implications?

mrzzzrm:
Before I review: The same compile-time optimization can be applied to the JNL as well, no?

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -28,66 +29,99 @@ std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsC
...
+    if (condition == PredicateCondition::GreaterThan || condition == PredicateCondition::GreaterThanEquals) {
+      condition = flip_predicate_condition(condition);
+      flipped = true;
+    }
+
+    auto erase_comparator_type = [](auto comparator, const auto& it1, const auto& it2) {
+      if constexpr (type_erasure == SegmentIterationTypeErasure::Always) {
+        return comparator;
+      } else {
+        return std::function<bool(const AbstractSegmentPosition<std::decay_t<decltype(it1->value())>>&,
```
Bensk1: Does wrapping it into std::function have any performance impact?

mrks:
> Before I review: The same compile-time optimization can be applied to the JNL as well, no?

You already erased the outer iterator there anyway.

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrks: Yes, because it type-erases the lambda implementation. This is one level of indirection, but more importantly, breaks the basic block and hinders inlining.

mrks:
Release. Here's debug on my 2012 Macbook:

```
⌂34% [:~/Projekte/Opossum/Git/build] $ time ninja src/lib/CMakeFiles/hyrise.dir/operators/join_nested_loop.cpp.o
[1/1] Building CXX object src/lib/CMakeFiles/hyrise.dir/operators/join_nested_loop.cpp.o

real	0m10.557s
user	0m10.087s
sys	0m0.430s
```

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -28,66 +29,99 @@ std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsC
 
 std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::scan_chunk(ChunkID chunk_id) const {
   const auto chunk = _in_table->get_chunk(chunk_id);
+  const auto& left_segment = *chunk->get_segment(_left_column_id);
+  const auto& right_segment = *chunk->get_segment(_right_column_id);
+
+  std::shared_ptr<PosList> result;
+
+  // If the left and the right segment and/or type are not the same, we erase the types even for the release build.
```
mrzzzrm: I'd prefix this comment with 
```
// Compile time optimization
// ....
```

mrzzzrm: Also, which types? Iterator type? DataType? ComparatorType?

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -28,66 +29,99 @@ std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsC
...
+  // For example, ValueSegment<int> == ValueSegment<float> will be erased. So will ValueSegment<int> ==
+  // DictionarySegment<int>. ReferenceSegments do not need to be handled differently because we expect a table to
+  // either have only ReferenceSegments or non-ReferenceSegments.
+  //
+  // We use type erasure here because we currently do not compare, e.g., a ValueSegment with a DictionarySegment,
```
mrzzzrm: "we currently do not" could be read as "Hyrise cannot do this". What is actually meant, is that we have no use cases where this happens. I'd make that clearer.

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -28,66 +29,99 @@ std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsC
...
+  // and we don't want the compiler to spend time instantiating unused templates. Whenever the types of the iterators
+  // is removed, we also erase the comparator lambda by wrapping it into an std::function. All of this brought the
+  // compile time down by a factor of 5. This is only relevant for the release build - in the debug build, iterators
```
mrzzzrm: Erasing the comparator type is relevant for debug as well, right?

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -28,66 +29,99 @@ std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsC
...
+  // C++ cannot compare strings and non-strings out of the box:
+  if constexpr (std::is_same_v<LeftType, std::string> == std::is_same_v<RightType, std::string>) {
+    bool flipped = false;
+    auto condition = _predicate_condition;
+    if (condition == PredicateCondition::GreaterThan || condition == PredicateCondition::GreaterThanEquals) {
+      condition = flip_predicate_condition(condition);
+      flipped = true;
+    }
+
+    auto erase_comparator_type = [](auto comparator, const auto& it1, const auto& it2) {
```
mrzzzrm: This reads as if you're NOT erasing the type if the comparator if `type_erasure == SegmentIterationTypeErasure::Always`?

mrzzzrm: Also, calling a function that does NOT erase the type under certain conditions "erase_comparator_type" is misleading. "erase_comparator_type_if_necessary"? "maybe_erase_..."?

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -28,66 +29,99 @@ std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsC
...
+    });
+  });
 
-  const auto left_segment = chunk->get_segment(_left_column_id);
-  const auto right_segment = chunk->get_segment(_right_column_id);
+  return result;
+}
+
+template <SegmentIterationTypeErasure type_erasure, typename LeftIterable, typename RightIterable>
+std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::_typed_scan_chunk(ChunkID chunk_id,
```
mrzzzrm: Would this function be shorter/easier to read if you'd do the flipping before calling it, passing it a flipped predicate condition and the iterables in the flipped order? Minor, but this would avoid mentioning `AbstractTableScanImpl::_scan_with_iterators<true>(erased_comparator, right_it, right_end, chunk_id_copy,
                                                              *matches_out_ref, left_it);`twice

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrks: I went for "reducing the compile time". For me, "optimization" usually hints at better run-time performance. I rather be careful with "Compile time optimization", which could be interpreted as "an optimization made at compile time"

mrks: "If the left and the right segment and/or data type are not the same" ?

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrks: like this?

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrks: d'oh. `conditionally_erase_comparator_type`?

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrks: At some point, we have to have a convert the runtime value of `flip` into a compile-time constant. If that happens here or in `ColumnVsColumnTableScanImpl::scan_chunk` shouldn't change the amount of code. I'd rather have `scan_chunk` only deal with the type erasure of the iterators.

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrzzzrm: + "we erase the **segment** types ..."

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrzzzrm: kk

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrks: the **segment iterable** types, no?

mrzzzrm: see, the devil's in the details.

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrzzzrm: nah, at first I didn't even see a difference. "actively use", maybe?

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrzzzrm: awkward, but certainly better than the current name

mrks:
This should be good to go

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -28,66 +29,102 @@ std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsC
...
+      }
+    });
+  });
 
-  const auto left_segment = chunk->get_segment(_left_column_id);
-  const auto right_segment = chunk->get_segment(_right_column_id);
+  return result;
+}
+
+template <SegmentIterationTypeErasure type_erasure, typename LeftIterable, typename RightIterable>
```
mrzzzrm: **Segment**IterationTypeErasure is used to control whether the **Comparator** is erased in "conditionally_erase_comparator_type". This is counterintuitive, even a boolean parameter, appropriately name "erase_comparator[_type]", would be a better fit

mrzzzrm: Type name == EraseType
variable name == the_actual_type_that_is_erased
?

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -28,66 +29,102 @@ std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsC
...
-        const auto condition = _predicate_condition;
-
-        with_comparator(condition, [&](auto predicate_comparator) {
-          auto comparator = [predicate_comparator](const auto& left, const auto& right) {
+  using LeftType = typename LeftIterable::ValueType;
+  using RightType = typename RightIterable::ValueType;
+
+  // C++ cannot compare strings and non-strings out of the box:
+  if constexpr (std::is_same_v<LeftType, std::string> == std::is_same_v<RightType, std::string>) {
+    bool flipped = false;
```
mrzzzrm: I know it's minor, but if you'd do the flipping outside of thing function, there wouldn't be a need for vaguely named variables "flipped" (what exactly is flipped?) and "condition" (how this is different from "_predicate_condition" is not immediately obvious)

mrzzzrm:
Should be ready to go after the remaining comments have been addressed/dismissed

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrks: Not sure how that improves things. We wouldn't have those variables, yes, but I would expect readers to be more surprised if an outside function magically flipped their predicate. Seeing the flip and its effect in close proximity seems to be better to me. I agree about the naming though, maybe `condition_was_flipped` and `potentially_flipped_condition`?

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrks: yepp, makes sense

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrzzzrm: yeah, at the moment I feel like awkward speaking names are better than short imprecise ones.

Bouncner:
Two stupid questions:
* what happens at the moment, when we compare an int and a float? Does the optimize add a cast or do we simply throw?
* "we expect a table to either have only ReferenceSegments or non-ReferenceSegments": could there be a case where we materialize a column? E.g., we have reference table (after some predicates) and need to evaluate something like `WEEK(shipdate)`. Wouldn't that create a value column?

mrks:
ad 1: We use c++'s int == float method
ad 2: That might make sense in the future, but doesn't happen right now. Currently, we have to materialize all segments.

PR #1419, opened 2019-01-17
Fix SQLiteWrapper loading for special chars

mrzzzrm:


src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -15,25 +15,29 @@ AbstractTableGenerator::AbstractTableGenerator(const std::shared_ptr<BenchmarkCo
 void AbstractTableGenerator::generate_and_store() {
   Timer timer;
 
-  std::cout << "- Loading/Generating tables " << std::flush;
+  std::cout << "- Loading/Generating tables " << std::endl;
   auto table_info_by_name = generate();
-  std::cout << "(" << format_duration(std::chrono::duration_cast<std::chrono::nanoseconds>(timer.lap())) << ")"
```
mrks: ?

src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -15,25 +15,29 @@ AbstractTableGenerator::AbstractTableGenerator(const std::shared_ptr<BenchmarkCo
...
-            << std::endl;
-
-  std::cout << "- Encoding tables " << std::flush;
 
   /**
    * Encode the Tables
    */
+  std::cout << "- Encoding tables if necessary" << std::endl;
```
mrks: "if requested"?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -61,10 +63,15 @@ void BenchmarkRunner::run() {
...
 
   if (_config.verify) {
+    std::cout << "- Loading tables into SQLite for verification." << std::endl;
     // Load the data into SQLite
     _sqlite_wrapper = std::make_unique<SQLiteWrapper>();
     for (const auto& [table_name, table] : StorageManager::get().tables()) {
+      std::cout << "- Loading '" << table_name << "' into SQLite " << std::flush;
+      Timer timer;
       _sqlite_wrapper->create_table(*table, table_name);
+      std::cout << "(" << format_duration(std::chrono::duration_cast<std::chrono::nanoseconds>(timer.lap())) << ")"
```
mrks: With this pattern being used over and over again, shouldn't the timer have a `lap_as_string` method? Name tbd?

src/bin/playground.cpp
```diff
@@ -1,10 +1,16 @@
...
+#include "utils/sqlite_wrapper.hpp"
 
 using namespace opossum;  // NOLINT
 
 int main() {
-  std::cout << "Hello world!!" << std::endl;
+  SQLiteWrapper wrapper;
+  wrapper.create_table(*load_table("resources/test_data/tbl/int_int_int.tbl"), "t");
+  Print::print(wrapper.execute_query("SELECT * FROM t"));
+
```
mrks: !

mrzzzrm:
WIP, gdmit

src/benchmarklib/abstract_table_generator.cpp
mrzzzrm: na, could be dictionary in the binary already. Then an encoding was requested, but there was no need to re-encode

src/benchmarklib/abstract_table_generator.cpp
mrzzzrm: Load times printed for each table now. Want an aggregation, too?

mrks:
Oups, sorry. Just saw the mail.

src/benchmarklib/abstract_table_generator.cpp
mrks: Yes. Remember that not every benchmark uses the FileBasedTableGenerator. Also, it helps to see how long I will have to wait the next time, and I don't want to sum up the times for that.

mrzzzrm: ack

src/benchmarklib/benchmark_runner.cpp
mrzzzrm: I like `lap_formatted()`, but whatever you prefer

mrks: Sounds good

src/benchmarklib/benchmark_runner.hpp
```diff
@@ -33,7 +33,7 @@ class BenchmarkRunner {
                   std::unique_ptr<AbstractTableGenerator> table_generator, const nlohmann::json& context);
   ~BenchmarkRunner();
 
-  void run();
+  int run();
```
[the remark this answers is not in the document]
mrzzzrm: I derived this from QApplication::run() (as it is similar to the BenchmarkRunner in that it takes over the application completely) returning the main() return code. Will doc, is that okay?

If main() computes its own return code, then we have 3 main()s doing this (file based, join order, tpch).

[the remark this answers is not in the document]
mrks: Mmmh, not happy with it, but if you doc it appropriately, that should be fine.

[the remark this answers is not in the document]
mrzzzrm: Assert at the end of "BenchmarkRunner::run()" is better?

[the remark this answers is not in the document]
mrks: Why not fail right when the validation fails?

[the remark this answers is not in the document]
mrzzzrm: Because I want to run the entire JOB and see which queries fail and which do not. I do not want the benchmark to crash on the first failed query, waiting for me to start it again with the remaining queries.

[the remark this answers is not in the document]
mrks: > Assert at the end of "BenchmarkRunner::run()" is better?

Yes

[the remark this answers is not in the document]
mrzzzrm: Looks weird if the app logs verification failure while the queries are running and then again, via the Assert, after all queries have finished, but have it as you like.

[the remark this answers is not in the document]
mrzzzrm: oh, that sounds sooo passive agressive :D

PR #1420, opened 2019-01-17
Fix logging whether tables were encoded or not during Benchmark setup

mrzzzrm:
embarrassing


mrks:
Could you add a title before you merge this?

PR #1422, opened 2019-01-17
RFC: LQP to CPP

mrzzzrm:
As discussed in the fortnightly, a way to turn query plans back into the cpp generating them might come in handy to manually construct alternative query plans. I went for LQPs, since they are way easier to manipulate than PQPs and gave this a quick shot. Your job is to decide whether it's worth it and whether I should continue. As you can see, this is probably going to be a ~1k cpp + tests in the end.

This is by no means done, lots of node types and expression types still print nothing. Also, subselects, unions and similar stuff is not supported yet.

BUT, it can turn simple queries back into LQPs: `SELECT * FROM lineitem, supplier WHERE s_suppkey = l_suppkey AND l_shipdate > 'than_ever'` becomes:

```
const auto lineitem = StoredTableNode::make("lineitem");
const auto supplier = StoredTableNode::make("supplier");

const auto l_shipdate = lqp_column_({lineitem, ColumnID{10}});
const auto s_suppkey = lqp_column_({supplier, ColumnID{0}});
const auto l_suppkey = lqp_column_({lineitem, ColumnID{2}});

const auto lqp = 
 JoinNode::make(JoinMode::Inner, equals_(s_suppkey, l_suppkey),
  PredicateNode::make(greater_than_(l_shipdate, "than_ever"),
    ValidateNode::make(
      lineitem)), 
  ValidateNode::make(
    supplier));
```

### DIFF
--- a/src/bin/playground.cpp
+++ b/src/bin/playground.cpp
@@ -1,10 +1,23 @@
 #include <iostream>
 
 #include "types.hpp"
+#include "tpch/tpch_table_generator.hpp"
+#include "tpch/tpch_query_generator.hpp"
+#include "sql/sql_pipeline_builder.hpp"
+#include "utils/plan_to_cpp.hpp"
 
 using namespace opossum;  // NOLINT
 
 int main() {
-  std::cout << "Hello world!!" << std::endl;
+  TpchTableGenerator{0.01f}.generate_and_store();
+  auto query = TPCHQueryGenerator{false, 0.01f}.build_deterministic_query(QueryID{6});
+  query = "SELECT * FROM lineitem, supplier WHERE s_suppkey = l_suppkey AND l_shipdate > 'than_ever'";
+
+  auto statement = SQLPipelineBuilder{query}.create_pipeline_statement();
+
+  statement.get_optimized_logical_plan()->print();
+
+  std::cout << lqp_to_cpp(statement.get_optimized_logical_plan()) << std::endl;
+
   return 0;
 }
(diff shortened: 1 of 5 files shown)


mrks:
Wow, you actually followed through on this :)

On first look, this looks like a good thing. It's definitely helpful and I like that it is limited to a single file. Just a couple of thoughts:

1. Is the idea that the LQP will be translated to the PQP without being optimized?
2. Should we automatically do this right away and test that we actually end up with the same PQP?
3. How can we make sure that new operators / options are correctly added?
4. Do you foresee any issues when it comes to more complex queries? I am specifically thinking about correlated subqueries.

mrzzzrm:
> Is the idea that the LQP will be translated to the PQP without being optimized?

I understand lqp_to_cpp to be used in a context where devs want to "play Optimizer themselves", so applying the Optimizer to the built LQP is probably not desirable?

> Should we automatically do this right away and test that we actually end up with the same PQP?

You mean "test" as in adding Asserts() or unit tests? How would we do either? This function generates a std::string with cpp code, after all - it needs to be c/p and compiled before it can become an actual LQP/PQP again. But I am probably misunderstanding the question? 

> How can we make sure that new operators / options are correctly added?

About "correctly" I do not know, but if there are new LQPNodes/Expressions, then the compiler will not compile plan_to_cpp.cpp until the new enum members representing these new LQPNodes/Expressions are handled.

> Do you foresee any issues when it comes to more complex queries? I am specifically thinking about correlated subqueries.

At the moment, I do not, but I haven't really thought about it yet. I think it's going to be a bulkload of code, but it's going to work. Probably... :D

mrks:
Disregard question 2. I haven't had my morning cup of coffee yet.

mrzzzrm:
Closing until

PR #1424, opened 2019-01-21
Track if tables are validated

### LINKED ISSUE #1344: MVCC: Store is_validated in Table?
A seminar group wrote the attached code, which fails with:

```
libc++abi.dylib: terminating with uncaught exception of type std::logic_error: /hyrise/src/lib/concurrency/transaction_context.cpp:150 All read/write operators need to be in state Executed (especially not Failed).
```

The reason for this is that the input to `Update` was not validated, thus they were trying to update the rows twice. This rightfully failed.

If we add a field `Table::_is_validated` which is only set by the `Validate` operator, we should be able to prevent this. Any objections?

[mvcc_benchmark.txt](https://github.com/hyrise/hyrise/files/2644786/mvcc_benchmark.txt)

mrks:
This makes sure that the input to the delete operator is always validated. Also, if a SQL Pipeline has a transaction context, all output are asserted to be validated as well. fixes #1344.


src/lib/operators/difference.cpp
```diff
@@ -35,6 +35,8 @@ std::shared_ptr<const Table> Difference::_on_execute() {
 
   auto output = std::make_shared<Table>(input_table_left()->column_definitions(), TableType::References);
 
+  if (input_table_left()->is_validated() && input_table_right()->is_validated()) output->mark_as_validated();
```
mrzzzrm: why do you have to do this in every individual operator? cannot AbstractOperator::execute() do this?

mrks: That's what I thought in between to, but the table is const once it leaves the operator.

mrks: I don't see how an `std::shared_ptr<const SomeThing>` is different from a `const SomeThing`. Unless you are suggesting to remove const from everything, I don't see why we should make a difference there.

Generally, I like the idea of "only the operator itself knows how to construct the result, so no one else should be allowed to touch it".

mrks: (But yes, it would be nicer for the AbstractOperator to handle that. It is just that given the option of saving those ~20 lines or weakening the const safety, I tend to value the latter more)

mrzzzrm: Ignore my rant about `std::shared_ptr<const SomeThing>` that's a different topic. 

I aggree that only the operators should touch the tables. But `AbstractOperator::execute()` IS still pretty close to being the operator, so i wouldn't see a problem with `AbstractOperator::_on_execute()` handing out a mutable table that  `AbstractOperator::execute()` can mark as validated.

Doing it as it is done right now is easily forgotten when implementing an operator, I'm sure you'll agree. If validated/not_validated is an integral part of the table state, then maybe it should be a constructor argument, so setting it cannot be missed?

mrzzzrm: >  I tend to value the latter more

Where is the value the Table being const as soon as it is obtained by AbstractOperator::execute()? I think the risk that AbstractOperator::execute() is going to do something invalid with it is very low.

Forgetting to call mark_as_validated in an operator is a very real risk, on the other hand, imho.

src/lib/storage/table.cpp
```diff
@@ -212,4 +212,12 @@ size_t Table::estimate_memory_usage() const {
   return bytes;
 }
 
+bool Table::is_validated() const { return _is_validated; }
```
mrzzzrm: This will always be the same as all_of(chunks, chunk -> !chunk.has_mvcc_data()), right? Maybe make that a debug assert so these two "states" cannot diverge?

mrzzzrm: Not that simple, right? Actually there are three possible states: Validated, NotValidated and NoMvccDataInTheFirstPlace.

src/lib/storage/table.hpp
```diff
@@ -167,6 +167,15 @@ class Table : private Noncopyable {
    */
   size_t estimate_memory_usage() const;
 
+  // Returns whether the table has already been validated (i.e., contains only rows visible to the current transaction)
+  bool is_validated() const;
```
mrzzzrm: Right now it is possible to append chunks WITH mvcc data to a "validated" Table without an assert firing, right?

src/lib/storage/table.cpp
mrks: Exactly, see #1344. The idea here would be for this to simply be a "Validate has seen this table" flag.

src/lib/storage/table.hpp
mrks: No, because a table can only be validated by the validate operator and once it leaves that operator, it is const. No snappishness intended. ;)

src/lib/operators/abstract_read_only_operator.cpp
```diff
@@ -7,6 +7,10 @@
 namespace opossum {
 
 std::shared_ptr<const Table> AbstractReadOnlyOperator::_on_execute(std::shared_ptr<TransactionContext>) {
+  DebugAssert(
+      !input_left() || !input_right() || input_table_left()->is_validated() == input_table_right()->is_validated(),
```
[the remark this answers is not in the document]
mrks: I hoped that the text of the assert would be speaking enough. Any idea how to improve this?

src/lib/operators/difference.cpp
mrks: I'll give this some thought in a quite minute.

src/lib/operators/abstract_read_only_operator.cpp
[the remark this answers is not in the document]
Bouncner: Let's talk in person.

FabianWiebe:
> @FabianWiebe - Any idea where we can set a flag `context.is_validated` in JitValidate?

If the JitValidate operator sets a flag, this would require that the flag is set for every processed tuple. Therefore, I propose that the JitOperatorWrapper has to set this flag in the output table, if a JitValidate exists in the jittable operator pipeline.

src/lib/operators/difference.cpp
mrzzzrm: We have never used Print's in the middle of plans and I wouldn't be able to see use for it, anyway. Speaking of which, we never use Print as an operator, do we? 

I wouldn't allow Print to be the issue here. As a band-aid, its on_execute should just return a nullptr

mrks: It's not automatically placed into a plan, but is a useful debug feature. I have placed Print into the middle of query plans quite often to have a look at the data that is passed around.

mrzzzrm: I am pretty sure I am going to be relentless here: Requiring operators to **not forget** to "forward" a property that has nothing, really nothing, to do with the logic they implement is a bad idea. If our architecture forces us to make this requirement, then it is our architecture is bad.

(In this case, I think, the bad architecture is either: Operators, glorified functions, after all, return an immutable object. OR: the "is_validated" property on a Table is not a good idea in the first place.)

mrzzzrm: The bug mentioned in the issue could have been avoided by Assert()ing that the Table with Insert-data has no MVCCData, right? Maybe just doing that would be the better approach to solve the issue? Tbh, I think the motivation in the issue is too weak to warrant hacky changes in all operators.

mrks: > The bug mentioned in the issue could have been avoided by Assert()ing that the Table with Insert-data has no MVCCData, right?

No, this is all about Delete and has nothing to do with the insert.

> Tbh, I think the motivation in the issue is too weak to warrant hacky changes in all operators.

Yes. And your suggestion above led me to think about this again. Simple solution coming up in a couple of minutes. Can't believe I didn't think of that before.

PR #1425, opened 2019-01-21
Fix NULL behavior (#1421)

### LINKED ISSUE #1421: MultipleChunkReferenceSegmentAccessor shows faulty behavior with NULL values
The sort-based aggregate seminar group found out that Hyrise crashes when they sorted the output of a full outer join. Apparently, @mrks has seen the problem before and pointed directly to the MultipleChunkReferenceSegmentAccessor.

While working on that, I found out that the SortOperator itself is broken and does not handle NULL values correctly. I have fixes on my machine and will file PRs tomorrow.

Bensk1:
This fixes #1421 and some other nasty NULL cases.

One problem was that the SegmentAccessors did not correctly check for null values and accessed chunks with `INVALID_CHUNK_ID`.

Also, ReferenceSegmentIterables created SegmentPositions with `chunk_offset = 0u` when they were NULL. This way it was impossible to correctly check for `is_null()` afterwards. 

Last but not least, some optimizations that were implemented for `references_single_chunk` assume now that no NULL values are involved.

### DIFF
--- a/src/lib/storage/reference_segment/reference_segment_iterable.hpp
+++ b/src/lib/storage/reference_segment/reference_segment_iterable.hpp
@@ -28,13 +28,12 @@ class ReferenceSegmentIterable : public SegmentIterable<ReferenceSegmentIterable
     const auto begin_it = pos_list.begin();
     const auto end_it = pos_list.end();
 
-    // If we are guaranteed that the reference segment refers to a single chunk, we can do some optimizations.
+    // If we are guaranteed that the reference segment refers to a single non-NULL chunk, we can do some optimizations.
     // For example, we can use a single, non-virtual segment accessor instead of having to keep multiple and using
     // virtual method calls.
 
-    if (pos_list.references_single_chunk() && pos_list.size() > 0) {
-      const auto first_chunk_id = begin_it->is_null() ? ChunkID {0u} : begin_it->chunk_id;
-      auto referenced_segment = referenced_table->get_chunk(first_chunk_id)->get_segment(referenced_column_id);
+    if (pos_list.references_single_chunk() && pos_list.size() > 0 && !begin_it->is_null()) {
+      auto referenced_segment = referenced_table->get_chunk(begin_it->chunk_id)->get_segment(referenced_column_id);
       resolve_segment_type<T>(*referenced_segment, [&](const auto& typed_segment) {
         using SegmentType = std::decay_t<decltype(typed_segment)>;
 
--- a/src/lib/storage/reference_segment/reference_segment_iterable.hpp
+++ b/src/lib/storage/reference_segment/reference_segment_iterable.hpp
@@ -30,7 +30,7 @@ class ReferenceSegmentIterable : public SegmentIterable<ReferenceSegmentIterable
 
     // If we are guaranteed that the reference segment refers to a single non-NULL chunk, we can do some optimizations.
     // For example, we can use a single, non-virtual segment accessor instead of having to keep multiple and using
-    // virtual method calls.
+    // virtual method calls. If begin_it is NULL, chunk_id will be INVALID_CHUNK_ID. Therefore, we skip this case.
 
     if (pos_list.references_single_chunk() && pos_list.size() > 0 && !begin_it->is_null()) {
       auto referenced_segment = referenced_table->get_chunk(begin_it->chunk_id)->get_segment(referenced_column_id);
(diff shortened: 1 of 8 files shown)


src/lib/storage/reference_segment/reference_segment_iterable.hpp
```diff
@@ -28,11 +28,11 @@ class ReferenceSegmentIterable : public SegmentIterable<ReferenceSegmentIterable
     const auto begin_it = pos_list.begin();
     const auto end_it = pos_list.end();
 
-    // If we are guaranteed that the reference segment refers to a single chunk, we can do some optimizations.
+    // If we are guaranteed that the reference segment refers to a single non-NULL chunk, we can do some optimizations.
     // For example, we can use a single, non-virtual segment accessor instead of having to keep multiple and using
     // virtual method calls.
 
-    if (pos_list.references_single_chunk() && pos_list.size() > 0) {
+    if (pos_list.references_single_chunk() && pos_list.size() > 0 && !begin_it->is_null()) {
```
mrks: Add a comment on why NULL is so mean?

src/lib/storage/segment_accessor.hpp
```diff
@@ -85,12 +87,20 @@ class SingleChunkReferenceSegmentAccessor : public BaseSegmentAccessor<T> {
   explicit SingleChunkReferenceSegmentAccessor(const ReferenceSegment& segment)
       : _segment{segment},
         _chunk_id((*_segment.pos_list())[ChunkOffset{0}].chunk_id),
-        _accessor{create_segment_accessor<T>(
-            segment.referenced_table()->get_chunk(_chunk_id)->get_segment(_segment.referenced_column_id()))} {}
+        // If *_segment.pos_list())[ChunkOffset{0}] is NULL, its chunk_id is INVALID_CHUNK_OFFSET. When the
```
mrks: extra `)`

src/lib/storage/segment_accessor.hpp
```diff
@@ -85,12 +87,20 @@ class SingleChunkReferenceSegmentAccessor : public BaseSegmentAccessor<T> {
...
+        // SingleChunkReferenceSegmentAccessor is used, all entries reference the same chunk_id (INVALID_CHUNK_OFFSET).
+        // Therefore, we can safely assume that all other entries are also NULL and always return std::nullopt.
+        _accessor((*_segment.pos_list())[ChunkOffset{0}].is_null()
+                      ? nullptr
+                      : create_segment_accessor<T>(segment.referenced_table()->get_chunk(_chunk_id)->get_segment(
+                            _segment.referenced_column_id()))) {}
 
   const std::optional<T> access(ChunkOffset offset) const final {
-    const auto referenced_chunk_offset = (*_segment.pos_list())[offset].chunk_offset;
+    if (!_accessor) {
```
mrks: Did you benchmark this? Adding a branch in a hot method seems like it might influence the performance. Maybe add a class-internal `NullAccessor` that always returns `std::nullopt`? That way we would simply direct the virtual method call somewhere else.

src/test/operators/sort_test.cpp
```diff
@@ -217,4 +223,19 @@ TEST_P(OperatorsSortTest, DescendingSortOfOneDictSegment) {
   EXPECT_TABLE_EQ_ORDERED(sort->get_output(), expected_result);
 }
 
+TEST_P(OperatorsSortTest, SortAfterOuterJoin) {
+  // auto join = std::make_shared<JoinNestedLoop>(_table_wrapper_join_1, _table_wrapper_join_2, JoinMode::Outer,
+  // ColumnIDPair(ColumnID{0}, ColumnID{0}), PredicateCondition::LessThan);
```
mrks: ?

src/test/storage/iterables_test.cpp
```diff
@@ -231,7 +231,24 @@ TEST_F(IterablesTest, ReferenceSegmentIteratorWithIterators) {
...
+      std::make_unique<ReferenceSegment>(table, ColumnID{0u}, std::make_shared<PosList>(std::move(pos_list)));
+
+  auto iterable = ReferenceSegmentIterable<int>{*reference_segment};
+
+  auto sum = uint32_t{0};
+  auto accessed_offsets = std::vector<ChunkOffset>{};
+  iterable.with_iterators(SumUpWithIterator{sum, accessed_offsets});
+
+  EXPECT_EQ(sum, 0u);
+  EXPECT_EQ(accessed_offsets, (std::vector<ChunkOffset>{ChunkOffset{0}}));
```
mrks: Mmh, I am not sure how safe this test is. A sum ending up to be zero can also happen if we read zero-initialized memory. Can't we explicitly check for NULL?

Bensk1: We check in line 251 if the row containing NULL was accessed. And the test is not about the sum, I don't know whether a change is necessary here.

PR #1427, opened 2019-01-21
Fix sort merge join when joining columns containing NULL

Bensk1:
Previously, we had `NULL_VALUE_ID{std::numeric_limits<ValueID::base_type>::max()}`. Now NULL values are represented by a ValueID that equals the size of the dictionary. The ColumnMaterializer did still use the old way to represent NULL values.

Looking over the code, there should be the opportunity for some simple optimizations taking into account if the column is nullable or not. E.g., the branch in the hot loop that is comparing `value_id != null_value_id` could be avoided.


src/test/operators/join_full_test.cpp
```diff
@@ -250,6 +250,13 @@ TYPED_TEST(JoinFullTest, SmallerOuterJoin) {
                                              "resources/test_data/tbl/joinoperators/int_smaller_outer_join.tbl", 1);
 }
 
+TYPED_TEST(JoinFullTest, SmallerOuterJoinWithNull) {
```
mrks: Any reason you're specifically testing the outer join?

Bensk1: Could also be left or right join. But only these three materialize NULL values.

Issue #1208, opened 2018-10-19
Weird code in radix_cluster_sort.hpp


mrks:
Something is fishy here:

https://github.com/hyrise/hyrise/blob/e5b6a4ce0123ade5a52b0cef622c9eff0a41371e/src/lib/operators/join_sort_merge/radix_cluster_sort.hpp#L261-L266

It takes`->second` of `max_element`, which I understand to be the number of occurences. Why is this stored in a `vector<T>`? Later, the clusterer takes this and compares it against a T:

https://github.com/hyrise/hyrise/blob/e5b6a4ce0123ade5a52b0cef622c9eff0a41371e/src/lib/operators/join_sort_merge/radix_cluster_sort.hpp#L299-L305

mrks:
@Bouncner: Please add `fixes #nnnn` or `closes #nnnn` to PRs when appropriate.

PR #1428, opened 2019-01-21
Fix sampling in sort-merge join

Bouncner:
This PR addresses #1208. Previously, the sampling (apart from being plain wrong) was using the frequency of values to determine splits in the data. Since the chance of sampling the same value multiple times is low (and even then, it does not need to be a good split value), the splitting resulted in highly skewed clusters.

Now, we gather a small `SampleRequest` in which each materialized segment puts some samples. Then, the sample requests are merged, sorted and we select the required number of split values.

I measured the performance with an active scheduler and the change is minimal. I expected it to be better since the values are now (almost) equally sized.

Now:
```
➜  OpossumDB git:(fix/sort_merge_join) ✗ ./rel/hyriseMicroBenchmarks  --benchmark_filter="BM_Sam" --benchmark_repetitions=10
Generating TPC-H data set with scale factor 1 and Dictionary encoding:
[...]
----------------------------------------------------------------------------------------
Benchmark                                                 Time           CPU Iterations
----------------------------------------------------------------------------------------
[...]
TPCHDataMicroBenchmarkFixture/BM_Sampling_mean   1468207936 ns 1006052700 ns          1
TPCHDataMicroBenchmarkFixture/BM_Sampling_median 1464570165 ns 1000790500 ns          1
TPCHDataMicroBenchmarkFixture/BM_Sampling_stddev   23104783 ns   11344929 ns          1
```

Before
```
➜  OpossumDB_master git:(master) ✗ ./rel/hyriseMicroBenchmarks  --benchmark_filter="BM_Sam" --benchmark_repetitions=10
Generating TPC-H data set with scale factor 1 and Dictionary encoding:
[...]
----------------------------------------------------------------------------------------
Benchmark                                                 Time           CPU Iterations
----------------------------------------------------------------------------------------
[...]
TPCHDataMicroBenchmarkFixture/BM_Sampling_mean   1502313768 ns 1017044400 ns          1
TPCHDataMicroBenchmarkFixture/BM_Sampling_median 1515102442 ns 1017976500 ns          1
TPCHDataMicroBenchmarkFixture/BM_Sampling_stddev   68923687 ns   15745084 ns          1
```

### DIFF
--- a/src/lib/operators/join_sort_merge/column_materializer.hpp
+++ b/src/lib/operators/join_sort_merge/column_materializer.hpp
@@ -30,11 +30,17 @@ using MaterializedSegment = std::vector<MaterializedValue<T>>;
 template <typename T>
 using MaterializedSegmentList = std::vector<std::shared_ptr<MaterializedSegment<T>>>;
 
+/**
+ * This data structure is passed as a reference to the jobs which materialize
+ * the chunks. Each job then adds `samples_to_collect` samples to its passed
+ * SampleRequest. All SampleRequests are later merged to gather a global sample
+ * list with which the split values for the radix partitioning are determined.
+ */
 template <typename T>
 struct SampleRequest {
-  explicit SampleRequest(uint32_t sample_count) : samples_to_collect(sample_count), collected_samples() {}
+  explicit SampleRequest(uint16_t sample_count) : samples_to_collect(sample_count), collected_samples() {}
 
-  const uint32_t samples_to_collect;
+  const uint16_t samples_to_collect;
   std::vector<T> collected_samples;
 };
 
@@ -55,7 +61,7 @@ class ColumnMaterializer {
    **/
   std::tuple<std::unique_ptr<MaterializedSegmentList<T>>, std::unique_ptr<PosList>, std::vector<T>> materialize(
       const std::shared_ptr<const Table> input, const ColumnID column_id) {
-    const uint32_t samples_per_chunk = 10;  // rather arbitrarily chosen number
+    const uint16_t samples_per_chunk = 10;  // rather arbitrarily chosen number
     const auto chunk_count = input->chunk_count();
 
     auto output = std::make_unique<MaterializedSegmentList<T>>(chunk_count);
@@ -66,8 +72,8 @@ class ColumnMaterializer {
 
     std::vector<std::shared_ptr<AbstractTask>> jobs;
     for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
-      const uint32_t samples_to_write =
-          static_cast<uint32_t>(std::min(samples_per_chunk, input->get_chunk(chunk_id)->size()));
+      const uint16_t samples_to_write =
+          static_cast<uint16_t>(std::min(samples_per_chunk, static_cast<uint16_t>(input->get_chunk(chunk_id)->size())));
       sample_requests.push_back(SampleRequest<T>(samples_to_write));
       jobs.push_back(
           _create_chunk_materialization_job(output, null_rows, chunk_id, input, column_id, sample_requests.back()));
--- a/src/lib/operators/join_sort_merge/column_materializer.hpp
+++ b/src/lib/operators/join_sort_merge/column_materializer.hpp
@@ -117,18 +117,20 @@ class ColumnMaterializer {
 
   /**
    * Samples values from a materialized segment.
-   * We collect locally write once to the global sample collection to limit non-local writes.
+   * We collect samples locally and write once to the global sample collection to limit non-local writes.
    */
-  void _gather_samples_from_segment(MaterializedSegment<T>& segment_output, SampleRequest<T>& sample_request) const {
+  void _gather_samples_from_segment(const MaterializedSegment<T>& segment, SampleRequest<T>& sample_request) const {
     const auto samples_to_collect = sample_request.samples_to_collect;
     std::vector<T> collected_samples;
     collected_samples.reserve(samples_to_collect);
 
-    if (segment_output.size() > 0 && samples_to_collect > 0) {
+    if (segment.size() > 0 && samples_to_collect > 0) {
+      const auto step_width = segment.size() / std::max(static_cast<uint16_t>(1u), samples_to_collect);
 
-      for (uint32_t sample_count = 0; sample_count < samples_to_collect; ++sample_count) {
-        // output vector does not contain NULL values, thus we do not have to check for NULL samples.
-        collected_samples.push_back(segment_output[static_cast<size_t>(sample_count * step_width)].value);
+      for (uint16_t sample_count = 0; sample_count < samples_to_collect; ++sample_count) {
+        // NULL values in passed `segment` vector have already been
+        // removed, thus we do not have to check for NULL samples.
+        collected_samples.push_back(segment[static_cast<size_t>(sample_count * step_width)].value);
       }
 
       // Sequential write of result
--- a/src/lib/operators/join_sort_merge/column_materializer.hpp
+++ b/src/lib/operators/join_sort_merge/column_materializer.hpp
@@ -32,16 +32,15 @@ using MaterializedSegmentList = std::vector<std::shared_ptr<MaterializedSegment<
 
 template <typename T>
 struct SampleRequest {
-  SampleRequest(uint32_t sample_count) : samples_to_collect(sample_count), collected_samples() { }
+  explicit SampleRequest(uint32_t sample_count) : samples_to_collect(sample_count), collected_samples() {}
 
   const uint32_t samples_to_collect;
   std::vector<T> collected_samples;
 };
 
 /**
- * Materializes a table for a specific segment and sorts it if required. Row-Ids are kept in order to enable
- * the construction of pos lists for the algorithms that are using this class.
- TODO: sampling explain
+ * Materializes a table for a specific segment and sorts it if required. Result is a triple of
+ * materialized values, positions of NULL values, and a list of samples.
  **/
 template <typename T>
 class ColumnMaterializer {
@@ -56,7 +55,7 @@ class ColumnMaterializer {
    **/
   std::tuple<std::unique_ptr<MaterializedSegmentList<T>>, std::unique_ptr<PosList>, std::vector<T>> materialize(
       const std::shared_ptr<const Table> input, const ColumnID column_id) {
-    const uint32_t samples_per_chunk = 10;
+    const uint32_t samples_per_chunk = 10;  // rather arbitrarily chosen number
     const auto chunk_count = input->chunk_count();
 
     auto output = std::make_unique<MaterializedSegmentList<T>>(chunk_count);
@@ -70,7 +69,8 @@ class ColumnMaterializer {
       const uint32_t samples_to_write =
           static_cast<uint32_t>(std::min(samples_per_chunk, input->get_chunk(chunk_id)->size()));
       sample_requests.push_back(SampleRequest<T>(samples_to_write));
-      jobs.push_back(_create_chunk_materialization_job(output, null_rows, chunk_id, input, column_id, sample_requests.back()));
+      jobs.push_back(
+          _create_chunk_materialization_job(output, null_rows, chunk_id, input, column_id, sample_requests.back()));
       jobs.back()->schedule();
     }
 
@@ -80,7 +80,8 @@ class ColumnMaterializer {
     gathered_samples.reserve(samples_per_chunk * chunk_count);
 
     for (const auto& sample_request : sample_requests) {
-      gathered_samples.insert(gathered_samples.end(), sample_request.collected_samples.begin(), sample_request.collected_samples.end());
+      gathered_samples.insert(gathered_samples.end(), sample_request.collected_samples.begin(),
+                              sample_request.collected_samples.end());
     }
 
     return {std::move(output), std::move(null_rows), std::move(gathered_samples)};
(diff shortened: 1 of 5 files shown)


src/lib/operators/join_sort_merge/column_materializer.hpp
```diff
@@ -30,9 +30,17 @@ using MaterializedSegment = std::vector<MaterializedValue<T>>;
 template <typename T>
 using MaterializedSegmentList = std::vector<std::shared_ptr<MaterializedSegment<T>>>;
 
+template <typename T>
+struct SampleRequest {
+  explicit SampleRequest(uint32_t sample_count) : samples_to_collect(sample_count), collected_samples() {}
+
+  const uint32_t samples_to_collect;
```
mrks: why uint32?

src/test/operators/join_equi_test.cpp
```diff
@@ -67,7 +67,8 @@ TYPED_TEST(JoinEquiTest, InnerJoinIntFloatRadixBit) {
     join->execute();
     EXPECT_TABLE_EQ_UNORDERED(join->get_output(), expected_result);
 
-    for (size_t radix_bits : {1, 2, 3, 10, 17}) {
+    for (size_t radix_bits : {1, 2, 8}) {
+      std::cout << radix_bits << std::endl;
```
mrks: ?

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
```diff
@@ -345,10 +321,15 @@ class RadixClusterSort {
     ColumnMaterializer<T> right_column_materializer(!_equi_case, _materialize_null_right);
     auto materialization_left = left_column_materializer.materialize(_input_table_left, _left_column_id);
     auto materialization_right = right_column_materializer.materialize(_input_table_right, _right_column_id);
-    auto materialized_left_segments = std::move(materialization_left.first);
-    auto materialized_right_segments = std::move(materialization_right.first);
-    output.null_rows_left = std::move(materialization_left.second);
-    output.null_rows_right = std::move(materialization_right.second);
+    auto materialized_left_segments = std::move(std::get<0>(materialization_left));
+    auto materialized_right_segments = std::move(std::get<0>(materialization_right));
```
mrks: structured binding?

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
```diff
@@ -345,10 +321,15 @@ class RadixClusterSort {
...
+    output.null_rows_left = std::move(std::get<1>(materialization_left));
+    output.null_rows_right = std::move(std::get<1>(materialization_right));
+    auto samples_left = std::get<2>(materialization_left);
+    auto samples_right = std::get<2>(materialization_right);
+
+    // Append right samples to left samples and sort
+    samples_left.insert(samples_left.end(), samples_right.begin(), samples_right.end());
```
mrks: reserve?

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
```diff
@@ -247,25 +247,25 @@ class RadixClusterSort {
...
+    std::sort(sample_values.begin(), sample_values.end());
+
+    if (sample_values.size() <= _cluster_count - 1) {
+      return sample_values;
+    }
+
+    std::vector<T> split_values;
+    split_values.reserve(_cluster_count - 1);
+    auto jump_width = sample_values.size() / _cluster_count;
+    for (auto i = size_t{0}; i < _cluster_count - 1; ++i) {
```
mrks: i? :(

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
```diff
@@ -171,7 +171,7 @@ class RadixClusterSort {
   * -> Reserve the appropriate space for each output cluster to avoid ongoing vector resizing.
   * -> At last, each value of each chunk is moved to the appropriate cluster.
   **/
-  std::unique_ptr<MaterializedSegmentList<T>> _cluster(std::unique_ptr<MaterializedSegmentList<T>>& input_chunks,
+  std::unique_ptr<MaterializedSegmentList<T>> _cluster(const std::unique_ptr<MaterializedSegmentList<T>>& input_chunks,
```
mrks: remember that if you really want no changes, you need the const inside, too

src/lib/operators/join_sort_merge/column_materializer.hpp
```diff
@@ -67,24 +93,51 @@ class ColumnMaterializer {
...
       } else {
-        (*output)[chunk_id] = _materialize_generic_segment(*segment, chunk_id, null_rows_output);
+        (*output)[chunk_id] = _materialize_generic_segment(*segment, chunk_id, null_rows_output, sample_request);
       }
     });
   }
 
+  /**
+   * Samples values from a materialized segment.
+   * We collect locally write once to the global sample collection to limit non-local writes.
```
mrks: missing word?

src/lib/operators/join_sort_merge/column_materializer.hpp
```diff
@@ -67,24 +93,51 @@ class ColumnMaterializer {
...
+   */
+  void _gather_samples_from_segment(MaterializedSegment<T>& segment_output, SampleRequest<T>& sample_request) const {
```
mrks: i don't like the signature of this. On first sight, I thought that the first parameter was the input.

src/lib/operators/join_sort_merge/column_materializer.hpp
```diff
@@ -67,24 +93,51 @@ class ColumnMaterializer {
...
+    const auto samples_to_collect = sample_request.samples_to_collect;
+    std::vector<T> collected_samples;
+    collected_samples.reserve(samples_to_collect);
+
+    if (segment_output.size() > 0 && samples_to_collect > 0) {
+      const auto step_width = segment_output.size() / std::max(1u, samples_to_collect);
+
+      for (uint32_t sample_count = 0; sample_count < samples_to_collect; ++sample_count) {
+        // output vector does not contain NULL values, thus we do not have to check for NULL samples.
```
mrks: I don't get this. What exactly is not checked?

src/lib/operators/join_sort_merge/column_materializer.hpp
```diff
@@ -67,24 +93,51 @@ class ColumnMaterializer {
    **/
   std::shared_ptr<AbstractTask> _create_chunk_materialization_job(std::unique_ptr<MaterializedSegmentList<T>>& output,
                                                                   std::unique_ptr<PosList>& null_rows_output,
-                                                                  ChunkID chunk_id, std::shared_ptr<const Table> input,
-                                                                  ColumnID column_id) {
-    return std::make_shared<JobTask>([this, &output, &null_rows_output, input, column_id, chunk_id] {
+                                                                  const ChunkID chunk_id,
```
mrks: If this is private, shouldnt it have an underscore? Why doesn't tidy complain?

src/test/operators/join_equi_test.cpp
Bouncner: The test took 1s in release. We built 2^17 partitions.
2^8 is sufficient to check whether we can deal with #partitions>#rows.

Shall I add a comment?

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
Bouncner: Not necessary here: https://stackoverflow.com/a/35359472/1147726

Shall I add a short comment?

src/lib/operators/join_sort_merge/column_materializer.hpp
Bouncner: Can't follow? Function has an underscore. Or what are you looking at?

mrks: nevermind

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
mrks: Learned something new. It'd add a comment, yes.

src/test/operators/join_equi_test.cpp
mrks: That explains the reduced test size, but what about the cout?

Bouncner: Narf. Didn't see that. Was just there for debugging.

Will add the comment and remove the cout.

src/lib/operators/join_sort_merge/column_materializer.hpp
Bouncner: Added a comment.

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
Bouncner: Added a comment.

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
Bouncner: Could you please take a look at 15f9797? I'd like to hear from someone who understands C++ that these changes do not have a negative impact performance.
I did not measure a difference, but we just might lack the right benchmarks.

src/test/operators/join_equi_test.cpp
Bouncner: Done.

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
mrks: Looks like you are using `left_column_materializer.materialize` twice. I wouldn't expect any differences though.

src/lib/operators/join_sort_merge/column_materializer.hpp
```diff
@@ -30,9 +30,17 @@ using MaterializedSegment = std::vector<MaterializedValue<T>>;
 template <typename T>
 using MaterializedSegmentList = std::vector<std::shared_ptr<MaterializedSegment<T>>>;
 
+template <typename T>
+struct SampleRequest {
+  explicit SampleRequest(uint32_t sample_count) : samples_to_collect(sample_count), collected_samples() {}
```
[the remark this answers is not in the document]
mrks: SubSample? It's not really a "request", just a data structure that results are stored in.

src/lib/operators/join_sort_merge/column_materializer.hpp
mrks: That wasn't my point. We shouldn't use naked data types without documenting why they are the best fit, or better, having a name for them.

src/lib/operators/join_sort_merge/column_materializer.hpp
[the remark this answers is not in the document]
Bouncner: `SubSample` sounds good.

src/lib/operators/join_sort_merge/column_materializer.hpp
Bouncner: A name would probably be an overkill here. I thought the name and the type are self-explanatory.  Any ideas how to improve that?

Bouncner: Changed to `SubSample` and using ChunkOffset now.

src/lib/operators/join_sort_merge/column_materializer.hpp
```diff
@@ -70,26 +66,29 @@ class ColumnMaterializer {
...
-    samples.reserve(chunk_count * samples_per_chunk);
-    auto samples_written = ChunkOffset{0};
+    std::vector<SubSample<T>> subsamples;
+    subsamples.reserve(chunk_count);
 
     std::vector<std::shared_ptr<AbstractTask>> jobs;
     for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
       const auto samples_to_write = std::min(samples_per_chunk, input->get_chunk(chunk_id)->size());
-      SubSample<T> sub_sample(samples_to_write, samples_written, samples);
+      subsamples.push_back(SubSample<T>(samples_to_write));
```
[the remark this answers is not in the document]
Bouncner: Listening to "Yellow Submarine", I decided to name the struct `Subsample`.

src/lib/operators/join_sort_merge/column_materializer.hpp
```diff
@@ -70,26 +66,29 @@ class ColumnMaterializer {
...
 
-    samples.resize(samples_written);
     CurrentScheduler::wait_for_tasks(jobs);
 
-    return {std::move(output), std::move(null_rows), std::move(samples)};
+    auto gathered_samples = std::vector<T>();
+    gathered_samples.reserve(samples_per_chunk * chunk_count);
+
+    for (const auto& subsample : subsamples) {
+      gathered_samples.insert(gathered_samples.end(), subsample.samples.begin(), subsample.samples.end());
```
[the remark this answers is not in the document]
Bouncner: Not really sure how to use `move` here. :-/
```
for (auto&& subsample : subsamples) {
  ????;
}
```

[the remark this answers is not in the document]
mrks: Ah, as we are working on iterators, this becomes a tad more complicated. You could do

```
gathered_samples.insert(gathered_samples.end(), std::make_move_iterator(subsample.samples.begin()), std::make_move_iterator(subsample.samples.end()));
```

Let's wait with that until it actually pops up in the profiler.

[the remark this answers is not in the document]
Bouncner: Since we collect a maximum of 10 items per chunk, we'll probably never see this part becoming an issue. Anyhow, interesting to see that such constructs exist.

PR #1429, opened 2019-01-21
Aggregate: Use accessors, not []

mrks:
Also gets rid of a lot of those pesky performance warnings

No real performance impact because aggregate has not that much of an impact on performance. I account queries 15 and 16 to alignment variations (hey, but at least we get better on average):

```
+-----------+----------------+------+-----------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s   | runs | new iter/s      | runs | change | p-value (significant if <0.001) |
+-----------+----------------+------+-----------------+------+--------+---------------------------------+
| TPC-H 1   | 0.575067281723 | 35   | 0.58910048008   | 36   | +2%    |                          0.0023 |
| TPC-H 2   | 8.29356479645  | 498  | 8.4175786972    | 506  | +1%    |                          0.0001 |
| TPC-H 3   | 5.28543233871  | 318  | 5.46232700348   | 328  | +3%    |                          0.0000 |
| TPC-H 4   | 1.85177838802  | 112  | 1.87058711052   | 113  | +1%    |                          0.0003 |
| TPC-H 5   | 3.74189043045  | 225  | 3.79483795166   | 228  | +1%    |                          0.0002 |
| TPC-H 6   | 15.5908937454  | 936  | 15.5625419617   | 934  | -0%    |     (run time too short) 0.2915 |
| TPC-H 7   | 1.3831166029   | 83   | 1.37914168835   | 83   | -0%    |                          0.3846 |
| TPC-H 8   | 4.72634124756  | 284  | 4.7217874527    | 284  | -0%    |                          0.8526 |
| TPC-H 9   | 1.492192626    | 90   | 1.51469421387   | 91   | +2%    |                          0.0007 |
| TPC-H 10  | 2.4881169796   | 150  | 2.4813785553    | 149  | -0%    |     (run time too short) 0.2666 |
| TPC-H 11  | 32.8169555664  | 1970 | 33.907497406    | 2035 | +3%    |     (run time too short) 0.0000 |
| TPC-H 12  | 7.53452110291  | 453  | 7.70902585983   | 463  | +2%    |                          0.0000 |
| TPC-H 13  | 1.99485766888  | 120  | 1.97830247879   | 119  | -1%    |                          0.3752 |
| TPC-H 14  | 14.5876226425  | 876  | 14.9294900894   | 896  | +2%    |     (run time too short) 0.0000 |
| TPC-H 15  | 8.91453742981  | 535  | 12.4969348907   | 750  | +40%   |     (run time too short) 0.0000 |
| TPC-H 16  | 7.48697853088  | 450  | 6.45993232727   | 388  | -14%   |                          0.0000 |
| TPC-H 17  | 0.128018930554 | 8    | 0.126433804631  | 8    | -1%    |        (not enough runs) 0.7011 |
| TPC-H 18  | 1.30469536781  | 79   | 1.40639257431   | 85   | +8%    |                          0.0000 |
| TPC-H 19  | 5.38754940033  | 324  | 5.44963645935   | 327  | +1%    |     (run time too short) 0.0003 |
| TPC-H 20  | 0.065749540925 | 4    | 0.0680251866579 | 5    | +3%    |        (not enough runs) 0.0620 |
| TPC-H 21  | 0.119401492178 | 8    | 0.12438698858   | 8    | +4%    |        (not enough runs) 0.0863 |
| TPC-H 22  | 11.84998703    | 711  | 12.1899709702   | 732  | +3%    |     (run time too short) 0.0000 |
| average   |                |      |                 |      | +3%    |                                 |
+-----------+----------------+------+-----------------+------+--------+---------------------------------+
```


src/lib/operators/aggregate.cpp
```diff
@@ -702,15 +691,43 @@ std::enable_if_t<func == AggregateFunction::Avg && !std::is_arithmetic_v<Aggrega
...
+      using ColumnDataType = typename decltype(typed_value)::type;
+
+      auto values = pmr_concurrent_vector<ColumnDataType>(pos_list.size());
+      auto null_values = pmr_concurrent_vector<bool>(pos_list.size());
+      std::vector<std::unique_ptr<BaseSegmentAccessor<ColumnDataType>>> accessors(input_table->chunk_count());
+
+      auto output_offset = ChunkOffset{0};
+
+      for (const auto& row_id : pos_list) {
+        // pos_list was generated as part of the aggregate operator. While it might point to rows that contain NULL
```
Bouncner: We are the aggregate operator? Maybe a little bit more context?

src/lib/operators/aggregate.cpp
```diff
@@ -702,15 +691,43 @@ std::enable_if_t<func == AggregateFunction::Avg && !std::is_arithmetic_v<Aggrega
...
+        // values, no new NULL values should have been added.
+        DebugAssert(!row_id.is_null(), "Did not expect NULL value here");
+
+        auto& accessor = accessors[row_id.chunk_id];
+        if (!accessor) {
+          accessor =
+              create_segment_accessor<ColumnDataType>(input_table->get_chunk(row_id.chunk_id)->get_segment(column_id));
```
Bouncner: Why not add to `accessors`?

Bouncner: Further, would it be possible to sort the groups? We would not switch the accessors that often.

src/lib/operators/aggregate.cpp
```diff
@@ -702,15 +691,43 @@ std::enable_if_t<func == AggregateFunction::Avg && !std::is_arithmetic_v<Aggrega
...
+        }
+
+        const auto& optional_value = accessor->access(row_id.chunk_offset);
+        if (!optional_value) {
+          null_values[output_offset] = true;
```
Bouncner: Can we get rid of the branch?

`null_values[output_offset] = optional_value;`

Bensk1:
Regarding the alignment variations, I think I saw something similar as well yesterday. A clean build (`ninja clean` and `ccache --clear`) resolved the problem for me.

src/lib/operators/aggregate.cpp
mrks: We would still need the branch because `*opt` is undefined if `opt` is empty. Also, it would hide our intent.

src/lib/operators/aggregate.cpp
mrks: > Why not add to accessors?

But... that is what we do?

> Further, would it be possible to sort the groups?

At the cost of `n log n` indirect accesses to the result list. I have no intuition on what is faster, but this should only play a role for really big aggregate outputs. In the long run, we will have to rewrite this anyway.

Bouncner: > But... that is what we do?

Am I blind? I don't see it.

> At the cost of n log n indirect accesses to the result list.
Then never mind. I thought we can still proceed as now, but can better cache the accessors. If it introduces an indirection that does no longer sound like a good idea. :)

mrks: accessor is a reference into the vector.

src/lib/operators/aggregate.cpp
Bouncner: Well, a comment would sufficiently cover our intend and if the branch could be removed, I'd at least try it.

So something (forgive my C++ knowledge) like `null_values[output_offset] = static_cast<bool>(optional_value);` does not work?

mrks: But how would you write to `values`?

Bouncner: Ok, I see. Doing that would be some really nasty branchless code. Not worth it.

src/lib/operators/aggregate.cpp
Bouncner: Got it. *facepalm*

src/lib/operators/aggregate.cpp
mrks: should be addressed

PR #1430, opened 2019-01-22
Fix Pagination CTRL+C behaviour

### LINKED ISSUE #1426: Exiting Pagination with Ctrl+C breaks the console
Preferably, it would do the same as pressing [q]. Alternatively, it could do nothing. What it should not do: Break everything.

lanice:
Closes #1426.

Just ignoring CTRL+C during pagination is simple (first commit of this PR), we just don't handle the signal.
To make CTRL+C behave the same way as 'q' during pagination, I made it push the curses character code for CTRL+C onto the input stack (with ungetch()), and also add a generous timeout to the getch() loop, so that the simulated keypress actually gets captured without a second input.

A bit hacky, but since we catch the signal in the console, it is not passed on to the pagination window, therefore we have to manually push it back.


mrks:
ncurses is weird...

PR #1431, opened 2019-01-22
Fix Delete'ing from pruned Table

### LINKED ISSUE #1083: ChunkPruningRule leads to issues with Delete
If the ChunkPruningRule matches, it creates a new table to be used by GetTable and only the shared_ptrs to the matching chunks are included. As a result, this table is not the same as Delete retrieves from the StorageManager and this line fails:

```
if (table != first_column->referenced_table()) return false;
```

Also, row ids do not match anymore (because chunk ids don't) and cannot be used for deletion anymore.

An easy fix would be to exclude the rule for updates and deletes, but maybe someone can come up with a better one?

### LINKED ISSUE #1269: Run TPC-H/SQLite Tests with encoding as part of system tests.
Anyone against this? This will get some coverage on both encodings and chunk pruning

Make sure to test VectorCompression schemes as well. https://github.com/hyrise/hyrise/pull/1323#discussion_r235933637

mrzzzrm:
Fix #1083 
Closes #1269

The `Delete` operator can now work on "pruned" versions of stored Tables. Such Tables contain a subset of the Chunks of a stored Table and are generated if `GetTable` is instructed to exclude Chunks.

* Removes the "table_name" fields from `Delete` and `DeleteNode`
* Enables the `SQLiteTestrunner` for all encodings (and JIT + Dictionary for good measure)
* Fix a bug in the `SQLiteTestrunner` that made it not reset the SQLite version of a Table when running the same query for multiple encodings.

### DIFF
--- a/src/lib/logical_query_plan/lqp_utils.cpp
+++ b/src/lib/logical_query_plan/lqp_utils.cpp
@@ -193,8 +193,8 @@ std::set<std::string> lqp_find_modified_tables(const std::shared_ptr<AbstractLQP
         break;
       case LQPNodeType::Delete: {
         visit_lqp(node->left_input(), [&](const auto& sub_delete_node) {
-          if (const auto stored_stable_node = std::dynamic_pointer_cast<StoredTableNode>(sub_delete_node)) {
-            modified_tables.insert(stored_stable_node->table_name);
+          if (const auto stored_table_node = std::dynamic_pointer_cast<StoredTableNode>(sub_delete_node)) {
+            modified_tables.insert(stored_table_node->table_name);
           } else if (const auto mock_node = std::dynamic_pointer_cast<MockNode>(sub_delete_node)) {
             if (mock_node->name) {
               modified_tables.insert(*mock_node->name);
--- a/src/lib/logical_query_plan/lqp_utils.cpp
+++ b/src/lib/logical_query_plan/lqp_utils.cpp
@@ -7,7 +7,9 @@
 #include "logical_query_plan/abstract_lqp_node.hpp"
 #include "logical_query_plan/delete_node.hpp"
 #include "logical_query_plan/insert_node.hpp"
+#include "logical_query_plan/mock_node.hpp"
 #include "logical_query_plan/predicate_node.hpp"
+#include "logical_query_plan/stored_table_node.hpp"
 #include "logical_query_plan/union_node.hpp"
 #include "logical_query_plan/update_node.hpp"
 #include "utils/assert.hpp"
@@ -189,9 +191,18 @@ std::set<std::string> lqp_find_modified_tables(const std::shared_ptr<AbstractLQP
       case LQPNodeType::Update:
         modified_tables.insert(std::static_pointer_cast<UpdateNode>(node)->table_name);
         break;
-      case LQPNodeType::Delete:
-        modified_tables.insert(std::static_pointer_cast<DeleteNode>(node)->table_name);
-        break;
+      case LQPNodeType::Delete: {
+        visit_lqp(node->left_input(), [&](const auto& sub_delete_node) {
+          if (const auto stored_stable_node = std::dynamic_pointer_cast<StoredTableNode>(sub_delete_node)) {
+            modified_tables.insert(stored_stable_node->table_name);
+          } else if (const auto mock_node = std::dynamic_pointer_cast<MockNode>(sub_delete_node)) {
+            if (mock_node->name) {
+              modified_tables.insert(*mock_node->name);
+            }
+          }
+          return LQPVisitation::VisitInputs;
+        });
+      } break;
       case LQPNodeType::CreateTable:
       case LQPNodeType::CreatePreparedPlan:
       case LQPNodeType::DropTable:
--- a/src/lib/operators/delete.cpp
+++ b/src/lib/operators/delete.cpp
@@ -12,8 +12,8 @@
 
 namespace opossum {
 
-Delete::Delete(const std::shared_ptr<const AbstractOperator>& _referencing_table)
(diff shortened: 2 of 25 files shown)


src/lib/operators/delete.cpp
```diff
@@ -13,127 +13,144 @@
...
+            referenced_table,
+        "All segments in values_to_delete must reference the same table");
+    DebugAssert(chunk->references_exactly_one_table(),
+                "All segments in values_to_delete must reference the same table");
+
     const auto first_segment = std::static_pointer_cast<const ReferenceSegment>(chunk->get_segment(ColumnID{0}));
     const auto pos_list = first_segment->pos_list();
 
-    _pos_lists.emplace_back(pos_list);
+    DebugAssert(std::all_of(chunk->segments().begin(), chunk->segments().end(),
```
mrzzzrm: This is silly, tbh. values_to_delete can contain multiple columns, but we just use the leftmost. To avoid confusion, we check the all other columns have the same poslist as the leftmost column...

src/lib/operators/update.cpp
```diff
@@ -32,15 +32,6 @@ std::shared_ptr<const Table> Update::_on_execute(std::shared_ptr<TransactionCont
   DebugAssert(input_table_left()->column_data_types() == input_table_right()->column_data_types(),
               "Update required identical layouts from its input tables");
 
-  for (const auto& chunk : input_table_left()->chunks()) {
```
mrzzzrm: Delete performs these Checks anyway

mrzzzrm:
yep, thought about that as well... We can discuss this in person tomorrow.

mrks:
Why can't we get the table by following the referenced_table in the obsolete rows? We might need a `std::weak_ptr<Table> _unpruned_table` in there.

src/test/statistics/chunk_statistics/histograms/histogram_utils_test.cpp
```diff
@@ -203,6 +203,6 @@ TEST_F(HistogramUtilsTest, NextValueBruteForce) {
                    supported_characters, prefix_length);
     EXPECT_EQ(number_string, next_value_of_previous_number);
   }
-}
+}gin
```
mrks: Yes, please!

mrzzzrm:
I'm not sure whether removing table_name from Insert and Update as part of this PR is a good idea. It probably requires further discussion. Remember, we only have a `const Table` - we can "remove" rows since we can write to the MVCCData - but we can absolutely not ADD rows, as this requires a mutable table.

src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -189,9 +191,18 @@ std::set<std::string> lqp_find_modified_tables(const std::shared_ptr<AbstractLQP
       case LQPNodeType::Update:
         modified_tables.insert(std::static_pointer_cast<UpdateNode>(node)->table_name);
         break;
-      case LQPNodeType::Delete:
-        modified_tables.insert(std::static_pointer_cast<DeleteNode>(node)->table_name);
-        break;
+      case LQPNodeType::Delete: {
+        visit_lqp(node->left_input(), [&](const auto& sub_delete_node) {
+          if (const auto stored_stable_node = std::dynamic_pointer_cast<StoredTableNode>(sub_delete_node)) {
```
mrks: typo

src/lib/operators/delete.cpp
mrks: This is to avoid cases like `DELETE FROM t t1, t t2 WHERE t1.a = 3` where it is unclear what is deleted.

src/lib/operators/delete.cpp
```diff
@@ -12,93 +12,113 @@
...
-          mvcc_data->tids[row_id.chunk_offset] == _transaction_id) {
-        // Make sure that even we don't see it anymore
-        mvcc_data->tids[row_id.chunk_offset] = TransactionManager::INVALID_TRANSACTION_ID;
-        continue;
-      }
 
-      // the row is already locked by someone else and the transaction needs to be rolled back
-      _mark_as_failed();
-      return nullptr;
+      // Scope for the lock on the MVCC data
```
mrks: What do we need this for? The scope would be the same without the extra block.

src/lib/operators/delete.hpp
```diff
@@ -11,15 +11,11 @@
 namespace opossum {
 
 /**
- * Operator that deletes a number of rows from one table.
- * Expects a table with one chunk referencing only one table which
- * is passed via the AbstractOperator in the constructor.
- *
- * Assumption: The input has been validated before.
```
mrks: Why did you remove the assumption?

src/lib/operators/delete.cpp
mrks: We might end up with cases where all reference segments point to the same table but to different rows. I am not sure how we deal with that right now, but as an extra safety, this debug assert is there.

mrzzzrm: (What I meant with "silly" is just that the Delete has to deal with the input table containing multiple columns in the first place, despite it never using anything but the first column - we need a multi-line assert to ensure the validity of data (columns 1->n) that we never use anyway - that's "silly", imo)

src/lib/operators/delete.hpp
mrzzzrm: Accidentally. But it makes a nice point: Verbose commenting leads to the important parts being overlooked and potentially deleted.

src/lib/operators/delete.cpp
mrks: > despite it never using anything but the first column

I would rather argue that it uses *all* columns.

Do you have a different suggestion though?

mrzzzrm: Na, just made the original comment to ensure the reader that I placed an awkward multiline assert intentionally.

PR #1435, opened 2019-01-25
Rename SelectExpression to SubqueryExpression

mrzzzrm:
Ages ago @Bouncner pointed out that "SelectExpression" is not really a good name, since "Select" is a word that has meaning in SQL, but our LQP/PQPs should be ignorant about SQL.

So I finally renamed it them to "SubQueryExpressions", is that better?

Feel free to reject if this makes code in the making mergehell....


mrks:
Let's discuss this on Wednesday.

Bouncner:
We stumpled about this issue again with my DYOB group. Sub(q/Q)uery sounds much better for me. At least now I understand its usage in the log. plans I have written so far.
But: I don’t dare to comment on the general term since I don’t know about all its usages. Also: I am not fluent enough in all the relation/logical terms to know whether we might mix up things here (logical vs. SQL world)...

But you guys will surely solve that on Wednesday. :-)
I will pass due to vacation this week.

mrzzzrm:
Rename to SubqueryExpression

mrzzzrm:
Its LQPSubqueryExpression and PQPSubqueryExpression now.

PR #1438, opened 2019-01-28
Proper libHyrise linking for plugins

### LINKED ISSUE #1434: Loading Plugins in hyrisePlayground fails
I'm trying to use the PluginManager to load a plugin in hyrisePlayground. Therefore, I added `CMAKE_DL_LIBS` and `FILESYSTEM_LIBRARY` to my `bin/CMakeLists.txt`, and defined `TEST_PLUGIN_DIR` in the same way it is defined in `test/CMakeLists.txt`. My hyrisePlayground looks like this:
```cpp
#ifdef __APPLE__
#define DYNAMIC_LIBRARY_SUFFIX ".dylib"
#elif __linux__
#define DYNAMIC_LIBRARY_SUFFIX ".so"
#endif

int main() {
  auto& pm = PluginManager::get();
  pm.load_plugin(std::string(TEST_PLUGIN_DIR) + "libTestPlugin" + DYNAMIC_LIBRARY_SUFFIX);
  pm.unload_plugin("TestPlugin");
  return 0;
}
```

If I execute the playground, I get the following error (the displayed path to the plugin is correct):
```shell
  what():  src/lib/utils/plugin_manager.cpp:28 Loading plugin failed: /home/david/Studium/HPI/01/DYOD/hyrise/hyrise/cmake-build-debug/lib/libTestPlugin.so:
undefined symbol: _ZN7opossum14StorageManager9add_tableERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESt10shared_ptrINS_5TableEE
```

Is there anything I have missed in order to use the PluginManager properly?

Bensk1:
Fixes #1434

### DIFF
--- a/src/bin/CMakeLists.txt
+++ b/src/bin/CMakeLists.txt
@@ -1,3 +1,10 @@
+# -force_load/--whole-archive are necessary to include all symbols so that dynamically loaded plugins can use them
+if (APPLE)
+    set(HYRISE_LIBRARY_WITH_ALL_SYMBOLS -force_load hyrise)
+else()
+    set(HYRISE_LIBRARY_WITH_ALL_SYMBOLS -Wl,--whole-archive hyrise -Wl,--no-whole-archive)
+endif()
+
 # Configure server
 add_executable(
     hyriseServer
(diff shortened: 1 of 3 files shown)


mrks:
https://stackoverflow.com/a/3718867/2204581

Bensk1:
Yeah, sure. I just don't see how the changes should lead to these errors and thought you might have seen that before.

mrks:
Mmmh, something something initialization order? That would explain why it only happens at the end of the tests.

Bensk1:
@mrks could you give it another look?

PR #1379, opened 2018-12-19
Change worker sleep time from 10ms to exponential backoff.

lanice:
Previously, workers slept for a constant 10ms when no task is available. This is quite long sometimes, especially if there are many small tasks. This PR changes that to follow an exponential backoff algorithm, starting with a sleep of 1ns, doubling every time there still is no work, with an upper limit of 10ms.

TPC-H comparison (on the current master including #1322) with 224 cores, 100 clients, sf 1:
```
+-----------+------------------+-------+------------------+-------+--------+---------------------------------+
| Benchmark | prev. iter/s     | runs  | new iter/s       | runs  | change | p-value (significant if <0.001) |
+-----------+------------------+-------+------------------+-------+--------+---------------------------------+
| TPC-H 1   | 11.2330493927002 | 674   | 11.816180229187  | 709   | +5%    |                          0.0000 |
| TPC-H 2   | 5.46587657928467 | 328   | 27.3655071258545 | 1642  | +401%  |                          0.0000 |
| TPC-H 3   | 66.1910705566406 | 3972  | 68.91650390625   | 4135  | +4%    |                          0.0242 |
| TPC-H 4   | 49.5639915466309 | 2974  | 50.8428764343262 | 3051  | +3%    |                          0.2231 |
| TPC-H 5   | 66.8958892822266 | 4014  | 66.5590744018555 | 3994  | -1%    |                          0.6094 |
| TPC-H 6   | 494.452728271484 | 10000 | 484.125305175781 | 10003 | -2%    |                          0.0000 |
| TPC-H 7   | 46.0259552001953 | 2762  | 45.8783874511719 | 2753  | -0%    |                          0.8490 |
| TPC-H 8   | 84.7737579345703 | 5087  | 88.2499542236328 | 5295  | +4%    |                          0.0710 |
| TPC-H 9   | 37.1288223266602 | 2228  | 36.931209564209  | 2216  | -1%    |                          0.6146 |
| TPC-H 10  | 69.9747924804688 | 4199  | 67.0482330322266 | 4023  | -4%    |                          0.0000 |
| TPC-H 11  | 796.446228027344 | 10004 | 1154.39624023438 | 10000 | +45%   |                          0.0000 |
| TPC-H 12  | 233.180847167969 | 10002 | 242.003021240234 | 10001 | +4%    |                          0.0000 |
| TPC-H 13  | 17.8472366333008 | 1071  | 14.8499202728271 | 891   | -17%   |                          0.0000 |
[...]

PR #1442, opened 2019-01-28
Worker sleep times (Follow-up on #1379)

toniSta:
I'm currently benchmarking the pipeline creation for my thesis. I noticed a big spreading of execution times (min: 78 µs, median: 2300 µs, max: 18000 µs for query `select 1;`). The strongly fluctuating times are a result of the worker sleep time of 10 ms. As discussed in PR #1379 we actually don't want to have a sleep timer for the workers. Hence, I implemented a solution using `std::condition_variable`s. I changed the current implementation in such a way, that the node's task queue notifies a worker (on the same node) as soon as a new job gets pushed into the queue. Work stealing is performed by waking up each worker every 100µs (debatable).

However, benchmark performance on numa systems using many clients will **decrease** tremendously (see screenshot below). This is because of the bad job scheduling behavior. Tasks are currently scheduled to Node 0 only (Correct me if I'm wrong or I missed a configuration option). All other nodes are performing work stealing only.

Comparison current master and this PR (SF1, 100 clients, 240 cores, 8 nodes)
screenshot from 2019-01-28 17-23-44

When we only use one node of the system (so no numa) we get the following results (SF1, 1 client, 30 cores, 1 node)
screenshot from 2019-01-28 17-25-56

Any thoughts about this?

### DIFF
--- a/src/lib/scheduler/worker.cpp
+++ b/src/lib/scheduler/worker.cpp
@@ -91,10 +91,6 @@ void Worker::start() { _thread = std::thread(&Worker::operator(), this); }
 
 void Worker::join() {
   Assert(!CurrentScheduler::get()->active(), "Worker can't be join()-ed while the scheduler is still active");
-  // The scheduler will only shut down if all tasks within the queue are done. After that, the worker threads will be
-  // stuck in the while loop waiting for a new task. Hence, we notify all workers, so that it runs the loop once.
-  // Since the loop condition will no longer be true, the worker is going to shutdown.
-  _queue->new_task.notify_all();
   _thread.join();
 }
 
--- a/src/lib/scheduler/worker.cpp
+++ b/src/lib/scheduler/worker.cpp
@@ -25,6 +25,9 @@ namespace {
 thread_local std::weak_ptr<opossum::Worker> this_thread_worker;
 }  // namespace
 
+// The sleep time was determined experimentally
+static constexpr auto WORKER_SLEEP_TIME = std::chrono::microseconds(300);
+
 namespace opossum {
 
 std::shared_ptr<Worker> Worker::get_this_thread_worker() { return ::this_thread_worker.lock(); }
@@ -74,7 +77,7 @@ void Worker::_work() {
     if (!work_stealing_successful) {
       {
         std::unique_lock<std::mutex> unique_lock(_queue->lock);
-        _queue->new_task.wait_for(unique_lock, std::chrono::microseconds(200));
+        _queue->new_task.wait_for(unique_lock, WORKER_SLEEP_TIME);
       }
       return;
     }
--- a/src/lib/scheduler/worker.cpp
+++ b/src/lib/scheduler/worker.cpp
@@ -45,20 +45,12 @@ void Worker::operator()() {
 
   _set_affinity();
 
-  std::mutex lock;
-  std::unique_lock<std::mutex> unique_lock(lock);
-  bool queues_are_empty;
-
   while (CurrentScheduler::get()->active()) {
-    queues_are_empty = _work();
-    if (queues_are_empty) {
-      // Wait for new task pushed to the current node's task queue or perform work stealing after 100 µs.
-      _queue->new_task.wait_for(unique_lock, std::chrono::microseconds(100));
-    }
+    _work();
   }
 }
 
-bool Worker::_work() {
+void Worker::_work() {
   auto task = _queue->pull();
 
   if (!task) {
@@ -77,9 +69,14 @@ bool Worker::_work() {
       }
     }
 
-    // Return if there is no ready task in our queue and work stealing was not successful.
+    // If there is no ready task neither in our queue nor in any other, worker waits for a new task to be pushed to the
+    // own queue or returns after timer exceeded (whatever occurs first).
     if (!work_stealing_successful) {
-      return true;
+      {
+        std::unique_lock<std::mutex> unique_lock(_queue->lock);
+        _queue->new_task.wait_for(unique_lock, std::chrono::microseconds(200));
+      }
+      return;
     }
   }
 
@@ -88,7 +85,6 @@ bool Worker::_work() {
   // This is part of the Scheduler shutdown system. Count the number of tasks a Worker executed to allow the
   // Scheduler to determine whether all tasks finished
   _num_finished_tasks++;
-  return false;
 }
 
 void Worker::start() { _thread = std::thread(&Worker::operator(), this); }
--- a/src/lib/scheduler/worker.cpp
+++ b/src/lib/scheduler/worker.cpp
@@ -49,7 +49,8 @@ void Worker::operator()() {
   std::unique_lock<std::mutex> unique_lock(lock);
 
   while (CurrentScheduler::get()->active()) {
-    _queue->new_task.wait(unique_lock);
+    // Wait for new task pushed to the current node's task queue or perform work stealing after 100 µs.
+    _queue->new_task.wait_for(unique_lock, std::chrono::microseconds(100));
     _work();
   }
 }
@@ -58,7 +59,25 @@ void Worker::_work() {
   auto task = _queue->pull();
 
   if (!task) {
-    return;
+    // Simple work stealing without explicitly transferring data between nodes.
+    auto work_stealing_successful = false;
+    for (auto& queue : CurrentScheduler::get()->queues()) {
+      if (queue == _queue) {
+        continue;
+      }
+
+      task = queue->steal();
+      if (task) {
+        task->set_node_id(_queue->node_id());
+        work_stealing_successful = true;
+        break;
+      }
+    }
+
+    // Return if there is no ready task in our queue and work stealing was not successful.
+    if (!work_stealing_successful) {
+      return;
+    }
   }
 
   task->execute();
--- a/src/lib/scheduler/worker.cpp
+++ b/src/lib/scheduler/worker.cpp
@@ -47,15 +47,18 @@ void Worker::operator()() {
 
   std::mutex lock;
   std::unique_lock<std::mutex> unique_lock(lock);
+  bool queues_are_empty;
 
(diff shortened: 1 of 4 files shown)


mrks:
ping @lanice

lanice:
One tip on benchmarking with many (e.g., 100) clients: If you want to save some time, skip queries 17, 20, 21. Their runtime is so bad, that it REALLY slows down "quick" benchmarks when having more that only 1 or a few clients, because for each client, the query has to finish after the time is up.

lanice:
And here a comparison with all 4 nodes (224 cores), sf 1, but only 1 client (left out queries 17, 20, 21, for mentioned reasons):
```
+-----------+-------------------+------+-------------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s      | runs | new iter/s        | runs | change | p-value (significant if <0.001) |
+-----------+-------------------+------+-------------------+------+--------+---------------------------------+
| TPC-H 1   | 0.399995595216751 | 24   | 0.316662728786469 | 19   | -21%   |     (run time too short) 0.0000 |
| TPC-H 2   | 0.266653656959534 | 16   | 2.78317260742188  | 167  | +944%  |     (run time too short) 0.0000 |
| TPC-H 3   | 5.44972991943359  | 327  | 10.7325420379639  | 644  | +97%   |     (run time too short) 0.0000 |
| TPC-H 4   | 4.18329334259033  | 251  | 4.8663501739502   | 292  | +16%   |     (run time too short) 0.0000 |
| TPC-H 5   | 3.49971985816956  | 210  | 4.34943723678589  | 261  | +24%   |     (run time too short) 0.0000 |
| TPC-H 6   | 14.3318510055542  | 860  | 33.6994552612305  | 2022 | +135%  |     (run time too short) 0.0000 |
| TPC-H 7   | 1.96661972999573  | 118  | 2.19987535476685  | 132  | +12%   |     (run time too short) 0.0000 |
| TPC-H 8   | 4.1665358543396   | 250  | 5.09921360015869  | 306  | +22%   |     (run time too short) 0.0000 |
| TPC-H 9   | 1.86657476425171  | 112  | 2.46627449989319  | 148  | +32%   |     (run time too short) 0.0000 |
| TPC-H 10  | 1.89996516704559  | 114  | 2.0663685798645   | 124  | +9%    |     (run time too short) 0.0000 |
| TPC-H 11  | 13.5992755889893  | 816  | 20.0324096679688  | 1202 | +47%   |     (run time too short) 0.0000 |
| TPC-H 12  | 7.33300399780273  | 440  | 26.681755065918   | 1601 | +264%  |     (run time too short) 0.0000 |
| TPC-H 13  | 3.16633725166321  | 190  | 3.11614966392517  | 187  | -2%    |     (run time too short) 0.0010 |
| TPC-H 14  | 9.04871463775635  | 543  | 21.3488864898682  | 1281 | +136%  |     (run time too short) 0.0000 |
| TPC-H 16  | 5.03321647644043  | 302  | 6.04925727844238  | 363  | +20%   |     (run time too short) 0.0000 |
| TPC-H 18  | 1.44988059997559  | 87   | 1.11648499965668  | 67   | -23%   |     (run time too short) 0.0000 |
| TPC-H 19  | 9.53213310241699  | 572  | 45.7123947143555  | 2743 | +380%  |     (run time too short) 0.0000 |
| TPC-H 22  | 12.3163185119629  | 739  | 21.7301216125488  | 1304 | +76%   |     (run time too short) 0.0000 |
| average   |                   |      |                   |      | +121%  |                                 |
+-----------+-------------------+------+-------------------+------+--------+---------------------------------+
```

As you can see, with only 1 client, your changes cause huge boost in performance even when using all 4 nodes.

So I don't think the numa nodes cause the problem here, it seems to be the number of concurrently running queries/tasks (as configured through the number of clients). Why your changes perform so poorly with many clients, I am not quite sure.

toniSta:
Good point. You are right. Let me change and benchmark it.

Besides that issue, wouldn't it be great to have some sort of round robin scheduling as default?

lanice:
> Besides that issue, wouldn't it be great to have some sort of round robin scheduling as default?

Sure, I have nothing against it. Make sure to benchmark that too, and see if there is any potential performance impact, e.g. when using an atomic counter for that. I don't think it would be a problem, but better be sure.

toniSta:
Benchmark performance hasn't changed significantly after my latest change. 
screenshot from 2019-01-29 15-34-09

Also applying round robin scheduling does't make a big difference.

src/lib/scheduler/worker.cpp
```diff
@@ -44,12 +45,20 @@ void Worker::operator()() {
...
+  std::mutex lock;
+  std::unique_lock<std::mutex> unique_lock(lock);
+  bool queues_are_empty;
+
   while (CurrentScheduler::get()->active()) {
-    _work();
+    queues_are_empty = _work();
+    if (queues_are_empty) {
+      // Wait for new task pushed to the current node's task queue or perform work stealing after 100 µs.
+      _queue->new_task.wait_for(unique_lock, std::chrono::microseconds(100));
```
mrks: I haven't worked with condition variables for a while, but are you sure that you are using them correctly?

> **std::condition_variable::wait_for**
> [...]
> Calling this function if lock.mutex() is not the same mutex as the one used by all other threads that are currently waiting on the same condition variable is undefined behavior.

src/lib/scheduler/worker.cpp
```diff
@@ -44,12 +45,20 @@ void Worker::operator()() {
 
   _set_affinity();
 
+  std::mutex lock;
+  std::unique_lock<std::mutex> unique_lock(lock);
+  bool queues_are_empty;
```
mrks: Why queue*s*? I only see a single one.

toniSta: Because, I am also checking the queues not located on the current node (work stealing).

mrks:
Maybe track how many tasks were stolen vs. executed on the "correct" node. If the amount of stealing increased due to your change, that could explain at least part of the difference.

src/lib/scheduler/worker.cpp
toniSta: You are right, I am using the wrong. I missed that note.

mrks: Maybe also consider if you can reduce the time the lock is held.

src/lib/scheduler/worker.cpp
```diff
@@ -68,10 +77,9 @@ void Worker::_work() {
       }
     }
 
-    // Sleep if there is no ready task in our queue and work stealing was not successful.
+    // Return if there is no ready task in our queue and work stealing was not successful.
     if (!work_stealing_successful) {
-      std::this_thread::sleep_for(std::chrono::milliseconds(10));
-      return;
+      return true;
```
lanice: I would maybe move the whole wait-for-condition-variable-thing to here, where previously the sleeping happened. That way you don't actually have to check whether the queues are empty or not, because that is implicit: If we reached this point, then a) there was no task in my queue, and b) we could not steal a task. Therefore we then go into wait mode until notified again.

toniSta:
Applied the feedback now. Overall performance gain is 17% now. However, some queries - especially TPC-H 4 still have a performance decrease.

Comparison to master (SF1, 100 clients, 240 cores, 8 nodes)
screenshot from 2019-01-30 16-30-15

mrks:
Surprisingly, the approach in #1379 still produces better numbers. Any idea why that might be the case?

toniSta:
Could that be a server related issue? I ran my benchmarks on rapa using clang compiler. I just checked out #1379, merged the latest master into it and ran the benchmarks. (SF1, 100 clients)

Master - Exponential backoff

screenshot from 2019-01-31 12-18-25

Exponential backoff - this PR
screenshot from 2019-01-31 12-18-50

lanice:
Another thing to note, we are waiting at least 200 microseconds now before trying to work steal again after it did not succeed, correct? Did you try different values there and see if it makes any difference? Like 10 nanoseconds for example, just to check if that is a delay.

mrks:
Convinced.

src/lib/scheduler/worker.cpp
```diff
@@ -68,9 +69,13 @@ void Worker::_work() {
...
     }
 
-    // Sleep if there is no ready task in our queue and work stealing was not successful.
+    // If there is no ready task neither in our queue nor in any other, worker waits for a new task to be pushed to the
+    // own queue or returns after timer exceeded (whatever occurs first).
     if (!work_stealing_successful) {
-      std::this_thread::sleep_for(std::chrono::milliseconds(10));
+      {
+        std::unique_lock<std::mutex> unique_lock(_queue->lock);
+        _queue->new_task.wait_for(unique_lock, std::chrono::microseconds(200));
```
mrks: Could you move that into a `static constexpr` so that it stands out as a configuration parameter?

Example: `static constexpr auto DEFAULT_LIKE_SELECTIVITY = 0.1f;`

mrks:
ping?

mrks:
Is this ready to merge?

toniSta:
I played around with different sleep times. 300µs seems to be the best performing value. Sleeping less than 50µs does not make sense at all. I benchmarked the execution of both 
`_queue->new_task.wait_for(unique_lock, std::chrono::nanoseconds(1));` and `std::this_thread::sleep_for(std::chrono::nanoseconds(1));`. Due to the syscall performed this "sleep" takes about 50µs.

Also a combination of the exponential back-off and the condition variable does not make a significant change.

Furthermore, I think benchmarking 100 clients does not make sense because workers barely go to sleep. Hence, I ran benchmarks once again using only one client. Here are the results (I can't explain results of TPC-H 2):

(SF1, 240 cores, 8 nodes)
screenshot from 2019-02-05 14-46-50

toniSta:
> Is this ready to merge?

I think so

mrks:
Nais. Don't worry about 2 and 15 too much. They like to step out of line. :/

PR #1444, opened 2019-01-28
Fix nullability of columns from null-supplying side of outer joins

### LINKED ISSUE #1051: LQP JoinNode returns wrong column definitions for outer joins
Output columns of joins can become `nullable`, if the join type is `outer`, `left` or `right`.
This behaviour is not implemented in the LQP `JoinNode`.
The current implementation of only the LQP `JoinNode` only copies the column definitions and does not change `nullable` behaviour of the column, if necessary.
I have not found a trivial solution for this problem as a column definition is represented by an `Expression` and I think that we currently cannot just create a column definition from scratch. I thought that this could be realised with a `MockNode` but it does not support `nullable`.

Bug found via running hyriseSystemTest with the jittable operators.
This bug was not found before as the normal hyrise operators take the column input definitions from the input table in comparison to the `JitAwareLQPTranslator`, which creates already specified type operators.

mrzzzrm:
Fix #1051


mrzzzrm:
@mrks @FabianWiebe 

I'd like to get rid of the concept of `Expression::is_nullable()`/`AbstractLQPNode::is_column_nullable()` altogether. Due to the outer joins, there is no practical value in `Expression::is_nullable` as we can not use it to determine whether a **column** is nullable. Additionally, it's tricky to determine whether a LQP column is nullable or not. But, this is my main point: we actually do not need this information at all. I refactored the ExpressionEvaluator and Projection so that they do no use `Expression::is_nullable` anymore. The only place where we need to determine the nullability of an expression/lqp column is the `JitAwareLQPTranslator` (see `is_column_nullable`):

```c++
std::shared_ptr<const JitExpression> JitAwareLQPTranslator::_try_translate_expression_to_jit_expression(
    const AbstractExpression& expression, JitReadTuples& jit_source,
    const std::shared_ptr<AbstractLQPNode>& input_node) const {
  const auto input_node_column_id = input_node->find_column_id(expression);
  if (input_node_column_id) {
    const auto tuple_value = jit_source.add_input_column(
        expression.data_type(), input_node->is_column_nullable(input_node->get_column_id(expression)),
        *input_node_column_id);
    return std::make_shared<JitExpression>(tuple_value);
  }
```

If we could modify the JIT to delay requiring the nullability of a column until the input table is present at execution time (then we can use `Table::column_is_nullable()`) we can do away with the concept of nullable Expressions.

The question is: can we? Does the JIT inherently have to know the nullability of a column at lqp translation time?

FabianWiebe:
> The question is: Does the JIT inherently have to know the nullability of a column at lqp translation time?

No, it is theoretically possible that the `JitAwareLQPTranslator` does not need this information. This would require changes within the `JitAwareLQPTranslator` and the jittable operators.
The big drawback of this approach is that it requires that a query can only be directly compiled before its execution. It makes future enhancements such as compiling a query-specific code for a stored procedure impossible.

I currently cannot see the benefit of why the expressions should not provide this information any longer. Maybe this information does not need to be stored within the expressions but could be computed recursively if required?
Should we maybe discuss this tomorrow at the meeting?

mrzzzrm:
@mrks (@FabianWiebe) ready for review. One ugliness I had to leave in: `CorrelatedParameterExpression` are always nullable (for now). Correctly determining the nullability of those is techically possible, but would require literally a hundred places to be changed where we construct such expressions, and they would all look like this: `correlated_parameter_(ParameterID{0}, *referenced_expression, *lqp_on_which_the_select_expression_to_which_this_is_a_parameter_is_executed)`.

src/lib/expression/arithmetic_expression.cpp
```diff
@@ -56,18 +56,18 @@ std::string ArithmeticExpression::as_column_name() const {
...
 bool ArithmeticExpression::_shallow_equals(const AbstractExpression& expression) const {
   return arithmetic_operator == static_cast<const ArithmeticExpression&>(expression).arithmetic_operator;
 }
 
 size_t ArithmeticExpression::_on_hash() const { return boost::hash_value(static_cast<size_t>(arithmetic_operator)); }
 
+bool ArithmeticExpression::_on_is_nullable_on_lqp(const AbstractLQPNode& lqp) const {
+  // We return NULL for divisions/modulo by 0
+  return AbstractExpression::_on_is_nullable_on_lqp(lqp) || arithmetic_operator == ArithmeticOperator::Division ||
+         arithmetic_operator == ArithmeticOperator::Modulo;
```
FabianWiebe: Is this the expected behaviour - that division and modulo result can result in null? If yes, this needs to be incorporated into the jittable operators.

src/lib/expression/lqp_select_expression.hpp
```diff
@@ -40,6 +39,8 @@ class LQPSelectExpression : public AbstractExpression {
  protected:
   bool _shallow_equals(const AbstractExpression& expression) const override;
   size_t _on_hash() const override;
+  // "lqp_" with underscore to avoid collision with member "lqp"
```
FabianWiebe: `lqp_` in comment is not found in code.

src/lib/expression/arithmetic_expression.cpp
mrzzzrm: Jep, using sqlite as a reference:

```
sqlite> SELECT 3/0 IS NULL;
1
sqlite> SELECT 3%0 IS NULL;
1
```

src/lib/expression/abstract_expression.hpp
```diff
@@ -73,9 +75,10 @@ class AbstractExpression : public std::enable_shared_from_this<AbstractExpressio
   virtual DataType data_type() const = 0;
 
   /**
-   * @return whether the result of the Expression MAY contain a NULL
+   * @return    whether an expression, executed on the output of a plan, would
+   *            produce a nullable result.
```
mrks: Could you doc why this needs the LQP?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -208,6 +209,12 @@ ColumnID AbstractLQPNode::get_column_id(const AbstractExpression& expression) co
   return *column_id;
 }
 
+bool AbstractLQPNode::is_column_nullable(const ColumnID column_id) const {
+  // Default behaviour: Forward from input
+  Assert(left_input() && !right_input(), "Can only from input iff there is a left input and no right input");
```
mrks: can only... what?

src/lib/expression/abstract_expression.hpp
mrzzzrm: Added 
```
   * @note      Determining whether an Expression (isolated from a plan to execute it on) is nullable is intentionally
   *            not supported: This is because "expression X is nullable" does NOT always equal "a column containing
   *            expression X is nullable": An expression `a + 5` (with `a` being a non-nullable column),
   *            e.g., is actually nullable if `a` comes from the null-supplying side of an outer join.
```

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrzzzrm: covfefe

mrzzzrm: "Can forward nullability from input iff there is a left input and no right input"

PR #1446, opened 2019-01-29
Add jittable Limit Operator

FabianWiebe:
This pr introduces a jittable Limit operator.
The `JitAwareLQPTranslator` adds the jittable Limit operator to the jittable operator pipeline if a Limit exists in the lqp.
The row count expression of the limit is evaluated before execution by the first operator in the pipeline, `JitReadTuples` operator.

### DIFF
--- a/src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
+++ b/src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
@@ -304,10 +304,12 @@ bool JitAwareLQPTranslator::_node_is_jittable(const std::shared_ptr<AbstractLQPN
           return aggregate_expression->aggregate_function == AggregateFunction::CountDistinct ||
                  aggregate_expression->arguments.empty();
         });
+    // Aggregate must be the last node in the jittable operator pipeline. Hence, the the root node in the lpp.
     return is_root_node && !has_unsupported_aggregate;
   }
 
   if (node->type == LQPNodeType::Limit) {
+    // Limit must be the last node in the jittable operator pipeline. Hence, the the root node in the lpp.
     return is_root_node;
   }
 
--- a/src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
+++ b/src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
@@ -15,6 +15,7 @@
 #include "expression/lqp_column_expression.hpp"
 #include "expression/value_expression.hpp"
 #include "logical_query_plan/aggregate_node.hpp"
+#include "logical_query_plan/limit_node.hpp"
 #include "logical_query_plan/lqp_utils.hpp"
 #include "logical_query_plan/predicate_node.hpp"
 #include "logical_query_plan/projection_node.hpp"
@@ -22,6 +23,7 @@
 #include "operators/jit_operator/operators/jit_aggregate.hpp"
 #include "operators/jit_operator/operators/jit_compute.hpp"
 #include "operators/jit_operator/operators/jit_filter.hpp"
+#include "operators/jit_operator/operators/jit_limit.hpp"
 #include "operators/jit_operator/operators/jit_read_tuples.hpp"
 #include "operators/jit_operator/operators/jit_validate.hpp"
 #include "operators/jit_operator/operators/jit_write_tuples.hpp"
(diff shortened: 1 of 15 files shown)


src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
```diff
@@ -116,11 +118,18 @@ std::shared_ptr<JitOperatorWrapper> JitAwareLQPTranslator::_try_translate_sub_pl
     return nullptr;
   }
 
+  // limit can only be the root node
+  const bool use_limit = node->type == LQPNodeType::Limit;
+  std::shared_ptr<AbstractExpression> row_count_expression = nullptr;
+  if (use_limit) {
+    row_count_expression = std::static_pointer_cast<LimitNode>(node)->num_rows_expression();
```
mrks: Is the translation part tested anywhere?

src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
```diff
@@ -293,7 +304,11 @@ bool JitAwareLQPTranslator::_node_is_jittable(const std::shared_ptr<AbstractLQPN
           return aggregate_expression->aggregate_function == AggregateFunction::CountDistinct ||
                  aggregate_expression->arguments.empty();
         });
-    return allow_aggregate_node && !has_unsupported_aggregate;
+    return is_root_node && !has_unsupported_aggregate;
+  }
+
+  if (node->type == LQPNodeType::Limit) {
+    return is_root_node;
```
mrks: Short comment?

src/lib/operators/jit_operator/operators/jit_limit.cpp
```diff
@@ -0,0 +1,15 @@
+#include "jit_limit.hpp"
+
+namespace opossum {
+
+std::string JitLimit::description() const { return "[Limit]"; }
+
+void JitLimit::_consume(JitRuntimeContext& context) const {
+  DebugAssert(context.limit_rows > 0, "JitLimit should not be called with limit_rows = 0");
+  if (--context.limit_rows == 0) {
+    context.chunk_size = 0;
```
mrks: Comment?

src/lib/operators/jit_operator/operators/jit_read_tuples.cpp
```diff
@@ -40,6 +42,18 @@ void JitReadTuples::before_query(const Table& in_table, JitRuntimeContext& conte
       });
     }
   }
+
+  if (_row_count_expression) {
```
mrks: Do I understand this block correctly in that is has nothing to do with reading the tuples but you needed a place in which to evaluate the limit expression?

src/test/operators/jit_operator/operators/jit_limit_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+TEST_F(JitLimitTest, FiltersTuplesAccordingToLimit) {
+  const uint32_t chunk_size{123};
+
+  JitRuntimeContext context;
+  context.chunk_size = chunk_size;
+  context.limit_rows = 2;
+
+  auto source = std::make_shared<MockSource>();
+  auto limit = std::make_shared<JitLimit>();
+  auto sink = std::make_shared<MockSink>();
```
mrks: Would it make sense for the sink to count the number of tuples it received?

src/lib/operators/jit_operator/operators/jit_read_tuples.cpp
FabianWiebe: Yes, the alternative approach would be to call a `before_query` function on every operator before execution. Then this code could be placed within the `JitLimit` operator. Should I change it to this approach that.

mrks: No, just wanted to make sure I understood it correctly. A comment should be sufficient.

src/test/operators/jit_operator/operators/jit_limit_test.cpp
FabianWiebe: That would be possible. There are different possibilities on how to implement a Limit operator. I chose the option to decrement the counter within the Limit operator as this only requires one access to the context (`--context.limit_rows`) for every tuple.

mrks: I mean just for the test.

PR #1449, opened 2019-01-31
Fix jemalloc linking under OS X

### LINKED ISSUE #1448: jemalloc linking problem on macOS?
After getting the latest master, hyriseTest and hyriseConsole report:

```
dyld: Library not loaded: /usr/local/lib/libjemalloc.2.dylib
  Referenced from: /Users/jan/Coding/hyrise_alter/./cmake-build-debug/hyriseTest
  Reason: image not found
[1]    89422 abort      ./cmake-build-debug/hyriseTest
```

when executed on macOS. This does also happen on a freshly checkout version.

mrks:
@Bensk1 - please try this.

fixes #1448


src/lib/CMakeLists.txt
```diff
@@ -592,7 +592,7 @@ set(
     sqlparser
     cqf
     uninitialized_vector
-    jemalloc
+    custom_jemalloc
```
mrks: This doesn't really *change* anything, but I wanted to make it clear that we are using our own one.

Bensk1:
That does solve it!

mrks:
I believe I called `ninja jemalloc` once - maybe that is why it was installed on my machine. I changed it to `/usr/bin/false` though so that people can realize that something is wrong.

lanice:
> I changed it to /usr/bin/false though so that people can realize that something is wrong.

Sounds good.

PR #1453, opened 2019-02-01
Add TPC-H example query file

### LINKED ISSUE #1411: TPC-H Query instance for testing/demo
Since 0bc175be95 tpch_queries.cpp does not contain actual TPC-H queries that can be executed in the console anymore. This was really handy for demos or to try out ideas. However, it's not always obvious which parameters make sense to fill in to form reasonably-sized result sets.

Any ideas whether/how we want to handle that? We could provide an example TPC-H query file or add a commented out executable version for each query to tpch_queries.cpp.

Bensk1:
Fixes #1411. If we decide to merge this, I will add a note to https://github.com/hyrise/hyrise/wiki/TPCH as well.

### DIFF
--- a/resources/tpch_validation_queries.sql
+++ b/resources/tpch_validation_queries.sql
@@ -1,0 +1,763 @@
+/**
+ * TPC-H 1
+ *
+ * Original:
+ *
+ * SELECT
+ *      l_returnflag,
+ *      l_linestatus,
+ *      sum(l_quantity) as sum_qty,
+ *      sum(l_extendedprice) as sum_base_price,
+ *      sum(l_extendedprice*(1-l_discount)) as sum_disc_price,
+ *      sum(l_extendedprice*(1-l_discount)*(1+l_tax)) as sum_charge,
+ *      avg(l_quantity) as avg_qty,
+ *      avg(l_extendedprice) as avg_price,
+ *      avg(l_discount) as avg_disc,
+ *      count(*) as count_order
+ * FROM
+ *      lineitem
+ * WHERE
+ *      l_shipdate <= date '1998-12-01' - interval '[DELTA]' day (3)
+ * GROUP BY
+ *      l_returnflag, l_linestatus
+ * ORDER BY
+ *      l_returnflag, l_linestatus
+ *
+ * Changes:
+ *  1. dates are not supported
+ *    a. use strings as data type for now
+ *    b. pre-calculate date operation
+ *  2. implicit type conversions for arithmetic operations are not supported
+ *    a. changed 1 to 1.0 explicitly
+ */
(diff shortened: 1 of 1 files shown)


resources/tpch_validation_queries.sql
```diff
@@ -0,0 +1,763 @@
...
+ * Changes:
+ *  1. Random values are hardcoded
+ *  2. dates are not supported
+ *    a. use strings as data type for now
+ *    b. pre-calculate date operation
+ */
+SELECT o_orderpriority, count(*) as order_count FROM orders WHERE o_orderdate >= '1996-07-01' AND o_orderdate < '1996-10-01' AND exists (SELECT *FROM lineitem WHERE l_orderkey = o_orderkey AND l_commitdate < l_receiptdate) GROUP BY o_orderpriority ORDER BY o_orderpriority;
+
+
+/**
```
mrzzzrm: can you make these sql-style comments so that this file is valid sql?

mrks: Haha, now we have two conflicting ideas on what this should look like. I seem to be missing in advantage of having this as a sql file.

mrzzzrm: image

I thought we consented on an SQL file? I do not have strong opinions here, but as a quick resource for c/p a non-cpp file seems lees intimidating, no?

mrks: Ah, ok. I understood this as a file with SQL commands, not necessarily an .sql file. In the end, it probably doesn't make that much of a difference anyway.

Bensk1: > can you make these c++-comments valid sql-style comments so that this file is valid sql?

Done. Still, we can't just load the file via `script resources/tpch_validation_queries.sql` in the console because there seems to be some newline handling issue.

mrks:
Could you add a comment to the top explaining the purpose of the file though and saying that it is not really "used" anywhere?

resources/tpch_validation_queries.sql
mrzzzrm: Open issue over at hyrise/sql-parser?

Bensk1: I don't know whether this is a parser issue or a problem with the console. I will open an issue after finding out.

Bensk1: This fails:
```
-- Comment

SELECT * FROM nation;
```

while this works:
```
-- Comment
SELECT * FROM nation;
```

I don't know if this is the intended parser bevaiour. I understand that blank lines in a statement are a problem. Blank lines between statements should not in my opinion.

Bouncner:
> I will add a note to https://github.com/hyrise/hyrise/wiki/TPCH as well

:)

Bensk1:
> > I will add a note to https://github.com/hyrise/hyrise/wiki/TPCH as well
> 
> @Bensk1 :)

Done

PR #1458, opened 2019-02-02
Rewrite IN expression to equality check if list has a single element

### LINKED ISSUE #1304: Rewrite trivial IN predicate to equals
Because IN is slow:

image

mrzzzrm:
Fix #1304 

I split up the `LogicalReductionRule` into a `ExpressionReductionRule` (which now covers the Distributivity Reduction, the IN-rewrite introduced by this PR and the former `ConstantCalculationRule`) and a `PredicateSplitUpRule` (which splits `PredicateNodes` with conjunctive chains into multiple `PredicateNodes`). The `ExpressionReductionRule` is supposed to be a general rule covering rewrites of Expressions - I'd rather not put each technique into its own rule and have a whole LQP traversal for each of them.


src/lib/optimizer/optimizer.cpp
```diff
@@ -105,7 +105,7 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {
   // for now we want the PredicateReorderingRule to have the final say on predicate positions
   optimizer->add_rule(std::make_shared<PredicatePlacementRule>());
 
-  // Bring predicates into the desired order once the PredicateReorderingRule has positioned them as desired
+  // Bring predicates into the desired order once the PredicatePlacementRule has positioned them as desired
   optimizer->add_rule(std::make_shared<PredicateReorderingRule>());
```
Bouncner: Not part of this PR, but why do we place the `IndexScanRule` behind the predicate reordering rule? @SvenLehmann : something to address?

SvenLehmann: I don't really like either way with Hyrise's current functionality.

If we first run the PredicateReorderingRule, we might miss a better order.
If we, however, first run the IndexScanRule and then reorder the nodes, we run into problems as IndexScans are only supported for non-referenced segments. Thus, we would have to check after reordering, whether there is an IndexScan that is run on referenced segments and rewrite it to a TableScan again. And if we revert the choice of using an Index, this might result in different costs for the chain of predicates. So reordering would be necessary again.

There are two solutions coming to my mind right now.
1. We could fix the IndexScan to handle ReferenceSegments (no idea whether this is desirable or what the design decisions were). In general, there should not be any differences in feature support between the TableScan operator and the IndexScan operator. Maybe, it would be good enough for now to fall back to the TableScan in the case of reference segments, similar to the IndexJoin falling back to the NestedLoopJoin in certain situations.

2. Merge both optimizer rules, so that the predicate reordering is aware of the possibility that the first predicate might use an index. I don't really like this approach, though, as it mingles two independent optimization steps.

Bouncner: Let's discuss this on Wednesday.

Bouncner:
I have a general question: according to one TPC-DS paper, there will be large IN lists. Such IN-Lists should be reformulated to joins. But if we (join reformulation seminar group) are not mistaken, such IN-Lists are currently not tables.

Does it make sense to see these IN lists as tables in general and then apply your rewrite in case the table's length is 1?

mrks:
Just to add an option to the discussion: Could we also do this in a similar fashion as `TableScan::_resolve_uncorrelated_subqueries`? The advantage would be that we could also deal with `WHERE a IN (SELECT ... LIMIT 1)`.

mrzzzrm:
> @Bouncner Does it make sense to see these IN lists as tables in general 

Maybe, not really related to this PR, let's discuss elsewhere/in person

> @Bouncner And is there any performance impact on TPC-H

This was motivated by a  JOB query, running it as we speak

> @mrks Could we also do this in a similar fashion as [...]

Sounds attractive, but I'd do the expression rewrite as proposed in this PR anyway, because other optimization rules (predicate ordering) will be able to work with `=` but not with `IN`

mrzzzrm:
@Bouncner with `-r 1` 15% improvement for JOB. Running with `-r 20` now.

mrzzzrm:
Around 13% improvement with JOB, but that is due to two queries, all others remain unchanged in perf. Can this go in? (@mrks)

mrks:
In my review queue.

src/lib/optimizer/strategy/expression_reduction_rule.cpp
```diff
@@ -131,4 +123,49 @@ std::shared_ptr<AbstractExpression> LogicalReductionRule::reduce_distributivity(
   }
 }
 
+std::shared_ptr<AbstractExpression> ExpressionReductionRule::reduce_in_with_single_list_element(
+    const std::shared_ptr<AbstractExpression>& input_expression) {
+  if (const auto in_expression = std::dynamic_pointer_cast<InExpression>(input_expression)) {
+    if (const auto list_expression = std::dynamic_pointer_cast<ListExpression>(in_expression->set())) {
```
mrks: `const auto&`?

src/lib/optimizer/strategy/expression_reduction_rule.cpp
```diff
@@ -14,48 +16,38 @@ namespace opossum {
...
+      expression = reduce_in_with_single_list_element(expression);
+
+      // We can't prune Aggregate arguments, because the operator doesn't support, e.g., `MIN(1)`, whereas it supports
+      // `MIN(2-1)`, since `2-1` becomes a column.
+      if (sub_node->type != LQPNodeType::Aggregate) {
+        // TODO(anybody) We can't prune top level expressions right now, because that breaks `SELECT MIN(1+2)...`
+        //               because if we rewrite that to `SELECT MIN(3)...` the input to the aggregate is not a column
+        //               anymore and the Aggregate operator cannot handle that
+        for (auto& argument : expression->arguments) {
+          reduce_constant_expression(argument);
```
mrks: Why do we modify in-place here but use return values in `reduce_in_with_single_list_element`?

src/lib/optimizer/strategy/expression_reduction_rule.cpp
```diff
@@ -131,4 +123,49 @@ std::shared_ptr<AbstractExpression> LogicalReductionRule::reduce_distributivity(
...
+
+  const auto all_arguments_are_values =
+      std::all_of(input_expression->arguments.begin(), input_expression->arguments.end(),
+                  [&](const auto& argument) { return argument->type == ExpressionType::Value; });
+
+  if (!all_arguments_are_values) return;
+
+  resolve_data_type(input_expression->data_type(), [&](const auto data_type_t) {
+    using ExpressionDataType = typename decltype(data_type_t)::type;
+    const auto result = ExpressionEvaluator{}.evaluate_expression_to_result<ExpressionDataType>(*input_expression);
```
mrks: I really like how easy these things became. :)

src/lib/optimizer/strategy/predicate_split_up_rule.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include "abstract_rule.hpp"
+
+namespace opossum {
+
+/**
+ * This rule turns PredicateNodes with conjunctive chains (e.g. `PredicateNode(a AND b AND c)`) as their scan expression
+ *    into multiple consecutive PredicateNodes (e.g. `PredicateNode(c) -> PredicateNode(b) -> PredicateNode(a)`).
+ *
+ * Doing so enables other Optimizer rules to process these PredicateNodes and split-up PredicateNodes might take a
+ *    faster operator exectution path.
```
mrks: typo

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Step 1:
+   *    - Collect PredicateNodes that can be split up into multiple ones into `predicate_nodes_to_flat_conjunctions`
+   */
+  auto predicate_nodes_to_flat_conjunctions =
+      std::vector<std::pair<std::shared_ptr<PredicateNode>, std::vector<std::shared_ptr<AbstractExpression>>>>{};
+
+  visit_lqp(root, [&](const auto& sub_node) {
+    // We only aim at PredicateNodes, since these are the nodes that primarily contain logical expressions.
```
mrks: Why "primarily"? What other cases are there?

src/lib/optimizer/strategy/expression_reduction_rule.cpp
mrzzzrm: We don't do `const auto& a = std::dynamic_pointer_cast` anywhere else. Whatever affection one can find for this reference lifetime extension thingy, I haven't stumbled upon it yet.

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
mrzzzrm: c/p nonesense, removed

src/lib/optimizer/strategy/expression_reduction_rule.cpp
mrks: I just had to check what `dynamic_pointer_cast` returns. Since it returns a new shared_ptr that shares ownership with the original pointer, it doesn't make a difference. If, however, it returned some type of reference, `const auto&` would be the better choice. And that is the beauty of lifetime extension - you can't do anything wrong here. No need to change it, though.

src/lib/optimizer/strategy/expression_reduction_rule.cpp
mrzzzrm: okay changed to in-place rewrite and also recursive application.

Tried this before but failed. probably just been tired.

PR #1459, opened 2019-02-02
Reduce hyriseTest runtime

mrzzzrm:
Decreased some magic numbers and got the test runtime (debug) from >5s down to 4.5s on my machine. For none of these magic numbers was a justification given, so I suppose they are fair game :man_shrugging:


mrzzzrm:
`OperatorsTableScanTest, ScanOnWideDictionarySegment` is another 0.5s but refactoring that is actual work.

PR #1461, opened 2019-02-03
Proper libHyrise linking for hyriseTest & pausable_loop_thread fix

julianmenzler:
Our DYOD18/19 group works on the MvccDeletePlugin. 
We experienced linking issues with the hyrisePlayground target, which now have been fixed (#1438).
Since we use our plugin in the hyriseTest target as well, we need the fix to be applied here as well.

Our plugin uses `utils/pausible_loop_thread.hpp`. The missing `#pragma once` led to compile issues on our front.

### DIFF
--- a/src/test/CMakeLists.txt
+++ b/src/test/CMakeLists.txt
@@ -1,3 +1,10 @@
+# -force_load/--whole-archive are necessary to include all symbols so that dynamically loaded plugins can use them
+if (APPLE)
+    set(HYRISE_LIBRARY_WITH_ALL_SYMBOLS -force_load hyrise)
+else()
+    set(HYRISE_LIBRARY_WITH_ALL_SYMBOLS -Wl,--whole-archive hyrise -Wl,--no-whole-archive)
+endif()
+
 set(
     SHARED_SOURCES
     base_test.hpp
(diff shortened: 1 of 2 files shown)


src/test/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
+# -force_load/--whole-archive are necessary to include all symbols so that dynamically loaded plugins can use them
```
mrzzzrm: @Bensk1 Conditionally defining HYRISE_LIBRARY_WITH_ALL_SYMBOLS in each CMakeLists.txt (the same cmake code as added in this PR appears in another CMakeLists.txt as well) in which libhyrise is linked to something is not a sustainable solution. Is there a way to modify the `hyrise` target in a way that linking to it automatically incurs these extra flags?

Bensk1: I would not have any idea how the modification of the `hyrise` target should work. Could define `HYRISE_LIBRARY_WITH_ALL_SYMBOLS` in the topmost CMakeLists.txt and have it available in the other ones?

Bensk1: I will have a look at it tomorrow.

PR #1462, opened 2019-02-03
[DYOD18/19] MvccDeletePlugin prep. (TransactionManager & -Context)

julianmenzler:
Our upcoming MvccDeletePlugin (DYOD18/19 project) requires some changes within hyrise.

- `TransactionManager` needs to keep track of running transactions resp. the snapshot-commit-ids being in use. 
This allows our MvccDeletePlugin to make sure that an already logically deleted chunk (fully invalidated) can be deleted physically (from memory) as well – without interfering with ongoing transactions.
- `TransactionContext` notifies `TransactionManager` that the transaction has finished. So the `TransactionManager` can "release" the snapshot-commit-id

### DIFF
--- a/src/lib/concurrency/transaction_context.cpp
+++ b/src/lib/concurrency/transaction_context.cpp
@@ -15,7 +15,7 @@ TransactionContext::TransactionContext(const TransactionID transaction_id, const
       _snapshot_commit_id{snapshot_commit_id},
       _phase{TransactionPhase::Active},
       _num_active_operators{0} {
-  TransactionManager::get().register_transaction(snapshot_commit_id);
+  TransactionManager::get()._register_transaction(snapshot_commit_id);
 }
 
 TransactionContext::~TransactionContext() {
@@ -46,7 +46,7 @@ TransactionContext::~TransactionContext() {
    * Tell the TransactionManager, which keeps track of active snapshot-commit-ids,
    * that this transaction has finished.
    */
-  TransactionManager::get().deregister_transaction(_snapshot_commit_id);
+  TransactionManager::get()._deregister_transaction(_snapshot_commit_id);
 }
 
 TransactionID TransactionContext::transaction_id() const { return _transaction_id; }
--- a/src/lib/concurrency/transaction_context.cpp
+++ b/src/lib/concurrency/transaction_context.cpp
@@ -10,11 +10,9 @@
 
 namespace opossum {
 
-TransactionContext::TransactionContext(const TransactionID transaction_id, const CommitID snapshot_commit_id,
-                                       const bool created_by_transaction_manager)
+TransactionContext::TransactionContext(const TransactionID transaction_id, const CommitID snapshot_commit_id)
     : _transaction_id{transaction_id},
       _snapshot_commit_id{snapshot_commit_id},
-      _created_by_transaction_manager{created_by_transaction_manager},
       _phase{TransactionPhase::Active},
       _num_active_operators{0} {}
 
@@ -42,13 +40,11 @@ TransactionContext::~TransactionContext() {
               }()),
               "Has registered operators but has neither been committed nor rolled back.");
 
-  if (_created_by_transaction_manager) {
-    /**
+  /**
      * Tell the TransactionManager, which keeps track of active snapshot-commit-ids,
      * that this transaction has finished.
      */
-    TransactionManager::get().remove_active_snapshot_commit_id(_snapshot_commit_id);
-  }
+  TransactionManager::get().deregister_transaction(_snapshot_commit_id);
 }
 
 TransactionID TransactionContext::transaction_id() const { return _transaction_id; }
--- a/src/lib/concurrency/transaction_context.cpp
+++ b/src/lib/concurrency/transaction_context.cpp
@@ -10,9 +10,11 @@
 
 namespace opossum {
 
-TransactionContext::TransactionContext(const TransactionID transaction_id, const CommitID snapshot_commit_id)
+TransactionContext::TransactionContext(const TransactionID transaction_id, const CommitID snapshot_commit_id,
+                                       const bool created_by_transaction_manager)
     : _transaction_id{transaction_id},
       _snapshot_commit_id{snapshot_commit_id},
+      _created_by_transaction_manager{created_by_transaction_manager},
       _phase{TransactionPhase::Active},
       _num_active_operators{0} {}
 
@@ -40,11 +42,13 @@ TransactionContext::~TransactionContext() {
               }()),
               "Has registered operators but has neither been committed nor rolled back.");
 
-  /**
-   * Tell the TransactionManager, which keeps track of active snapshot-commit-ids,
-   * that this transaction has finished.
-   */
-  TransactionManager::get().remove_active_snapshot_commit_id(_snapshot_commit_id);
+  if (_created_by_transaction_manager) {
+    /**
+     * Tell the TransactionManager, which keeps track of active snapshot-commit-ids,
+     * that this transaction has finished.
+     */
+    TransactionManager::get().remove_active_snapshot_commit_id(_snapshot_commit_id);
+  }
 }
 
 TransactionID TransactionContext::transaction_id() const { return _transaction_id; }
--- a/src/lib/concurrency/transaction_context.cpp
+++ b/src/lib/concurrency/transaction_context.cpp
@@ -14,7 +14,9 @@ TransactionContext::TransactionContext(const TransactionID transaction_id, const
     : _transaction_id{transaction_id},
       _snapshot_commit_id{snapshot_commit_id},
       _phase{TransactionPhase::Active},
-      _num_active_operators{0} {}
+      _num_active_operators{0} {
+  TransactionManager::get().register_transaction(snapshot_commit_id);
+}
 
 TransactionContext::~TransactionContext() {
   DebugAssert(([this]() {
@@ -41,9 +43,9 @@ TransactionContext::~TransactionContext() {
               "Has registered operators but has neither been committed nor rolled back.");
 
   /**
-     * Tell the TransactionManager, which keeps track of active snapshot-commit-ids,
-     * that this transaction has finished.
-     */
+   * Tell the TransactionManager, which keeps track of active snapshot-commit-ids,
+   * that this transaction has finished.
+   */
   TransactionManager::get().deregister_transaction(_snapshot_commit_id);
 }
 
--- a/src/lib/concurrency/transaction_context.cpp
+++ b/src/lib/concurrency/transaction_context.cpp
@@ -39,6 +39,12 @@ TransactionContext::~TransactionContext() {
                 return !has_registered_operators || committed_or_rolled_back;
               }()),
               "Has registered operators but has neither been committed nor rolled back.");
+
+  /**
+   * Tell the TransactionManager, which keeps track of active snapshot-commit-ids,
+   * that this transaction has finished.
+   */
+  TransactionManager::get().remove_active_snapshot_commit_id(_snapshot_commit_id);
 }
 
 TransactionID TransactionContext::transaction_id() const { return _transaction_id; }
@@ -67,7 +73,6 @@ bool TransactionContext::rollback() {
   }
 
   _mark_as_rolled_back();
-
   return true;
 }
 
--- a/src/lib/concurrency/transaction_manager.cpp
+++ b/src/lib/concurrency/transaction_manager.cpp
@@ -24,31 +24,36 @@ CommitID TransactionManager::last_commit_id() const { return _last_commit_id; }
 
 std::shared_ptr<TransactionContext> TransactionManager::new_transaction_context() {
   std::unique_lock<std::mutex> lock(_mutex_active_snapshot_commit_ids);
-  TransactionID snapshot_commit_id = _last_commit_id;
+  const TransactionID snapshot_commit_id = _last_commit_id;
   _active_snapshot_commit_ids.insert(snapshot_commit_id);
   return std::make_shared<TransactionContext>(_next_transaction_id++, snapshot_commit_id, true);
 }
 
 void TransactionManager::remove_active_snapshot_commit_id(CommitID snapshot_commit_id) {
   std::unique_lock<std::mutex> lock(_mutex_active_snapshot_commit_ids);
-  for (auto it = _active_snapshot_commit_ids.begin(); it != _active_snapshot_commit_ids.end(); ++it) {
-    if (*it == snapshot_commit_id) {
-      it = _active_snapshot_commit_ids.erase(it);
-      return;
-    }
+
+  auto it =
+      std::find(std::begin(_active_snapshot_commit_ids), std::end(_active_snapshot_commit_ids), snapshot_commit_id);
+
+  if (it != std::end(_active_snapshot_commit_ids)) {
+    _active_snapshot_commit_ids.erase(it);
+    return;
   }
   DebugAssert(false,
               "Could not find snapshot_commit_id in TransactionManager's "
               "_active_snapshot_commit_ids. Therefore, the removal failed and "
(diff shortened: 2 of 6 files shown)


julianmenzler:
Please add @m0r13 and @Philipp-Bode to the hyrise organization so I can add them as reviewers.

mrks:
done

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -1,8 +1,7 @@
 #include "transaction_manager.hpp"
 
-#include <memory>
```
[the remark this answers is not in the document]
julianmenzler: In my eyes it is redundant since it is already included in `transaction_manager.hpp`.

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -23,7 +23,32 @@ TransactionManager::TransactionManager()
...
+
+void TransactionManager::remove_active_snapshot_commit_id(CommitID snapshot_commit_id) {
+  std::unique_lock<std::mutex> lock(_mutex_active_snapshot_commit_ids);
+  for (auto it = _active_snapshot_commit_ids.begin(); it != _active_snapshot_commit_ids.end(); ++it) {
+    if (*it == snapshot_commit_id) {
+      it = _active_snapshot_commit_ids.erase(it);
+      return;
+    }
+  }
+  DebugAssert(false,
```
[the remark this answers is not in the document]
julianmenzler: This is good hint. Admittedly, it already felt a little "dirty" when I was writing this 😃

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -23,7 +23,32 @@ TransactionManager::TransactionManager()
...
+              "Could not find snapshot_commit_id in TransactionManager's "
+              "_active_snapshot_commit_ids. Therefore, the removal failed and "
+              "the function should not have been called.")
+}
+
+CommitID TransactionManager::get_lowest_active_snapshot_commit_id() const {
+  std::unique_lock<std::mutex> lock(_mutex_active_snapshot_commit_ids);
+  CommitID lowest_id = MvccData::MAX_COMMIT_ID;
+  for (auto it = _active_snapshot_commit_ids.begin(); it != _active_snapshot_commit_ids.end(); it++)
```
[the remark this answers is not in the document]
julianmenzler: Not really, thank you. I'll look into this. In this context `std::min_element` looks more promising though.

src/lib/concurrency/transaction_manager.cpp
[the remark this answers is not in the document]
julianmenzler: Just looked up the `Assert` macro:
```cpp
#define Assert(expr, msg)                                                                                 \
  if (!static_cast<bool>(expr)) {                                                                         \
    opossum::Fail(opossum::trim_source_file_path(__FILE__) + ":" BOOST_PP_STRINGIZE(__LINE__) " " + msg); \
  }
```
When calling `Fail("...")` directly we're missing `__FILE__`and `__LINE__` in the output.
In addition, the check will also happen in the release build, which one can argue about.

src/lib/concurrency/transaction_context.cpp
```diff
@@ -10,9 +10,11 @@
 
 namespace opossum {
 
-TransactionContext::TransactionContext(const TransactionID transaction_id, const CommitID snapshot_commit_id)
+TransactionContext::TransactionContext(const TransactionID transaction_id, const CommitID snapshot_commit_id,
+                                       const bool created_by_transaction_manager)
     : _transaction_id{transaction_id},
       _snapshot_commit_id{snapshot_commit_id},
+      _created_by_transaction_manager{created_by_transaction_manager},
```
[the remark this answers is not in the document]
julianmenzler: There are several tests, which create TransactionContexts on their own. E.g. `src/test/operators/get_table_test.cpp`.
We introduced this flag for safety reasons. Shall we delete it?

src/lib/concurrency/transaction_manager.cpp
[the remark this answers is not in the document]
julianmenzler: 👍

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -13,6 +12,7 @@ void TransactionManager::reset() {
   manager._next_transaction_id = INITIAL_TRANSACTION_ID;
   manager._last_commit_id = INITIAL_COMMIT_ID;
   manager._last_commit_context = std::make_shared<CommitContext>(INITIAL_COMMIT_ID);
+  manager._active_snapshot_commit_ids.clear();
```
mrks: Just saw this. Does this actually make sense? Shouldn't we assert that this already is empty?

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -23,7 +23,41 @@ TransactionManager::TransactionManager()
...
+  auto it =
+      std::find(std::begin(_active_snapshot_commit_ids), std::end(_active_snapshot_commit_ids), snapshot_commit_id);
+
+  if (it != std::end(_active_snapshot_commit_ids)) {
+    _active_snapshot_commit_ids.erase(it);
+    return;
+  }
+
+  /**
+   * TODO(all) Change Google tests like get_table_test.cpp to always use the TransactionManager for creating
```
julianmenzler: Update: I tried to change the tests → https://github.com/d-justen/hyrise/issues/33#issuecomment-463806010

What do you suggest @mrks?

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -23,7 +23,37 @@ TransactionManager::TransactionManager()
 CommitID TransactionManager::last_commit_id() const { return _last_commit_id; }
 
 std::shared_ptr<TransactionContext> TransactionManager::new_transaction_context() {
-  return std::make_shared<TransactionContext>(_next_transaction_id++, _last_commit_id);
+  std::unique_lock<std::mutex> lock(_mutex_active_snapshot_commit_ids);
```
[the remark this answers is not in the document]
julianmenzler: Intel writes following [in the docs](https://software.intel.com/en-us/node/506181):
> concurrent_unordered_set and concurrent_unordered_multiset support concurrent insertion and traversal, but not concurrent erasure.

Since we are also erasing elements from the multiset, we cannot get rid of the locks, it seems.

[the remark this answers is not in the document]
mrks: Got it.

src/lib/concurrency/transaction_manager.cpp
mrks: Would it be possible to create *and register* these by hand?

julianmenzler: Yep, possible. I already started implementing this. Then, a thought came to my mind: Instead of calling `register_transaction()` manually at different places, we could also call it in the `TransactionContext`-constructor once.
In addition, we could make `TransactionContext` a friend class of `TransactionManager`. Then, `register_transaction()` and `deregister_transaction()` could be made private and thus removed from the public interface of `TransactionManager`.

src/lib/concurrency/transaction_manager.cpp
julianmenzler: True. This code was written before calling `deregister_transaction` from the `TransactionContext` destructor.
So I'll replace it as follows:
```suggestion
  DebugAssert(manager._active_snapshot_commit_ids.empty(),
          "Some transactions do not seem to have finished yet, as they are still registered as active.")
```

src/lib/concurrency/transaction_manager.cpp
mrks: Sounds even better.

src/lib/concurrency/transaction_manager.hpp
```diff
@@ -80,5 +95,8 @@ class TransactionManager : public Singleton<TransactionManager> {
   static constexpr auto INITIAL_COMMIT_ID = CommitID{1};
 
   std::shared_ptr<CommitContext> _last_commit_context;
+
+  mutable std::mutex _mutex_active_snapshot_commit_ids;
+  std::unordered_multiset<CommitID> _active_snapshot_commit_ids;
```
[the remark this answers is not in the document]
julianmenzler: A test for this has been added.

julianmenzler:
All changes are applied. So this pull request should be ready for merging.

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -40,13 +44,9 @@ void TransactionManager::deregister_transaction(CommitID snapshot_commit_id) {
...
   }
 
-  /**
-   * TODO(all) Change Google tests like get_table_test.cpp to always use the TransactionManager for creating
-   * TransactionContexts. After dong that, enable the following check:
-   */
-  //  Fail(opossum::trim_source_file_path(__FILE__) + ":" BOOST_PP_STRINGIZE(__LINE__) +
-  //       " Could not find snapshot_commit_id in TransactionManager's _active_snapshot_commit_ids. Therefore," +
-  //       " the removal failed and the function should not have been called.");
+  Fail(opossum::trim_source_file_path(__FILE__) + ":" BOOST_PP_STRINGIZE(__LINE__) +
```
mrks: That was not what I meant. Have a look at assert.hpp. There, `Assert` builds the file/line/msg string and passes it to `Fail`. If we move that part to `Fail`, we would have line information for both asserts and fails. Only `assert.hpp` needs to change.

mrks:
@nilsthamm and team: You are blocking this PR. Is everything you were concerned about addressed?

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -13,6 +12,8 @@ void TransactionManager::reset() {
   manager._next_transaction_id = INITIAL_TRANSACTION_ID;
   manager._last_commit_id = INITIAL_COMMIT_ID;
   manager._last_commit_context = std::make_shared<CommitContext>(INITIAL_COMMIT_ID);
+  DebugAssert(manager._active_snapshot_commit_ids.empty(),
```
mrzzzrm: As this is a fast, seldomly executed check that could potentially point to a grave problem, I'd make this an `Assert` so release builds profit from it as well.

src/lib/concurrency/transaction_manager.cpp
mrzzzrm: In any case, you should probably just do `Assert(it != _active_snapshot_commit_ids.end(), "...")`;

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -23,7 +24,40 @@ TransactionManager::TransactionManager()
...
+void TransactionManager::register_transaction(const CommitID snapshot_commit_id) {
+  std::unique_lock<std::mutex> lock(_mutex_active_snapshot_commit_ids);
+  _active_snapshot_commit_ids.insert(snapshot_commit_id);
+}
+
+void TransactionManager::deregister_transaction(const CommitID snapshot_commit_id) {
+  std::unique_lock<std::mutex> lock(_mutex_active_snapshot_commit_ids);
+
+  auto it =
+      std::find(std::begin(_active_snapshot_commit_ids), std::end(_active_snapshot_commit_ids), snapshot_commit_id);
```
mrzzzrm: I know it is more generic the was you did it, but the rest of the code base would do `_active_snapshot_commit_ids.begin()` over `std::begin(...)`, so I'd opt for the former for consistency reasons. Same in the `if` below.

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -23,7 +24,40 @@ TransactionManager::TransactionManager()
...
+  Fail(opossum::trim_source_file_path(__FILE__) + ":" BOOST_PP_STRINGIZE(__LINE__) +
+       " Could not find snapshot_commit_id in TransactionManager's _active_snapshot_commit_ids. Therefore," +
+       " the removal failed and the function should not have been called.");
+}
+
+CommitID TransactionManager::get_lowest_active_snapshot_commit_id() const {
+  std::unique_lock<std::mutex> lock(_mutex_active_snapshot_commit_ids);
+
+  if (_active_snapshot_commit_ids.empty()) {
+    return MvccData::MAX_COMMIT_ID;
```
mrzzzrm: Maybe `std::nullopt` (and `std::optional<CommitID>` as the function return type), instead of a somewhat magic number?

src/test/concurrency/transaction_manager_test.cpp
```diff
@@ -0,0 +1,88 @@
...
+  EXPECT_EQ(manager.get_lowest_active_snapshot_commit_id(), t2_context->snapshot_commit_id());
+
+  t2_context->commit();
+  deregister_transaction(t2_context->snapshot_commit_id());
+
+  EXPECT_EQ(get_active_snapshot_commit_ids().size(), 0);
+  EXPECT_EQ(manager.get_lowest_active_snapshot_commit_id(), MvccData::MAX_COMMIT_ID);
+
+  // To prevent exceptions in TransactionContext destructor
+  register_transaction(t1_snapshot_commit_id);
```
mrzzzrm: Shouldn't it actually be `TransactionContext::commit` and `TransactionContext::rollback` that end the lifecycle of a transaction and de-register the transaction from the TransactionManager? This would reduce the lifetime of the Transaction, which is imo a good thing, and render void the necessity to have these hacky `register_transaction()` calls here.

src/lib/concurrency/transaction_manager.cpp
mrks: Good point, especially since we only call this in test code anyway

src/test/concurrency/transaction_manager_test.cpp
julianmenzler: We've had a discussion regarding this a few weeks back: To keep `TransactionManager`'s `_active_snapshot_commit_ids` consistent, we need to make sure that `deregister_transaction` is called only once per `TransactionContext`. Since `commit` and `abort` can be called multiple times we would need a bool flag  and additional checks to ensure `exactly-once` semantics. 
When using the destructor of `TransactionContext` to call `deregister_transaction`, we do not need any checks. In addition, the code stays clean and compact. 
The delayed call of `deregister_transaction`is feasible.

Remember, the hacky use of `register_transaction` here is required for this test only since it breaks with conventions. `TransactionManager` is usually the only class calling `register_transaction`.

mrks: > Since `commit` and `abort` can be called multiple times

This is news to me. Also, this seems to suggest otherwise: 
https://github.com/hyrise/hyrise/blob/master/src/lib/concurrency/transaction_context.hpp#L133

src/lib/concurrency/transaction_manager.cpp
julianmenzler: > That was not what I meant. Have a look at assert.hpp. There, `Assert` builds the file/line/msg string and passes it to `Fail`. If we move that part to `Fail`, we would have line information for both asserts and fails. Only `assert.hpp` needs to change.

Ok, I did not succeed in changing this right away. I opened a ticket and will take care of it in a week or so. 

For now, I'll change this as @mrzzzrm suggested.

src/test/concurrency/transaction_manager_test.cpp
julianmenzler: Thanks for the hint @mrks, you are right! In consequence, there is one argument left for the current implementation: Minimal use of code – one line of code in the destructor vs. multiple spread across `TransactionContext`.

So, should we switch again, to the approach @mrzzzrm suggested?

mrks: I think both have their advantages and disadvantages. Once we work more with modifying transactions, we will see if there is a material difference between the two. We might also see questions like "What happens to contexts of lost network connections" and alike.

For now, I'd leave it the way it is. In order to slowly reduce the number of open PRs, I wouldn't see this as a blocker.

PR #1466, opened 2019-02-04
Fix false call in jemalloc build

janehmueller:
references #1448 ([comment](https://github.com/hyrise/hyrise/issues/1448#issuecomment-460221250))

On Ubuntu (e.g., the build vms for the DYOD seminar) the following does not work:
`BUILD_COMMAND bash -c "output=$(make 2>&1) || (printf \"$output\\n\" && /usr/bin/false)"`

The false binary is located in `/usr/bin/false` on macOS while it is located in `/bin/false` on linux (or at least ubuntu).

This PR just uses `false` instead.


janehmueller:
The build fails due to a missing authorisation token. No clue what to do about that.

mrks:
I'll have to add you to the organization

PR #1467, opened 2019-02-04
[DYOD18/19] Add LZ4 Encoding Scheme

janehmueller:
This PR adds the heavyweight compression library LZ4 to the existing hyrise Encoding Scheme. This happens in 3 parts:

- The `LZ4Encoder` encodes ValueSegments into LZ4Segments.
- The `LZ4Iterable` implements the Encoding Iterator scheme, which provides a normal iterator as well as a point-access iterator. The point-access iterator currently still decompresses the whole segment since point-access is not a feature of LZ4. We will extend it in the next few weeks with a proper implementation.
- The `LZ4Segment` which stores the compressed data, the necessary metadata for it (compressed and uncompressed size) and provides the decompression needed to access the values again.


janehmueller:
@Bouncner could you please add @Felerius, @janetzki and Pascal (don't know his GitHub name - maybe David and Jonathan can tell us) as reviewers.

src/lib/storage/create_iterable_from_segment.hpp
```diff
@@ -72,6 +73,15 @@ auto create_iterable_from_segment(const FrameOfReferenceSegment<T>& segment) {
   }
 }
 
+template <typename T, bool EraseSegmentType = HYRISE_DEBUG>
+auto create_iterable_from_segment(const LZ4Segment<T>& segment) {
+  if constexpr (EraseSegmentType) {
+    return create_any_segment_iterable<T>(segment);
+  } else {
+    return LZ4Iterable<T>{segment};
```
mrzzzrm: Imo, as LZ4 is slow and inlining will both be costly (in terms of build time) and of little benefit (in terms of runtime performance), its type should always be erased. i.e. just `return create_any_segment_iterable<T>(segment);` from this function

src/lib/storage/lz4_segment.hpp
```diff
@@ -0,0 +1,71 @@
...
+   * @{
+   */
+
+  EncodingType encoding_type() const final;
+  std::optional<CompressedVectorType> compressed_vector_type() const final;
+
+  /**@}*/
+
+ private:
+  const std::shared_ptr<const pmr_vector<char>> _compressed_data;
```
mrzzzrm: Why not just `pmr_vector<char> _compressed_data`. Which other data structure is this vector shared with?

src/lib/storage/lz4_segment.hpp
```diff
@@ -0,0 +1,71 @@
...
+namespace opossum {
+
+class BaseCompressedVector;
+
+template <typename T>
+class LZ4Segment : public BaseEncodedSegment {
+ public:
+  explicit LZ4Segment(const std::shared_ptr<const pmr_vector<char>>& compressed_data,
+                      const std::shared_ptr<const pmr_vector<bool>>& null_values,
+                      const std::shared_ptr<const pmr_vector<size_t>>& offsets, const int compressed_size,
```
mrzzzrm: Can you doc these params, e.g., `offsets`, pls? What is `compressed_size`? Is it equal to `compressed_data.size()`

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,125 @@
...
+                                               input_size, num_elements);
+  }
+
+  std::shared_ptr<BaseEncodedSegment> _on_encode(
+      const std::shared_ptr<const ValueSegment<std::string>>& value_segment) {
+    const auto alloc = value_segment->values().get_allocator();
+    const auto num_elements = value_segment->size();
+
+    // copy values and null flags from value segment
+    auto values = pmr_vector<char>{alloc};
```
Felerius: Have you tested reserving for values as well (By summing over the lengths of all strings in the segment in the beginning)?

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,125 @@
...
+    offsets.reserve(num_elements);
+
+    auto iterable = ValueSegmentIterable<std::string>{*value_segment};
+    iterable.with_iterators([&](auto it, auto end) {
+      size_t offset = 0u;
+      bool is_null;
+      for (; it != end; ++it) {
+        auto segment_value = *it;
+        is_null = segment_value.is_null();
+        null_values.push_back(is_null);
```
Felerius: See question in the method above

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,125 @@
...
+        offsets.emplace_back(offset);
+        if (!is_null) {
+          auto c_string = segment_value.value().c_str();
+          auto length = strlen(c_string);
+          values.insert(values.cend(), c_string, c_string + length);
```
Felerius: What is the advantage of using C style strings over using `value.insert(values.cend(), str.begin(), str.end())` and `offset += str.size()`? Also, if null bytes are allowed in the input strings, `strlen` might cut off the string early.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,125 @@
...
+          offset += length;
+        }
+      }
+    });
+
+    // lz4 compression
+    const auto input_size = static_cast<int>(values.size());
```
Felerius: This should probably be protected by a DebugAssert (as in the method above)

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,125 @@
...
+    auto output_size = LZ4_compressBound(input_size);
+    auto compressed_data = pmr_vector<char>{alloc};
+    compressed_data.reserve(static_cast<size_t>(output_size));
+    const int compression_result =
+        LZ4_compress_HC(values.data(), compressed_data.data(), input_size, output_size, LZ4HC_CLEVEL_MAX);
+    if (compression_result <= 0) {
+      Fail("LZ4 compression failed");
+    }
+
+    auto data_ptr = std::allocate_shared<pmr_vector<char>>(alloc, std::move(compressed_data));
```
Felerius: See comment above about using `reserve` and how it affects `data_ptr->size()`

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,125 @@
...
+    auto null_values_ptr = std::allocate_shared<pmr_vector<bool>>(alloc, std::move(null_values));
+    auto offset_ptr = std::allocate_shared<pmr_vector<size_t>>(alloc, std::move(offsets));
+
+    return std::allocate_shared<LZ4Segment<std::string>>(alloc, data_ptr, null_values_ptr, offset_ptr,
+                                                         compression_result, input_size, num_elements);
```
Felerius: Consider factoring out the actual compression into a helper method

src/lib/storage/lz4_segment.cpp
```diff
@@ -0,0 +1,152 @@
...
+  if (decompressed_result <= 0) {
+    Fail("LZ4 decompression failed");
+  }
+
+  auto string_data = std::make_shared<std::vector<std::string>>();
+  for (auto it = _offsets->cbegin(); it != _offsets->cend(); ++it) {
+    auto begin = *it;
+    size_t end;
+    if (it + 1 == _offsets->cend()) {
+      end = static_cast<size_t>(_decompressed_size);
```
Felerius: Consider making `_decompressed_size` and `_compressed_size` size_t's, and only casting to ints when calling lz4 functions. Or are negative values for these values actually possible?

src/lib/storage/lz4_segment.cpp
```diff
@@ -0,0 +1,152 @@
...
+      end = *(it + 1);
+    }
+
+    std::string current_element;
+    const auto data_begin = decompressed_data->cbegin() + begin;
+    const auto data_end = decompressed_data->cbegin() + end;
+    for (auto data_it = data_begin; data_it != data_end; ++data_it) {
+      current_element += (*data_it);
+    }
+    string_data->emplace_back(current_element);
```
Felerius: `std::string` has a constructor taking a begin and end iterator. Consider just using `string_data->emplace_back(data_begin, data_end)`. This would also avoid unnecessary, temporary allocations.

src/lib/storage/lz4_segment.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+template <typename T>
+std::shared_ptr<BaseSegment> LZ4Segment<T>::copy_using_allocator(const PolymorphicAllocator<size_t>& alloc) const {
+  auto new_compressed_data = pmr_vector<char>{*_compressed_data, alloc};
+  auto new_null_values = pmr_vector<bool>{*_null_values, alloc};
+
+  std::shared_ptr<pmr_vector<size_t>> new_offsets_ptr = nullptr;
+  if (_offsets != nullptr) {
+    auto new_offsets = pmr_vector<size_t>{*_offsets, alloc};
+    new_offsets_ptr = std::allocate_shared<pmr_vector<size_t>>(alloc, std::move(new_offsets));
```
Felerius: Consider just using `allocate_shared<...>(alloc, *_offsets)`

src/lib/storage/lz4_segment.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+  return string_data;
+}
+
+template <typename T>
+std::shared_ptr<BaseSegment> LZ4Segment<T>::copy_using_allocator(const PolymorphicAllocator<size_t>& alloc) const {
+  auto new_compressed_data = pmr_vector<char>{*_compressed_data, alloc};
+  auto new_null_values = pmr_vector<bool>{*_null_values, alloc};
+
+  std::shared_ptr<pmr_vector<size_t>> new_offsets_ptr = nullptr;
```
Felerius: The default constructor already initializes the pointer to nullptr

third_party/CMakeLists.txt
```diff
@@ -101,6 +101,30 @@ if(${NUMA_FOUND})
...
+
+    ${LZ4_LIBRARY_DIR}/lz4.h
+    ${LZ4_LIBRARY_DIR}/lz4hc.h
+)
+
+target_compile_definitions(
+    lz4
+
+    PUBLIC
+    LZ4_DISABLE_DEPRECATE_WARNINGS=1)
```
Felerius: Is this a workaround for an lz4 bug or are you actually using deprecated lz4 functions? If so, why?

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,125 @@
...
+    auto iterable = ValueSegmentIterable<T>{*value_segment};
+    iterable.with_iterators([&](auto it, auto end) {
+      for (; it != end; ++it) {
+        auto segment_value = *it;
+        values.emplace_back(segment_value.value());
+        null_values.push_back(segment_value.is_null());
+      }
+    });
+
+    // lz4 compression
```
janetzki: Please consider a consistent uppercase spelling for LZ4.

janetzki:
You could also be the first contributors to https://github.com/hyrise/hyrise/wiki/Compression. ;)

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,125 @@
...
+    auto null_values = pmr_vector<bool>{alloc};
+    null_values.reserve(num_elements);
+
+    // copy values and null flags from value segment
+    auto iterable = ValueSegmentIterable<T>{*value_segment};
+    iterable.with_iterators([&](auto it, auto end) {
+      for (; it != end; ++it) {
+        auto segment_value = *it;
+        values.emplace_back(segment_value.value());
+        null_values.push_back(segment_value.is_null());
```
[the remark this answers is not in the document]
janehmueller: No that's just a remnant of copy and paste 👍

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: No we haven't benchmarked the performance difference but now that you mention it, we will do so 👍

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: That's a leftover of the previous version of encoding strings into a char vector. Thanks for the hint 👍

src/lib/storage/lz4_segment.cpp
janehmueller: Negative values indicate errors and are caught in the asserts. We can change them to size_t's

third_party/CMakeLists.txt
janehmueller: No that's just a copy and paste remnant from when we fixed our lz4 integration (i.e., compiling and linking). We'll remove it.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: Doing that would save the duplication of around 4 lines while we will have to pass lots of values as parameters around. I don't think that overhead is worth saving so few lines of code duplication.

src/lib/storage/lz4_segment.cpp
janehmueller: Our implementation is based on the implementation of the other encoding scheme (e.g., run length segment). There, the data is also copied instead of just creating a new object.

src/lib/storage/create_iterable_from_segment.hpp
janehmueller: Hm, removing the inlining causes a segmentation fault in the LZ4 tests. I don't really know why 😅

src/lib/storage/lz4_segment.hpp
janehmueller: Having it as a pointer makes it easier to move around instead of calling `std::move` all the time (since it's created in the encoder and then moved to the LZ4Segment).

src/lib/storage/lz4_segment.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+namespace opossum {
+
+class BaseCompressedVector;
+
+template <typename T>
+class LZ4Segment : public BaseEncodedSegment {
+ public:
+  /*
+   * An LZ4 segment is a container for an LZ4 compressed segment. It contains the compressed data, the necessary
```
mrks: The tautology fan club is a club for tautology fans. ;)

src/lib/storage/lz4_segment.hpp
```diff
@@ -0,0 +1,85 @@
...
+   * metadata and the ability to decompress the data again.
+   *
+   * @param compressed_data The char vector that contains the LZ4 compressed segment data as binary blob.
+   * @param null_values Boolean vector that contains the information which row is null and which is not null.
+   * @param offsets If this segment is not a std::string segment this will be a null pointer. Otherwise it contains
+   *                the offsets for the compressed strings. The offset at position 0 is the character index of the
+   *                string at index 0. Its (exclusive) end is at the offset at position 1. The last string ends at the
+   *                end of the compressed data (since there is offset after it that specifies the end offset).
```
mrks: Please add a note that strings are not null-terminated.

src/lib/storage/lz4_segment.hpp
```diff
@@ -0,0 +1,85 @@
...
+   * @param compressed_size The size of the compressed data vector (the return value of LZ4)
+   * @param decompressed_size The size in bytes of the decompressed data vector.
+   * @param num_elements The number of rows that are compressed in this segment.
```
mrks: Why do I need this if I have null_values.size?

src/lib/storage/lz4_segment.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+  /**@}*/
+
+ private:
+  const std::shared_ptr<const pmr_vector<char>> _compressed_data;
+  const std::shared_ptr<const pmr_vector<bool>> _null_values;
+  const std::shared_ptr<const pmr_vector<size_t>> _offsets;
+  const int _compressed_size;
+  const int _decompressed_size;
+  size_t _num_elements;
```
mrks: Why isn't this const?

src/lib/storage/lz4_segment.hpp
mrks: this is still open, isn't it?

src/lib/storage/lz4_segment.hpp
mrks: Mmh, I didn't see this and added a similar comment. I'm not really convinced yet - avoiding a couple of moves doesn't seem like a good reason to introduce shared ownership.

src/lib/storage/lz4_segment.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include "utils/performance_warning.hpp"
+
+#include "lib/lz4.h"
+
+namespace opossum {
+
+template <typename T>
+LZ4Segment<T>::LZ4Segment(const std::shared_ptr<const pmr_vector<char>>& compressed_data,
+                          const std::shared_ptr<const pmr_vector<bool>>& null_values,
+                          const std::shared_ptr<const pmr_vector<size_t>>& offsets, const int compressed_size,
```
Bouncner: Why is `compressed_size` an int? I see why the return values in the code have to be signed, but here?

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,126 @@
...
+    const auto alloc = value_segment->values().get_allocator();
+    const auto num_elements = value_segment->size();
+    DebugAssert(num_elements <= std::numeric_limits<int>::max(),
+                "Trying to compress a ValueSegment with more "
+                "elements than fit into an int.");
+
+    // copy values and null flags from value segment
+    auto values = pmr_vector<char>{alloc};
+    auto null_values = pmr_vector<bool>{alloc};
+    null_values.reserve(num_elements);
```
Bouncner: 1.) Please also reserve (or resize according to @Felerius's point) `values`. Something sane as `4*num_elements` should already have an good impact.
2.) Please ensure that this encoding as well as decoding is tested with vectors of empty strings. This is very common in real-world data and might yield problems with your handling.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,126 @@
...
+    });
+
+    // LZ4 compression
+    const auto input_size = static_cast<int>(values.size());
+    auto output_size = LZ4_compressBound(input_size);
+    auto compressed_data = pmr_vector<char>{alloc};
+    compressed_data.resize(static_cast<size_t>(output_size));
+    const int compression_result =
+        LZ4_compress_HC(values.data(), compressed_data.data(), input_size, output_size, LZ4HC_CLEVEL_MAX);
+    Assert(compression_result > 0, "LZ4 compression failed");
```
Bouncner: Please use something as `shrink_to_fit` (or whatever fits when you decided for resize/reserve) to size down the compressed vector to its actual size instead of the estimated upper bound.

src/lib/storage/lz4_segment.cpp
```diff
@@ -0,0 +1,147 @@
...
+    : BaseEncodedSegment{data_type_from_type<T>()},
+      _compressed_data{compressed_data},
+      _null_values{null_values},
+      _offsets{offsets},
+      _compressed_size{compressed_size},
+      _decompressed_size{decompressed_size},
+      _num_elements{num_elements} {}
+
+template <typename T>
+std::shared_ptr<const pmr_vector<char>> LZ4Segment<T>::compressed_data() const {
```
Bouncner: Are these Java'esk accessors necessary?

janehmueller: Probably not. We'll remove them.

src/lib/storage/lz4_segment.hpp
janehmueller: That's true. We'll remove it.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,126 @@
...
+                "elements than fit into an int.");
+
+    // copy values and null flags from value segment
+    auto values = pmr_vector<char>{alloc};
+    auto null_values = pmr_vector<bool>{alloc};
+    null_values.reserve(num_elements);
+
+    /**
+     * These offsets mark the beginning of strings (and therefore end of the previous string) in the data vector.
+     * These offsets are character offsets. The string at position 0 starts at the offset stored at position 0.
```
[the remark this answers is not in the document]
janehmueller: Yes, indeed. I will add that to the comment.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,126 @@
...
+    offsets.reserve(num_elements);
+
+    auto iterable = ValueSegmentIterable<std::string>{*value_segment};
+    iterable.with_iterators([&](auto it, auto end) {
+      size_t offset = 0u;
+      bool is_null;
+      for (; it != end; ++it) {
+        auto segment_value = *it;
+        is_null = segment_value.is_null();
+        null_values.emplace_back(is_null);
```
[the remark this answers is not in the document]
janehmueller: Yes this could be done in one line. But these variables are used in other places of the loop (e.g., the `is_null` variable). I don't think that minimizing the number of lines in this case will improve readability.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -0,0 +1,126 @@
...
+        offsets.emplace_back(offset);
+        if (!is_null) {
```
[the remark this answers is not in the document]
janehmueller: This is handled in the `SegmentPosition` since the `null_value` boolean is a separate variable that is set by/in the source segment.

But we can add a test that tests that (which is similar to Martin's request)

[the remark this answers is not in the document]
mrks: Correct, but in the whole encoding/decoding shebang, this might get lost. If there is a general test (for all encodings) that deals with empty and NULL strings, you're good. Otherwise, please add one.

src/lib/storage/lz4/lz4_encoder.hpp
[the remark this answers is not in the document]
mrks: Missed that. Nevermind.

src/lib/storage/lz4_segment.cpp
```diff
@@ -0,0 +1,147 @@
...
+  auto new_null_values_ptr = std::allocate_shared<pmr_vector<bool>>(alloc, std::move(new_null_values));
+
+  return std::allocate_shared<LZ4Segment>(alloc, new_compressed_data_ptr, new_null_values_ptr, new_offsets_ptr,
+                                          _decompressed_size, _compressed_size, _num_elements);
+}
+
+template <typename T>
+size_t LZ4Segment<T>::estimate_memory_usage() const {
+  auto bool_size = (_null_values ? _null_values->size() * sizeof(bool) : 0u);
+  auto offset_size = (_offsets ? _offsets->size() * sizeof(size_t) : 0u);
```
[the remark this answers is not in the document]
janehmueller: The _null_values vector is never unset that's true. The offset vector is unset if the segment is not a string-segment.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: I don't think the reserve will make much of a difference. Take a look at this [example benchmark](http://quick-bench.com/aplSNPOV_-7cKL6sP_gNhFqBtKg).

src/lib/storage/lz4_segment.hpp
janehmueller: Okay, then we'll change it to just a normal vector.

src/lib/storage/lz4/lz4_encoder.hpp
mrks: Valid in this case, because we expect the compressed vector to live for a long time, but please remember that shrink_to_fit might result in the entire vector being copied again. That might not justify its advantages for ephemeral data structures.

src/lib/storage/lz4_segment.hpp
janehmueller: Yes, compressed_size should equal the size of the vector. We'll remove the compressed_size parameter/member since we can just access the vector size.

src/lib/storage/lz4_segment.cpp
Bouncner: Needed or used?

src/lib/storage/lz4/lz4_encoder.hpp
Bouncner: I think this benchmark is fairer: http://quick-bench.com/CRpktHw2mMVZL94P7v_7n6ixpGA

And I'd argue it's worth the added single line.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: We found that the difference is negligible ([benchmark](http://quick-bench.com/aplSNPOV_-7cKL6sP_gNhFqBtKg)) but we can add it still add it (since there is a small performance improvement).

mrks: resize + []= is even faster than reserving if you know the length in advance and the default constructor is trivial.

src/lib/storage/lz4_segment.hpp
janehmueller: We deleted it (since null_values.size is the same).

src/lib/storage/lz4_segment.cpp
janehmueller: Alright 👍  ([issue link](https://github.com/hyrise/hyrise/issues/1514))

src/lib/storage/lz4_segment.cpp
```diff
@@ -0,0 +1,147 @@
...
+  return (*decompressed_segment)[chunk_offset];
+}
+
+template <typename T>
+size_t LZ4Segment<T>::size() const {
+  return _num_elements;
+}
+
+template <typename T>
+std::shared_ptr<std::vector<T>> LZ4Segment<T>::decompress() const {
```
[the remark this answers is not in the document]
janehmueller: No, this doesn't really need to be a pointer. We'll remove it.

src/lib/storage/lz4_segment.cpp
janehmueller: The null value accessor is needed in the iterators and the offset accessor is needed for testing.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: We do this via `resize` which should just cut off the elements add the end, right?

mrks: Those are two different things. `resize` to a smaller size than current leaves the address and capacity untouched. `shrink_to_fit` frees up the unused capacity, possibly causing a reallocation.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: We summed the total number of characters since we also use that number to handle the edge case of only empty strings.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: Yes, but I don't think replacing multiple characters at once is possible (in a vector of characters). At least I didn't find a way to do that. And iterating also over the single characters in the strings is very slow ([benchmark](http://quick-bench.com/ARtSFQTuCxt8fnsgQWhVoJvZkfA)).

src/lib/storage/lz4/lz4_encoder.hpp
Bouncner: But why not collecting the number during construction?

src/lib/storage/lz4_segment.cpp
Bouncner: Ack.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -28,75 +28,102 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
       const std::shared_ptr<const ValueSegment<std::string>>& value_segment) {
     const auto alloc = value_segment->values().get_allocator();
     const auto num_elements = value_segment->size();
-    DebugAssert(num_elements <= std::numeric_limits<int>::max(),
-                "Trying to compress a ValueSegment with more "
-                "elements than fit into an int.");
+
+    /**
+     * First iterate over the values for two reasons.
+     * 1) If all the strings are empty LZ4 will try to compress an empty vector which will cause a segmentation fault.
```
Bouncner: What about a vector with only empty strings except for one single cell, which contains exactly one char? Does LZ4 have something like a minimum data size which it needs in order to work?

src/test/storage/lz4_segment_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+    // Test offsets
+    EXPECT_TRUE(offsets[index] == expected_offsets[index]);
+  }
+}
+
+TEST_F(StorageLZ4SegmentTest, CompressEmptySegmentString) {
+  vs_str->append("");
+  vs_str->append("");
+  vs_str->append("");
+  vs_str->append("");
```
Bouncner: See question about empty string vector with one 1-char element.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: I added an extra test case for that. A single char works.

src/lib/storage/lz4_segment.cpp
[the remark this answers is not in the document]
mrks: Please add a comment for that.

src/lib/storage/lz4/lz4_encoder.hpp
mrks: ok

src/lib/storage/lz4_segment.cpp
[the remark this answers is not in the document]
mrks: I took the liberty of adding that myself as this was the only open thing I considered missing.

src/lib/storage/lz4/lz4_encoder.hpp
mrks: If you still want to look into this, `std::copy` instead of `v.insert` should do the job.

src/lib/storage/lz4_segment.cpp
```diff
@@ -13,8 +13,8 @@ template <typename T>
 LZ4Segment<T>::LZ4Segment(const pmr_vector<char>& compressed_data, const pmr_vector<bool>& null_values,
                           const std::shared_ptr<const pmr_vector<size_t>>& offsets, const size_t decompressed_size)
     : BaseEncodedSegment{data_type_from_type<T>()},
-      _compressed_data{std::move(compressed_data)},
-      _null_values{std::move(null_values)},
+      _compressed_data{compressed_data},
+      _null_values{null_values},
```
mrks: Maybe change the signature to accept by universal reference; that way you can actually move the data instead of copying it.

src/lib/storage/create_iterable_from_segment.hpp
janehmueller: The error causing line seems to be the [following in `segment_statistics.cpp`](https://github.com/phnm/hyrise/blob/7634ea8b6d42fa179802af136abd7d4ab421ba16/src/lib/statistics/chunk_statistics/segment_statistics.cpp#L58):
`auto iterable = create_iterable_from_segment<DataTypeT>(typed_segment);`

I'm not quite what the alternative here is though.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: `std::copy` seems to be [super slow](http://quick-bench.com/EOqwtNvJk6R8caV4O3fMt6kSjgM) (this is only with 10 elements btw).

src/lib/storage/lz4_segment.cpp
janehmueller: Hm, wouldn't changing the parameters to non-const make them movable? I can do that (and that is what happens in other encoded segments as well).

src/lib/storage/lz4/lz4_encoder.hpp
mrks: Never mind then.

src/lib/storage/create_iterable_from_segment.hpp
mrks: Do you have a stack trace of the segfault?

src/lib/storage/lz4_segment.cpp
mrks: http://blogs.microsoft.co.il/sasha/2014/08/21/c-sink-parameter-passing/

Making it a regular reference would consume it without letting the caller know that you do so. Example (from [here](https://stackoverflow.com/questions/45514116/have-rvalue-reference-instead-of-forwarding-reference-with-variadic-template)):

```
auto foo(X&& arg) -> void {...};

X x;
foo(x);            // compile error [OK]
foo(std::move(x)); // accepted      [OK]
foo(X{});          // accepted      [OK]
```

This is good because you don't inadvertently destroy x.

src/lib/storage/create_iterable_from_segment.hpp
janehmueller: No, unfortunately I don't.

src/lib/storage/lz4_segment.cpp
janehmueller: I can't quite get that working (maybe since the move is in the constructor?). The final solution(s) of that blog post seem kind of overkill and the segment should only ever be created by the encoder (which has no other job and does not use the resulting encoded segment).

Since the other implementations of encoded segments either do the same (move a non-const value) or use shared pointers, I would propose to keep it as it is and maybe create an issue for it (covering all encoded segments).

src/lib/storage/lz4_segment.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include "resolve_type.hpp"
+#include "storage/vector_compression/base_compressed_vector.hpp"
+#include "utils/assert.hpp"
+#include "utils/performance_warning.hpp"
+
+namespace opossum {
+
+template <typename T>
+LZ4Segment<T>::LZ4Segment(pmr_vector<char> compressed_data, pmr_vector<bool> null_values,
+                          const std::shared_ptr<const pmr_vector<size_t>>& offsets, const size_t decompressed_size)
```
mrks: Why is offsets a shared_ptr again? Who is it shared with? If it is just because it is only used for strings, `optional` would be the better fit.

src/lib/storage/create_iterable_from_segment.hpp
mrks: It would have shown you that you have a stack overflow. `create_iterable_from_segment` is called by `CreateAnySegmentIterable<T>::create` with `false` as the second parameter. This should work:

```
template <typename T, bool EraseSegmentType = true>
auto create_iterable_from_segment(const LZ4Segment<T>& segment) {
  // We always erase the type here because LZ4 is too slow to notice a difference anyway.
  if constexpr (EraseSegmentType) {
    return create_any_segment_iterable<T>(segment);
  } else {
    return LZ4Iterable<T>{segment};
  }
}
```

src/lib/storage/lz4_segment.cpp
mrks: Not sure what the problem is:

```
diff --git a/src/lib/storage/lz4_segment.cpp b/src/lib/storage/lz4_segment.cpp
index e42d60926..d444e1454 100644
--- a/src/lib/storage/lz4_segment.cpp
+++ b/src/lib/storage/lz4_segment.cpp
@@ -10,7 +10,7 @@
 namespace opossum {

 template <typename T>
-LZ4Segment<T>::LZ4Segment(pmr_vector<char> compressed_data, pmr_vector<bool> null_values,
+LZ4Segment<T>::LZ4Segment(pmr_vector<char>&& compressed_data, pmr_vector<bool>&& null_values,
                           const std::shared_ptr<const pmr_vector<size_t>>& offsets, const size_t decompressed_size)
     : BaseEncodedSegment{data_type_from_type<T>()},
       _compressed_data{std::move(compressed_data)},
@@ -121,7 +121,7 @@ std::shared_ptr<BaseSegment> LZ4Segment<T>::copy_using_allocator(const Polymorph
     new_offsets_ptr = std::allocate_shared<pmr_vector<size_t>>(alloc, *_offsets);
   }

-  return std::allocate_shared<LZ4Segment>(alloc, new_compressed_data, new_null_values, new_offsets_ptr,
+  return std::allocate_shared<LZ4Segment>(alloc, std::move(new_compressed_data), std::move(new_null_values), new_offsets_ptr,
                                           _decompressed_size);
 }

diff --git a/src/lib/storage/lz4_segment.hpp b/src/lib/storage/lz4_segment.hpp
index 0eb978fc5..a509c6d1d 100644
--- a/src/lib/storage/lz4_segment.hpp
+++ b/src/lib/storage/lz4_segment.hpp
@@ -35,7 +35,7 @@ class LZ4Segment : public BaseEncodedSegment {
    * @param compressed_size The size of the compressed data vector (the return value of LZ4)
    * @param decompressed_size The size in bytes of the decompressed data vector.
    */
-  explicit LZ4Segment(pmr_vector<char> compressed_data, pmr_vector<bool> null_values,
+  explicit LZ4Segment(pmr_vector<char>&& compressed_data, pmr_vector<bool>&& null_values,
                       const std::shared_ptr<const pmr_vector<size_t>>& offsets, const size_t decompressed_size);

   const pmr_vector<bool>& null_values() const;
```

janehmueller: Oh ic. I forgot to change it in `copy_using_allocator` as well 😅 .

src/lib/storage/create_iterable_from_segment.hpp
janehmueller: Thanks 👍

src/lib/storage/lz4_segment.cpp
janehmueller: No other reason than it being easy to use when not set. We'll change it to `optional`

mrks:
@Bouncner Except for the optional thing, do you have anything left?

PR #1468, opened 2019-02-05
[DYOD18/19] MvccDeletePlugin prep. (Chunk invalidation & delete) 

julianmenzler:
This is the second chunk of HYRISE preparations for our upcoming MvccDeletePlugin. It mainly implements support for physical & logical chunk deletion. An overview of the most important changes as follows:

- class `Table` gets
  - a `delete_chunk(...)` function, which deletes a chunk from the chunk-vector resp. memory. In speak of our MvccDeletePlugin this is termed a _"physical delete"_.
  - an adjusted `row_count()` function, which bypasses physically deleted chunks

- class `Chunk` gets
  - the new attribute `_invalid_row_count`, which is used by the MvccDeletePlugin to determine whether chunks should be cleaned up or not. 
  - the new attribute `_cleanup_commit_id` attribute, which represents the commit ID of a logical delete transaction. The latter is carried out by the MvccDeletePlugin and invalidates a chunk as a whole. This is a necessity before a physical deletion can take place.

- class `GetTable` 
  - is adjusted to exclude logically and physically deleted chunks when used in conjunction with a transaction context. 
    - GetTable still works without transaction context, but in this case it should not be combined with the MVCC Delete functionality.
  - obtains two new Google tests covering the added functionality


src/lib/operators/get_table.cpp
```diff
@@ -42,24 +42,52 @@ std::shared_ptr<AbstractOperator> GetTable::_on_deep_copy(
 
 void GetTable::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeVariant>& parameters) {}
 
-std::shared_ptr<const Table> GetTable::_on_execute() {
+std::shared_ptr<const Table> GetTable::_on_execute() { Fail("GetTable can't be called without transaction context."); }
```
m0r13: It's a bit confusing that this read only operator can be called only with a transaction context, or that the transaction context can be `nullptr` only if there are no discarded chunks. Perhaps some more documentation on this logic would good if there is no better way to handle the required transaction context here. Also we feel that it is a bit inconsistent that you're using `Fail` here but only a `DebugAssert` below to allow `nullptr` transaction context if there are no discarded chunks.

src/lib/operators/get_table.cpp
```diff
@@ -42,24 +42,52 @@ std::shared_ptr<AbstractOperator> GetTable::_on_deep_copy(
...
+
+std::shared_ptr<const Table> GetTable::_on_execute(std::shared_ptr<TransactionContext> transaction_context) {
   auto original_table = StorageManager::get().get_table(_name);
+
+  DebugAssert(transaction_context != nullptr || !_table_contains_discarded_chunks(original_table),
```
m0r13: Consider using a lambda instead of `_table_contains_discarded_chunks` which is only called from here.

src/lib/operators/delete.cpp
```diff
@@ -98,6 +98,7 @@ void Delete::_on_commit_records(const CommitID cid) {
       auto referenced_chunk = referenced_table->get_chunk(row_id.chunk_id);
 
       referenced_chunk->get_scoped_mvcc_data_lock()->end_cids[row_id.chunk_offset] = cid;
+      referenced_chunk->increase_invalid_row_count(1);
```
mrks: Would it be sufficient to do this once at the end?

src/lib/operators/get_table.cpp
mrks: I had to work my way to the class hierarchy to understand what was happening. So AbstractOperator only knows `_on_execute(tx_context)`. `AbstractReadOnlyOperator` introduces `_on_execute() -> _on_execute(nullptr)` and does little more. This operator makes `_on_execute()` illegal. See what's smells here? :)

src/lib/operators/get_table.cpp
```diff
@@ -42,24 +42,52 @@ std::shared_ptr<AbstractOperator> GetTable::_on_deep_copy(
...
+              "Some chunks haven been discarded: They were deleted either logically (fully invalidated) or physically "
+              "(removed from memory)"
+              "Thus, GetTable needs to have a transaction context set to filter discarded chunks out.");
```
mrks: Code style: We keep assertion texts short (so that they fit into one line) and add a code comment if necessary.

src/lib/operators/get_table.cpp
```diff
@@ -42,24 +42,52 @@ std::shared_ptr<AbstractOperator> GetTable::_on_deep_copy(
...
+
+  if (transaction_context) {
```
mrks: A comment on what happens in this block would be nice.

src/lib/operators/get_table.cpp
```diff
@@ -42,24 +42,52 @@ std::shared_ptr<AbstractOperator> GetTable::_on_deep_copy(
...
   const auto pruned_table = std::make_shared<Table>(original_table->column_definitions(), TableType::Data,
                                                     original_table->max_chunk_size(), original_table->has_mvcc());
+
   const auto excluded_chunks_set =
       std::unordered_set<ChunkID>(_excluded_chunk_ids.cbegin(), _excluded_chunk_ids.cend());
   for (ChunkID chunk_id{0}; chunk_id < original_table->chunk_count(); ++chunk_id) {
-    if (excluded_chunks_set.find(chunk_id) == excluded_chunks_set.end()) {
-      pruned_table->append_chunk(original_table->get_chunk(chunk_id));
+    const auto chunk = original_table->get_chunk(chunk_id);
+    if (chunk && excluded_chunks_set.find(chunk_id) == excluded_chunks_set.end()) {
```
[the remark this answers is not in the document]
mrks: I seem to remember that libc++ didn't yet implement that.

src/lib/storage/chunk.hpp
```diff
@@ -151,6 +152,19 @@ class Chunk : private Noncopyable {
    */
   size_t estimate_memory_usage() const;
 
+  void increase_invalid_row_count(uint64_t count) const;
+
+  uint64_t invalid_row_count() const { return _invalid_row_count; }
+
+  /**
+   * Returns the commit-id of the MvccDeletePlugin-logical-delete-transaction
+   * in case it has been carried out.
```
mrks: Please be more precise that "has been carried out". Since this is one of the main header files, it wouldn't hurt to explain the concept a bit more.

src/lib/storage/table.cpp
```diff
@@ -132,6 +133,17 @@ std::shared_ptr<const Chunk> Table::get_chunk(ChunkID chunk_id) const {
   return _chunks[chunk_id];
 }
 
+void Table::delete_chunk(ChunkID chunk_id) {
```
mrks: I believe "remove" is more commonly used. Let's try to reserve "delete" for things that correlate with SQL's delete.

src/lib/storage/table.cpp
```diff
@@ -132,6 +133,17 @@ std::shared_ptr<const Chunk> Table::get_chunk(ChunkID chunk_id) const {
...
+  DebugAssert(chunk_id < _chunks.size(), "ChunkID " + std::to_string(chunk_id) + " out of range");
+  DebugAssert(_chunks[chunk_id]->invalid_row_count() == _chunks[chunk_id]->size(),
+              "Physical delete of chunk prevented: Chunk needs to be fully invalidated before.");
```
mrks: Is `invalid_row_count` thread-safe? If not, this might never be true. Also, we might have other implications that we should be aware of.

src/test/operators/get_table_test.cpp
```diff
@@ -52,4 +56,53 @@ TEST_F(OperatorsGetTableTest, ExcludedChunks) {
...
+  auto delete_all = std::make_shared<opossum::Delete>(vt);
+  delete_all->set_transaction_context(context);
+  delete_all->execute();
+  EXPECT_FALSE(delete_all->execute_failed());
+  context->commit();
+  // not setting cleanup-commit-ids is intentional here
+
+  // Delete chunks physically
+  original_table->delete_chunk(ChunkID{0});
+  original_table->delete_chunk(ChunkID{2});
```
mrks: Test for nullptr?

src/lib/operators/delete.hpp
```diff
@@ -26,12 +26,11 @@ class Delete : public AbstractReadWriteOperator {
       const std::shared_ptr<AbstractOperator>& copied_input_left,
       const std::shared_ptr<AbstractOperator>& copied_input_right) const override;
   void _on_set_parameters(const std::unordered_map<ParameterID, AllTypeVariant>& parameters) override;
-  void _on_commit_records(const CommitID cid) override;
+  void _on_commit_records(CommitID cid) override;
```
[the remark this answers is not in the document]
julianmenzler: The argument `CommitID` is of primitive type and will be copied by value. Therefore, marking it as `const` ist not relevant to function callers. This would be different though, if objects were passed by reference.
Using `const` in the function definition has an effect. It makes `cid` const within the implementation, which is a design choice.

src/lib/storage/chunk.cpp
```diff
@@ -201,5 +201,11 @@ void Chunk::set_statistics(const std::shared_ptr<ChunkStatistics>& chunk_statist
               "ChunkStatistics must have same number of segments as Chunk");
   _statistics = chunk_statistics;
 }
+void Chunk::increase_invalid_row_count(const uint64_t count) const { _invalid_row_count += count; }
+
+void Chunk::set_cleanup_commit_id(CommitID cleanup_commit_id) {
+  DebugAssert(_cleanup_commit_id == MvccData::MAX_COMMIT_ID, "Cleanup commit ID can only be set once.")
```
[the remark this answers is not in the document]
julianmenzler: Correct, I wonder how this compiled 🤔

src/lib/operators/delete.cpp
julianmenzler: Within each iteration `referenced_chunk` changes. Therefore, to perform a "batch-increase" of a chunk's `invalid_row_count`, a container would be required to store the accumulated increase-count per chunk (thinking of `unordered_set`). Then, a single function call of `increase_invalid_row_count(..)` per chunk would be sufficient.

Might be more performant. Do you prefer this approach?

src/lib/operators/get_table.cpp
julianmenzler: Well, this took me some time to understand as well. And I agree with both of you, pretty strange code. It does not really make sense what we built here! 😃 (@d-justen, please also have a look)
We'll address this.

src/lib/operators/get_table.cpp
julianmenzler: Good suggestion!

src/lib/operators/delete.cpp
julianmenzler: we discussed this.

src/lib/operators/get_table.cpp
```diff
@@ -42,24 +42,52 @@ std::shared_ptr<AbstractOperator> GetTable::_on_deep_copy(
...
   if (_excluded_chunk_ids.empty()) {
     return original_table;
   }
 
   // we create a copy of the original table and don't include the excluded chunks
   const auto pruned_table = std::make_shared<Table>(original_table->column_definitions(), TableType::Data,
                                                     original_table->max_chunk_size(), original_table->has_mvcc());
+
   const auto excluded_chunks_set =
       std::unordered_set<ChunkID>(_excluded_chunk_ids.cbegin(), _excluded_chunk_ids.cend());
```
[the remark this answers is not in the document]
engelfa: We did not implement `excluded_chunks_set`. It was already part of Hyrise (83127a543f438dfd05f8a67194fe3e1d96c0c07c). However, we will check the performance impact.

src/lib/statistics/table_statistics.hpp
```diff
@@ -56,6 +56,9 @@ class TableStatistics final {
   // Increases the (approximate) count of invalid rows in the table (caused by deletes).
   void increase_invalid_row_count(uint64_t count);
 
+  // Decreases the (approximate) count of invalid rows in the table (caused by deleted chunks).
```
mrks: "Caused by chunks that were entirely invalid and thus deleted"?

src/lib/operators/get_table.cpp
```diff
@@ -44,18 +44,45 @@ void GetTable::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeV
...
+                    for (ChunkID chunk_id{0}; chunk_id < original_table->chunk_count(); ++chunk_id) {
+                      // Check, whether chunk got deleted either physically (removed from chunk-vector resp. memory).
+                      // or logically (fully invalidated)
+                      if (!original_table->get_chunk(chunk_id) ||
+                          original_table->get_chunk(chunk_id)->get_cleanup_commit_id() < MvccData::MAX_COMMIT_ID)
+                        return false;
+                    }
+                    return true;
+                  }(),
+              "GetTable needs to have a transaction context set to filter out deleted chunks.");
```
mrks: "For tables with logically or physically deleted chunks, the transaction context must be set"?

src/lib/storage/chunk.hpp
```diff
@@ -151,6 +152,19 @@ class Chunk : private Noncopyable {
...
+  void increase_invalid_row_count(uint64_t count) const;
+
+  uint64_t invalid_row_count() const { return _invalid_row_count; }
+
+  /**
+   * The MvccDeletePlugin implements a logical chunk deletion (full chunk invalidation) as a transaction.
+   * In case a clean-up transaction has been performed successfully its commit-id will be returned via
+   * this function. Otherwise, MvccData::MAX_COMMIT_ID is returned.
+   */
+  CommitID get_cleanup_commit_id() const { return _cleanup_commit_id; }
```
mrks: Is the concept of cleanup commit IDs documented somewhere? If not, this would be a good place to do so.

src/lib/storage/table.hpp
```diff
@@ -85,6 +85,7 @@ class Table : private Noncopyable {
   // returns the chunk with the given id
   std::shared_ptr<Chunk> get_chunk(ChunkID chunk_id);
   std::shared_ptr<const Chunk> get_chunk(ChunkID chunk_id) const;
+  void remove_chunk(ChunkID chunk_id);
```
mrzzzrm: This function decl seems a bit squeezed in here... Can you give it its own block and document it?

src/lib/storage/table.cpp
```diff
@@ -132,6 +133,17 @@ std::shared_ptr<const Chunk> Table::get_chunk(ChunkID chunk_id) const {
   return _chunks[chunk_id];
 }
 
+void Table::remove_chunk(ChunkID chunk_id) {
+  DebugAssert(chunk_id < _chunks.size(), "ChunkID " + std::to_string(chunk_id) + " out of range");
+  DebugAssert(_chunks[chunk_id]->invalid_row_count() == _chunks[chunk_id]->size(),
+              "Physical delete of chunk prevented: Chunk needs to be fully invalidated before.");
+  if (_table_statistics) {
+    uint64_t invalidated_rows_count = _chunks[chunk_id]->size();
```
mrzzzrm: just use auto to get rid of the pain of having to pick a data type here ;)

src/lib/storage/chunk.hpp
```diff
@@ -162,6 +176,8 @@ class Chunk : private Noncopyable {
   pmr_vector<std::shared_ptr<BaseIndex>> _indices;
   std::shared_ptr<ChunkStatistics> _statistics;
   bool _is_mutable = true;
+  mutable uint64_t _invalid_row_count = 0;
+  CommitID _cleanup_commit_id = MvccData::MAX_COMMIT_ID;
```
mrzzzrm: I've said it elsewhere, but this is probably the right place: this would probably better be a std::optional

src/lib/storage/chunk.hpp
```diff
@@ -162,6 +176,8 @@ class Chunk : private Noncopyable {
   pmr_vector<std::shared_ptr<BaseIndex>> _indices;
   std::shared_ptr<ChunkStatistics> _statistics;
   bool _is_mutable = true;
+  mutable uint64_t _invalid_row_count = 0;
```
mrzzzrm: It seems like this should be atomic, right?

src/lib/storage/chunk.hpp
```diff
@@ -151,6 +152,19 @@ class Chunk : private Noncopyable {
    */
   size_t estimate_memory_usage() const;
 
+  void increase_invalid_row_count(uint64_t count) const;
```
mrzzzrm: can you doc this function and the function below (briefly?) as to what purpose they serve?

src/test/operators/get_table_test.cpp
```diff
@@ -52,4 +56,55 @@ TEST_F(OperatorsGetTableTest, ExcludedChunks) {
...
+  EXPECT_EQ(gt->get_output()->chunk_count(), 4);
+  auto vt = std::make_shared<opossum::Validate>(gt);
+  vt->set_transaction_context(context);
+  vt->execute();
+  auto delete_all = std::make_shared<opossum::Delete>(vt);
+  delete_all->set_transaction_context(context);
+  delete_all->execute();
+  EXPECT_FALSE(delete_all->execute_failed());
+  context->commit();
+  // not setting cleanup-commit-ids is intentional here
```
mrzzzrm: and why? ;)

src/test/operators/get_table_test.cpp
```diff
@@ -52,4 +56,55 @@ TEST_F(OperatorsGetTableTest, ExcludedChunks) {
...
+  // Invalidate all records to be able to call remove_chunk()
+  auto context = std::make_shared<TransactionContext>(1u, 1u);
+  auto gt = std::make_shared<opossum::GetTable>("tableWithValues");
+  gt->set_transaction_context(context);
+  gt->execute();
+  EXPECT_EQ(gt->get_output()->chunk_count(), 4);
+  auto vt = std::make_shared<opossum::Validate>(gt);
+  vt->set_transaction_context(context);
+  vt->execute();
+  auto delete_all = std::make_shared<opossum::Delete>(vt);
```
mrzzzrm: Maybe add doc "// Delete all rows from table so calling original_table->remove_chunk() below is legal"?

src/test/operators/get_table_test.cpp
```diff
@@ -52,4 +56,55 @@ TEST_F(OperatorsGetTableTest, ExcludedChunks) {
...
+  chunk->set_cleanup_commit_id(CommitID{2u});
+
+  gt->set_transaction_context(context);
+  gt->execute();
+
+  auto table = gt->get_output();
+  EXPECT_EQ(original_table->chunk_count(), 4);
+  EXPECT_EQ(table->chunk_count(), 3);
+}
+
```
mrzzzrm: (Random line)

It seems like you're missing a test that excludes chunks both via `GetTable::set_excluded_chunk_ids()` and the pruning of deleted chunks.

src/lib/operators/get_table.cpp
```diff
@@ -44,18 +44,45 @@ void GetTable::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeV
 
 std::shared_ptr<const Table> GetTable::_on_execute() {
   auto original_table = StorageManager::get().get_table(_name);
+
+  DebugAssert(this->transaction_context_is_set() ||
```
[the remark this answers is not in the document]
mrks: Can't have that constexpr there, but the compiler should be smart enough to figure out what you want

src/lib/operators/get_table.cpp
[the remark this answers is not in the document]
nilsthamm: I know we talked about this during our last meeting. I think you wanted to try out if it works. Have you tried it? @d-justen @engelfa @julianmenzler

src/lib/storage/chunk.cpp
```diff
@@ -201,5 +201,11 @@ void Chunk::set_statistics(const std::shared_ptr<ChunkStatistics>& chunk_statist
               "ChunkStatistics must have same number of segments as Chunk");
   _statistics = chunk_statistics;
 }
+void Chunk::increase_invalid_row_count(const uint64_t count) const { _invalid_row_count += count; }
```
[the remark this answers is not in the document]
engelfa: This was done in the style of `void TableStatistics::increase_invalid_row_count(uint64_t count)`. Should we change it there as well?

src/lib/operators/get_table.cpp
```diff
@@ -44,18 +44,45 @@ void GetTable::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeV
...
+                  }(),
+              "GetTable needs to have a transaction context set to filter out deleted chunks.");
+
+  // Add logically & physically deleted chunks to list of excluded chunks
+  if (this->transaction_context_is_set()) {
+    for (ChunkID chunk_id{0}; chunk_id < original_table->chunk_count(); ++chunk_id) {
+      const auto chunk = original_table->get_chunk(chunk_id);
+
+      if (!chunk || chunk->get_cleanup_commit_id() <= this->transaction_context()->snapshot_commit_id()) {
+        _excluded_chunk_ids.emplace_back(chunk_id);
```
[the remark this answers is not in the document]
engelfa: So you would keep the set unlike [this](https://github.com/hyrise/hyrise/pull/1468/files#r253969489) comment proposed?
We could create a second vector instead of a set, which is initalized with the `_excluded_chunk_ids`, append our chunk ids and then apply sort and unique on the new vector. This could be faster than using a set.

src/lib/operators/get_table.cpp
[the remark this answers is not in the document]
engelfa: Yes, we tried it. It works, but in our test it performed a bit slower than using find.

src/lib/operators/get_table.cpp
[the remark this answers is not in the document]
mrzzzrm: whatever is faster and not too verbose

src/lib/storage/chunk.cpp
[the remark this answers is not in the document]
mrzzzrm: nope, TableStatistics is refactored in another PR

src/lib/storage/chunk.hpp
```diff
@@ -151,6 +152,21 @@ class Chunk : private Noncopyable {
...
+
+  uint64_t invalid_row_count() const { return _invalid_row_count; }
+
+  /**
+   * The MvccDeletePlugin implements a logical chunk deletion (full chunk invalidation) as a transaction.
+   * In case a clean-up transaction has been performed successfully, its commit id will be returned via
+   * this function. Otherwise, an empty _cleanup_commit_id is returned.
+   * The _cleanup_commit_id is used to determine if the chunk got already fully invalidated by the logical
+   * delete and can be physically deleted.
+   */
```
mrks: This comment does not appear to be really helpful for someone looking at the chunk interface. What is a clean-up transaction? Why are chunks "fully invalidated"? ...?

I was thinking about something like "Chunks with few visible entries can be periodically cleaned up by the MvccDeletePlugin. In one step, it deletes the rows from this chunk and re-inserts them add the end of the table. Thus, future transactions will find the still valid rows at the end of the table and do not have to look at this chunk. The cleanup commit ID is the highest snapshot commit ID for which there are still visible."

src/lib/storage/chunk.cpp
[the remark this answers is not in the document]
mrks: `uint64_t Table::row_count() const;` :/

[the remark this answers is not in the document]
mrks: I wouldn't let this block us.

src/lib/operators/get_table.cpp
```diff
@@ -43,19 +43,47 @@ std::shared_ptr<AbstractOperator> GetTable::_on_deep_copy(
...
+    }
+  }
+
+  // Add logically & physically deleted chunks to list of excluded chunks
+  if (transaction_context_is_set()) {
+    for (ChunkID chunk_id{0}; chunk_id < original_table->chunk_count(); ++chunk_id) {
+      const auto chunk = original_table->get_chunk(chunk_id);
+
+      if (!chunk || (chunk->get_cleanup_commit_id() &&
+                     chunk->get_cleanup_commit_id().value() <= transaction_context()->snapshot_commit_id())) {
```
mrks: use `op*` instead of `value()` if you already checked for existence.

mrks:
@mrzzzrm Could you have a final pass or dismiss your review. On first sight, I believe that only the last two points are open. Once those are done, I'd really like to get this merged

src/lib/storage/chunk.hpp
julianmenzler: Great, thank you for your detailed suggestion.

src/lib/storage/table.cpp
julianmenzler: Currently, `invalid_row_count` is not thread-safe. So this `DebugAssert` is a problem. Right!
We have two possibilities:
1. Implement locks to make it thread-safe. This might cost us some performance.
2. Accept that `invalid_row_count` is not thread-safe and live with it as a lazy approach. Thus, we'd have to remove this `DebugAssert`.

What do you suggest?

julianmenzler:
Alright, all changes have been applied. From our side, this PR is ready for merging.

mrks:
Irregularly at my laptop at the moment. Someone press that button when it is enabled, please.

PR #1469, opened 2019-02-05
JIT - Enable outputting table data by reference

FabianWiebe:
This PR introduces a new jittable `JitWriteReference` operator which allows JIT to output table data by reference.

The addition requires changes to the `AbstractJittableSink` interface (implemented by `JitWriteTuples`, `JitAggregate` and now also `JitWriteReference`) to provide access to the input table.
The `JitReadTuples` interface was also modified to include the current chunk id.

### DIFF
--- a/src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
+++ b/src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
@@ -236,7 +236,7 @@ std::shared_ptr<JitOperatorWrapper> JitAwareLQPTranslator::_try_translate_sub_pl
           jit_operator->add_jit_operator(std::make_shared<JitCompute>(jit_expression));
         }
 
-        write_table->add_output_column(column_expression->as_column_name(), jit_expression->result());
+        write_table->add_output_column_definition(column_expression->as_column_name(), jit_expression->result());
       }
 
       jit_operator->add_jit_operator(write_table);
@@ -247,7 +247,7 @@ std::shared_ptr<JitOperatorWrapper> JitAwareLQPTranslator::_try_translate_sub_pl
       for (const auto& column : node->column_expressions()) {
         const auto column_id = input_node->find_column_id(*column);
         DebugAssert(column_id, "Output column must reference an input column");
-        write_table->add_output_column(column->as_column_name(), *column_id);
+        write_table->add_output_column_definition(column->as_column_name(), *column_id);
       }
 
       jit_operator->add_jit_operator(write_table);
--- a/src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
+++ b/src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
@@ -28,7 +28,7 @@
 #include "operators/jit_operator/operators/jit_limit.hpp"
 #include "operators/jit_operator/operators/jit_read_tuples.hpp"
 #include "operators/jit_operator/operators/jit_validate.hpp"
-#include "operators/jit_operator/operators/jit_write_reference.hpp"
+#include "operators/jit_operator/operators/jit_write_references.hpp"
 #include "operators/jit_operator/operators/jit_write_tuples.hpp"
 #include "operators/operator_scan_predicate.hpp"
 #include "storage/storage_manager.hpp"
@@ -216,7 +216,9 @@ std::shared_ptr<JitOperatorWrapper> JitAwareLQPTranslator::_try_translate_sub_pl
   } else {
     if (use_limit) jit_operator->add_jit_operator(std::make_shared<JitLimit>());
 
-    // Check if the output has to be materialized
+    // Data must be materialized if an output column has to be computed. If this is not the case, the data is
+    // outputted by reference which is more efficient as the writing of tuple values into the output table is very
+    // expensive.
     const auto output_must_be_materialized = std::find_if(
         node->column_expressions().begin(), node->column_expressions().end(),
         [&input_node](const auto& column_expression) { return !input_node->find_column_id(*column_expression); });
@@ -240,7 +242,7 @@ std::shared_ptr<JitOperatorWrapper> JitAwareLQPTranslator::_try_translate_sub_pl
       jit_operator->add_jit_operator(write_table);
     } else {
       // Output data by reference
-      auto write_table = std::make_shared<JitWriteReference>();
+      auto write_table = std::make_shared<JitWriteReferences>();
 
       for (const auto& column : node->column_expressions()) {
         const auto column_id = input_node->find_column_id(*column);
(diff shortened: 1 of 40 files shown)


src/lib/operators/jit_operator/operators/jit_write_reference.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+namespace opossum {
+
+/* JitWriteReference must be the last operator in any chain of jit operators.
+ * It is responsible for
+ * 1) adding column definitions to the output table
+ * 2) appending the current row id to the current output chunk
+ * 3) creating a new output chunk and adding output chunks to the output table
+ */
+class JitWriteReference : public AbstractJittableSink {
```
mrks: Maybe add a comment that this is called "JitWriteReference" because it writes one reference for each call of `_consume`. This is probably one of only few places where we don't use the plural.

src/lib/operators/jit_operator/operators/jit_write_reference.hpp
```diff
@@ -0,0 +1,38 @@
...
+ public:
+  struct OutputColumn {
+    const std::string column_name;
+    const ColumnID referenced_column_id;
+  };
+
+  std::string description() const final;
+
+  std::shared_ptr<Table> create_output_table(const Table& in_table) const final;
```
mrks: Please add some documentation here. The interactions of this and `add_output_column` are not clear - I would expect a method `create_output_table` to also create the output columns.

src/lib/operators/jit_operator/operators/jit_write_reference.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+        const auto table_out = ref_segment_in->referenced_table();
+        const auto column_id_out = ref_segment_in->referenced_column_id();
+
+        auto& filtered_pos_list = filtered_pos_lists[pos_list_in];
+
+        if (!filtered_pos_list) {
+          filtered_pos_list = std::make_shared<PosList>();
+          if (pos_list_in->references_single_chunk()) {
+            filtered_pos_list->guarantee_single_chunk();
```
mrks: nice

src/lib/operators/jit_operator/operators/jit_write_reference.cpp
```diff
@@ -0,0 +1,107 @@
...
+  if (context.output_pos_list->size() > 0) {
+    Segments out_segments;
+
+    out_segments.reserve(_output_columns.size());
+
+    const auto chunk_in = in_table->get_chunk(context.chunk_id);
+    if (in_table->type() == TableType::References) {
+      auto filtered_pos_lists = std::map<std::shared_ptr<const PosList>, std::shared_ptr<PosList>>{};
+
+      for (const auto& output_column : _output_columns) {
```
mrks: The order in which methods are called seems to be important, but is not documented / asserted.

src/lib/operators/jit_operator/operators/jit_write_reference.cpp
```diff
@@ -0,0 +1,107 @@
...
+        auto segment_in = chunk_in->get_segment(output_column.referenced_column_id);
+
+        auto ref_segment_in = std::dynamic_pointer_cast<const ReferenceSegment>(segment_in);
+        DebugAssert(ref_segment_in != nullptr, "All columns should be of type ReferenceColumn.");
```
mrks: Did you copy this from somewhere? We stopped having a "ReferenceColumn" a while ago. Also, the reader might miss that this is only true for the `in_table->type() == TableType::References` case.

src/lib/operators/jit_operator/operators/jit_write_reference.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+        const auto pos_list_in = ref_segment_in->pos_list();
+
+        const auto table_out = ref_segment_in->referenced_table();
+        const auto column_id_out = ref_segment_in->referenced_column_id();
+
+        auto& filtered_pos_list = filtered_pos_lists[pos_list_in];
```
mrks: Please add some comments to this method. You can't expect people to know and recognize the (not so pretty) logic of the table scan.

src/lib/operators/jit_operator/operators/jit_write_reference.cpp
```diff
@@ -0,0 +1,107 @@
...
+      context.output_pos_list->guarantee_single_chunk();
+      for (const auto& output_column : _output_columns) {
+        auto ref_segment_out =
+            std::make_shared<ReferenceSegment>(in_table, output_column.referenced_column_id, context.output_pos_list);
+        out_segments.push_back(ref_segment_out);
+      }
+    }
+    out_table.append_chunk(out_segments);
+
+    // Check if the current chunk is the last one
```
mrks: And then... what?

src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
```diff
@@ -203,21 +204,40 @@ std::shared_ptr<JitOperatorWrapper> JitAwareLQPTranslator::_try_translate_sub_pl
...
-    auto write_table = std::make_shared<JitWriteTuples>();
-    for (const auto& column_expression : node->column_expressions()) {
-      const auto jit_expression =
-          _try_translate_expression_to_jit_expression(*column_expression, *read_tuples, input_node);
-      if (!jit_expression) return nullptr;
-      // If the JitExpression is of type JitExpressionType::Column, there is no need to add a compute node, since it
-      // would not compute anything anyway
-      if (jit_expression->expression_type() != JitExpressionType::Column) {
-        jit_operator->add_jit_operator(std::make_shared<JitCompute>(jit_expression));
+    // Check if the output has to be materialized
```
mrks: Yes, that's what I would expect a variable `output_must_be_materialized` to do. A better comment would explain the reasoning behind that.

src/lib/operators/jit_operator/operators/jit_write_reference.hpp
FabianWiebe: Add comments. I also added further comments in `JitReadTuples`.

src/lib/operators/jit_operator/operators/jit_write_reference.hpp
FabianWiebe: I changed `JitWriteReference` to `JitWriteReferences` to keep consistent with the naming convention of `JitReadTuples` and `JitWriteTuples`. `JitWriteTuples` also just writes a single tuple with multiple values in `_consume` but uses the plural of tuple.

src/lib/operators/jit_operator/operators/jit_write_references.hpp
```diff
@@ -0,0 +1,48 @@
...
+  void before_query(Table& out_table, JitRuntimeContext& context) const override;
+
+  // Is called by the JitOperatorWrapper after all tuples of one chunk have been consumed.
+  // This is used to create a reference segment for every output column. The segments are added to a new chunk which is
+  // added to the output table.
+  void after_chunk(const std::shared_ptr<const Table>& in_table, Table& out_table,
+                   JitRuntimeContext& context) const final;
+
+  // The order in which the columns are added defines the order of the columns in the output table.
+  void add_output_column(const std::string& column_name, const ColumnID referenced_column_id);
```
mrks: This is still unclear. `create_output_table` adds column definitions - why do we have `add_output_column` as well?

FabianWiebe: I added a comment explaining that the function is called by the jit-aware LQP translator and that it only defines which columns are in the output table. I renamed `add_output_column` to `add_output_column_definition` to make clear that the function does not add table data.

PR #1473, opened 2019-02-05
[DYOD18/19] Subquery to Join Reformulation

pascalhpi:
This PR adds the subselect to join reformulation rule to the Hyrise optimizer, which allows the reformulation of most `(NOT) EXISTS`, `(NOT) IN` and comparator (`=`, `<>`, `<`, `<=`, `>=`, `>`) subqueries to semi-/anti-multi-predicate-joins.

## Steps
- Identify (in case of `EXISTS`) or derive (in case of `IN` and comparator) a correlated predicate, which is suitable for a join.
- Pull up the predicate until it can replace the correlated predicate with a join.
- Modify aggregate, alias and projection nodes during pull-up to retain all columns needed for the predicate.

## Benchmarks
```
+-----------+-----------------+------+----------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s    | runs | new iter/s     | runs | change | p-value (significant if <0.001) |
+-----------+-----------------+------+----------------+------+--------+---------------------------------+
| TPC-H 1   | 0.534649312496  | 33   | 0.536757349968 | 33   | +0%    |                          0.6507 |
| TPC-H 2   | 8.26157283783   | 496  | 8.45827293396  | 508  | +2%    |                          0.0000 |
| TPC-H 3   | 6.5545258522    | 394  | 6.549451828    | 393  | -0%    |                          0.7039 |
| TPC-H 4   | 1.84837341309   | 111  | 1.89364421368  | 114  | +2%    |                          0.0000 |
| TPC-H 5   | 4.40783548355   | 265  | 4.39982414246  | 264  | -0%    |                          0.6501 |
| TPC-H 6   | 26.7634906769   | 1606 | 26.6178474426  | 1598 | -1%    |                          0.0000 |
| TPC-H 7   | 1.30873858929   | 79   | 1.32820630074  | 80   | +1%    |                          0.0001 |
| TPC-H 8   | 5.68102788925   | 341  | 5.54596567154  | 333  | -2%    |                          0.0021 |
| TPC-H 9   | 1.58829391003   | 96   | 1.51046192646  | 91   | -5%    |                          0.0000 |
| TPC-H 10  | 2.69659328461   | 162  | 2.7321202755   | 164  | +1%    |                          0.0000 |
| TPC-H 11  | 32.5931358337   | 1956 | 32.9183769226  | 1976 | +1%    |                          0.0000 |
| TPC-H 12  | 7.04911851883   | 423  | 7.19940519333  | 432  | +2%    |                          0.0000 |
| TPC-H 13  | 2.19362092018   | 132  | 2.19558572769  | 132  | +0%    |                          0.9026 |
| TPC-H 14  | 17.2678222656   | 1037 | 17.564125061   | 1054 | +2%    |                          0.0000 |
| TPC-H 15  | 12.8383817673   | 771  | 13.2922258377  | 798  | +4%    |                          0.0000 |
| TPC-H 16  | 7.67736816406   | 461  | 7.84662151337  | 471  | +2%    |                          0.0000 |
| TPC-H 17  | 0.149778231978  | 9    | 1.41914093494  | 86   | +847%  |        (not enough runs) 0.0000 |
| TPC-H 18  | 1.89899134636   | 115  | 2.02508997917  | 122  | +7%    |                          0.0000 |
| TPC-H 19  | 5.2240357399    | 314  | 5.3042974472   | 319  | +2%    |                          0.0000 |
| TPC-H 20  | 0.0749611109495 | 5    | 1.97328615189  | 119  | +2532% |        (not enough runs) 0.0000 |
| TPC-H 21  | 0.125653415918  | 8    | 0.429250568151 | 26   | +242%  |        (not enough runs) 0.0000 |
| TPC-H 22  | 9.87333011627   | 593  | 10.0980024338  | 606  | +2%    |                          0.0000 |
| average   |                 |      |                |      | +166%  |                                 |
+-----------+-----------------+------+----------------+------+--------+---------------------------------+
```

## Work in progress
- Overhead from join might slow down execution for small inputs
  - Heuristic should detect these and prevent reformulation
- Testing on LQP level
  - Depends on the LQP implementation of multi-predicate-joins
- Resolve optimizer rule interactions
  - The subselect to join and join reordering reformulation rule interfere with each other in some cases

## Limitations
- Requires semi-/anti-joins, which are currently only supported by hash-joins, so it needs at least one equals-predicate


Felerius:
@Bouncner: Please add @janehmueller, @phnm and Lisa (couldn't find her Github handle) as reviewers.

mrks:
Uuuh, now you ruined my surprise for tomorrow. :( Nice results!
I added @phnm. @janehmueller was already in the group; I requested his review.

src/test/optimizer/strategy/subquery_to_join_reformulation_rule_test.cpp
```diff
@@ -0,0 +1,284 @@
...
+                                                                                  node_table_a,
+                                                                                  node_table_b))));
+  // clang-format on
+  const auto actual_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
+
+  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+}
+
+// We currently do not support this reformulation, because an anti join would not preserve the columns from the right sub-tree.
+TEST_F(SubqueryToJoinReformulationRuleTest, ShouldNotReformulateSimpleCorrelatedNotInWithEqualityPredicate) {
```
janehmueller: If this test doesn't work yet you could also just skip it (via `GTEST_SKIP();`) and don't have commented out code.

src/test/optimizer/strategy/subquery_to_join_reformulation_rule_test.cpp
```diff
@@ -0,0 +1,284 @@
...
+  //                                                                                  node_table_a,
+  //                                                                                  node_table_b))));
+  // clang-format on
+  const auto actual_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
+
+  EXPECT_LQP_EQ(actual_lqp, input_lqp);
+}
+
+// We currently do not support this reformulation, because an anti join would not preserve the columns from the right sub-tree.
+TEST_F(SubqueryToJoinReformulationRuleTest, ShouldNotReformulateSimpleCorrelatedNotInWithLessThanPredicate) {
```
janehmueller: Same here with skipping

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+
+/**
+ * Replace aggregate nodes by ones with additional group by expressions added to allow predicates to be pulled up.
+ */
+void replace_aggregate_nodes(PredicatePullUpInfo& pull_up_info) {
+  // Walk the predicates from bottom to top while keeping track of the column expressions they need to be available.
+  // Whenever an aggregate node is encountered, replace it by a new one with the column expressions of all the
+  // predicates below it added as group by expressions.
+
+  // TODO(anybody): This code does quite a bit of unnecessary copying around because the AggregateNode constructor
```
julisa99: Please complete for final submission and then remove.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+   *
+   * The nodes are ordered top to bottom.
+   */
+  std::vector<std::shared_ptr<AliasNode>> alias_nodes;
+
+  /**
+   * Aggregates which need to be patched allow the predicates to be pulled up.
+   *
+   * Ordered by depth in the sub-tree, from top to bottom. The number notes the number of predicates above this
+   * aggregate in the tree, which won't need be considered when patching the aggregate.
```
julisa99:
```suggestion
   * aggregate in the tree, which won't need to be considered when patching the aggregate.
```

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+  // Walk the predicates from bottom to top while keeping track of the column expressions they need to be available.
+  // Whenever an aggregate node is encountered, replace it by a new one with the column expressions of all the
+  // predicates below it added as group by expressions.
+
+  // TODO(anybody): This code does quite a bit of unnecessary copying around because the AggregateNode constructor
+  // expects two expression vectors (which it then copies again).
+  std::vector<std::shared_ptr<AbstractExpression>> group_by_expressions;
+  auto aggregate_it = pull_up_info.aggregate_nodes.rbegin();
+  auto aggregate_it_end = pull_up_info.aggregate_nodes.rend();
+  for (auto predicate_idx = pull_up_info.predicates.size(); predicate_idx--;) {
```
mrks: This... does not look like a nice for loop. Few people expect the middle part to have side effects.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+    group_by_expressions.emplace_back(pull_up_info.predicates[predicate_idx].right_operand);
+
+    while (aggregate_it != aggregate_it_end && aggregate_it->second == predicate_idx) {
```
mrks: I am having a hard time following this method. Why is the predicate loop the outer one? I would have expected the execution to first iterate over all aggregates, then over all predicates.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+ */
+template <class ParameterPredicate>
+bool uses_correlated_parameters(const std::shared_ptr<AbstractLQPNode>& node,
+                                ParameterPredicate&& is_correlated_parameter) {
+  bool is_correlated = false;
+  for (const auto& expression : node->node_expressions) {
+    visit_expression(expression, [&](const auto& sub_expression) {
+      // We already know that the node is correlated, so we can skip the rest of the expression
+      if (is_correlated) {
+        return ExpressionVisitation::DoNotVisitArguments;
```
mrks: When would this be called?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+      }
+
+      if (sub_expression->type == ExpressionType::LQPSubquery) {
+        // Need to check whether the subquery uses correlated parameters
+        const auto& lqp_select_expression = std::static_pointer_cast<LQPSubqueryExpression>(sub_expression);
+        visit_lqp(lqp_select_expression->lqp, [&](const auto& sub_node) {
+          if (is_correlated) {
+            return LQPVisitation::DoNotVisitInputs;
```
mrks: Also: When would this be called?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+ *   - Usages of correlated parameters in predicate nodes nested below nodes they cannot be pulled up past.
+ */
+template <class ParameterPredicate>
+bool contains_unoptimizable_correlated_parameter_usages(
+    const std::shared_ptr<AbstractLQPNode>& lqp, ParameterPredicate&& is_correlated_parameter,
+    const std::set<std::shared_ptr<PredicateNode>> optimizable_predicate_nodes) {
+  bool optimizable = true;
+  visit_lqp(lqp, [&](const auto& node) {
+    if (!optimizable) {
+      return LQPVisitation::DoNotVisitInputs;
```
mrks: When would this be called?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+          if (is_correlated) {
+            return LQPVisitation::DoNotVisitInputs;
+          }
+
+          is_correlated |= uses_correlated_parameters(sub_node, is_correlated_parameter);
+          return is_correlated ? LQPVisitation::DoNotVisitInputs : LQPVisitation::VisitInputs;
+        });
+      } else if (sub_expression->type == ExpressionType::CorrelatedParameter) {
+        const auto& parameter_expression = std::static_pointer_cast<CorrelatedParameterExpression>(sub_expression);
+        is_correlated |= is_correlated_parameter(parameter_expression->parameter_id);
```
mrks: @mrzzzrm Is this really the best way to do this?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+  // We are only interested in predicate, projection, validate, aggregate and sort nodes. These only ever have one
+  // input, thus we can scan the path of nodes linearly. This makes it easy to track which projection nodes actually
+  // need to be removed, and which predicates need to be considered when patching aggregates.
+  PredicatePullUpInfo info{};
+  size_t num_projections_to_remove = 0;
+  size_t num_aggregates_to_patch = 0;
+
+  auto node = lqp;
+  while (node != nullptr) {
+    if (node->type == LQPNodeType::Projection) {
```
mrks: switch?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+    } else if (node->type == LQPNodeType::Aggregate) {
+      info.aggregate_nodes.emplace_back(std::static_pointer_cast<AggregateNode>(node), info.predicates.size());
+    } else if (node->type == LQPNodeType::Alias) {
+      info.alias_nodes.emplace_back(std::static_pointer_cast<AliasNode>(node));
+    } else if (node->type != LQPNodeType::Validate && node->type != LQPNodeType::Sort) {
+      // It is not safe to pull up predicates past this node, stop scanning
+      break;
+    }
+
+    DebugAssert(!node->right_input(), "Scan only implemented for nodes with one input");
```
mrks: What "scan" do you refer to?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+   */
+  std::vector<std::pair<std::shared_ptr<AggregateNode>, size_t>> aggregate_nodes;
+};
+
+/**
+ * Checks whether a predicate node should be turned into a join predicate.
+ */
+std::optional<PredicateInfo> should_become_join_predicate(
+    const std::shared_ptr<PredicateNode>& predicate_node,
+    const std::map<ParameterID, std::shared_ptr<AbstractExpression>>& parameter_mapping, bool is_below_aggregate) {
```
mrks: Not necessarily related to this line: LQPs can contain multiple aggregates - is that something that is correctly covered?
Example: `SELECT AVG(x) FROM (SELECT SUM(f) AS x FROM t GROUP BY i) t2`

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+    info.right_operand = right_side;
+  } else if (right_side->type == ExpressionType::CorrelatedParameter) {
+    info.condition = flip_predicate_condition(info.condition);
+    parameter_id = std::static_pointer_cast<CorrelatedParameterExpression>(right_side)->parameter_id;
+    info.right_operand = left_side;
+  } else {
+    return std::nullopt;
+  }
+
+  if (!predicate_node->find_column_id(*info.right_operand)) {
```
mrks: Please document this failure case.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+    return std::nullopt;
+  }
+
+  auto expression_it = parameter_mapping.find(parameter_id);
```
mrks: Please document this failure case.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+  for (auto it = join_predicates.begin(), end = join_predicates.end(); it != end; ++it) {
+    const auto& predicate = **it;
+    if (predicate.predicate_condition == PredicateCondition::Equals) {
+      std::iter_swap(join_predicates.begin(), it);
+      break;
+    }
+  }
+
+  if (join_predicates.empty() || join_predicates.front()->predicate_condition != PredicateCondition::Equals) {
+    return _apply_to_inputs(node);
```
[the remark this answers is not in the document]
Bouncner: Yes. But at the moment only for single-predicate joins.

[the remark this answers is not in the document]
Bouncner: A comment would be good though.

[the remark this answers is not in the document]
mrks: There is a comment a couple of lines above. Will you keep this in mind once both are merged?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+  // Collect all join predicates into one place
+  for (auto& predicate_info : pull_up_info.predicates) {
+    // Note that we cannot remove the predicate nodes here just yet, because the reformulation might still fail when no
+    // valid primary join predicate is found.
+    join_predicates.emplace_back(std::make_shared<BinaryPredicateExpression>(
+        predicate_info.condition, predicate_info.left_operand, predicate_info.right_operand));
+  }
+
+  // Semi and anti joins are currently only implemented by hash joins. These need an equals comparison as the primary
+  // join predicate.
```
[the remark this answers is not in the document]
Bouncner: And even when having multiple join predicates and the sort-merge join is fully supporting all the remaining modes: using equality predicates as the main join predicate should be a good heuristic that should make sense in the vast majority of cases. So you might want to keep that even with the improved SM join.

src/lib/optimizer/strategy/subquery_to_join_reformulation_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+    group_by_expressions.emplace_back(pull_up_info.predicates[predicate_idx].right_operand);
+
+    while (aggregate_it != aggregate_it_end && aggregate_it->second == predicate_idx) {
+      auto& aggregate_node = aggregate_it->first;
+      // Add all group by expressions of the original aggregate node unless they are already in the vector. After
+      // creating the new aggregate node, the added expressions will be removed again.
+      auto original_size = group_by_expressions.size();
+      for (size_t group_by_idx = 0; group_by_idx < aggregate_node->aggregate_expressions_begin_idx; ++group_by_idx) {
+        auto& group_by_expression = aggregate_node->node_expressions[group_by_idx];
+        auto it = std::find(group_by_expressions.begin(), group_by_expressions.end(), group_by_expression);
```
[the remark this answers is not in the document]
Felerius: In my tests sets started having faster lookup somewhere between 10-20 items. Not only should 10+ GROUP BY expressions should be incredibly rare, but also all sets have a much higher setup cost than an unsorted vector.

src/lib/optimizer/strategy/subquery_to_join_reformulation_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+  }
+}
+
+/**
+ * Replace alias nodes by ones which preserve all columns from below.
+ */
+void replace_alias_nodes(PredicatePullUpInfo& pull_up_info) {
+  // Replace the nodes bottom to top, in case of two alias nodes being adjacent.
+  auto& alias_nodes = pull_up_info.alias_nodes;
+  for (auto alias_it = alias_nodes.rbegin(), alias_it_end = alias_nodes.rend(); alias_it != alias_it_end; ++alias_it) {
```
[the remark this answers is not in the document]
Felerius: The contributor guidelines mention to prefer calling `end()` only once

src/lib/optimizer/strategy/subquery_to_join_reformulation_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+  auto right_tree_root = subquery_expression->lqp;
+
+  // We cannot support anti joins right now (see large comment below on multi-predicate joins for the reasons).
+  if (join_mode != JoinMode::Semi) {
+    _apply_to_inputs(node);
+    return;
+  }
+
+  // Do not reformulate if expected output is small.
+  if (node->get_statistics()->row_count() < 100.0f) {
```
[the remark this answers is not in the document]
Felerius: 100 is only a placeholder until we have proper benchmarks for a threshold, which is still work in progress. Should probably have been commented.

resources/test_data/sqlite_testrunner_queries.sql
```diff
@@ -231,6 +231,15 @@ SELECT * FROM id_int_int_int_100 WHERE a * 10 NOT IN (SELECT b FROM mixed)
 SELECT a FROM id_int_int_int_100 WHERE a IN (SELECT b FROM mixed)
 SELECT a, b FROM id_int_int_int_100 WHERE a IN (SELECT b FROM mixed)
 
+SELECT * FROM id_int_int_int_100 WHERE a NOT IN (SELECT a FROM id_int_int_int_50)
+SELECT * FROM id_int_int_int_100 AS r WHERE a IN (SELECT s.a FROM id_int_int_int_50 AS s WHERE s.b = r.b)
+SELECT * FROM id_int_int_int_100 AS r WHERE a IN (SELECT s.a FROM id_int_int_int_50 AS s WHERE s.b < r.b)
+SELECT a FROM id_int_int_int_100 AS r WHERE a IN (SELECT s.a FROM id_int_int_int_50 AS s WHERE s.b < r.b) -- fails because there is a duplicate in r that is mistakenly removed
```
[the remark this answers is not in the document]
Felerius: This references the behavior of the multi-predicate-join workaround/hack to remove duplicates from the left subtree of the created join. I removed the comment since the test doesn't fail anymore and the workaround is temporary anyway.

src/test/optimizer/strategy/subquery_to_join_rule_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+                                                                                  node_table_a,
+                                                                                  node_table_b))));
+  // clang-format on
+  const auto actual_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
+
+  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+}
+
+// We currently do not support this reformulation, because an anti join would
+// not preserve the columns from the right sub-tree.
```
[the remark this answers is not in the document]
Felerius: Yes (the comment references the fact that we cannot put additional predicates above an anti-join to simulate multi-predicate joins)

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+    }
+
+    // Check that the comparison expression is a column expression of the left input tree so that it can be turned into
+    // a join predicate.
+    if (!left_tree_root->find_column_id(*comparison_expression)) {
+      _apply_to_inputs(node);
+      return;
+    }
+
+    // Check that the subquery returns a single column, and build a join predicate with it.
```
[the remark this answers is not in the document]
Felerius: We weren't quite sure if INs with tuples might be allowed (e.g. `WHERE (a, b) IN (SELECT c, d FROM ...`). From all that I can find SQL doesn't support this, so an assert is indeed better.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+  //    - Projection nodes might remove the required columns, so they are simply removed. Since we join with a
+  //      semi/anti join, this does not change the semantics. We run before ColumnPruningRule, which then re-adds
+  //      appropriate pruning projections.
+  //    - Alias nodes are due to the same reasons as projection nodes. Since we can't remove them (that would remove
+  //      the aliased columns) we simply extend them to only add and not remove any columns.
+  //    - Aggregate nodes also remove columns. For now, we only support pulling equals predicates above them. For that
+  //      to work, we extend the aggregate nodes to also group by the columns referenced by all predicates pulled from
+  //      below to above them.
+  //
+  // The only other nodes we pull predicates past are sort and validate nodes, which don't need to be adjusted.
```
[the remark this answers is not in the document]
Felerius: This means that any sub-LQP has to stay the same semantically, since it might be the input to multiple nodes, right? In this case I think we need to rewrite how we remove predicate nodes, since right now we just remove them from the input LQP (we would probably need to clone everything above the predicate nodes in the tree).

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+  auto pull_up_info = prepare_predicate_pull_up(right_tree_root, parameter_mapping);
+  auto is_correlated_parameter = [&](ParameterID id) { return parameter_mapping.find(id) != parameter_mapping.end(); };
+  if (contains_unoptimizable_correlated_parameter_usages(right_tree_root, is_correlated_parameter,
+                                                         pull_up_info.predicate_nodes)) {
```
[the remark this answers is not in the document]
Felerius: I changed it to pass `parameter_mapping` around directly (we introduced the lambda at some point because the check was used in a lot of places, now its only used once)

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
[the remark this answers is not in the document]
Felerius: Shouldn't the optimized version still be faster even if there isn't an equals predicate? It will probably still be slow in most cases, but should still be faster than the unoptimized version, right?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+    const auto secondary_predicate_node = PredicateNode::make(std::move(secondary_join_predicate));
+    lqp_insert_node(parent, LQPInputSide::Left, secondary_predicate_node);
+    parent = secondary_predicate_node;
+  }
+
+  lqp_insert_node(parent, LQPInputSide::Left, join_node);
+  join_node->set_right_input(right_tree_root);
+
+  // By removing/replacing nodes only now, the right input of the join node will be automatically kept up to date when
+  // we remove/replace the root of the previous subquery.
```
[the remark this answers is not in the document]
Felerius: If we remove the nodes before building the join node (which a previous version did), we need to keep track of the root of the right sub-tree. Doing it this way means that `lqp_remove_node` does that for us, since it is registered as an input to the join node. I admit the comment is hard to understand without the context, and maybe also unnecessary.

mrzzzrm:
Can you give me a short summary about the synchronization of this PR and the MPJ PR? As to my understanding this rule is highly dependant on MPJ as soon as it wants to rewrite a subquery with multiple predicates (I am unconvinced that the insertion of the AggregateNode is semantically correct, but that's another point). Will we merge the MPJ PR and the finish this one, or how are we going to integrate this PR?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
mrzzzrm: just use reverse iterators

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
mrzzzrm: Giving up here as well (might be because it's late...) 

please revise this function, provide examples of the rewrite happening here

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,573 @@
...
+
+/**
+ * Replace aggregate nodes by ones with additional group by expressions added to allow predicates to be pulled up.
+ */
+void replace_aggregate_nodes(PredicatePullUpInfo& pull_up_info) {
+  // Walk the predicates from bottom to top while keeping track of the column expressions they need to be available.
+  // Whenever an aggregate node is encountered, replace it by a new one with the column expressions of all the
+  // predicates below it added as group by expressions.
+
+  // TODO(anybody): This code does quite a bit of unnecessary copying around because the AggregateNode constructor
```
mrzzzrm: The AggregateNode is not going to change in that regard, pls adjust the rule and/or remove this TODO

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+      group_by_expressions.resize(original_size);
+      aggregate_it++;
+    }
+  }
+}
+
+/**
+ * Replace alias nodes by ones which preserve all columns from below.
+ */
+void replace_alias_nodes(PredicatePullUpInfo& pull_up_info) {
```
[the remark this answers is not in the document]
mrzzzrm: Absolutely agree!

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+  // Do not reformulate if expected output is small.
+  if (node->get_statistics()->row_count() < 100.0f) {
+    _apply_to_inputs(node);
+    return;
+  }
+  std::cout << "node cost before in reformulation: " << CostModelLogical().estimate_plan_cost(node) << '\n';
+
+  // TODO(anybody): Is this check actually necessary, or is this always true for correlated parameters?
+  // Check that all correlated parameters expressions are column expressions of the left sub-tree. Otherwise, we won't
+  // be able to use them in join predicates. Also build up a map from parameter ids to their respective expressions.
```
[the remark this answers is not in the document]
mrks: This one is still open.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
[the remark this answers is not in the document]
mrks: Diamond shapes are quite uncommon (at least at the moment). You could also let your rule skip a node if it has multiple outputs.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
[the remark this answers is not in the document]
mrks: If it just explains history that we don't know about, I'd remove it

Felerius:
Note to anyone reviewing: Since the multi-predicate join PR (#1482) seems largely done, we merged its changes into this branch so that we can start writing LQP tests and running benchmarks. Until #1482 is merged, this means that their changes also appear in the diff of this PR

src/lib/optimizer/optimizer.hpp
```diff
@@ -21,6 +22,14 @@ class Optimizer final {
 
   void add_rule(std::unique_ptr<AbstractRule> rule);
 
+  template <class T>
```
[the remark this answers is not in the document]
Felerius: This is used in our benchmark in the micro benchmarks to create a default optimizer without the `subquery_to_join` rule. Is there any better way to do this?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,573 @@
...
+#include "utils/assert.hpp"
+
+namespace opossum {
+
+namespace {
+
+/**
+ * Info about a predicate that needs to be pulled up into a join predicate.
+ */
+struct PredicateInfo {
```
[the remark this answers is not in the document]
Felerius: The idea was to have a struct where the operands are ordered by the subtrees they come from. Since we only use it in one vector in `PredicatePullUpInfo`, we can just have it as a comment there.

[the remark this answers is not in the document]
Felerius: On a related note: `BinaryPredicateExpression` does not seem to be very usable without a `shared_ptr`/`unique_ptr` around it. For me it was throwing compile errors every time the move constructor was used, for example when swapping or when assigning an `std::optional<BinaryPredicateExpression>`. Is this intentional?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
Felerius: Multiple aggregates are supported, yes. For example, the following query gets reformulated correctly:
```sql
SELECT *
FROM customer
WHERE c_acctbal > (
  SELECT SUM(a.x)
  FROM (
    SELECT AVG(o_totalprice) AS x
    FROM orders
    WHERE o_custkey = c_custkey
    GROUP BY o_orderpriority
  ) AS a
)
```
We will also have tests for cases like this very soon.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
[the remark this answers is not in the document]
mrzzzrm: Not by design, but as the expression calsses have some const members they cannot be copied/moved into. If you need swapping/assigning you're probably actually better off using your own struct to represent the expression. The expression classes could be turned into copyable types (and probably should at some point) but I wouldn't want to do that as part of this PR if there is only a single use case for this.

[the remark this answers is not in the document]
Felerius: I wrapped them in shared_ptrs for now. Since `JoinNode` expect its predicates as `shared_ptr<AbstractExpression>` anyways, there shouldn't be much overhead except for a few additional indirections.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,573 @@
...
+  // contains_unoptimizable_correlated_parameter_usages.
+  if (predicate_node->predicate()->type != ExpressionType::Predicate) {
+    return std::nullopt;
+  }
+
+  const auto& predicate_expression = std::static_pointer_cast<AbstractPredicateExpression>(predicate_node->predicate());
+
+  // Joins only support these six binary predicates. We rely on PredicateSplitUpRule having split up ANDed chains of
+  // such predicates previously, so that we can process them separately.
+  auto cond_type = predicate_expression->predicate_condition;
```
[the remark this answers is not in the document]
Felerius: I replaced it by a switch. Maybe this could be factored into a helper method, since `OperatorJoinPredicate::from_expression` does the exact same check.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,573 @@
...
+  if (cond_type != PredicateCondition::Equals && cond_type != PredicateCondition::NotEquals &&
+      cond_type != PredicateCondition::LessThan && cond_type != PredicateCondition::LessThanEquals &&
+      cond_type != PredicateCondition::GreaterThan && cond_type != PredicateCondition::GreaterThanEquals) {
+    return std::nullopt;
+  }
+
+  // We can currently only pull equals predicates above aggregate nodes. The other predicate types could be supported
+  // but require more sophisticated reformulations.
+  if (is_below_aggregate && cond_type != PredicateCondition::Equals) {
```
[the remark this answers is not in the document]
Felerius: We group by the attribute that gets compared in the equals predicate (so instead of summing all rows where that attribute has a specific value, we group by the attribute and build the sum for each group)

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,573 @@
...
+
+  info.left_operand = expression_it->second;
+  return info;
+}
+
+/**
+ * Finds predicate nodes to pull up to become join predicates and other nodes that need to be removed or patched.
+ *
+ * This selects all predicates that can be safely pulled up and turned into join predicates. It also collects all
+ * projection nodes above the selected predicates. These need to be removed to make sure that the required columns are
```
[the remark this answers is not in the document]
Felerius: Right now re remove the projection nodes entirely and rely on `ColumnPruningRule` to put sensible projections back in for us

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,573 @@
...
+        // All projections/aggregates found so far need to be removed/patched
+        num_projections_to_remove = info.projection_nodes.size();
+        num_aggregates_to_patch = info.aggregate_nodes.size();
+      }
+    } else if (node->type == LQPNodeType::Aggregate) {
+      info.aggregate_nodes.emplace_back(std::static_pointer_cast<AggregateNode>(node), info.predicates.size());
+    } else if (node->type == LQPNodeType::Alias) {
+      info.alias_nodes.emplace_back(std::static_pointer_cast<AliasNode>(node));
+    } else if (node->type != LQPNodeType::Validate && node->type != LQPNodeType::Sort) {
+      // It is not safe to pull up predicates past this node, stop scanning
```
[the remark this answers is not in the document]
Felerius: Yes (note the `!=` instead of `==`). This should be far more readable in the new version which uses `switch` instead of an `if` chain

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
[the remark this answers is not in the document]
mrzzzrm: that justification reads a bit odd. Are you aware that the ProjectionNode might be used to calculate expressions (such as `a + b - c`) for subsequent scans/aggregates/joins/sorts?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
Felerius: This references an old name for this method (scan_for_join_predicates or something similar). I replaced it with a more sensible message

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
Felerius: This was so that the rest of the visitation was cancelled as fast as possible after we found a correlated parameter (since returning `DoNotVisitArguments` only applies to the arguments of the correlated parameter expression)

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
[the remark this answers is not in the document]
Felerius: For some reason we didn't consider this at all. In this case we need to handle projection nodes more like aggregate nodes and only add the columns we need. This might actually be a chance to clean up our LQP manipulation (by having less separate logic for alias, aggregate and projection nodes) and handling the diamond LQP case.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,573 @@
...
+    visit_expression(expression, [&](const auto& sub_expression) {
+      // We already know that the node is correlated, so we can skip the rest of the expression
+      if (is_correlated) {
+        return ExpressionVisitation::DoNotVisitArguments;
+      }
+
+      if (sub_expression->type == ExpressionType::LQPSubquery) {
+        // Need to check whether the subquery uses correlated parameters
+        const auto& lqp_select_expression = std::static_pointer_cast<LQPSubqueryExpression>(sub_expression);
+        visit_lqp(lqp_select_expression->lqp, [&](const auto& sub_node) {
```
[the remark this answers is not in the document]
Felerius: I thought we had an example at some point that required this, but that must have been an unrelated bug then.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
Felerius: As in the case above, this is to stop the visitation as fast as possible once we have found something that prevents optimization.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,573 @@
...
+
+  const auto predicate_node = std::static_pointer_cast<PredicateNode>(node);
+  const auto predicate_node_predicate = predicate_node->predicate();
+
+  const auto& left_tree_root = node->left_input();
+  std::vector<std::shared_ptr<BinaryPredicateExpression>> join_predicates;
+  std::shared_ptr<LQPSubqueryExpression> subquery_expression;
+  JoinMode join_mode;
+
+  if (predicate_node_predicate->type != ExpressionType::Predicate &&
```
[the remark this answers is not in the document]
Felerius: This was an attempt to break up a very long if - else if - else construct by having the last else case first. As you've noted below, a switch is probably better, so I'll change it to that

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,564 @@
...
+    }
+
+    auto new_alias_node = AliasNode::make(std::move(expressions), std::move(aliases));
+    lqp_replace_node(alias_node, new_alias_node);
+  }
+}
+
+std::string SubqueryToJoinRule::name() const { return "Subquery to Join Rule"; }
+
+void SubqueryToJoinRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node) const {
```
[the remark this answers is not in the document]
Felerius: Do you think the comment below works as a first overview?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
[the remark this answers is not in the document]
Felerius: The check is removed now

pascalhpi:
Regarding the heuristic when to use the subquery to join rule and when not: We did some benchmarks and got strange results. For the situation we benchmarked (see subquery_to_join_benchmark.cpp) the only case where performance is worse with our rule is when the left side is very small (< 5) and the right side is about 1000. Does someone have an idea how this non-monotone behaviour can be explained? Further, I would suggest we always apply our rule for now. This seems to be the safest strategy performance wise, since even in these rare cases described above performance is still good, just a little worse compared to not applying our rule.

query:
```
SELECT *
FROM T1
WHERE EXISTS (
  SELECT *
  FROM T2
  WHERE T1.id == T2.id
)
```

t1 and t2 each have a single int column, filled with 0s.

```
t1_size/t2_size  without  with
--------------------------------------
1/1              5360     7457
1/5              5418     7854
1/10             5331     7639
1/50             5387     6808
1/100            5256     7495
1/500            5034     >     3584
1/1000           4686     >     3465
1/5000           2099     >     1752
2/5              4547     7725
2/10             4822     7766
2/50             4544     7653
2/100            4665     7277
2/500            4285     4384
2/1000           3747     >     3478
2/5000           1388     1819
3/1              4349     7835
3/5              4445     7310
3/10             4451     7702
3/50             4339     7636
3/100            4286     7042
3/500            3730     4491
3/1000           3257     >     2713
3/5000           1107     1745
4/1              4234     7696
4/5              4088     7728
4/10             4180     7515
4/50             4143     7632
4/100            3297     7498
4/500            3478     4497
4/1000           2920     >     2705
4/5000           870      1797
5/1              4111     7796
5/5              4044     6889
5/10             4074     7126
5/50             3834     7237
5/100            3840     7373
5/500            3204     4185
5/1000           2643     3872
5/5000           772      1697
10/1             3163     7810
10/5             3107     6696
10/10            3051     7195
10/50            2955     7667
10/100           2845     7482
10/500           2221     4365
10/1000          1728     4015
10/5000          359      1711
50/1             705      7478
50/5             703      7589
50/10            1112     7600
50/50            663      7548
50/100           511      7350
50/500           392      4422
50/1000          341      3959
50/5000          93       1848
100/1            377      7519
100/5            445      7507
100/10           371      7376
100/50           410      7378
100/100          333      7179
100/500          205      4385
100/1000         170      3934
100/5000         45       1772
500/1            96       6384
500/5            96       6357
500/10           94       6363
500/50           88       6369
500/100          83       5474
500/500          42       3936
500/1000         32       3571
500/5000         13       1541
1000/1           41       5342
1000/5           69       5491
1000/10          40       5398
1000/50          63       5391
1000/100         58       5355
1000/500         25       3468
1000/1000        18       2367
1000/5000        7        2162
5000/1           8        1680
5000/5           14       1773
5000/10          13       1740
5000/50          8        1692
5000/100         8        1737
5000/500         7        1647
5000/1000        3        1585
5000/5000        1        937
```

Felerius:
@Bouncner @mrks @mrzzzrm: I have rewritten the way we do predicate pull-up and LQP manipulation to fix some issues (see below) and would really appreciate some feedback on it. 90% of the changes are in `attempt_predicate_pull_up` and its sub-methods. Basically, we now first scan for all usages of correlated parameters, counting the number of correlated predicate nodes. We then recursively scan for correlated predicate nodes that we can pull up, and (if we can pull them all up), change the LQP while unwinding from the recursion.

Things fixed by this:
 * Removing projections adding new columns (discussed [here](https://github.com/hyrise/hyrise/pull/1473#discussion_r261889159)): now they are adapted instead of removed
 * Handling of nodes with more than one output (discussed [here](https://github.com/hyrise/hyrise/pull/1473#discussion_r259010192)): the changed part of subqueries LQP is completely copied and adapted, so references to nodes in the original LQP work as expected
 * The overly convoluted patching of aggregate nodes (discussed [here](https://github.com/hyrise/hyrise/pull/1473#discussion_r259024890)): this is now implemented much simpler and more readable

Felerius:
> Can you please add a short comment that we willingly never check for table sizes and reformulate, because benchmarks showed that it’s not necessary?

Added a comment in fe602c9 (and a33ab26, because I can't type)

Felerius:
One question regarding testing: Right now our LQP tests are not consistent, since `EXPECT_LQP_EQ` tests the order of node expressions in projection nodes (also alias and aggregate nodes). However, the order in which our rule adds them depends on the iteration order of an `unordered_set` (`PredicatePullUpInfo::required_column_expressions`), which, as far as I know, is unspecified.

Performance wise it would probably be fine to replace the `unordered_set` by a `vector` and to do a linear search when inserting to avoid duplicates, but it feels wrong to do this just for the sake of testing. Whats your take on this?

mrzzzrm:
> Performance wise it would probably be fine to replace the unordered_set by a vector and to do a linear search when inserting to avoid duplicates, but it feels wrong to do this just for the sake of testing. Whats your take on this?

Use a std::vector (or otherwise order-stable data structure) in your code if perf is not critical, which in this case I think it isn't. A pain, I know...

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,583 @@
...
+  // Attempt to pull up all correlated predicate nodes.
+  auto maybe_pull_up_info = attempt_predicate_pull_up(input_info.subquery_expression.lqp, parameter_mapping,
+                                                      correlated_predicate_node_count, false);
+  if (!maybe_pull_up_info) {
+    _apply_to_inputs(node);
+    return;
+  }
+
+  auto pull_up_info = *maybe_pull_up_info;
+  if (input_info.additional_join_predicate) {
```
mrzzzrm: why would there be another join predicate in input_info that is not in pull_up_info? Please doc this.

src/lib/optimizer/strategy/subquery_to_join_rule.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <unordered_map>
+
+#include "abstract_rule.hpp"
+#include "expression/abstract_expression.hpp"
+
+namespace opossum {
+
+class AbstractLQPNode;
+
+// Optimizes (NOT) IN and (NOT) EXISTS expressions into semi/anti joins.
```
mrzzzrm: Should also optimize >,>=, etc with subquery, no?

src/test/optimizer/strategy/subquery_to_join_rule_test.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+using namespace opossum::expression_functional;  // NOLINT
+
+namespace opossum {
+
+class SubqueryToJoinRuleTest : public StrategyBaseTest {
+ public:
+  void SetUp() override {
+    node_a = MockNode::make(MockNode::ColumnDefinitions{{DataType::Int, "a"}, {DataType::Int, "b"}}, "a");
+    a_a = {node_a, ColumnID{0}};
```
mrzzzrm: you could also do `node_a->get_column("a")`, which imo reads a bit nicer and is safer

Bouncner:
> Use a std::vector (or otherwise order-stable data structure) in your code if perf is not critical, which in this case I think it isn't. A pain, I know...

What about std::set? As far as I know, iteration is guaranteed to be ordered (haven’t looked at the code though).

mrzzzrm:
> What about std::set? As far as I know, iteration is guaranteed to be ordered

It is, but neither AbstractLQPNode nor AbstractExpression support operator< atm

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,583 @@
...
+  }
+
+  // Check that one side of the expression is a correlated parameter and the other a column expression of the LQP below
+  // the predicate node (required for turning it into a join predicate). Also order the left/right operands by the
+  // subtrees they originate from.
+  const auto& binary_predicate_expression = std::static_pointer_cast<BinaryPredicateExpression>(predicate_expression);
+  const auto& left_side = binary_predicate_expression->left_operand();
+  const auto& right_side = binary_predicate_expression->right_operand();
+  auto ordered_predicate = binary_predicate_expression->deep_copy();
+  ParameterID parameter_id;
```
pascalhpi:
```suggestion
  ParameterID correlated_parameter_id;
```

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,583 @@
...
+      return nullptr;
+  }
+
+  // We can currently only pull equals predicates above aggregate nodes (by grouping by the column that the predicate
+  // compares with). The other predicate types could be supported but would require more sophisticated reformulations.
+  if (is_below_aggregate && predicate_condition != PredicateCondition::Equals) {
+    return nullptr;
+  }
+
+  // Check that one side of the expression is a correlated parameter and the other a column expression of the LQP below
```
pascalhpi: If I understand the code correctly we just make sure we get one correlated parameter and then assume the other side must be "a column expression of the LQP below the predicate node". This is however not checked, contrary to what this comment implies, right? Also, what if left_side and right_side are both correlated parameters?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,583 @@
...
+ */
+std::shared_ptr<AggregateNode> adapt_aggregate_node(
+    const AggregateNode& node, const std::vector<std::shared_ptr<AbstractExpression>>& required_column_expressions) {
+  std::vector<std::shared_ptr<AbstractExpression>> group_by_expressions(
+      node.node_expressions.cbegin(), node.node_expressions.cbegin() + node.aggregate_expressions_begin_idx);
+  ExpressionUnorderedSet original_group_by_expressions(group_by_expressions.cbegin(), group_by_expressions.cend());
+
+  const auto not_found_it = original_group_by_expressions.cend();
+  for (const auto& expression : required_column_expressions) {
+    if (original_group_by_expressions.find(expression) == not_found_it) {
```
pascalhpi: Although I don't like this pattern to check whether something is in a container I'm still expecting to see it, so the extra variable confused me unnecessarily.
```suggestion
    if (original_group_by_expressions.find(expression) == original_group_by_expressions.cend()) {
```

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,583 @@
...
+}
+
+/**
+ * Copy an aggregate node and adapt it to group by all required columns.
+ */
+std::shared_ptr<AggregateNode> adapt_aggregate_node(
+    const AggregateNode& node, const std::vector<std::shared_ptr<AbstractExpression>>& required_column_expressions) {
+  std::vector<std::shared_ptr<AbstractExpression>> group_by_expressions(
+      node.node_expressions.cbegin(), node.node_expressions.cbegin() + node.aggregate_expressions_begin_idx);
+  ExpressionUnorderedSet original_group_by_expressions(group_by_expressions.cbegin(), group_by_expressions.cend());
```
pascalhpi: Just wanted to mention again that I prefer the "auto to stick"-style, see [here why](https://www.fluentcpp.com/2018/09/28/auto-stick-changing-style/).
```suggestion
  auto original_group_by_expressions = ExpressionUnorderedSet(group_by_expressions.cbegin(), group_by_expressions.cend());
```

mrzzzrm:
@Felerius & co. I have locally merged the master into this via 
```
git revert -m 1 f37d61fe2
git pull origin master
```

which had just minimal conflicts. I don't want to push to you branch to avoid conflicts with anything you might have locally. Can you run the listed commands and push?
Then I'll look intensely at this PR, so we can get it in asap.

Felerius:
> Can you run the listed commands and push?
> Then I'll look intensely at this PR, so we can get it in asap.
@mrzzzrm Done. I only had to fix the anti-join names manually.

Felerius:
@mrzzzrm Going back to what we discussed [here](https://github.com/hyrise/hyrise/pull/1473#discussion_r261889645): I think I found a case where we need to crawl subquery LQPs of predicate nodes we check:
```sql
SELECT * FROM a WHERE a.a IN (SELECT b.b FROM b WHERE EXISTS (SELECT c.c IN C WHERE c.c = a.a))
```
While we optimize the outer `IN`, we currently do not consider the `EXISTS` predicate node as using a correlated parameter, since we don't scan the innermost LQP and don't find the use of `a.a`.

mrks:
Shouldn't `LQPSubqueryExpression::is_correlated` be sufficient?

mrks:
Mmmh, you somehow killed TPC-H query 2 (and a couple others as well):

lqp

For some reason, there are cross joins emitted here.
@mrzzzrm - any idea why this passed all tests?

Felerius:
> Shouldn't `LQPSubqueryExpression::is_correlated` be sufficient?

I don't think so, since `is_correlated` only checks whether the subquery creates new correlated parameters, not whether it uses ones from an outer subquery.

Felerius:
> For some reason, there are cross joins emitted here.

This is an incompatibility between this rule and `JoinOrderingRule` (we wanted to open an issue about this a while ago but forgot about it unfortunately). In short, if `SubqueryToJoinRule` runs before `JoinOrderingRule`, the created semi-/anti-joins prevent the `JoinOrderingRule` from reordering joins and merging them with predicate nodes into inner joins (which is what happened in your example)

If it is the other way around, `JoinOrderingRule` places predicate nodes as low as possible, which might prevent `SubqueryToJoinRule` to pull them up into join predicates (for example if they are below a join node). TPC-H query 2 is an example where both orderings fail right now.

mrks:
> If it is the other way around, JoinOrderingRule places predicate nodes as low as possible, which might prevent SubqueryToJoinRule to pull them up into join predicates (for example if they are below a join node).

How does this prevent this rule from pulling the predicates up? It just needs to check if the pull-up is legal.

Felerius:
> > If it is the other way around, JoinOrderingRule places predicate nodes as low as possible, which might prevent SubqueryToJoinRule to pull them up into join predicates (for example if they are below a join node).
> 
> How does this prevent this rule from pulling the predicates up? It just needs to check if the pull-up is legal.

Depends on how hard that is to check. For inner joins it should be relatively easy (I think it might even be always possible for the kind of predicates that we pull up, but I'm not quite sure yet), however for outer joins I'm unsure about how nullability needs to be handled.

@mrzzzrm How does the join reordering decide which predicates are safe to move around? Is there any logic that we could apply inversely?

mrks:
On first thought, everything that is not null-producing should be fine to be moved up and down. E.g, the left side of a `LEFT JOIN` can be filtered both before and after, while pulling the right side up would lead to different results. For semi/anti joins, I would only look at the left side (the side that is forwarded).

@mrzzzrm is certainly more familiar with that though.

README.md
```diff
@@ -99,30 +99,35 @@ Contact: firstname.lastname@hpi.de
```
mrzzzrm: (Random line) @mrks @Felerius 
> @mrzzzrm Going back to what we discussed here: I think I found a case where we need to crawl subquery LQPs of predicate nodes we check:
`SELECT * FROM a WHERE a.a IN (SELECT b.b FROM b WHERE EXISTS (SELECT c.c IN C WHERE c.c = a.a))`

I don't think you should need to look at `SELECT c.c IN C WHERE c.c = a.a`. The PredicateNode processing `EXISTS (SELECT c.c IN C WHERE c.c = a.a))` will have a parameter `a.a` in its `predicate`. There should never be a need for you to go into `SELECT c.c IN C WHERE c.c = a.a`, however deeply nested of a subquery it might be.

> For some reason, there are cross joins emitted here.
@mrzzzrm - any idea why this passed all tests?

Because... there are just no tests for the JoinOrderingRule pulling predicates past Semi/Anti/Outer joins.

> Depends on how hard that is to check. For inner joins it should be relatively easy (I think it might even be always possible for the kind of predicates that we pull up, but I'm not quite sure yet), however for outer joins I'm unsure about how nullability needs to be handled.

You can always pull up predicates past Inner Joins, from the outer side of a left/right outer join and from the left side of a semi/anti join.
Both rules could be blamed here: The JoinOrderingRule (more specifically: the JoinGraph(Builder)) should be able to "look" past Semi/Anti joins, the SubqueryToJoinRule should be able to look past inner joins. For the moment, I think the SubqueryToJoinRule should be run before the JoinOrderingRule and should be fixed to be able to pull up predicates past inner joins, which is really easy (at least easier than what we'd need to do to JoinGraph and its builder in order to make it work with non-inner joins)

mrzzzrm:
Merged the master in in order to make the diff clean

mrzzzrm:
Are you okay with me committing on your branch? I'd rather do that then post bazillions of comments like "I don't like that variable name", "that test is redundant, imo".

Felerius:
> Are you okay with me committing on your branch?

No problem, go ahead.

> For the moment, I think the SubqueryToJoinRule should be run before the JoinOrderingRule and should be fixed to be able to pull up predicates past inner joins, which is really easy (at least easier than what we'd need to do to JoinGraph and its builder in order to make it work with non-inner joins)

This sounds like you want to run the SubqueryToJoinRule *after* the JoinOrderingRule (the one running later is the one that needs fixes). I agree though, this shouldn't be hard to implement in this rule.

> The PredicateNode processing `EXISTS (SELECT c.c IN C WHERE c.c = a.a))` will have a parameter `a.a` in its predicate.

When using this [playground.cpp](https://gist.github.com/Felerius/47fab10f21b30aa5618d8cb6bd21b5eb) I don't find one. What am I missing here?

Felerius:
@mrzzzrm @mrks @Bouncner: Short update from our side: I'm working on implementing the pulling past joins we discussed above, which should be done by tomorrow. We also want to write tests for our helper methods as suggested, including changing some of the `NoRewrite...` tests into tests for the relevant helper method. The bulk of that should be done tomorrow or Sunday.

Bouncner:
Sounds superb! 👍

Felerius:
Pulling predicates past joins is implemented now. I included support for all join modes since the only thing needed is a few extra lines in a switch.

On another note: I had to replace a `std::map` with a `std::vector` again, this time to keep the order of join predicates reproducible. Is there any reason that expressions/LQP nodes cannot have an `operator <` (except that nobody has needed it enough to implemented it)? I can't imagine this is the only place where this would be useful.

mrzzzrm:
No real reason, except that's like 30 classes needing 30 new functions.

src/test/optimizer/strategy/subquery_to_join_rule_test.cpp
```diff
@@ -0,0 +1,1048 @@
...
+// HELPER FUNCTION TESTS
+
+TEST_F(SubqueryToJoinRuleTest, AssessCorrelatedParameterUsageCountsNodesNotUsages) {
+  const auto parameter1 = correlated_parameter_(ParameterID{0}, a_a);
+  const auto parameter2 = correlated_parameter_(ParameterID{1}, a_b);
+  const std::map<ParameterID, std::shared_ptr<AbstractExpression>> parameter_map = {{ParameterID{0}, a_a_expression},
+                                                                                    {ParameterID{1}, a_b_expression}};
+
+  // clang-format off
+  const auto lqp =
```
mrzzzrm: @Felerius don't know how WIP this is, so sorry if you wanted to change this anyway: please format LQPs in the tests with an indentation of two.

resources/test_data/sqlite_testrunner.tables
```diff
@@ -1,5 +1,6 @@
 resources/test_data/tbl/sqlite/mixed_types_100.tbl mixed
```
mrzzzrm: (Random Line) @Felerius 

> When using this playground.cpp I don't find one. What am I missing here?

You are right. Since this is a nested subquery it will not have the parameter of the outer query in its arguments, and rightly so. We need to recursively craw subquery-LQPs....

mrzzzrm: Done

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,543 @@
...
+    case LQPNodeType::Sort:
+    case LQPNodeType::Validate:
+      return {true, false};
+    default:
+      return {false, false};
+  }
+  Fail("GCC thinks this is reachable");
+}
+
+void find_pullable_predicate_nodes_recursive(
```
mrzzzrm: This will currently find PredicateNodes below diamond shapes multiple times...

src/test/optimizer/strategy/subquery_to_join_rule_test.cpp
Felerius: They must have gotten accidentally auto-formatted at one point. Fixed now, thanks!

Bouncner:
That's the current result:
image

Q4 does not use a semi-join any longer: 
image

For Q22, I'd guess the exists has been reformulated before as well but it is now no longer.

Is this still WIP and expected at the current stage?

Felerius:
> Is this still WIP and expected at the current stage?

I think I know whats happening here: When we moved the `SubqueryToJoinRule` to run after the `JoinOrderingRule`, we also moved it to run after the `InsertLimitInExistsRule`, and limit nodes stop us from pulling up predicates. Could we move `InsertLimitInExistsRule` to just after `SubqueryToJoinRule`, or is there anything requiring it to be run so early?

Bouncner:
> I think I know whats happening here: When we moved the `SubqueryToJoinRule` to run after the `JoinOrderingRule`, we also moved it to run after the `InsertLimitInExistsRule`, and limit nodes stop us from pulling up predicates. Could we move `InsertLimitInExistsRule` to just after `SubqueryToJoinRule`, or is there anything requiring it to be run so early?

@FabianWiebe & @mrks ?

mrks:
Mmmmh... it won't work, because it won't find the subquery BUT that's a good thing because we can't use that limit once we have a join, anyway. In fact, are there any EXISTS left where the subquery cannot be rewritten into a join and where that rule makes sense?

That also raises a different point. How do you currently handle `WHERE EXISTS (SELECT * FROM t WHERE t.a = outer.a LIMIT 1)`?

FabianWiebe:
> Could we move `InsertLimitInExistsRule` to just after `SubqueryToJoinRule`, or is there anything requiring it to be run so early?

That would solve this. Change it. The `InsertLimitInExistsRule` rule should just run after the `ExistsReformulationRule`/`SubqueryToJoinRule` rule.

FabianWiebe:
> In fact, are there any EXISTS left where the subquery cannot be rewritten into a join and where that rule makes sense?

I assume that there is a limit on how many correlated parameters can be within a subquery.
So I would not expect this query to be rewritten by the `SubqueryToJoinRule` rule:
`SELECT * FROM T1 WHERE EXISTS
(SELECT * FROM T2 WHERE T1.a = T2.a AND T1.b = T2.b AND T1.c = T2.c)`

@Felerius Is that correct?

Bouncner:
> That also raises a different point. How do you currently handle `WHERE EXISTS (SELECT * FROM t WHERE t.a = outer.a LIMIT 1)`?

Do I understand our current state of Hyrise correct that a limit does not really provide any advantages in this case (assuming query is not JIT'ted) and consequently we can also reformulate such subqueries and drop the `limit`?

FabianWiebe:
> > That also raises a different point. How do you currently handle `WHERE EXISTS (SELECT * FROM t WHERE t.a = outer.a LIMIT 1)`?
> 
> Do I understand our current state of Hyrise correct that a limit does not really provide any advantages in this case (assuming query is not JIT'ted) and consequently we can also reformulate such subqueries and drop the `limit`?

Yes, that is true. (Except the limit is zero)

mrks:
> > That also raises a different point. How do you currently handle `WHERE EXISTS (SELECT * FROM t WHERE t.a = outer.a LIMIT 1)`?
> 
> Do I understand our current state of Hyrise correct that a limit does not really provide any advantages in this case (assuming query is not JIT'ted) and consequently we can also reformulate such subqueries and drop the `limit`?

The LIMIT does not have any benefit here, but if we simply keep it in the rewritten LQP that is the right (left? well, the hashed, #1557) input of the join, we end up with wrong results.

Felerius:
> > In fact, are there any EXISTS left where the subquery cannot be rewritten into a join and where that rule makes sense?
> 
> I assume that there is a limit on how many correlated parameters can be within a subquery.
> So I would not expect this query to be rewritten by the `SubqueryToJoinRule` rule:
> `SELECT * FROM T1 WHERE EXISTS (SELECT * FROM T2 WHERE T1.a = T2.a AND T1.b = T2.b AND T1.c = T2.c)`
> 
> @Felerius Is that correct?

Right now we reformulate regardless of the number of correlated parameters, so that `EXISTS` would be turned into a semi join with 3 predicates.

mrks:
Could you give an example of an EXISTS subquery that can not be reformulated?

Felerius:
> Could you give an example of an EXISTS subquery that can not be reformulated?

One example would be a subquery that contains a correlated predicate below a full outer join:
```sql
SELECT * FROM a WHERE EXISTS (SELECT * FROM b FULL OUTER JOIN (SELECT c.a FROM c WHERE c.a < a.a) AS s ON (b.a = s.a))
```

The LQP is probably far easier to understand:
unoptlqp

mrks:
Ok, so we still want that rule, but want to execute it after any rewriting.

My LIMIT question from above remains nevertheless.

Felerius:
> That also raises a different point. How do you currently handle `WHERE EXISTS (SELECT * FROM t WHERE t.a = outer.a LIMIT 1)`?

Right now we don't reformulate queries like this because we don't pull predicate nodes past limit nodes. We could probably remove (non-zero) limits in some cases when reformulating an exists, but I'm not sure when it would be safe. One example where we could not remove them would be below a join, because we might go from no join matches to some.

mrks:
Ok, just wanted to make sure.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -0,0 +1,537 @@
...
+  if (predicate_node->predicate()->type != ExpressionType::Predicate) {
+    return nullptr;
+  }
+
+  const auto& predicate_expression = std::static_pointer_cast<AbstractPredicateExpression>(predicate_node->predicate());
+
+  // We rely on PredicateSplitUpRule having split up ANDed chains of such predicates previously, so that we can process
+  // them separately.
+  auto predicate_condition = predicate_expression->predicate_condition;
+  if (!is_binary_predicate_condition(predicate_condition)) {
```
Felerius: @mrzzzrm This matches more predicate conditions than we want, since it includes `(NOT) LIKE` and `(NOT) IN`. It also causes the sanitizer errors on CI, since we cast an `InExpression` to a `BinaryPredicateExpression`, which it does not derive from

mrzzzrm: Mmpf, we need a function that does what I thought `is_binary_predicate_condition` would do. Will fix this later, sorry for breaking things...

Bouncner:
With the following change, everything appears to be just fine.
```@@ -92,8 +92,6 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {

   optimizer->add_rule(std::make_unique<ColumnPruningRule>());

-  optimizer->add_rule(std::make_unique<InsertLimitInExistsRule>());
-
   optimizer->add_rule(std::make_unique<ChunkPruningRule>());

   // Run before SubqueryToJoinRule, since the Semi/Anti Joins it introduces are opaque to the JoinOrderingRule
@@ -102,6 +100,8 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {
   // Relies on being run after PredicateSplitUpRule to avoid handling ANDed expressions.
   optimizer->add_rule(std::make_unique<SubqueryToJoinRule>());

+  optimizer->add_rule(std::make_unique<InsertLimitInExistsRule>());
+
   optimizer->add_rule(std::make_unique<LikeReplacementRule>());

   // Position the predicates after the JoinOrderingRule ran. The JOR manipulates predicate placement as well, but
```

image

I don't want to change the order of rules myself and push into your repository though.

mrks:
Sounds perfectly reasonable. If we can get rid of the EXISTS altogether, we don't need to insert a Limit into the now non-existing subquery.

mrks:
@Bouncner: Could you please update the results in the PR description with a text version of this? The PR description still contains the improvements for Q4.

Bouncner:
Will do as soon as the PR fixed the rule order.

mrks:
Could you give a quick summary on where we are with this? What's still coming/missing?

janetzki:
I'm currently adding more complex test cases, like the TPC-H 20.

Felerius:
> Could you give a quick summary on where we are with this? What's still coming/missing?

I'm working on making the pull-up phase safe for diamond LQPs, which should be done later today. Oh, and I also need to fix `is_binary_predicate_condition` thing, but that shouldn't take long.

mrzzzrm:
> Oh, and I also need to fix is_binary_predicate_condition thing, but that shouldn't take long

I fixed that, just pushed to the wrong remote :see_no_evil:

Felerius:
> I'm working on making the pull-up phase safe for diamond LQPs, which should be done later today. 

Done.

> Will do as soon as the PR fixed the rule order.

@Bouncner: I changed the order in f53177b

mrzzzrm:
Think I found a bug in there, if the subquery has a predicate `a=parameter AND b IN (1,2,3)`, `b IN (1,2,3)` gets dropped.

Felerius:
> Think I found a bug in there, if the subquery has a predicate `a=parameter AND b IN (1,2,3)`, `b IN (1,2,3)` gets dropped.

You're right, I didn't consider that. I will fix this later today, should be relatively easy.

Felerius:
> > Think I found a bug in there, if the subquery has a predicate `a=parameter AND b IN (1,2,3)`, `b IN (1,2,3)` gets dropped.
> 
> You're right, I didn't consider that. I will fix this later today, should be relatively easy.

Done. @mrzzzrm I fixed what looked like a copy-paste error in the test you added. Could you have a look just in case?

src/test/optimizer/strategy/subquery_to_join_rule_test.cpp
```diff
@@ -1116,12 +1131,12 @@ TEST_F(SubqueryToJoinRuleTest, SubqueryUsesConjunctionOfCorrelatedAndLocalPredic
   PredicateNode::make(greater_than_(d_a, subquery),
     node_d);
 
-  const auto join_predicates = expression_vector(equals_(d_c, e_c), equals_(d_b, e_b));
+  const auto join_predicates = expression_vector(equals_(d_b, e_b), greater_than_(d_a, sum_(e_a)));
```
mrzzzrm: yes, makes sense

mrks:
Sitting with @mrzzzrm here. Just checking - can I press the button?

Felerius:
> Sitting with @mrzzzrm here. Just checking - can I press the button?

I think so, we're done with everything we wanted to do.

mrks:
I updated the description with a current PR. Could you remove the three WIPs if they are done? The cost model did not make it in, correct? Not asking for it to be added, just wanting to finalize the description.

Felerius:
> I updated the description with a current benchmark. Could you remove the three WIPs if they are done? The cost model did not make it in, correct? Not asking for it to be added, just wanting to finalize the description.

The WIPs are all done and can be removed (I can't edit the description unfortunately). The cost model wasn't implemented because our benchmarks showed that the reformulation was always beneficial.

mrks:
:)

mrzzzrm:
@mrks are the perf numbers in the description from the benchmark you ran today?

mrks:
Yes

Am 8. April 2019 18:35:01 schrieb Moritz Eyssen <notifications@github.com>:

PR #1474, opened 2019-02-06
[DYOD18/19] MvccDeletePlugin

julianmenzler:
This pull request includes the implementation of our `MvccDeletePlugin`, some tests and the required changes within the hyrise codebase (compare Pull Requests #1462 #1468).

As @mrks suggested down in the comments, review PR #1462 and #1468 first. As soon as they are merged you can start reviewing this PR.

### DIFF
--- a/src/lib/concurrency/transaction_context.cpp
+++ b/src/lib/concurrency/transaction_context.cpp
@@ -39,6 +39,12 @@ TransactionContext::~TransactionContext() {
                 return !has_registered_operators || committed_or_rolled_back;
               }()),
               "Has registered operators but has neither been committed nor rolled back.");
+
+  /**
+     * Tell the TransactionManager, which keeps track of active snapshot-commit-ids,
+     * that this transaction has finished.
+     */
+  TransactionManager::get().deregister_transaction(_snapshot_commit_id);
 }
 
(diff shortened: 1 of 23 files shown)


src/lib/concurrency/transaction_context.cpp
```diff
@@ -39,6 +39,12 @@ TransactionContext::~TransactionContext() {
                 return !has_registered_operators || committed_or_rolled_back;
               }()),
               "Has registered operators but has neither been committed nor rolled back.");
+
+  /**
+     * Tell the TransactionManager, which keeps track of active snapshot-commit-ids,
+     * that this transaction has finished.
+     */
+  TransactionManager::get().deregister_transaction(_snapshot_commit_id);
```
mrks: This looks familiar. If this PR doesn't make sense before the other two are merged, let's wait for that. Otherwise, reviewers will have to look at the same things twice and consider changes that might be made in the other two PRs.

julianmenzler: Makes sense. I added a note at the top to clarify this circumstance.

mrks:
I'll close this for now so that the list only has items that someone can act on. I appreciate the PR being on time though!

PR #1477, opened 2019-02-06
Fix ColumnPruningRule for Insert and Delete

### LINKED ISSUE #1440: ColumnPruningRule partially broken
This looks wrong:

https://github.com/hyrise/hyrise/blob/e9036384dd30c462637e07ca4aa80d6d615edf39/src/lib/optimizer/strategy/column_pruning_rule.cpp#L76

It should try a cast to `InsertNode`, not `UpdateNode`. Also, there should be a test for this.

mrzzzrm:
Fix #1440

There was already a ColumnPruningRule test for Insert, but it did not fail because, by coincidence, another bug fixed things: The Delete/Insert/UpdateNodes were assumed to output all their input columns, causing the ColumnPruningRule to not prune in LQPs with Insert/Update/DeleteNodes. Actually, the Delete/Insert/Update **operators** return a nullptr, which is now reflected in the LQP as well.

### DIFF
--- a/src/lib/logical_query_plan/delete_node.cpp
+++ b/src/lib/logical_query_plan/delete_node.cpp
@@ -22,8 +22,8 @@ std::string DeleteNode::description() const {
 bool DeleteNode::is_column_nullable(const ColumnID column_id) const { Fail("Delete does not output any columns"); }
 
 const std::vector<std::shared_ptr<AbstractExpression>>& DeleteNode::column_expressions() const {
-  static std::vector<std::shared_ptr<AbstractExpression>> dummy_column_expressions;
-  return dummy_column_expressions;
+  static std::vector<std::shared_ptr<AbstractExpression>> empty_vector;
+  return empty_vector;
 }
 
 std::shared_ptr<AbstractLQPNode> DeleteNode::_on_shallow_copy(LQPNodeMapping& node_mapping) const {
--- a/src/lib/logical_query_plan/delete_node.cpp
+++ b/src/lib/logical_query_plan/delete_node.cpp
@@ -21,6 +21,10 @@ std::string DeleteNode::description() const {
 
 bool DeleteNode::is_column_nullable(const ColumnID column_id) const { Fail("Delete does not output any columns"); }
 
+const std::vector<std::shared_ptr<AbstractExpression>>& DeleteNode::column_expressions() const {
+  return _dummy_column_expressions;
+}
+
 std::shared_ptr<AbstractLQPNode> DeleteNode::_on_shallow_copy(LQPNodeMapping& node_mapping) const {
   return DeleteNode::make();
 }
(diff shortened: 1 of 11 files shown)


src/lib/logical_query_plan/delete_node.cpp
```diff
@@ -21,6 +21,10 @@ std::string DeleteNode::description() const {
 
 bool DeleteNode::is_column_nullable(const ColumnID column_id) const { Fail("Delete does not output any columns"); }
 
+const std::vector<std::shared_ptr<AbstractExpression>>& DeleteNode::column_expressions() const {
+  return _dummy_column_expressions;
```
mrks:
```
  static auto empty_vector = std::vector<std::shared_ptr<AbstractExpression>>{};
  return empty_vector;
```
?

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -51,31 +51,56 @@ ExpressionUnorderedSet ColumnPruningRule::_collect_actually_used_columns(const s
...
+      // prune) by only searching the arguments of expression.
+      case LQPNodeType::Projection: {
+        for (const auto& expression : node->node_expressions) {
+          for (const auto& argument : expression->arguments) {
+            collect_consumed_columns_from_expression(argument);
+          }
+        }
+      } break;
+
+        // No pruning of the input columns to Delete, Update and Insert, they need them all.
```
mrks: format

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -51,31 +51,56 @@ ExpressionUnorderedSet ColumnPruningRule::_collect_actually_used_columns(const s
...
   // the "functioning" of the LQP.
-  // For ProjectionNodes, ignore forwarded columns (since they would include all columns and we wouldn't be able to
-  // prune) by only searching the arguments of expression.
   visit_lqp(lqp, [&](const auto& node) {
-    for (const auto& expression : node->node_expressions) {
-      if (node->type == LQPNodeType::Projection) {
-        for (const auto& argument : expression->arguments) {
-          collect_consumed_columns_from_expression(argument);
+    switch (node->type) {
+      case LQPNodeType::Aggregate:
```
mrks: Maybe add a comment that this is the "default" case?

src/lib/logical_query_plan/delete_node.cpp
```diff
@@ -22,7 +22,8 @@ std::string DeleteNode::description() const {
 bool DeleteNode::is_column_nullable(const ColumnID column_id) const { Fail("Delete does not output any columns"); }
 
 const std::vector<std::shared_ptr<AbstractExpression>>& DeleteNode::column_expressions() const {
-  return _dummy_column_expressions;
+  static std::vector<std::shared_ptr<AbstractExpression>> dummy_column_expressions;
+  return dummy_column_expressions;
```
mrks: When I read "dummy" here, I thought it had something to do with the dummy node. Now that it is in the same method, it's not that much of a problem, but in my opinion `empty_vector` would express better that we would like to `return {}` but can't. Won't mind if you disagree and simply merge though.

PR #1479, opened 2019-02-07
Extend compare_benchmarks.py script with --ignore-naming parameter

Miroka96:
Regarding comparison of benchmark results from FileBasedBenchmarks

Previously it was only possible to compare results of the same benchmark. This failed in our setup where we rewrote SQL queries and stored them in a second file. When running the benchmarks, this caused a different name, although the queries were doing the same in two different manners.

To be able to compare them anyway using scripts/compare_benchmarks.py just add the parameter --ignore-naming

### DIFF
--- a/scripts/compare_benchmarks.py
+++ b/scripts/compare_benchmarks.py
@@ -35,15 +35,34 @@ def calculate_and_format_p_value(old, new):
 	color = 'green' if is_significant else 'white'
 	return colored(notes + "{0:.4f}".format(p_value), color)
 
-
-if(len(sys.argv) != 3):
-	print("Usage: " + sys.argv[0] + " benchmark1.json benchmark2.json")
+def print_usage():
+	print("Usage: " + sys.argv[0] + " [--ignore-naming]" + " benchmark1.json benchmark2.json")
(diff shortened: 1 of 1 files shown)


Bouncner:
Please add a description.

Miroka96:
How can I retrigger Jenkins?

Bouncner:
`git commit --allow-empty -m 'Trigger CI'`

Bensk1:
Has anyone else ever encountered this problem? Do we really need it?

mrks:
Ok, the explanation in the description makes sense. Instead of adding a parameter that we are parsing by hand and that is semi-documented, why don't we simply print both query names (maybe in red?) in cases where the names don't match?

Miroka96:
Sorry for my delay in answering...
Do you mean, just set the parameter by default and do not fail, if the names differ? This would be a solution

mrks:
Almost - I meant not having a parameter/flag/variable at all and simply print a warning.

mrks:
I'm trying to reduce the list of open PRs, so I took the liberty of making that change myself.

PR #1480, opened 2019-02-07
[DYOD18/19] Like Replacement Rule

simsieg:
This PR introduces the like replacement optimizer rule. The rule takes a LQP, looks for string patterns of the type ` a LIKE "abc%"` and replaces them with two predicates ` a >= "abc" AND a < "abd"`.
 This PR adds the between composition rule to the Hyrise optimizer, which allows the reformulation of binary predicate expressions in a specific pattern (a >= 0 AND a <= 10) to between expressions (a BETWEEN 0 AND 10)

# Steps

- Identify `LIKE` scans of of the pattern `"abc%"`
- Remove those nodes from the LQP
- Create predicate node for first string without wildcard (`"abc"`) and string with increased last character before the wildcard (`"abd"`)
- Create predicate expressions in the way `a >= "abc" AND a <= "abd"`
- Insert the predicate nodes into the LQP

# Benchmarks

Results of a microbenchmark on a generated lineitem table (scale factor: 1):

| Standard (Iter/s) | Optimized (Iter/s) | Selectivity (%) |
|-------------------|--------------------|-----------------|
| 13                | 128                | 2,19            |
| 13                | 129                | 3,25            |
| 16                | 100                | 8,72            |
| 16                | 162                | 11,97           |

# Limitations

- The algorithm can only be applied to ascii character sets


mrks:
Please ping your partner group and ask them to start reviewing, if you have not already done so.

src/test/optimizer/strategy/like_replacement_rule_test.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+  std::shared_ptr<StoredTableNode> node;
+  LQPColumnReference a;
+  std::shared_ptr<LikeReplacementRule> _rule;
+};
+
+TEST_F(LikeReplacementTest, LikeReplacement) {
+  const auto input_lqp = PredicateNode::make(like_(a, "RED%"), node);
+
+  // clang-format off
```
querenker: I don't know how clang-format would format this code snippet, but is it so bad, that you have to disable the formatter?

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,73 @@
...
+            upper_bound_node->set_left_input(lower_bound_node);
+            for (size_t output_idx = 0; output_idx < outputs.size(); ++output_idx) {
+              outputs[output_idx]->set_input(input_sides[output_idx], upper_bound_node);
+            }
+          }
+        }
+      }
+    }
+  }
+  _apply_to_inputs(node);
```
aloeser: I think you might have an unintended abort condition here.
From my understanding, you want this rule to be applied to all `LIKE` nodes. However, I think it is applied only to the first `LIKE` node in the LQP.

Reasoning:
L. 26 declares `predicate_node` as type-casted version of  `node`, so they should both point to the same object.
L. 57 calls `lqp_remove_node(predicate_node)`, which sets the (left) input of that node to `nullptr`.
L. 70 calls `_apply_to_inputs(node)`, which applies the rule to all inputs. However, since `node` references the same object as `predicate_node`, it has no longer any inputs, and the recursion stops.
I must admit that I did not wrote a test to prove it, however I think it would be good to have a test that covers the double-like-edgecase.

To fix this, one could maybe pull this line up, or save the original input in a variable and apply the rule to it.

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,73 @@
...
+        const auto value = boost::get<std::string>(value_expression->value);
+        const auto offset = value.find_first_of("%");
+        // Check if the string has an % wildcard and ends with the % wildcard
+        if (offset != std::string::npos && offset > 0 && value.length() == offset + 1) {
+          // Calculate lower and upper bound of the string
+          const auto lower_bound = value.substr(0, offset);
+          const auto current_character_value = static_cast<int>(lower_bound.at(lower_bound.length() - 1));
+          // Find next value according to ASCII-table
+          if (current_character_value > 0 && current_character_value < 127) {
+            const auto next_character = static_cast<char>(current_character_value + 1);
```
aloeser: You stated in your description that it works for ASCII only. Maybe you could outsource this line to a function that "knows" about the charset and how to get the next character? It might not make much sense for ascii only, but e.g. if we consider supporting UTF-8 with its var-length characters, the next-char-function might get more complicated.
I feel like this could be easier to extend

src/lib/optimizer/optimizer.cpp
```diff
@@ -101,6 +102,8 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {
 
   optimizer->add_rule(std::make_shared<JoinOrderingRule>(std::make_shared<CostModelLogical>()));
 
+  optimizer->add_rule(std::make_shared<LikeReplacementRule>());
```
querenker: Is there a special reason why this rule will be applied after the `JoinOrderingRule`? I am not sure, if there is a preferred order (as long as it works). But it seems that in general, rules that replace nodes are executed before rules that change the order of nodes?

src/lib/optimizer/strategy/like_replacement_rule.cpp
aloeser: I do not know how Hyrise deals with collations (especially case-(in)sensitivity), so maybe the comment below is not yet relevant, and maybe will never be.

Anyway, consider the following scenario: we get a query containing `name LIKE "Z%"`.
Your rule will increment `Z` (90) to `[` (91). While this might look a bit weird, it should probably still work fine, as long as strings are compared charwise by their ascii code points (no idea if this is the case for all collations?).
Further, there could be a problem when using case-insensitive comparisons. "Could" because it depends on the concrete implementation, but assume we achieve case-insensitivity by e.g. lowercasing everything, and we do this when the LQP is done with optimization.
In that case, the `Z` (90) will become `z` (122), while `[` stays at 91. After your rule it would look like `name >= 122 AND name < 91`, which will never evaluate to true.

Again, I know there's a lot of ifs in there, so I am not sure if this really needs fixing. If it needs a fix, one idea could be to make the condition in L. 41 stricter, e.g. only apply the rule if the last char is in A-Y or a-y.

src/lib/optimizer/strategy/like_replacement_rule.cpp
simsieg: Thanks a lot!

src/lib/optimizer/strategy/like_replacement_rule.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include "abstract_rule.hpp"
+#include "expression/expression_functional.hpp"
+
+#include "types.hpp"
+
+namespace opossum {
+
+/**
+ * This rule determines special cases of like ("abc%") which can be rewritten into two BinaryPredicateExpressions (column >= "abc" and column < "abd")
```
mrks: **that** can be rewritten (restrictive clause)

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+using namespace opossum::expression_functional;  // NOLINT
+
+namespace opossum {
+
+std::string LikeReplacementRule::name() const { return "Like Replacement Rule"; }
+
+void LikeReplacementRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node) const {
+  const auto predicate_node = std::dynamic_pointer_cast<PredicateNode>(node);
+  // Continue only if predicte_node is defined
```
mrks: Typo. Also, this comment could be less syntactical and more semantical. What does it mean that predicate_node is not defined?

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,88 @@
...
+    return;
+  }
+  const auto offset = value.find_first_of("%");
+
+  // Continue only if the string has a % wildcard and ends with the % wildcard ("asdf%")
+  if (offset == std::string::npos || offset <= 0 || value.length() != offset + 1) {
+    _apply_to_inputs(node);
+    return;
+  }
+  // Calculate lower and upper bound of the string
```
mrks: A string has no "lower and upper bound". A search pattern has.

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,88 @@
...
+  if (current_character_value > 0 && current_character_value < 127) {
+    const auto next_character = static_cast<char>(current_character_value + 1);
+    const auto upper_bound = lower_bound.substr(0, offset - 1) + next_character;
+    const auto lower_bound_node = PredicateNode::make(
+        std::make_shared<BinaryPredicateExpression>(PredicateCondition::GreaterThanEquals, column_expression,
+                                                    std::make_shared<ValueExpression>(lower_bound)));
+    const auto upper_bound_node = PredicateNode::make(std::make_shared<BinaryPredicateExpression>(
+        PredicateCondition::LessThan, column_expression, std::make_shared<ValueExpression>(upper_bound)));
+
+    // Store the input and outputs of the node
```
mrks: Which node? You just added two new nodes that I have to keep track of.

src/test/optimizer/strategy/like_replacement_rule_test.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+TEST_F(LikeReplacementTest, MultipleLikes) {
+  const auto input_lqp = PredicateNode::make(like_(a, "RED%"), PredicateNode::make(like_(a, "BLUE%"),node));
+
+  const auto expected_lqp =
+      PredicateNode::make(less_than_(a, "REE"),
+      PredicateNode::make(greater_than_equals_(a, "RED"),
+      PredicateNode::make(less_than_(a, "BLUF"),
+      PredicateNode::make(greater_than_equals_(a, "BLUE"), node))));
```
mrks: Extract into variables or use indentation with `clang-format off` (see LQP tests in other files)

src/test/optimizer/strategy/like_replacement_rule_test.cpp
```diff
@@ -0,0 +1,124 @@
...
+  const auto result_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
+
+  EXPECT_LQP_EQ(result_lqp, expected_lqp);
+}
+
+TEST_F(LikeReplacementTest, MultipleLikes) {
+  const auto input_lqp = PredicateNode::make(like_(a, "RED%"), PredicateNode::make(like_(a, "BLUE%"), node));
+
+  // clang-format off
+  const auto expected_lqp =
```
mrzzzrm: Just to be clear what Markus meant was a formatting like:

```c++
const auto expected_lqp =
PredicateNode::make(less_than_(a, "REE"),
  PredicateNode::make(greater_than_equals_(a, "RED"),
      PredicateNode::make(less_than_(a, "BLUF"),
        PredicateNode::make(greater_than_equals_(a, "BLUE"), node))));
```

src/test/optimizer/strategy/like_replacement_rule_test.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+#include "utils/assert.hpp"
+
+#include "logical_query_plan/mock_node.hpp"
+
+using namespace opossum::expression_functional;  // NOLINT
+
+namespace opossum {
+
+class LikeReplacementTest : public StrategyBaseTest {
```
mrzzzrm: pls test that NOT LIKE remains unaffected

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+  const auto value = boost::get<std::string>(value_expression->value);
+  // Filter strings containing "_", as they cannot be reformulated using this approach
+  if (value.find_first_of("_") != std::string::npos) {
+    _apply_to_inputs(node);
+    return;
+  }
+  const auto offset = value.find_first_of("%");
+
+  // Continue only if the string has a % wildcard and ends with the % wildcard ("asdf%")
```
mrzzzrm: "... and has a non-empty prefix before the wildcard"

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,87 @@
...
+  if (offset == std::string::npos || offset <= 0 || value.length() != offset + 1) {
+    _apply_to_inputs(node);
+    return;
+  }
+  // Calculate lower and upper bound of the search pattern
+  const auto lower_bound = value.substr(0, offset);
+  const auto current_character_value = static_cast<int>(lower_bound.at(lower_bound.length() - 1));
+  // Find next value according to ASCII-table
+  if (current_character_value > 0 && current_character_value < 127) {
```
mrzzzrm: Why the "current_character_value > 0" check? you never substract anything, do you? Why would a lower bounds check be necessary?

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,87 @@
...
+    const auto next_character = static_cast<char>(current_character_value + 1);
+    const auto upper_bound = lower_bound.substr(0, offset - 1) + next_character;
+    const auto lower_bound_node = PredicateNode::make(std::make_shared<BinaryPredicateExpression>(
```
mrzzzrm: Please add a comment why you cannot just emit a BETWEEN here. (and a TODO that we soon might after #1481 )

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,87 @@
...
+        PredicateCondition::GreaterThanEquals, column_expression, std::make_shared<ValueExpression>(lower_bound)));
+    const auto upper_bound_node = PredicateNode::make(std::make_shared<BinaryPredicateExpression>(
+        PredicateCondition::LessThan, column_expression, std::make_shared<ValueExpression>(upper_bound)));
+
+    // Store the input and outputs of the like predicate node
```
mrzzzrm: Adding the new nodes below the old predicate node and then removing the old node would be a bit less code

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+using namespace opossum::expression_functional;  // NOLINT
+
+namespace opossum {
+
+std::string LikeReplacementRule::name() const { return "Like Replacement Rule"; }
+
+void LikeReplacementRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node) const {
+  const auto predicate_node = std::dynamic_pointer_cast<PredicateNode>(node);
+  // Continue only on predicate nodes
```
Bensk1: I don't know if we really need this (same applies to L33) comment, line 26 and 28 are pretty easy to understand and it should be self-explanatory. But I would not block this PR if you decide to leave it like this.

Bouncner:
Can you help to remember what the final decision on reformulating `col like 'a%'` to `col > 'a' and col < 'b'` was? The discussion what about collations and safe assumptions about the char set.

src/lib/optimizer/strategy/like_replacement_rule.cpp
simsieg: We decided not to rewrite like expressions to between expressions. There are two different rules.
Consider this query part: `column LIKE "a%" AND column <= "asdf"`
If we would reformulate the like to a between in this rule, the outcome would be:
`column BETWEEN "a" AND "b" (EX)  AND column <= "asdf"`
If we have two different rules applied in sequence, the outcome is:
`column BETWEEN "a" AND "asdf"`

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,87 @@
...
+    _apply_to_inputs(node);
+    return;
+  }
+  const auto expression = std::dynamic_pointer_cast<BinaryPredicateExpression>(predicate_node->predicate());
+  // Continue only if expressions predicate condition is of type PredicateCondition::Like
+  if (expression == nullptr || expression->predicate_condition != PredicateCondition::Like) {
+    _apply_to_inputs(node);
+    return;
+  }
+  const auto column_expression = std::dynamic_pointer_cast<LQPColumnExpression>(expression->left_operand());
```
mrzzzrm: the left operand being a column expression is by no means a necessity - please allow any expression type here.

src/lib/optimizer/strategy/like_replacement_rule.cpp
mrzzzrm: By "comment" I meant "in the code", sorry to be unclear. I am fine with not generating a BETWEEN for now, but in general: Rules should not know about each other. A BETWEEN scan is faster than two separate scans, so not emitting a *faster* predicate because another rule cannot handle it in an obscure case is a weak argument.

src/lib/optimizer/strategy/like_replacement_rule.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include "abstract_rule.hpp"
+#include "expression/expression_functional.hpp"
+
+#include "types.hpp"
+
+namespace opossum {
+
+/**
+ * This optimizer rule replaces PredicateNodes with a like condition by BinaryPredicateExpressions for some special cases.
```
mrzzzrm: "replaces ... with", not "replaces ... by", right?

src/test/optimizer/strategy/like_replacement_rule_test.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+  std::shared_ptr<StoredTableNode> node;
+  LQPColumnReference a;
+  std::shared_ptr<LikeReplacementRule> _rule;
+};
+
+TEST_F(LikeReplacementTest, LikeReplacement) {
+  const auto input_lqp = PredicateNode::make(like_(a, "RED%"), node);
+
+  const auto expected_lqp =
```
mrzzzrm: minor, but please format the LQP as you did below, one node per line

src/lib/optimizer/strategy/like_replacement_rule.hpp
Bensk1: I thought `replace` could be used with both prepositions, `with` and `by`.

mrzzzrm: just sounded weird to me, not about to look up english grammar rules.

src/lib/optimizer/strategy/like_replacement_rule.cpp
simsieg: @mrks @Bensk1 What is your opinion?

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,87 @@
...
+  // Continue only if the string has a % wildcard and ends with the % wildcard ("asdf%") and has a non-empty prefix before the wildcard
+  if (offset == std::string::npos || offset <= 0 || value.length() != offset + 1) {
+    _apply_to_inputs(node);
+    return;
+  }
+  // Calculate lower and upper bound of the search pattern
+  const auto lower_bound = value.substr(0, offset);
+  const auto current_character_value = static_cast<int>(lower_bound.at(lower_bound.length() - 1));
+  // Find next value according to ASCII-table
+  if (current_character_value < 127) {
```
Bensk1: Prof. Hirschfeld would probably suggest to replace the `127` by a constant `MAX_ASCII_VALUE`.

src/lib/optimizer/strategy/like_replacement_rule.cpp
Bensk1: Moritz is right. Ideally, the BETWEEN rule would also be able to incorporate other betweens when calculating the minimal necessary between range. I think both are fine for now since it should really be an edge case.

Either way, we should put a comment on that and/or open an issue for improvement to either incorporate existing betweens in the other rule or if you decide to keep it as is that instead a between could be emitted and betweens should be incorporated by the between rule.

src/lib/optimizer/strategy/like_replacement_rule.cpp
mrks: Mmh, not sure if I agree. While the transformation would still be valid, it might not be beneficial.

For now, I'd keep it simple and stick to the default case of having a column on the left.

src/lib/optimizer/strategy/like_replacement_rule.cpp
mrzzzrm: BETWEEN, apart from being faster to execute, can also be used to prune entire Chunks. Two separate scans are less likely to result in a pruned Chunk. I think we want BETWEEN here.

src/lib/optimizer/strategy/like_replacement_rule.cpp
mrzzzrm: For which expression `x` we want `x LIKE "a%"` over `x BETWEEN a AND b (EX)`?

src/lib/optimizer/strategy/like_replacement_rule.hpp
mrks: There used to be a page where you could see the most frequently used prepositions for a verb, but I can't find it anymore. This suggests that 'with' is more common: 
https://english.stackexchange.com/a/151352

src/lib/optimizer/strategy/like_replacement_rule.cpp
mrks: We could also do it the Java way and create a CharacterSetMaximumIntRepresentationCalculatorFactory. ;)

src/lib/optimizer/strategy/like_replacement_rule.hpp
Bensk1: In this case it could even be understood that `PredicateNodes` are replaced with `a like condition`. So maybe `PredicateNodes that contain a like condition`

src/lib/optimizer/strategy/like_replacement_rule.cpp
mrzzzrm: > MAX_ASCII_VALUE

that would actually help me

Bensk1: @mrks often likes to make fun of that stuff, but in this case a plain `127` (even with the comment) is not partly as descriptive as a speaking constant.

src/lib/optimizer/strategy/like_replacement_rule.cpp
mrks: Mmh, now that I think about it a bit longer, I can't come up with anything. Yes, I suppose we can simply get rid of the check for a ColumnExpression on the left.

src/lib/optimizer/strategy/like_replacement_rule.cpp
mrks: Looking at this again, I'd vote for BETWEEN, too. If we want to cover that edge case where we have another limiting predicate on the same column, that should be the job of the between rule, not this one.

src/lib/optimizer/strategy/like_replacement_rule.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include "storage/table.hpp"
+#include "utils/assert.hpp"
+
+using namespace opossum::expression_functional;  // NOLINT
+
+namespace opossum {
+
+std::string LikeReplacementRule::name() const { return "Like Replacement Rule"; }
+
+const MAX_ASCII_VALUE = 127;
```
mrks: Brrr. Implicit int is... not nice. Please add a data type (or auto). Also, constexpr?

mrzzzrm:
What's the state of this PR? Anything I can help with?

mrks:
A final review (or dismissal of your review) would be nice.

mrzzzrm:
Added two tests (DiamondShape + LastAsciiChar), ready to go in.

mrks:
Fire at will.

PR #1481, opened 2019-02-07
[DYOD18/19] Between Composition Rule

simsieg:
This PR adds the between composition rule to the Hyrise optimizer, which allows the reformulation of binary predicate expressions in a specific pattern (a >= 0 AND a <= 10) to between expressions (a BETWEEN 0 AND 10)

To represent single/double side exclusive binary predicate expression the Between Table Scan was extended, too. The BetweenExpression now supports two additional constructor parameters (left_inclusive and right_inclusive), which default to true to comply with the SQL standard.

# Steps

- Identify a contiguous row of predicate nodes
- Remove those nodes from the LQP
- Identify possible scan boundaries that result from the predicate nodes
- Match boundaries according to the rule pattern (a >= 0 AND a <= 10)
- Generate between nodes for every match
- Insert the between nodes and the remaining predicate nodes into the LQP

# Benchmarks

Results of a microbenchmark on a generated lineitem table (scale factor: 1):

| Standard (Iter/s) | Optimized (Iter/s) | Selectivity |
|-------------------|--------------------|-------------|
| 19                | 58                 | 5,879526063 |
| 18                | 54                 | 7,010163775 |
| 16                | 41                 | 13,86367594 |
| 13                | 29                 | 25,89510624 |
| 11                | 29                 | 27,80375307 |
| 10                | 19                 | 47,60347696 |
| 10                | 22                 | 68,37687035 |
| 12                | 27                 | 89,26302424 |
| 12                | 30                 | 95,23059914 |
| 12                | 32                 | 98,68333329 |
| 13                | 31                 | 100         |

# TPC-H
```
+-----------+----------------+------+----------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s   | runs | new iter/s     | runs | change | p-value (significant if <0.001) |
+-----------+----------------+------+----------------+------+--------+---------------------------------+
| TPC-H 1   | 0.524423062801 | 32   | 0.537236154079 | 33   | +2%    |                          0.0026 |
| TPC-H 2   | 8.20963001251  | 493  | 8.51211643219  | 511  | +4%    |                          0.0000 |
| TPC-H 3   | 6.56763982773  | 395  | 6.50631523132  | 391  | -1%    |                          0.0014 |
| TPC-H 4   | 1.86995303631  | 113  | 1.92370796204  | 116  | +3%    |                          0.0000 |
| TPC-H 5   | 4.51079559326  | 271  | 4.26213264465  | 256  | -6%    |                          0.0000 |
| TPC-H 6   | 26.2078990936  | 1573 | 29.8916110992  | 1794 | +14%   |                          0.0000 |
| TPC-H 7   | 1.322229743    | 80   | 1.26839864254  | 77   | -4%    |                          0.0024 |
| TPC-H 8   | 5.44385528564  | 327  | 5.70827865601  | 343  | +5%    |                          0.0000 |
| TPC-H 9   | 1.49139273167  | 90   | 1.58014404774  | 95   | +6%    |                          0.0000 |
| TPC-H 10  | 2.67921328545  | 161  | 2.75179862976  | 166  | +3%    |                          0.0000 |
| TPC-H 11  | 32.6247787476  | 1958 | 32.7540435791  | 1966 | +0%    |                          0.0047 |
| TPC-H 12  | 7.02723455429  | 422  | 8.33574676514  | 501  | +19%   |                          0.0000 |
| TPC-H 13  | 2.22448968887  | 134  | 2.21665239334  | 133  | -0%    |                          0.6583 |
| TPC-H 14  | 17.198759079   | 1032 | 24.7140789032  | 1483 | +44%   |                          0.0000 |
| TPC-H 15  | 10.677573204   | 641  | 19.9230823517  | 1196 | +87%   |                          0.0000 |
| TPC-H 16  | 7.7480340004   | 465  | 7.80457401276  | 469  | +1%    |                          0.0004 |
| TPC-H 17  | 1.41240048409  | 85   | 1.449639678    | 87   | +3%    |                          0.0000 |
| TPC-H 18  | 2.01038455963  | 121  | 2.01796936989  | 122  | +0%    |                          0.6082 |
| TPC-H 19  | 5.2063832283   | 313  | 5.2030134201   | 313  | -0%    |                          0.8478 |
| TPC-H 20  | 1.96740233898  | 119  | 2.02884721756  | 122  | +3%    |                          0.0000 |
| TPC-H 21  | 0.419977486134 | 26   | 0.41627201438  | 25   | -1%    |                          0.1030 |
| TPC-H 22  | 9.95100402832  | 598  | 9.99539279938  | 600  | +0%    |                          0.0014 |
| average   |                |      |                |      | +8%    |                                 |
+-----------+----------------+------+----------------+------+--------+---------------------------------+

```

# Limitations

- The algorithm can only be applied to a contiguous row of predicate nodes


src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  // Apply boundary check again for the last column
+  if (lower_bound_value_expression != nullptr && upper_bound_value_expression != nullptr) {
+    const auto between_node = PredicateNode::make(
+        std::make_shared<BetweenExpression>(current_column_expression, lower_bound_value_expression,
+                                            upper_bound_value_expression, left_inclusive, right_inclusive));
+    between_nodes.push_back(between_node);
+  } else {
+    predicate_nodes.insert(std::end(predicate_nodes), std::cbegin(node_scope), std::cend(node_scope));
```
aloeser: From our understanding, `node_scope` contains all nodes that were at some point an upper or lower bound for an between expression.
It does make sense to track them, as we might not find a corresponding lower/upper bound and need to re-insert them. However, don't we have to remove them from `node_scope` as soon as we have both a lower and an upper bound? From our understanding, this should lead to additional nodes in the LQP, causing all tests for this rule to fail.
We could not build the tests (we are getting the same compiler error as the [ci-server](https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise/hyrise/detail/PR-1481/1/pipeline)), so we are not perfectly sure about it.
Maybe there is there something we missed or misunderstood?

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+void BetweenCompositionRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node) const {
+  if (node->type == LQPNodeType::Predicate) {
+    std::vector<std::shared_ptr<AbstractLQPNode>> predicate_nodes;
+
+    // Gather adjacent PredicateNodes
+    auto current_node = node;
+    while (current_node->type == LQPNodeType::Predicate) {
+      // Once a node has multiple outputs, we're not talking about a predicate chain anymore
+      if (current_node->outputs().size() > 1 || current_node->right_input() != nullptr) {
```
aloeser: Just curious, but what would be examples for predicates with more than one output and/or more than one input?

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,219 @@
...
+  }
+}
+
+void BetweenCompositionRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node) const {
+  if (node->type == LQPNodeType::Predicate) {
+    std::vector<std::shared_ptr<AbstractLQPNode>> predicate_nodes;
+
+    // Gather adjacent PredicateNodes
+    auto current_node = node;
+    while (current_node->type == LQPNodeType::Predicate) {
```
aloeser: Probably out of scope, just wondering if the following scenario can happen:
Consider the following sequence of LQP nodes (Predicate `A > 10`), (Projection `A`),(Predicate `A < 20`).
If we get it right, such a case would not be optimized by your rule, allthough its possible in this specific case. Do you know by chance if the given sequence is actually valid, i.e. can it appear in that order or are there some optimizer/sql-parser-intern assumptions that ensure predicate chains are as long as possible? (Not saying you have to handle this)

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -0,0 +1,258 @@
...
+    _rule = std::make_shared<BetweenCompositionRule>();
+    node = StoredTableNode::make("a");
+
+    a = LQPColumnReference{node, ColumnID{0}};
+    b = LQPColumnReference{node, ColumnID{1}};
+    c = LQPColumnReference{node, ColumnID{2}};
+  }
+
+  std::shared_ptr<StoredTableNode> node;
+  LQPColumnReference a, b, c;
```
ekrebs5: In l. 46 and 47 an "_" is probably missing before the variable names. Also, according to the [style guide](https://github.com/hyrise/hyrise/blob/master/CONTRIBUTING.md), the names could be more descriptive.

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+  const auto expected_lqp = input_lqp;
+  // clang-format on
+
+  const auto result_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
+
+  EXPECT_LQP_EQ(result_lqp, expected_lqp);
+}
+
+TEST_F(BetweenCompositionTest, ImpossibleColumnRange) {
```
ekrebs5: We are not what the expected behaviour is here, should an impossible range still lead to a between node?

src/test/base_test.hpp
```diff
@@ -112,15 +112,28 @@ class BaseTestWithParam
...
+      return create_between_table_scan(in, column_id, value, value2);
     } else {
       predicate = std::make_shared<BinaryPredicateExpression>(predicate_condition, column_expression, value_(value));
     }
 
     return std::make_shared<TableScan>(in, predicate);
   }
 
+  // Utility to create between table scans
+  static std::shared_ptr<TableScan> create_between_table_scan(
```
ekrebs5: What is the advantage of making this static?
If it just shouldn't be used from outside the class, wouldn't a protected function work?
(We aren't sure how static function work in combination with subclasses)

src/test/operators/table_scan_between_test.cpp
```diff
@@ -26,7 +26,7 @@ class TableScanBetweenTest : public TypedOperatorBaseTest {
     // ...
     // 30.25         10
     //
-    // As the first column is type casted, it contains 10 for an int column, the string "10.25" for a string column etc.
+    // As the first column is TYPE CASTED, it contains 10 for an int column, the string "10.25" for a string column etc.
```
ekrebs5: Why did you capitalize this?

src/test/operators/table_scan_between_test.cpp
```diff
@@ -57,50 +57,91 @@ class TableScanBetweenTest : public TypedOperatorBaseTest {
     _data_table_wrapper = std::make_shared<TableWrapper>(data_table);
     _data_table_wrapper->execute();
   }
+
+  void _test_between_boundaries(std::vector<std::tuple<AllTypeVariant, AllTypeVariant, std::vector<int>>>& tests,
+                                bool left_inclusive = true, bool right_inclusive = true) {
+    const auto& [data_type, encoding, nullable] = GetParam();
+    std::ignore = encoding;
```
ekrebs5: If we understand correctly, this means the variable is not used.
Using __maybe_unused may be better for this?

src/test/operators/table_scan_between_test.cpp
```diff
@@ -57,50 +57,91 @@ class TableScanBetweenTest : public TypedOperatorBaseTest {
...
+    resolve_data_type(data_type, [&, nullable = nullable](const auto type) {
```
ekrebs5: We are not sure what the ``nullable = nullable`` does. We think you could use this to rename the variable, but here everything stays the same?

src/test/operators/table_scan_between_test.cpp
```diff
@@ -57,50 +57,91 @@ class TableScanBetweenTest : public TypedOperatorBaseTest {
...
+      for (const auto& [left, right, expected_with_null] : tests) {
+        SCOPED_TRACE(std::string("BETWEEN ") + std::to_string(boost::get<double>(left)) +
+                     (left_inclusive ? " (inclusive)" : " (exclusive)") + " AND " +
+                     std::to_string(boost::get<double>(right)) + (right_inclusive ? " (inclusive)" : " (exclusive)"));
```
ekrebs5: ``boost::get<type>`` is used multiple times in this file, why not use ``type_cast_variant<type>``?

src/test/operators/table_scan_between_test.cpp
```diff
@@ -57,50 +57,91 @@ class TableScanBetweenTest : public TypedOperatorBaseTest {
...
+          for (auto offset = ChunkOffset{0}; offset < segment_b->size(); ++offset) {
+            result_ints.emplace_back(boost::get<int>((*segment_b)[offset]));
+          }
+        }
+        std::sort(result_ints.begin(), result_ints.end());
+
+        auto expected = expected_with_null;
+        if (nullable) {
+          // Remove the positions that should not be included because they are NULL
+          expected.erase(std::remove_if(expected.begin(), expected.end(), [](int x) { return x % 3 == 2; }),
```
ekrebs5: ``x % 3 == 2`` how does this work?
Maybe this could be changed to something more readable.

src/lib/expression/between_expression.hpp
```diff
@@ -10,17 +10,22 @@ class BetweenExpression : public AbstractPredicateExpression {
  public:
   BetweenExpression(const std::shared_ptr<AbstractExpression>& value,
                     const std::shared_ptr<AbstractExpression>& lower_bound,
-                    const std::shared_ptr<AbstractExpression>& upper_bound);
+                    const std::shared_ptr<AbstractExpression>& upper_bound, bool left_inclusive = true,
+                    bool right_inclusive = true);
```
mrks: Please update the PR description and especially the title and add the inclusive/exclusive part. When looking at the git log, this would otherwise be missed.

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -54,19 +57,56 @@ void ColumnBetweenTableScanImpl::_scan_generic_segment(const BaseSegment& segmen
...
-      return position.value() >= typed_left_value && position.value() <= typed_right_value;
-    };
 
-    _scan_with_iterators<true>(comparator, it, end, chunk_id, matches);
+    if (_left_inclusive) {
+      if (_right_inclusive) {
+        auto comparator = [typed_left_value, typed_right_value](const auto& position) {
+          return position.value() >= typed_left_value && position.value() <= typed_right_value;
+        };
+        _scan_with_iterators<true>(comparator, it, end, chunk_id, matches);
```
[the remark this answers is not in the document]
mrks: You don't want to wrap it into an `std::function` though because those are slow as hell. Still, I am a bit afraid on what this does to the compile time. :/

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -54,19 +57,56 @@ void ColumnBetweenTableScanImpl::_scan_generic_segment(const BaseSegment& segmen
...
+      } else {
```
mrks: I think having four ifs `_left_inclusive && _right_inclusive`, `_left_inclusive && !...` would be easier to read.

src/lib/optimizer/optimizer.cpp
```diff
@@ -101,6 +102,8 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {
 
   optimizer->add_rule(std::make_shared<JoinOrderingRule>(std::make_shared<CostModelLogical>()));
 
+  optimizer->add_rule(std::make_shared<BetweenCompositionRule>());
```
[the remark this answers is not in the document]
mrks: Not really. Whatever works, I guess.

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,219 @@
...
+#include <vector>
+
+#include "constant_mappings.hpp"
+#include "logical_query_plan/abstract_lqp_node.hpp"
+#include "logical_query_plan/lqp_utils.hpp"
+#include "logical_query_plan/predicate_node.hpp"
+#include "statistics/table_statistics.hpp"
+#include "utils/assert.hpp"
+
+using namespace opossum::expression_functional;  // NOLINT
```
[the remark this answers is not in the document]
mrks: Nah, this one is fine. The style guide rule is more concerned about `using namespace std;`.

src/lib/optimizer/strategy/between_composition_rule.hpp
```diff
@@ -0,0 +1,52 @@
+#pragma once
```
mrks: The comments in this file can be improved. I have no idea what a column boundary is, why it needs to hold onto a predicate node and why it has a ValueExpression.

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,219 @@
...
+      upper_bound_value_expression = nullptr;
+      lower_bound_value_expression = nullptr;
+      left_inclusive = true;
+      right_inclusive = true;
+      node_scope = std::vector<std::shared_ptr<AbstractLQPNode>>();
+      current_column_expression = boundary.column_expression;
+    }
+
+    if (boundary.type == ColumnBoundaryType::UpperBoundaryInclusive) {
+      if (upper_bound_value_expression == nullptr ||
```
mrks: No need for `== nullptr`

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,219 @@
...
+      if (lower_bound_value_expression == nullptr ||
+          lower_bound_value_expression->value <= boundary.value_expression->value) {
+        lower_bound_value_expression = boundary.value_expression;
+        left_inclusive = false;
+        node_scope.push_back(boundary.node);
+      }
+    }
+  }
+
+  // Apply boundary check again for the last column
```
mrks: What's special about the last column?

src/lib/optimizer/strategy/between_composition_rule.cpp
mrks: Consistency: We use `x.begin()` instead of `std::begin(x)`, even if the latter is not wrong.

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,219 @@
+#include "between_composition_rule.hpp"
```
mrks: This file is missing a whole bunch of comments. There is no high-level explanation of the algorithm and central data structures are not explained. What is a "node scope"? Why can column boundaries be compared by looking at their original column id?

src/test/operators/table_scan_between_test.cpp
mrks: Because the type is already known, so we can save us some overhead.

src/test/operators/table_scan_between_test.cpp
```diff
@@ -57,50 +57,91 @@ class TableScanBetweenTest : public TypedOperatorBaseTest {
     _data_table_wrapper = std::make_shared<TableWrapper>(data_table);
     _data_table_wrapper->execute();
   }
+
+  void _test_between_boundaries(std::vector<std::tuple<AllTypeVariant, AllTypeVariant, std::vector<int>>>& tests,
```
mrks: Naming: Does this really test the boundaries?

mrks: Do you expect someone to be able to use this function just by looking at its signature?

src/test/operators/table_scan_between_test.cpp
```diff
@@ -57,50 +57,91 @@ class TableScanBetweenTest : public TypedOperatorBaseTest {
...
+      }
+    });
+  }
 };
 
 TEST_P(TableScanBetweenTest, ExactBoundaries) {
-  auto tests = std::vector<std::tuple<AllTypeVariant, AllTypeVariant, std::vector<int>>>{
+  // BEWARE: The data values (see above) are type casted, latest when they get dictionary compressed.
+  // The same happens to the boundary values, that is why, for convenience, some boundaries were
+  // pushed up/down to the next leading int.
```
mrks: I can't follow. Pushed up/down from where?

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -0,0 +1,258 @@
...
+      greater_than_equals_(a, 100),
+      PredicateNode::make(
+        greater_than_equals_(a, 200),
+        PredicateNode::make(
+          less_than_equals_(a, 400),
+          PredicateNode::make(
+            less_than_equals_(a, 300),
+            node))));
+
+  const auto expected_lqp = PredicateNode::make(between_(a, 200, 300), node);
```
mrks: Until now, I had no idea that your rule did this. *Please* improve its documentation.

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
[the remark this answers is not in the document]
Miroka96: as far as I know, all lambdas have a unique type, so it seemed impossible to store them into a common variable. Having a method class could be a solution, but I don't think it would be better

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
Miroka96: wouldn't it be less performant?

mrks: I have no idea how the compiler translates those two versions, but I wouldn't be surprised if the result doesn't look anything like either. Given that this is control-level code, not data-level code, readability seems to be of higher performance than a minimal performance improvement that will probably not be discernible from noise.

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  std::shared_ptr<StoredTableNode> node;
+  LQPColumnReference a, b, c;
+  std::shared_ptr<BetweenCompositionRule> _rule;
+};
+
+TEST_F(BetweenCompositionTest, ColumnExpressionLeft) {
+  // clang-format off
+  const auto input_lqp = PredicateNode::make(
```
mrzzzrm: The usual and imo more reable formatting of LQPs written in c++ is one node per line:

```
const auto input_lqp = 
PredicateNode::make(greater_than_equals_(a, 200),
  PredicateNode::make(less_than_equals_(a, 300),
    node))  
```

Please update both your prs accordingly.

README.md
```diff
@@ -112,12 +112,15 @@ Contact: firstname.lastname@hpi.de
 -	Michael  Janke
```
mrzzzrm: (random line)
your rule would not create a predicate expression if a single node has a predicate like `PredicateNode::make(and_(greater_than_(a, 5), less_than_(a, 10))`, right? At least I haven't seen a test for it. 

Note that you can use `flatten_logical_expressions(expr, LogicalOperator::And) to untangle arbitrarily complex AND nestings.

README.md
```diff
@@ -112,12 +112,15 @@ Contact: firstname.lastname@hpi.de
...
 -	Max      Jendruk
 -	Marvin   Keller
+- Mirko    Krause
 -	Sven     Lehmann
+- Tom      Lichtenstein
 -	Jan      Mattfeld
 -	Arne     Mayer
```
mrzzzrm: (random line)
with the changes from this PR, `PredicateCondition::Between` itself becomes ambiguous and always needs the two "inclusion-bools" to convey a precise meaning. Thus still having `PredicateCondition::Between` makes us prone to bugs. Should we have `PredicateCondition::Between`, `PredicateCondition::BetweenLowerExclusive`, `PredicateCondition::BetweenUpperExclusive` and `PredicateCondition::BetweenBothExclusive` @Bensk1

Bouncner:
Can you help to remember what the final decision on reformulating `col like 'a%'` to `col > 'a' and col < 'b'` was? The discussion what about collations and safe assumptions about the char set.

mrks:
@Bouncner You are thinking about #1480

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,219 @@
...
+    } else if (expression->predicate_condition == PredicateCondition::LessThan) {
+      type = ColumnBoundaryType::LowerBoundaryExclusive;
+    }
+    return {
+        node,
+        right_column_expression,
+        left_value_expression,
+        type,
+    };
+  }
```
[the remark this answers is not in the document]
tom-lichtenstein: The similar code is a result of the lazy pointer casting in l. 53-54, which is not necessary if the first condition ist true. The new variable declaration in those lines should increase the understandability of the second, almost identical check and help to understand why it is different to the first one. Therefore we need additional return statements.

README.md
Miroka96: Should the other PredicateConditions be added? SQL only supports the double-inclusive meaning, anyway.

mrzzzrm: Yes, I think @Bensk1 agreed they should be added as well. And just to be clear: `PredicateCondition` is an entirely different level than SQL.

src/lib/expression/between_expression.hpp
Miroka96: @simsieg You are the one, that can update the description?
I think of something like:
To represent single/double side exclusive binary predicate expression the Between Table Scan was extended, too. The BetweenExpression now supports two additional constructor parameters (left_inclusive and right_inclusive), which default to true to comply with the SQL standard.

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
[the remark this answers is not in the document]
Miroka96: There might be some dirty solution (more information in this SO: https://stackoverflow.com/questions/28746744/passing-capturing-lambda-as-function-pointer), but I am not sure whether I want to try it... 
The compile time problem probably persists as long as _scan_with_iterators is templated. A templated wrapper class could be a solution, but would it be any better?
Please tell me, if you have any solutions for this.

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -54,19 +57,56 @@ void ColumnBetweenTableScanImpl::_scan_generic_segment(const BaseSegment& segmen
...
+  }
+
+  ValueID right_value_id;
+  if (_right_inclusive) {
+    right_value_id = segment.upper_bound(_right_value);
+  } else {
+    right_value_id = segment.lower_bound(_right_value);
+  }
+
+  if (left_value_id == INVALID_VALUE_ID || left_value_id == right_value_id || left_value_id >= right_value_id) {
```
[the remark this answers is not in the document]
Miroka96: the second condition is unnecessary, sure. but what do you mean with moving the third condition up in the code? I just get the values in the lines above. I could move it into an assert, but because it is possible and no error case, this does not seem useful.

src/test/base_test.hpp
Miroka96: static just means, it does not depend on any object variables. Therefore, it is moved into the static part of the class. And we just followed the pattern of utility functions above and below, which are all tagged static ;)

src/test/operators/table_scan_between_test.cpp
Miroka96: It took me some time to realize its meaning, because it turned out to be very important for the test results. Otherwise I would add another comment starting with 'Attention', but I found it already described enough. One just needs to read it.
Whatever you like more, I am flexible, but I would like to highlight this fact

src/test/operators/table_scan_between_test.cpp
Miroka96: have a look at line 38: `nullable=nullable` is a dirty hack to work around C++ defect 2313.
I have just copied this code and have not spent any further time questioning it.

Miroka96: It seems like there is still no proper solution: http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html

src/test/operators/table_scan_between_test.cpp
Miroka96: this code snipped was again just copied. anyway, I have tried:
error: expected ‘]’ before ‘__maybe_unused’
     const auto& [data_type, encoding __maybe_unused, nullable] = GetParam();

https://stackoverflow.com/questions/40673080/stdignore-with-structured-bindings
It seems like I can either mark all of the variables as maybe_unused, declare them separately or just keep the current solution.

src/test/operators/table_scan_between_test.cpp
Miroka96: okay, I have added a comment to it. Once again, it was copied/moved from its original place
Anyway, this code just removes approximately every third value. It just generates test data

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+  const auto result_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
+
+  EXPECT_LQP_EQ(result_lqp, expected_lqp);
+}
+
+TEST_F(BetweenCompositionTest, FindOptimalBetween) {
+  // clang-format off
+  const auto input_lqp =
+  PredicateNode::make(greater_than_equals_(_column_a, 100),
+    PredicateNode::make(greater_than_equals_(_column_a, 200),
```
mrzzzrm: Pls test what your algo does when the right argument is NOT a literal.

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+#include "utils/assert.hpp"
+
+#include "logical_query_plan/mock_node.hpp"
+
+using namespace opossum::expression_functional;  // NOLINT
+
+namespace opossum {
+
+class BetweenCompositionTest : public StrategyBaseTest {
```
mrzzzrm: Please test that your rule does not
* Pull predicates past nodes that it cannot legally pull past (Aggregates, Non-Inner joins, Union(?))
* Think about what the behaviour should be in diamond-shaped LQPs (Unions) and test accordingly.
* As mentioned elsewhere, test what happens if the right argument to a Predicate is not a Value or the left is not a Column.

src/lib/optimizer/strategy/between_composition_rule.hpp
```diff
@@ -0,0 +1,61 @@
...
+namespace opossum {
+
+class AbstractLQPNode;
+class PredicateNode;
+
+/**
+ * The ColumnBoundaryType defines whether a value represents a boundary for a column or not
+ * and if it is a boundary it also defines which kind of boundary it is including the inclusive and exclusive property.
+**/
+enum ColumnBoundaryType {
```
mrzzzrm: The enum (pls use enum class!) and the struct exposed here are internal to the rule, pls put them into an anonymous namespace in the cpp. No one else should know about them

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+void BetweenCompositionRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node) const {
+  if (node->type == LQPNodeType::Predicate) {
+    std::vector<std::shared_ptr<AbstractLQPNode>> predicate_nodes;
+
+    // Gather adjacent PredicateNodes
+    auto current_node = node;
+    while (current_node->type == LQPNodeType::Predicate) {
+      // Once a node has multiple outputs, we're not talking about a predicate chain anymore
+      if (current_node->outputs().size() > 1 || current_node->right_input() != nullptr) {
```
mrzzzrm: the topmost node could very legally have multiple outputs and it would still be a chain!

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,265 @@
...
+    }
+    // Remove node from lqp in order to rearrange them later
+    lqp_remove_node(predicate);
+  }
+
+  // sort boundaries according to their columns to group boundaries referring to the same column
+  std::sort(boundaries.begin(), boundaries.end(), _column_boundary_comparator);
+
+  // Store the highest lower bound and the lowest upper bound for a column in order to get an optimal BetweenExpression
+  std::shared_ptr<LQPColumnExpression> current_column_expression = nullptr;
```
mrzzzrm: not needed and anti-idiomatic to intitalize smart prts with nullptr

src/lib/optimizer/strategy/between_composition_rule.hpp
```diff
@@ -0,0 +1,61 @@
...
+struct ColumnBoundary {
+  std::shared_ptr<PredicateNode> node;
+  std::shared_ptr<LQPColumnExpression> column_expression;
+  std::shared_ptr<ValueExpression> value_expression;
+  ColumnBoundaryType type;
+};
+
+/**
+ * The BetweenCompositionRule searches for a chain of predicate nodes and substitutes BinaryPredicateConditions
+ * to BetweenExpressions within this chain. The algorithm checks wether two or more BinaryPredicateConditions
```
mrzzzrm: typo in "wether"

src/lib/optimizer/strategy/between_composition_rule.hpp
```diff
@@ -0,0 +1,61 @@
...
+**/
+struct ColumnBoundary {
+  std::shared_ptr<PredicateNode> node;
+  std::shared_ptr<LQPColumnExpression> column_expression;
+  std::shared_ptr<ValueExpression> value_expression;
+  ColumnBoundaryType type;
+};
+
+/**
+ * The BetweenCompositionRule searches for a chain of predicate nodes and substitutes BinaryPredicateConditions
```
mrzzzrm: pls introduce your rule with an example like "Rewrite a>=5 AND a<=10 to a BETWEEN 5 AND 10". Makes it far easier to get an idea of what it does

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,265 @@
...
+      default:
+        break;
+    }
+  } else {
+    value_expression = std::dynamic_pointer_cast<ValueExpression>(expression->left_operand());
+    column_expression = std::dynamic_pointer_cast<LQPColumnExpression>(expression->right_operand());
+
+    // Case: "ValueExpression [CONDITION] ColumnExpression" will be checked
+    // Boundary type will be set according to this order
+    if (value_expression != nullptr && column_expression != nullptr) {
```
mrzzzrm: So the rule does NOT work for things like `a >= b AND a <= c`? I think this would be more in its scope than pruning expression (referring to the "find optimal between" thing). If you have to sacrifice one thing to make the other work, pls sacrifice the latter instead of the first.

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,265 @@
...
+  std::shared_ptr<ValueExpression> lower_bound_value_expression = nullptr;
+  std::shared_ptr<ValueExpression> upper_bound_value_expression = nullptr;
+  auto left_inclusive = true;
+  auto right_inclusive = true;
+
+  // Store all nodes of a column in a temporary node_scope vector.
+  // If no boundary could be created for this column, all nodes in the node_scope have to be recovered.
+  auto node_scope = std::vector<std::shared_ptr<AbstractLQPNode>>();
+
+  for (const auto& boundary : boundaries) {
```
mrzzzrm: I think a normal `column -> constraints` ("constraint" == "ColumnBounds" in your terminology) map would be much simpler than this sorted vector iteration thingy with a lot of temporary state that needs to be "reset" occasionally.

README.md
Miroka96: Thanks for the explanation. I will add it soon.

I have to talked with Jan and assume that PredicateCondition::Between should be renamed to BetweenBothInclusive, too. Right?

mrzzzrm: BetweenInclusive would do it for me, but no strong opinions

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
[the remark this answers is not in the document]
aloeser: Dropping the second condition should be fine for us

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+}
+
+TEST_F(BetweenCompositionTest, NoColumnRange) {
+  // clang-format off
+  const auto input_lqp =
+  PredicateNode::make(greater_than_equals_(_column_a, 200),
+    PredicateNode::make(less_than_equals_(_column_b, 300),
+      _node));
+
+  const auto expected_lqp = input_lqp;
```
Bensk1: You want to test that nothing changes, but I believe instead you will always compare the LQP with itself since `expected_lqp` and `input_lqp` point to the same LQP and, therefore, `expected_lqp` will also change if `input_lqp` changes. 

Use `const auto expected_lqp = input_lqp->deep_copy()` instead. There are more cases further down.

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+  // clang-format on
+
+  const auto result_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
+
+  EXPECT_LQP_EQ(result_lqp, expected_lqp);
+}
+
+TEST_F(BetweenCompositionTest, FindOptimalBetween) {
+  // clang-format off
+  const auto input_lqp =
```
Bensk1: Does it make sense to mix in some `greater_than` without the `equals` in here?

src/test/base_test.hpp
ekrebs5: yes, thanks for the explanation

src/lib/optimizer/strategy/between_composition_rule.cpp
aloeser: In that case the given scenario will (probably) not occur. I think we can mark this as resolved then?

README.md
Bouncner: Can somebody list the final decision here? I am confused by those `both/neither/left/right/exclusive`s. :)

Miroka96: I want to turn it into: BetweenInclusive, BetweenLowerExclusive, BetweenUpperExclusive, BetweenExclusive

mrzzzrm: and i agree with that

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -54,19 +57,52 @@ void ColumnBetweenTableScanImpl::_scan_generic_segment(const BaseSegment& segmen
 
     auto typed_left_value = type_cast_variant<ColumnDataType>(_left_value);
     auto typed_right_value = type_cast_variant<ColumnDataType>(_right_value);
-    auto comparator = [typed_left_value, typed_right_value](const auto& position) {
-      return position.value() >= typed_left_value && position.value() <= typed_right_value;
-    };
 
-    _scan_with_iterators<true>(comparator, it, end, chunk_id, matches);
+    if (_left_inclusive && _right_inclusive) {
```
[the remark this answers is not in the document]
Miroka96: The first step is done, renaming Between to BetweenInclusive and it still compiles.
Btw, regarding the parameter naming in with_comparator, why is it called 'func' instead of 'comparator' as in the method name?

README.md
Miroka96: Please have a look at it again

src/lib/expression/between_expression.cpp
```diff
@@ -6,27 +6,48 @@
 
 namespace opossum {
 
+static PredicateCondition _get_between_predicate_expression(bool left_inclusive, bool right_inclusive) {
```
Miroka96: btw, should left_inclusive and right_inclusive be renamed to lower_inclusive and upper_inclusive?

src/lib/optimizer/strategy/between_composition_rule.cpp
Miroka96: is it resolved?

mrks:
Is this ready for another look or is it WIP?

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -547,7 +547,10 @@ ExpressionEvaluator::_evaluate_predicate_expression<ExpressionEvaluator::Bool>(
       return _evaluate_binary_predicate_expression<ExpressionEvaluator::Bool>(
           static_cast<const BinaryPredicateExpression&>(predicate_expression));
 
-    case PredicateCondition::Between:
+    case PredicateCondition::BetweenInclusive:
```
mrzzzrm: This will not work, how are all four Between-variant to be handled by the same code?!? Please also add tests for this to the ExpressionEvaluatorTests, should be straight forward.

mrzzzrm: Let me know if you need help adjusting the ExpressionEvaluator.

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -1032,7 +1035,10 @@ PosList ExpressionEvaluator::evaluate_expression_to_pos_list(const AbstractExpre
           });
         } break;
 
-        case PredicateCondition::Between:
+        case PredicateCondition::BetweenInclusive:
+        case PredicateCondition::BetweenLowerExclusive:
```
mrzzzrm: (same here, you need to actually handle these four different cases, they can't just be routed to the exact same code path)

src/lib/expression/between_expression.cpp
```diff
@@ -6,27 +6,48 @@
...
+    return PredicateCondition::BetweenExclusive;
+  }
+  Fail("Unreachable Case");
+}
+
 BetweenExpression::BetweenExpression(const std::shared_ptr<AbstractExpression>& value,
                                      const std::shared_ptr<AbstractExpression>& lower_bound,
-                                     const std::shared_ptr<AbstractExpression>& upper_bound)
-    : AbstractPredicateExpression(PredicateCondition::Between, {value, lower_bound, upper_bound}) {}
+                                     const std::shared_ptr<AbstractExpression>& upper_bound, bool left_inclusive,
```
mrzzzrm: the left exclusive and right exclusive properties need to go from the BetweenExpression. Instread, it should take a PredicateCondition as a constructor parameter and check whether it is a valid between predicate condition

src/lib/optimizer/strategy/between_composition_rule.hpp
mrzzzrm: pls make this an enum class and remove it from the opossum namespace, instead make it a member type of BetweenCompositionRule

src/lib/optimizer/strategy/between_composition_rule.hpp
```diff
@@ -0,0 +1,61 @@
...
+  UpperBoundaryExclusive,
+};
+
+/**
+ * A column boundary is a normalized format for further computation
+ * that allows us to store a column and a value expression of a predicate node. The value represents a boundary
+ * for the column, if the ColumnBoundaryType does not equal None.
+ * The PredicateNode has to be stored so it can be removed if a substitution of two ColumnBoundaries is possible.
+**/
+struct ColumnBoundary {
```
mrzzzrm: pls remove it from the opossum namespace, instead make it a member type of BetweenCompositionRule

src/test/optimizer/strategy/between_composition_rule_test.cpp
tom-lichtenstein: I will add another test for this case

src/test/optimizer/strategy/between_composition_rule_test.cpp
tom-lichtenstein: Do you mean a column_a >= column_b relation?

src/lib/optimizer/strategy/between_composition_rule.hpp
tom-lichtenstein: Please check again if the documentation is sufficient now

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,219 @@
...
+    const auto expression = std::dynamic_pointer_cast<BinaryPredicateExpression>(predicate_node->predicate());
+    if (expression != nullptr) {
+      const auto boundary = _get_boundary(expression, predicate_node);
+      if (boundary.type != ColumnBoundaryType::None) {
+        boundaries.push_back(boundary);
+      } else {
+        predicate_nodes.push_back(predicate);
+      }
+    } else {
+      predicate_nodes.push_back(predicate);
```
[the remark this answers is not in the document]
tom-lichtenstein: Nothing happens. the case a >= 200 OR a <= 300 is covered in the test: OrExpression

README.md
tom-lichtenstein: We added the support and tests for 'and expressions'.

src/lib/expression/evaluation/expression_evaluator.cpp
Miroka96: should be fixed

src/lib/expression/evaluation/expression_evaluator.cpp
Miroka96: should be fixed

src/test/optimizer/strategy/between_composition_rule_test.cpp
Miroka96: is it resolved?

src/lib/expression/between_expression.cpp
```diff
@@ -6,30 +6,91 @@
 
 namespace opossum {
 
-BetweenExpression::BetweenExpression(const std::shared_ptr<AbstractExpression>& value,
-                                     const std::shared_ptr<AbstractExpression>& lower_bound,
-                                     const std::shared_ptr<AbstractExpression>& upper_bound)
-    : AbstractPredicateExpression(PredicateCondition::Between, {value, lower_bound, upper_bound}) {}
+PredicateCondition BaseBetweenExpression::get_between_predicate_expression(bool left_inclusive, bool right_inclusive) {
```
mrzzzrm: Thus function is only used in the rule, I think it should be there instead of here. We have to try to keep the Expression interfaces small and clean

src/lib/expression/between_expression.hpp
```diff
@@ -6,15 +6,26 @@
 
 namespace opossum {
 
-class BetweenExpression : public AbstractPredicateExpression {
+class BaseBetweenExpression : public AbstractPredicateExpression {
```
mrzzzrm: As far as I can tell there is no need to have four(!) distinct classes for Between expressions. We cluster similar expressions into the same expression-class - and just as >, >=, <, ... are all "BinaryPredicateExpression", all four between-types are "BetweenExpressions". Please revert this back to a single BetweenExpression-class.

src/lib/expression/between_expression.hpp
```diff
@@ -6,15 +6,26 @@
...
  public:
-  BetweenExpression(const std::shared_ptr<AbstractExpression>& value,
-                    const std::shared_ptr<AbstractExpression>& lower_bound,
-                    const std::shared_ptr<AbstractExpression>& upper_bound);
+  static PredicateCondition get_between_predicate_expression(bool left_inclusive, bool right_inclusive);
```
mrzzzrm: these four static functions don't have too much to do with the class they are in and should be - just like `is_binary_predicate_condition` - free functions. If they are only used in a single cpp, put them there instead. Again, this is to keep the interfaces of the expression classes noise-free

src/lib/expression/between_expression.hpp
```diff
@@ -23,4 +34,32 @@ class BetweenExpression : public AbstractPredicateExpression {
   ExpressionPrecedence _precedence() const override;
 };
 
+class BetweenInclusiveExpression : public BaseBetweenExpression {
```
mrzzzrm: Just to stress the point, look at these 26 lines + the corresponding definitions in the cpp - they are just code-noise but have no functional purpose and therefore only put cognitive load on the reader.

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -71,15 +71,67 @@ std::shared_ptr<AbstractExpression> rewrite_between_expression(const AbstractExp
...
-  const auto lte_expression = less_than_equals_(between_expression->value(), between_expression->upper_bound());
+  const auto gte_expression =
+      greater_than_equals_(between_inclusive_expression->value(), between_inclusive_expression->lower_bound());
+  const auto lte_expression =
+      less_than_equals_(between_inclusive_expression->value(), between_inclusive_expression->upper_bound());
 
   return and_(gte_expression, lte_expression);
 }
 
+std::shared_ptr<AbstractExpression> rewrite_between_lower_exclusive_expression(const AbstractExpression& expression) {
```
mrzzzrm: Please use a single method for rewriting the Between, this is just too much code duplication right now

```c++

std::shared_ptr<AbstractExpression> rewrite_between_expression(const AbstractExpression& expression) {
  const auto* between_expression = dynamic_cast<const BetweenExpression*>(&expression);
  Assert(between_expression, "Expected Between Expression");

  const auto predicate_condition = between_expression->predicate_condition;
  
  const auto lower_bound_condition = is_between_lower_inclusive(predicate_condition) ? 
                                         PredicateCondition::GreaterThanEquals : PredicateCondition::GreaterThan;
  const auto upper_bound_condition = is_between_upper_inclusive(predicate_condition) ? 
                                         PredicateCondition::LessThanEquals : PredicateCondition::LessThan;

  const auto lower_bound = std::make_shared<BinaryPredicateExpression>(lower_bound_condition, 
    between_expression->value(), between_expression->lower_bound());
  const auto upper_bound = std::make_shared<BinaryPredicateExpression>(upper_bound_condition, 
    between_expression->value(), between_expression->lower_bound());

  return and_(lower_bound, upper_bound);
}
```

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -547,10 +599,22 @@ ExpressionEvaluator::_evaluate_predicate_expression<ExpressionEvaluator::Bool>(
       return _evaluate_binary_predicate_expression<ExpressionEvaluator::Bool>(
           static_cast<const BinaryPredicateExpression&>(predicate_expression));
 
-    case PredicateCondition::Between:
+    case PredicateCondition::BetweenInclusive:
       return evaluate_expression_to_result<ExpressionEvaluator::Bool>(
           *rewrite_between_expression(predicate_expression));
 
+    case PredicateCondition::BetweenLowerExclusive:
```
mrzzzrm: once you just have a single rewrite method there is less code here as well

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -1032,9 +1096,18 @@ PosList ExpressionEvaluator::evaluate_expression_to_pos_list(const AbstractExpre
           });
         } break;
 
-        case PredicateCondition::Between:
+        case PredicateCondition::BetweenInclusive:
           return evaluate_expression_to_pos_list(*rewrite_between_expression(expression));
 
+        case PredicateCondition::BetweenLowerExclusive:
```
mrzzzrm: once you just have a single rewrite method there is less code here as well

src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
```diff
@@ -270,16 +273,49 @@ std::shared_ptr<const JitExpression> JitAwareLQPTranslator::_try_translate_expre
...
-                                            jit_expression_arguments[1], jit_source.add_temporary_value());
-        auto upper_bound_check =
-            std::make_shared<JitExpression>(jit_expression_arguments[0], JitExpressionType::LessThanEquals,
-                                            jit_expression_arguments[2], jit_source.add_temporary_value());
-
+        std::shared_ptr<JitExpression> lower_bound_check;
+        std::shared_ptr<JitExpression> upper_bound_check;
+
+        switch (jit_expression_type) {
+          case JitExpressionType::BetweenInclusive:
```
mrzzzrm: similar to the rewrite in the ExpressionEvaluator, this can be done with way less code.

src/test/expression/expression_evaluator_to_pos_list_test.cpp
```diff
@@ -99,6 +108,12 @@ TEST_F(ExpressionEvaluatorToPosListTest, PredicatesWithOnlyLiterals) {
   EXPECT_TRUE(test_expression(table_b, ChunkID{1}, *greater_than_(5, 1), {0, 1, 2}));
   EXPECT_TRUE(test_expression(table_b, ChunkID{1}, *between_(2, 5, 6), {}));
   EXPECT_TRUE(test_expression(table_b, ChunkID{0}, *between_(2, 1, 6), {0, 1, 2, 3}));
+  EXPECT_TRUE(test_expression(table_b, ChunkID{1}, *between_lower_exclusive_(2, 5, 6), {}));
+  EXPECT_TRUE(test_expression(table_b, ChunkID{0}, *between_lower_exclusive_(2, 1, 6), {0, 1, 2, 3}));
```
mrzzzrm: the tests you added would pass even if all between types were rewritten to the same expression. Please write tests that actually test the lower and upper bound.

src/test/expression/expression_evaluator_to_values_test.cpp
```diff
@@ -272,6 +272,30 @@ TEST_F(ExpressionEvaluatorToValuesTest, PredicatesLiterals) {
   EXPECT_TRUE(test_expression<int32_t>(*between_(5.1f, null_(), 5), {0}));
   EXPECT_TRUE(test_expression<int32_t>(*between_(null_(), 3.1, 5), {std::nullopt}));
   EXPECT_TRUE(test_expression<int32_t>(*between_(null_(), null_(), null_()), {std::nullopt}));
+  EXPECT_TRUE(test_expression<int32_t>(*between_lower_exclusive_(3, 3.0, 5.0), {0}));
```
mrzzzrm: (II )the tests you added would pass even if all between types were rewritten to the same expression. Please write tests that actually test the lower and upper bound.

src/test/expression/expression_evaluator_to_values_test.cpp
```diff
@@ -289,6 +313,17 @@ TEST_F(ExpressionEvaluatorToValuesTest, PredicatesSeries) {
   EXPECT_TRUE(test_expression<int32_t>(table_a, *between_(b, a, c), {1, std::nullopt, 1, std::nullopt}));
   EXPECT_TRUE(test_expression<int32_t>(table_a, *between_(e, a, f), {1, 0, 0, 0}));
   EXPECT_TRUE(test_expression<int32_t>(table_a, *between_(3.3, a, b), {0, 0, 1, 0}));
+  EXPECT_TRUE(
```
mrzzzrm: (III) the tests you added would pass even if all between types were rewritten to the same expression. Please write tests that actually test the lower and upper bound.

src/lib/expression/evaluation/expression_evaluator.cpp
Miroka96: should be fixed

README.md
```diff
@@ -99,33 +99,36 @@ Contact: firstname.lastname@hpi.de
...
--   David     Schumann
--   Arthur    Silber
--   Daniel    Stolpe
--   Jonathan  Striebel
--   Nils      Thamm
--   Carsten   Walther
--   Lukas     Wenzel
--   Fabian    Wiebe
--   Tim       Zimmermann
+-	Yannick   Bäumer
```
mrks: We changed this to single space on purpose because people were having difficulties with maintaing the formatting.

src/lib/constant_mappings.cpp
```diff
@@ -26,7 +26,7 @@ const boost::bimap<PredicateCondition, std::string> predicate_condition_to_strin
         {PredicateCondition::LessThanEquals, "<="},
         {PredicateCondition::GreaterThan, ">"},
         {PredicateCondition::GreaterThanEquals, ">="},
-        {PredicateCondition::Between, "BETWEEN"},
+        {PredicateCondition::BetweenInclusive, "BETWEEN"},
```
mrks: Add a comment on why the others are missing here?

src/lib/expression/expression_functional.hpp
```diff
@@ -109,6 +109,14 @@ struct ternary final {
   }
 };
 
+template <auto t, typename E>
+struct typed_ternary final {
```
mrks: doc?

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -54,19 +56,34 @@ void ColumnBetweenTableScanImpl::_scan_generic_segment(const BaseSegment& segmen
...
+    });
   });
 }
 
 void ColumnBetweenTableScanImpl::_scan_dictionary_segment(const BaseDictionarySegment& segment, const ChunkID chunk_id,
                                                           PosList& matches,
                                                           const std::shared_ptr<const PosList>& position_filter) const {
-  auto left_value_id = segment.lower_bound(_left_value);
-  auto right_value_id = segment.upper_bound(_right_value);
+  // naming assumption: the left value is always the lower one (otherwise the result is empty)
```
mrks: where is this enforced?

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -86,7 +103,10 @@ void ColumnBetweenTableScanImpl::_scan_dictionary_segment(const BaseDictionarySe
     return;
   }
 
-  if (left_value_id >= static_cast<ValueID>(segment.unique_values_count()) || left_value_id == right_value_id) {
+  if (left_value_id == INVALID_VALUE_ID || left_value_id >= static_cast<ValueID>(segment.unique_values_count()) ||
+      left_value_id >= right_value_id) {
+    // TODO(all)
+    // if (left_value_id >= static_cast<ValueID>(segment.unique_values_count()) || left_value_id == right_value_id) {
```
mrks: ?

src/lib/optimizer/strategy/between_composition_rule.cpp
mrks: Except for the comments below, yes. :)

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+using namespace opossum::expression_functional;  // NOLINT
+
+namespace opossum {
+
+std::string BetweenCompositionRule::name() const { return "Between Composition Rule"; }
+
+/**
+ * _get_boundary takes a BinaryPredicateExpression and the corresponding PredicateNode
+ * as its input and returns a normalized ColumnBoundary. This function checks where the
```
mrks: Still not sure what you mean by "normalized"

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,284 @@
...
+ * LQPColumnExpression and where the ValueExpression is stored in the BinaryPredicateExpression.
+ * The expressions are transferred to a normalized ColumnBoundary format
+ * and are labled with a ColumnBoundaryType, that depends on their positions and the predicate condition
```
mrks: typo, also no comma before that

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  return {
+      column_expression,
+      value_expression,
+      type,
+  };
+}
+
+static PredicateCondition get_between_predicate_condition(bool left_inclusive, bool right_inclusive) {
```
mrks: Could we bring the free-floating functions that have to do with handling the PredicateConditions together in one place?

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,284 @@
...
+      if (boundary.type != ColumnBoundaryType::None) {
+        if (column_boundaries.find(boundary.column_expression->column_reference) == column_boundaries.end()) {
+          column_boundaries[boundary.column_expression->column_reference] = std::vector<ColumnBoundary>();
+        }
+        column_boundaries[boundary.column_expression->column_reference].push_back(boundary);
+      } else {
+        predicate_nodes.push_back(predicate);
+      }
+    }
+    // Remove node from lqp in order to rearrange them later
```
mrks: node vs them?

src/lib/optimizer/strategy/between_composition_rule.hpp
```diff
@@ -0,0 +1,60 @@
...
+**/
+  enum class ColumnBoundaryType {
+    None,
+    LowerBoundaryInclusive,
+    LowerBoundaryExclusive,
+    UpperBoundaryInclusive,
+    UpperBoundaryExclusive,
+  };
+
+  /**
```
mrks: did you run the formatter? These comments usually have their *s aligned.

src/lib/type_comparison.hpp
```diff
@@ -115,6 +115,28 @@ void with_comparator_light(const PredicateCondition predicate_condition, const F
   }
 }
 
+// Function that calls a given functor with the correct std comparator. The light version is not instantiated for
+// > and >=, reducing the number of instantiated templates by a third.
```
mrks: comment looks misplaced

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
Miroka96: I did not know how to merge this statement with the if-clause above, because I do not fully understand, what is happening here

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -86,7 +103,10 @@ void ColumnBetweenTableScanImpl::_scan_dictionary_segment(const BaseDictionarySe
     return;
   }
 
-  if (left_value_id >= static_cast<ValueID>(segment.unique_values_count()) || left_value_id == right_value_id) {
+  if (left_value_id == INVALID_VALUE_ID || left_value_id >= static_cast<ValueID>(segment.unique_values_count()) ||
+      left_value_id >= right_value_id) {
+    // TODO(all)
```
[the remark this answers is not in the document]
Miroka96: unfortunately not, I do not know how to merge the two if-statements and need help for that

src/lib/type_comparison.hpp
```diff
@@ -115,6 +115,28 @@ void with_comparator_light(const PredicateCondition predicate_condition, const F
...
+      return func(std::greater_equal<void>{}, std::less_equal<void>{});
+
+    case PredicateCondition::BetweenLowerExclusive:
+      return func(std::greater<void>{}, std::less_equal<void>{});
+
+    case PredicateCondition::BetweenUpperExclusive:
+      return func(std::greater_equal<void>{}, std::less<void>{});
+
+    case PredicateCondition::BetweenExclusive:
+      return func(std::greater<void>{}, std::less<void>{});
```
Miroka96: void is default sind C++14, should it be kept or better removed?

src/lib/type_comparison.hpp
Miroka96: good point, partially removed

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -54,19 +56,34 @@ void ColumnBetweenTableScanImpl::_scan_generic_segment(const BaseSegment& segmen
 
     auto typed_left_value = type_cast_variant<ColumnDataType>(_left_value);
     auto typed_right_value = type_cast_variant<ColumnDataType>(_right_value);
-    auto comparator = [typed_left_value, typed_right_value](const auto& position) {
-      return position.value() >= typed_left_value && position.value() <= typed_right_value;
-    };
 
-    _scan_with_iterators<true>(comparator, it, end, chunk_id, matches);
+    with_comparator_between(_predicate_condition, [&](auto lower_comparator, auto upper_comparator) {
```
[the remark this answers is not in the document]
Miroka96: okay, combined into one function

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
Miroka96: nowhere, because it is valid input. This comment only struggles with the confusing naming of left_values and lower_values...
My problem with this naming is, that the left_value is always the left_value, but the lower_value is not necessarily the lower one.

My confusion is e.g. line 75-76 where the check is for lower-inclusiveness and then takes the left_value...

Solution? For me, the comment seemed to be the easiest solution to put attention on this

src/lib/optimizer/strategy/between_composition_rule.hpp
Miroka96: yes, but maybe the IDE directly indented it again.

somehow, the linter did not complain, but I think it should in such situations? Our working CI is a proof for the linter not checking this.

Should this be a new issue? As I am running out of time I do not want it to be in our own scope, sorry.

src/test/expression/expression_evaluator_to_values_test.cpp
Miroka96: what about now?

src/test/expression/expression_evaluator_to_values_test.cpp
Miroka96: what about now?

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
mrzzzrm: I think there is no reason for confusion here. If lower_bound > upper_bound than that's just the case and the result of the Between is empty. Still, lower bound is the lower bound and upper bound is the upper bound. They just do not happen to produce a non-empty intersection :man_shrugging:

mrzzzrm:
@simsieg @Miroka96 to save you a bit of time I have made some adjustments, mostly related to the newly introduced Between-types, myself. Some things (like adjusting the `OperatorScanPredicate`) I changed on your branch, for others (#1552, #1553, #1554) I opened follow-up issues.

@Bouncner @Bensk1 @mrks I am happy with everything outside of the BetweenCompositonRule and its Test, both of which I haven't looked at properly yet. IMO we can focus on these two and hopefully get this PR merged quickly.

mrzzzrm:
While reading the rule, I noticed it might drop predicates, i.e. the IN is dropped from this LQP, apparently. Test added

```
  const auto input_lqp =
  PredicateNode::make(and_(greater_than_(_a_a, 100), in_(_a_a, list_(1, 2, 3))),
    PredicateNode::make(less_than_(_a_a, 200),
      _node_a));
```

src/lib/optimizer/strategy/between_composition_rule.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * corresponding (exclusive or inclusive) BetweenExpression. All obsolete BinaryPredicateConditions are removed
+ * after the substitution.
+**/
+class BetweenCompositionRule : public AbstractRule {
+ public:
+  std::string name() const override;
+  void apply_to(const std::shared_ptr<AbstractLQPNode>& node) const override;
+
+ private:
+  /**
```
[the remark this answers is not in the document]
mrks: Ok, weird. But if that's what the formatter wants todo...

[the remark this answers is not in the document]
Bensk1: As far as I can see, comments in other files are also indented.

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,414 @@
...
+  // Store original input and output
+  auto input = predicates.back()->left_input();
+  const auto outputs = predicates.front()->outputs();
+  const auto input_sides = predicates.front()->get_input_sides();
+
+  auto between_nodes = std::vector<std::shared_ptr<AbstractLQPNode>>();
+  auto predicate_nodes = std::vector<std::shared_ptr<AbstractLQPNode>>();
+  std::unordered_map<LQPColumnReference, std::vector<ColumnBoundary>> column_boundaries;
+
+  // Filter predicates with a boundary to the boundaries vector
```
Bensk1: I don't understand that comment

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,414 @@
...
+  for (auto& predicate : predicates) {
+    // A logical expression can contain multiple binary predicate expressions
+    std::vector<std::shared_ptr<BinaryPredicateExpression>> expressions;
+    const auto predicate_node = std::static_pointer_cast<PredicateNode>(predicate);
```
Bensk1: The data type of the provided `predicates` is `AbstractLQPNode` which is also stated by the method comment. However, the calling method ensures that all provided nodes are `PredicateNodes`, right? Should we clarify that we can safely use a static cast here or should we do the cast in the calling method and provide a vector or `PredicateNodes`?

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,414 @@
...
+        predicate_nodes.push_back(predicate);
+      }
+    }
+
+    for (const auto& expression : expressions) {
+      const auto boundary = _get_boundary(expression);
+      if (boundary.type != ColumnBoundaryType::None) {
+        if (boundary.boundary_is_column_expression) {
+          const auto inverse_boundary = _create_inverse_boundary(boundary);
+          if (column_boundaries.find(inverse_boundary.column_expression->column_reference) == column_boundaries.end()) {
```
Bensk1: `.cend()`?

src/lib/optimizer/strategy/between_composition_rule.cpp
```diff
@@ -0,0 +1,414 @@
...
+      } else {
+        predicate_nodes.push_back(predicate);
+      }
+    }
+
+    for (const auto& expression : expressions) {
+      const auto boundary = _get_boundary(expression);
+      if (boundary.type != ColumnBoundaryType::None) {
+        if (boundary.boundary_is_column_expression) {
+          const auto inverse_boundary = _create_inverse_boundary(boundary);
```
Bensk1: Maybe add a comment here to give some intuition about the following?

mrks:
It appears as if you have a performance regression for TPC-H 12, where a single execution was previously 392.534 ms and now is 1505.157 ms. This might, however, be just because your trunk is a bit old. Could you please post a `compare_benchmarks.py` of the current master with the current master + this PR once you have merged those?

mrzzzrm:
I think I see room for making the rule shorter by a good portion, and I am a little concerned that it does effectively do two different things: Between composition and - sometimes? - predicate pruning. But I won't have the time in the intermediate future to address this and apart from the mentioned things the code is fine and seems to work. Can be merged from my perspective.

mrks:
For the TPC-H 12 regression, we should only create BETWEENs when the two values are fixed values (or correlated parameter expressions). The ColumnBetweenTableScanImpl cannot handle columns.

mrks:
(on it right now)

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -54,53 +56,70 @@ void ColumnBetweenTableScanImpl::_scan_generic_segment(const BaseSegment& segmen
...
   }
 
-  const auto value_id_diff = right_value_id - left_value_id;
+  /**
+   * No early out possible: Actually scan the attribute vector
+   */
 
-  const auto comparator = [left_value_id, value_id_diff](const auto& position) {
-    // Using < here because the right value id is the upper_bound. Also, because the value ids are integers, we can do
-    // a little hack here: (x >= a && x < b) === ((x - a) < (b - a)); cf. https://stackoverflow.com/a/17095534/2204581
```
mrks: Any reason you removed this? This makes the difference between 21.255 iter/s and 29.1051 iter/s. Admittedly, the comment should have explained that this was done for performance reasons.

mrks: @mrzzzrm

mrzzzrm: Distant memories.... I think there was a reason, but it may well be that it became obsolete with the adjustment of upper_bound_value_id:

```c++
  if (upper_bound_value_id == INVALID_VALUE_ID) {
    upper_bound_value_id = segment.unique_values_count();
  }
```

Feel free to revert, if it doesn't trigger any tests it should be fine.

mrks:
Added TPC-H to PR description

PR #1484, opened 2019-02-08
Optimizer: store rules as unique_ptr

### LINKED ISSUE #1476: Optimizer keeps Rules as shared_ptr
The optimizer stores the rules as shared_ptr. Is this necessary? Could these be unique_ptr?

mrks:
> The optimizer stores the rules as shared_ptr. Is this necessary? Could these be unique_ptr?

No. Yes.

Fixes #1476.

### DIFF
--- a/src/test/optimizer/optimizer_test.cpp
+++ b/src/test/optimizer/optimizer_test.cpp
@@ -48,7 +48,7 @@ TEST_F(OptimizerTest, OptimizesSubqueries) {
   // A "rule" that just collects the nodes it was applied to
   class MockRule : public AbstractRule {
    public:
-    MockRule(std::unordered_set<std::shared_ptr<AbstractLQPNode>>& nodes) : nodes(nodes) {}
+    explicit MockRule(std::unordered_set<std::shared_ptr<AbstractLQPNode>>& nodes) : nodes(nodes) {}
     std::string name() const override { return "Mock"; }
 
     void apply_to(const std::shared_ptr<AbstractLQPNode>& root) const override {
@@ -126,7 +126,7 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
   size_t counter{0};
   class MockRule : public AbstractRule {
    public:
-    MockRule(size_t& counter) : counter(counter) {}
+    explicit MockRule(size_t& counter) : counter(counter) {}
     std::string name() const override { return "Mock"; }
 
     void apply_to(const std::shared_ptr<AbstractLQPNode>& root) const override { ++counter; }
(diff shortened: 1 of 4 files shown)


src/test/optimizer/optimizer_test.cpp
```diff
@@ -120,26 +123,28 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
...
-    mutable size_t counter{0};
+    size_t& counter;
   };
 
-  const auto rule = std::make_shared<MockRule>();
+  auto rule = std::make_unique<MockRule>(counter);
 
   Optimizer optimizer{};
-  optimizer.add_rule(rule);
+  optimizer.add_rule(std::move(rule));
```
Bensk1: `add_rule()` uses `std::move()` internally. Is the move here necessary or convention?

mrzzzrm: always gotta move unique_ptrs around, cannot do anything else with those cumbersome simpletons

mrks: How else would `rule` become an xvalue suitable for being moved?

mrzzzrm:
I mildly disagree with the presupposition of this PR. unique_ptrs, due to having to be moved around (and I hate xvalues in my debugger, it's a pain!!), are not a good default.

src/test/optimizer/optimizer_test.cpp
Bensk1: Right, unique_ptrs are used way too rarely.

mrks:
On the other hand, paying `shared_ptr`s' overhead of reference counting and atomic accesses just to  save us from having to write `std::move` doesn't sound right either. What's up with the debbugger? I haven't experienced that issue yet.

mrzzzrm:
> What's up with the debbugger? I haven't experienced that issue yet.

Breakpoint. Some frames up the stack, an empty object that you're really sure shouldn't be empty. You see the initialisation code just there. Confused as hell, stack corruption, maybe? Turns out: Was moved out of somewhere deeper into the stack.

Move-semantics for performance critical stuff.

mrzzzrm:
I while back we still had some code in which people diligently moved their shared_ptrs around in performance non-critical code, probably because they followed the rule of thumbs that it's faster. But debugging was made harder for no real performance gain and more noise in the code. 

So I am just saying, we shouldn't just start walking around replacing shared_ptrs with unique_ptrs just because we can. My experience is that it can make working with the code harder and this just ends in way more meaningful optimizations not getting done because people got stuck for longer in needlessly noisy/complex code.

This PR is a minor thing and I am obviously not arguing about it in particular, just about the precedent that it sets.

mrks:
I would argue that the example where something is moved somewhere deep in the stack is flawed in a different way. In any case, this should not be taken as a precedent, and I would agree that we should check future PRs that change shared_ptrs to unique_ptrs only based on their particular merit.

PR #1491, opened 2019-02-13
Remove obsolete comment in SQLiteTestrunnerTest

mrzzzrm:


src/test/sql/sqlite_testrunner/sqlite_testrunner_encodings.cpp
```diff
@@ -4,7 +4,6 @@
 
 namespace opossum {
 
-
```
mrzzzrm: triggering fullci like a pro

Issue #1393, opened 2019-01-09
Transactions with failed operators can be committed


mrks:
Example:

src/test/operators/delete_test.cpp
```
@@ -184,7 +184,7 @@ TEST_F(OperatorsDeleteTest, UpdateAfterDeleteFails) {

   delete_op->execute();

-  t1_context->commit();
+  ASSERT_FALSE(t1_context->commit());
```

(1) This should fail.
(2) Operators should not execute if it is known that the transaction is a goner.
(3) Enforce checking of the transaction result: `[[nodiscard]] bool TransactionContext::commit();`
(4) If `commit()` fails the transaction should have already been rolled back. This should be covered by an assertion.

d-justen:
I have looked into this now, and for me it is not clear, why this should fail. If I understand this test correctly, the deletion is supposed to work fine, but only the update that comes after should fail, because its rows to update are already deleted. 
Trying to commit this update actually does fail, because of the DebugAssert in `_prepare_commit()`:
```
DebugAssert(([this]() {
                for (const auto& op : _rw_operators) {
                  if (op->state() != ReadWriteOperatorState::Executed) return false;
                }
                return true;
              }()),
              "All read/write operators need to be in state Executed (especially not Failed).");
```
But I guess then the `commit()` should rather return false, than to raise an exception, right? 

Regarding (4): If you are trying to commit a transaction that is already rolled back, you will get a logic error, because it is an illegal phase transition (the transaction context will try to transition from RolledBack to Commiting). Isn't this the wanted behavior? Or is this a misunderstanding and the operators are actually supposed to rollback the transaction, if they fail?

mrks:
It actually appears as if the situation is not as bad as originally expected. I am having some trouble defining what the expected / best behavior is, though.

> But I guess then the commit() should rather return false, than to raise an exception, right?

Correct. During regular execution, we should never see a single exception, even if we have a transaction conflict. After all those conflicts are expected to some degree.

Here's what we have:
1. Once a Tx is known to have failed, operators that were already scheduled should not be executed. This is already done in abstract_operator.cpp.
2. `SQLPipeline::get_result_tables` checks whether the context has been aborted and returns an empty result table.

What I think should change:
1. The return value of `commit()` should not be ignored.
[...]

PR #1493, opened 2019-02-13
[DYOD18/19] MvccDeletePlugin with tests

julianmenzler:
This pull request includes the implementation of our `MvccDeletePlugin` and associated tests. 
Necessary changes within the hyrise codebase are addressed in separate pull requests (#1462 and #1468) to keep this PR rather small.


src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "operators/validate.hpp"
+#include "storage/pos_list.hpp"
+#include "storage/reference_segment.hpp"
+#include "storage/table.hpp"
+
+namespace opossum {
+
+MvccDeletePlugin::MvccDeletePlugin()
+    : _sm(StorageManager::get()),
+      _delete_threshold_rate_invalidated_rows(0.8),
```
nilsthamm: Have you thought about using constants here so the parameter can be better configured?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+void MvccDeletePlugin::stop() {
+  // Call destructor of PausableLoopThread to terminate its thread
+  _loop_thread_logical_delete.reset();
+  _loop_thread_physical_delete.reset();
+}
+
+/**
+ * This function analyzes each chunk of every table and triggers a chunk-cleanup-procedure if a certain threshold of invalidated rows is exceeded.
```
nilsthamm: That line looks a bit long. Have you used the formater?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+ */
+void MvccDeletePlugin::_logical_delete_loop() {
+  for (auto& [table_name, table] : _sm.tables()) {
+    // Check all chunks, except for the last one, which is currently used for insertions
+    if (table->chunk_count() == 1) return;
```
nilsthamm:
```suggestion
    if (table->chunk_count() <= 1) return;
```

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+  if (success) {
+    DebugAssert(StorageManager::get().get_table(table_name)->get_chunk(chunk_id)->get_cleanup_commit_id() !=
+                    MvccData::MAX_COMMIT_ID,
+                "Chunk needs to be deleted logically before deleting it physically.")
+
+        std::unique_lock<std::mutex>
+            lock(_mutex_physical_delete_queue);
+    _physical_delete_queue.emplace(table_name, chunk_id);
+  } else {
+    std::cout << "Logical delete of chunk " << chunk_id << " failed." << std::endl;
```
nilsthamm: I think it would be better if this throws an exception.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+void MvccDeletePlugin::_delete_chunk(const std::string& table_name, const ChunkID chunk_id) {
+  // Delete chunk logically
+  bool success = _delete_chunk_logically(table_name, chunk_id);
+
+  // Queue physical delete
+  if (success) {
+    DebugAssert(StorageManager::get().get_table(table_name)->get_chunk(chunk_id)->get_cleanup_commit_id() !=
+                    MvccData::MAX_COMMIT_ID,
+                "Chunk needs to be deleted logically before deleting it physically.")
```
nilsthamm:
```suggestion
                "Chunk needs to be deleted logically before deleting it physically.");
```

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+  } else {
+    std::cout << "Logical delete of chunk " << chunk_id << " failed." << std::endl;
+  }
+}
+
+bool MvccDeletePlugin::_delete_chunk_logically(const std::string& table_name, const ChunkID chunk_id) {
+  const auto& table = StorageManager::get().get_table(table_name);
+  const auto& chunk = table->get_chunk(chunk_id);
+
+  DebugAssert(chunk != nullptr, "Chunk does not exist. Physical Delete can not be applied.")
```
nilsthamm:
```suggestion
  DebugAssert(chunk != nullptr, "Chunk does not exist. Physical Delete can not be applied.");
```

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+      DebugAssert(chunk_id < (table->chunk_count() - 1),
+                  "MVCC Logical Delete should not be applied on the last/current mutable chunk.")
```
nilsthamm:
```suggestion
                  "MVCC Logical Delete should not be applied on the last/current mutable chunk.");
```

src/plugins/mvcc_delete_plugin.cpp
nilsthamm: And in my opinion, this could also be a normal Assert (it is cheap and crucial for the function)

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+      // Create temporary referencing table that contains the given chunk only
+      auto transaction_context = TransactionManager::get().new_transaction_context();
+  auto gt = std::make_shared<GetTable>(table_name);
+  gt->set_transaction_context(transaction_context);
+
+  auto chunk_count = table->chunk_count();
+  std::vector<ChunkID> excluded_chunk_ids;
+  for (ChunkID excluded_chunk_id = ChunkID{0}; excluded_chunk_id < chunk_count; ++excluded_chunk_id) {
```
nilsthamm: I'm wondering if it would be easier to take all chunk_ids and then remove `chunk_id` from this vector

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+    chunk->set_cleanup_commit_id(transaction_context->commit_id());
+    std::cout << "Deleted chunk " << chunk_id << " logically." << std::endl;
+    return true;
+  }
+}
+
+bool MvccDeletePlugin::_delete_chunk_physically(const std::string& table_name, const ChunkID chunk_id) {
+  const auto& table = StorageManager::get().get_table(table_name);
+
+  DebugAssert(table->get_chunk(chunk_id) != nullptr, "Chunk does not exist. Physical Delete can not be applied.")
```
nilsthamm:
```suggestion
  DebugAssert(table->get_chunk(chunk_id) != nullptr, "Chunk does not exist. Physical Delete can not be applied.");
```

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+      // Check whether there are still active transactions that might use the chunk
+      CommitID cleanup_commit_id = table->get_chunk(chunk_id)->get_cleanup_commit_id();
+  CommitID lowest_snapshot_commit_id = TransactionManager::get().get_lowest_active_snapshot_commit_id();
+  if (cleanup_commit_id < lowest_snapshot_commit_id) {
+    DebugAssert(table->chunks()[chunk_id].use_count() == 1,
+                "At this point, the chunk should be referenced by the "
+                "Table-chunk-vector only.")
```
nilsthamm:
```suggestion
                "Table-chunk-vector only.");
```

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+        // Usage checks have been passed. Apply physical delete now.
+        table->remove_chunk(chunk_id);
+    std::cout << "Deleted chunk " << chunk_id << " physically." << std::endl;
```
nilsthamm:
```suggestion
    std::cout << "Deleted chunk " << chunk_id << " of table " << table_name << " physically." << std::endl;
```

src/test/CMakeLists.txt
```diff
@@ -187,7 +188,7 @@ set(
     utils/plugin_test_utils.hpp
     utils/singleton_test.cpp
     utils/string_utils_test.cpp
-)
+    )
```
nilsthamm: ?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+      _delete_threshold_rate_invalidated_rows(0.8),
+      _delete_threshold_commit_diff_factor(1.5),
+      _idle_delay_logical_delete(std::chrono::milliseconds(1000)),
+      _idle_delay_physical_delete(std::chrono::milliseconds(1000)) {}
+
+const std::string MvccDeletePlugin::description() const { return "Physical MVCC delete plugin"; }
+
+void MvccDeletePlugin::start() {
+  _loop_thread_logical_delete = std::make_unique<PausableLoopThread>(_idle_delay_logical_delete,
+                                                                     [this](size_t) { this->_logical_delete_loop(); });
```
mrks: Code style: We like to capture `&` and call `_logical_delete_loop();` just like that. Not that it is any better, just for consistency's sake.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * This function analyzes each chunk of every table and triggers a chunk-cleanup-procedure if a certain threshold of invalidated rows is exceeded.
+ */
+void MvccDeletePlugin::_logical_delete_loop() {
+  for (auto& [table_name, table] : _sm.tables()) {
+    // Check all chunks, except for the last one, which is currently used for insertions
+    if (table->chunk_count() == 1) return;
+    ChunkID max_chunk_id_to_check = static_cast<ChunkID>(table->chunk_count() - 2);
+
+    for (ChunkID chunk_id = ChunkID{0}; chunk_id <= max_chunk_id_to_check; chunk_id++) {
```
mrks: No need to repeat the type: `auto chunk_id = ...`

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+      const auto& chunk = table->get_chunk(chunk_id);
+      if (chunk && chunk->get_cleanup_commit_id() == MvccData::MAX_COMMIT_ID) {
+        // Calculate metric 1 – Chunk invalidation level
+        const double rate_of_invalidated_rows = static_cast<double>(chunk->invalid_row_count()) / chunk->size();
```
mrks: Did you mean "ratio"?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+        // Calculate metric 2 – Chunk Hotness
+        const CommitID lowest_end_commit_id =
+            *std::min_element(std::begin(chunk->mvcc_data()->end_cids), std::end(chunk->mvcc_data()->end_cids));
```
mrks: This is a bit late for me to realize, but if we are looking at the CIDs anyway, what do we save by maintaining the `invalid_row_count`?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+        const CommitID commit_id_diff = TransactionManager::get().last_commit_id() - lowest_end_commit_id;
+        const CommitID max_commit_id_diff =
+            table->max_chunk_size() * static_cast<CommitID>(_delete_threshold_commit_diff_factor);
+
+        // Evaluate metrics
+        const bool criterion1 = _delete_threshold_rate_invalidated_rows <= rate_of_invalidated_rows;
+        const bool criterion2 = max_commit_id_diff <= commit_id_diff;
+        if (criterion1 || criterion2) _delete_chunk(table_name, chunk_id);
```
mrks: Didn't we plan to use both criteria in conjunction?

mrks: `_delete_chunk_logically`?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+    if (success) {
+      _physical_delete_queue.pop();
+    } else {
+      return;  // wait for more transactions to finish
+    }
+  }
+}
+
+void MvccDeletePlugin::_delete_chunk(const std::string& table_name, const ChunkID chunk_id) {
```
mrks: The names of the methods are unclear to me. Why does `_delete_chunk` call `_delete_chunk_logically`? Maybe have `_delete_chunk_logically` and `_queue_physical_delete`, both called from within the loop?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+    DebugAssert(table->chunks()[chunk_id].use_count() == 1,
+                "At this point, the chunk should be referenced by the "
+                "Table-chunk-vector only.")
+        // Usage checks have been passed. Apply physical delete now.
+        table->remove_chunk(chunk_id);
+    std::cout << "Deleted chunk " << chunk_id << " physically." << std::endl;
+    return true;
+  } else {
+    // Chunk might still be in use. Wait with physical delete.
+    return false;
```
mrks: A method `_delete_chunk_physically` should not decide if it deletes the chunk or not. Either call it `_try_...` or extract the part that checks if it can be deleted.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,177 @@
...
+bool MvccDeletePlugin::_delete_chunk_physically(const std::string& table_name, const ChunkID chunk_id) {
+  const auto& table = StorageManager::get().get_table(table_name);
+
+  DebugAssert(table->get_chunk(chunk_id) != nullptr, "Chunk does not exist. Physical Delete can not be applied.")
+
+      // Check whether there are still active transactions that might use the chunk
+      CommitID cleanup_commit_id = table->get_chunk(chunk_id)->get_cleanup_commit_id();
+  CommitID lowest_snapshot_commit_id = TransactionManager::get().get_lowest_active_snapshot_commit_id();
+  if (cleanup_commit_id < lowest_snapshot_commit_id) {
+    DebugAssert(table->chunks()[chunk_id].use_count() == 1,
```
mrks: Store `const auto& chunk = ...` in the beginning?

src/test/plugins/mvcc_delete_plugin_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+#include "storage/storage_manager.hpp"
+#include "storage/table.hpp"
+#include "utils/load_table.hpp"
+#include "utils/plugin_manager.hpp"
+
+namespace opossum {
+
+class MvccDeletePluginTest : public BaseTest {
+ protected:
+  void load_and_update_table(const std::string& name, const uint8_t val) {
```
mrks: ?

src/test/plugins/mvcc_delete_plugin_test.cpp
```diff
@@ -0,0 +1,210 @@
...
+    const auto segment = table->get_chunk(chunk_id)->get_segment(column_id);
+    const auto& value_alltype = static_cast<const AllTypeVariant&>((*segment)[chunk_offset]);
+    return boost::lexical_cast<int>(value_alltype);
+  }
+
+  std::string _table_name{"mvccTestTable"};
+  inline static std::shared_ptr<AbstractExpression> _column_a;
+};
+
+TEST_F(MvccDeletePluginCoreTest, LogicalDelete) {
```
mrks: As these tests are quite complex, please add some comments on what you actually what to test.

src/plugins/mvcc_delete_plugin.cpp
d-justen: We actually don't want this to throw an exception, because we are expecting transactions to fail, since it's a natural part of the optimistic transaction management approach. That's why we would just rollback our Tx and try it again.

mrks: Maybe write "failed because of MVCC conflict, retrying"?

src/plugins/mvcc_delete_plugin.cpp
d-justen: Formatter didn't change anything, but you're right, it looks too long

src/plugins/mvcc_delete_plugin.cpp
d-justen: Done

src/plugins/mvcc_delete_plugin.cpp
d-justen: Hmm, I get your point, but then we still would have to fill a vector with the chunk ids first, and remove one of the values afterwards. What do you think about an approach like this (iota just fills up a vector with ascending numbers)?
```c++
std::vector<ChunkID> excluded_chunk_ids(table->chunk_count());
std::iota(std::begin(excluded_chunk_ids), std::end(excluded_chunk_ids), 0);
excluded_chunk_ids.erase(std::remove(std::begin(excluded_chunk_ids), std::end(excluded_chunk_ids), chunk_id));
```

src/plugins/mvcc_delete_plugin.cpp
d-justen: Yes, that's right. It wouldn't be too wise to delete a chunk that only fits one of the criteria.

src/test/plugins/mvcc_delete_plugin_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include "operators/update.hpp"
+#include "operators/validate.hpp"
+#include "storage/storage_manager.hpp"
+#include "storage/table.hpp"
+#include "utils/load_table.hpp"
+#include "utils/plugin_manager.hpp"
+
+namespace opossum {
+
+class MvccDeletePluginTest : public BaseTest {
```
mrks: I just realzed that you are not testing the plugin as such. Could you add a test that simply runs the plugin and checks for the expected effects? You might want to put this into hyriseSystemTest and reduce the cycle times for the test.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,158 @@
...
+    transaction_context->rollback();
+    return false;
+  } else {
+    // TODO(all): Check for success of commit, currently (2019-01-11) not yet possible.
+    transaction_context->commit();
+
+    // Mark chunk as logically deleted
+    chunk->set_cleanup_commit_id(transaction_context->commit_id());
+    std::cout << "Deleted chunk " << chunk_id << " logically." << std::endl;
+    return true;
```
mrks: You updated delete_chunk_physically, but this one still returns a bool. At least add "try"

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+            std::unique_lock<std::mutex> lock(_mutex_physical_delete_queue);
+            _physical_delete_queue.emplace(table_name, chunk_id);
+          } else {
+            std::cout << "Logical delete of chunk " << chunk_id << " failed because of MVCC conflict. Retrying..."
+                      << std::endl;
+          }
+        }
+      }
+    }  // for each chunk
```
mrzzzrm: comment not inline without style

src/plugins/mvcc_delete_plugin.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+ public:
+  const std::string description() const final;
+
+  void start() final;
+
+  void stop() final;
+
+ private:
+  struct ChunkSpecifier {
```
mrzzzrm: I do not like names that I need to lookup in order to know what they contain. `using TableAndChunkID = std::pair<std::shared_ptr<Table>, ChunkID>` would sound a bit awkward, but would be more speaking.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,158 @@
...
+    _delete_chunk_physically(chunk_spec.table_name, chunk_spec.chunk_id);
+    _physical_delete_queue.pop();
+  }
+}
+
+bool MvccDeletePlugin::_delete_chunk_logically(const std::string& table_name, const ChunkID chunk_id) {
+  const auto& table = StorageManager::get().get_table(table_name);
+  const auto& chunk = table->get_chunk(chunk_id);
+
+  DebugAssert(chunk != nullptr, "Chunk does not exist. Physical Delete can not be applied.");
```
mrzzzrm: These two asserts could point to serious problems (the plugin running more than once, someone else tinkering with the chunks). I'd make them `Asserts` so they are included in release builds as well.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,158 @@
...
+  DebugAssert(chunk_id < (table->chunk_count() - 1),
+              "MVCC Logical Delete should not be applied on the last/current mutable chunk.");
+
+  // Create temporary referencing table that contains the given chunk only
+  auto transaction_context = TransactionManager::get().new_transaction_context();
+  auto gt = std::make_shared<GetTable>(table_name);
+  gt->set_transaction_context(transaction_context);
+
+  std::vector<ChunkID> excluded_chunk_ids(table->chunk_count());
```
mrzzzrm: what are these three lines doing? They look like they convolutedly construct a vector that contains all ChunkID's but `chunk_id`? Can you doc this? Also be aware that iota and std::remove implies you're iterating over the ChunkID range twice, whereas a simple for loop could do what you want to do in be a single iteration.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,158 @@
...
+  if (update_table->execute_failed()) {
+    transaction_context->rollback();
+    return false;
+  } else {
+    // TODO(all): Check for success of commit, currently (2019-01-11) not yet possible.
+    transaction_context->commit();
+
+    // Mark chunk as logically deleted
+    chunk->set_cleanup_commit_id(transaction_context->commit_id());
+    std::cout << "Deleted chunk " << chunk_id << " logically." << std::endl;
```
mrzzzrm: We don't do logging anywhere else. Do we want this @mrks @Bensk1 ?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,158 @@
...
+    return true;
+  }
+}
+
+void MvccDeletePlugin::_delete_chunk_physically(const std::string& table_name, const ChunkID chunk_id) {
+  const auto table = StorageManager::get().get_table(table_name);
+  const auto chunk = table->get_chunk(chunk_id);
+
+  DebugAssert(chunk.use_count() == 2,
```
mrzzzrm: For the same reason as above, maybe make this an Assert?

src/plugins/mvcc_delete_plugin.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+  static bool _delete_chunk_logically(const std::string& table_name, ChunkID chunk_id);
+  static void _delete_chunk_physically(const std::string& table_name, ChunkID chunk_id);
+
+  std::unique_ptr<PausableLoopThread> _loop_thread_logical_delete, _loop_thread_physical_delete;
+
+  std::mutex _mutex_physical_delete_queue;
+  std::queue<ChunkSpecifier> _physical_delete_queue;
+
+  const double _DELETE_THRESHOLD_RATE_INVALIDATED_ROWS = 0.6;
```
mrzzzrm: constexpr?

src/plugins/mvcc_delete_plugin.cpp
mrks: I thought about this, too. As this won't run automatically but needs to be explicitly enabled, maybe some sort of feedback is not wrong. Once we have actual users of this (TPC-C), we can decide if we want that output or not.

julianmenzler: Indeed, these printouts were for debugging and should not be in this PR. Sorry!

src/plugins/mvcc_delete_plugin.cpp
julianmenzler: Legit point! We should keep the `invalid_row_count` metric though, because It is very efficient. The second metric in comparison is very resource-intensive, since we're evaluating all rows over and over again. This should only be done, if the first criterion (threshold `invalid_row_count` surpassed) becomes positive.

mrks: Yes, keeping it, but only evaluating it if necessary sounds good.

src/plugins/mvcc_delete_plugin.cpp
julianmenzler: Oh, I seem to have misread @mrks's statement. So, to make sure, we should keep these printouts? 
We could also just comment them out.

mrks: Commented out code is always the worst of any set of options. That's bound to become outdated soon. In any case, this is a one-liner that is easy to put back in. However you like.

julianmenzler: Good. I'll remove them.

src/plugins/mvcc_delete_plugin.cpp
julianmenzler: Makes sense, will be addressed.

src/plugins/mvcc_delete_plugin.hpp
julianmenzler: Thanks. Good suggestion!

src/test/plugins/mvcc_delete_plugin_test.cpp
engelfa: We are on it, but we encountered 3 instead of 2 references to the chunk which should be physically deleted next. Allowing 3 references does not seem to cause any problems but we still try to figure out where the additional reference comes from.

src/test/plugins/mvcc_delete_plugin_system_test.cpp
```diff
@@ -0,0 +1,120 @@
...
+    deleted_chunks = 0;
+
+    for (auto chunk_id = ChunkID{0}; chunk_id < table->chunk_count(); ++chunk_id) {
+      if (!table->get_chunk(chunk_id)) {
+        deleted_chunks++;
+      }
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+  }
+
```
mrks: Do I understand you correctly that this is still WIP? Because right now, you're not testing anything.

engelfa: This test only ends if the plugin deleted four chunks successfully, otherwise it stays in the while loop. We will add update and time constraints so the test will always stop.

engelfa:
All requested changes are included now and the pull request can be merged.

src/test/plugins/mvcc_delete_plugin_system_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+       ++loop_count) {
+    const auto table = StorageManager::get().get_table("mvcc_test");
+    _deleted_chunks = 0;
+
+    for (auto chunk_id = ChunkID{0}; chunk_id < table->chunk_count(); ++chunk_id) {
+      if (!table->get_chunk(chunk_id)) {
+        _deleted_chunks++;
+      }
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(1000));
```
mrks: The thread is not guaranteed to be executed exactly after 1s. I'd at least wait for 50% longer.

src/plugins/mvcc_delete_plugin.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+  static bool _try_logical_delete(const std::string& table_name, ChunkID chunk_id);
+  static void _delete_chunk_physically(const std::shared_ptr<Table>& table, ChunkID chunk_id);
+
+  std::unique_ptr<PausableLoopThread> _loop_thread_logical_delete, _loop_thread_physical_delete;
+
+  std::mutex _mutex_physical_delete_queue;
+  std::queue<TableAndChunkID> _physical_delete_queue;
+
+  constexpr static double _DELETE_THRESHOLD_RATE_INVALIDATED_ROWS = 0.6;
```
mrks: comment?

src/plugins/mvcc_delete_plugin.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include "gtest/gtest_prod.h"
+#include "storage/chunk.hpp"
+#include "storage/storage_manager.hpp"
+#include "utils/abstract_plugin.hpp"
+#include "utils/pausable_loop_thread.hpp"
+#include "utils/singleton.hpp"
+
+namespace opossum {
+
+class MvccDeletePlugin : public AbstractPlugin, public Singleton<MvccDeletePlugin> {
```
mrks: Please add a couple of lines on what the plugin does.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,167 @@
...
+    // Mark chunk as logically deleted
+    chunk->set_cleanup_commit_id(transaction_context->commit_id());
+    return true;
+  }
+}
+
+void MvccDeletePlugin::_delete_chunk_physically(const std::shared_ptr<Table>& table, const ChunkID chunk_id) {
+  const auto& chunk = table->get_chunk(chunk_id);
+
+  Assert(chunk.use_count() == 3,
```
mrks: Describe where this number comes from?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,167 @@
...
+  auto update_table = std::make_shared<Update>(table_name, validate_table, validate_table);
+  update_table->set_transaction_context(transaction_context);
+  update_table->execute();
+
+  // Check for success
+  if (update_table->execute_failed()) {
+    transaction_context->rollback();
+    return false;
+  } else {
+    // TODO(all): Check for success of commit, currently (2019-01-11) not yet possible.
```
mrks: reference this in the issue?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,167 @@
...
+      }
+    }
+  }
+}
+
+bool MvccDeletePlugin::_try_logical_delete(const std::string& table_name, const ChunkID chunk_id) {
+  const auto& table = StorageManager::get().get_table(table_name);
+  const auto& chunk = table->get_chunk(chunk_id);
+
+  Assert(chunk != nullptr, "Chunk does not exist. Physical Delete can not be applied.");
```
mrks: logical?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,167 @@
...
+    // Check all chunks, except for the last one, which is currently used for insertions
+    const auto max_chunk_id = static_cast<ChunkID>(table->chunk_count() - 1);
+    for (auto chunk_id = ChunkID{0}; chunk_id < max_chunk_id; chunk_id++) {
+      const auto& chunk = table->get_chunk(chunk_id);
+      if (chunk && !chunk->get_cleanup_commit_id()) {
+        // Calculate metric 1 – Chunk invalidation level
+        const double invalidated_rows_ratio = static_cast<double>(chunk->invalid_row_count()) / chunk->size();
+        const bool criterion1 = _DELETE_THRESHOLD_RATE_INVALIDATED_ROWS <= invalidated_rows_ratio;
+
+        if (criterion1) {
```
mrks: Change this and criterion2 to guard clause

engelfa:
#1393

src/lib/storage/chunk.hpp
```diff
@@ -171,7 +171,9 @@ class Chunk : private Noncopyable {
       * have to look at this chunk anymore.
       * The cleanup commit id represents the snapshot commit id at which transactions can ignore this chunk.
       */
-  const std::optional<CommitID>& get_cleanup_commit_id() const { return _cleanup_commit_id; }
+  const std::optional<CommitID>& get_cleanup_commit_id() const {
+    std::shared_lock lock(_mutex_cleanup_commit_id);
+    return _cleanup_commit_id; }
```
mrks: This looks like you haven't formatted the code.

src/lib/storage/table.cpp
```diff
@@ -125,11 +126,13 @@ uint32_t Table::max_chunk_size() const { return _max_chunk_size; }
 
 std::shared_ptr<Chunk> Table::get_chunk(ChunkID chunk_id) {
   DebugAssert(chunk_id < _chunks.size(), "ChunkID " + std::to_string(chunk_id) + " out of range");
+  std::unique_lock lock(_chunk_mutex);
```
mrks: Our logic guarantees that we do not try to get the pointer of a chunk that is about to be physically deleted, correct? TSan cannot know about this, so I would rather suppress that warning (with a comment here) instead of locking the chunks. For fun, you could run the TPC-H in multi-threaded mode on a beefy machine and you will see that this largely kills the performance.

src/lib/storage/chunk.cpp
```diff
@@ -202,6 +202,7 @@ void Chunk::increase_invalid_row_count(const uint64_t count) const { _invalid_ro
 
 void Chunk::set_cleanup_commit_id(const CommitID cleanup_commit_id) {
   DebugAssert(!_cleanup_commit_id, "Cleanup commit ID can only be set once.");
+  std::unique_lock lock(_mutex_cleanup_commit_id);
```
mrks: We can't run into any issues here, can we? The 64-bit value is guaranteed not to be read partially* (note that this is different from it being read atomically). A reader might see the old or the new version and either is fine. TSan doesn't know this, so using a suppression should be fine. Otherwise, at least use an atomic, which is significantly cheaper than a lock.

*) At least on Intel architectures.

src/plugins/mvcc_delete_plugin.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include "utils/abstract_plugin.hpp"
+#include "utils/pausable_loop_thread.hpp"
+#include "utils/singleton.hpp"
+
+namespace opossum {
+
+/*
+ * One disadvantage of insert only databases like Hyrise is the accumulation of invalidated
+ * rows, which have to be removed from the final result for every transaction.
+ * This plugin deletes chunks with large numbers of invalidated rows. Thus, it keeps the
```
mrks: ... and reinserts them at the end of the table. These rows are either visible at their original position (for old transactions) or their new position (for new transactions)

src/lib/storage/table.cpp
mrzzzrm: I have recently watched [Herb Sutter's atomic weapons talks](https://www.youtube.com/watch?v=A8eCGOqgvH4) and I feel more insecure about my MT knowledge than ever. My takeaway is that there is no such thing as "our logic" because sequential consistency is limited to one thread - it only applies to multi-threading if you synchronize properly. Otherwise, here be dragons, or something.

I won't give my blessing to anything that doesn't synchronize via atomic or mutex and to nothing that relies on the impartiality of 64 bit reads, simply because I don't know enough about the topic. 

It seems unwise to me to silence TSAN on such a trivial matter, I mean this is just multiple threads rw-accessing the same vector. If TSAN detects a race here, there probably is one.

mrks: The internet is reeeeeaaaaally sloooooow here, so I can't properly work with SSH. Maybe you could start the TPC-H MT benchmark for ac06e5d^ and ac06e5d and paste it here?

mrks: Well it doesn't properly kill it, but I still think this hit is avoidable:

```
./hyriseBenchmarkTPCH --scheduler --clients 100 -o after.json -q 1,2,3,4,5,6,7,8,9,10

+-----------+---------------+-------+---------------+-------+--------+---------------------------------+
| Benchmark | prev. iter/s  | runs  | new iter/s    | runs  | change | p-value (significant if <0.001) |
+-----------+---------------+-------+---------------+-------+--------+---------------------------------+
| TPC-H 1   | 16.0326118469 | 962   | 15.3656196594 | 922   | -4%    |     (run time too short) 0.6157 |
| TPC-H 2   | 11.38326931   | 683   | 9.93315410614 | 596   | -13%   |     (run time too short) 0.1662 |
| TPC-H 3   | 181.402236938 | 10000 | 174.466079712 | 10000 | -4%    |     (run time too short) 0.0083 |
| TPC-H 4   | 97.333114624  | 5840  | 94.1878051758 | 5652  | -3%    |     (run time too short) 0.0000 |
| TPC-H 5   | 144.038482666 | 8643  | 143.321044922 | 8600  | -0%    |     (run time too short) 0.0876 |
| TPC-H 6   | 467.382324219 | 10001 | 410.694335938 | 10003 | -12%   |     (run time too short) 0.0000 |
| TPC-H 7   | 49.525642395  | 2972  | 48.1816558838 | 2891  | -3%    |     (run time too short) 0.3558 |
| TPC-H 8   | 16.5375728607 | 993   | 16.4296684265 | 986   | -1%    |                          0.6665 |
| TPC-H 9   | 68.806427002  | 4129  | 69.1422424316 | 4149  | +0%    |     (run time too short) 0.5791 |
| TPC-H 10  | 104.351455688 | 6262  | 103.053039551 | 6184  | -1%    |     (run time too short) 0.6114 |
| average   |               |       |               |       | -4%    |                                 |
+-----------+---------------+-------+---------------+-------+--------+---------------------------------+
```

I'll give this some thought in the next days.

mrks: Not getting through with this as fast as I hoped for. Let's bring the version with locks in and look at this separately.

mrks:
There seem to remain some threading issues. Let's get those worked out, merge this, and I'll have a look at how to avoid those locks later.

mrks:
Some commits back, a merge ran astray and Github showed ~20k modified rows. That was when this PR was at commit ae6491f5cf7313be4eb40337e9a0d2c0601c462a. I reset it to origin/master, squashed ae6491f5cf7313be4eb40337e9a0d2c0601c462a and force-pushed it.

This is to avoid merge conflicts down the roads when files that are not part of this PR as well as to keep the history of those files clean.

src/lib/CMakeLists.txt
```diff
@@ -682,30 +682,37 @@ add_custom_command (
...
+
+# -force_load/--whole-archive are necessary to include all symbols so that dynamically loaded plugins can use them
+if (APPLE)
+    set_property(TARGET hyrise PROPERTY INTERFACE_LINK_LIBRARIES -force_load $<TARGET_FILE:hyrise_impl>)
+else()
+    set_property(TARGET hyrise PROPERTY INTERFACE_LINK_LIBRARIES -Wl,--whole-archive,$<TARGET_FILE:hyrise_impl>,--no-whole-archive)
+endif()
+
+# Link against all libraries used by hyrise_impl - do this after INTERFACE_LINK_LIBRARIES is initially set so that it is not overwritten
+target_link_libraries(hyrise INTERFACE hyrise_impl)
```
mrks: @Bensk1 I had to play around with this a bit as `whole-archive` was not completely propagated to downstream targets (here hyriseTest). This is tested on OS X and Ubuntu, but I'd like to make you at least aware of it.

mrks:
blocked by #1603

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+    const auto max_chunk_id = static_cast<ChunkID>(table->chunk_count() - 1);
+    for (auto chunk_id = ChunkID{0}; chunk_id < max_chunk_id; chunk_id++) {
+      const auto& chunk = table->get_chunk(chunk_id);
+      if (chunk && !chunk->get_cleanup_commit_id()) {
+        // Calculate metric 1 – Chunk invalidation level
+        const double invalidated_rows_ratio = static_cast<double>(chunk->invalid_row_count()) / chunk->size();
+        const bool criterion1 = _DELETE_THRESHOLD_PERCENTAGE_INVALIDATED_ROWS <= invalidated_rows_ratio;
+
+        if (!criterion1) {
+          return;
```
mrks: This fails for multiple tables

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        // Calculate metric 2 – Chunk Hotness
+        const CommitID lowest_end_commit_id =
+            *std::min_element(std::begin(chunk->mvcc_data()->end_cids), std::end(chunk->mvcc_data()->end_cids));
+        const CommitID commit_id_diff = TransactionManager::get().last_commit_id() - lowest_end_commit_id;
+        auto max_commit_id_diff = static_cast<CommitID>(table->max_chunk_size() * _DELETE_THRESHOLD_COMMIT_DIFF_FACTOR);
```
mrks: Where does max_chunk_size come from here?

mrks:
Could you please review the changes made in d3d0978? I hope that the new system test is more comprehensive.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,184 @@
...
+void MvccDeletePlugin::_delete_chunk_physically(const std::shared_ptr<Table>& table, const ChunkID chunk_id) {
+  const auto& chunk = table->get_chunk(chunk_id);
+
+  /**
+   * The chunk should be referenced only by the plugin in methods _delete_chunk_physically()
+   * & _physical_delete_loop() and the table's chunk vector.
+   */
+  Assert(chunk.use_count() == 3,
+         "At this point, the chunk should be referenced only by the plugin in methods _delete_chunk_physically() "
+         "& _physical_delete_loop() and the table's chunk vector.");
```
mrks: What if a different thread just now looked at the chunk, made its copy, and has not yet decided that it doesn't have to process it as the cleanup commit id is older than its snapshot?

engelfa: Yeah, that's a problem. So, we really need a mutex? It can also fail if get_chunk already returns a nullptr when the chunk is logically deleted.

mrks: I think we are talking about different things. I am just saying that the use_count might be higher than 3. Is that really a problem though?

src/test/plugins/mvcc_delete_plugin_system_test.cpp
```diff
@@ -0,0 +1,208 @@
...
+ public:
+  // Create table with three (INITIAL_CHUNK_COUNT) chunks of CHUNK_SIZE rows each and increasing values.
+  void SetUp() override {
+    TableColumnDefinitions column_definitions;
+    column_definitions.emplace_back("number", DataType::Int);
+
+    _table = std::make_shared<Table>(column_definitions, TableType::Data, CHUNK_SIZE, UseMvcc::Yes);
+
+    auto begin_value = 0;
+    for (auto segment_id = size_t{0}; segment_id < INITIAL_CHUNK_COUNT; ++segment_id) {
```
engelfa:
```suggestion
    for (auto chunk_id = size_t{0}; chunk_id < INITIAL_CHUNK_COUNT; ++chunk_id) {
```

src/plugins/mvcc_delete_plugin.cpp
julianmenzler: Such a thread should not take ownership of the chunk. Saving a `chunk_id` is the better approach.

julianmenzler:
Alright, CI has finally passed. Any other change requests? 
If not, let's merge this! It became quite a pain merging in `hyrise/master` again and again. 😅

julianmenzler:
Well, quite interesting: Adding text to `README.md` leads to this: 

```
[----------] 1 test from MvccDeletePluginSystemTest
[ RUN      ] MvccDeletePluginSystemTest.CheckPlugin
../../../src/test/plugins/mvcc_delete_plugin_system_test.cpp:228: Failure
Expected: (attempts_remaining) > (-1), actual: -1 vs -1
../../../src/test/plugins/mvcc_delete_plugin_system_test.cpp:243: Failure
Expected: (attempts_remaining) > (-1), actual: -1 vs -1
[  FAILED  ] MvccDeletePluginSystemTest.CheckPlugin (14136 ms)
[----------] 1 test from MvccDeletePluginSystemTest (14136 ms total)
```

🤔

julianmenzler:
```
[ RUN      ] MvccDeletePluginSystemTest.CheckPlugin
script returned exit code 139
```

Looks like we still have some serious multithreading issues. I'll fix #1686 first. Then I will have a look at this again.

scripts/cpplint.py
```diff
@@ -366,6 +366,7 @@
     'ratio',
     'regex',
     'scoped_allocator',
+    'shared_mutex',
```
julianmenzler: I'll remove this.

src/lib/utils/pausable_loop_thread.hpp
```diff
@@ -15,7 +15,7 @@ namespace opossum {
 // paused state.
 struct PausableLoopThread {
  public:
-  explicit PausableLoopThread(std::chrono::milliseconds loop_sleep_time, const std::function<void(size_t)>& loop_func);
```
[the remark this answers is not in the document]
mrks: We mostly use explicit to avoid implicit conversions. If the constructor has multiple arguments, implicit conversions will not happen anyway.

julianmenzler:
Currently, we have some Linker errors regarding 8-byte-atomics:
```
/usr/bin/ld: /var/lib/jenkins/workspace/hyrise_hyrise_PR-1493@2/clang-debug/src/lib/libhyrise_impl.a(chunk.cpp.o): in function `std::atomic<std::optional<unsigned int> >::load(std::memory_order) const':
/usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/atomic:250: undefined reference to `__atomic_load_8'
/usr/bin/ld: /var/lib/jenkins/workspace/hyrise_hyrise_PR-1493@2/clang-debug/src/lib/libhyrise_impl.a(chunk.cpp.o): in function `std::atomic<std::optional<unsigned int> >::store(std::optional<unsigned int>, std::memory_order)':
/usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/atomic:239: undefined reference to `__atomic_store_8'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
It seems, these errors are limited to the CI platform only. On Nemea and my Mac everything compiles seemlessly.

I did some research and found out that  atomics are implemented using lockless-hardware instructions. However, not every platform does have support for this.
In this case, we have to link against `libatomics` to use a different lock-based approach.

I could hack `std::atomic<std::optional<CommitID>>` to something more simple that is supported by the CI platform.  (4-byte atomics have passed the CI recently).

We might want to use 8-byte atomics in the future though. So I think it makes sense to support this edge-case. I found [this CMake module](https://github.com/llvm-mirror/llvm/blob/master/cmake/modules/CheckAtomic.cmake). Shall I open an issue for this or do we want a solution for this now?

julianmenzler:
Discussed this with @mrks. For this PR, we do not implement `libatomic`. Instead, I'll omit the 64bit-atomic using an alternate implementation.

mrks:
Just tested this on TPC-C. We have at least a 10% performance increase, most likely more once the fundamental problems in the TPC-C implementation are resolved. I played a bit with the two parameters and found that both are reasonably chosen. The second criterion is vital - when I removed it, conflicts flew everywhere.

Two remaining things that I found on the way:

1) The `use_count` assert was triggered during some of the deletes. As that was flaky anyway, I'd simply remove it.
2) We should include failed (i.e., rolled back) inserts in the invalid row count metric (and add a test for that).

src/lib/storage/chunk.hpp
```diff
@@ -190,7 +190,9 @@ class Chunk : private Noncopyable {
   bool _is_mutable = true;
   std::optional<std::pair<ColumnID, OrderByMode>> _ordered_by;
   mutable std::atomic_uint64_t _invalid_row_count = 0;
-  std::atomic<std::optional<CommitID>> _cleanup_commit_id{std::nullopt};
+
+  // Default value of zero means "not set"
+  std::atomic_uint32_t _cleanup_commit_id{0};
```
mrks: Add a static assert that this and the base type of commit ID match?

src/test/plugins/mvcc_delete_plugin_system_test.cpp
```diff
@@ -0,0 +1,250 @@
...
+       ++transaction_idx) {
+    TransactionManager::get().new_transaction_context()->commit();
+  }
+
+  // (9) Wait for the MvccDeletePlugin to delete chunk three logically
+  {
+    auto attempts_remaining = 10;
+    while (attempts_remaining--) {
+      // The second chunk should have been logically deleted by now
+      if (_table->get_chunk(ChunkID{2})->get_cleanup_commit_id()) break;
```
mrks: @julianmenzler Who guarantees that chunk 2 is not physically deleted as well? There should be a transaction that still to which the chunk is still visible. After step 9, we commit that transaction, so that the physical delete can run.

julianmenzler: I think you are right. The physical delete might happen too early. We have no guarantee to see the cleanup-commit-id of the logical delete. 
I'll remove the logical-delete-checks, so that we only check for the end result, the physical delete.

mrks: No need to remove it, just hold a tx

julianmenzler: I just saw our "blocker condition" for the first physical delete. Therefore, we should be able to check for the cleanup-commit-id of the first logical delete. But the second logical delete cannot be checked.

julianmenzler:
I don't trust the CI yet because I still have some ThreadSanatizer issues locally.

src/test/plugins/mvcc_delete_plugin_system_test.cpp
mrks: Why not? If we create a transaction before step 10 and hold it until we have verified the logical delete, wouldn't the physical delete be blocked?

mrks:
> I don't trust the CI yet because I still have some ThreadSanatizer issues locally.
Are you passing the suppressions-file? Could you post a gist with the errors?

julianmenzler:
> > I don't trust the CI yet because I still have some ThreadSanatizer issues locally.
> 
> Are you passing the suppressions-file? Could you post a gist with the errors?

Oh, good hint! I just passed the file and the errors were gone. Nevertheless, I restarted the CI to make sure.

mrks:
This will probably not merge well with the current master (which got rid of `StorageManager::get()` etc.). Could you update this branch, please?

src/test/plugins/mvcc_delete_plugin_system_test.cpp
mrks: My last comment still appears unaddressed.

julianmenzler:
Sure, I'll finish this today.

src/test/plugins/mvcc_delete_plugin_system_test.cpp
julianmenzler: Now it is.

mrks:
Screenshot 2019-08-19 at 18 36 07

Really excited to finally get this in :)

mrks:
The lamentable decrease in the coverage is because the following files were previously ignored:

```
server/server.cpp
utils/null_streambuf.cpp
utils/pausable_loop_thread.cpp
visualization/abstract_visualizer.hpp
visualization/join_graph_visualizer.cpp
visualization/lqp_visualizer.cpp
visualization/pqp_visualizer.cpp
visualization/viz_record_layout.cpp
```

Those "add" 520 uncovered lines.

mrks:
This sounds like the thing that tsan could not possibly understand. Did you modify the tsan-ignore file? I'm not seeing any changes there.

julianmenzler:
> This sounds like the thing that tsan could not possibly understand. Did you modify the tsan-ignore file? I'm not seeing any changes there.



> This sounds like the thing that tsan could not possibly understand. Did you modify the tsan-ignore file? I'm not seeing any changes there.

I just added to `.tsan-ignore.txt`, not before.

mrks:
> Is there anything I can do in terms of code coverage?

Nah, we previously just lied about it. I don't think we will write tests for the visualization.

resources/.tsan-ignore.txt
```diff
@@ -9,4 +9,8 @@ race:shared_count
 
 # Boost is not built with tsan support, either
 race:boost::detail::thread_data_base
-race:thread_data
\ No newline at end of file
+race:thread_data
+
+# Hyrise is able to handle concurrent access to MVCC-data correctly.
+# However, the MvccDeletePluginSystemTest lead to several TSan warnings. (#1493)
```
mrks: > Hyrise is able to handle concurrent access to MVCC-data correctly.

Even if we could be sure that this was true, it would be a really bad reason to suppress a warning. The comment should explain what the suspected data race looks like, why it is actually thread-safe and why tsan still reports a false positive. Maybe we should also extract `end_cids[row_id.chunk_offset] = cid;` into its own method so that we can suppress just that.

src/test/plugins/mvcc_delete_plugin_system_test.cpp
```diff
@@ -158,7 +158,7 @@ TEST_F(MvccDeletePluginSystemTest, CheckPlugin) {
 
   // (5) Wait for the MvccDeletePlugin to delete chunk 2 logically
   {
-    auto attempts_remaining = 10;
+    auto attempts_remaining = 20;
```
mrks: Put a constant somewhere?

resources/.tsan-ignore.txt
julianmenzler: A hacky first try, I agree. If we get TSan under control with this, I'll try to document it better.

mrks: Nope, the idea is good

src/test/plugins/mvcc_delete_plugin_system_test.cpp
mrks: Maybe simply skip this test with asan?

julianmenzler: For now, raising `attempts_remaining` to 100 should help. If we ever run into this issue again, skipping the test would be our best option.

julianmenzler:
The additions to `tsan_ignore.txt` in this PR are redundant to #1909 – at least they become irrelevant. So I suggest we merge #1909 first and this PR follows?

mrks:
Sounds good. However, there seem to be some issues introduced with the latest merges that are independent of the tsan issue.

julianmenzler:
> Sounds good. However, there seem to be some issues introduced with the latest merges that are independent of the tsan issue.

Correct, this is next.

mrks:
@Bensk1 touched the invalid_row_count a couple of days ago. If you run into any issues there, maybe you two should have a chat.

julianmenzler:
Thanks, now everything is fixed. The finalization of chunks broke some tests, which I was not aware about. 
Also talked to @j-tr, whose tests are also affected by this change.

mrks:
CI is still unhappy :/

julianmenzler:
> CI is still unhappy :/

Yes :/ The system test needs an overhaul as well. I'm on it.

mrks:
@julianmenzler - I had to fix a couple of things that came up in the stress test. Please have a look. The plugin has been running for a good hour by now, so I am reasonably confident that we can get this in once you greenlight my changes.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -0,0 +1,171 @@
...
+  update->set_transaction_context(transaction_context);
+  update->execute();
+
+  // Check for success
+  if (update->execute_failed()) {
+    // Transaction conflict. Usually, the OperatorTask would call rollback, but as we executed Update directly, that is
+    // our job.
+    transaction_context->rollback();
+    return false;
+  }
```
mrks: I couldn't come up with a good way to test this. Any ideas?

julianmenzler: Ouch! Was not on my radar that we need an OperatorTask for Auto-Rollback in case of failures. 

With the current implementation, there is no way for testing transaction conflicts. 
An interface change might help though: We should create the transaction context outside of `_try_logical_delete(..)` and pass it in. 
I'll try that.

julianmenzler: Done, works!

mrks: Good idea, thanks :)

julianmenzler:
Oh what happened here?
```
[----------] 1914 tests from SQLiteTestRunnerEncodings/SQLiteTestRunner
[ RUN      ] SQLiteTestRunnerEncodings/SQLiteTestRunner.CompareToSQLite/876
script returned exit code 139
```

Bouncner:
> decreased by 1.81 compared to master and is now 93.07%

😲

mrks:
On the segfault:
```
Repeating all tests (iteration 2) . . .

Note: Randomizing tests' orders with a seed of 66055 .
```

On the coverage: That's mentioned somewhere - not really a change in this PR. For some reason, the entire `visualization/` folder was invisible to the coverage tool before

mrks:
Even if CI is green now, I'd like to check the setup with `--gtest_seed=66055` a couple of times before proceeding. As the plugin tests ran just before the segfault, this does not look like a coincidence to me.

mrks:
Running the system tests both in debug and release modes over night - if nothing comes up, I am not sure how to investigate that hickup.

julianmenzler:
> Running the system tests both in debug and release modes over night - if nothing comes up, I am not sure how to investigate that hickup.

Alright, so how did that test go?

mrks:
99% sure that this is #1643

julianmenzler:
Finally, merged! 🥳

Bouncner:
> 99% sure that this is #1643

The current state of the SMJ really annoys me, but I have quite a lot on the table right now. Ping me, if it is of great help to approach this problem timely. Otherwise, it's something the students of the sorting group should address.

PR #1496, opened 2019-02-18
JIT Improve JitExpression Evaluation

FabianWiebe:
This PR improves the evaluation of `JitExpressions` by reducing the number of accesses to the runtime tuple in the `JitRuntimeContext` in three ways:

1. `JitExpressions` can now directly return their result value instead of storing it in the runtime tuple.
This is achieved by directly returning a `JitValue<ResultValueType>` struct from a `compute<ResultValueType>()` function. The result type of a JitExpression is known by the caller so that it can call the compute function with the correct function template parameter.
2. A `JitExpression` can now directly store a literal value. Therefore, the expression does not need to read the value from the runtime tuple as it does not change during execution.
3. The `JitFilter` operator can now directly compute its filter `JitExpression` and does not depend on a prior `JitCompute` operator for the computation of the `JitExpression`. This prevents that the `JitCompute` and `JitFilter` operators needlessly store and read the result value from the runtime tuple.

The PR also improves the evaluation of expressions with non-string operands by using the old lambda functions for non-string operands. This allows a better specialization of these expressions.


src/lib/operators/jit_operator/jit_operations.cpp
```diff
@@ -84,14 +127,20 @@ bool jit_not_like(const std::string& a, const std::string& b) {
   return !std::regex_match(a, regex);
 }
 
-void jit_is_null(const JitTupleValue& lhs, const JitTupleValue& result, JitRuntimeContext& context) {
-  result.set_is_null(false, context);
-  result.set<bool>(lhs.is_null(context), context);
+JitValue<bool> jit_is_null(const JitExpression& left_side, JitRuntimeContext& context) {
+  switch (left_side.result().data_type()) {
+    BOOST_PP_SEQ_FOR_EACH_PRODUCT(JIT_IS_NULL_CASE, (JIT_DATA_TYPE_INFO))
```
mrks: Could you add a comment on why you need this <s>ugly</s> beautiful macro-based meta programming?

src/lib/operators/jit_operator/jit_operations.cpp
```diff
@@ -25,49 +25,92 @@ namespace opossum {
...
               "invalid type for jit operation and");
 
-  // three-valued logic AND
-  if (lhs.is_null(context)) {
-    result.set<bool>(false, context);
-    result.set_is_null(rhs.is_null(context) || rhs.get<bool>(context), context);
-  } else {
-    result.set<bool>(lhs.get<bool>(context) && rhs.get<bool>(context), context);
-    result.set_is_null(lhs.get<bool>(context) && rhs.is_null(context), context);
+  const auto left_result = left_side.compute<bool>(context);
```
mrks: Naming: left_result vs lhs? Also, how can you access `result` before you call `compute`?

src/lib/operators/jit_operator/jit_operations.cpp
```diff
@@ -25,49 +25,92 @@ namespace opossum {
...
+  // Computation of right hand side can be pruned if left result is false and not null
+  if (!left_result.value) {                            // Left result is false
+    if (!lhs.is_nullable() || !left_result.is_null) {  // Left result is not null
+      return {false, false};
```
mrks: The constructor is not intuitive. I would have expected NULL to be the second parameter, but either way, it's not good to read. Do you run into specialization issues even with `std::optional`?

src/lib/operators/jit_operator/jit_operations.cpp
```diff
@@ -25,49 +25,92 @@ namespace opossum {
...
+  DebugAssert((left_side.result().data_type() == DataType::Bool || left_side.result().data_type() == DataType::Int),
+              "invalid type for jit operation not");
+  const auto value = left_side.compute<bool>(context);
+  return {value.is_null, !value.value};
 }
 
-void jit_and(const JitTupleValue& lhs, const JitTupleValue& rhs, const JitTupleValue& result,
-             JitRuntimeContext& context) {
-  // If the input values are computed by non-jit operators, their data type is int but they can be read as bool values.
+JitValue<bool> jit_and(const JitExpression& left_side, const JitExpression& right_side, JitRuntimeContext& context) {
```
mrks: Please include some comments saying that this uses short-circuit evaluation.

src/lib/operators/jit_operator/jit_operations.hpp
```diff
@@ -71,14 +73,7 @@ namespace opossum {
...
 /* Arithmetic operators */
-struct JitAddition {
-  template <typename T1, typename T2,
-            typename = typename std::enable_if_t<std::is_scalar_v<T1> == std::is_scalar_v<T2>>>
-  auto operator()(const T1 a, const T2 b) const {
-    return a + b;
-  }
-};
-const JitAddition jit_addition{};
+const auto jit_addition = [](const auto a, const auto b) -> decltype(a + b) { return a + b; };
```
mrks: Did we need to explicitly state the return type here?

src/lib/operators/jit_operator/jit_operations.hpp
```diff
@@ -223,13 +241,14 @@ DataType jit_compute_type(const T& op_func, const DataType lhs, const DataType r
...
 }
 
-void jit_not(const JitTupleValue& lhs, const JitTupleValue& result, JitRuntimeContext& context);
-void jit_and(const JitTupleValue& lhs, const JitTupleValue& rhs, const JitTupleValue& result,
-             JitRuntimeContext& context);
-void jit_or(const JitTupleValue& lhs, const JitTupleValue& rhs, const JitTupleValue& result,
-            JitRuntimeContext& context);
-void jit_is_null(const JitTupleValue& lhs, const JitTupleValue& result, JitRuntimeContext& context);
-void jit_is_not_null(const JitTupleValue& lhs, const JitTupleValue& result, JitRuntimeContext& context);
+__attribute__((always_inline)) JitValue<bool> jit_not(const JitExpression& left_side, JitRuntimeContext& context);
```
mrks: Please add a comment on why you need to always inline this

src/lib/operators/jit_operator/operators/jit_expression.cpp
```diff
@@ -1,13 +1,70 @@
...
+#include <string>
 
 #include "../jit_constant_mappings.hpp"
 #include "../jit_operations.hpp"
+#include "../jit_types.hpp"
 
 namespace opossum {
 
+// Returns the enum value (e.g., DataType::Int, DataType::String) of a data type defined in the DATA_TYPE_INFO sequence
+#define JIT_GET_ENUM_VALUE(index, s) APPEND_ENUM_NAMESPACE(_, _, BOOST_PP_TUPLE_ELEM(3, 1, BOOST_PP_SEQ_ELEM(index, s)))
```
mrks: This looks familiar. Don't we have this in jit_operations.cpp, too?

src/lib/operators/jit_operator/operators/jit_expression.cpp
```diff
@@ -1,13 +1,70 @@
...
+
+// Returns the data type (e.g., int32_t, std::string) of a data type defined in the DATA_TYPE_INFO sequence
+#define JIT_GET_DATA_TYPE(index, s) BOOST_PP_TUPLE_ELEM(3, 0, BOOST_PP_SEQ_ELEM(index, s))
+
+// Call the compute function with the correct template parameter if compute is called without a template parameter and
+// store the result in the runtime tuple.
+#define JIT_COMPUTE_CASE(r, types)                                         \
```
mrks: This is also defined in jit_operations.hpp. Please clean up the macro use.

src/lib/operators/jit_operator/operators/jit_expression.cpp
```diff
@@ -1,13 +1,70 @@
...
+BOOST_PP_SEQ_FOR_EACH(JIT_VARIANT_GET, _, JIT_DATA_TYPE_INFO)
+BOOST_PP_SEQ_FOR_EACH(JIT_VARIANT_SET, _, JIT_DATA_TYPE_INFO)
+
 JitExpression::JitExpression(const JitTupleValue& tuple_value)
     : _expression_type{JitExpressionType::Column}, _result_value{tuple_value} {}
 
+JitExpression::JitExpression(const JitTupleValue& tuple_value, const AllTypeVariant& variant)
+    : _expression_type{JitExpressionType::Value}, _result_value{tuple_value} {
+  _variant.is_null = variant_is_null(variant);
+  if (!_variant.is_null) {
```
mrks: Guard clause?

src/lib/operators/jit_operator/operators/jit_expression.cpp
```diff
@@ -1,13 +1,70 @@
...
+    resolve_data_type(data_type_from_all_type_variant(variant), [&](const auto current_data_type_t) {
+      using CurrentType = typename decltype(current_data_type_t)::type;
+      _variant.set<CurrentType>(boost::get<CurrentType>(variant));
```
mrks: If you simply want to get the type of the variant, boost::apply_visitor might be better.

src/lib/operators/jit_operator/operators/jit_expression.cpp
```diff
@@ -1,13 +1,70 @@
...
+      // Non-jit operators store bool values as int values
+      if constexpr (std::is_same_v<CurrentType, Bool>) {
+        _variant.set<bool>(boost::get<CurrentType>(variant));
```
mrks: How can this be the case if CurrentType is the type from the variant and AllTypeVariants can't hold bools?

src/lib/operators/jit_operator/operators/jit_expression.cpp
```diff
@@ -25,6 +82,17 @@ JitExpression::JitExpression(const std::shared_ptr<const JitExpression>& left_ch
...
     return "x" + std::to_string(_result_value.tuple_index());
+  } else if (_expression_type == JitExpressionType::Value) {
+    std::stringstream str;
+    if (_result_value.data_type() != DataType::Null) {
+      resolve_data_type(_result_value.data_type(), [&](const auto current_data_type_t) {
+        using CurrentType = typename decltype(current_data_type_t)::type;
+        str << _variant.get<CurrentType>();
+      });
+    } else {
+      str << "null";
```
mrks: We use "NULL".

src/lib/operators/jit_operator/operators/jit_expression.cpp
```diff
@@ -25,6 +82,17 @@ JitExpression::JitExpression(const std::shared_ptr<const JitExpression>& left_ch
 std::string JitExpression::to_string() const {
   if (_expression_type == JitExpressionType::Column) {
     return "x" + std::to_string(_result_value.tuple_index());
+  } else if (_expression_type == JitExpressionType::Value) {
+    std::stringstream str;
+    if (_result_value.data_type() != DataType::Null) {
+      resolve_data_type(_result_value.data_type(), [&](const auto current_data_type_t) {
+        using CurrentType = typename decltype(current_data_type_t)::type;
+        str << _variant.get<CurrentType>();
```
mrks: Why can't you simply use the operator<< implementation of the all type variant?

src/lib/operators/jit_operator/operators/jit_expression.hpp
```diff
@@ -29,22 +54,36 @@ class JitExpression {
...
   const JitTupleValue& result() const { return _result_value; }
 
-  /* Triggers the (recursive) computation of the value represented by this expression.
-   * The result is not returned, but stored in the _result_value tuple value.
-   * The compute() function MUST be called before the result value in the runtime tuple can safely be accessed through
-   * the _result_value.
+  // The compute() and compute<ResultValueType>() functions trigger the (recursive) computation of the value
+  // represented by this expression.
+
+  /* The compute() function does not return the result, but stores it in the _result_value tuple value.
```
mrks: Maybe rename to `compute_and_store`? I don't like having two functions with the same name that behave fundamentally different and are only differentiated by their template argument.

src/lib/operators/jit_operator/operators/jit_expression.cpp
FabianWiebe: I removed `JIT_GET_ENUM_VALUE` and `JIT_GET_DATA_TYPE` macros from `jit_operations.hpp`, `jit_operations.cpp` and `jit_expressions.cpp` and added them to `jit_types.hpp` instead.

src/lib/operators/jit_operator/operators/jit_expression.cpp
FabianWiebe: Renamed `JIT_COMPUTE_CASE` macro to `JIT_EXPRESSION_COMPUTE_CASE` in this file.

src/lib/operators/jit_operator/operators/jit_expression.cpp
FabianWiebe: Condition removed during the replacement of `resolve_data_type(...` with `boost::apply_visitor(...`.

src/lib/operators/jit_operator/jit_operations.cpp
mrks: This has not really improved. Calling `data_type()` on something called `left_type` should result in some meta type, shouldn't it? Also, I still don't like how we call `result()` on something that has not been calculated yet.

Suggestion: `const auto left_type = left_side.result().data_type();` with a comment that while we don't have the actual result yet, we can already access the type information.

mrks:
Ah, I should have waited a couple of minutes. Regarding the open question, please see my comment above.

mrks:
Once that is changed and the lint error is fixed, this should be ready to go.

mrks:
Still the same problem: `result_value_type().data_type()` has "type" twice, so is it the type of a type? Also, if you need to call `data_type`, what is the object itself?

FabianWiebe:
> Still the same problem: `result_value_type().data_type()` has "type" twice, so is it the type of a type? 

No, it is the type of the value from the left/right operand. I added the word type to the function name during the first renaming to make clear that it does not return the actual value.
Should `result_value_type()` maybe be renamed to `result_tuple_value()`?

> Also, if you need to call `data_type`, what is the object itself?

Not sure if I understand the question correctly. The object returned from `result()`/`result_value_type()`/`result_tuple_value()` is a `JitTupleValue`, which contains all meta data for one tuple value.

mrks:
Let's take this offline. I'll be in the office by 9, or we do this tomorrow.

FabianWiebe:
I renamed `JitTupleValue` to `JitTupleEntry` and also renamed `JitHashmapValue` to `JitHashmapEntry` for consistency.

It caused a few more changes than I expected. :/

mrks:
Phew. I didn't expect that many changes either. This significantly cleans up the whole naming related to the tuple usage.

mrks:
>  ** Coverage ** — decreased by .92 compared to master and is now 91.30%

Is this on a current master?

FabianWiebe:
> > ** Coverage ** — decreased by .92 compared to master and is now 91.30%
> 
> Is this on a current master?

I merged the master a couple of days ago. I did another merge.

FabianWiebe:
The coverage still decreases:
> Coverage - decreased by .27 compared to master and is now 91.95%

The reason for this is that not every instantiation of a templated function is tested explicitely for every combination. I will look into what could be additionally tested.
I suggest to discuss this after the Fortnightly tomorrow.

To cover all paths in the `JitExpression::compute<T>()` function we need to test any combination of:
`expression type * left input type  * left input nullalbe * right input type * right input nullable * result type * result nullable`

mrks:
As far as I know, the percentage tracks source code lines, not instantiations.

If you could keep this decrease in the back of your mind, that would be great.

PR #1482, opened 2019-02-07
[DYOD18/19] Multi Predicate Join Support

Kolpa:
This PR adds support for the new Multi Predicate Join functionality.

This allows us to execute Joins with multiple predicates without creating an intermediate result table and calling an additional scan operator for each predicate by instead checking the additional predicates during the Probing Phase of the hash join. Additionally we also allow for Semi and Anti joins to be executed.

It modifies the HashJoin Operator to accept an additional parameter based on this Interface:
```c++
JoinHash::JoinHash(
  const shared_ptr<const AbstractOperator>& left,
  const shared_ptr<const AbstractOperator>& right,
  const JoinMode mode,
  const ColumnIDPair& column_ids,
  const PredicateCondition predicate_condition,
  const optional<size_t>& radix_bits,
  vector<JoinPredicate> additional_join_predicates = {}) {}
```

# Performance
Benchmarks related to the performance impact can be found [here](https://github.com/mweisgut/hyrise/wiki/Benchmark-results).

[JOB and TPC-H](https://gist.github.com/mrzzzrm/b55d39a1214acf9208a473e5d222a8bc)

# Currently Supported
- Multiple additional predicates for hash join
  - these predicates support the following conditions: 
`Equals GreaterThan GreaterThanEquals LessThan LessThanEquals NotEquals`
  - since it is based on the hash join the primary predicate has to be of type `Equals`
- Anti and Semi join support for the `HashJoin` operator with multiple predicates

# TODO
- Support more join types
  - Support for sort merge join
  - Support for nested loop join
- Evaluate mpj comparison function (is there a better way to do this?)

PR #1497, opened 2019-02-18
Add semi/anti join support in sort-merge join

### LINKED ISSUE #1499: Sort-Merge Join does not complain about unsupported join modes
In case the sort-merge join is used, TPC-H 4 with a semi join fails since the sort-merge join does not support the semi join, yet.

#1497 should fix this issue, but in case this PR will not be merge we should at least assert() that no unsupported modes are executed.
I consider this a bug (even though an unmodified Hyrise will always use the HashJoin for such queries) since I expect the sort-merge join to support all modes that it does not explicitly excludes via asserts or similar measures.

```
- Benchmarking Query TPC-H 4
- Running query with SQLite (1 s 938 ms).
- Comparing Hyrise and SQLite result tables
===================== Tables are not equal =====================
------------------------- Actual Result ------------------------
    o_orderpriority order_count
      string     long
<<<<<
   01-URGENT      279
   1  2-HIGH      284
   23-MEDIUM      250
   34-NOT SPECIFIED      329
   4   5-LOW      284
----------------------------------------------------------------

------------------------ Expected Result -----------------------
    o_orderpriority order_count
      string      int
<<<<<
   01-URGENT      109
   1  2-HIGH       99
   23-MEDIUM       90
   34-NOT SPECIFIED      116
   4   5-LOW      105
----------------------------------------------------------------
Type of error: Cell data mismatch
================================================================

Mismatched cells (row,column): (0,1) (1,1) (2,1) (3,1) (4,1)

- Verification failed (76 µs 184 ns)
- Running query with SQLite (1 s 746 ms).
- Comparing Hyrise and SQLite result tables
===================== Tables are not equal =====================
------------------------- Actual Result ------------------------
```

Bouncner:
This PR adds the ability to execute semi and anti joins in the sort merge join. Main motivation was an observation that came up during @SvenLehmann 's master thesis. In some cases, we have joins where the sides cannot be switched (e.g., certain outer joins or semi/anti) and in these cases the hash join performs poorly when the probe side is much smaller that the build side.
This limitation does not exist for the sort-merge join.

**On top of that, this PR includes some additional changes:** 
- a saner determination of cluster counts which improve performance significantly for small joins
- significantly improve performance by not merging pos lists in the end of the operator which saves runtime and further ensure that each output chunk is guaranteed to be sorted (exploitable by the current seminar groups), which was not the case before
- much more constness, again surprising performance gains
- some simplifications, e.g., removing the dictionary specialisations for the materialization phase (bucket sort is not faster than a final std::sort)
- make skipped tests more explicit by using `GTEST_SKIP()` over returns; we skipped join tests for most join implementations where columns types differed; no these tests are skipped and thus show that we lack support for certain joins

**Todo:**
- `src/lib/operators/join_sort_merge/radix_cluster_sort.hpp:158` does some ugly passing of vectors which I don't like but I am not able to improve that code. Any help is welcome here.

**Benchmarks:**

Here is a simple join benchmark for a `lineitem.l_orderkey` join on SF=1. The hash join is still the fastest join, but when the sides cannot be switched, the sort-merge join can be 2x faster.
This PR does not address this problem from the optimizer side.
```
-----------------------------------------------------------------------------------------------------------------
Benchmark                                                                          Time           CPU Iterations
-----------------------------------------------------------------------------------------------------------------
TPCHDataMicroBenchmarkFixture/BM_HashSemiProbeRelationSmaller_mean         516987105 ns  509959200 ns          1
TPCHDataMicroBenchmarkFixture/BM_HashSemiProbeRelationLarger_mean          234356279 ns  230752325 ns          4
TPCHDataMicroBenchmarkFixture/BM_SortMergeSemiProbeRelationSmaller_mean    254436540 ns  253390633 ns          3
TPCHDataMicroBenchmarkFixture/BM_SortMergeSemiProbeRelationLarger_mean     292223744 ns  291456767 ns          3
TPCHDataMicroBenchmarkFixture/BM_HashAntiProbeRelationSmaller_mean         411283087 ns  409554000 ns          1
TPCHDataMicroBenchmarkFixture/BM_HashAntiProbeRelationLarger_mean          167975552 ns  167341800 ns          4
TPCHDataMicroBenchmarkFixture/BM_SortMergeAntiProbeRelationSmaller_mean    265472312 ns  264762950 ns          2
TPCHDataMicroBenchmarkFixture/BM_SortMergeAntiProbeRelationLarger_mean     332750246 ns  329975550 ns          2
```

This is TPC-H SF=1 on nemea. On the left is the current master that always uses the sort-merge join. On the right is this PR.
```
+-----------+--------------------+------+--------------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s       | runs | new iter/s         | runs | change | p-value (significant if <0.001) |
+-----------+--------------------+------+--------------------+------+--------+---------------------------------+
| TPC-H 1   | 0.582531273365021  | 35   | 0.599179029464722  | 36   | +3%    |                          0.0002 |
| TPC-H 2   | 2.61675119400024   | 158  | 2.71726107597351   | 164  | +4%    |                          0.0000 |
| TPC-H 3   | 3.4964451789856    | 210  | 3.78179144859314   | 227  | +8%    |                          0.0000 |
| TPC-H 4   | 2.92605233192444   | 176  | 3.19961619377136   | 193  | +9%    |                          0.0000 |
| TPC-H 5   | 1.42285799980164   | 86   | 1.53936815261841   | 93   | +8%    |                          0.0000 |
| TPC-H 6   | 25.2097625732422   | 1513 | 25.3011417388916   | 1519 | +0%    |                          0.0063 |
| TPC-H 7   | 0.736438393592834  | 45   | 0.701482534408569  | 43   | -5%    |                          0.0000 |
| TPC-H 8   | 1.83221352100372   | 110  | 1.88769721984863   | 114  | +3%    |                          0.0000 |
| TPC-H 9   | 0.858280956745148  | 52   | 0.91465151309967   | 55   | +7%    |                          0.0000 |
| TPC-H 10  | 2.32912540435791   | 140  | 2.37053990364075   | 143  | +2%    |                          0.0000 |
| TPC-H 11  | 11.1193714141846   | 668  | 12.1195287704468   | 728  | +9%    |                          0.0000 |
| TPC-H 12  | 5.2208104133606    | 314  | 5.30656576156616   | 319  | +2%    |                          0.0000 |
| TPC-H 13  | 1.82378625869751   | 110  | 1.96606266498566   | 118  | +8%    |                          0.0000 |
| TPC-H 14  | 15.5952310562134   | 936  | 15.7704248428345   | 947  | +1%    |                          0.0000 |
| TPC-H 15  | 10.3300695419312   | 620  | 12.0110139846802   | 721  | +16%   |                          0.0000 |
| TPC-H 16  | 6.6278920173645    | 398  | 6.72749662399292   | 404  | +2%    |                          0.0000 |
| TPC-H 17  | 0.146589562296867  | 9    | 0.146263644099236  | 9    | -0%    |        (not enough runs) 0.9572 |
| TPC-H 18  | 1.37537503242493   | 83   | 1.37091338634491   | 83   | -0%    |                          0.5590 |
| TPC-H 19  | 5.32353925704956   | 320  | 5.38209962844849   | 323  | +1%    |                          0.0000 |
| TPC-H 20  | 0.0759870260953903 | 5    | 0.0779949873685837 | 5    | +3%    |        (not enough runs) 0.1984 |
| TPC-H 21  | 0.122050642967224  | 8    | 0.122789047658443  | 8    | +1%    |        (not enough runs) 0.8324 |
| TPC-H 22  | 9.37542819976807   | 563  | 11.0998754501343   | 666  | +18%   |                          0.0000 |
| average   |                    |      |                    |      | +4%    |                                 |
+-----------+--------------------+------+--------------------+------+--------+---------------------------------+
```

The changes mainly impact very small joins which do not a significant impact on TPC-H. Might be different for TPC-C and alike.

Fixes #1499.


Bouncner:
Fixes #1499.

Bouncner:
I won't find the time this week. Will reopen next week or so.

Bouncner:
Reopening to use the CI (having problems reproducing the CI failures locally).

src/lib/operators/join_sort_merge.cpp
```diff
@@ -160,15 +166,33 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
 
   /**
   * Determines the number of clusters to be used for the join.
-  * The number of clusters must be a power of two, i.e. 1, 2, 4, 8, 16...
-  * TODO(anyone): How should we determine the number of clusters?
+  * This task is not trivial as multiple aspects have to be considered: (i) the system's cache
+  * size, (ii) potential partitioning overhead, and (iii) the impact on successive operators.
+  * As of now, the cache can only be estimated. A size of 256k is used as this should be close 
+  * to the working machine of the students. For servers, however, this number might be vastly off.
+  * Aspects (i) and (ii) determine the performance of the join alone. Many partitions
```
mrks: How much of a difference does correctly "choosing" the cache size make? Adding a helper somewhere that tells us exactly that should not be more than a hundred lines.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -160,15 +166,33 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
+  * usually work well for sequential as well as parallel execution as the actual join phas is
```
mrks: Typo

src/lib/operators/join_sort_merge.cpp
```diff
@@ -160,15 +166,33 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
+  * faster, setting off the partitioning overhead.
+  * However, each cluster results in an output chunk. As such, to limit the potential
+  * negative impact of too many small chunks for the following operators, the cluster
+  * count is limited (to avoid expensive merges in the end).
+  * This is achieved by allowing the cluster count to grow linear up to 16 in every case, but
+  * adding only squareroot(clusters beyong 16) after that.
```
mrks: Typo. Also, I don't get the calculation here. Maybe an example would help?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -122,12 +127,12 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
   struct TableRange;
   struct TablePosition {
     TablePosition() = default;
-    TablePosition(size_t cluster, size_t index) : cluster{cluster}, index{index} {}
+    TablePosition(const size_t cluster, const size_t index) : cluster{cluster}, index{index} {}
```
mrks: Do you have a benchmark of just the constness? That would be interesting for me, but also a good motivator for the seminar.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -87,7 +91,8 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
         _right_column_id{right_column_id},
         _op{op},
         _mode{mode} {
-    _cluster_count = _determine_number_of_clusters();
+    _cluster_count = _determine_number_of_clusters(_sort_merge_join.input_table_left()->row_count(),
+                                                   _sort_merge_join.input_table_right()->row_count());
```
mrks: This is weirdly coupled with the non-Impl class? Any chance that should happen there and be passed into the Impl?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -256,26 +280,32 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
+    if (_mode != JoinMode::Semi && _mode != JoinMode::Anti) {
+      left_range.for_every_row_id(_sorted_left_table, [&](RowID left_row_id) {
+        right_range.for_every_row_id(_sorted_right_table, [&](RowID right_row_id) {
+          _emit_combination(output_cluster, left_row_id, right_row_id);
+        });
       });
-    });
+    } else {
+      left_range.for_every_row_id(_sorted_left_table, [&](RowID left_row_id) {
+        _output_pos_lists_left[output_cluster]->push_back(left_row_id);
```
mrks: emplace?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -355,13 +385,110 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
-    auto left_rest = TableRange(cluster_number, left_run_start, left_size);
+    const auto right_rest = TableRange(cluster_number, right_run_start, right_size);
+    const auto left_rest = TableRange(cluster_number, left_run_start, left_size);
     if (left_run_start < left_size) {
       _join_runs(left_rest, right_rest, CompareResult::Less);
     } else if (right_run_start < right_size) {
       _join_runs(left_rest, right_rest, CompareResult::Greater);
     }
+
+    // Short cut implementation for Anti joins. Implementing anti joins within the current sort-merge join
```
mrks: How can this be a short cut if the method ends right below/

src/lib/operators/join_sort_merge.cpp
Bouncner: It's dozens of percent for the sole join runtime. For the hash join, I did some testing last week and when we chose the "correct" size for nemea, it brings more then 10% for Q9.

src/lib/operators/join_sort_merge.cpp
Bouncner: Short cut in a sense that we did not implement the anti join within the "actual" sort-merge join, but as a postprocessing step where we negate the semi-join result. Happy to take better terms.

src/lib/operators/join_sort_merge.cpp
mrks: Do you want to open an issue for this (or do it right away)? We should introduce some `HardwareIdentifier` (name tbd) that exposes not only this, but also the number of cores as used by the scheduler and whatever else we might need.

src/lib/operators/join_sort_merge.cpp
Bouncner: Oh ... that might be something I'll test at the very end.

src/lib/operators/join_sort_merge.cpp
Bouncner: I'll dig up the number for the hash join again and open an issue.

Bouncner:
Closing again until both new anti-modes and SMJ adaptions are merged into the master (#1482).

PR #1498, opened 2019-02-19
LQP Utils: Correct order of expressions in the translation of an LQP to a boolean expression

FabianWiebe:
This PR corrects the LQP helper function, `lqp_subplan_to_boolean_expression()`.
The function can combine multiple Predicate and Union nodes in an LQP to one boolean expression and is used by the jit-aware LQP translator.

The old implementation reversed the order of the predicates which results in the last predicate of an LQP being evaluated first. This was caused by the fact that an LQP is evaluated from first to last node but `lqp_subplan_to_boolean_expression` traverses the LQP from last to first node.
The new implementation traverses the LQP in the same way but reverses the order in which predicates are added to the result expression.

### DIFF
--- a/src/lib/logical_query_plan/lqp_utils.cpp
+++ b/src/lib/logical_query_plan/lqp_utils.cpp
@@ -239,7 +239,7 @@ namespace {
  * translated expressions to the translation of its children nodes which enables to add the translated expression of
  * child node before its parent node to the output expression.
  */
-std::shared_ptr<AbstractExpression> _lqp_subplan_to_boolean_expression(
+std::shared_ptr<AbstractExpression> lqp_subplan_to_boolean_expression_impl(
     const std::shared_ptr<AbstractLQPNode>& begin, const std::optional<const std::shared_ptr<AbstractLQPNode>>& end,
     const std::optional<const std::shared_ptr<AbstractExpression>>& subsequent_expression) {
   if (end && begin == *end) return nullptr;
@@ -249,7 +249,7 @@ std::shared_ptr<AbstractExpression> _lqp_subplan_to_boolean_expression(
       const auto predicate_node = std::dynamic_pointer_cast<PredicateNode>(begin);
       const auto predicate = predicate_node->predicate();
       const auto expression = subsequent_expression ? and_(predicate, *subsequent_expression) : predicate;
-      const auto left_input_expression = _lqp_subplan_to_boolean_expression(begin->left_input(), end, expression);
+      const auto left_input_expression = lqp_subplan_to_boolean_expression_impl(begin->left_input(), end, expression);
       if (left_input_expression) {
         return left_input_expression;
       } else {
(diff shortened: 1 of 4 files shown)


src/test/logical_query_plan/jit_aware_lqp_translator_test.cpp
```diff
@@ -282,20 +310,20 @@ TEST_F(JitAwareLQPTranslatorTest, ConsecutivePredicatesGetTransformedToConjuncti
   ASSERT_EQ(a_gt_b->expression_type(), JitExpressionType::GreaterThan);
   ASSERT_EQ(a_gt_b->left_child()->expression_type(), JitExpressionType::Column);
   ASSERT_EQ(a_gt_b->right_child()->expression_type(), JitExpressionType::Column);
-  ASSERT_EQ(jit_read_tuples->find_input_column(a_gt_b->left_child()->result()), ColumnID{0});
-  ASSERT_EQ(jit_read_tuples->find_input_column(a_gt_b->right_child()->result()), ColumnID{1});
+  ASSERT_EQ(*jit_read_tuples->find_input_column(a_gt_b->left_child()->result()), ColumnID{0});
+  ASSERT_EQ(*jit_read_tuples->find_input_column(a_gt_b->right_child()->result()), ColumnID{1});
```
FabianWiebe: `find_input_column()` function returns an optional (`std::optional<ColumnID>`). I dereferenced the optional in all tests to make the comparison to the `ColumnID` explicit.

src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -231,45 +231,75 @@ std::set<std::string> lqp_find_modified_tables(const std::shared_ptr<AbstractLQP
...
+ */
+std::shared_ptr<AbstractExpression> _lqp_subplan_to_boolean_expression(
+    const std::shared_ptr<AbstractLQPNode>& begin, const std::optional<const std::shared_ptr<AbstractLQPNode>>& end,
+    const std::optional<const std::shared_ptr<AbstractExpression>>& subsequent_expression) {
   if (end && begin == *end) return nullptr;
 
-  static const auto whitelist =
-      std::set<LQPNodeType>{LQPNodeType::Projection, LQPNodeType::Sort, LQPNodeType::Validate, LQPNodeType::Limit};
-
-  if (whitelist.count(begin->type)) return lqp_subplan_to_boolean_expression(begin->left_input(), end);
```
FabianWiebe: A lookup in the `whitelist` is not necessary as these node types have a case in the switch statement below.

src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -231,45 +231,75 @@ std::set<std::string> lqp_find_modified_tables(const std::shared_ptr<AbstractLQP
...
+ *            LQP    --- lqp_subplan_to_boolean_expression(Predicate B) --->    boolean expression
+ *
+ *        Predicate B                                                    Predicate A       Predicate B
+ *             |                                                                \             /
+ *        Predicate A                                                             --- AND ---
+ *             |                                                                       |
+ *       Stored Table                                                          returned expression
+ *
+ * An lqp is evaluated from bottom to top. However, this function traverses an lqp from top to bottom. In the given lqp,
+ * Predicate B is reached first druing traversial, however, Predicate B can only be added to output expression once
```
mrks: during, traversal

src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -231,45 +231,75 @@ std::set<std::string> lqp_find_modified_tables(const std::shared_ptr<AbstractLQP
   return modified_tables;
 }
 
-std::shared_ptr<AbstractExpression> lqp_subplan_to_boolean_expression(
-    const std::shared_ptr<AbstractLQPNode>& begin, const std::optional<const std::shared_ptr<AbstractLQPNode>>& end) {
+namespace {
+/**
+ * Function creates a boolean expression from an lqp whilst preserving the same order of how expressions are evaluated.
```
mrzzzrm: I approve the change of this PR, but omg, this comment block is soooo long, if there is any important info in it noone will ever notice it. Can you maybe shorten it a bit?

FabianWiebe: I shortened it.

FabianWiebe:
Clang tidy does not like the prefix underscore in `_lqp_subplan_to_boolean_expression`.
I did not want that the public and private function share the same name and are only distinguishable by the number of parameters.
So I now renamed `_lqp_subplan_to_boolean_expression` to `lqp_subplan_to_boolean_expression_impl`.

Do you have a better alternative function name?
@mrks

mrks:
Works for me.

PR #1503, opened 2019-02-21
Compare Benchmarks: Handle mismatching names gracefully

mrks:
replaces #1479 

screenshot 2019-02-21 at 15 14 27


scripts/compare_benchmarks.py
```diff
@@ -52,9 +52,9 @@ def calculate_and_format_p_value(old, new):
 average_diff_sum = 0.0
 
 for old, new in zip(old_data['benchmarks'], new_data['benchmarks']):
+	name = old['name']
```
Bouncner: `old_name` and `new_name` in order to not have use `old['name']` and `name` simultaneously?

mrks: But `name` contains both if they are divergent!?

Bouncner: https://www.youtube.com/watch?v=8b4n1CDuR64

PR #1506, opened 2019-02-24
Introduce pmr_string

mrks:
This introduces polymorphic memory resources to all strings stored in tables. This is necessary whenever we want to store strings (not only in ValueSegments, but also in dictionaries etc.) somewhere else (specific NUMA resource, NVM, ...)

A couple of notes:
- This introduces a handful of unnecessary std::string <-> pmr_vector conversions, but none of these should be in a hot path.
- pmr_string violates our naming scheme. I'll be more than happy to change it to PmrString, but then we should change pmr_vector, too. I'd rather do that when the number of open PRs is lower.
- This PR is admittedly a whole lot of find-and-replace and compiler-error-driven development. I started by changing all_type_variant.hpp and worked my way through whatever broke. Finally, I grepped over src/lib to make sure no obvious cases remained.

Also, there are some minor changes that don't really warrant their own PR:
- Replace some uses of pmr_vector that were pointless. If you don't pass in a memory resource (explicitly or implicitly), using PMR doesn't give you anything.
- Moves boost_default_memory_resource.cpp and numa_memory_resource.cpp to a new folder "memory". That folder is not really interesting in the master yet, but I will fill it as part of my NVM work.
- Adds missing "Rule" to InsertLimitInExists.
- Sort src/lib/CMakeLists.txt
- Fixes build problems with clang on SKX

This costs a little performance due to the allocation indirection. Never mind TPC-H 15. It's the new TPC-H 2 and a bit flaky. :/

```
+-----------+-----------------+------+-----------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s    | runs | new iter/s      | runs | change | p-value (significant if <0.001) |
+-----------+-----------------+------+-----------------+------+--------+---------------------------------+
| TPC-H 1   | 0.506398439407  | 31   | 0.502039074898  | 31   | -1%    |                          0.2654 |
| TPC-H 2   | 8.47386264801   | 509  | 8.51299285889   | 511  | +0%    |                          0.2862 |
| TPC-H 3   | 6.93229055405   | 417  | 6.73408699036   | 405  | -3%    |                          0.0000 |
| TPC-H 4   | 2.11455535889   | 127  | 2.16181731224   | 130  | +2%    |                          0.0000 |
| TPC-H 5   | 4.47945642471   | 269  | 4.44505119324   | 267  | -1%    |                          0.1239 |
| TPC-H 6   | 25.1952991486   | 1512 | 25.4780387878   | 1529 | +1%    |                          0.0000 |
| TPC-H 7   | 1.33077466488   | 80   | 1.25796794891   | 76   | -5%    |                          0.0003 |
| TPC-H 8   | 5.7656583786    | 346  | 5.47066116333   | 329  | -5%    |                          0.0000 |
| TPC-H 9   | 1.59352552891   | 96   | 1.58227908611   | 95   | -1%    |                          0.3093 |
| TPC-H 10  | 2.66753149033   | 161  | 2.61108589172   | 157  | -2%    |                          0.0000 |
| TPC-H 11  | 33.8230400085   | 2030 | 34.1950187683   | 2052 | +1%    |                          0.0000 |
| TPC-H 12  | 7.26847410202   | 437  | 7.25145864487   | 436  | -0%    |                          0.4998 |
| TPC-H 13  | 2.08166623116   | 125  | 2.06420302391   | 124  | -1%    |                          0.2415 |
| TPC-H 14  | 17.1506462097   | 1030 | 16.6865158081   | 1002 | -3%    |                          0.0000 |
| TPC-H 15  | 11.9797992706   | 719  | 9.88319778442   | 594  | -18%   |                          0.0000 |
| TPC-H 16  | 7.11579847336   | 427  | 7.07193231583   | 425  | -1%    |                          0.0113 |
| TPC-H 17  | 0.152687713504  | 10   | 0.151234105229  | 10   | -1%    |                          0.7634 |
| TPC-H 18  | 1.8621762991    | 112  | 1.830042243     | 110  | -2%    |                          0.0561 |
| TPC-H 19  | 5.3910984993    | 324  | 5.2994389534    | 318  | -2%    |                          0.0000 |
| TPC-H 20  | 0.0751384943724 | 5    | 0.0768906325102 | 5    | +2%    |        (not enough runs) 0.1790 |
| TPC-H 21  | 0.12662781775   | 8    | 0.124699838459  | 8    | -2%    |        (not enough runs) 0.5766 |
| TPC-H 22  | 9.88627433777   | 594  | 9.93940353394   | 597  | +1%    |                          0.0008 |
| average   |                 |      |                 |      | -2%    |                                 |
+-----------+-----------------+------+-----------------+------+--------+---------------------------------+
```


src/lib/type_cast.hpp
```diff
@@ -51,23 +51,23 @@ inline __attribute__((always_inline)) T type_cast(const U& value) {
...
 // convert from T to string
 template <typename T, typename U,
-          typename = std::enable_if_t<std::is_same_v<std::decay_t<T>, std::string> &&
-                                      !std::is_same_v<std::decay_t<U>, std::string>>>
-inline __attribute__((always_inline)) std::string type_cast(const U& value) {
-  return std::to_string(value);
+          typename = std::enable_if_t<std::is_same_v<std::decay_t<T>, pmr_string> &&
+                                      !std::is_same_v<std::decay_t<U>, pmr_string>>>
+inline __attribute__((always_inline)) pmr_string type_cast(const U& value) {
+  return pmr_string{std::to_string(value)};
```
mrzzzrm: `std::to_string` has no pmr variant? saaaaad....

src/test/expression/like_matcher_test.cpp
```diff
@@ -12,7 +12,7 @@ class LikeMatcherTest : public ::testing::Test {
  public:
   bool match(const std::string& value, const std::string& pattern) const {
```
mrzzzrm: any special reason why these are not pm_strings?

mrzzzrm:
I have scrolled over the changes and not noticed anything suspicious. I feel unsure about two things:
* Why call this `pmr_string` if the `using` is intended as switch to quickly change the string type to be used? May just `String` is better, albeit looking like Java?

* What is the guideline on whether newly implemented stuff should use `pmr_string` or `std::string`? "Try it with std::string and if that fails to compile use pmr_string". Not implying this would be a bad guideline, just asking.

src/test/expression/like_matcher_test.cpp
mrks: Because it makes the test easier. Otherwise we would have to construct the pmr_string in every caller.

mrks:
>     * Why call this `pmr_string` if the `using` is intended as switch to quickly change the string type to be used? May just `String` is better, albeit looking like Java?

The "switch" capability was only an option in case the performance hit was too big. I would expect this to remain PMR.

>     * What is the guideline on whether newly implemented stuff should use `pmr_string` or `std::string`? "Try it with std::string and if that fails to compile use pmr_string". Not implying this would be a bad guideline, just asking.

What would you add to this comment?
```
// The string type that is used internally to store data. It's hard to draw the line between this and std::string.
// Generally, everything that is user-supplied data is a pmr_string. String that are built, e.g., for debugging, do not
// need to use PMR. This might sound complicated, but since the Hyrise data type registered in all_type_variant.hpp is
// pmr_string, the compiler will complain if you use std::string when you should use pmr_string.
```

mrzzzrm:
>  It's hard to draw the line between this and std::string.

May "It's hard to provide clear instructions of when to use pmr_string and when to use std::string"? 

> Generally, everything that is user-supplied

".... or column data"? I mean, probably "user-supplied" covers this, but for me it would be clearer this way.

mrks:
```
// The string type that is used internally to store data. It's hard to draw the line between this and std::string or
// give advice when to use what. Generally, everything that is user-supplied data (mostly, data stored in a table) is a
// pmr_string. Also, the string literals in SQL queries will get converted into a pmr_string (and then stored in an
// AllTypeVariant). This way, they can be compared to the pmr_string stored in the table. Strings that are built, e.g.,
// for debugging, do not need to use PMR. This might sound complicated, but since the Hyrise data type registered in
// all_type_variant.hpp is pmr_string, the compiler will complain if you use std::string when you should use pmr_string.
```

mrzzzrm:
One question, though: just switching to pmr_string for all strings is not a good idea because?

mrks:
There is no advantage to it if we don't pass in a memory source.

Also, some things come in as strings (SQL) and some interfaces are geared towards strings (e.g., stringstream).

mrzzzrm:
Time for pmr_stringstream then? ;)

Just speculating here, but a code base where two string types are used would probably confuse me, so using pmr_string everywhere (converting it immediately whenever we get an std::string from a library) might give us the advantage of not confusing people. Future PRs would also need to be checked for "using pmr_string where std::string would be sufficient" because if we don't we, again, build a confusing code base.

mrks:
That also has it's advantages, but it would require new people to forget about std::string. Following your argumentation, we would have to at least consider the same for vectors. I'd rather stick close to the standard.

The good thing here is that AllTypeVariant ensures that pmr is used whenever necessary. As we can see here, using it a little bit too much doesn't hurt.

mrzzzrm:
Okay, sufficiently convinced for now.

PR #1508, opened 2019-02-24
Retire NUMAPlacementManager

lanice:
Long awaited by @mrzzzrm, long delayed by me, this retires the NUMAPlacementManager and its associates (ChunkMigrationTask, MigrationPreparationTask), since they were never actually used.

For the case someone wants to use them in the future, we could probably add a snapshot to the [hyrise-archive](https://github.com/hyrise/hyrise-archive) repo, but I'm not sure if that would actually be necessary. Since this feature actually was in the main repo, we have it in the git history anyway, and one would only have to revert this PR to be able to work with it.


mrzzzrm:
Whole lotta love :heart: :heart: 

Can the ChunkAccessCounter and the ProxyChunk thingies go as well?

Bouncner:
Can you give us more information about what the manager did and whybwe don’t need it any longer?

Concerning counters, that something @bensk1 and me are very interested in. So please don’t remove them without further discussion.

lanice:
Ah yes you're right, I don't see any place where they are used anymore. Same with the ChunkAccessCounter if I'm not mistaken.

lanice:
> Can you give us more information about what the manager did and whybwe don’t need it any longer?

It was a while back (maybe half a year ago) where that was brought up. Basically, the NUMAPlacementManager was added a long time ago to enable dynamic migration of chunks between NUMA nodes, based on their access metrics. But, that

1. was never used in practice. The NUMAPlacementManager and all the other classes coming with that feature were just lingering around in the repo.
2. we had some weird (random) CI fails with some of tests that test the MigrationTask if I recall correctly.

Based on that we decided, that we will move all of that into the archive, and when (if) somebody wants to actually use it, then they can bring it back and/or re-implement it properly. Like I said, it was a while ago, but I think that's the general round-up.

lanice:
> Concerning counters, that something @Bensk1 and me are very interested in. So please don’t remove them without further discussion.

Same was true for the counters then, they were created and passed along everywhere, but never used. But you say you and/or @Bensk1 have started using them actually?

mrks:
Fully in favor of this. This was a reasonable first attempt, but as @lanice showed, our scalability issues lie somewhere else and need a different approach.

A future version of this should work as a plugin.

Concerning the counters, I believe that that approach is outsated as well. Now that we have the iterators, we should use those to get a better idea of how a chunk/segment is accessed. `on_with_iterators` tells us exactly how many rows are accessed at a minimal, constant cost.

Bouncner:
I am not aware how the current counters are implemented. But if they don't use the iterators am I fine with removing them.

src/benchmarklib/table_generator.cpp
```diff
@@ -112,11 +113,15 @@ std::shared_ptr<Table> TableGenerator::generate_table(
...
+#if HYRISE_NUMA_SUPPORT
+  auto node_id = 0;
+#endif
 
   for (ChunkID chunk_index{0}; chunk_index < num_chunks; ++chunk_index) {
 #if HYRISE_NUMA_SUPPORT
     if (numa_distribute_chunks) {
-      auto memory_resource = NUMAPlacementManager::get().get_next_memory_resource();
+      auto memory_resource = Topology::get().get_memory_resource(node_id);
+      node_id = (node_id + 1) % Topology::get().nodes().size();
```
mrks: Is round robin really what we want to do? Does this have any benefits?

lanice: > Does this have any benefits?

Questionable so far. But that is only applied if `numa_distribute_chunks=True`. Or do you want to get rid of that option entirely?

mrks: Is it ever set? That in itself has only limited value.

lanice: In the code in the repository, not that I know of. It was introduced as an option to play around with when using the table generator in a playground context.

mrks: I don't really mind either way. It's only a couple of lines in this file and has no external implications.

PR #1489, opened 2019-02-11
CardinalityEstimator utilizing Histograms

mrzzzrm:
Replaces the old cardinality estimation paths via `AbstractLQPNode::derive_statistics_from` with a dedicated `CardinalityEstimator` component. This cardinality estimator works primarily with Histograms spanning entire columns (as opposed to per-segment histograms). Inner Join and Scan operations are estimated more accurately than before. Scans on string columns are estimated reasonably well for the first time, thanks to @timzimmermann. This really helps in the Join Order Benchmark, which has many local predicates on string colums.

* Removes the `EqualWidthHistogram` and `EqualHeightHistogram` (for now), as we only actively use the `EqualDistinctCountHistogram`.
* The old cardinality estimation system had specialized code paths for Outer and Semi Joins - this functionality was removed, mainly, to reduce the amount of code to port over, test and maintain. Outer Joins are now estimated just like inner Joins and Semi/Anti Joins have a selectivity of 1. We don't "need" estimations for these join types anyway, right now. 
* We currently generate 5 (for < 10'000 rows) to 100 (for >= 20'000'000 rows) Bins. These numbers are rather arbitrarily chosen. More bins mean poorer optimization runtime (esp for Joins), but potentially better optimized query plans. As our error margins are measured in factors, it's probably not worth allocating more many bins and hope for better optimization results while sacrificing Optimizer runtime.

### Benchmark results
[Join Order Benchmark](https://gist.github.com/mrzzzrm/cd45479951e44403e3843bcc44732057) improves nicely, which imo is the best argument for the usefulness of this CardinalityEstimator.
[TPC-H with prepared statements](https://gist.github.com/mrzzzrm/4033bd2217148876366a1d65ecca5f15) never hits the Query Cache and therefore profits a bit from the reduced optimization times.

Fix #1153
Fix #1513 
Fix #1310

PR #1510, opened 2019-02-25
Finer grained benchmark metrics

mrzzzrm:
Since optimization duration may be (though shouldn't be) a significant portion of the query runtime, this PR makes the benchmarks write out more finer grained statistics for each query execution.

Part of #1489 as well, but I had to transfer these changes to the master anyway, so why not have a smaller PR for this ;)

```
{
          "parse_duration": 81168,
          "statements": [
            {
              "execution_duration": 1988159100,
              "lqp_translation_duration": 52451,
              "optimization_duration": 0,
              "query_plan_cache_hit": true,
              "sql_translation_duration": 0
            }
          ]
        },

```


Bouncner:
Does the JSON include some information about the unit? If not I’d vote for a `_ns` suffix.

I don’t get why parse duration is separated. Further, why is it called `statements`?
From the current naming I don’t get whether, e.g., execution does include optimization etc.

And how can SQL translation be 0?

mrzzzrm:
> Does the JSON include some information about the unit?
There is `"time_unit": "ns"`, though redundantly specified for each query.... :thinking: 

> I don’t get why parse duration is separated
It is only done once for a **sql-string/query**, but optimization etc is done for each **statement**

> Further, why is it called `statements`?
Each iteration of a **query** implies the optimization/translation/execution of multiple statements. See TPC-H 15

> From the current naming I don’t get whether, e.g., execution does include optimization etc.
Proposals? "plan_execution_duration"?

> And how can SQL translation be 0?
`"query_plan_cache_hit": true`


This was just meant as a small PR improving the reported benchmark results gradually, but I'd be happy to sit down and structure the JSON report properly, once and for all ;)

Bouncner:
> There is `"time_unit": "ns"`, though redundantly specified for each query.... 🤔

One should suffice, right?
 
> It is only done once for a **sql-string/query**, but optimization etc is done for each **statement**

Makes sense. Is the order of statements always clear or might it be necessary to include the statement?

> Proposals? "plan_execution_duration"?

Hmm, not sure. `_duration` might be unnecessary when the whole block is about time measurements. But I think `plan_execution` makes it definitely clearer.

> `"query_plan_cache_hit": true`

Makes sense. But I was wondering why there still is an LQP translation time then. Don't we cache string  ->> PQP plans as well?

mrzzzrm:
> One should suffice, right?

Yep, I changed all other timing information to ns and put a `{"time_unit": "ns"}` field into the `context`.


> Is the order of statements always clear

Always the same order as in the query.

mrzzzrm:
> Makes sense. But I was wondering why there still is an LQP translation time then. Don't we cache `SQL string ->> PQP` plans as well?

There is a PQP cache, I feel there might be a bug here, will try to fix it as part of this PR

mrzzzrm:
lqp_translation_duration not being 0 is due to the fact the this time includes the cache lookup duration, whereas sql_translation_duration does NOT include the cache lookup duration. I don't think we should just drop the cache lookup duration and will include the sql_translation_duration with it. Ack @Bouncner ?

src/benchmarklib/benchmark_runner.hpp
```diff
@@ -50,15 +52,18 @@ class BenchmarkRunner {
   void _warmup_query(const QueryID query_id);
 
   // Calls _schedule_query if the scheduler is active, otherwise calls _execute_query and returns no tasks
-  std::vector<std::shared_ptr<AbstractTask>> _schedule_or_execute_query(const QueryID query_id,
-                                                                        const std::function<void()>& done_callback);
+  std::vector<std::shared_ptr<AbstractTask>> _schedule_or_execute_query(
+      const QueryID query_id, const std::shared_ptr<SQLPipelineMetrics>& metrics,
```
mrks: Mmmh, passing in the metrics here is really not nice. Could these be a member of the class?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -349,13 +356,16 @@ std::vector<std::shared_ptr<AbstractTask>> BenchmarkRunner::_schedule_query(
     query_tasks.insert(query_tasks.end(), tasks.begin(), tasks.end());
   }
 
+  *metrics = pipeline.metrics();
```
mrks: On mobile, so I'm probably missing something. Why do we need to dereference metrics?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -244,19 +247,19 @@ void BenchmarkRunner::_benchmark_individual_queries() {
       if (currently_running_clients.load(std::memory_order_relaxed) < _config.clients) {
         currently_running_clients++;
 
+        const auto pipeline_metrics = std::make_shared<SQLPipelineMetrics>();
+
         // The on_query_done callback will be appended to the last Task of the query,
         // to measure its duration as well as signal that the query was finished
-        const auto query_run_begin = std::chrono::steady_clock::now();
-        auto on_query_done = [query_run_begin, &currently_running_clients, &result, &state]() {
+        auto on_query_done = [/*query_run_begin,*/ pipeline_metrics, &currently_running_clients, &result, &state]() {
```
mrks: ?

src/benchmarklib/abstract_table_generator.hpp
```diff
@@ -43,14 +53,9 @@ class AbstractTableGenerator {
    */
   virtual std::unordered_map<std::string, BenchmarkTableInfo> generate() = 0;
 
- protected:
-  /**
-   * The AbstractTableGenerator stores the whole BenchmarkConfig, yet not all fields in BenchmarkConfig influence
-   * table generation.
-   * This helper function creates a BenchmarkConfig with the fields that affect table generation
-   */
-  static std::shared_ptr<BenchmarkConfig> _create_minimal_benchmark_config(uint32_t chunk_size);
```
mrks: what happened to this?

src/benchmarklib/benchmark_runner.cpp
mrzzzrm: to assign by value. could assign by shared_ptr probably wouldn't make a difference.

src/benchmarklib/abstract_table_generator.hpp
mrzzzrm: Apparently noone was calling it, it wasn't implemented, so why keep it.

src/benchmarklib/benchmark_runner.hpp
mrzzzrm: We want to add them IFF and precisely when the query finished before its time was up. So we have to add them to the result in the `on_query_done` callback. I do not like it either, but haven't found a quick and clean way around it.

src/benchmarklib/benchmark_runner.cpp
mrks: This comment was marked as resolved, but I can still see `/*query_run_begin*/`. Did anything go wrong here?

src/benchmarklib/benchmark_runner.hpp
mrks: The callback already has the metrics, correct? Maybe pass the pipeline as a parameter (not a capture) into the callback?

src/benchmarklib/benchmark_runner.cpp
mrzzzrm: apparently I didn't push :(

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -320,28 +334,20 @@ void BenchmarkRunner::_warmup_query(const QueryID query_id) {
 }
 
 std::vector<std::shared_ptr<AbstractTask>> BenchmarkRunner::_schedule_or_execute_query(
-    const QueryID query_id, const std::function<void()>& done_callback) {
-  // Some queries (like TPC-H 15) require execution before we can call get_tasks() on the pipeline.
```
mrzzzrm: That comment seemed totally out of place here, so I removed it.

src/benchmarklib/benchmark_runner.hpp
mrzzzrm: I hope what I know changed it to is what you were looking for because the BenchmarkRunner is an otter pain to work on :(

src/benchmarklib/benchmark_runner.cpp
mrzzzrm: removed

src/benchmarklib/benchmark_runner.hpp
mrks: Argh, crap. I have only seen `BenchmarkRunner::_execute_query` and didn't realize that only there, `done_callback();` could have been extended with the pipeline as a parameter.

This new way feels a bit weird in that we are passing in the `query_id` and a pipeline which was already tailored to that particular query. Also, having three places in which we build the pipeline makes it easy to change one and forget about the others.

Please decide which way you like better - I am fine with both.

mrzzzrm: There is now `BenchmarkRunner::_build_sql_pipeline(QueryID query_id)` that should do it.

PR #1511, opened 2019-02-26
Add -Wshadow for GCC

mrks:
I bloody hate `-W` options. GCC doesn't include `-Wshadow` in `-Wall` or `-Wextra`, which would have caught an issue in #1495.


mrks:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78147 -.-

PR #1515, opened 2019-02-28
Make Segment iterators LegacyForwardIterator compatible

### LINKED ISSUE #1512: SegmentIterators are not suited for standard algorithms
Working at #1394, we are currently refactoring the code to use the SegmentIterators instead of custom code for each segment. When a segment is sorted, we'd like to call `std::lower_bound` on the segment iterator. However, this currently leads to compile-time errors, because the iterators are not copy-assignable. 

Code example of what we try to achieve:
```cpp
auto segment_iterable = create_iterable_from_segment(typed_segment);
segment_iterable.with_iterators(position_filter, [&](auto begin, auto end) {
auto result = std::lower_bound(begin, end, 27);
});
```

Excerpt of the error message:

```
/home/Christian.Flach/hyrise/src/lib/storage/segment_sorted_search.hpp:39:12: error: use of deleted function 'opossum::ValueSegmentIterable<int>::PointAccessIterator& opossum::ValueSegmentIterable<int>::PointAccessIterator::operator=(opossum::ValueSegmentIterable<int>::PointAccessIterator&&)'
     _begin = std::lower_bound(_begin, _end, false,
     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                               [](const auto& segment_position, const auto& _) { return segment_position.is_null(); });
                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/Christian.Flach/hyrise/src/lib/storage/create_iterable_from_segment.hpp:7,
                 from /home/Christian.Flach/hyrise/src/lib/operators/table_scan/column_vs_value_table_scan_impl.cpp:8:
/home/Christian.Flach/hyrise/src/lib/storage/value_segment/value_segment_iterable.hpp:167:9: note: 'opossum::ValueSegmentIterable<int>::PointAccessIterator& opossum::ValueSegmentIterable<int>::PointAccessIterator::operator=(opossum::ValueSegmentIterable<int>::PointAccessIterator&&)' is implicitly deleted because the default definition would be ill-formed:
   class PointAccessIterator : public BasePointAccessSegmentIterator<PointAccessIterator, SegmentPosition<T>> {
         ^~~~~~~~~~~~~~~~~~~
/home/Christian.Flach/hyrise/src/lib/storage/value_segment/value_segment_iterable.hpp:167:9: error: use of deleted function 'opossum::BasePointAccessSegmentIterator<opossum::ValueSegmentIterable<int>::PointAccessIterator, opossum::SegmentPosition<int> >& opossum::BasePointAccessSegmentIterator<opossum::ValueSegmentIterable<int>::PointAccessIterator, opossum::SegmentPosition<int> >::operator=(opossum::BasePointAccessSegmentIterator<opossum::ValueSegmentIterable<int>::PointAccessIterator, opossum::SegmentPosition<int> >&&)'
```

The `operator=` is implicitly deleted, because the `PointAccessIterator` (and almost all segment iterators) contains `const` members: 
https://github.com/hyrise/hyrise/blob/262ead52992bb61558d8e63a72510b08cb44b9fa/src/lib/storage/value_segment/value_segment_iterable.hpp#L195-L196

We thought of several possible solutions:
1. Making all members of all segment iterators non-const, so that they can be assigned in the `operator=`. This, however, also requires that the segments the iterators are created of are non-const.
2. Transforming all members `const T` into `std::shared_ptr<const T>`, so that the shared pointer can be assigned, but the underlying objects remain const.
3. Using `const_cast` in `operator=` to force-assign the values.

We don't like any of the options. Maybe you have an idea?

For now, we can also implement `std::lower_bound` and `std::upper_bound` in a way that doesn't require copy-assignment ourselves. But that certainly should only be an intermediary solution.

mrzzzrm:
Fix #1512

Makes the Segment iterators usable with STL algorithms (such as `std::lower_bounds`) that require copyable iterators.


mrzzzrm:
@cmfcmf all iterators should work with the STL algorithms requiring LegacyForwardIterator now.

@mrks you might wanna check whether you're fine with my usage of raw pointers into the Segments from the Iterators. The STL does the same and it's the best way to ensure iterator copying is cheap.

I want to open a/multiple separate PR which reduces the number of members that segment iterators have; reduces the number of string copies and maybe restructures the `src/storage/` directory a bit. This PR is to get @cmfcmf running as fast as possible ;) .

src/lib/storage/segment_iterables/any_segment_iterator.hpp
```diff
@@ -116,19 +116,27 @@ class AnySegmentIterator : public BaseSegmentIterator<AnySegmentIterator<T>, Seg
 
  public:
   AnySegmentIterator(const AnySegmentIterator& other) : _wrapper{other._wrapper->clone()} {}
+  AnySegmentIterator& operator=(const AnySegmentIterator& other) {
+    _wrapper = other._wrapper->clone();
```
cmfcmf:
```suggestion
    if (this == &other) { return *this; }
    _wrapper = other._wrapper->clone();
```

This page says that the object should perform no action on self-assignment: https://en.cppreference.com/w/cpp/language/operators

src/lib/storage/vector_compression/simd_bp128/simd_bp128_iterator.cpp
```diff
@@ -22,6 +22,16 @@ SimdBp128Iterator::SimdBp128Iterator(const SimdBp128Iterator& other)
       _current_meta_block{std::make_unique<std::array<uint32_t, Packing::meta_block_size>>(*other._current_meta_block)},
       _current_meta_block_index{other._current_meta_block_index} {}
 
+SimdBp128Iterator& SimdBp128Iterator::operator=(const SimdBp128Iterator& other) {
+  _data = other._data;
```
cmfcmf:
```suggestion
  if (this == &other) { return *this; }
  _data = other._data;
```

src/lib/storage/segment_iterables/any_segment_iterator.hpp
mrks: Nice. Self-assignment is something that I never really gave any attention to.
https://isocpp.org/wiki/faq/assignment-operators#self-assignment-how

mrks:
I don't like it, but I suppose it's reasonable. In the end we explicitly don't want the iterators to hold / share ownership.

mrzzzrm:
> I don't like it, but I suppose it's reasonable. In the end we explicitly don't want the iterators to hold / share ownership.

As discussed, I thought we could reduce the iterator state to a minimum and keep temporary materialized data in the Iterable. Worked fine, except for the JIT, which keeps an iterator for each column and throws the iterable away. 

Then I thought about it again and came to the conclusion that having to have living iterables in order to use the iterators is a pain. Especially then you'd probably want safe pointers from iterator to iterable and from iterable to segment. Would mean adjusting a lot of code and just feels cumbersome. So I decided a bit more state in the iterators doesn't hurt too much. If we want to we can probably implement the copy/move constructors of those iterators and explicitly not copy the shared state.

PR #1394, opened 2019-01-09
[DYOD18/19] Sorted Chunks

cmfcmf:
This is our team's (@hendraet, @johannes-schneider, @cmfcmf) sorted ~segment~ chunk implementation. 

The Chunknow has an optional member to store the a pair of `ColumnID` and `OrderByMode` to indicate a sorted column. Only one column can be marked as sorted.

We have enhanced the ColumnVsValueTableScan to take advantage of that flag.

### Microbenchmark

We have created a microbenchmark to measure if and how much we gain by our optimization. The benchmarks each create a table with `1'000'000` rows. Each row contains a column with ascendingly sorted numbers (`0` up to `rows - 1`). Depending on the benchmark, the column is either an integer or a string column (in which case leading 0s are prepended). We then scan for less than X, where X depends on the selectivity used in the benchmark. 
For our benchmarks, we made the strings 512 bytes long, which makes our implementation way faster regardless of segment type. The integer benchmarks are more interesting and not always faster.

The following diagrams compare our speed gains for
1. integer segments marked as sorted
2. integer segments not marked as sorted (although they are)
3. integer segments not marked as sorted and shuffled

2 and 3 illustrate how much is already gained "for free" without any code changes by taking advantage of processor optimization techniques like branch prediction.

grafik
grafik
grafik
grafik
grafik

PR #1517, opened 2019-03-01
Rewindable segment iterators

cmfcmf:
While working at #1394, we noticed that it would be nice to be able to rewind segment iterators. This PR makes all iterators rewindable.

### DIFF
--- a/src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
+++ b/src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
@@ -70,11 +70,26 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
       }
     }
 
+    void decrement() {
+      --_current_position;
+
+      if (_current_position <= *(_end_position_it - 1)) {
+        --_value_it;
+        --_null_value_it;
+        --_end_position_it;
+      }
+    }
+
     void advance(std::ptrdiff_t n) {
-      DebugAssert(n >= 0, "Rewinding iterators is not implemented");
       // The easy way for now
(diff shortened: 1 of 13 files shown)


mrzzzrm:
Which operator(/operation) would realistically be able to profit from this?

mrks:
I also found the need for this when I built the SIMD stuff. In the end, I chose another approach, but for ease of development, I do believe that our iterators should be rather complete, even if we don't use every method. #1515

mrks:
On first look, this appears to be in order.

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -70,11 +70,26 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
       }
     }
 
+    void decrement() {
+      --_current_position;
+
+      if (_current_position <= *(_end_position_it - 1)) {
```
mrzzzrm: `_end_position_it - 1` will be an illegal operation if _end_position_it is the `begin` iterator

mrzzzrm:
okay then

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
cmfcmf: Good catch! Fixed in https://github.com/hyrise/hyrise/pull/1517/commits/7a2cb883246c0a28f506b909d07e4d95f5ff045b

cmfcmf:
Tests have been added; this should be ready for review now :tada:

PR #1520, opened 2019-03-03
Force erasure of LZ4Segment

mrzzzrm:
I was a bit uneasy that it might still happen that code gets instantiated for the LZ4Iterable and higher compile times might be creeping in. This PR makes sure the compiler will tell us if someone attempts to do this and only allows type-erased iteration of the LZ4Segment


mrzzzrm:
Done

mrzzzrm:
Not possible to do this due to the create_any_segment_iterable->create_iterable_from_segment->create_any_segment_iterable recursion.

Issue #1321, opened 2018-11-22
Don't use random values in tests


mrzzzrm:
I just got this in CI:
![image](https://user-images.githubusercontent.com/2679018/48894494-2ffb0680-ee43-11e8-9b14-30c2f70978f3.png)

the test in question wasn't touched by me and uses random value generation. This is highly anti idiomatic, imo. Test should test explicit values, random values are not reproducible. And no, I don't think fixed seeds solve that, test values should be hand picked and thus comprehensible just from reading the test code.

*Update MD: Using a fixed-seed PRNG is also a bad fix because it involves implementation-dependent behavior.*

Bouncner:
Well, it helped a lot recently to find all sorts of edge-cases. I agree for unit tests, but it would still be nice to have some more fuzzing in Hyrise's tests.

I redo the test next week.

mrzzzrm:
I am not opposed to "fuzzing" tests at a higher level, maybe in the system tests. But the unit tests should not use randomly generated values.

Just as random values can (accidentally one could say) trigger edge cases, they might also test entirely uninteresting values and specifically not the edge cases. I think we're always better of to put some brains into a handful or two handpicked, explicit test values than to rely on random generation to find edge cases for us.

Then there is the case of test-readability. I will always prefer explicit values over a for loop, for some reasons:

E.g., here you can get an intuitive idea of which values are covered, i.e., how "good" the test tests.
```
EXPECT_TRUE(can_prune(filter, 5.0));
EXPECT_TRUE(can_prune(filter, 5.1)); // You get a test Failure here, you know exactly which value doesn't work
EXPECT_FALSE(can_prune(filter, 5.2));
EXPECT_FALSE(can_prune(filter, 15.3)):
```

Here, just from having a glimpse at this code you cannot tell how good the test is.
```
// You get a test Failure here, you have no idea what value didn't work. Run the test again in the debugger, you might get a different random value. Place a breakpoint in the loop body and you have to skip over a ton of values that do work.
for (auto v : gen_random_prunable_values()) EXPECT_TRUE(can_prune(filter, v)) ;
for (auto v : gen_random_non_prunable_values()) EXPECT_FALSE(can_prune(filter, v));
```

Also, with randomly generated values, there is always the temptation to just generate many, many random values because "better coverage" which results in the test running longer than it would need to, but no one daring to reduce the number of values generated because the original author surely had a reason to include so many.

Bouncner:
Moritz:
[...]

PR #1521, opened 2019-03-04
[DYOD18/19] Add LZ4 Point Access (also fixes #1522 and fixes #1516)

### LINKED ISSUE #1516: NULL vector size in ValueSegment<T>::estimate_memory_usage()
If I am not mistaken, the current implementation of `_null_values ? _null_values->size() * sizeof(bool) : 0u` overestimates the null_vector (`std::vector<bool>`) since `sizeof(bool)` is 1 but it's actually being optimized (afaik in all relevant implementations) to a vector of one bit per entry.

### LINKED ISSUE #1522: Run Length encoding and SIMDBP128 compression can't handle empty integer segments
In #1521 a test for the encoding of empty integer segments is added. This causes a segfault for both the Dictionary- and Frame Of Reference-Encoding with SIMDBP128 compression as well as the Run Length Encoding.

janehmueller:
This PR adds point access for LZ4 Segments via the LZ4 streaming API.

- For that we also use [zstd](https://github.com/facebook/zstd) to pre-train a dictionary on the uncompressed data. Using this dictionary as "state" for LZ4 we can compress the segment in independent blocks while maintaining a (hopefully) good compression ratio.

- ~~We will add benchmarks for the changes in this PR (and our seminar results in general) in next few days.~~ (see the results below) 

- These will include benchmarking different block sizes. The block size should be reasonably large (currently we chose 16 KB) since zstd dictionaries can't handle very small datasets (i.e., don't output a dictionary).

- There is some ugly code duplication in the LZ4Segment, since we use `char` vectors for `pmr_string` segments (and therefore the type parameter equals `pmr_string` and not `char`). We haven't found a nice workaround for that yet.

- We will also added simple caching to the point access iterator so that consecutively accessing the same block does not cause decompression of the same block multiple times. ~~(This will be added to the PR in the next few days)~~ (we added caching)

@Bouncner could you please add the appropriate reviewers?

# Benchmarks
The benchmarks were done with the same configuration as the ones we showed in our final presentation. The only new addition is, that we also compare the two vector compression methods (fixed size byte aligned (FSBA) and SimdBp128 (SIMD)). The default (which was used in the data we showed in our final presentation) is FSBA.
The column LZ4 (old) refers to the benchmarks we did in our final presentation. That implementation was not yet reviewed and not yet changed due to PR reviews.

## Sequential and Point Access
Compared to LZ4 (old), the sequential access worsened a little, which is expected since the data is split across multiple blocks. The point access improved, but unfortunately not by much.
According to Martin, this is mainly due to how strings are copied out of the decompressed data (which contains all the row values concatenated).
compression_access_times

## Compression Time
This shows compression times with the vector compression FSBA. The performance of SIMD was basically the same (a little faster with large amounts of data and a little slower with little data).
Compared to the LZ4 (old) the string compression worsened a lot (factor 5), while the other two were affected relatively little. This is explained by multiple factors. First, in string segments the offsets of the strings are additionally compressed with a vector compression method. Secondly, the training of the zstd dictionary could take is also trained before the compression happens. While this also happens for non-string segments, the make-up of the LZ4 segment members (shown at the end) shows that only for the string columns a dictionary is actually produced.
compression_time

### A more detailed view at the float and int columns
compression_time_detailed

## Compression Ratio
This compares the memory usage of the different encoding schemes when applied to different string columns.

### String Columns
In the previous benchmark we did (for the final presentation), the unencoded string column had a size of 3,2 MB instead of 4 MB. I don't know why that was the case (we used the same scale factor).
But even with possibly more data, this PR performs better than the previous one. LZ4 obviously has the least memory usage in the l_comment case, which is the worst case for dictionary encoding.
In the other two cases, LZ4 improved by factor 2 (FSBA) and 3 (SIMD) compared to LZ4 (old). It also comes close to the memory usage of the dictionary encoding. This is the result of reducing the size of the segment metadata considerably that happened in this PR.
compression_memory_usage_string

### Int and Float Columns
In these two cases the actual amount of data is relatively small (400 KB). The result is, that in the previous PR the size of the LZ4 segment was determined more by the amount of metadata than by the amount of actual compressed data. That is the main reason, why LZ4 improves by that much even though the compression happened without dictionary (can be seen in the make-up of the segments).
In the case of integers LZ4 performs best and in the case of floats it comes second behind dictionary encoding with SIMD.
compression_memory_usage_float_int

## Make-Up of LZ4Segment
Displayed is the amount of memory the different member variables of the LZ4 segment use. The object itself with all its members has a constant size of 184 bytes and is not displayed.

When looking at the string columns, it becomes obvious that the offsets make up a large part of the segment. When there is a lot of data (l_comment), this is still in proportion to the actual data. In the opposite case (the other two columns), they overshadow the data lot.
Also interesting to note is, that FSBA does not actually achieve any compression, since the offsets are originally 32-bit unsigned integers and there are as many offsets as there are row values (100.000). Therefore, the offsets also have an uncompressed size of `100.000 * 4 byte = 400 KB`.
Opposed to FSBA, SIMD actually compresses the offsets and nearly halves their size.

The zstd dictionary is very small compared to the compressed data and has basically no impact on the size of the segments.
The null values are not saved, since in the TPCH table we used (L_LINEITEM) no null values occur. This is a bias that occurs in every benchmark. If there were any or many null values, the LZ4 segment would have to actually store the null value vector (with a size of 12.5 KB).
member_memory_usage

Fix #1522 
Fix #1516


src/test/storage/encoded_string_segment_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+ protected:
+  /**
+   * Generate a random string with a given length from a set charset.
+   * This solution is taken from: https://stackoverflow.com/a/24586587
+   */
+  std::string random_string(std::string::size_type length) {
+    static auto& charset = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+    static std::random_device random_device;
+    static std::mt19937 generator(random_device());
+    static std::uniform_int_distribution<std::string::size_type> char_distribution(0, sizeof(charset) - 2);
```
mrzzzrm: pls dont use random data in test, it makes it hard to reproduce bugs

mrzzzrm:
> @Bouncner could you please add the appropriate reviewers?

Potential reviewer here :wave: How WIP is this PR? "(This will be added to the PR in the next few days)" makes it sound like you're still actively working on it.

janehmueller:
Implementation wise it's pretty much finished. What is left to to do, is the evaluation (via benchmarks) and implementing the point access caching (which will only add a few lines to both LZ4Segment and the point access iterator creation).

src/test/storage/encoded_segment_test.cpp
```diff
@@ -129,7 +129,14 @@ INSTANTIATE_TEST_CASE_P(
     formatter);
 
 TEST_P(EncodedSegmentTest, SequentiallyReadEmptyIntSegment) {
-  GTEST_SKIP();
+  const auto encoding_spec = GetParam();
+
+  // 1522
```
Bouncner: Please add a `#` and a short comment.

src/test/storage/encoded_segment_test.cpp
```diff
@@ -128,6 +128,33 @@ INSTANTIATE_TEST_CASE_P(
...
+       *encoding_spec.vector_compression_type == VectorCompressionType::SimdBp128)) {
+    GTEST_SKIP();
+  }
+
+  auto value_segment = std::make_shared<ValueSegment<int32_t>>(pmr_concurrent_vector<int32_t>{});
+  auto base_encoded_segment = this->encode_value_segment(DataType::Int, value_segment);
+
+  EXPECT_EQ(value_segment->size(), base_encoded_segment->size());
+
+  resolve_encoded_segment_type<int32_t>(*base_encoded_segment, [&](const auto& encoded_segment) {
```
mrzzzrm: any reason to leave these lines here if "nothing happens here"? If so, please doc, if no, please remove

src/lib/storage/lz4_segment.cpp
```diff
@@ -120,27 +140,222 @@ std::vector<pmr_string> LZ4Segment<pmr_string>::decompress() const {
...
   } else {
-    return std::allocate_shared<LZ4Segment>(alloc, std::move(new_compressed_data), std::move(new_null_values),
-                                            _decompressed_size);
+    return std::allocate_shared<LZ4Segment>(alloc, std::move(new_lz4_blocks), std::move(new_null_values),
+                                            std::move(new_dictionary), _block_size, _last_block_size, _compressed_size);
   }
 }
 
 template <typename T>
 size_t LZ4Segment<T>::estimate_memory_usage() const {
```
Bouncner: Image a column with 100 chunks of default size of 100,000 tuples. Assuming this column is a completely empty string column, we are storing `100,000 * sizeof(size_t) * 100 = 80MB`.
Is that correct? Compared to the actual data that's kinda ... crazy.

Some (but less significant) goes for the `null_values` vector.

If I am not mistaken and the structures are that large: do you have any ideas what to do here?

mrzzzrm:
> For that we also use zstd to pre-train a dictionary on the uncompressed data. Using this dictionary as "state" for LZ4 we can compress the segment in independent blocks while maintaining a (hopefully) good compression ratio.

Before I start deducing things from code, can you elaborate a bit on that approach? This PR proposes merging code from facebook, after all...

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -111,6 +119,12 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
     auto offsets = pmr_vector<size_t>{alloc};
     offsets.resize(num_elements);
 
+    /**
+     * These are the lengths of each string. They are needed to train the zstd dictionary.
```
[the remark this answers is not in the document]
janehmueller: The method of the zstd API that we use is called `ZDICT_trainFromBuffer`, therefore I'll change the generate to train.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -49,30 +54,33 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+     * instead of iterating over all blocks when the memory consumption of the LZ4 segment is estimated.
+     */
+    auto lz4_blocks = pmr_vector<pmr_vector<char>>{alloc};
+    _compress(values, lz4_blocks, dictionary);
+
+    auto last_block_size = input_size % _block_size ? input_size % _block_size : _block_size;
+
+    size_t total_compressed_size{};
+    for (const auto& block : lz4_blocks) {
+      total_compressed_size += block.size();
```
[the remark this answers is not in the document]
janehmueller: `block_size` is the uncompressed size of each block that will be compressed.
Here we add the sizes of the compressed blocks, hence `total_compressed_size`.

I'll change `block` to `compressed_block`.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +136,340 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+    /**
+     * If the generation still failed with increased input data, then compress without a dictionary (the compression
+     * ratio will suffer).
+     */
+    if (ZDICT_isError(dictionary_size)) {
+      return pmr_vector<char>{};
+    }
+
+    DebugAssert(dictionary_size <= max_dictionary_size,
+                "Generated ZSTD dictionary in LZ4 compression is larger than "
```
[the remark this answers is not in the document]
janehmueller: I don't quite understand this? The code line you commented on is in uppercase. As for the variables with lz4 in their names, the code style specifies lowercase variable names.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +136,340 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+
+ private:
+  static constexpr size_t _maximum_dictionary_size = 10000000u;
+  static constexpr size_t _maximum_value_size = 1000000u;
+  static constexpr size_t _minimum_value_size = 20000u;
+
+  /**
+   * Use the LZ4 high compression stream API to compress the input values. The data is separated into different
+   * blocks that are compressed independently. To maintain a high compression ratio and indepdence of these blocks
+   * we use dictionary generated via zstd. LZ4 can use the dictionary "learned" on the column to compress the data
```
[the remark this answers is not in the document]
janehmueller: Do you mean the missing article?

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +136,340 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+
+    // This magic line is needed for zstd to suceed.
+    sample_sizes.emplace_back(num_values);
+
+    /**
+     * Also add the whole data range in samples of size 10 to the sample size vector. This is also needed in combination
+     * with the line above. The idea here, is too provide larger samples to zstd. The input data is only too small to
+     * generate a dictionary when the strings are very short (i.e., single character values).
+     */
+    for (size_t i = 0u; i < num_values; i += 10u) {
```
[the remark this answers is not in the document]
janehmueller: We found it working well, via trial and error. I added a comment for that.

src/test/storage/encoded_segment_test.cpp
```diff
@@ -128,6 +128,33 @@ INSTANTIATE_TEST_CASE_P(
...
+
+  // 1522
+  if (encoding_spec.encoding_type == EncodingType::RunLength ||
+      (encoding_spec.vector_compression_type.has_value() &&
+       *encoding_spec.vector_compression_type == VectorCompressionType::SimdBp128)) {
+    GTEST_SKIP();
+  }
+
+  auto value_segment = std::make_shared<ValueSegment<int32_t>>(pmr_concurrent_vector<int32_t>{});
+  auto base_encoded_segment = this->encode_value_segment(DataType::Int, value_segment);
```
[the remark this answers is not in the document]
janehmueller: That's the code style of the other tests.

src/test/storage/encoded_string_segment_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+    return result_string;
+  }
+
+  std::shared_ptr<ValueSegment<pmr_string>> create_empty_string_value_segment() {
+    auto values = pmr_concurrent_vector<pmr_string>(row_count);
+    return std::make_shared<ValueSegment<pmr_string>>(std::move(values));
+  }
+
+  std::shared_ptr<ValueSegment<pmr_string>> create_empty_string_w_null_value_segment() {
```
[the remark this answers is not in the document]
janehmueller: That's the naming convention of the same methods for integers in the `EncodedSegmentTest`

src/test/storage/encoded_string_segment_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+    std::uniform_int_distribution<uint32_t> dist{0u, max_length};
+
+    for (auto& elem : values) {
+      elem = random_string(dist(engine));
+    }
+
+    return std::make_shared<ValueSegment<pmr_string>>(std::move(values));
+  }
+
+  std::shared_ptr<ValueSegment<pmr_string>> create_string_w_null_value_segment() {
```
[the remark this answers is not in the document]
janehmueller: That's the naming convention of the same methods for integers in the `EncodedSegmentTest`.

src/test/storage/encoded_string_segment_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+INSTANTIATE_TEST_CASE_P(
+    SegmentEncodingSpecs, EncodedStringSegmentTest,
+    ::testing::Values(SegmentEncodingSpec{EncodingType::Dictionary, VectorCompressionType::SimdBp128},
+                      SegmentEncodingSpec{EncodingType::Dictionary, VectorCompressionType::FixedSizeByteAligned},
+                      SegmentEncodingSpec{EncodingType::RunLength}, SegmentEncodingSpec{EncodingType::LZ4}),
+    formatter);
+
+TEST_P(EncodedStringSegmentTest, SequentiallyReadNotNullableEmptyStringSegment) {
+  auto value_segment = this->create_empty_string_value_segment();
```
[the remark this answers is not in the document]
janehmueller: That's the code style of the same methods for integers in the `EncodedSegmentTest`.

src/test/storage/encoded_string_segment_test.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+  EXPECT_EQ(value_segment->size(), base_encoded_segment->size());
+
+  resolve_encoded_segment_type<pmr_string>(*base_encoded_segment, [&](const auto& encoded_segment) {
+    auto value_segment_iterable = create_iterable_from_segment(*value_segment);
+    auto encoded_segment_iterable = create_iterable_from_segment(encoded_segment);
+
+    value_segment_iterable.with_iterators([&](auto value_segment_it, auto value_segment_end) {
+      encoded_segment_iterable.with_iterators([&](auto encoded_segment_it, auto encoded_segment_end) {
+        auto row_idx = 0;
```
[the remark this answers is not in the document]
janehmueller: That's the naming convention of the same methods for integers in the `EncodedSegmentTest`.

src/test/storage/encoded_string_segment_test.cpp
```diff
@@ -0,0 +1,313 @@
...
```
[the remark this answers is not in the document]
janehmueller: That's the code style of the same methods for integers in the `EncodedSegmentTest`.

src/test/storage/encoded_segment_test.cpp
janehmueller: No since just the compression is tested. I'll remove them.

janehmueller: As it turns out, this part causes the SIMDBP128 encoded segments to crash. Trying to iterate over the segment causes that. I added a comment.

src/lib/storage/lz4_segment.cpp
janehmueller: In the case of a totally empty string segment we could just make the offset vector empty, since they are never used. But as soon as only a single character is somewhere in the column, it is needed.

As for the null_values: the segment could also be made not nullable (same as the value segment) but that doesn't really solve the problem.

janehmueller:
We just added the caching implementation. Implementation wise this PR is done (i.e., not wip anymore). What's left are the benchmarks and refactoring due to PR feedback (i.e., code deduplication).

src/lib/storage/lz4_segment.cpp
Bouncner: As the purpose of this PR is to reduce the footprint of a segment as much as possible, it would be really nice to have both null values and the offsets only stored when necessary. Do you think you can cover this in this PR or is it a lot of work? I don't want to punish you with such things on the last mile. ;)

Bouncner: Oh and one thing: why use `size_t` for the offets over `ChunkOffset`?

phnm:
> 
> 
> > For that we also use zstd to pre-train a dictionary on the uncompressed data. Using this dictionary as "state" for LZ4 we can compress the segment in independent blocks while maintaining a (hopefully) good compression ratio.
> 
> Before I start deducing things from code, can you elaborate a bit on that approach? This PR proposes merging code from facebook, after all...

#### About zstd
Zstandard is about another compression algorithm from same family of algorithms as LZ4. More importantly, it also offers a dictionary builder that generates a dictionary from data that is to be compressed and that dictionary is also supported by LZ4. Zstd's and LZ4's reference implementations are written by the same guy at facebook, after all 😏 

#### Why train a dictionary
LZ4 offers two modes: block and stream compression. Stream compression is made for compressing large files as a stream of blocks, while still taking advantage of information redundancy in the whole file - single block compression can't do that and thus can not compress with as good ratios. For stream compression LZ4 builds up a dictionary internally that is incremented with every block that is compressed. In consequence of that, those blocks can only be decompressed in the same order they were compressed, making it useless for our point access case, where we only want to decompress the one block containing the requested element (or several blocks for strings larger than one block size). This is where the pre-trained dictionary comes to use: instead of building up a dictionary internally, it is already provided for compression and decompression. That makes it possible decompress a block independently and should also be slightly faster at that at the same time.

We did have some trouble with the dictionary training though, as it fails when it doesn't find enough redundant information in the given data for the given target dictionary size and thus wouldn't bring any benefit compared to the standard single block compression.

src/lib/storage/lz4_segment.cpp
```diff
@@ -120,27 +140,257 @@ std::vector<pmr_string> LZ4Segment<pmr_string>::decompress() const {
...
+  }
   auto new_null_values = pmr_vector<bool>{_null_values, alloc};
+  auto new_dictionary = pmr_vector<char>{_dictionary, alloc};
 
-  if (_offsets.has_value()) {
-    auto new_offsets = pmr_vector<size_t>(*_offsets, alloc);
-    return std::allocate_shared<LZ4Segment>(alloc, std::move(new_compressed_data), std::move(new_null_values),
-                                            std::move(new_offsets), _decompressed_size);
+  if (_string_offsets.has_value()) {
+    auto new_string_offsets = pmr_vector<size_t>(*_string_offsets, alloc);
```
pascalhpi: As I mentioned I would prefer calling all container constructors with normal braces, but if you disagree you should use curly braces here as well for consistency.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -49,30 +53,35 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
-    const int compression_result = LZ4_compress_HC(reinterpret_cast<char*>(values.data()), compressed_data.data(),
-                                                   static_cast<int>(input_size), output_size, LZ4HC_CLEVEL_MAX);
-    Assert(compression_result > 0, "LZ4 compression failed");
-
-    // shrink the vector to the actual size of the compressed result
-    compressed_data.resize(static_cast<size_t>(compression_result));
-    compressed_data.shrink_to_fit();
-
-    return std::allocate_shared<LZ4Segment<T>>(alloc, std::move(compressed_data), std::move(null_values), input_size);
+    auto dictionary = pmr_vector<char>{};
```
pascalhpi: I personally would prefer using uint8_t over char, since char is not guaranteed to be 8bits and because the semantic of a char is, well, a character, but what you want here is just a byte, without any semantics. To really make that clear you could even do `using byte = uint8_t;` and use byte all the time.

src/lib/storage/run_length_segment/run_length_encoder.hpp
```diff
@@ -28,6 +28,11 @@ class RunLengthEncoder : public SegmentEncoder<RunLengthEncoder> {
     auto iterable = ValueSegmentIterable<T>{*value_segment};
 
     iterable.with_iterators([&](auto it, auto end) {
+      // Early out for empty segments, code below assumes it to be non-empty
+      if (it == end) {
+        return;
+      }
+
       // Init is_current_null such that it does not equal the first entry
       auto current_value = T{};
```
pascalhpi: Since you are calling default constructors like this most of the time, for consistency:
```suggestion
      T current_value;
```
Although I prefer `auto current_value = T{}`, see [here why](https://www.fluentcpp.com/2018/09/28/auto-stick-changing-style/).

src/test/storage/encoded_string_segment_test.cpp
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  std::shared_ptr<ValueSegment<pmr_string>> create_empty_string_w_null_value_segment() {
+    auto values = pmr_concurrent_vector<pmr_string>(row_count);
+    auto null_values = pmr_concurrent_vector<bool>(row_count);
+
+    std::default_random_engine engine{};
+    std::bernoulli_distribution bernoulli_dist{0.3};
+
+    for (auto i = 0u; i < row_count; ++i) {
```
pascalhpi: There are several loops where you use i, but in most you use index. Decide for one and use it consistently.
```suggestion
    for (auto index = size_t{ 0 }; index < row_count; ++index) {
```

src/test/storage/encoded_string_segment_test.cpp
```diff
@@ -0,0 +1,301 @@
...
+auto formatter = [](const ::testing::TestParamInfo<SegmentEncodingSpec> info) {
+  const auto spec = info.param;
+
+  auto stream = std::stringstream{};
+  stream << encoding_type_to_string.left.at(spec.encoding_type);
+  if (spec.vector_compression_type) {
+    stream << "-" << vector_compression_type_to_string.left.at(*spec.vector_compression_type);
+  }
+
+  auto string = stream.str();
```
pascalhpi: I would suggest using another, probably more descriptive name, but definitely one which is not called like a type.

src/lib/storage/lz4_segment.cpp
```diff
@@ -120,27 +140,257 @@ std::vector<pmr_string> LZ4Segment<pmr_string>::decompress() const {
...
+     * Multiple blocks need to be decompressed. Iterate over all relevant blocks and append the result to this string
+     * stream.
+     */
+    std::stringstream result_string;
+
+    // These are the character offsets that need to be read in every block.
+    size_t block_start_offset = start_offset % _block_size;
+    size_t block_end_offset = _block_size;
+
+    const auto use_caching =
```
Bouncner: Comment please.

src/lib/storage/lz4_segment.cpp
Bouncner: Thinking about: leave the `size_t`. Actually, `uint32_t` should be sufficient, but we never know what cases we'll later see.
We should think about compression of offsets anyways at a later point in time.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +137,348 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+  pmr_vector<char> _train_dictionary(const pmr_vector<T>& values) {
+    /**
+     * The minimum sample size is 8 bytes. Since non-string data has a constant size for each value we can just set
+     * the sample size to 8 for all values.
+     */
+    const size_t min_sample_size = 8u;
+    const auto values_size = values.size() * sizeof(T);
+    const auto sample_size = std::max(sizeof(T), min_sample_size);
+    const auto num_samples = values_size / sample_size;
+    const std::vector<size_t> sample_lens(num_samples, sample_size);
```
Bouncner: Wha is `sample_lens` supposed to mean?

Further, for a `long` column, we gather the full segment as samples (100k) and for `int`, we gather 50k samples. Why that?

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +137,348 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+   *               n times (i.e., its size equals n * num_values). This method appends items to this vector.
+   * @param sample_sizes The sample sizes provided to zstd. This method appends items to this vector.
+   * @param num_values The number of characters in the original data.
+   */
+  void _increase_dictionary_input_data(pmr_vector<char>& values, pmr_vector<size_t>& sample_sizes, size_t num_values) {
+    // The first num_values values in the value data is the original input. It is just copied and appended again.
+    values.insert(values.end(), values.begin(), values.begin() + num_values);
+
+    // This magic line is needed for zstd to succeed.
+    sample_sizes.emplace_back(num_values);
```
Bouncner: Leads to segfaults.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -49,30 +53,35 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
     });
 
     /**
-     * Use the LZ4 high compression API to compress the copied values. As C-library LZ4 needs raw pointers as input
-     * and output. To avoid directly handling raw pointers we use std::vectors as input and output. The input vector
-     * contains the data that needs to be compressed and the output vector is allocated enough memory to contain
-     * the compression result. Via the .data() call we can supply LZ4 with raw pointers to the memory the vectors use.
-     * These are cast to char-pointers since LZ4 expects char pointers.
+     * Pre-compute a zstd dictionary if the input data is split among multiple blocks. This dictionary allows
+     * independent compression of the blocks, while maintaining a good compression ratio.
```
mrks: why is that good?

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -104,13 +113,19 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
      * These offsets mark the beginning of strings (and therefore end of the previous string) in the data vector.
      * These offsets are character offsets. The string at position 0 starts at the offset stored at position 0, which
      * will always be 0.
-     * Its exclusive end is the offset stored at position 1 (i.e. offsets[1] - 1 is the last character of the string
+     * Its exclusive end is the offset stored at position 1 (i.e., offsets[1] - 1 is the last character of the string
      * at position 0).
      * In case of the last string its end is determined by the end of the data vector.
      */
     auto offsets = pmr_vector<size_t>{alloc};
```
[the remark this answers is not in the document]
mrks: Negative. That would only happen if you had a user-defined conversion from alloc to size_t. But yes, that ambiguity sucks.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +136,340 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+
+    size_t total_compressed_size{};
+    for (const auto& block : lz4_blocks) {
+      total_compressed_size += block.size();
+    }
+
+    return std::allocate_shared<LZ4Segment<pmr_string>>(alloc, std::move(lz4_blocks), std::move(null_values),
+                                                        std::move(dictionary), std::move(offsets), _block_size,
+                                                        last_block_size, total_compressed_size);
+  }
```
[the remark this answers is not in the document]
mrks: Offline right now, so I can't add a thumbs up.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +137,348 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+                "the memory allocated for it.");
+    dictionary.resize(dictionary_size);
+    dictionary.shrink_to_fit();
+
+    return dictionary;
+  }
+
+  /**
+   * When training a dictionary for strings, each row element is a sample. This is not done when training a
+   * dictionary for non-strings since zstd's dictionary is made for strings (and the samples can have different sizes).
```
mrks: I can't follow this part. Do you mean your sample density is 100%? Also, could you try to explain the second sentence in a different way/

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +137,348 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+   *
+   * @param values The input data that will be compressed (i.e., all strings concatenated).
+   * @param sample_sizes A vector of sample lengths. Each length corresponds to a substring in the values vector. These
+   *                     should correspond to the length of each row's value.
+   * @return The trained dictionary or in the case of failure an empty vector.
+   */
+  pmr_vector<char> _train_string_dictionary(const pmr_vector<char>& values, const pmr_vector<size_t>& sample_sizes) {
+    /**
+     * The recommended dictionary size is about 1/100th of size of all samples combined, but he size also has to be at
+     * least 1KB. Smaller dictionaries won't work.
```
mrks: typo: the

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +137,348 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+    DebugAssert(dictionary_size <= max_dictionary_size,
+                "Generated ZSTD dictionary in LZ4 compression is larger than "
+                "the memory allocated for it.");
+
+    // Shrink the allocated dictionary size to the actual size.
+    dictionary.resize(dictionary_size);
+    dictionary.shrink_to_fit();
+
+    return dictionary;
+  }
```
mrks: Could you reduce duplication with above by using `if constexpr` or alike?

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +137,348 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+
+  /**
+   * Increase the dictionary input data by appending the original input data (linear increase) and increasing the
+   * maximum dictionary size (exponential increase). This only happens if the input data is too small for zstd to
+   * successfully train a dictionary.
```
mrks: Is this a standard approach or did you come up with it?

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +137,348 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+    dictionary.shrink_to_fit();
+
+    return dictionary;
+  }
+
+  /**
+   * Increase the dictionary input data by appending the original input data (linear increase). This only happens if
+   * the input data is too small for zstd to successfully train a dictionary.
+   *
+   * The sample sizes are not copied directly. First, we add a sample size of the whole input vector (somehow this is
```
mrks: Do you mean 'the samples are not...'?

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +137,348 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+   * important for zstd to successfully train a dictionary).
+   * Afterwards, we again add the whole range of the input data as sample sizes of size 10 (and the last sample size
+   * varies according to the size of the input data).
```
mrks: tbh, I can't follow the compression magic that you are doing here. From a code and documentation perspective, it looks good. If I worked my way deeper into the lz7 lib, this would probably make more sense to me. @Bouncner: I'd appreciate a check of the algorithms used here, if you can.

src/lib/storage/lz4/lz4_encoder.hpp
phnm: I'm not sure whether you've seen my comment from a few days ago where I tried to explain why we train a dictionary: https://github.com/hyrise/hyrise/pull/1521#issuecomment-470761144
If that leaves you with questions, what exactly are they?

src/lib/storage/lz4/lz4_encoder.hpp
phnm: The method to train a dictionary provided by zstd expects samples, which can be of varying length. We use a fixed length of samples for data types of fixed size and the size of each of string for string segments, so every string is one sample. So yes, we train on 100% of the segment.
Zstd / LZ4 are made for huge amounts of data, so segments with 100k rows are still small and should not be a too large amount of samples. All the guideline on the amount of samples we found was "a few thousand samples" if I remember correctly.

src/lib/storage/lz4/lz4_encoder.hpp
phnm: It's our own workaround. We had some trouble with the dictionary training, as it fails when it doesn't find enough redundant information in the given data for the given target dictionary size, which it's documentation says it can just do. We found that increasing the size of the dictionary and/or duplicating the input helps with that, but it seams that there are still cases where it will just fail for reasons not really not known to us. We will discuss that at our next meeting on Tuesday, it might be wiser to not train a dictionary if it fails the first time and just take a loss in compression ratio that comes with not using one.
Just fyi, my teammates and me are all currently in different places on different missions this week and can't really work on implementing your reviews, but we will catch up on the weekend / beginning of next week.

src/lib/storage/lz4_segment.cpp
```diff
@@ -120,27 +140,257 @@ std::vector<pmr_string> LZ4Segment<pmr_string>::decompress() const {
...
+                                               const std::optional<size_t> previous_block_index,
+                                               std::vector<char>& previous_block) const {
+  const auto memory_offset = chunk_offset * sizeof(T);
+  const auto block_index = memory_offset / _block_size;
+
+  /**
+   * If the previously decompressed block was a different block than the one accessed now, overwrite it with the now
+   * decompressed block.
+   */
+  if (!previous_block_index.has_value() || block_index != *previous_block_index) {
```
Bouncner: I don't get the comment. It sounds opposite to what is done.

src/lib/storage/lz4_segment.cpp
```diff
@@ -120,27 +140,257 @@ std::vector<pmr_string> LZ4Segment<pmr_string>::decompress() const {
...
+  // Assure that the decompressed data fits into the vector.
+  if (decompressed_data.size() != _block_size) {
+    decompressed_data.resize(_block_size);
+  }
+
+  // We use the string method since we handle a char-vector (even though the data is no necessarily string data).
+  _decompress_block_to_bytes(block_index, decompressed_data, 0u);
+
+  /**
+    * In the case of the last block, the decompressed data is possibly smaller than _block_size (it is _last_block_size
```
Bouncner: I don't get the comment. It sounds opposite to what is done.

src/lib/storage/lz4_segment.cpp
```diff
@@ -120,27 +140,257 @@ std::vector<pmr_string> LZ4Segment<pmr_string>::decompress() const {
...
+
+  Assert(decompressed_result > 0, "LZ4 stream decompression failed");
+  DebugAssert(static_cast<size_t>(decompressed_result) == decompressed_block_size,
+              "Decompressed LZ4 block has different size than the initial source data.")
+}
+
+template <typename T>
+std::pair<T, size_t> LZ4Segment<T>::decompress(const ChunkOffset& chunk_offset,
+                                               const std::optional<size_t> previous_block_index,
+                                               std::vector<char>& previous_block) const {
```
Bouncner: This is called by `LZ4Segment<T>::decompress(const ChunkOffset& chunk_offset)` and an empty vector is passed as a reference. Why is this vector called `previous_block`?

src/lib/storage/lz4/lz4_encoder.hpp
mrks: This should be included in the comment. Once it#s merged, anything in the pr comments will never be looked again.

src/lib/storage/lz4_segment.cpp
```diff
@@ -120,27 +140,257 @@ std::vector<pmr_string> LZ4Segment<pmr_string>::decompress() const {
...
+    decompressed_data.resize(_block_size);
+  }
+
+  // We use the string method since we handle a char-vector (even though the data is no necessarily string data).
+  _decompress_block_to_bytes(block_index, decompressed_data, 0u);
+
+  /**
+    * In the case of the last block, the decompressed data is possibly smaller than _block_size (it is _last_block_size
+    * large). However, when using _last_block_size as size for the decompression buffer, the LZ4 decompression fails.
+    * Therefore, the block is resized and shrunk afterwards.
```
mrks: Where is it shrunk? And why would we want that?

src/lib/storage/lz4_segment.hpp
```diff
@@ -20,29 +21,72 @@ class BaseCompressedVector;
...
-  explicit LZ4Segment(pmr_vector<char>&& compressed_data, pmr_vector<bool>&& null_values, pmr_vector<size_t>&& offsets,
-                      const size_t decompressed_size);
+  explicit LZ4Segment(pmr_vector<pmr_vector<char>>&& lz4_blocks, pmr_vector<bool>&& null_values,
+                      pmr_vector<char>&& dictionary, const size_t block_size, const size_t last_block_size,
+                      const size_t compressed_size);
 
-  explicit LZ4Segment(pmr_vector<char>&& compressed_data, pmr_vector<bool>&& null_values,
-                      const size_t decompressed_size);
+  /**
+   * This is a container for an LZ4 compressed segment. It contains the compressed data in blocks, the necessary
```
mrks: remove repeated comment?

src/test/storage/encoded_string_segment_test.cpp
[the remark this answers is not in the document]
mrks: Which does not necessarily mean it's a good thing ;)

src/test/storage/encoded_string_segment_test.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+  std::shared_ptr<ValueSegment<pmr_string>> create_string_w_null_value_segment() {
+    auto values = pmr_concurrent_vector<pmr_string>(row_count);
+    auto null_values = pmr_concurrent_vector<bool>(row_count);
+
+    std::default_random_engine engine{};
+    std::bernoulli_distribution bernoulli_dist{0.3};
+
+    for (auto i = 0u; i < row_count; ++i) {
+      null_values[i] = bernoulli_dist(engine);
```
mrks: Can we make this predictable? Random behavior in tests is hard to debug #1321

src/lib/storage/lz4_segment.cpp
Bouncner: Please, consider using SIMDBP128 vector compression.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +137,348 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
      */
     if (!num_chars) {
-      return std::allocate_shared<LZ4Segment<pmr_string>>(alloc, pmr_vector<char>{alloc}, std::move(null_values),
-                                                          std::move(offsets), 0u);
+      auto empty_blocks = pmr_vector<pmr_vector<char>>{alloc};
+      auto empty_dictionary = pmr_vector<char>{};
+      auto empty_offsets = pmr_vector<size_t>{};
+      return std::allocate_shared<LZ4Segment<pmr_string>>(alloc, std::move(empty_blocks), std::move(null_values),
+                                                          std::move(empty_dictionary), std::move(empty_offsets),
+                                                          _block_size, 0u, 0u);
```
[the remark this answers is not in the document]
janehmueller: Imo, it improves readability (and catching errors when changing the constructor is easier).

src/lib/storage/lz4/lz4_encoder.hpp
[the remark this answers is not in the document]
janehmueller: We removed the padding of the input data for the dictionary (due to a segfault with more input data).

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -122,48 +136,340 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+    const auto values_size = values.size() * sizeof(T);
+    const auto sample_size = sizeof(T) > 8 ? sizeof(T) : 8;
+    const auto num_samples = values_size / sample_size;
+    const std::vector<size_t> sample_lens(num_samples, sample_size);
+
+    /**
+     * The recommended dictionary size is about 1/100th of size of all samples combined, but he size also has to be at
+     * least 1KB. Smaller dictionaries won't work.
+     */
+    auto max_dictionary_size = num_samples * sample_size / 100;
```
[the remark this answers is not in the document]
janehmueller: Since this only happens in one line and it is simple math I don't think this needs to be its own variable.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: We renamed it to `sample_sizes`. It is the length of each sample. I.e., the value at index 0 is the size of the first sample in bytes. The samples are stored in the `values` vector.

Since the minimum sample size is 8 byte and a (32 bit) integer is only 4 bytes large, each sample consists of 2 integers. Thus we only have half as many samples.

src/lib/storage/lz4_segment.cpp
janehmueller: We removed the shrinking since the vector is re-used to store each decompressed block.

src/lib/storage/lz4_segment.cpp
janehmueller: I renamed it to `cached_block`. The vector is used to store the previously decompressed blocks (and to skip decompression if the same block is accessed).

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: We removed this part since Martin found out that it segfaults for a large amount of input data. Now, we just don't use a dictionary if the training fails.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: We removed this part of the code.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: Since this workaround segfaults with a large amount of data and doesn't really work without the magic line, we removed it.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: We removed the code duplication when we removed the padding of the input data.

src/lib/storage/lz4/lz4_encoder.hpp
janehmueller: We use char, since the underlying libraries (both lz4 and zstd) expect the dictionary to be a char-pointer. A char is guaranteed to be 1 byte large (which is at least 8 bits large).

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -23,6 +24,9 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
  public:
   static constexpr auto _encoding_type = enum_c<EncodingType, EncodingType::LZ4>;
   static constexpr auto _uses_vector_compression = false;
+  static constexpr size_t _block_size = 16384u;
+  static_assert(_block_size <= std::numeric_limits<int>::max(),
```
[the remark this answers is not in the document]
janehmueller: We use int, since lz4 expects a value of type int. If int is not 32 bit long, then the library will also expect/use the different size.

janehmueller:
As a result of our latest meeting with Martin, we will still add a few optimisations for the memory usage of the lz4 segment.

mrks:
Can we do that in a separate PR?

Bouncner:
Yes, we can. It’s “just” a memory consumption reduction (a significant though).

But I won’t find the time for a full review until next week.

src/lib/storage/lz4_segment.cpp
```diff
@@ -69,31 +82,38 @@ size_t LZ4Segment<T>::size() const {
 
 template <typename T>
 std::vector<T> LZ4Segment<T>::decompress() const {
-  auto decompressed_data = std::vector<T>(_decompressed_size / sizeof(T));
-  auto compressed_size = static_cast<int>(_compressed_data.size());
-  const int decompressed_result =
-      LZ4_decompress_safe(_compressed_data.data(), reinterpret_cast<char*>(decompressed_data.data()), compressed_size,
-                          static_cast<int>(_decompressed_size));
-  Assert(decompressed_result > 0, "LZ4 decompression failed");
+  auto decompressed_data = std::vector<T>(size());
```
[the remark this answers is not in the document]
janehmueller: It's part of the segment interface (it returns the number of rows in the segment).

src/lib/storage/lz4_segment.cpp
janehmueller: I clarified the comment.

The last block's size can be smaller than the block size. Say the block size equals 16000 and the last block has a size of 4000. When we try to decompress the last block into a vector of size 4000, the LZ4 decompression method returns an error code. However, it works with the block size of 16000.
Our workaround for that, is to decompress the last block into a vector with the full block size to prevent that error. We resize it afterwards, so that it can be accessed correctly.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -180,12 +196,14 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
       auto empty_dictionary = pmr_vector<char>{};
-      auto empty_offsets = pmr_vector<size_t>{};
       return std::allocate_shared<LZ4Segment<pmr_string>>(alloc, std::move(empty_blocks), std::move(null_values),
-                                                          std::move(empty_dictionary), std::move(empty_offsets),
-                                                          _block_size, 0u, 0u);
+                                                          std::move(empty_dictionary), nullptr, _block_size, 0u, 0u);
     }
 
+    // Compress the offsets with SIMDBP128 to reduce the memory footprint of the LZ4 segment.
+    auto vector_compressor = SimdBp128Compressor{};
```
mrks: With the dictionary encoding, we allow different compressions to be used. If we bring in vector compression here, we should follow the pattern that it is decoupled from the chosen segment encoding.

janehmueller:
Please do. It will take some time for Martin to review this PR anyway.

src/lib/storage/lz4/lz4_encoder.hpp
```diff
@@ -144,6 +149,9 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
     auto null_values = pmr_vector<bool>{alloc};
     null_values.resize(num_elements);
 
+    // If the segment does not contain any null values, we don't have to store the null values vector and can use.
```
mrks: use...?

src/lib/storage/lz4/lz4_iterable.hpp
```diff
@@ -22,8 +22,11 @@ class LZ4Iterable : public PointAccessibleSegmentIterable<LZ4Iterable<T>> {
 
     auto decompressed_segment = _segment.decompress();
 
-    auto begin = Iterator<ValueIterator>{decompressed_segment.cbegin(), _segment.null_values().cbegin()};
-    auto end = Iterator<ValueIterator>{decompressed_segment.cend(), _segment.null_values().cend()};
+    // If the null values vector doesn't exist, then there are no null values in the segment/
```
mrks: `/`?

src/lib/storage/lz4/lz4_iterable.hpp
```diff
@@ -47,9 +50,12 @@ class LZ4Iterable : public PointAccessibleSegmentIterable<LZ4Iterable<T>> {
       cached_block_index = block_index;
     }
 
-    auto begin = PointAccessIterator<ValueIterator>{decompressed_filtered_segment, &_segment.null_values(),
+    // If the null values vector doesn't exist, then there are no null values in the segment/
```
mrks: dto

src/lib/storage/lz4_segment.hpp
```diff
@@ -72,18 +80,23 @@ class LZ4Segment : public BaseEncodedSegment {
...
+   *                       The offsets are compressed using a vector compression method to reduce their memory footprint.
    * @param block_size The decompressed size of each full block in bytes. This can be numeric_limits<int>::max() at max.
    * @param last_block_size The size of the last block in bytes. It is a separate value since the last block is not
    *                        necessarily full.
    * @param compressed_size The sum of the compressed size of all blocks. This is a separate argument so that
    *                        there is no need to iterate over all blocks when estimating the memory usage.
+   * @param num_elements The number of elements in this segment. This needs to be stored in its own variable, since
+   *                     the other variables might not be set or stored to reduce the memory footprint. E.g., a string
+   *                     segment with only empty strings as elements would have no other way to know how many rows there
+   *                     are.
```
mrks: Is there a test for those cases?

Bouncner:
Apart from the few comments by mrks, this PR looks ready to go.

src/lib/storage/lz4_segment.hpp
janehmueller: Now there is 👍

src/lib/storage/lz4/lz4_iterable.hpp
```diff
@@ -22,24 +22,48 @@ class LZ4Iterable : public PointAccessibleSegmentIterable<LZ4Iterable<T>> {
...
     auto decompressed_segment = _segment.decompress();
 
-    auto begin = Iterator<ValueIterator>{decompressed_segment.cbegin(), _segment.null_values().cbegin()};
-    auto end = Iterator<ValueIterator>{decompressed_segment.cend(), _segment.null_values().cend()};
+    /**
+     * If the null value vector doesn't exist, then the segment does not have any row value that is null. In that case,
+     * we can just use a default initialized boolean vector.
+     */
+    const auto null_values =
+        _segment.null_values().has_value() ? _segment.null_values().value() : pmr_vector<bool>(_segment.size());
```
mrks: Why did you change this? `value()` makes no sense if you have already checked for its existence. How about
`const auto null_values = _segment.null_values() ? *_segment.null_values() : pmr_vector<bool>(_segment.size());

janehmueller: To keep it consistent with the other cases of accessing an `std::optional`'s value.

mrks: In the rest of the code base, there are more cases of the implicit bool and the dereference operator being used. Even better, I just remembered that `optional` has `value_or`:

`const auto null_values = _segment.null_values().value_or(pmr_vector<bool>(_segment.size()))`

(untested)

mrks: Ah, wait. That one would probably always create the vector. Maybe

`const auto null_values = _segment.null_values().value_or(_segment.size())`?

Haven't worked with that myself yet.

mrks: But then, grepping through the code base, I find more of those unnecessary uses of `value()`. Apparently, it's not as consistent as I hoped it would be.

mrks:
Also, thanks for updating the graphs. Those are quite interesting! Good work!

mrks:
Bring in the cake! :)

PR #1524, opened 2019-03-05
Allow ChunkPruningRule to prune past non-filtering operators 

### LINKED ISSUE #1509: ChunkPruningRule is kind of toothless for "realistic" queries
For the TPC-H benchmark, the ChunkPruningRule does almost never prune. The reason being that the ChunkPruningRule only works for a predicate chain that directly connects to a StoredTable node. Most queries have a ProjectionNode (e.g., introduced by the ColumnPruningRule) between Predicate and StoredTable nodes that in the end prevents the pruning of chunks.

mrzzzrm:
Fix #1509

### DIFF
--- a/src/lib/optimizer/strategy/chunk_pruning_rule.cpp
+++ b/src/lib/optimizer/strategy/chunk_pruning_rule.cpp
@@ -28,21 +28,20 @@ void ChunkPruningRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node) co
   // try to find a chain of predicate nodes that ends in a leaf
   std::vector<std::shared_ptr<PredicateNode>> predicate_nodes;
 
-  // Gather consecutive PredicateNodes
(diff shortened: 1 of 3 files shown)


src/lib/optimizer/strategy/chunk_pruning_rule.cpp
```diff
@@ -106,4 +105,8 @@ std::set<ChunkID> ChunkPruningRule::_compute_exclude_list(
   return result;
 }
 
+bool ChunkPruningRule::_is_non_filtering_node(const AbstractLQPNode& node) const {
```
Bensk1: Is that something we can recycle somewhere else?

mrzzzrm: The rule of thumb that I've grown to like is that if something is used three times it's okay to generalize it, everything below that is overengineering.

PR #1530, opened 2019-03-07
Remove tpch_tbl_header.py

mrzzzrm:
...as it was used nowhere and two "scripts" directories ("/scripts" and "/src/scripts") are too much imo.


Bouncner:
What was the purpose of this function? Instead of using TPCH dbgen "directly", rather write CSV file and prepend the header with this script?

mrks:
I guess. But if we are in that situation now, it's probably easier to simply copy those lines from existing tbl files.

Issue #1531, opened 2019-03-07
SegmentIterators are only LegacyInputIterators


cmfcmf:
After #1512, SegmentIterators became usable for standard algorithms. However, all SegmentIterators are still only LegacyInputIterators to the STL. As an example, `std::advance(it, 5)` will call `it++;` 5 times instead of calling `it += 5;`. This might also be the reason why sorted LZ4 segments perform so poorly: https://github.com/hyrise/hyrise/pull/1394#discussion_r263089195

It took me almost the whole day to figure out why that is the case: [LegacyForwardIterators require](https://en.cppreference.com/w/cpp/named_req/ForwardIterator):

> The type std::iterator_traits<It>::reference must be exactly 
> - T& if It satisfies LegacyOutputIterator (It is mutable)
> - const T& otherwise (It is constant), 
> (where T is the type denoted by std::iterator_traits<It>::value_type) 

And that's not the case for any of our SegmentIterators, because the `BaseSegmentIterator` is defined like this:
https://github.com/hyrise/hyrise/blob/5998cf202c2632ac89b1fea2169a69ca488a9665/src/lib/storage/segment_iterables/base_segment_iterators.hpp#L48-L50

Now look at the fourth parameter of `boost::iterator_facade`. This parameter denotes the "Reference" type, which is [described like this](https://www.boost.org/doc/libs/1_69_0/libs/iterator/doc/iterator_facade.html#id12):

> The Reference argument becomes the type returned by node_iterator's dereference operation, and will also be the same as std::iterator_traits<node_iterator>::reference. The library's default for this parameter is Value&; [...]

Our implementation, instead of specifying nothing or `Value&`, specifies just `Value`, which is in conflict with the STLs definition.

I've created a playground that illustrates the issue:

https://github.com/cmfcmf/hyrise/blob/iterator-issues/src/bin/playground.cpp

It creates an iterator for a ValueSegment and a custom iterator and prints the detected iterator types for both the STL and Boost. The custom iterator comes with 4 different possible definitions, out of which the first one matches our current `BaseSegmentIterator`.

When defining the iterator like this:
```cpp
class MyIterator : public boost::iterator_facade<MyIterator, int32_t, boost::random_access_traversal_tag, int32_t>,
                   public JitBaseSegmentIterator {
// ...
}
```

the result is:

```
# value segment iterator #

stl input
boost random

# custom iterator #

stl input
boost random
```

And when defining it like this:

```cpp
[...]

PR #1533, opened 2019-03-07
Remove JitBaseSegmentIterator

mrzzzrm:
Code smell, imo, to have a special base class of a core class, dedicated to a particular component. Thankfully it wasn't used.

### DIFF
--- a/src/lib/storage/segment_iterables/base_segment_iterators.hpp
+++ b/src/lib/storage/segment_iterables/base_segment_iterators.hpp
@@ -9,6 +9,14 @@
 namespace opossum {
 
 /**
+ * @brief template-free base class of all iterators used by iterables
+ *
+ * The class allows the JitOperatorWrapper to keep pointers to differently specialized versions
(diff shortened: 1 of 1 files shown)


mrks:
Mmh? Did you just re-add everything?

mrzzzrm:
If you need to access iterators polymorphically you can always wrap them - which I think is what the JIT does anyway now (cf. `JitSegmentReader`)

cmfcmf:
Ahh, I was wondering whether I was missing some sort of voodoo C++ JIT magic, because I couldn't find any usages of the `JitBaseSegmentIterator` when debugging segment iterators for #1531. Good to know it wasn't used :+1:

mrzzzrm:
> Ahh, I was wondering whether I was missing some sort of voodoo C++ JIT magic, because I couldn't find any usages of the JitBaseSegmentIterator when debugging segment iterators for #1531. Good to know it wasn't used

If in doubt, always ask - Issues are one option, PMs another!

PR #1537, opened 2019-03-10
Reduce ColumnVsColumnScan compile time

mrzzzrm:
Single-threaded release build with clang on pella: 57m39,206s (from 68m28,928s on the master)

Instead of always instanciating `#LeftDataType*#LeftSegmentType*#RightDataType+#RightSegmentType` versions of the inner lambda before deciding to type-erase, only instantiate `#LeftDataType*#RightDataType` (for the type-erased case) plus `#LeftDataType*#LeftSegmentType` (for the non-type erased case).

This PR makes it harder to switch back to never-erase in code, but those days are over for good imo.

### DIFF
--- a/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
@@ -67,7 +67,7 @@ std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::scan_chunk(ChunkID chunk_i
     });
 
     // `result` will still be nullptr if the SegmentTypes were not the same - if that's the case we have to take the
-    // "slow" further down
+    // "slow" path further down to perform the scan
     if (result) {
       return result;
     }
--- a/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
@@ -34,20 +34,25 @@ std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::scan_chunk(ChunkID chunk_i
 
   std::shared_ptr<PosList> result;
 
-  // Reducing the compile time:
-  //
-  // If the left and the right segment and/or data type are not the same, we erase the segment iterable types EVEN for
-  // the release build. For example, ValueSegment<int> == ValueSegment<float> will be erased. So will ValueSegment<int>
-  // == DictionarySegment<int>. ReferenceSegments do not need to be handled differently because we expect a table to
-  // either have only ReferenceSegments or non-ReferenceSegments.
-  //
-  // We use type erasure here because we currently do not actively use comparisons between, e.g., a ValueSegment and a
-  // DictionarySegment. While it is supported, it is not executed, so we don't want the compiler to spend time
-  // instantiating unused templates. Whenever the types of the iterators is removed, we also erase the comparator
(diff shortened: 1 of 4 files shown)


src/lib/types.hpp
```diff
@@ -219,7 +219,7 @@ enum class CleanupTemporaries : bool { Yes = true, No = false };
 // Used as a template parameter that is passed whenever we conditionally erase the type of a template. This is done to
 // reduce the compile time at the cost of the runtime performance. Examples are iterators, which are replaced by
 // AnySegmentIterators that use virtual method calls.
-enum class EraseTypes { OnlyInDebug, Always };
+enum class EraseTypes { OnlyInDebugBuild, Always };
```
mrzzzrm: For me, this reads nicer in code, but will revert without discussion if anyone thinks differently. :speak_no_evil:

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -29,41 +29,56 @@ std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsC
...
+      if (const auto right_typed_segment = std::dynamic_pointer_cast<SegmentType>(right_segment)) {
+        // Same segment types - do not erase types in Release builds
+        result = _typed_scan_chunk<EraseTypes::OnlyInDebugBuild>(
+            chunk_id, create_iterable_from_segment<ColumnDataType>(left_typed_segment),
+            create_iterable_from_segment<ColumnDataType>(*right_typed_segment));
       }
     });
+
+    if (result) {
+      return result;
```
mrks: Is it really expected for this to fall through into the next block? If yes, please doc.
Otherwise, I'd move the block below into an else-block. This is not really a guard clause.

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -29,41 +29,56 @@ std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsC
...
-            create_iterable_from_segment<RightType>(right_typed_segment));
-      } else {
-        PerformanceWarning("ColumnVsColumnTableScan using type-erased iterators");
-        result = _typed_scan_chunk<EraseTypes::Always>(chunk_id, create_any_segment_iterable<LeftType>(left_segment),
-                                                       create_any_segment_iterable<RightType>(right_segment));
+  // lambda by wrapping it into an std::function. All of this brought the compile time of this translation unit down
+  // significantly. This is only really relevant for the release build - in the debug build, iterators are always
+  // erased. The comparator type is being erased in the debug build as well.
+
+  if (left_segment->data_type() == right_segment->data_type()) {
```
mrks: Please add comments to the block if this is the erased or the non-erased path

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrzzzrm: Yes, it is expected to fall through, doc'ed: `// result will still be nullptr if the SegmentTypes were not the same - if that's the case we have to take the "slow" path further down to perform the scan`

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
mrzzzrm: Done

PR #1541, opened 2019-03-13
Fix SortMerge/NestedLoop joins for Outer Joins with <,<=,>,>= 

mrzzzrm:
Actually, I just wanted to decrease the compile time for the JNL (which is why there is some JNL refactoring in this PR already). Reading it I noticed it swapped the input tables, but did not flip the predicate condition when performing right outer joins. Writing a test for it, it emerged that JSM was broken, too. I am going to open a separate issue about the state of our join operator testing and what needs to be done about it.

@Bouncner should be relatively easy to review and is probably necessary for #1473


src/lib/operators/join_sort_merge.cpp
```diff
@@ -441,11 +441,11 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
   }
 
   /**
-  * Adds the rows without matches for left outer joins for non-equi operators (<, <=, >, >=).
+  * Adds the rows without matches for right outer joins for non-equi operators (<, <=, >, >=).
   * This method adds those rows from the left table to the output that do not find a join partner.
   * The outer join for the equality operator is handled in _join_runs instead.
   **/
-  void _left_outer_non_equi_join() {
+  void _right_outer_non_equi_join() {
```
Bouncner: Srsly? 😂

mrzzzrm: The code read as if it would be doing the exact opposite of what the name implied, so I switched the name and the tests passed. Didn't go deeper, though

Issue #848, opened 2018-04-25
Move code to hpp / lint


mrks:
I found some places where we have code in the hpp file even though it is not part of a template.

Two things:

* @mjendruk - Is there any reason why `SimdBp128Decompressor::get` is in the hpp file or is that just an oversight?
* Can we get the linter to complain about that?

mjendruk:
Yes, it is this way so that it can be inlined by the compiler.

mrks:
Let's build a benchmark that tests for this and see if it actually makes a difference. If not, let's move it to the cpp file.

mrks:
@nilsthamm - You mentioned that you already have made some changes for htis?

mweisgut:
Does anyone already know an automatic static analysis tool/method that is able to detect function definitions in header files? I currently try to find these header locations using the regular expression shown below. However, it's not really sufficient yet.

```
(^(?!( |\t)*template[^\n]*)|})\n( |\t)*\w+\s\w+\([^\n\t]*\)( const | ){
```

mrks:
This looks promising:

https://clang.llvm.org/extra/clang-tidy/checks/misc-definitions-in-headers.html

(#1901)

mrks:
Ah, no, it focuses on ODR and doesn't complain about code within classes.

Bouncner:
The SIMDBP128 part has been removed. Does anybody see value in this issue? @mweisgut @dey4ss

dey4ss:
Hm, we still have header files with non-templated function definitions, right? Not sure if there is a performance impact when moving them to cpp due to inlining. We could try and measure it though.

PR #1550, opened 2019-03-22
Safer handling of cross-type operations with lossless_[variant_]cast

### LINKED ISSUE #1306: Type conversions are incorrect
Currently, the scan converts the parameter into the type of the column that we scan on. This means that for an int column, the search parameter is first converted to an int - this is a problem if it was originally a long:

```
CREATE TABLE t (a INT, b LONG);
INSERT INTO t (a, b) VALUES (1000, 1000);
SELECT * FROM t WHERE a < 1000000000000
-- no results
SELECT * FROM t WHERE b < 1000000000000
-- one result
```

Pretty sure this is the case for other operators as well. We should have a careful look at what the SQL standard does, fix the issues, and write tests for mismatching data types.

In #1148, we have disabled a test `SELECT * FROM mixed WHERE a = 'a' AND c < 65.31;` that showed differing behavior for JIT and non-JIT.

### LINKED ISSUE #1525: Comparators (<, >=) ignore decimal places
**Query**
```sql
SELECT c_custkey FROM customer WHERE c_custkey < 5.5
```

**Actual result**
```
=== Columns
|c_custkey|
|      int|
| not null|
=== Chunk 0 ===
|<ReferS> |
|        1|
|        2|
|        3|
|        4|
===
4 rows total
```

**Expected result**
```
=== Columns
|c_custkey|
|      int|
| not null|
=== Chunk 0 ===
|<ReferS> |
|        1|
|        2|
|        3|
|        4|
|        5|
===
5 rows total
```

**Other cases**
The same happens for
```sql
SELECT c_custkey FROM customer WHERE c_custkey >= 5.5 AND c_custkey <= 10
```
but not for
```sql
SELECT c_custkey FROM customer WHERE c_custkey BETWEEN 5.5 AND 10
```

### LINKED ISSUE #1526: Inconsistent treatment of NULL from subqueries
**Actual behavior**
```sql
SELECT MIN(c_custkey) FROM customer WHERE c_custkey < -1
```
correctly returns `NULL`.

```sql
SELECT MIN(c_custkey) FROM customer WHERE c_custkey < (
  SELECT MIN(c_custkey) FROM customer WHERE c_custkey < -1
)
```
If I nest the query in itself, it still returns `NULL`.

```sql
SELECT MIN(c_custkey) FROM customer WHERE c_custkey < NULL
```
If I substitute the subquery, it breaks the Hyrise console:
```
  what():  Cannot convert from NullValue to anything but string
Aborted
```

**Expected behavior**
The second and third query should have the same result.

mrzzzrm:
Fix #1306
Fix #1525
Fix #1526 

This PR proposes to essentially remove `type_cast` and `type_cast_variant` from libhyrise.
`type_cast` incurs a possible loss of information, which has and will potentially lead to bugs, e.g, when scanning `<int_column> = type_cast<int>(16.25)` which will scan for "16" instead for "16.25".

To replace `type_cast`, this PR introduces `lossless_cast` which allows casting only when **no** information is lost by the cast and returns `std::nullopt` if no lossless conversion is possible. I hope that this forces people to **think** what they are casting instead of blindly using `type_cast`.

* `type_cast` is removed entirely as it is bug-enabling
* JoinHash of non-string with string columns was disabled out of convenience and because I think we should not support such obscure joins. This will also safe a bit of compile time.

```
| TPC-H 1        | 0.243358373642 | 15   | 0.24547201395  | 15   | +1%        |                          0.0056 |
| TPC-H 2        | 4.22778701782  | 254  | 4.23762273788  | 255  | +0%        |                          0.2459 |
| TPC-H 3        | 2.98765826225  | 180  | 3.03642725945  | 183  | +2%        |                          0.0000 |
| TPC-H 4        | 1.09118509293  | 66   | 1.093121171    | 66   | +0%        |                          0.3174 |
| TPC-H 5        | 2.0307199955   | 122  | 2.03820848465  | 123  | +0%        |                          0.5053 |
| TPC-H 6        | 10.2018680573  | 613  | 9.65229320526  | 580  | -5%        |                          0.0000 |
| TPC-H 7        | 0.62652438879  | 38   | 0.637421905994 | 39   | +2%        |                          0.0000 |
| TPC-H 8        | 2.29938220978  | 138  | 2.311160326    | 139  | +1%        |                          0.2153 |
| TPC-H 9        | 0.708157122135 | 43   | 0.706121623516 | 43   | -0%        |                          0.4542 |
| TPC-H 10       | 1.34893643856  | 81   | 1.34856569767  | 81   | -0%        |                          0.8702 |
| TPC-H 11       | 13.3296890259  | 800  | 13.2862958908  | 798  | -0%        |                          0.1436 |
| TPC-H 12       | 3.53764843941  | 213  | 3.60569214821  | 217  | +2%        |                          0.0000 |
| TPC-H 13       | 1.15354943275  | 70   | 1.16146063805  | 70   | +1%        |                          0.3989 |
| TPC-H 14       | 11.7027168274  | 703  | 11.5323457718  | 692  | -1%        |                          0.0000 |
| TPC-H 15       | 7.84137010574  | 471  | 7.9008808136   | 475  | +1%        |                          0.0000 |
| TPC-H 16       | 3.85413551331  | 232  | 3.87913537025  | 233  | +1%        |                          0.0009 |
| TPC-H 17       | 0.714478194714 | 43   | 0.714770495892 | 43   | +0%        |                          0.8116 |
| TPC-H 18       | 0.762093782425 | 46   | 0.771434724331 | 47   | +1%        |                          0.0497 |
| TPC-H 19       | 2.71924829483  | 164  | 2.69532871246  | 162  | -1%        |                          0.0000 |
| TPC-H 20       | 1.02411973476  | 62   | 1.02805352211  | 62   | +0%        |                          0.2342 |
| TPC-H 21       | 0.230422705412 | 14   | 0.232140690088 | 14   | +1%        |                          0.0054 |
| TPC-H 22       | 5.20759963989  | 313  | 5.18549728394  | 312  | -0%        |                          0.3454 |
| geometric mean |                |      |                |      | +0%        |                                 |
```


src/lib/lossless_cast.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+  // Safe casting from NULL to NULL is always NULL. (Cannot be handled below as resolve_data_type()
+  // doesn't resolve NULL)
+  if (source_data_type == DataType::Null && target_data_type == DataType::Null) {
+    return NullValue{};
+  }
+
+  // Safe casting between NULL and non-NULL type is not possible. (Cannot be handled below as resolve_data_type()
+  // doesn't resolve NULL)
+  if ((source_data_type == DataType::Null) != (target_data_type == DataType::Null)) {
```
mrks: I feel like the caller should have ruled out NULL before calling this method.

src/lib/lossless_cast.cpp
```diff
@@ -0,0 +1,37 @@
...
+    return std::nullopt;
+  }
+
+  std::optional<AllTypeVariant> result;
+
+  resolve_data_type(source_data_type, [&](auto source_data_type_t) {
```
mrks: why not standard visitor pattern?

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,229 @@
...
+ * "Lossless" means just that: No information must be lost during casting. Typical lossful operations would be
+ *      - float-to-int casts with the float having a non-zero fractional part
+ *      - long-to-int casts where the long was > MAX_INT
+ *      - string-to-int where the string contained no number, or a number > MAX_INT.
+ *      - double-to-float where the significand of the double has bits set that cannot be represented in float
+ *
+ * Whenever possible you should use these casts over "comfortable" casts such as static_cast or boost::lexical_cast.
+ * Those might incur information loss and might thus introduce bugs. See #1306, #1525.
+ *
+ * THINK ABOUT AN ALGORITHM'S DATA, THE POSSIBLE STATES IT MIGHT HAVE, AND HANDLE THEM PROPERLY.
```
mrks: This applies everywhere, doesn't it?

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,229 @@
...
+ */
+
+namespace opossum {
+
+inline std::tuple<bool, int32_t, uint32_t> decompose_floating_point(float f) {
```
mrks: 1. Why is this necessary?
2. "The value representation of floating-point types is implementation-defined" - This should be at least documented and statically(?) asserted.

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,229 @@
...
+std::enable_if_t<std::is_same_v<Target, Source>, std::optional<Target>> lossless_cast(const Source& source) {
+  return source;
+}
+
+// Long to Int
+template <typename Target, typename Source>
+std::enable_if_t<std::is_same_v<int64_t, Source> && std::is_same_v<int32_t, Target>, std::optional<Target>>
+lossless_cast(const Source& source) {
+  if (source < std::numeric_limits<int32_t>::min() || source > std::numeric_limits<int32_t>::max()) {
+    return std::nullopt;
```
mrks: What does SQL do in this case?

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// Int to Long
+template <typename Target, typename Source>
+std::enable_if_t<std::is_same_v<int32_t, Source> && std::is_same_v<int64_t, Target>, std::optional<Target>>
+lossless_cast(const Source& source) {
+  return static_cast<Target>(source);
+}
+
+// NULL to anything but NULL
```
mrks: Didn't you cover that in the cpp?

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,229 @@
...
+  auto fraction = (*bits) & 0xFFFFFFFFFFFFF;
+  auto sign = ((*bits) & 0x80000000) != 0;
+
+  return {sign, exponent, fraction};
+}
+
+// Identity
+template <typename Target, typename Source>
+std::enable_if_t<std::is_same_v<Target, Source>, std::optional<Target>> lossless_cast(const Source& source) {
+  return source;
```
mrks: Could universal references be beneficial here?

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// Anything but NULL to NULL
+template <typename Target, typename Source>
+std::enable_if_t<!std::is_same_v<NullValue, Source> && std::is_same_v<NullValue, Target>, std::optional<Target>>
+lossless_cast(const Source& source) {
+  return std::nullopt;
+}
+
+// String to Integral
```
mrks: is this the way SQL does it?

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,229 @@
...
+std::enable_if_t<std::is_integral_v<Source> && std::is_same_v<pmr_string, Target>, std::optional<Target>> lossless_cast(
+    const Source& source) {
+  return pmr_string{std::to_string(source)};
+}
+
+// Floating Point to String
+template <typename Target, typename Source>
+std::enable_if_t<std::is_floating_point_v<Source> && std::is_same_v<pmr_string, Target>, std::optional<Target>>
+lossless_cast(const Source& source) {
+  // TODO(moritz) find a lossless float-to-string converter
```
mrks: Should be impossible given that not all floats can be expressed as a terminating decimal

mrzzzrm: I know you are probably right, but could you give me a quick example?

src/lib/lossless_cast.cpp
mrzzzrm: na, they haven't. E.g., for a query `... WHERE a > NULL` the call to lossless_variant_cast is precisely where hyrise figures out that NULL has no representation in whatever data type `a` is.

src/lib/lossless_cast.hpp
mrzzzrm: lossless_**variant**_cast and lossless_cast are both entry points. 

Ideally I wouldn't need to handle NULL in lossless_**variant**_cast, but resolve_data_type doesn't (and probably shouldn't) resolve to DataType::Null, so lossless_**variant**_cast has to do an explicit check for it.

src/lib/lossless_cast.hpp
mrzzzrm: Hyrise will refuse to process `SELECT * FROM Table1 WHERE a > '3öp'` as "3öp" cannot be losslessly converted to integer.
MySQL will process `SELECT * FROM Table1 WHERE a > '3öp'` as `SELECT * FROM Table1 WHERE a > 3`. It will also process `SELECT * FROM Table1 WHERE a > 'öp'`, though, returning all tuples
Postgres will not process `SELECT * FROM Table1 WHERE a > '3öp'`, but behaves as Hyrise/MySQL does for `SELECT * FROM Table1 WHERE a > '3'`.

I don't know what the standard says, but there seems to be no consesus anyway. Hyrise now is strict and only allows string-int-conversion if all characters are processed and the int is in range.


(Funnily enough, postgres allows `SELECT * FROM Table1 WHERE a < 123456789123456789123456789123456789` but not `SELECT * FROM Table1 WHERE a < '123456789123456789123456789123456789'`, complaining the number is no integer.)

src/lib/lossless_cast.hpp
mrzzzrm: idk, to what benefit? This is in a header, I'd expect this to get inlined anyway?

src/lib/lossless_cast.cpp
mrzzzrm: I think this looks ugly, with FIRST the visitor arg and THEN the variant, but the difference is debatable, so I changed it:

```c++
  boost::apply_visitor([&](const auto& source_value) {
    using SourceDataType = std::decay_t<decltype(source_value)>;

    resolve_data_type(target_data_type, [&](auto target_data_type_t) {
      using TargetDataType = typename decltype(target_data_type_t)::type;
      const auto source = boost::get<SourceDataType>(variant);
      result = lossless_cast<TargetDataType>(source);
    });
  }, variant);
```

src/lib/lossless_cast.hpp
mrzzzrm: Ah, sure... I guess I was just surprised by what I perceived as careless usage of the lossy type_cast all over the code base.  
Removed the comment.

src/lib/lossless_cast.hpp
mrzzzrm: 1. To check whether the float can be losslessly converted to an integer. I have searched a bit and found - to my surprise and anger - no premade function that does this. 

TBH, the float->int conversion is the shakiest part of this PR and requires more thought and testing.

Can you do `f <= INT_MAX && f >= INT_MIN` followed by a `static_cast<int>(f)`? `f <= INT_MAX` will implicitly cast INT_MAX to float to perform the comparison, as I understand it. If roundup happens during this conversion `f <= INT_MAX` will return true for some floats that are larger than INT_MAX. Actually looking at the bits seemed the best way around this.

I COULD also study IEEE754 with respect to rounding from int to float, maybe `f <= INT_MAX && f >= INT_MIN` is guaranteed to work. 

2. Added `static_assert(std::numeric_limits<float>::is_iec559, "IEEE 754 floating point representation expected.");`

src/lib/lossless_cast.hpp
mrzzzrm: Would need to check but let me be clear first: lossless_cast is NOT there to mirror SQL conversions in any way. It really just checks whether a conversion can be performed losslessly and does so if possible. Doesn't care about SQL at all.

src/lib/lossless_cast.cpp
mrks: What do you need `SourceDataType` for? Isn't `source_value` already the type that you want?

src/lib/lossless_cast.hpp
mrks: Avoiding a string copy. Just a quick thought though.

src/lib/lossless_cast.cpp
mrzzzrm: you're right, removing that makes apply_visitor indeed shorter than resolve_data_type

src/lib/lossless_cast.hpp
mrks: https://stackoverflow.com/questions/1089018/why-cant-decimal-numbers-be-represented-exactly-in-binary

mrks: Or try entering .1 here: https://www.h-schmidt.net/FloatConverter/

mrzzzrm: this function is about FLOAT to STRING, not the other way round! Of course I know that 0.1 has no exact representation in binary...

mrzzzrm: ...but is there any binary float that has no exact representation in decimal. E.g., is there an `n` for which `2^(-n)` is no representable in decimal?

mrzzzrm: 2^(-n) always ends with 5 in decimal, so 2^(-n-1) should, too.

mrks: How would you express the float representation of .1 as a string?

mrks: Ah, guess I'm wrong. Still, I am not sure if we want to start staring into that abyss: https://stackoverflow.com/questions/7153979/algorithm-to-convert-an-ieee-754-double-to-a-string

mrzzzrm: >  I am not sure if we want to start staring into that abyss: 

Absolutely not sure, either. Which is why I left it open for the moment. I would just declare string-to-float as too cumbersome to do losslessly and thus leave this specialization returning std::nullopt.

src/lib/lossless_cast.hpp
mrks: I haven't had a coffee yet, so this might be stupid but why don't we cast the number in question into the other type, cast it back, and see if it changed?

mrzzzrm: I did it like that before. You'll get UBSan, rightfully, complaining about casts from floats >INT_MAX to int, e.g. See my previous answer why I am hesitant to check for f > INT_MAX.

mrks: Can't you exclude that case by comparing it to `< (float)(INT_MAX)` first?

mrzzzrm: After some more thought I couldn't find a case where a check `f < (float)INT_MAX` yields to false positives/negatives do to conversion. The float-decomposition, thus, was removed. Much cleaner.

src/lib/lossless_cast.hpp
mrzzzrm: I decided we do not want to stare into that abyss, lossless float-string conversion is not supported.

src/lib/lossless_cast.hpp
mrzzzrm: Nope, doesn't work, here is why:

```c++
#include <limits>
#include <iostream>

int main(int argc, char **argv) {
  // INT_MAX is 2'147'483'647

  std::cout << "INT_MAX is " << std::numeric_limits<int32_t>::max() << std::endl;
  std::cout << "INT_MAX casted to float is " << std::fixed << static_cast<float>(std::numeric_limits<int32_t>::max()) << std::endl;
  std::cout << "Thus 2'147'483'648.0f, which should be greater than INT_MAX, is actually not greater: " << (2'147'483'648.0f > std::numeric_limits<int32_t>::max()) << std::endl;
  std::cout << "So any lossless_cast algorithm based on comparison would think it can convert 2'147'483'648.0f to a 32bit int, but it actually cannot." << std::endl;

}
```

2'147'483'648.0f, which is greater than INT_MAX, would not be detected as such.

mrzzzrm:
`type_cast` now removed entirely, I think we are better off without it, given how easy it is to produce bugs with it.

mrzzzrm:
As soon as CI passes I am happy with this. The ieee754-decomposition is a bit cumbersome but, as the tests show, it works pretty well.

src/lib/lossless_cast.hpp
mrks: Mmh, how about checking if `modf` returns `0.0`?

mrzzzrm: that would still not tell us whether the float is bigger than INT_MAX and thus doesn't tell us whether the `static_cast<int>(some_float)` is safe...

```c++
  auto input_float = 2'147'483'648.0f;

  float integral;
  auto fractional = std::modf(input_float, &integral);

  std::cout << fractional << std::endl;
  std::cout << integral << std::endl;
  std::cout << (integral > std::numeric_limits<int32_t>::max()) << std::endl;
  std::cout << (input_float > std::numeric_limits<int32_t>::max()) << std::endl;
```

gives 

```
0
2.14748e+09
0
0
```

mrks: Haha, well - the problem is that `std::numeric_limits<int32_t>::max()` cannot be expressed as a float. `std::numeric_limits<int32_t>::max() + 1`, however, can:

```
#include <cmath>
#include <limits>
#include <iostream>

bool is_int(float f) {
	float integral_part;
	if(std::modf(f, &integral_part) != 0.0) return false;
	if(integral_part >= (float)0x80000000u) return false;
	// TODO handle the negative side here
	return true;
}

int main(int argc, char **argv) {
  for (auto f : {5.0f, 5.5f, 2'147'483'647.0f, 2'147'483'648.0f, std::nexttoward(2'147'483'648.0f, 0.f)}) {
  	std::cout << std::fixed << f << " " << is_int(f) << std::endl;
  }
}
```

```
5.000000 1
5.500000 0
2147483648.000000 0
2147483648.000000 0
2147483520.000000 1
```

mrzzzrm: > the problem is that std::numeric_limits<int32_t>::max() cannot be expressed as a float

Yes, that's what I've been trying to express the entire time - I thought the whole discussion in this thread and to some degree the whole purpose of this PR is just that: There are floats that aren't representable as ints, there are ints that aren't representable as float...

mrks: Sure, I'm just trying to find a way around having to play with the bit representation of the float. Maybe it's easier to talk about this in person.

mrks: Let me just add one more thought though, lest I forget:

> There are floats that aren't representable as ints

Yes, but those are all outside of [INT_MIN, INT_MAX]. Checking if the float is within those boundaries and has no fractional part should thus be sufficient.

mrzzzrm: What do you think of this?

```c++
// Floating Point Type to 32/64 bit Integral Type
template <typename Target, typename Source>
std::enable_if_t<std::is_floating_point_v<Source> && (std::is_same_v<Target, int32_t> || std::is_same_v<Target, int64_t>), std::optional<Target>>
lossless_cast(const Source& source) {
  static_assert(std::numeric_limits<float>::is_iec559, "IEEE 754 floating point representation expected.");

  auto integral_part = Source{};

  // No lossless float-to-int conversion possible if the source float has a fractional part
  if (std::modf(source, &integral_part) != 0.0) {
    return std::nullopt;
  }

  // Check the floating point value against explicitly defined boundary values. The boundary values are the
  // the closest-to-zero floating point values that are NOT representable in the given integral type.
  // (Note: We evaluated multiple approaches to identify whether a float/double is representable as a int32/64. None 
  //        of them was really portable/readable/"clean". The bounding values approach at least makes intuitive sense.) 
  if constexpr (std::is_same_v<Source, float> && std::is_same_v<Target, int32_t>) {
    if (source >= 2'147'483'648.0f || source <= -2'147'483'904.0f) return std::nullopt;
  } else if constexpr (std::is_same_v<Source, double> && std::is_same_v<Target, int32_t>) {
    if (source >= 2'147'483'648.0 || source <= -2'147'483'649.0) return std::nullopt;
  } else if constexpr (std::is_same_v<Source, float> && std::is_same_v<Target, int64_t>) {
    if (source >= 9'223'372'036'854'775'808.0f || source <= -9'223'373'136'366'403'584.0f) return std::nullopt;
  } else if constexpr (std::is_same_v<Source, double> && std::is_same_v<Target, int64_t>) {
    if (source >= 9'223'372'036'854'775'808.0 || source <= -9'223'372'036'854'777'856.0) return std::nullopt;
  }

  return static_cast<Target>(source);
}
```

mrzzzrm:
@FabianWiebe the JIT cannot handle some SQL queries I added in this PR, e.g. 
```
===================== Tables are not equal =====================
------------------------- Actual Result ------------------------
          id        a        b        c        d 
         int   string      int    float   string 
   0       5        a       11    43.55    ahpll 
   1       6        c       42    18.33     qunl 
   2       7        c       49    10.92  smpizda 
   3       8        b       16    28.25     caoe 
----------------------------------------------------------------

------------------------ Expected Result -----------------------
          id        a        b        c        d 
         int   string      int   double   string 
   0       6        c       42    18.33     qunl 
   1       7        c       49    10.92  smpizda 
   2       8        b       16    28.25     caoe 
----------------------------------------------------------------
Type of error: Row count mismatch
================================================================

Actual number of rows: 4
Expected number of rows: 3

/home/moritz/Coding/hyrise/src/test/sql/sqlite_testrunner/sqlite_testrunner.cpp:184: Failure
Value of: check_table_equal(result_table, sqlite_result_table, order_sensitivity, TypeCmpMode::Lenient, FloatComparisonMode::RelativeDifference)
  Actual: false
Expected: true
Query failed: SELECT * FROM mixed WHERE id BETWEEN 5.5 AND 8;
Google Test trace:
/home/moritz/Coding/hyrise/src/test/sql/sqlite_testrunner/sqlite_testrunner.cpp:154: Query 'SELECT * FROM mixed WHERE id BETWEEN 5.5 AND 8;' with JIT and encoding Dictionary
```

It casts 5.5 down to 5 and then obviously generates a different result set. Do you think a quick fix within this PR would be possible?

mrzzzrm:
@FabianWiebe I have disabled the JIT for expressions like `int > double`. #1606

FabianWiebe:
@mrzzzrm: I fixed the jit code on my branch for now `fabian/safe_type_cast`: ad4214040e29afbf2240c5cbc517a2580f8647dd
However, I do not have time to write tests for this in the next days. I can address this next week.
The question is whether my new code should also be used by the normal table scan as it allows to exploit dictionary-encoding in more places.
For example, the table scan on int column < `2.4` will use the expression evaluator without exploiting dictionary-encoding. My new code now updates `2.4` to `3` and uses dictionary-encoding.

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,214 @@
...
+std::enable_if_t<std::is_same_v<pmr_string, Source> && std::is_integral_v<Target>, std::optional<Target>> lossless_cast(
+    const Source& source) {
+  static_assert(std::is_same_v<int32_t, Target> || std::is_same_v<int64_t, Target>, "Expected int32_t or int64_t");
+
+  // We don't want std::stol's exceptions to occur when debugging, thus we're going the c-way: strtol, which
+  // communicates conversion errors via errno. See http://c-faq.com/misc/errno.html on why we're setting `errno = 0`
+  errno = 0;
+  char* end;
+
+  const auto integral = std::strtol(source.c_str(), &end, 10);
```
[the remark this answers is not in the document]
mrzzzrm: switched to boost::lexical_cast

mrzzzrm:
I am happy with this PR - that the JIT doesn't translate `<int_column> > 5.5` is okay. Certainly for now, probably forever. The non-fallback TableScans cannot handle it either.

mrzzzrm:
> I am surprised that this problem does not occur in the non-JIT code.

It certainly would. We should move towards not-allowing float-vs-double comparisons - or at least issue a big warning when they are used.

mrks:
Do you have TPC-H results for this?

mrzzzrm:
Generating them...

src/lib/optimizer/strategy/chunk_pruning_rule.cpp
```diff
@@ -87,21 +88,47 @@ std::set<ChunkID> ChunkPruningRule::_compute_exclude_list(
   std::set<ChunkID> result;
 
   for (const auto& operator_predicate : *operator_predicates) {
+    // Cannot prune column-to-column predicates atm. Column-to-placeholder predicates are never prunable.
```
mrks: atm?

src/lib/optimizer/strategy/chunk_pruning_rule.cpp
```diff
@@ -87,21 +88,47 @@ std::set<ChunkID> ChunkPruningRule::_compute_exclude_list(
...
+    // If `value` cannot be converted losslessly to the column data type, we rather skip pruning than running into
+    // errors with lossful casting and pruning Chunks that we shouldn't have pruned.
+    auto value = lossless_variant_cast(boost::get<AllTypeVariant>(operator_predicate.value), column_data_type);
+    if (!value) {
+      continue;
+    }
+
+    auto value2 = std::optional<AllTypeVariant>{};
+    if (operator_predicate.value2) {
+      // Cannot prune column-to-column predicates atm. Column-to-placeholder predicates are never prunable.
```
mrks: atm?

src/lib/statistics/column_statistics.cpp
```diff
@@ -26,52 +26,55 @@ std::shared_ptr<BaseColumnStatistics> ColumnStatistics<ColumnDataType>::clone()
 template <typename ColumnDataType>
 FilterByValueEstimate ColumnStatistics<ColumnDataType>::estimate_predicate_with_value(
     const PredicateCondition predicate_condition, const AllTypeVariant& variant_value,
-    const std::optional<AllTypeVariant>& value2) const {
-  const auto value = type_cast_variant<ColumnDataType>(variant_value);
+    const std::optional<AllTypeVariant>& variant_value2) const {
+  const auto value = lossless_variant_cast<ColumnDataType>(variant_value);
+  if (!value) return {non_null_value_ratio(), without_null_values()};
```
mrks: Does this mean that all comparisons that use different types will have no statistics? Should we care if we lose the double precision here?

src/lib/storage/lz4/lz4_iterable.hpp
```diff
@@ -48,7 +48,9 @@ class LZ4Iterable : public PointAccessibleSegmentIterable<LZ4Iterable<T>> {
     auto cached_block_index = std::optional<size_t>{};
     for (auto index = size_t{0u}; index < position_filter->size(); ++index) {
       const auto& position = (*position_filter)[index];
-      auto [value, block_index] = _segment.decompress(position.chunk_offset, cached_block_index, cached_block);  // NOLINT
+      // NOLINTNEXTLINE
+      auto [value, block_index] =
+          _segment.decompress(position.chunk_offset, cached_block_index, cached_block);  // NOLINT
```
mrks: This NOLINT can go, right?

src/lib/storage/table.hpp
```diff
@@ -135,7 +136,7 @@ class Table : private Noncopyable {
       size_t current_size = chunk->size();
       row_counter += current_size;
       if (row_counter > row_number) {
-        return get<T>(
+        return boost::get<T>(
```
mrks: @Bensk1 and I spend a while on Thursday trying to understand why a similar thing worked in student code :D

src/lib/utils/load_table.cpp
```diff
@@ -57,11 +60,15 @@ std::shared_ptr<Table> load_table(const std::string& file_name, size_t chunk_siz
...
     for (auto column_id = ColumnID{0}; column_id < string_values.size(); ++column_id) {
-      if (table->column_is_nullable(column_id) && string_values[column_id] == "null") {
-        variant_values[column_id] = NULL_VALUE;
-      } else {
-        variant_values[column_id] = AllTypeVariant{pmr_string{string_values[column_id]}};
-      }
+      resolve_data_type(table->column_data_type(column_id), [&](auto data_type_t) {
+        using ColumnDataType = typename decltype(data_type_t)::type;
+
+        if (table->column_is_nullable(column_id) && string_values[column_id] == "null") {
```
mrks: Move out of the lambda?

src/lib/all_type_variant.hpp
```diff
@@ -119,6 +119,16 @@ static const auto NULL_VALUE = AllTypeVariant{};
   BOOST_PP_SEQ_FOR_EACH(EXPLICIT_INSTANTIATION, template_class, DATA_TYPES) \
   static_assert(true, "End call of macro with a semicolon")
 
+/**
+ * This function returns the DataType of an AllTypeVariant
+ *
+ * Note: DataType and AllTypeVariant are defined in a way such that
+ *       the indices in DataType and AllTypeVariant match.
+ */
+inline DataType data_type_from_all_type_variant(const AllTypeVariant& all_type_variant) {
```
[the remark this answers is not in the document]
mrzzzrm: Not from this PR, but it'd bet it's in the hpp for "performance reasons". idc, but I can move it if you want to.

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+// String to Integral
+template <typename Target, typename Source>
+std::enable_if_t<std::is_same_v<pmr_string, Source> && std::is_integral_v<Target>, std::optional<Target>> lossless_cast(
+    const Source& source) {
+  static_assert(std::is_same_v<int32_t, Target> || std::is_same_v<int64_t, Target>, "Expected int32_t or int64_t");
+
+  Target result{};
+
+  if (boost::conversion::detail::try_lexical_convert(source, result)) {
```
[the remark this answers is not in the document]
mrzzzrm: as the tests (which include numbers outside of int) all pass, I'd say: yes. Is that good enough or want me to crawl the boost docs (which I'd prefer not having to do)?

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,211 @@
...
+      return std::nullopt;
+    }
+  }
+}
+
+template <typename Target>
+std::optional<Target> lossless_variant_cast(const AllTypeVariant& variant) {
+  std::optional<Target> result;
+
+  const auto source_data_type = data_type_from_all_type_variant(variant);
```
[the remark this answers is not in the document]
mrzzzrm: would incur unneeded overhead on the commented-on function, but I can change it.

src/lib/operators/join_hash.cpp
```diff
@@ -89,10 +88,26 @@ std::shared_ptr<const Table> JoinHash::_on_execute() {
...
+    resolve_data_type(probe_input->column_data_type(probe_column_id), [&](const auto probe_data_type_t) {
+      using ProbeColumnDataType = typename decltype(probe_data_type_t)::type;
+
+      constexpr auto BOTH_ARE_STRING =
+          std::is_same_v<pmr_string, BuildColumnDataType> && std::is_same_v<pmr_string, ProbeColumnDataType>;
+      constexpr auto NEITHER_IS_STRING =
+          !std::is_same_v<pmr_string, BuildColumnDataType> && !std::is_same_v<pmr_string, ProbeColumnDataType>;
+
+      if constexpr (BOTH_ARE_STRING || NEITHER_IS_STRING) {
+        _impl = std::make_unique<JoinHashImpl<BuildColumnDataType, ProbeColumnDataType>>(
```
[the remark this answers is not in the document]
mrzzzrm: less. string-vs-non-string is not compiled anymore.

src/lib/all_type_variant.hpp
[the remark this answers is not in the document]
mrks: Nah, if we ever get to #848, we'll check if this makes sense.

src/lib/lossless_cast.hpp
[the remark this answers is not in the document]
mrks: If it's tested, I'm fine if you just add a comment saying so.

src/lib/lossless_cast.hpp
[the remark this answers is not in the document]
mrks: Had another look, you're right.

src/lib/statistics/column_statistics.cpp
mrzzzrm: One of the rougher edges of this PR, I agree. But I really wanted to remove type_cast and the bug proneness that it brings and was ready to sacrifice estimations on non-equal operand types.

There is a (potential) follow-up to this PR: `std::pair<PredicateCondition, Target> predicate_operand_cast<Target>(Source, PredicateCondition)` (exact name to be determined). This will take, e.g., a predicate `<int_column> > 5.5` and turn it into `<int_column> > 5` or `<int_column> >= 5.3` into `<int_column> >= 6`.

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+/**
+ * This file defines the following casting functions that guarantee **lossless** conversion between data types and
+ * return std::nullopt if no lossless conversion was possible.
+ *      lossless_cast<Target>(Source)
+ *      lossless_variant_cast<Target>(AllTypeVariant)
+ *      lossless_variant_cast(Source, DataType)
+ *
+ * "Lossless" means just that: No information must be lost during casting. Typical lossful operations would be
+ *      - float-to-int casts with the float having a non-zero fractional part
```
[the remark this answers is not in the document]
mrzzzrm: Covered in the tests: 

```c++
  EXPECT_EQ(lossless_cast<int32_t>(std::numeric_limits<float>::quiet_NaN()), std::nullopt);
  EXPECT_EQ(lossless_cast<int32_t>(std::numeric_limits<double>::quiet_NaN()), std::nullopt);
  EXPECT_EQ(lossless_cast<int32_t>(std::numeric_limits<float>::infinity()), std::nullopt);
  EXPECT_EQ(lossless_cast<int32_t>(std::numeric_limits<double>::infinity()), std::nullopt);
```

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,211 @@
...
+ *      - long-to-int casts where the long was > MAX_INT
```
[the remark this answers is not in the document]
mrzzzrm: those are just meant to be examples ("typical [...] would be:")

src/lib/statistics/column_statistics.cpp
mrks: With the number of open issues (and the histogram PR touching on this as well), I am not sure if we should focus on that or wait for that. The statistics will always be a best-effort thing, and accepting type conversion errors here will certainly be better than not returning any statistics at all.

mrks: Maybe we don't need all back-and-forths that the old type_cast had, but if static_cast can deal with it, we should be able to use it.

src/lib/lossless_cast.hpp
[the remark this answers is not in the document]
mrks: I read that, but still included this comment as it might also confuse other readers. Also, why "was"? How about "is not inside [MIN_INT, MAX_INT]?

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+  if (boost::conversion::detail::try_lexical_convert(source, result)) {
+    return result;
+  } else {
+    return std::nullopt;
+  }
+}
+
+// String to Floating Point
+// NOT SUPPORTED: Some strings (e.g., "5.5") have lossless float representations, others (e.g., "5.3") do not. Allowing
```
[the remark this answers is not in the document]
mrzzzrm: "All over the place"? And now I am left guessing what actually bothers you. 

(I all-capped "NOT SUPPORTED" because I judged the following comment to be more important than the others in this file.)

src/lib/operators/jit_operator/operators/jit_read_tuples.cpp
```diff
@@ -257,9 +257,18 @@ bool JitReadTuples::before_chunk(const Table& in_table, const ChunkID chunk_id,
...
-        resolve_data_type(jit_input_column.tuple_entry.data_type, [&](const auto current_data_type_t) {
-          using CurrentType = typename decltype(current_data_type_t)::type;
-          casted_value = type_cast_variant<CurrentType>(value);
+        resolve_data_type(jit_input_column.tuple_entry.data_type, [&](const auto column_data_type_t) {
+          using ColumnDataType = typename decltype(column_data_type_t)::type;
+
+          resolve_data_type(data_type_from_all_type_variant(value), [&](const auto value_data_type_t) {
+            using ValueDataType = typename decltype(value_data_type_t)::type;
+
+            if constexpr (std::is_same_v<ColumnDataType, pmr_string> == std::is_same_v<ValueDataType, pmr_string>) {
```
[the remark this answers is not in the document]
mrzzzrm: since ValueDataType is needed for the boost::get<> anyway.

src/lib/optimizer/strategy/chunk_pruning_rule.cpp
mrzzzrm: "Cannot prune column-to-column predicates" would mean there is some fundamental impossibility in pruning chunks with column-vs-column predicates - which is not the case (though it might be complicated) 

Add the "atm" and it becomes clear that the current code base of hyrise just doesn't do it.

How would you formulate it?

mrzzzrm: or did you take issue with the abbreviation? i am all for compact commenting but not if it leaves me guessing too much ;)

src/lib/lossless_cast.hpp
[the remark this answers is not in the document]
mrks: "String" vs "strings"
"Floating Point" vs "floating point" vs "float"

src/lib/optimizer/strategy/chunk_pruning_rule.cpp
mrks: > or did you take issue with the abbreviation?

Yepp, sorry.

src/lib/operators/jit_operator/operators/jit_read_tuples.cpp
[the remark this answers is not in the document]
mrks: Ah. Maybe it's just me, but I'm usually misreading `bool == bool` on the first attempt. Maybe add a comment on what you are checking?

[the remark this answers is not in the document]
mrzzzrm: I am tempted to quote your famous "every programmer..." quote. bool == bool should, imo, be so basic knowledge that I do not feel it should be documented every time it's used.

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -138,7 +137,7 @@ RadixContainer<T> materialize_input(const std::shared_ptr<const Table>& in_table
           ++it;
 
           if (!value.is_null() || retain_null_values) {
-            const Hash hashed_value = hash_function(type_cast<HashedType>(value.value()));
+            const Hash hashed_value = hash_function(static_cast<HashedType>(value.value()));
```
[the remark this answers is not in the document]
mrzzzrm: We should be safe here. JoinHash always picks a "big enough" HashedType. I.e. for a double-vs-float scan, HashedType would be double and no conversion loss would happen.

src/lib/lossless_cast.hpp
```diff
@@ -0,0 +1,211 @@
...
+  } else if constexpr (std::is_same_v<Source, double> && std::is_same_v<Target, int64_t>) {
+    if (source >= 9'223'372'036'854'775'808.0 || source <= -9'223'372'036'854'777'856.0) return std::nullopt;
+  }
+
+  return static_cast<Target>(source);
+}
+
+// float to double
+template <typename Target, typename Source>
+std::enable_if_t<std::is_same_v<float, Source> && std::is_same_v<double, Target>, std::optional<Target>> lossless_cast(
```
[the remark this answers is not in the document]
mrzzzrm: I'd love to change it, but I cannot conclusively resolve the c++ template errors that ensue once I do.

src/lib/lossless_cast.hpp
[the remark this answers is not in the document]
mrzzzrm: hopefully better now. "floating point" (which includes double and float) and "float" is something different intentionally, btw.

src/lib/statistics/column_statistics.cpp
mrzzzrm: Normally I am all for merging PRs fast and opening follow-up issues to keep my pipeline small.

But cases where actual functionality is (temporarily) removed by a PR shouldn't happen. I introduced `static_variant_cast` as a local helper for the statistics. If I need it in #1489 I'll refactor it into its own file.

src/lib/operators/join_hash/join_hash_steps.hpp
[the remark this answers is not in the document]
mrzzzrm: double-vs-long is the only case I can think of (HashedType being double in this case) where a loss would occur. I'd leave it as it is for now - just to be clear: this PR doesn't remove all conversion losses from hyrise, aside from the double-vs-long join, the ExpressionEvaluator is also heavily lossy (e.g. comparing float with int) and full of potential UB (e.g. INT_MAX + INT_MAX). This PR just removes some information-loss related bugs.

src/lib/operators/jit_operator/operators/jit_read_tuples.cpp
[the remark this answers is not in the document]
mrks: Won't keep me up at night ;)

src/lib/operators/join_hash/join_hash_steps.hpp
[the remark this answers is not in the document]
mrks: Ok. Maybe add what you just said as a comment?

src/lib/lossless_cast.hpp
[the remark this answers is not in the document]
mrks: Not important enough to spend more than a couple of minutes on.

src/lib/lossless_cast.cpp
```diff
@@ -7,7 +7,7 @@ namespace opossum {
 std::optional<AllTypeVariant> lossless_variant_cast(const AllTypeVariant& variant, DataType target_data_type) {
   const auto source_data_type = data_type_from_all_type_variant(variant);
 
-  // Safe casting from NULL to NULL is always NULL. (Cannot be handled below as resolve_data_type()
+  // Lossless castin from NULL to NULL is always NULL. (Cannot be handled below as resolve_data_type()
```
mrks: typo

src/lib/lossless_cast.hpp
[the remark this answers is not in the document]
mrks: Absolutely.

> "floating point" (which includes double and float) and "float" is something different intentionally, btw.

Yes, which is why the case was more confusing.

mrzzzrm:
Perf looks okay (see OP). TPC-H 6 uses the ExpressionEvaluator for one scan, wouldn't know how to avoid that.

image

src/lib/operators/join_hash/join_hash_steps.hpp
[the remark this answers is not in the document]
mrzzzrm: static_cast to HashedType is used in 4 places. Where should I put the comment?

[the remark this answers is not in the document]
mrzzzrm: I am not sure a comment helps at all. I'd categorize the fact that we're inaccurate when it comes to type-interoperability as general hyrise-knowledge that's documented in our collective consciousness. ;)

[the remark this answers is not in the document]
mrks: One of those four places, preferably the first, should be fine. It's not perfect, but better than nothing.

mrzzzrm:
Okay, this was to be expected: UB triggered in the statistics, now that they use static_cast again:

```
[ RUN      ] SQLiteTestRunnerUnencoded/SQLiteTestRunner.CompareToSQLite/42

../../../src/lib/statistics/column_statistics.cpp:23:36: runtime error: 1.23457e+39 is outside the range of representable values of type 'float'

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior ../../../src/lib/statistics/column_statistics.cpp:23:36 in 
```

Without getting pretty involved, lossless_cast (instead of static_cast) seems to be the only way?

mrzzzrm:
Reverted back, but feel free to propose another solution

mrks:
Any reason we can't check for `> float_max`?

FabianWiebe:
> Perf looks okay (see OP). TPC-H 6 uses the ExpressionEvaluator for one scan, wouldn't know how to avoid that.

 In this query the double `0.042095` or `0.082105` cannot be lossless casted to a float so that the `ExpressionEvaluator` is used. This can be avoided if the predicate condition is known during casting as then the next representable float number can be used.

mrzzzrm:
@FabianWiebe Yep, as mentioned here: https://github.com/hyrise/hyrise/pull/1550#discussion_r277177632 - I am probably going to create an issue about that soon.

src/lib/statistics/column_statistics.cpp
```diff
@@ -2,9 +2,46 @@
...
+    if constexpr (std::is_same_v<Target, SourceDataType>) {
+      result = boost::get<SourceDataType>(source);
+    } else {
+      if constexpr (std::is_same_v<pmr_string, SourceDataType> == std::is_same_v<pmr_string, Target>) {
+        const auto source_value = boost::get<SourceDataType>(source);
+        if (source_value <= std::numeric_limits<Target>::max() &&
+            source_value >= std::numeric_limits<Target>::lowest()) {
+          result = static_cast<Target>(boost::get<SourceDataType>(source));
+        } else {
+          std::cout << "Fuck" << std::endl;
```
mrks: ;)

Either use those max values or return empty statistics.

mrzzzrm: Sorry. Doing hacky fixes before my first coffee....

mrzzzrm: empty statistics it is

mrzzzrm: changed my mind and went for min/max values

PR #1551, opened 2019-03-22
Remove unused return value of TransactionContext::commit()/rollback()

### LINKED ISSUE #1393: Transactions with failed operators can be committed

d-justen:
This commit is originated by #1393 and its former purpose was to make the result of the `commit` method in `TransactionContext` non-discardable. This has changed to make `commit`, `rollback` and others void, since we have found out that there is currently no way that these methods can fail, if the operators executed successfully and if the transaction is committed when it's active.

fixes #1393


src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -194,8 +194,17 @@ const std::shared_ptr<const Table>& SQLPipelineStatement::get_result_table() {
...
+    _mark_as_failed();
+    return _result_table;
+  }
+
   if (_auto_commit) {
-    _transaction_context->commit();
+    const auto success = _transaction_context->commit();
+    if (!success) {
+      _mark_as_failed();
+      return _result_table;
```
mrks: This means `return nullptr`, correct? Rather be explicit.Couple of lines above as well.

src/lib/sql/sql_pipeline_statement.hpp
```diff
@@ -71,7 +71,11 @@ class SQLPipelineStatement : public Noncopyable {
 
   const std::shared_ptr<SQLPipelineStatementMetrics>& metrics() const;
 
+  bool has_failed() const;
+
  private:
+  void _mark_as_failed();
```
mrks: the changes to this and the pipelineotself seem untested.

src/lib/server/server_session.cpp
```diff
@@ -265,7 +265,8 @@ template <typename TConnection, typename TTaskRunner>
 boost::future<void> ServerSessionImpl<TConnection, TTaskRunner>::_handle_sync_command() {
   if (!_transaction) return boost::make_ready_future();
 
-  _transaction->commit();
+  const auto success = _transaction->commit();
+  Assert(success, "Commit has failed.");
```
mrks: Transaction conflicts are, to some degree, expected in multi-threaded environments. Killing the server is thus not thecorrect answer. The server should respond with whatever psql does in this case. Also, this should be tested.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -194,8 +194,17 @@ const std::shared_ptr<const Table>& SQLPipelineStatement::get_result_table() {
                 reinterpret_cast<uintptr_t>(this));
   CurrentScheduler::schedule_and_wait_for_tasks(tasks);
 
+  if (_transaction_context && _transaction_context->aborted()) {
+    _mark_as_failed();
+    return _result_table;
```
mrks: This means `return nullptr`, correct? Rather be explicit. Couple of lines below as well.

src/lib/sql/sql_pipeline_statement.cpp
d-justen: Yes, that's correct. But in order to change this we could not return  `shared_ptr<Table>&` anymore as it cannot be null. Would it be better then to return a `shared_ptr<Table>` instead?

mrks: Got it. Maybe add a comment `// de facto return nullptr, but as a reference`?

d-justen: Or what about `return _result_table = nullptr`?

src/lib/server/server_session.cpp
d-justen: Right now this is pretty difficult for me to test, because (at least until unique constraints are available) there is no way I can think of to provoke such an unsuccessful commit without having access to the actual transaction context.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -195,7 +195,15 @@ const std::shared_ptr<const Table>& SQLPipelineStatement::get_result_table() {
   CurrentScheduler::schedule_and_wait_for_tasks(tasks);
 
   if (_auto_commit) {
-    _transaction_context->commit();
+    const auto success = _transaction_context->commit();
+
+    if (!success) {
+      _query_has_output = false;
+      return _result_table = nullptr;
```
mrks: Rather do only one thing per line. Also, readers who have not seen our discussion will have a hard time understanding why this is done.

src/lib/server/server_session.cpp
mrks: With the rare exception of `AssertInput`, we don't want to use exceptions as part of regular execution. `Fail` should always mean "Something is broken, we don't know why, and we go belly-up".

Can we pass the information to `_handle_client_requests` in a better way?

> Right now this is pretty difficult for me to test, because (at least until unique constraints are available) there is no way I can think of to provoke such an unsuccessful commit without having access to the actual transaction context.

Can't you just insert a hard-coded transaction failure?

src/lib/sql/sql_pipeline_statement.cpp
mrks: Rather do only one thing per line. Also, readers who have not seen our discussion will have a hard time understanding why this is done.

mrks:
Could you update the description please?
Also, does this fix #1393? I don't see a test where we call commit on a context that has a failed operator.

mrks:
The whole handling of failed transactions is still incomplete. The SQLPipeline, for instance, can simply not deal with failed transactions:

```
std::shared_ptr<const Table> SQLPipeline::get_result_table() {
  const auto& tables = get_result_tables();
  Assert(!tables.empty(), "No result tables");
  return tables.back();
}
```

I inserted the following line to make it work for a single query:

```
  if (transaction_context()->phase() == TransactionPhase::RolledBack) return nullptr;
```

Still, if we have multiple queries, it simply does not check for rolled back transactions and tries to execute the next query.

This should be fixed and tested.

src/lib/concurrency/transaction_context.cpp
```diff
@@ -189,16 +176,13 @@ void TransactionContext::_wait_for_active_operators_to_finish() const {
...
+void TransactionContext::_transition(TransactionPhase from_phase, TransactionPhase to_phase,
                                      TransactionPhase end_phase) {
   auto expected = from_phase;
   const auto success = _phase.compare_exchange_strong(expected, to_phase);
 
-  if (success) {
-    return true;
-  } else {
+  if (!success) {
     Assert((expected == to_phase) || (expected == end_phase), "Invalid phase transition detected.");
```
mrks: I don't understand these cases. Why is it ok if the phase is in those states? Also, what is the "end phase" anyway? Can you doc that?

mrzzzrm: I don't think this is part of this PR, that code was there before. Essentially what this is testing is multiple people calling commit() - something we're not sure we want, anyway. Only one of them should perform the transition and everyone else checks whether the the transaction is one of the expected follow-up states.

mrks: > I don't think this is part of this PR

That is rarely a good reason. Either fix it in this PR or open an issue.

> Essentially what this is testing is multiple people calling commit()

I'd make this illegal and see if it causes any problems.

mrks:
Is the end_phase still used somewhere else?

mrzzzrm:
> Is the end_phase still used somewhere else?

Nope, I removed it. All points addressed?

PR #1562, opened 2019-03-27
Replace boost::variant with std::variant in LikeMatcher

FabianWiebe:
This pr replaces `boost::variant` with `std::variant` in `LikeMatcher` class.

It is a preparation for a future JIT pr improving the evaluation of like queries.
JIT cannot handle `boost::variant` code during specialization but it can handle `std::variant` code.

### DIFF
--- a/src/lib/expression/evaluation/like_matcher.hpp
+++ b/src/lib/expression/evaluation/like_matcher.hpp
@@ -2,9 +2,9 @@
 
 #include <regex>
 #include <string>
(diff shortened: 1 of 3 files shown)


src/lib/expression/evaluation/like_matcher.hpp
```diff
@@ -2,9 +2,9 @@
 
 #include <regex>
 #include <string>
+#include <variant>  // NOLINT(build/include_order) Linter assumes variant is a c library
```
mrks: Feel free to add it to cpplint.py

PR #1563, opened 2019-03-27
Use implicit pointer/optional-bool conversion wherever possible

mrzzzrm:
...for consistency.

### DIFF
--- a/src/lib/operators/jit_operator/operators/jit_filter.cpp
+++ b/src/lib/operators/jit_operator/operators/jit_filter.cpp
@@ -14,7 +14,7 @@ std::shared_ptr<const JitExpression> JitFilter::expression() { return _expressio
 
 void JitFilter::_consume(JitRuntimeContext& context) const {
   const auto result = _expression->compute<bool>(context);
-  if ((!_expression->result_entry().is_nullable() || result.has_value()) && result.value()) {
+  if ((!_expression->result_entry().is_nullable() || result) && result.value()) {
     _emit(context);
   }
 }
--- a/src/lib/storage/table.cpp
+++ b/src/lib/storage/table.cpp
@@ -150,7 +150,7 @@ void Table::append_chunk(const Segments& segments, const std::optional<Polymorph
 #if HYRISE_DEBUG
   for (const auto& segment : segments) {
     DebugAssert(segment->size() == chunk_size, "Segments don't have the same length");
-    const auto is_reference_segment = std::dynamic_pointer_cast<ReferenceSegment>(segment) != nullptr;
+    const auto is_reference_segment = std::dynamic_pointer_cast<ReferenceSegment>(segment);
     switch (_type) {
       case TableType::References:
         DebugAssert(is_reference_segment, "Invalid segment type");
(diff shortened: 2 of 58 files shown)


src/lib/operators/jit_operator/operators/jit_filter.cpp
```diff
@@ -14,7 +14,7 @@ std::shared_ptr<const JitExpression> JitFilter::expression() { return _expressio
 
 void JitFilter::_consume(JitRuntimeContext& context) const {
   const auto result = _expression->compute<bool>(context);
-  if ((!_expression->result_entry().is_nullable() || result.has_value()) && result.value()) {
```
lanice: Question here, which is not directly related to your change: if `!_expression->result_entry().is_nullable()` is True, then `&& result.value()` will be called no matter if `result.has_value()` is True or False, correct? And in that case, it throws an exception. Shouldn't this if statement require the check for `result.has_value()` (or rather, just `result` with your changes) in any case?

This comes up in a few other places as well.

src/lib/storage/table.cpp
```diff
@@ -150,7 +150,7 @@ void Table::append_chunk(const Segments& segments, const std::optional<Polymorph
 #if HYRISE_DEBUG
   for (const auto& segment : segments) {
     DebugAssert(segment->size() == chunk_size, "Segments don't have the same length");
-    const auto is_reference_segment = std::dynamic_pointer_cast<ReferenceSegment>(segment) != nullptr;
```
lanice: For something like this, I actually don't like using the automatic conversion. The name `is_reference_segment` clearly implies a boolean, but the actual type of the variable is a pointer. So I would either rename it to just `reference_segment`, or keep the `!= nullptr`. I prefer the latter, because we actually only want a boolean, and don't use the pointer even if it is valid.

src/lib/storage/table.cpp
```diff
@@ -174,7 +174,7 @@ void Table::append_chunk(const Segments& segments, const std::optional<Polymorph
 void Table::append_chunk(const std::shared_ptr<Chunk>& chunk) {
 #if HYRISE_DEBUG
   for (const auto& segment : chunk->segments()) {
-    const auto is_reference_segment = std::dynamic_pointer_cast<ReferenceSegment>(segment) != nullptr;
```
lanice: same as above

src/lib/storage/table.cpp
mrzzzrm: the bane of Search&Replace...

src/lib/storage/table.cpp
mrzzzrm: fixed

src/lib/operators/jit_operator/operators/jit_filter.cpp
mrzzzrm: @FabianWiebe this is JIT, maybe you know if the `if` is doing what it intends to?

Yes, `result.value()` may be called even with `result.has_value()`. Unless there is a dependency that makes sure that `!_expression->result_entry().is_nullable() -> result.has_value()`...

mrzzzrm:
> If this is something we want to have consistency in, should we put it into the contribution guidelines?

Don't really have an opinion on this, for me this is more a case of "try to not type things you don't need to type"... Do you make new people read Contributions.md, @mrks?

lanice:
Looking good to me :+1: The unresolved question is, like mentioned, not actually related to this PR, and maybe @FabianWiebe can say more about it.

src/lib/operators/jit_operator/operators/jit_filter.cpp
FabianWiebe: @lanice If a value is not nullable, the optional is always set so that the` has_value()` check is not necessary.

mrks:
Could you clarify that by removing the `has_value` check and adding a comment? Feel free to do this directly in this PR.

PR #1565, opened 2019-03-29
[DYOD18/19] [constraints/master <--] Refactored Commit Phase

nilsthamm:
This PR introduces our last change. As shown in our talk it is necessary to reorder the commands executed during the commit. In the new state, the checks for uniqueness are executed when the transaction manager fires the callback. This ensures that only one check at a time is executed. The test `ConcurrentInsertInsertRace` shows how a multithreaded scenario would look like, in which the old implementation would have failed.


src/lib/concurrency/transaction_context.cpp
```diff
@@ -182,7 +162,34 @@ void TransactionContext::_mark_as_pending_and_try_commit(std::function<void(Tran
...
+      // If the constraint check fails, set the commit as failed.
+      for (const auto& op : context_ptr->_rw_operators) {
+        const auto& type = op->type();
+        // TOOD(all): This is a dirty hack necessary because, currently, the transaction phase model does not support
+        // the abort of a commit during the "committing" phase. If there is a change to this phase model this code
+        // needs to be refactored
+        if (type == OperatorType::Insert) {
+          auto insert_op = std::dynamic_pointer_cast<Insert>(op);
+          if (!insert_op) {
+            Fail(opossum::trim_source_file_path(__FILE__) + ":" BOOST_PP_STRINGIZE(__LINE__) " " +
```
mrks: DebugAssert?

src/lib/concurrency/transaction_context.cpp
```diff
@@ -182,7 +162,34 @@ void TransactionContext::_mark_as_pending_and_try_commit(std::function<void(Tran
...
+          if (!constraints_satisfied) {
+            context_ptr->_transition(TransactionPhase::Committing, TransactionPhase::Active,
+                                     TransactionPhase::RolledBack);
+            context_ptr->rollback();
+            success = false;
+            break;
+          }
+        }
+      }
+      if (success) context_ptr->_phase = TransactionPhase::Committed;
```
mrks: Add a comment that in the failure case, the transition is made above?

src/lib/concurrency/transaction_context.cpp
```diff
@@ -182,7 +162,34 @@ void TransactionContext::_mark_as_pending_and_try_commit(std::function<void(Tran
   _commit_context->make_pending(_transaction_id, [context_weak_ptr, callback](auto transaction_id) {
     // If the transaction context still exists, set its phase to Committed.
     if (auto context_ptr = context_weak_ptr.lock()) {
-      context_ptr->_phase = TransactionPhase::Committed;
+      auto success = true;
```
mrks: of what?

src/lib/concurrency/transaction_manager.cpp
```diff
@@ -72,10 +72,14 @@ void TransactionManager::_try_increment_last_commit_id(const std::shared_ptr<Com
   while (current_context->is_pending()) {
     auto expected_last_commit_id = current_context->commit_id() - 1;
 
-    if (!_last_commit_id.compare_exchange_strong(expected_last_commit_id, current_context->commit_id())) return;
+    if (!_last_commit_id.compare_exchange_strong(expected_last_commit_id, expected_last_commit_id)) return;
```
mrks: exchange it with... itself?

PR #1568, opened 2019-03-29
[DYOD18/19] Add LZ4 test for (de)compressing with a dictionary

janehmueller:
This PR just adds two test cases (one for strings, one for non-strings) that test segments with enough values, so that a dictionary can be used. This is not the case in any of the tests in EncodedSegmentTest or EncodedStringSegmentTest. 

@Bouncner

### DIFF
--- a/src/lib/storage/lz4_segment.cpp
+++ b/src/lib/storage/lz4_segment.cpp
@@ -94,10 +94,11 @@ std::vector<T> LZ4Segment<T>::decompress() const {
 
   const auto num_blocks = _lz4_blocks.size();
 
+  // This offset is needed to write directly into the decompressed data vector.
+  auto decompression_offset = size_t{0u};
   for (auto block_index = size_t{0u}; block_index < num_blocks; ++block_index) {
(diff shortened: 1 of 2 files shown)


Bouncner:
Is it possible to test the dictionary without creating such large segments?

```
[ RUN      ] StorageLZ4SegmentTest.CompressDictionaryStringSegment
[       OK ] StorageLZ4SegmentTest.CompressDictionaryStringSegment (352 ms)
[ RUN      ] StorageLZ4SegmentTest.CompressDictionaryIntSegment
[       OK ] StorageLZ4SegmentTest.CompressDictionaryIntSegment (130 ms)
```

janehmueller:
In theory yes. I‘ll try to find smaller sizes via trial and error. But the performance will still be bad since we need a lot of data to actually train a dictionary.

janehmueller:
I reduced the sizes a little (halved in both cases). But I also added decompression of the whole segment, since I saw that missing in the coverage. So that might bump the runtimes a little 😅

PR #1575, opened 2019-04-05
Make Assert() and DebugAssert() require semicolon

mrzzzrm:
Before, this may have caused formatting/linting confusion.

The do-while(false) thing is the standard pattern for requiring macros to be concluded with a semicolon just like a normal function call. Pls don't question this. ;)


mrks:
In `strong_typedef.hpp`, we use `static_assert(true, "End call of macro with a semicolon")`. Any reason to prefer the do/while approach?

mrks:
> Pls don't question this.

I don't really care too much which way we use, just wanted to give an alternative. If we stick with do/while, we should use it in `strong_typedef.hpp` too and add a comment. If I saw just the header, I would be very confused about that loop.

mrzzzrm:
2xu02u

Issue #1409, opened 2019-01-15
hyriseBenchmarkJoinOrder --verify fails


mrzzzrm:
Update: SQLite is working since #1419. Now we got different problems, see comments

mrzzzrm:
After #1419, some queries get successfully verified. Some (e.g. 10a) do not finish within reasonable time (left running 3h), some report mismatches (e.g. 7a). Will continue investigating

mrks:
> Interestingly 10a runs almost instantly when issued from the sqlite command prompt.

How did you load the data here?

mrzzzrm:
The JOB has a schema.sql - after running that I probably just loaded the imdb data as normal csv into sqlite

mrks:
That schema.sql uses `PRIMARY KEY`, so you can join on an index.

mrks:
Are JOB queries supposed to return empty results (e.g., 12b)? While sqlite and Hyrise agree, we might want to check if the data is loaded correctly.

mrzzzrm:
Until now, I have kind of lived with some queries returning empty results... Unprofessional as it is. Since the queries are about movies/actors etc, it should be possible to figure out whether there should be matches or not.

mrks:
If loading the data into sqlite / mysql also results in empty results, everything is fine. No need to actually look at the queries.

mrzzzrm:
I think - but this is a year ago! - that running the queries via the sqlite3 cli also produced the occasional empty result set. But we should re-verify this before we forever shrug off these resultless queries as "that's just the way it is"

mrks:
Current results so that you don't have to wait an hour: https://gist.github.com/mrks/3b28c649e328e87078bde4db8cb5ecfd

mrks:
Any chance the original export into CSV files lost something?

```
$ wc -l ~/imdb_data/movie_info_idx.csv
1380035 /home/Markus.Dreseler/imdb_data/movie_info_idx.csv
$ wc -l movie_info_idx.csv
2367725 movie_info_idx.csv
```

However, [following the online guide](https://github.com/gregrahn/join-order-benchmark), I could not get it to finish without problems:

```
# TIME createIndexes() : 7min, 15sec (wall) 0min, 0sec (user) 0min, 0sec (system)
adding foreign keys (this may take a while)
ERROR caught exception creating a foreign key: insert or update on table "aka_title" violates foreign key constraint "movie_id_exists"
[...]

PR #1577, opened 2019-04-05
Add JoinTestRunner to thoroughly test join operators

### LINKED ISSUE #1542: Join Operator tests need improvement
As mentioned in #1541 our join tests failed to catch some relatively simple bugs. Such bugs in our basic operators **need to stop happening**. JoinFullTest, which is responsible for testing non-equi joins tests only 55 JoinMode/Predicate combinations, some of which are "wasted" on testing dictionary inputs, which the join operators are agnostic about. I have two ideas for this for improving this.

## Brute force testing with pre-generated tables
Write a performance-agnostic JNL in python, let it generate tbl files for manifold (all) combinations of predicatecondition/joinmode/{empty-table,small-table,medium-table) for each DataType. Make sure null values are emitting for non-matching tuples in outer joins. The cpp-test code handles flipping input sides, generating reference-segment input and such.

## Cross Validation between Join Operators.
Maybe in the system test, generate A LOT of join operations and verify all our operators (as they support them) produce the same output. This would have caught #1541, e.g., and should be relatively simple to do.

Once this is finished for the Join Operators, consider a similar approach for the TableScan. Once again, imo, we cannot afford having bugs in these fundamental operators. We had two seminar groups working with joins, we should have given them an absolutely stable basis to build on.

### LINKED ISSUE #1557: (Hash?)Join: Clarify comments
Comments talking about the "left" and "right" table are confusing as tables may be flipped. This is not only the case for regular joins, but also for anti joins such as `WHERE a IN (1, 2)`. I would never call `a` the "right side".

Maybe "probed" and "hashed side" are better?

Not sure if this applies to other operators that deal with multiple columns, too.

mrzzzrm:
This is to fix #1542, it runs each of our JoinOperators (NestedLoop, Hash, SortMerge, MPSM, Index-without-indices-for-now) in up to ~1000 configurations (depending on the operations supported by the Operator).

Also, fix #1557

### Bugs hunted down so far:
* JoinNestedLoop had problems with Right/Full Outer Joins and an empty left table
* JoinIndex didn't support Semi/Anti in its fallback-to-jnl mode
* JoinSortMerge couldn't deal with empty input table(s) Part I: `ClusterCount must be > 0`
* JoinSortMerge couldn't deal with empty input table(s) Part II: `Every partition is empty`
* JoinSortMerge sometimes emitted tuples filled up with NULL multiple times.
* JoinHash was buggy for an obscure AntiNullAsTrue cases
* JoinSortMerge/MPSM had UB when columns contained values > UINT_MAX, since `get_radix` would cast every arithmetic value to `uint32_t`


mrks:
> Ship result tables in git repo or generate them on the fly as the tests are running? If we ship the tables that's a lot storage required (1.1 MB) and, more importantly, the generator scripts would not be run in CI - we don't want code in our repo that's not run in CI.

How would these tables be different from the test tables that we already have? It's not that these will change any time soon. [Github considers](https://help.github.com/en/articles/working-with-large-files) files with more than 50 MB as big and with more than 100 MB as too big. I'm slightly leaning towards simply committing them, even if this means that you will save your most-lines-added-spot in the contributor list for years to come. ;)

> Generate input/output tables as well as configurations with python scripts or with c++? The latter would be cleaner to invoke from c++.

This largely depends on the answer to the previous question. If they need to be generated as part of the test, it should be C++. Otherwise, I would think that its easier to write in Python. If there is a python-supported database that supports all join types (i.e., not sqlite), we could even cross-check our generation there.

> If we generate from c++, do we generate tables and configurations as files? I find this important because having actual tbl files to look at is a great thing when debugging failing tests. Having the configurations as a file is great for validating that a case is tested.

Another reason to generate this outside of the tests.

> On my 4Ghz machine the tests take around 3s, significantly more on my laptop. Do we want to burden this on the unit test suite, or do we run the JoinTestRunner as part of the system test?

I can live with either. Maybe we can execute some configurations (e.g., segments other than ValueSegments) only in the system test?

mrks:
If the coverage provided by these tests is good enough, #1409 should be fixed, correct?

mrzzzrm:
Nope, only partially fixed in the best case. The main point of #1409 is "Some (e.g. 10a) do not finish within reasonable time (left running 3h) [in sqlite]"

mrks:
See #1617.

mrzzzrm:
Now - given the info that a PR fixing the other half of #1409 exists - the answer to your question would be: Yes ;)

mrzzzrm:
@mrks ready for review. Better name for `JoinReferenceOperator` is still pending, I haven't come up with anything yet. Paging @Bouncner to notify you that all join operators changed at least a bit.

src/lib/operators/join_reference_operator.cpp
```diff
@@ -0,0 +1,220 @@
+#include "join_reference_operator.hpp"
```
mrzzzrm: (random line) @Bouncner @mrks What do you think of `JoinVerificationOperator` as a name? The term "verify" is used with similar semantic in the `--verify` option of the benchmarks, that's why I think it may be a good fit for something that provides a ground truth.

Bouncner: Sounds good to me. I like the term especially since I don’t think somebody will use this operator accidentally. It’s clearer than reference.

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,530 @@
...
+          }
+        }
+      }
+    }
+
+    // JoinHash specific configurations varying the number of radix_bits.
+    // The number of radix bits affects the partitioning inside the JoinHash operator, thus table/chunk sizes and
+    // different join modes need to be tested in combination, since they all use this partitioning.
+    if constexpr (std::is_same_v<JoinOperator, JoinHash>) {
+      for (const auto radix_bits : {1, 2, 5}) {
```
Bouncner: Is `radix_bits = 0` tested as well somewhere?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -371,10 +371,12 @@ void BenchmarkRunner::_execute_query(const QueryID query_id, const std::shared_p
...
         _query_results[query_id].verification_passed = false;
         std::cout << "- Verification failed: Hyrise returned a result, but SQLite didn't" << std::endl;
-      } else if (!check_table_equal(hyrise_result, sqlite_result, OrderSensitivity::No, TypeCmpMode::Lenient,
-                                    FloatComparisonMode::RelativeDifference)) {
+      } else if (const auto table_difference_message =
+                     check_table_equal(hyrise_result, sqlite_result, OrderSensitivity::No, TypeCmpMode::Lenient,
+                                       FloatComparisonMode::RelativeDifference)) {
         _query_results[query_id].verification_passed = false;
-        std::cout << "- Verification failed (" << timer.lap_formatted() << ")" << std::endl;
+        std::cout << "- Verification failed (" << timer.lap_formatted() << ")" << std::endl
```
Bouncner: Any reason to use `std::endl` in between over `\n`?

src/lib/operators/join_hash.cpp
```diff
@@ -21,54 +21,69 @@
...
+  auto probe_column_id = ColumnID{};
+
+  /**
+   * Build and probe side are assigned as follows (depending only on JoinMode, except for inner joins where input
+   * relation sizes are considered):
+   *
+   * JoinMode::Inner        The smaller relation becomes the build side, the bigger the probe side
+   * JoinMode::Left/Right   The outer relation becomes the probe side, the inner relation becomes the build side
+   * JoinMode::FullOuter    Not supported by JoinHash
+   * JoinMode::Semi/Anti*   The inner relation becomes the build side, the outer relation becomes the probe side
```
Bouncner: Is there an inner and outer relation for semi/anti joins?

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -211,44 +211,45 @@ std::vector<std::optional<HashTable<HashedType>>> build(const RadixContainer<Lef
...
+    if (build_partition_size == 0) {
       continue;
     }
 
-    jobs.emplace_back(std::make_shared<JobTask>([&, partition_left_begin, partition_left_end, current_partition_id,
-                                                 partition_size]() {
-      auto& partition_left = static_cast<Partition<LeftType>&>(*radix_container.elements);
+    jobs.emplace_back(std::make_shared<JobTask>(
+        [&, build_partition_begin, build_partition_end, current_partition_id, build_partition_size]() {
+          auto& build_partition = static_cast<Partition<BuildColumnType>&>(*radix_container.elements);
```
Bouncner: Is that indentation off?

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -533,57 +533,80 @@ void probe_semi_anti(const RadixContainer<RightType>& radix_container,
...
         // Accessors are not thread-safe, so we create one evaluator per job
-        MultiPredicateJoinEvaluator multi_predicate_join_evaluator(left, right, secondary_join_predicates);
+        MultiPredicateJoinEvaluator multi_predicate_join_evaluator(build_table, probe_table, mode,
+                                                                   secondary_join_predicates);
 
         for (size_t partition_offset = partition_begin; partition_offset < partition_end; ++partition_offset) {
-          auto& row = partition[partition_offset];
+          auto& probe_column_element = partition[partition_offset];
 
-          if constexpr (retain_null_values) {
```
Bouncner: A particular reason to remove the template parameter?

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -596,20 +619,20 @@ void probe_semi_anti(const RadixContainer<RightType>& radix_container,
...
+using PosListsByChunk = std::vector<std::shared_ptr<PosLists>>;
 
 /**
  * Returns a vector where each entry with index i references a PosLists object. The PosLists object
  * contains the position list of every segment/chunk in column i.
  * @param input_table
  */
 // See usage in _on_execute() for doc.
-inline PosListsBySegment setup_pos_lists_by_segment(const std::shared_ptr<const Table>& input_table) {
+inline PosListsByChunk setup_pos_lists_by_segment(const std::shared_ptr<const Table>& input_table) {
```
Bouncner: Return type and function name are confusing. Is it now by chunk or segment?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -39,19 +46,13 @@ JoinSortMerge::JoinSortMerge(const std::shared_ptr<const AbstractOperator>& left
                              const OperatorJoinPredicate& primary_predicate,
                              const std::vector<OperatorJoinPredicate>& secondary_predicates)
     : AbstractJoinOperator(OperatorType::JoinSortMerge, left, right, mode, primary_predicate, secondary_predicates) {
-  // Validate the parameters
-  Assert(mode != JoinMode::Cross, "This operator does not support cross joins.");
-  Assert(left, "The left input operator is null.");
-  Assert(right, "The right input operator is null.");
   Assert(primary_predicate.predicate_condition == PredicateCondition::Equals ||
```
Bouncner: Your reason to keep that assert?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -182,7 +189,8 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
     // Note: this is only provisional. There should be a reasonable calculation here based on hardware stats.
     size_t chunk_count_left = _sort_merge_join.input_table_left()->chunk_count();
     size_t chunk_count_right = _sort_merge_join.input_table_right()->chunk_count();
-    return static_cast<size_t>(std::pow(2, std::floor(std::log2(std::max(chunk_count_left, chunk_count_right)))));
```
Bouncner: Reason for that change? Since the chunk counts can never be negative, there shouldn't be values smaller than 2^0. One problem that might occur are empty tables with chunk counts of 0 such that `log2()` fails, but that's still not covered (can that even happen?).

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
```diff
@@ -66,6 +68,26 @@ class RadixClusterSort {
...
+    PerformanceWarning("Using hash to perform bit_cast/radix partitioning of floating point number");
+    return std::hash<T2>{}(value)&radix_bitmask;
+  }
+
+  // Radix calculation for non-arithmetic types
+  template <typename T2>
+  static std::enable_if_t<std::is_same_v<T2, pmr_string>, uint32_t> get_radix(T2 value, size_t radix_bitmask) {
+    PerformanceWarning("Using memcpy to perform bit_cast/radix partitioning of strings");
+    uint32_t radix;
+    std::memcpy(&radix, value.c_str(), std::min(value.size(), sizeof(radix)));
```
Bouncner: Reason to not use std::hash here?

src/lib/operators/print.hpp
```diff
@@ -8,7 +8,8 @@
 #include "abstract_read_only_operator.hpp"
 
 namespace opossum {
-enum PrintFlags { PrintMvcc = 1 << 1 };
+
+enum PrintFlags { PrintMvcc = 1 << 1, PrintIgnoreChunkBoundaries = 1u << 2u };
```
Bouncner: Haven't heard of these chunk boundaries before. It's just about whether we print where a chunk ends and the other begins?

src/lib/storage/table.cpp
```diff
@@ -193,6 +194,51 @@ void Table::append_chunk(const std::shared_ptr<Chunk>& chunk) {
   _chunks.push_back(chunk);
 }
 
+std::vector<AllTypeVariant> Table::get_row(size_t row_idx) const {
+  for (const auto& chunk : _chunks) {
```
Bouncner: Performance warnings for both new methods? They look by far too nice to be not used by students.

src/test/operators/join_test_runner.cpp
mrzzzrm: I probably assumed that was the default, but it is not. I added it here, let's see whether there are bugs ;)

src/benchmarklib/benchmark_runner.cpp
mrzzzrm: Na, just habit. Wouldn't change it here to avoid conflicts with #1597

src/lib/operators/join_hash.cpp
mrzzzrm: Mmh, they are not called like that, right? reverting to left/right then.

src/lib/operators/join_hash/join_hash_steps.hpp
mrzzzrm: Na, it's okay. I wouldn't clang-format-off the entire Job-lambda just to get the lambda header right

src/lib/operators/join_sort_merge.cpp
mrzzzrm: I moved it to AbstractJoinOperator's constructor. We should have some place making sure JoinOps are not called with IN/EXISTS predicates

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
mrzzzrm: strings used memcpy before. But this is actually UB, as the memcpy might access memory it shouldn't for strings <= 3 chars. will use std::hash then.

src/lib/operators/join_sort_merge.cpp
mrzzzrm: I think I changed this precisely because Tables might be empty, i.e., containing no Chunks. Yes, imo (and I *think* @mrks agrees) this is a very valid state and operators need to handle it properly

src/lib/operators/print.hpp
mrzzzrm: precisely. For chunk_size = 1 (which happens in tests) it's hard to decipher table printings without this flag. will doc it.

src/lib/storage/table.cpp
mrzzzrm: yesss

src/lib/operators/join_hash/join_hash_steps.hpp
mrzzzrm: Yep, changed return type name without changing the function name. Fixed

src/lib/operators/join_hash/join_hash_steps.hpp
mrzzzrm: It just wasn't needed anymore.

mrzzzrm: I replaced it with a JoinMode template param now.

src/test/operators/join_test_runner.cpp
mrzzzrm: all seems fine

src/lib/operators/join_sort_merge.cpp
Bouncner: Not what I meant. I see the point for empty chunks, no question.

But what the current change does it to ensure that at least a value of 1 is returned. However, 2^0 is 1, so that's ensured anyhow. Of course assuming we will never see negative chunk counts (should be save to assume). Another problem that is currently not handled (and hasn't been before) is the `log(0)` case, which would fail (both tables have 0 chunks). So what I am proposing is to move the` max(1)` to the chunk count max: `std::log2(std::max({1, chunk_count_left, chunk_count_right})`. For all values >= 1, the following `2^n` will always yield a valid chunk count.

mrzzzrm: gottcha. changed.

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
```diff
@@ -66,6 +68,17 @@ class RadixClusterSort {
 
   virtual ~RadixClusterSort() = default;
 
+  template <typename T2>
+  static std::enable_if_t<std::is_integral_v<T2>, size_t> get_radix(T2 value, size_t radix_bitmask) {
+    return static_cast<int64_t>(value) & radix_bitmask;
+  }
+
+  template <typename T2>
+  static std::enable_if_t<!std::is_integral_v<T2>, size_t> get_radix(T2 value, size_t radix_bitmask) {
```
Bouncner: `is_floating_point`?

Bouncner: And function for `pmr_string` appears to be missing?

src/lib/operators/operator_join_predicate.cpp
```diff
@@ -53,4 +53,8 @@ void OperatorJoinPredicate::flip() {
   predicate_condition = flip_predicate_condition(predicate_condition);
 }
 
+bool operator<(const OperatorJoinPredicate& l, const OperatorJoinPredicate& r) {
+  return std::make_tuple(l.column_ids, l.predicate_condition) < std::make_tuple(r.column_ids, r.predicate_condition);
```
Bouncner: Is the creation of a tuple really necessary here?

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
mrzzzrm: since float and string now both use std::hash, there is no need anymore for dedicated pmr_string/is_floating_point impls. they just share this one

src/lib/operators/operator_join_predicate.cpp
mrzzzrm: less writing. It most probably won't be "created" at runtime

Bouncner: I'd guess it's seldomly used? And just within the optimizer? 

If not, I would rather make sure that we don't create tuple unnecessarily.

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
Bouncner: 🤦

src/lib/operators/operator_join_predicate.cpp
mrzzzrm: I'd bet a good chunk of... let's say a finger of my left hand that no extra machine-code is generated for this in release builds. And even if it were, this is not used in hot loops.

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,527 @@
...
+  return l.to_tuple() < r.to_tuple();
+}
+
+class BaseJoinOperatorFactory;
+
+struct JoinTestConfiguration {
+  InputTableConfiguration input_left;
+  InputTableConfiguration input_right;
+  JoinMode join_mode{JoinMode::Inner};
+  DataType data_type_left{DataType::Int};
```
mrks: These defaults seem somewhat arbitrary. Would it be clearer if those parameters were explicitly set?

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+template <typename JoinOperator>
+class JoinOperatorFactory : public BaseJoinOperatorFactory {
+ public:
+  std::shared_ptr<AbstractJoinOperator> create_operator(const std::shared_ptr<const AbstractOperator>& left,
+                                                        const std::shared_ptr<const AbstractOperator>& right,
+                                                        const JoinMode mode,
+                                                        const OperatorJoinPredicate& primary_predicate,
+                                                        const std::vector<OperatorJoinPredicate>& secondary_predicates,
+                                                        const JoinTestConfiguration& configuration) override {
```
mrks: Aren't all of the above part of the configuration?

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+class JoinTestRunner : public BaseTestWithParam<JoinTestConfiguration> {
+ public:
+  template <typename JoinOperator>
+  static std::vector<JoinTestConfiguration> create_configurations() {
+    auto configurations = std::vector<JoinTestConfiguration>{};
+
+    /**
+     * For each `all_*` set, the first element is the default - which is chosen to be sensible. Thus the elements are
+     * not necessarily sorted.
```
mrks: What do "default" and "sorted mean in this context?

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,527 @@
...
+    JoinTestConfiguration default_configuration{
+      InputTableConfiguration{
+        InputSide::Left, all_chunk_sizes.front(), all_table_sizes.front(), all_input_table_types.front()},
+      InputTableConfiguration{
+        InputSide::Right, all_chunk_sizes.front(), all_table_sizes.front(), all_input_table_types.front()},
+      JoinMode::Inner,
+      DataType::Int,
+      DataType::Int,
+      true,
+      true,
```
mrks: Enable c++20 and bring in designated initializers?

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,527 @@
...
+              join_test_configuration.swap_input_sides();
+            }
+
+            add_configuration_if_supported(join_test_configuration);
+          }
+        }
+      }
+    }
+
+    // JoinHash specific configurations varying the number of radix_bits.
```
mrks: Does it make sense to test pre-sorted / unsorted for JoinSortMerge @Bouncner ?

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,527 @@
...
+      for (const auto& right_input_table_type : all_input_table_types) {
+        auto join_test_configuration = default_configuration;
+        join_test_configuration.input_left.table_type = left_input_table_type;
+        join_test_configuration.input_right.table_type = right_input_table_type;
+
+        add_configuration_if_supported(join_test_configuration);
+      }
+    }
+
+    // Test MPJ support for all join modes. Swap the input tables to trigger cases especially in the Anti* modes
```
mrks: Abbrevation

resources/test_data/sqlite_testrunner_queries.sql
```diff
@@ -136,8 +136,8 @@ SELECT * FROM mixed AS t1 LEFT JOIN mixed_null AS t2 ON t1.a = t2.a AND t1.b <=
 SELECT * FROM mixed AS t1 LEFT JOIN mixed_null AS t2 ON t1.a = t2.a AND t1.b > t2.b;
 SELECT * FROM mixed AS t1 LEFT JOIN mixed_null AS t2 ON t1.a = t2.a AND t1.b >= t2.b;
 SELECT * FROM id_int_int_int_100 AS t1 LEFT JOIN id_int_int_int_100 AS t2 ON t1.a = t2.a;
--- #1527 SELECT * FROM id_int_int_int_100 AS t1 LEFT JOIN id_int_int_int_100 AS t2 ON t1.a > t2.a AND t1.b >= t2.b AND t1.c < t2.c;
--- #1527 SELECT * FROM id_int_int_int_100 AS t1 LEFT JOIN id_int_int_int_100 AS t2 ON t1.a <= t2.a AND t1.b > t2.b AND t1.c < t2.c;
+SELECT * FROM id_int_int_int_100 AS t1 LEFT JOIN id_int_int_int_100 AS t2 ON t1.a > t2.a AND t1.b >= t2.b AND t1.c < t2.c;
```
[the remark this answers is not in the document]
mrzzzrm: Doesn't even have to be bigger input tables, imo. And yes, should be possible. But I'd not do it as part of this PR, because I don't want to fix #1643 as part of this PR and because I'd like to keep my pipeline small. Opened issue #1659

src/lib/operators/join_hash.cpp
```diff
@@ -21,54 +21,76 @@
...
-         "AntiNullAsTrue joins are not supported by JoinHash with secondary predicates.");
-}
+      _radix_bits(radix_bits) {}
 
 const std::string JoinHash::name() const { return "JoinHash"; }
 
+const std::string JoinHash::description(DescriptionMode description_mode) const {
+  std::ostringstream stream;
+  stream << AbstractJoinOperator::description(description_mode);
+  stream << " RadixBits: " << (_radix_bits ? std::to_string(*_radix_bits) : "Unset");
```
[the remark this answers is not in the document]
mrzzzrm: No particular reason. Changed to "Radix bits"

src/lib/operators/join_hash.cpp
```diff
@@ -21,54 +21,76 @@
...
+  const auto inputs_swapped =
       _mode == JoinMode::Left || _mode == JoinMode::AntiNullAsTrue || _mode == JoinMode::AntiNullAsFalse ||
       _mode == JoinMode::Semi ||
       (_mode == JoinMode::Inner && _input_left->get_output()->row_count() > _input_right->get_output()->row_count());
 
   if (inputs_swapped) {
     // We don't have to swap the operation itself here, because we only support the commutative Equi Join.
-    build_operator = _input_right;
-    probe_operator = _input_left;
+    build_input_operator = _input_right;
```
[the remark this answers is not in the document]
mrzzzrm: was there before, but I'll refactor it

src/lib/operators/join_hash.cpp
```diff
@@ -113,21 +136,21 @@ std::shared_ptr<const Table> JoinHash::_on_execute() {
 
 void JoinHash::_on_cleanup() { _impl.reset(); }
 
-template <typename LeftType, typename RightType>
+template <typename BuildColumnType, typename ProbeColumnType>
```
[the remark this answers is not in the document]
mrzzzrm: yes, thoroughly I'd claim

src/lib/operators/join_hash.cpp
```diff
@@ -138,19 +161,20 @@ class JoinHash::JoinHashImpl : public AbstractJoinOperatorImpl {
 
  protected:
   const JoinHash& _join_hash;
-  const std::shared_ptr<const AbstractOperator> _left, _right;
+  const std::shared_ptr<const AbstractOperator> _build_input_operator, _probe_input_operator;
   const JoinMode _mode;
   const ColumnIDPair _column_ids;
   const PredicateCondition _predicate_condition;
-  const bool _inputs_swapped;
+  // If true, the output table will have the probe side columns left of the build side columns.
```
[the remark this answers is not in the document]
mrzzzrm: Changed it to an enum class to correctly image the ternary nature of which input table gets output where:

```c++
// Depending on which input table became the build/probe table we have to order the columns of the output table.
// Semi/Anti* Joins only emit tuples from the probe table
enum class OutputColumnOrder {
  BuildFirstProbeSecond,
  ProbeFirstBuildSecond,
  ProbeOnly
};
```

src/lib/operators/join_hash.cpp
```diff
@@ -258,133 +292,154 @@ class JoinHash::JoinHashImpl : public AbstractJoinOperatorImpl {
...
+        materialized_build_column = materialize_input<BuildColumnType, HashedType, false>(
+            build_table, _column_ids.first, build_chunk_offsets, histograms_build_column, _radix_bits);
+      }
 
       if (_radix_bits > 0) {
-        // radix partition the left table
-        radix_left = partition_radix_parallel<LeftType, HashedType, false>(materialized_left, left_chunk_offsets,
-                                                                           histograms_left, _radix_bits);
+        // radix partition the build table
+        if (keep_nulls_build_column) {
```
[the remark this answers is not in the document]
mrzzzrm: wholeheartedly agree. I want `std::resolve(<enum_or_bool>, Functor functor)`

src/lib/operators/join_hash.cpp
```diff
@@ -200,49 +224,59 @@ class JoinHash::JoinHashImpl : public AbstractJoinOperatorImpl {
...
 
     _output_table = _join_hash._initialize_output_table();
 
-    /*
-     * This flag is used in the materialization and probing phases.
-     * When dealing with an OUTER join, we need to make sure that we keep the NULL values for the outer relation.
-     * In the current implementation, the relation on the right is always the outer relation.
-     * The AntiNullAsFalse-Join, too, will emit tuples with a NULL
+    /**
+     * Discard NULLs from build and probe columns as follows
```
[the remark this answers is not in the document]
mrzzzrm: Changed to "Keep/Discard NULLs from build and probe columns as follows"

src/lib/operators/join_hash.cpp
```diff
@@ -258,133 +292,154 @@ class JoinHash::JoinHashImpl : public AbstractJoinOperatorImpl {
...
     CurrentScheduler::wait_for_tasks(jobs);
 
-    // (Hacky) short cut for AntiNullAsTrue
-    //          If there is any NULL value on the left side, do not bother probe as no tuples can be emitted
-    //          anyway. Doing this early out here is hacky, but during probing we assume NULL values on the left
-    //          side do not matter, so we'd have no chance detecting a NULL value on the left side there.
+    // (Hacky) Short cut for AntiNullAsTrue
+    //          If there is any NULL value on the build side, do not bother probe as no tuples can be emitted
+    //          anyway (as long as JoinHash/AntiNullAsTrue doesn't support secondary predicates). Doing this early out
+    //          here is hacky, but during probing we assume NULL values on the build
```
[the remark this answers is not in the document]
mrzzzrm: Changed as suggested

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -189,19 +189,19 @@ RadixContainer<T> materialize_input(const std::shared_ptr<const Table>& in_table
...
 enum class JoinHashBuildMode { AllPositions, SinglePosition };
 
-template <typename LeftType, typename HashedType, JoinHashBuildMode mode>
-std::vector<std::optional<HashTable<HashedType>>> build(const RadixContainer<LeftType>& radix_container) {
+template <typename BuildColumnType, typename HashedType, JoinHashBuildMode mode>
+std::vector<std::optional<HashTable<HashedType>>> build(const RadixContainer<BuildColumnType>& radix_container) {
   /*
   NUMA notes:
-  The hashtables for each partition P should also reside on the same node as the two vectors leftP and rightP.
+  The hashtables for each partition P should also reside on the same node as the two vectors buildP and probeP.
```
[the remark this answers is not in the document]
mrzzzrm: As far as I can tell they never did, I took these "NUMA notes" comments as "look out for this if you ever make this operator NUMA friendly". @lanice.

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -533,57 +533,80 @@ void probe_semi_anti(const RadixContainer<RightType>& radix_container,
...
+      } else if constexpr (mode == JoinMode::AntiNullAsFalse) {  // NOLINT - doesn't like else if constexpr
+        // no hashtable on other side, but we are in AntiNullAsFalse mode which means all tuples from the probing side
+        // get emitted.
+        pos_list_local.reserve(partition_end - partition_begin);
+        for (size_t partition_offset = partition_begin; partition_offset < partition_end; ++partition_offset) {
+          auto& probe_column_element = partition[partition_offset];
+          pos_list_local.emplace_back(probe_column_element.row_id);
+        }
+      } else if constexpr (mode == JoinMode::AntiNullAsTrue) {  // NOLINT - doesn't like else if constexpr
         // no hashtable on other side, but we are in Anti mode which means all tuples from the probing side get emitted
```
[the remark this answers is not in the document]
mrzzzrm: The subtle difference is that for JoinAntiNullAsTrue: `... all tuples from the probing side get emitted. That is, except NULL values, which only get emitted if the build table is empty.`

src/lib/operators/join_index.cpp
```diff
@@ -56,18 +66,23 @@ void JoinIndex::_perform_join() {
   _right_matches.resize(input_table_right()->chunk_count());
   _left_matches.resize(input_table_left()->chunk_count());
 
-  const auto track_left_matches = (_mode == JoinMode::Left || _mode == JoinMode::FullOuter);
+  const auto is_outer_join = _mode == JoinMode::Left || _mode == JoinMode::Right || _mode == JoinMode::FullOuter;
```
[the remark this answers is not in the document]
mrzzzrm: Only used by JNL/JI and JI is under refactoring/revival anyway. Wouldn't change/refactor for something so minor. One could even claim they are the same anyway.

src/lib/operators/join_index.cpp
```diff
@@ -231,39 +262,6 @@ void JoinIndex::_join_two_segments_using_index(LeftIterator left_it, LeftIterato
   }
 }
 
-// join loop that joins two segments of two columns via their iterators
```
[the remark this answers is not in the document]
mrzzzrm: apparently so.

src/lib/operators/join_mpsm.cpp
```diff
@@ -36,19 +36,30 @@ STRONG_TYPEDEF(uint32_t, ClusterID);
...
-  // Validate the parameters
-  DebugAssert(mode != JoinMode::Cross, "This operator does not support cross joins.");
-  DebugAssert(left, "The left input operator is null.");
-  DebugAssert(right, "The right input operator is null.");
-  DebugAssert(primary_predicate.predicate_condition == PredicateCondition::Equals,
-              "Only Equi joins are supported by MPSM join.");
+                   const OperatorJoinPredicate& primary_predicate,
+                   const std::vector<OperatorJoinPredicate>& secondary_predicates)
+    : AbstractJoinOperator(OperatorType::JoinMPSM, left, right, mode, primary_predicate, secondary_predicates) {
+  // TODO(moritz) incorporate into supports()?
```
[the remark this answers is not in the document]
mrzzzrm: Dear past self, it's creepy how you do that.

src/lib/operators/join_mpsm/column_materializer_numa.hpp
```diff
@@ -16,25 +18,25 @@
 namespace opossum {
 
 template <typename T>
-struct MaterializedValue {
-  MaterializedValue() = default;
-  MaterializedValue(RowID row, T v) : row_id{row}, value{v} {}
+struct MaterializedValueNUMA {
```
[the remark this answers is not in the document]
mrzzzrm: The *NUMA suffix was used before to differentiate JMPSM and JSM, e.g. ColumnMaterializationNUMA. Imo there is little code-cleanliness to gain here from minor naming things, JMPSM/JSM need a down-to-earth refactorwrite.

src/lib/storage/table.cpp
```diff
@@ -193,6 +194,55 @@ void Table::append_chunk(const std::shared_ptr<Chunk>& chunk) {
   _chunks.push_back(chunk);
 }
 
+std::vector<AllTypeVariant> Table::get_row(size_t row_idx) const {
```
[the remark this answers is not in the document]
mrzzzrm: This is supposed to be used in nice-to-read code, so I'd keep the name simple. You'll get a performance warning, anyway.

src/lib/storage/table.cpp
```diff
@@ -193,6 +194,55 @@ void Table::append_chunk(const std::shared_ptr<Chunk>& chunk) {
...
+      return row;
+    } else {
+      row_idx -= chunk->size();
+    }
+  }
+
+  Fail("row_idx out of bounds");
+}
+
+std::vector<std::vector<AllTypeVariant>> Table::get_rows() const {
```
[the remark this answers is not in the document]
mrzzzrm: because I need it elsewhere, too, e.g., the JoinVerification[Operator].

src/test/operators/join_test_runner.cpp
mrzzzrm: Given how red/orange clang still is, I'd say no: https://clang.llvm.org/cxx_status.html

src/lib/utils/check_table_equal.hpp
```diff
@@ -11,8 +11,8 @@ namespace opossum {
 enum class OrderSensitivity { Yes, No };
 
 /**
- * "Strict" enforces that both tables have precisely the same column types, "Lenient" allows float instead of double, double
- * instead of float, long instead of int, int instead of long
+ * "Strict" enforces that both tables have precisely the same column types, "Lenient" allows float instead of double,
+ * double instead of float, long instead of int, int instead of long
```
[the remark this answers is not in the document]
mrzzzrm: Generally, I'd like for PR comments to only refer to things actually changed by that PR, not formatting changes etc. Otherwise people might get discouraged to work on existing unclean code because they fear they might be made responsible for the entire mess in their PR.

Anyway, added "We need this for comparing Hyrise with SQLite 
 * since the column types of the latter might differ from Hyrise's."

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,527 @@
...
+namespace {
+
+using namespace opossum;  // NOLINT
+
+enum class InputSide { Left, Right };
+
+enum class InputTableType {
+  // Input Tables are data
+  Data,
+  // Input Tables are reference Tables with all Segments of a Chunk having the same PosList
```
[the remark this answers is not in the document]
mrzzzrm: Added 
```
// Join operators will build internal PosLists that they have to de-reference when assembling the output Table if the 
// input itself is already a reference Table.
```

src/test/operators/join_test_runner.cpp
mrzzzrm: well, at least join_mode and seconary_predicates are direct members of the configuration. All other ones need to be built from info in the configuration.

src/test/operators/join_test_runner.cpp
mrzzzrm: Unless there is a special "detect_sorted_input" path in the sorting algo, I'd say "no". Firing random configuration combinations into test runner makes its runtime explode pretty quickly.

src/lib/operators/join_hash.cpp
```diff
@@ -21,54 +21,76 @@
 
 namespace opossum {
 
+bool JoinHash::supports(JoinMode join_mode, PredicateCondition predicate_condition, DataType left_data_type,
```
mweisgut: What's the crucial benefit with declaring the `supports`-function as static? Since all `AbstractJoinOperator`-derived classes implement this function, `supports(…)` could be implemented in `AbstractJoinOperator` as (non-static) pure virtual function to guarantee that possible upcoming derived classes have implemented this function.
Imo `supports` is in the same semantic level as `name()` of `AbstractOperator` but for join operators.

mrzzzrm: nothings "crucial" ;)

The reasons I choose static over virtual are:
* With static there is no need to build a join operator that we never use; no need to build a join operator with a configuration that it doesn't support. Not (temporarily) constructing invalid state is what's probably most important to me here.
* Implementation of supports() is sufficiently (implicitly) enforced by places calling `JoinOperator::supports(...)` - so there is no realtistic way you could add a new join operator (which is added to the JoinTestRunner) without implementing supports()
* You'd save passing join_mode, secondary_predicates and predicate_condition with virtual, granted. But you'd not get around to pass left/right_data_type.

src/lib/operators/join_index.cpp
```diff
@@ -56,18 +66,23 @@ void JoinIndex::_perform_join() {
   _right_matches.resize(input_table_right()->chunk_count());
   _left_matches.resize(input_table_left()->chunk_count());
 
-  const auto track_left_matches = (_mode == JoinMode::Left || _mode == JoinMode::FullOuter);
+  const auto is_outer_join = _mode == JoinMode::Left || _mode == JoinMode::Right || _mode == JoinMode::FullOuter;
+  const auto is_semi_or_anti_join =
+      _mode == JoinMode::Semi || _mode == JoinMode::AntiNullAsFalse || _mode == JoinMode::AntiNullAsTrue;
+
+  const auto track_left_matches = is_outer_join || is_semi_or_anti_join;
```
mweisgut: Tracking left matches if `_mode == JoinMode::Right` is not necessary.

src/lib/utils/check_table_equal.hpp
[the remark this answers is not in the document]
mrks: I try not to place an undue burden upon whoever opens the PR when it comes to improving legacy code. At the same time, opening an issue for something this small seems like overkill, too.

I’ll keep your concern in mind and will both add “(or open an issue)” to more of the comments and try to limit the requests that actually take time.

src/test/operators/join_test_runner.cpp
mrks: For me, it’s confusing to pass the same parameter twice. What keeps someone from using the `mode` from `configuration`, which might differ from the one passed in?

src/lib/utils/check_table_equal.hpp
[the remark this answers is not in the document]
mrzzzrm: I'd just ignore minor things.

src/test/operators/join_test_runner.cpp
mrzzzrm: sorry, I wasn't clear. Of course I removed `join mode` and `secondary predicates` from the parameter list and use the one from `configuration`

src/lib/operators/join_index.cpp
mrzzzrm: right!

src/lib/operators/join_reference_operator.cpp
mrzzzrm: `JoinVerification` without "*Operator" suffix it is. Suffix dropped due to @mrks implicit dismissal https://github.com/hyrise/hyrise/pull/1577#discussion_r279195268

src/lib/CMakeLists.txt
```diff
@@ -206,6 +206,8 @@ set(
     operators/join_mpsm/radix_cluster_sort_numa.hpp
     operators/join_nested_loop.cpp
     operators/join_nested_loop.hpp
+    operators/join_verification_operator.cpp
```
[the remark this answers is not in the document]
mrzzzrm: suffix dropped. I thought it made it more clear that... this is an operator. But then, neither do "Sort", "GetTable", etc.

src/test/operators/join_test_runner.cpp
mrzzzrm: When varying only a subset of the dimensions of the configuration, there have to be *some* defaults, we just don't get around that. Explicitly typing all defaults of in each place where a configuration is generated, is not better: It generates noise and makes it look like all those values were somehow chosen deliberately when they were actually not:

Compare this
```c++
    // Test all combinations of reference/data input tables. This tests mostly the composition of the output table
    for (const auto& left_input_table_type : all_input_table_types) {
      for (const auto& right_input_table_type : all_input_table_types) {
        auto join_test_configuration = default_configuration;
        join_test_configuration.input_left.table_type = left_input_table_type;
        join_test_configuration.input_right.table_type = right_input_table_type;

        add_configuration_if_supported(join_test_configuration);
      }
    }
```

to this
```c++
 // Test all combinations of reference/data input tables. This tests mostly the composition of the output table
for (const auto& left_input_table_type : all_input_table_types) {
  for (const auto& right_input_table_type : all_input_table_types) {
    auto join_test_configuration = JoinTestConfiguration;
    join_test_configuration.input_left.input_side = InputSide::Left;
    join_test_configuration.input_left.chunk_size = all_chunk_sizes.front();
    join_test_configuration.input_left.table_size = 10;
    join_test_configuration.input_left.table_type = left_input_table_type;
    join_test_configuration.input_left.input_side = InputSide::Right;
    join_test_configuration.input_left.chunk_size = all_chunk_sizes.front();
    join_test_configuration.input_left.table_size = 10;
    join_test_configuration.input_right.table_type = right_input_table_type;
    join_test_configuration.input_right.join_mode = JoinMode::Inner;
    join_test_configuration.input_right.data_type_left = DataType::Int;
    join_test_configuration.input_right.data_type_right = DataType::Int;
    join_test_configuration.input_right.nullable_left = true;
    join_test_configuration.input_right.nullable_right = true;
    join_test_configuration.predicate_condition = PredicateCondition::Equals;
    join_test_configuration.secondary_predicates = all_secondary_predicate_sets.front();
    join_test_configuration.operator_factory = std::make_shared<JoinOperatorFactory<JoinOperator>>();

    add_configuration_if_supported(join_test_configuration);
  }
}

I want it to be discernable which configuration parameter was chosen deliberately and which was not.
```

src/test/operators/join_test_runner.cpp
mrzzzrm: Update comment to 
```
    /**
     * For each `all_*` set, the first element is used to build the default/base configuration 
     * (`JoinTestConfiguration default_configuration`). Thus the first elements are chosen to be sensible defaults.
     */
```

("sorted" meant just that: "Don't confused if it's not sorted, the elements are ordered such that the first is a good default")

mrzzzrm:
> Any ideas why the coverage decreased?

Deleted some tests that the runner doesn't handle (e.g., for `description()`, `deep_copy()`). Added those again and also some more configurations. Coverage should be up slightly now.

mrzzzrm:
What does everyone think of the name "JoinTestRunner"? I don't like "Runner" - maybe just "Join[Operator]Test"?

mrks:
I‘m fine with `JoinTestRunner`. We should have something that shows that tests are generated, not hand-written. `AutomatedJoinTest`?

src/test/operators/join_test_runner.cpp
mrks: Ok

mrzzzrm:
"Automated" makes me think there is some intelligent logic generating the configurations. I'll think about it a bit and if nothing comes up I'll keep it as it is.

Bouncner:
Will take another look at it this evening. Please wait with the merge.

mrzzzrm:
> Please wait with the merge.

Will do, but if that's the case you can also just "Request Changes" to this PR with a pseudo-message which will automatically block merging.

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,590 @@
...
+        join_test_configuration.join_mode = join_mode;
+        join_test_configuration.nullable_left = true;
+        join_test_configuration.nullable_right = true;
+        join_test_configuration.predicate_condition = predicate_condition;
+
+        add_configuration_if_supported(join_test_configuration);
+      }
+    }
+
+    // The input tables are designed to have exclusive values. Test that these are handled correctly for different
```
Bouncner: What are exclusive values? Why does swapping (potentially) change the tested behavior?

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,590 @@
...
+          join_test_configuration.input_left.table_size = table_size;
+          join_test_configuration.input_right.table_type = right_input_table_type;
+          join_test_configuration.input_right.table_size = table_size;
+
+          add_configuration_if_supported(join_test_configuration);
+        }
+      }
+    }
+
+    // Test multi-predicate jin support for all join modes. Swap the input tables to trigger cases especially in the
```
Bouncner: "jin"

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,590 @@
...
+              join_test_configuration.swap_input_sides();
+            }
+
+            add_configuration_if_supported(join_test_configuration);
+          }
+        }
+      }
+    }
+
+    // Materialization phases might take advantage of certain encodings. JSM, e.g., has a special path for Dictionaries
```
Bouncner: I think I've never seen the abbr. "JSM" before.

Bouncner: finish with a dot.

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,590 @@
...
+                  add_configuration_if_supported(join_test_configuration);
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+
+    return configurations;
```
Bouncner: std::sort(std::unique(configurations))???

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,590 @@
...
+      }
+
+      if (JoinOperator::supports(configuration.join_mode, configuration.predicate_condition,
+                                 configuration.data_type_left, configuration.data_type_right,
+                                 !configuration.secondary_predicates.empty())) {
+        configurations.emplace_back(configuration);
+      }
+    };
+
+    // JoinOperators (e.g. JoinHash) might pick a "common type"
```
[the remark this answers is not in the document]
mrzzzrm: Altered to 
```
    // JoinOperators (e.g. JoinHash) might pick a "common type" from the columns used by the primary predicate and cast
    // all values to that (e.g., joining float + long will cast everything to double)
```

src/test/operators/join_test_runner.cpp
mrzzzrm: Added `// The input tables are designed to have exclusive values (i.e. values not contained in the other tables).`

Think of Semi/Anti joins. An exclusive value on the left side will not appear in the output of a Semi, but in the output of an Anti. Similarily, an exclusive value on the left side will have to be filled up with NULLs by a LeftOuter join, but discarded by a RightOuter join.

src/test/operators/join_test_runner.cpp
mrzzzrm: gin

src/test/operators/join_test_runner.cpp
mrzzzrm: Strange, I use it all the time in my head. Changed to JoinSortMerge

src/test/operators/join_test_runner.cpp
mrzzzrm: You mean the beauty that is this?

```c++
    std::sort(configurations.begin(), configurations.end());
    configurations.erase(std::unique(configurations.begin(), configurations.end()), configurations.end());
```

But yeah, helps:

Before: 1115
After: 1003

Before: 827
After: 459

Before: 740
After: 682

Before: 653
After: 562

Before: 272
After: 233

src/test/operators/join_test_runner.cpp
```diff
@@ -0,0 +1,590 @@
...
+        {{{ColumnID{0}, ColumnID{0}}, PredicateCondition::LessThan}},
+        {{{ColumnID{0}, ColumnID{0}}, PredicateCondition::GreaterThanEquals}},
+        {{{ColumnID{0}, ColumnID{0}}, PredicateCondition::NotEquals}}};
+
+    const auto all_input_table_types =
+        std::vector{InputTableType::Data, InputTableType::IndividualPosLists, InputTableType::SharedPosList};
+
+    // TODO(anybody) include FrameOfReferenceEncoding once #1662 is resolved
+    const auto all_encoding_types =
+        std::vector{EncodingType::Unencoded, EncodingType::RunLength, EncodingType::Dictionary,
```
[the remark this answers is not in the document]
mrzzzrm: I know what you mean, but a "more noise!" approach ("use two encodings here for the default configuration") is snakeoil and shouldn't be necessary. We generate configurations that should trigger all paths and if we don't something is wrong with the configurations that we generate.

src/test/operators/join_test_runner.cpp
Bouncner: Beautiful. :)

src/test/operators/join_test_runner.cpp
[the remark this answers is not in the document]
mrks: Agreed

PR #1578, opened 2019-04-06
Compile time: More type erasure in ColumnVsColumnScan

mrks:
In ColumnVsColumnScans, we previously generated code for comparing `ReferenceSegmentIterable::SingleChunkIterator<SegmentAccessor<int, FrameOfReferenceSegment<int>>>` with `ReferenceSegmentIterable::SingleChunkIterator<SegmentAccessor<int, LZ4Segment<int>>>`. No more.

The PerformanceWarning is never printed, and except for Q15 acting weird again, the TPC-H results remain the same, too.

```
8df96d5075443fde6dc98d41fce711e8cf02b969 (master)
real    10m22,591s
user    86m39,677s
sys     12m16,553s

d8a3a44920f4e61d9109aff50150cce97d050c01 (Type-erase the accessor in type-erased reference segments, erase more aggressively in ColumnVsColumnScan)
real    5m35,670s
user    82m8,795s
sys     12m35,370s
```

### DIFF
--- a/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
@@ -27,9 +27,6 @@ ColumnVsColumnTableScanImpl::ColumnVsColumnTableScanImpl(const std::shared_ptr<c
 
 std::string ColumnVsColumnTableScanImpl::description() const { return "ColumnVsColumn"; }
 
-template<typename... T>
-struct whatis;
-
 std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::scan_chunk(ChunkID chunk_id) const {
   const auto chunk = _in_table->get_chunk(chunk_id);
   const auto left_segment = chunk->get_segment(_left_column_id);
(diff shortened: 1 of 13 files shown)


src/lib/storage/create_iterable_from_segment.hpp
```diff
@@ -28,7 +28,7 @@ class ReferenceSegmentIterable;
  * @{
  */
 
-template <typename T, bool EraseSegmentType = HYRISE_DEBUG>
+template <typename T, bool EraseSegmentType = HYRISE_DEBUG, bool EraseReferencedSegmentType = false /*ignored*/>
```
mrzzzrm: a boolean parameter. Even worse, a boolean parameter for six(!) functions that are not concerned at all with whatever this boolean parameter is concerned with. 
There has to be a way around this, pretty please? :fearful:

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -52,17 +52,35 @@ std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::scan_chunk(ChunkID chunk_i
...
-            chunk_id, create_iterable_from_segment<ColumnDataType>(left_typed_segment),
-            create_iterable_from_segment<ColumnDataType>(*right_typed_segment));
+        if constexpr (std::is_same_v<SegmentType, ReferenceSegment>) {
+          // For reference segments check if the iterables end up using the same type of iterator.
+          auto left_iterable = ReferenceSegmentIterable<ColumnDataType, false>{left_typed_segment};
+          auto right_iterable = ReferenceSegmentIterable<ColumnDataType, false>{*right_typed_segment};
+
+          left_iterable.with_iterators([&](auto left_it, const auto left_end) {
+            right_iterable.with_iterators([&](auto right_it, const auto right_end) {
+              if constexpr (std::is_same_v<std::decay_t<decltype(left_it)>, std::decay_t<decltype(right_it)>>) {
```
mrzzzrm: Just a comment: The lengths that we need to go to to save compile time are getting out of hand. We should drop the type inter-operability, it's just not achievable in a static-polymorphic world...

I mean, look at this, how many people in the radius of 50km around HPI can decipher what we're doing here? :(

src/lib/storage/create_iterable_from_segment.hpp
mrks: Yes, but I have to copy two lines. See cae829dd6

src/test/storage/iterables_test.cpp
```diff
@@ -263,7 +263,24 @@ TEST_F(IterablesTest, ReferenceSegmentIteratorWithIteratorsSingleChunk) {
   auto reference_segment =
       std::make_unique<ReferenceSegment>(table, ColumnID{0u}, std::make_shared<PosList>(std::move(pos_list)));
 
-  auto iterable = ReferenceSegmentIterable<int>{*reference_segment};
+  auto iterable = ReferenceSegmentIterable<int, false>{*reference_segment};
```
mrzzzrm: Coming from an outsider, I really do not know what I am reading here. What is a `ReferenceSegmentIterable<int, false>` why is it not a `ReferenceSegmentIterable<int, EraseReferencedSegmentType::No>`?

mrks:
>  is_reference_segment_iterable 

good idea.

PR #1581, opened 2019-04-08
Update SQLParser

mrzzzrm:
So we can get multi-threaded parsing on the master as well.


mrks:
You'll have to update the handling of LIMIT/OFFSET from ints to expressions.

mrzzzrm:
Dune

PR #1582, opened 2019-04-08
Fix CI build that claimed to be Release, but wasn't actually configur…

mrzzzrm:
…ed as such


mrks:
wat? We need tests for the CI. And a CI that regularly tests those tests.

PR #1584, opened 2019-04-09
Improve benchmark script

Bouncner:
This PR introduces a new column to the result of the benchmark compare script and removes the percentage change aggregation as it is not symmetric (see example below). This PR is more a request for comments as I'd like to get your opinions.

I've talked to a couple of people and we could not come up with a better idea than the proposed "change factors" we discussed a few weeks ago. Change factors are always positive floating point numbers (basically `max(old,new) / min(old,new)`) but are can be a degradation or improvement which is signaled by colour and an up/down arrow.

```
Martin.Boissier@nemea:~/hyrise_master$ python3 ./scripts/compare_benchmarks.py 20180329_master_with_dict.json ../tmt_hyrise_join/20180403_join_reformulation.json
+-----------+--------------------+------+-------------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s       | runs | new iter/s        | runs | change | p-value (significant if <0.001) |
+-----------+--------------------+------+-------------------+------+--------+---------------------------------+
...
| average   |                    |      |                   |      | +162%  |                                 |

Martin.Boissier@nemea:~/hyrise_master$ python3 ./scripts/compare_benchmarks.py ../tmt_hyrise_join/20180403_join_reformulation.json  20180329_master_with_dict.json

+-----------+-------------------+------+--------------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s      | runs | new iter/s         | runs | change | p-value (significant if <0.001) |
+-----------+-------------------+------+--------------------+------+--------+---------------------------------+
...
| average   |                   |      |                    |      | -13%   |                                 |
+-----------+-------------------+------+--------------------+------+--------+---------------------------------+
```

This is the current output:
Screen Shot 2019-04-08 at 11 14 18

PS: there is still an issue with Python2 and encoding. Will look at it as soon as we agree to not throw awasy this PR.


scripts/compare_benchmarks.py
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 
+# -*- coding: utf-8 -*-
```
mrks: > To define a source code encoding, a magic comment must be placed into the source files either as first or second line in the file

scripts/compare_benchmarks.py
```diff
@@ -12,10 +14,16 @@
...
-        return colored('+' + "{0:.0%}".format(diff), 'green')
-    else:
+    if diff < 0:
         return colored("{0:.0%}".format(diff), 'red')
+    else:
+        return colored("+{0:.0%}".format(diff), 'green')
+
+def format_change(change):
+    if change < 0:
+        return colored("▼ {:.4f}".format(abs(change)), 'red')
```
mrks: u"..."

mrks:
Vital feature request:

scripts/compare_benchmarks.py
```diff
@@ -12,10 +14,16 @@
...
+    else:
+        return colored("▲ {:.4f}".format(abs(change)), 'green')
```
mrks: grey for `< 1.05`?

mrks: Can we find symbols that have the same size in the default font?

scripts/compare_benchmarks.py
mrks: On a more serious note, I'm finding it hard to give a narrative for those numbers. Personally, I work with statements like "TPC-H x became 20% faster / twice as fast". `+6%` was clear - it is now 6% faster. What does "1.0587" mean? I end up translating those numbers back into something I can understand, and will get it wrong for red numbers. I wonder if this might be the case for most people.

For single queries, I have no problem understanding that the old "-50%" and "+100%" mean the same thing, just in opposite directions. The problem only comes when those are, quite erroneously, averaged. I wonder if using the factor only for the average might make things easier.

Furthermore, I wonder if we would break anything by subtracting 1. I expect most people to understand red as negative, so having a scale `[-infty, -1], [1, infty]` is confusing.

mrks: Ah, I just understood that your suggestion is to keep both columns. Isn't that more confusing for people who have not seen this discussion? The table should answer the question if the PR improved things or not without a statistics degree.

scripts/compare_benchmarks.py
Bouncner: Rather standard white for "non significant" results? Grey might look strange when everything else is white?

mrks: Sure

mrks: Careful though. I'm talking about irrelevant changes (just doesn't change a lot), not insignificant changes (p-value)

scripts/compare_benchmarks.py
Bouncner: Personally, I understand 17.04 much better than 1804%. And yes, I don't want to replace the percentages, they are especially good to get a quick first impression. I still see a problem for the average reader that -80% look by far less significant than its positive pendent.

Don't get what you mean by substracting 1. Where exactly?

scripts/compare_benchmarks.py
Bouncner: Not required for python3 and not helping for python2. I'll take a look at this issue again when we have agreed on what to do with this PR.

scripts/compare_benchmarks.py
mrks: > Personally, I understand 17.04 much better than 1804%.

Mmmh, but wouldn't it be 1604%?

With big numbers, this is easier, because it is x times faster. For me, it's small red numbers that I can't wrap my head around. Together with the green/red coloring and the arrows, I would expect people to relate this to a percentage change on first sight.

Maybe this is something that could also be taken to StackOverflow? The problem seems to be easy enough to define, and I can't believe that it has no general solution.

scripts/compare_benchmarks.py
mrks: But I have it working with that? Did you also fix the magic comment?

scripts/compare_benchmarks.py
Bouncner: That’s what I meant as well.

scripts/compare_benchmarks.py
```diff
@@ -66,26 +73,37 @@ def calculate_and_format_p_value(old, new):
     new_data = json.load(new_file)
 
 table_data = []
-table_data.append(["Benchmark", "prev. iter/s", "runs", "new iter/s", "runs", "change", "p-value (significant if <" + str(p_value_significance_threshold) + ")"])
+table_data.append(["Benchmark", "prev. iter/s", "runs", "new iter/s", "runs", "change [%]", "change factor", "p-value (significant if <" + str(p_value_significance_threshold) + ")"])
 
 average_diff_sum = 0.0
+changes = []
 
 for old, new in zip(old_data['benchmarks'], new_data['benchmarks']):
```
mrzzzrm: (random line)

I like the general idea of having a "symmetric" metric - and I was originally in favour of the "change factors" proposed here. Now I am wondering: Since the days of improving TPC-H by orders of magnitude are likely over, I wonder if we really want a "main" metric which is likely to always starts with "1.0*" or in good cases with "1.1*" - this would make solid improvements look negligible... But probably this is okay, and we'll just need to get used to/learn to read these numbers.

Unless we come with a better idea now-ish I'd say we merge this PR as a incremental improvement and maybe further discuss this in person and on a whiteboard?

mrks:
There are a couple of problems with the presented approach.

**First, the "average" is inconsistent with commits that leave the overall performance unchanged.** How is "performance did not change" represented? Is it a green 1 or is it 0 (which above was said to not exist)?

Exhibit A:
Screenshot 2019-04-12 at 09 37 53

More generalized speaking, **while the value range of the factors for the queries is limited to [1, infty), that of the average is not**. If the "average" can take numbers that cannot be represented in the range of the numbers it is taken from, it might be everything, but it is not an average. Additionally, the signed factor introduces inconsistencies.

Exhibit B:
Screenshot 2019-04-12 at 09 31 09
In the top example, the performance of one query is doubled, the other one is untouched. The new "average" is 1,5. Next, the second query loses just a tiny bit. This flips the sign on the factor and changes the overall "average".

**Finally, this gives weird results for successive commits.**

Exhibit C:
Screenshot 2019-04-12 at 09 46 47

In the first two tables, the two queries go from 10 iter/s to 60 iter/s via an intermediate step. In the last table, it is direct. I don't see how 2,5 and 2,5 can be mathematicatized to end up as 6.

In summary, this is utterly broken.

mrks:
This is embarrassing. By rephrasing the problem slightly, we will see that this is almost 1:1 a high school maths problem.

*An amount x (initial performance) is invested at a floating rate. Over 22 years (queries), the interest rates (performance improvements) are 4%, 3%, 50%. What is the fixed interest rate that would have yielded the same interest?*

We have been looking for the geometric mean, where the geometric mean of two values is such that both values deviate from the mean by the same factor.

Screenshot 2019-04-12 at 10 57 20

Bouncner:
Let’s discuss this next week or so. Averaging with transposing does not make sense.
And IMO any mean that’s not arithmetic will not be understood, I’d guess.

I am still think we desperately need symmetric metrics in addition because we will sooner or later leave the ground where changes are always good (e.g., join reformulations) and have changes that have both negative and positive effects. Here, discussing with a asymmetrical metric is anything but helpful.

mrks:
Discussed this with Johannes. Mathematically,  this is the way to go. I'll be happy to discuss this next week if you have a different suggestion.  Right now we believe that your requirements are incompatible.

mrks:
Also, geomean is not asymmetric. Instead of addition, you just need machine learning aka. multiplication. 

Just as you wouldn't add the percentage points of two discounts, addition is simply the wrong operation here.

Bouncner:
Just to keep this issue up: I will update the PR this week with the requested color changes. Further, I added the geometric mean as proposed by @mrks as I generally sounds like a good idea.

Concerning the factors, I think we partially talked about different things and the requirement of "transitive" (probably the wrong term) averages was new to me, but definitely sounds like a good idea. Let's talk about it in person after I committed my current changes.

Bouncner:
Talked with @mrks and while I am still not happy with the current situation but I understand that the current proposal have problems than I thought. Other means for symmetric results exist but are (IMO) too hard to understand.

The current status of the PR changes the average result to be a geometric mean and prints changes in white if less than 5%.

Screen Shot 2019-04-18 at 13 58 49

Bouncner:
What's still missing is the "vital feature request" though.

scripts/compare_benchmarks.py
```diff
@@ -12,10 +12,20 @@
 min_runtime_ns = 59 * 1000 * 1000 * 1000
 
 def format_diff(diff):
-    if diff >= 0:
-        return colored('+' + "{0:.0%}".format(diff), 'green')
+    select_color = lambda value, color: color if abs(value) > 0.05 else 'white'
```
mrks: Technically, this is still biased. The white range should be (1/1,05, 1,05). The difference is 0.0023809524, so...

Issue #749, opened 2018-03-01
Get rid of concurrent_vector


mrzzzrm:
Yesterday @mjendruk and I discussed the possibility of abandoning dynamic row counts in Chunks and instead having a fixed number of maximum row, instead of `Chunk::MAX_SIZE`.

In operators the chunk size is known before the Chunk is created, it's just tests that rely on dynamic `append()`. 

Doing so would mean we could get rid of the concurrent vector usage, but would have to use some more memory for the tables in the StorageManager (to allow Insert).

mjendruk:
It also so sloooooooow …

#### Benchmark using `std::vector`
![screen shot 2018-04-05 at 17 59 40](https://user-images.githubusercontent.com/2190123/38377499-5d84fc58-38fb-11e8-9de7-c5dc3160828d.png)

#### Benchmark using `tbb::concurrent_vector`
![screen shot 2018-04-05 at 18 00 01](https://user-images.githubusercontent.com/2190123/38377508-653c9974-38fb-11e8-9bf3-98b7f543b019.png)

![screen shot 2018-04-05 at 18 00 06](https://user-images.githubusercontent.com/2190123/38377550-79a0cb56-38fb-11e8-9e24-022570ae6a31.png)

mrks:
I wrote a lot about this and then I lost it. :(

We should discuss this in the round table. I can see the advantages, but we have to see if they outweigh the additional memory usage. For an ERP with 10,000 tables, 20 columns on average, preallocating 100,000 rows would be ~75 GB.

Besides the performance gain, we can also get rid of the PMR issues that we have with TBB.

mrks:
- We believe that the space overhead is negligible (max 32GB for a database with 10,000 string columns)
- Once we have TPC-C, compare it to `std::vector` and `uninitialized_vector`, then get rid of it.

Bensk1:
Maybe, this could be helpful: http://www.vldb.org/pvldb/vol9/p768-schuhknecht.pdf

mrzzzrm:
@mrks name.bin takes roughly 70s to load into ValueSegments and 18s to load into a DictionarySegment - even though the actual filesize doesn't differ much between the encodings. Getting rid of concurrent_vector could speed up loading binary tables by a factor of >3! (Similar numbers for other imdb tables)

mrks:
We should also compare `std::vector` to `uninitialized_vector` when we work on this.

mrks:
[...]

PR #1585, opened 2019-04-09
Fix Multi-threading support for Insert

### LINKED ISSUE #1086: Multithreaded inserts lead to ChunkID out of range
Example: Connect twice to the server and run multiple inserts simultaneously.

Error: Table throws ChunkID out of range.

Insert.cpp has following code:
```
    // while target chunk is not full
    while (target_start_index != target_chunk->size()) {
      const auto source_chunk = input_table_left()->get_chunk(source_chunk_id);

      ...

      bool source_chunk_depleted = source_chunk_start_index == source_chunk->size();
      if (source_chunk_depleted) {
        source_chunk_id++;
        source_chunk_start_index = 0u;
      }
```

If the target_chunk is expanded during an insertion, the while loop doesn't stop and a chunk from input_table_left is requested that does not exist.

Hotfix: 

 In the while loop there is 
```
still_to_insert -= num_to_insert;
```
which indicates the remaining number of rows that have to be inserted. Break the loop if all rows have been inserted:
```
still_to_insert -= num_to_insert;
if (still_to_insert <= 0) {
   break;
}
```

mrzzzrm:
...and general cleanup. The Insert code is now much more pleasant to read.

Also, fix #1086 

Here is the playground that I used to "test" whether multi-threaded insert works:

```c++
#include <atomic>
#include <iostream>
#include <thread>

#include "operators/print.hpp"
#include "sql/sql_pipeline_builder.hpp"
#include "storage/table.hpp"
#include "types.hpp"

using namespace opossum;               // NOLINT
using namespace std::string_literals;  // NOLINT

int main() {
  const auto table =
      std::make_shared<Table>(TableColumnDefinitions{{"a", DataType::Int}}, TableType::Data, 5, UseMvcc::Yes);
  //table->add_unique_constraint({ColumnID{0}}, false);

  StorageManager::get().add_table("T0", table);

  std::atomic<int32_t> global_variable{0};

  std::vector<std::thread> threads;

  for (size_t t{0}; t < 10; ++t) {
    threads.emplace_back([&global_variable, t]() {
      while (global_variable.load() < 50) {
        const auto query = "INSERT INTO T0 VALUES("s + std::to_string(global_variable.load()) + ")";

        auto statement = SQLPipelineBuilder{query}.dont_cleanup_temporaries().create_pipeline_statement();
        statement.get_result_table();

        if (t == 0) global_variable++;
      }
    });
  }

  for (auto& thread : threads) thread.join();

  Print::print(table);

  return 0;
}
```

### DIFF
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -119,9 +119,18 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
           ChunkRange{target_chunk_id, target_chunk->size(),
                      static_cast<ChunkOffset>(target_chunk->size() + num_rows_for_target_chunk)});
 
+      // Grow MVCC vectors and mark new (but still empty) rows as being under modification by current transaction.
+      // Do so before resizing the Segments, because the resize of `Chunk::_segments.front()` is what releases the
+      // new row count.
+      target_chunk->get_scoped_mvcc_data_lock()->grow_by(num_rows_for_target_chunk, context->transaction_id());
+
       // Grow data Segments.
+      // Do so in REVERSE column order so that the resize of `Chunk::_segments.front()` happens last. It is this last
+      // resize that makes the new row count visible to the outside world.
       auto old_size = target_chunk->size();
-      for (ColumnID column_id{0}; column_id < target_chunk->column_count(); ++column_id) {
+      for (ColumnID reverse_column_id{0}; reverse_column_id < target_chunk->column_count(); ++reverse_column_id) {
+        const auto column_id = static_cast<ColumnID>(target_chunk->column_count() - reverse_column_id - 1);
+
         resolve_data_type(_target_table->column_data_type(column_id), [&](const auto data_type_t) {
           using ColumnDataType = typename decltype(data_type_t)::type;
 
@@ -138,9 +147,6 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
           }
         });
       }
-
-      // Grow MVCC vectors and mark new (but still empty) rows as being under modification by current transaction.
-      target_chunk->get_scoped_mvcc_data_lock()->grow_by(num_rows_for_target_chunk, context->transaction_id());
 
       remaining_rows -= num_rows_for_target_chunk;
     }
@@ -209,12 +215,28 @@ void Insert::_on_rollback_records() {
     const auto target_chunk = _target_table->get_chunk(target_chunk_range.chunk_id);
     auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();
 
+    /**
+     * !!! Crucial comment, PLEASE READ AND _UNDERSTAND_ before altering any of the following code !!!
+     *
+     * Set end_cids to 0 (effectively making the rows invisible for everyone) BEFORE setting the begin_cids to 0.
+     *
+     * Otherwise, another thread might observe a row with `begin_cid == 0` and `end_cid == MAX_COMMIT_ID` which would
+     * make this other thread believe it can see a row that was actually rolled back and should never become visible.
+     *
+     * We need to set `begin_cid = 0` so that the ChunkCompressionTask can identify "completed" Chunks.
+     */
+
     for (auto chunk_offset = target_chunk_range.begin_chunk_offset; chunk_offset < target_chunk_range.end_chunk_offset;
          ++chunk_offset) {
-      // Set the begin and end cids to 0 (effectively making it invisible for everyone).
-      // Make sure that the end is written before the begin.
       mvcc_data->end_cids[chunk_offset] = 0u;
-      std::atomic_thread_fence(std::memory_order_release);
+    }
+
+    // This fence guarantees that no other thread will ever observe `begin_cid = 0 && end_cid != 0` for rolled-back
+    // records
+    std::atomic_thread_fence(std::memory_order_release);
+
+    for (auto chunk_offset = target_chunk_range.begin_chunk_offset; chunk_offset < target_chunk_range.end_chunk_offset;
+         ++chunk_offset) {
       mvcc_data->begin_cids[chunk_offset] = 0u;
       mvcc_data->tids[chunk_offset] = 0u;
     }
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -97,10 +97,6 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
    *    Do so while locking the table to prevent multiple threads modifying the table's size simultaneously.
    *    Since allocation is expected to be faster than writing to the memory, allocating under lock and then writing -
    *    in a second step - without lock will minimize the time that the Table's append_mutex is locked.
-   *
-   *    In the rare case where the input table is the table that we write to (OperatorsInsertTest.SelfInsert), this
-   *    modifies input_table_left(), so calling row_count() on it or size() on its segment is no longer a valid way
-   *    to retrieve the number of rows that are to be inserted.
    */
   {
     const auto append_lock = _target_table->acquire_append_mutex();
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -109,8 +109,7 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
   // Create ColumnTypeWrappers
   auto column_type_wrappers = std::vector<std::unique_ptr<BaseColumnTypeWrapper>>();
   for (const auto& column_type : _target_table->column_data_types()) {
-    column_type_wrappers.emplace_back(
-        make_unique_by_data_type<BaseColumnTypeWrapper, ColumnTypeWrapper>(column_type));
+    column_type_wrappers.emplace_back(make_unique_by_data_type<BaseColumnTypeWrapper, ColumnTypeWrapper>(column_type));
   }
 
   /**
@@ -153,7 +152,7 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
       auto old_size = target_chunk->size();
       for (ColumnID column_id{0}; column_id < target_chunk->column_count(); ++column_id) {
         column_type_wrappers[column_id]->resize_value_segment(target_chunk->get_segment(column_id),
-                                                                  old_size + target_chunk_num_inserted_rows);
+                                                              old_size + target_chunk_num_inserted_rows);
       }
 
       remaining_rows -= target_chunk_num_inserted_rows;
@@ -184,8 +183,8 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
       for (ColumnID column_id{0}; column_id < target_chunk->column_count(); ++column_id) {
         const auto source_segment = source_chunk->get_segment(column_id);
         column_type_wrappers[column_id]->copy(source_segment, source_row_id.chunk_offset,
-                                                  target_chunk->get_segment(column_id), target_chunk_offset,
-                                                  source_chunk_num_rows);
+                                              target_chunk->get_segment(column_id), target_chunk_offset,
+                                              source_chunk_num_rows);
       }
 
       if (source_chunk_num_rows == source_chunk_remaining_rows) {
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -19,21 +19,12 @@ namespace {
 using namespace opossum;  // NOLINT
 
 template <typename T>
-void resize_value_segment(std::shared_ptr<BaseSegment> segment, size_t new_size) {
-  const auto value_segment = std::dynamic_pointer_cast<ValueSegment<T>>(segment);
-  Assert(value_segment, "Cannot insert into non-ValueColumns");
+void copy_value_range(const std::shared_ptr<const BaseSegment>& source_base_segment, ChunkOffset source_begin_offset,
+                      const std::shared_ptr<BaseSegment>& target_base_segment, ChunkOffset target_begin_offset,
+                      ChunkOffset length) {
+  DebugAssert(source_base_segment->size() >= source_begin_offset + length, "Source Segment out-of-bounds");
+  DebugAssert(target_base_segment->size() >= target_begin_offset + length, "Target Segment out-of-bounds");
 
-  value_segment->values().resize(new_size);
-
-  if (value_segment->is_nullable()) {
-    value_segment->null_values().resize(new_size);
-  }
-}
-
-template <typename T>
-void copy(const std::shared_ptr<const BaseSegment>& source_base_segment, ChunkOffset source_begin_offset,
-          const std::shared_ptr<BaseSegment>& target_base_segment, ChunkOffset target_begin_offset,
-          ChunkOffset length) {
   const auto target_value_segment = std::dynamic_pointer_cast<ValueSegment<T>>(target_base_segment);
   Assert(target_value_segment, "Cannot insert into non-ValueSegments");
 
@@ -58,7 +49,7 @@ void copy(const std::shared_ptr<const BaseSegment>& source_base_segment, ChunkOf
       std::copy(nulls_begin_iter, nulls_end_iter, target_value_segment->null_values().begin() + target_begin_offset);
     }
   } else {
-    segment_with_iterators<T>(*source_base_segment, [&](auto source_begin, const auto source_end) {
+    segment_with_iterators<T>(*source_base_segment, [&](const auto source_begin, const auto source_end) {
       auto source_iter = source_begin + source_begin_offset;
       auto target_iter = target_values.begin() + target_begin_offset;
 
@@ -121,15 +112,12 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
         target_chunk = _target_table->get_chunk(target_chunk_id);
       }
 
-      const auto target_chunk_num_inserted_rows =
+      const auto num_rows_for_target_chunk =
           std::min<size_t>(_target_table->max_chunk_size() - target_chunk->size(), remaining_rows);
 
       _target_chunk_ranges.emplace_back(
           ChunkRange{target_chunk_id, target_chunk->size(),
-                     static_cast<ChunkOffset>(target_chunk->size() + target_chunk_num_inserted_rows)});
-
-      // Grow MVCC columns while simultaneously locking them for the current transaction.
-      target_chunk->get_scoped_mvcc_data_lock()->grow_by(target_chunk_num_inserted_rows, context->transaction_id());
+                     static_cast<ChunkOffset>(target_chunk->size() + num_rows_for_target_chunk)});
 
       // Grow data Segments.
       auto old_size = target_chunk->size();
@@ -137,18 +125,29 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
         resolve_data_type(_target_table->column_data_type(column_id), [&](const auto data_type_t) {
           using ColumnDataType = typename decltype(data_type_t)::type;
 
-          resize_value_segment<ColumnDataType>(target_chunk->get_segment(column_id),
-                                               old_size + target_chunk_num_inserted_rows);
+          const auto value_segment =
+              std::dynamic_pointer_cast<ValueSegment<ColumnDataType>>(target_chunk->get_segment(column_id));
+          Assert(value_segment, "Cannot insert into non-ValueColumns");
+
+          const auto new_size = old_size + num_rows_for_target_chunk;
+
+          value_segment->values().resize(new_size);
+
+          if (value_segment->is_nullable()) {
+            value_segment->null_values().resize(new_size);
+          }
         });
       }
 
-      remaining_rows -= target_chunk_num_inserted_rows;
+      // Grow MVCC vectors and mark new (but still empty) rows as being under modification by current transaction.
+      target_chunk->get_scoped_mvcc_data_lock()->grow_by(num_rows_for_target_chunk, context->transaction_id());
+
+      remaining_rows -= num_rows_for_target_chunk;
     }
   }
 
   /**
-   * 2. Insert the Data into the memory allocated in the first step. Write the transaction_context's transaction_id into
-   *    all allocated rows.
+   * 2. Insert the Data into the memory allocated in the first step without holding a lock on the Table.
    */
   auto source_row_id = RowID{ChunkID{0}, ChunkOffset{0}};
 
@@ -171,8 +170,8 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
 
         resolve_data_type(_target_table->column_data_type(column_id), [&](const auto data_type_t) {
           using ColumnDataType = typename decltype(data_type_t)::type;
-          copy<ColumnDataType>(source_segment, source_row_id.chunk_offset, target_segment, target_chunk_offset,
-                               num_rows_current_iteration);
+          copy_value_range<ColumnDataType>(source_segment, source_row_id.chunk_offset, target_segment,
+                                           target_chunk_offset, num_rows_current_iteration);
         });
       }
 
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -18,7 +18,7 @@ namespace {
 
 using namespace opossum;  // NOLINT
 
-class BaseColumnTypeWrapper : public Noncopyable {
+class BaseColumnTypeWrapper {
  public:
   virtual ~BaseColumnTypeWrapper() = default;
 
@@ -45,7 +45,7 @@ class ColumnTypeWrapper : public BaseColumnTypeWrapper {
             const std::shared_ptr<BaseSegment>& target_base_segment, ChunkOffset target_begin_offset,
             ChunkOffset length) override {
     const auto target_value_segment = std::dynamic_pointer_cast<ValueSegment<T>>(target_base_segment);
-    Assert(target_value_segment, "Cannot insert into non-ValueColumns");
+    Assert(target_value_segment, "Cannot insert into non-ValueSegments");
 
     auto& target_values = target_value_segment->values();
     const auto target_is_nullable = target_value_segment->is_nullable();
@@ -145,7 +145,7 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
           ChunkRange{target_chunk_id, target_chunk->size(),
                      static_cast<ChunkOffset>(target_chunk->size() + target_chunk_num_inserted_rows)});
 
-      // Grow MVCC columns.
+      // Grow MVCC columns while simultaneously locking them for the current transaction.
       target_chunk->get_scoped_mvcc_data_lock()->grow_by(target_chunk_num_inserted_rows, context->transaction_id());
 
       // Grow data Segments.
@@ -158,8 +158,6 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
       remaining_rows -= target_chunk_num_inserted_rows;
     }
   }
-
-  // TODO(all): make compress chunk thread-safe; if it gets called here by another thread, things will likely break.
 
   /**
    * 2. Insert the Data into the memory allocated in the first step. Write the transaction_context's transaction_id into
@@ -177,26 +175,26 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
     while (target_chunk_range_remaining_rows > 0) {
       const auto source_chunk = input_table_left()->get_chunk(source_row_id.chunk_id);
       const auto source_chunk_remaining_rows = source_chunk->size() - source_row_id.chunk_offset;
-      const auto source_chunk_num_rows = std::min(source_chunk_remaining_rows, target_chunk_range_remaining_rows);
+      const auto num_rows_current_iteration = std::min(source_chunk_remaining_rows, target_chunk_range_remaining_rows);
 
       // Copy from the source into the target Segments
       for (ColumnID column_id{0}; column_id < target_chunk->column_count(); ++column_id) {
         const auto source_segment = source_chunk->get_segment(column_id);
-        column_type_wrappers[column_id]->copy(source_segment, source_row_id.chunk_offset,
-                                              target_chunk->get_segment(column_id), target_chunk_offset,
-                                              source_chunk_num_rows);
+        const auto target_segment = target_chunk->get_segment(column_id);
+        column_type_wrappers[column_id]->copy(source_segment, source_row_id.chunk_offset, target_segment,
+                                              target_chunk_offset, num_rows_current_iteration);
       }
 
-      if (source_chunk_num_rows == source_chunk_remaining_rows) {
+      if (num_rows_current_iteration == source_chunk_remaining_rows) {
         // Proceed to next source Chunk
         ++source_row_id.chunk_id;
         source_row_id.chunk_offset = 0;
       } else {
-        source_row_id.chunk_offset += source_chunk_num_rows;
+        source_row_id.chunk_offset += num_rows_current_iteration;
       }
 
-      target_chunk_offset += source_chunk_num_rows;
-      target_chunk_range_remaining_rows -= source_chunk_num_rows;
+      target_chunk_offset += num_rows_current_iteration;
+      target_chunk_range_remaining_rows -= num_rows_current_iteration;
     }
   }
 
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -85,6 +85,12 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
   _target_table = StorageManager::get().get_table(_target_table_name);
 
   Assert(_target_table->max_chunk_size() > 0, "Expected max chunk size of target table to be greater than zero");
+  for (ColumnID column_id{0}; column_id < _target_table->column_count(); ++column_id) {
+    // This is not really a strong limitation, we just did not want the compile time of all type combinations.
+    // If you really want this, it should be only a couple of lines to implement.
+    Assert(input_table_left()->column_data_type(column_id) == _target_table->column_data_type(column_id),
+           "Cannot handle inserts into column of different type");
+  }
 
   /**
    * 1. Allocate the required rows in the target Table, without actually copying data to them.
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -18,77 +18,66 @@ namespace {
 
 using namespace opossum;  // NOLINT
 
-class BaseColumnTypeWrapper {
- public:
-  virtual ~BaseColumnTypeWrapper() = default;
+template <typename T>
+void resize_value_segment(std::shared_ptr<BaseSegment> segment, size_t new_size) {
+  const auto value_segment = std::dynamic_pointer_cast<ValueSegment<T>>(segment);
+  Assert(value_segment, "Cannot insert into non-ValueColumns");
 
-  virtual void resize_value_segment(std::shared_ptr<BaseSegment> segment, size_t new_size) = 0;
-  virtual void copy(const std::shared_ptr<const BaseSegment>& source, ChunkOffset source_start_index,
-                    const std::shared_ptr<BaseSegment>& target, ChunkOffset target_start_index, ChunkOffset length) = 0;
-};
+  value_segment->values().resize(new_size);
+
+  if (value_segment->is_nullable()) {
+    value_segment->null_values().resize(new_size);
+  }
+}
 
 template <typename T>
-class ColumnTypeWrapper : public BaseColumnTypeWrapper {
- public:
-  void resize_value_segment(std::shared_ptr<BaseSegment> segment, size_t new_size) override {
-    const auto value_segment = std::dynamic_pointer_cast<ValueSegment<T>>(segment);
-    Assert(value_segment, "Cannot insert into non-ValueColumns");
+void copy(const std::shared_ptr<const BaseSegment>& source_base_segment, ChunkOffset source_begin_offset,
+          const std::shared_ptr<BaseSegment>& target_base_segment, ChunkOffset target_begin_offset,
+          ChunkOffset length) {
+  const auto target_value_segment = std::dynamic_pointer_cast<ValueSegment<T>>(target_base_segment);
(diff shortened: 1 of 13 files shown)


src/lib/operators/insert.cpp
```diff
@@ -8,94 +8,80 @@
...
 template <typename T>
-class TypedSegmentProcessor : public AbstractTypedSegmentProcessor {
- public:
-  void resize_vector(std::shared_ptr<BaseSegment> segment, size_t new_size) override {
-    auto value_segment = std::dynamic_pointer_cast<ValueSegment<T>>(segment);
-    DebugAssert(value_segment, "Cannot insert into non-ValueColumns");
-    auto& values = value_segment->values();
+void resize_value_segment(std::shared_ptr<BaseSegment> segment, size_t new_size) {
+  const auto value_segment = std::dynamic_pointer_cast<ValueSegment<T>>(segment);
+  Assert(value_segment, "Cannot insert into non-ValueColumns");
```
mrks: Haha, I just thought about refactoring an hour ago and how the column-to-segment thing was largely without complications...

src/lib/operators/insert.cpp
```diff
@@ -8,94 +8,80 @@
...
+using namespace opossum;  // NOLINT
 
 template <typename T>
-class TypedSegmentProcessor : public AbstractTypedSegmentProcessor {
- public:
-  void resize_vector(std::shared_ptr<BaseSegment> segment, size_t new_size) override {
-    auto value_segment = std::dynamic_pointer_cast<ValueSegment<T>>(segment);
-    DebugAssert(value_segment, "Cannot insert into non-ValueColumns");
-    auto& values = value_segment->values();
+void resize_value_segment(std::shared_ptr<BaseSegment> segment, size_t new_size) {
```
mrks: Add a comment that it is the callers responsibility to guard the entry to this critical section?

src/lib/operators/insert.cpp
```diff
@@ -8,94 +8,80 @@
...
-    auto casted_target = std::dynamic_pointer_cast<ValueSegment<T>>(target);
-    DebugAssert(casted_target, "Cannot insert into non-ValueColumns");
-    auto& values = casted_target->values();
-
-    auto target_is_nullable = casted_target->is_nullable();
-
-    if (auto casted_source = std::dynamic_pointer_cast<const ValueSegment<T>>(source)) {
-      std::copy_n(casted_source->values().begin() + source_start_index, length, values.begin() + target_start_index);
+template <typename T>
+void copy(const std::shared_ptr<const BaseSegment>& source_base_segment, ChunkOffset source_begin_offset,
```
mrks: copy what? I agree that data doesn't mean too much either. copy attribute? That's what the relational algebra would call it.

src/lib/operators/insert.cpp
```diff
@@ -8,94 +8,80 @@
...
+      const auto nulls_begin_iter = source_value_segment->null_values().begin() + source_begin_offset;
+      const auto nulls_end_iter = nulls_begin_iter + length;
+
+      Assert(
+          target_is_nullable || std::none_of(nulls_begin_iter, nulls_end_iter, [](const auto& null) { return null; }),
+          "Trying to insert NULL into non-NULL segment");
+      std::copy(nulls_begin_iter, nulls_end_iter, target_value_segment->null_values().begin() + target_begin_offset);
+    }
+  } else {
+    segment_with_iterators<T>(*source_base_segment, [&](auto source_begin, const auto source_end) {
```
mrks: Wouldn't `source_it` be more consistent?

src/lib/operators/insert.cpp
```diff
@@ -8,94 +8,80 @@
...
-    auto target_is_nullable = casted_target->is_nullable();
-
-    if (auto casted_source = std::dynamic_pointer_cast<const ValueSegment<T>>(source)) {
-      std::copy_n(casted_source->values().begin() + source_start_index, length, values.begin() + target_start_index);
+template <typename T>
+void copy(const std::shared_ptr<const BaseSegment>& source_base_segment, ChunkOffset source_begin_offset,
+          const std::shared_ptr<BaseSegment>& target_base_segment, ChunkOffset target_begin_offset,
+          ChunkOffset length) {
+  const auto target_value_segment = std::dynamic_pointer_cast<ValueSegment<T>>(target_base_segment);
+  Assert(target_value_segment, "Cannot insert into non-ValueSegments");
```
mrks: DebugAssert that the vector has enough space?

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
 
     if (_target_table->chunk_count() == 0) {
       _target_table->append_mutable_chunk();
     }
 
-    start_chunk_id = _target_table->chunk_count() - 1;
-    end_chunk_id = start_chunk_id + 1;
-    auto last_chunk = _target_table->get_chunk(start_chunk_id);
-    start_index = last_chunk->size();
+    while (remaining_rows > 0) {
```
mrks: An assumption all through the code base is that the last chunk is the one that will be inserted into. Here, we could end up with a full last chunk that might even later be compressed. Not sure about the implications, but maybe we want to avoid this?

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
+        _target_table->append_mutable_chunk();
+        ++target_chunk_id;
+        target_chunk = _target_table->get_chunk(target_chunk_id);
+      }
 
-    auto remaining_rows = total_rows_to_insert;
-    while (remaining_rows > 0) {
-      auto current_chunk = _target_table->get_chunk(static_cast<ChunkID>(_target_table->chunk_count() - 1));
-      auto rows_to_insert_this_loop = std::min(_target_table->max_chunk_size() - current_chunk->size(), remaining_rows);
+      const auto target_chunk_num_inserted_rows =
```
mrks: `rows_to_insert_in_target_chunk` has the same length, but might be clearer

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
+          std::min<size_t>(_target_table->max_chunk_size() - target_chunk->size(), remaining_rows);
 
-      // Resize MVCC vectors.
-      current_chunk->get_scoped_mvcc_data_lock()->grow_by(rows_to_insert_this_loop, MvccData::MAX_COMMIT_ID);
+      _target_chunk_ranges.emplace_back(
```
mrks: `reserve(input_rows / max_chunk_size + 2)`?

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
+          ChunkRange{target_chunk_id, target_chunk->size(),
+                     static_cast<ChunkOffset>(target_chunk->size() + target_chunk_num_inserted_rows)});
```
mrks: Is the end position really necessary?

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
 
-      // Resize current chunk to full size.
-      auto old_size = current_chunk->size();
-      for (ColumnID column_id{0}; column_id < current_chunk->column_count(); ++column_id) {
-        typed_segment_processors[column_id]->resize_vector(current_chunk->get_segment(column_id),
-                                                           old_size + rows_to_insert_this_loop);
-      }
+      // Grow MVCC columns while simultaneously locking them for the current transaction.
+      target_chunk->get_scoped_mvcc_data_lock()->grow_by(target_chunk_num_inserted_rows, context->transaction_id());
```
mrks: I first thought "locking them" referred to `get_scoped_mvcc_data_lock`. Maybe "Grow MVCC vectors and mark new (but still empty) rows as being under modification by current transaction"

mrks: Explain why we grow MVCC vectors before we touch the segments?

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
 
-      remaining_rows -= rows_to_insert_this_loop;
+      // Grow data Segments.
```
mrks: Should `Chunk::size()` be changed so that it returns the size of the *last* segment? Accessing the "new" last row is only valid once the MVCC vectors and all segments are extended.

src/lib/operators/insert.cpp
mrks: Having seen the only call of this method, I wonder if inlining this method could improve readability (and make the previous comment superfluous).

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
+      auto old_size = target_chunk->size();
+      for (ColumnID column_id{0}; column_id < target_chunk->column_count(); ++column_id) {
+        resolve_data_type(_target_table->column_data_type(column_id), [&](const auto data_type_t) {
```
mrks: Use `BaseValueSegment`?

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
     }
   }
-  // TODO(all): make compress chunk thread-safe; if it gets called here by another thread, things will likely break.
 
-  // Then, actually insert the data.
-  auto input_offset = 0u;
-  auto source_chunk_id = ChunkID{0};
-  auto source_chunk_start_index = 0u;
+  /**
+   * 2. Insert the Data into the memory allocated in the first step. Write the transaction_context's transaction_id into
```
mrks: Add a comment that the lock has been released?

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
+   *    all allocated rows.
```
mrks: Didn't you write the txid before?

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
+   */
+  auto source_row_id = RowID{ChunkID{0}, ChunkOffset{0}};
 
-  for (auto target_chunk_id = start_chunk_id; target_chunk_id < end_chunk_id; target_chunk_id++) {
-    auto target_chunk = _target_table->get_chunk(target_chunk_id);
+  for (const auto& target_chunk_range : _target_chunk_ranges) {
```
mrks: This loop feels unintuitive to me. I wouldn't have expected a loop over the chunks that we insert into as the outermost loop. Also, resolving the data type for each value seems to be quite expensive. Isn't it comparable to using AllTypeVariants?

In pseudocode, this seems to be easier:

```
next_target_chunk_range = target_chunk_ranges[0]
next_target_chunk_offset = next_target_chunk_range.begin
foreach column:
  foreach input_chunk:
    iterate_segment<column.type>(input_chunk.get_segment(column), input_it):
      segment = static_cast<ValueSegment<...>>(next_target_chunk_range.chunk.get_segment(column))
      segment.values[next_target_chunk_offset] = *input_it
      next_target_chunk_offset++
      if (next_target_chunk_offset > next_target_chunk_range.end):
        // update next_target_chunk_range
```

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
 void Insert::_on_commit_records(const CommitID cid) {
-  for (auto row_id : _inserted_rows) {
-    auto chunk = _target_table->get_chunk(row_id.chunk_id);
-
-    auto mvcc_data = chunk->get_scoped_mvcc_data_lock();
-    mvcc_data->begin_cids[row_id.chunk_offset] = cid;
-    mvcc_data->tids[row_id.chunk_offset] = 0u;
+  for (const auto& target_chunk_range : _target_chunk_ranges) {
+    const auto target_chunk = _target_table->get_chunk(target_chunk_range.chunk_id);
+    auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();
```
mrks: Do we need to hold the lock for the entire inner loop? (below too)

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
-    // We set the begin and end cids to 0 (effectively making it invisible for everyone) so that the ChunkCompression
-    // does not think that this row is still incomplete. We need to make sure that the end is written before the begin.
-    chunk->get_scoped_mvcc_data_lock()->end_cids[row_id.chunk_offset] = 0u;
-    std::atomic_thread_fence(std::memory_order_release);
-    chunk->get_scoped_mvcc_data_lock()->begin_cids[row_id.chunk_offset] = 0u;
-
-    chunk->get_scoped_mvcc_data_lock()->tids[row_id.chunk_offset] = 0u;
+  for (const auto& target_chunk_range : _target_chunk_ranges) {
+    const auto target_chunk = _target_table->get_chunk(target_chunk_range.chunk_id);
+    auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();
```
mrks: Do we need to hold the lock for the entire inner loop?

src/lib/operators/insert.cpp
```diff
@@ -107,136 +93,132 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
+
+    for (auto chunk_offset = target_chunk_range.begin_chunk_offset; chunk_offset < target_chunk_range.end_chunk_offset;
+         ++chunk_offset) {
+      // Set the begin and end cids to 0 (effectively making it invisible for everyone).
+      // Make sure that the end is written before the begin.
```
mrks: Write end_cids first, issue a single fence, then write the others?

src/lib/storage/chunk.cpp
```diff
@@ -43,7 +43,7 @@ void Chunk::append(const std::vector<AllTypeVariant>& values) {
   DebugAssert(is_mutable(), "Can't append to immutable Chunk");
 
   // Do this first to ensure that the first thing to exist in a row are the MVCC data.
-  if (has_mvcc_data()) get_scoped_mvcc_data_lock()->grow_by(1u, MvccData::MAX_COMMIT_ID);
+  if (has_mvcc_data()) get_scoped_mvcc_data_lock()->grow_by(1u, TransactionID{0});
```
mrks: **is** the MVCC data

src/lib/storage/mvcc_data.cpp
```diff
@@ -6,7 +6,7 @@
 
 namespace opossum {
 
-MvccData::MvccData(const size_t size) { grow_by(size, 0); }
+MvccData::MvccData(const size_t size) { grow_by(size, TransactionID{0}); }
```
mrks: INVALID_TRANSACTION_ID?

src/lib/storage/mvcc_data.hpp
```diff
@@ -36,9 +36,9 @@ struct MvccData {
   /**
    * Grows mvcc data by the given delta
    *
-   * @param begin_cid value all new begin_cids will be set to
+   * @param transaction_id    value all new tids will be set to
    */
-  void grow_by(size_t delta, CommitID begin_cid);
+  void grow_by(size_t delta, TransactionID transaction_id);
```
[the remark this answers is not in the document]
mrzzzrm: I don't think so, the issue was a race of two Insert::_on_execute() running at the same time, both allocating space and then trying to insert **until the table was "full"** instead of stopping when all their input was consumed.

src/lib/operators/insert.cpp
mrzzzrm: ;)

src/lib/operators/insert.cpp
mrzzzrm: yes, inlining is cool.

src/lib/operators/insert.cpp
mrzzzrm: hard one to name right, it is very similar to, e.g., `std::copy`/`memcpy` - just that it operates on segments. `copy_value_range`?

src/lib/operators/insert.cpp
mrzzzrm: Nah, why? `segment_with_iterators` gives us the entire Segment's range, but the function just wants to write to a sub-range of it.

mrks: Aaah, I see what you're doing. Const it?

src/lib/operators/insert.cpp
mrzzzrm: done

src/lib/operators/insert.cpp
mrzzzrm: done

src/lib/operators/insert.cpp
mrzzzrm: better indeed, some of the namings, old and new, are pretty awkward...

src/lib/operators/insert.cpp
mrzzzrm: I'd never expect more than one or two ranges to appear in practice. Each range will cover up to max_chunk_size rows, after all!

src/lib/operators/insert.cpp
mrzzzrm: "Not necessary" as in "there are other ways to solve this". "Necessary" as in "this is how I chose to solve this". ;)

Having the end position prevents us from having to check for "have we consumed all input" and "how many rows can we copy into the current chunk" when we actually copy the data in the `2.` step.

src/lib/operators/insert.cpp
mrzzzrm: done

src/lib/operators/insert.cpp
mrks: I like it

src/lib/operators/insert.cpp
mrks: The discussion about the general structure needs to be resolved first. Generally, I'd like to keep temporary data structures small, especially if we can easily recreate that information later.

src/lib/operators/insert.cpp
mrzzzrm: To do what? If want to grow ValueSegment::values(), we need to know its type.

src/lib/operators/insert.cpp
mrzzzrm: done

src/lib/operators/insert.cpp
mrzzzrm: yep, removed

src/lib/operators/insert.cpp
mrks: Remember that you're forcing the reader to remember all the things you put into that struct. If you recreate that information later, the reader does not need as much heap space and the information will be fresh. ;)

src/lib/operators/insert.cpp
mrzzzrm: Gods, I am not resolving the data type for every value - just for every target chunk range, covering hundred thousands of rows!

mrzzzrm: I am certain there are other ways to structure the loops. The way I do it is that I first build the target-chunk ranges (aka "this is where the inserted data goes") and then I fill these ranges in the seconds step. I think it's okay enough.

src/lib/operators/insert.cpp
mrks: Not if `BaseValueSegment` has a virtual `resize` method. Why should I have to know the type of a segment to resize it?

src/lib/operators/insert.cpp
mrks: Oups. Might have misread that. That makes the performance aspect void, but I'd still like to discuss the general approach from a readability perspective.

src/lib/operators/insert.cpp
mrzzzrm: Don't see a problem, can you elaborate? Ending up with full Chunks is okay, is it not?

src/lib/operators/insert.cpp
mrks: Didn't see the second comment. Let's work through the rest and I'll have another look at this during my second pass.

src/lib/operators/insert.cpp
mrks: Not sure if there is something in the code that would actually break right now. I am just wondering if we should always have a modifiable chunk at the end as this is how we communicate it, e.g., in our chunk visualization. Saying "the last chunk is where we insert" seems clearer that "the last chunk might be an encoded segment or a ValueSegment, and if its size is less than the maximum_chunk_size, this is where we insert".

mrks: @Bouncner was involved in a couple of chunk layout things, too.

src/lib/operators/insert.cpp
mrzzzrm: tbh I don't even know why we need that lock. The MvccData uses concurrent_vectors, after all, and Insert holds a lock over the entire table when it resizes them...

src/lib/operators/insert.cpp
mrzzzrm: Having the least amount of runtime-constraints like "you can rely on the last chunk being non-full and modifiable" seems desirable for me. Remember how well this "a table will always have at least one chunk"-thing went?

src/lib/operators/insert.cpp
mrzzzrm: The readers heap/stack is one of my core concerns, yes. But their compute power is as well. (Re)constructing **where data went/goes** needs to happen in four places: When allocating, when copying, when committing and when rolling back. I'd rather require my reader to have a model of a vector of ranges within a Chunks then having them to parse redundant computations in four different places.

src/lib/operators/insert.cpp
mrzzzrm: No hard opinion on this, but there is only one place where we grow/resize segments: `Insert`. I'd rather not add a virtual function which is then only used in one place.I'd inline it.

src/lib/operators/insert.cpp
mrzzzrm: Why the fencing? From around 5h of watching c++ memory model talks and very little security gained on the topic I'd say `mvcc_data->tids[chunk_offset] = 0u;` at the very end is the release operation that makes the inserted row visible. No fences needed....

src/lib/operators/insert.cpp
mrzzzrm: The table is locked, no one else can resize its segments or mvcc-data.
Every query will see the data through the lenses of mvcc, it does not matter whether the mvcc data is not grown or whether it is grown but locked.

If anything I'd say we should resize the MVCC vectors AFTER the value segments, as the resize of the tid vector will be a release operations and "those should always happen at the end", informally speaking.

mrzzzrm: @mrks I think we should take 5 minutes tomorrow to synchronize our understanding of the c++ memory model.

mrks: Mmh, I thought this was done on purpose; so this was a request to add a comment.

>  Every query will see the data through the lenses of mvcc, it does not matter whether the mvcc data is not grown or whether it is grown but locked.

But how do the "lenses of MVCC" work? The validate operator looks at all rows (i.e., `chunks->size()`). If we resize the segments first, it might end up accessing non-existing MVCC info.

> I think we should take 5 minutes tomorrow to synchronize our understanding of the c++ memory model.

The C++ memory model in five minutes. I giggled.

src/lib/operators/insert.cpp
mrks: You added the fencing. My suggestion is to remove it from the loop for performance reasons.

src/lib/operators/insert.cpp
mrks: Works for me.

src/lib/operators/insert.cpp
mrks: Works for me, still wanted to bring it up.

src/lib/operators/insert.cpp
mrzzzrm: the fence was there before, I doubt it was I who added it originally.

src/lib/operators/insert.cpp
mrzzzrm: > Mmh, I thought this was done on purpose

c/p from the previous impl


> But how do the "lenses of MVCC" work? The validate operator looks at all rows (i.e., chunks->size()). If we resize the segments first, it might end up accessing non-existing MVCC info.

Then Validate *should* probably look at the mvcc-data-length... Really looking forward to tpcc with 100 clients to stress test all this...

src/lib/operators/insert.cpp
mrks: I believe it has something to do with the MVCC data being migrated if a chunk is being encoded. Will have to look into it again, though.

src/lib/operators/insert.cpp
mrks: Ah, ok. As this PR rewrote basically everything, I used split view and didn't check the left side too much.

src/lib/operators/insert.cpp
mrks: I think `Chunk::size` should return how many valid rows there are and no one should be allowed to touch any offsets in any vectors beyond that.

> Then Validate should probably look at the mvcc-data-length...

Which one of the three vectors?

mrzzzrm: > Which one of the three vectors?

the tid vector, I guess. it should always be written to as the last operation and read from as the first operation - it contains atomics so this should get us closer to a safe acquire/release pattern

mrzzzrm: (I am being vague here as really want to discuss this in the presence of a whiteboard...)

src/lib/operators/insert.cpp
mrzzzrm: This is roughly how on_commit_records will look like with and without explicitly storing the ChunkRanges. This increased complexity would be there for commit_records/rollback_records and the copy step. I think it is desirable to store the ChunkRanges, given the increased complexity incurred without them

```c++

void Insert::_on_commit_records(const CommitID cid) {
  // Without explicitly storing the target ChunkRanges
  for (auto target_chunk_id = _target_table_begin.chunk_id; target_chunk_id < _target_table_end.chunk_id;
       ++target_chunk_id) {
    const auto target_chunk = _target_table->get_chunk(target_chunk_id);

    const auto target_chunk_begin_offset =
        target_chunk_id == _target_table_begin.chunk_id ? _target_table_begin.chunk_offset : ChunkOffset{0};
    const auto target_chunk_end_offset =
        target_chunk_id == _target_table_end.chunk_id ? _target_table_end.chunk_offset : target_chunk->size();

    auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();

    for (auto chunk_offset = target_chunk_begin_offset; chunk_offset < target_chunk_end_offset; ++chunk_offset) {
      mvcc_data->begin_cids[chunk_offset] = cid;
      mvcc_data->tids[chunk_offset] = 0u;
    }
  }

  // With explicitly storing the target ChunkRanges
  for (const auto& target_chunk_range : _target_chunk_ranges) {
    const auto target_chunk = _target_table->get_chunk(target_chunk_range.chunk_id);
    auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();

         ++chunk_offset) {
      mvcc_data->begin_cids[chunk_offset] = cid;
      mvcc_data->tids[chunk_offset] = 0u;
    }
  }
}
```

src/lib/operators/insert.cpp
mrks: If we wrote the begin CID before we write the end CID, we might end up in a situation where the row (with begin 0, end infty) is indisinguishable from a row that is visible to everyone. This should be added to the comment.

src/lib/operators/insert.cpp
```diff
@@ -107,136 +84,140 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
-    chunk->get_scoped_mvcc_data_lock()->begin_cids[row_id.chunk_offset] = 0u;
-
-    chunk->get_scoped_mvcc_data_lock()->tids[row_id.chunk_offset] = 0u;
+  for (const auto& target_chunk_range : _target_chunk_ranges) {
+    const auto target_chunk = _target_table->get_chunk(target_chunk_range.chunk_id);
+    auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();
+
+    for (auto chunk_offset = target_chunk_range.begin_chunk_offset; chunk_offset < target_chunk_range.end_chunk_offset;
+         ++chunk_offset) {
+      // Set the begin and end cids to 0 (effectively making it invisible for everyone).
```
mrks: Bring back the reference to the ChunkCompressionTASK

src/lib/storage/chunk_encoder.hpp
```diff
@@ -32,6 +32,9 @@ using ChunkEncodingSpec = std::vector<SegmentEncodingSpec>;
 /**
  * @brief Interface for encoding chunks
  *
+ * TODO(all): make compress chunk thread-safe; if it gets called while another thread is inserting, things will likely
+ *            break.
```
mrks: Update comment: This thingy does not guarantee thread-safety, the ChunkEncoderTask does

src/lib/operators/insert.cpp
mrks: It is needed when we shrink the MVCC data:

```
  /**
   * Compacts the internal representation of
   * the mvcc data in order to reduce fragmentation
   * Locks mvcc data exclusively in order to do so
   */
  void shrink();
```

However, 9b15cafcc98e8f82ae731e7601a2fa701ee047bb removed the exclusive lock by mistake. It should be reintroduced (and then removed again by #749).

mrzzzrm: done

src/lib/operators/insert.cpp
mrzzzrm: Added this comment, what do you think?

```c++
    /**
     * !!! Crucial comment, PLEASE READ AND _UNDERSTAND_ before altering any of the following code !!!
     *
     * Set end_cids to 0 (effectively making the rows invisible for everyone) BEFORE setting the begin_cids to 0.
     *
     * Otherwise, another thread might observe a column with `begin_cid == 0` and `end_cid == MAX_COMMIT_ID` which would
     * make this other thread believe it can see a row that was actually rolled back and should never become visible.
     *
     * We need to set `begin_cid = 0` so that the ChunkCompressionTask can identify "completed" Chunks.
     */
```

src/lib/operators/insert.cpp
mrzzzrm: done

src/lib/storage/chunk_encoder.hpp
mrzzzrm: Altered to `NOT thread-safe. In a multi-threaded context, the ChunkCompressionTask should invoke the ChunkEncoder.`

src/lib/operators/insert.cpp
mrzzzrm: Apparently we do, otherwise someone might call "shrink" on the mvcc data while we access it. Acquiring the lock for each row individually seems bad as well.

src/lib/operators/insert.cpp
mrzzzrm: resolved?

src/lib/operators/insert.cpp
mrzzzrm: Regarding 

> Explain why we grow MVCC vectors before we touch the segments? 

I added comments:

```c++
      // Grow MVCC vectors and mark new (but still empty) rows as being under modification by current transaction.
      // Do so before resizing the Segments, because the resize of `Chunk::_segments.front()` is what releases the
      // new row count.
      target_chunk->get_scoped_mvcc_data_lock()->grow_by(num_rows_for_target_chunk, context->transaction_id());

      // Grow data Segments.
      // Do so in REVERSE column order so that the resize of `Chunk::_segments.front()` happens last. It is this last
      // resize that makes the new row count visible to the outside world.
      auto old_size = target_chunk->size();
      for (ColumnID reverse_column_id{0}; reverse_column_id < target_chunk->column_count(); ++reverse_column_id) {
```

src/lib/operators/insert.cpp
mrzzzrm: done

src/lib/operators/insert.cpp
```diff
@@ -119,9 +119,18 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
                      static_cast<ChunkOffset>(target_chunk->size() + num_rows_for_target_chunk)});
 
+      // Grow MVCC vectors and mark new (but still empty) rows as being under modification by current transaction.
+      // Do so before resizing the Segments, because the resize of `Chunk::_segments.front()` is what releases the
+      // new row count.
+      target_chunk->get_scoped_mvcc_data_lock()->grow_by(num_rows_for_target_chunk, context->transaction_id());
+
       // Grow data Segments.
+      // Do so in REVERSE column order so that the resize of `Chunk::_segments.front()` happens last. It is this last
+      // resize that makes the new row count visible to the outside world.
```
mrks: Don't we need a fence here? Right now, there is nothing that keeps the compiler from resizing the segments in any order.

src/lib/operators/insert.cpp
```diff
@@ -209,12 +215,28 @@ void Insert::_on_rollback_records() {
     const auto target_chunk = _target_table->get_chunk(target_chunk_range.chunk_id);
     auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();
 
+    /**
+     * !!! Crucial comment, PLEASE READ AND _UNDERSTAND_ before altering any of the following code !!!
+     *
+     * Set end_cids to 0 (effectively making the rows invisible for everyone) BEFORE setting the begin_cids to 0.
+     *
+     * Otherwise, another thread might observe a row with `begin_cid == 0` and `end_cid == MAX_COMMIT_ID` which would
+     * make this other thread believe it can see a row that was actually rolled back and should never become visible.
```
mrks: Mmh, this is not that easy to understand. Maybe:
Otherwise, another transaction/thread might observe a row with `begin_cid == 0`, `end_cid == MAX_COMMIT_ID` and a foreign tid which is what a visible row that is being deleted by a different transaction looks like. Thus, the other transaction would consider the row that is in the process of being rolled back and should have never been visible as visible.

src/lib/storage/mvcc_data.cpp
```diff
@@ -12,6 +12,15 @@ MvccData::MvccData(const size_t size) { grow_by(size, INVALID_TRANSACTION_ID); }
 size_t MvccData::size() const { return _size; }
 
 void MvccData::shrink() {
+  // tbb::concurrent_vector::shrink_to_fit() is not thread-safe, we need a unique lock to it.
+  //
+  // https://software.intel.com/en-us/node/506205
+  //   "Concurrent invocation of these operations on the same instance is not safe."
+  // https://software.intel.com/en-us/node/506203
+  //   "The method shrink_to_fit()merges several smaller arrays into a single contiguous array, which may improve access
```
mrks: typo

src/lib/tasks/chunk_compression_task.cpp
```diff
@@ -42,6 +42,9 @@ bool ChunkCompressionTask::_chunk_is_completed(const std::shared_ptr<Chunk>& chu
   auto mvcc_data = chunk->get_scoped_mvcc_data_lock();
 
   for (const auto begin_cid : mvcc_data->begin_cids) {
+    // TODO(anybody) Reading the non-atomic begin_cid (which is written to in Insert without a write lock) is likely UB
+    //               When activating the ChunkCompressionTask, please look for a different means of determining whether
+    //               All Inserts to a Chunk finished.
```
mrks: all

src/lib/operators/insert.cpp
mrzzzrm: I probably thought of tbb::concurrent_vector::resize() as a full-seq-consistency operation that doesn't need fences....

Added this after the column resize, but I have no idea whether it is necessary or not or whether a release-barrier would be enough...:

```c++

        // Make sure the first columns rewrite actually happens last and doesn't get reordered.
        std::atomic_thread_fence(std::memory_order_seq_cst);
```

src/lib/operators/insert.cpp
mrzzzrm: mmh, okay, changed.

src/lib/operators/insert.cpp
mrzzzrm: I feel like it has to be a seq_cst barrier to make sure the first column's new size only gets observed after the other resizes completed, but it feels awkward...

mrks: Yes, the concurrent_vector should be fine for now, but once we put in std::vectors, I can almost guarantee that this will be forgotten.

src/lib/operators/insert.cpp
mrks: I tried quite a while to get my version to work. Turns out your approach is better suited to handle self-inserts.

mrks:
Inserts into a table with four ints on nemea:

```
./hyriseBenchmarkFileBased --table_path=../foo --query_path=../foo --mvcc --scheduler --clients=100 --runs=100000
Executed 100000 times in 1.89297 seconds (52827.2 iter/s)
```

Not bad given that we haven't really optimized the OLTP part too much yet.

mrzzzrm:
How many iter/s with a single client?

mrks:
100 clients, all cores: 52827.2 iter/s
single client, all cores: 99.1388 iter/s
single client, single-threaded: 71554.3 iter/s

src/lib/operators/insert.cpp
```diff
@@ -97,6 +97,10 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
    *    Do so while locking the table to prevent multiple threads modifying the table's size simultaneously.
    *    Since allocation is expected to be faster than writing to the memory, allocating under lock and then writing -
    *    in a second step - without lock will minimize the time that the Table's append_mutex is locked.
+   *
+   *    In the rare case where the input table is the table that we write to (OperatorsInsertTest.SelfInsert), this
```
mrzzzrm: I am not sure such vague comments pointing in the direction of possible complications if a problem would have been attempted to be solved in a different manner are very helpful...

mrks:
Added a run with a single NUMA node

mrks:
I have no idea what happened here, but this reduced the TPC-H Q2 performance from 8.51236 to 7.18498. I don't see a single line in here that is used for Q2, so this might be another alignment issue. :/

PR #1595, opened 2019-04-12
Unified handling of operator<<(std::ostream, ...), to_string() and print()

### LINKED ISSUE #716: print vs to_string vs operator<<
I feel like we should decide how we want to get a string representation of our classes. Currently, we do all types of things:

```
void AbstractLQPNode::print(std::ostream& out);
std::string to_string(const AllParameterVariant& x);
std::ostream& operator<<(std::ostream& os, const LQPColumnReference& column_reference);
```

I like `to_string` because it is easy to use in tests: `EXPECT_EQ(column_expression1->to_string(), "ColumnID #1 * 2");`
`operator<<` is nice, too, because that's what everyone is used to.
For `print`, I can't come up with something particularly good about it.

mrzzzrm:
Fix #716

Every relevant enum/class/struct is now print-able to streams like so: `std::cout << data_type << lqp_node;`. Previously this was different for each data type: `std::cout << data_type_to_string.left.at(data_type); lqp_node.print(std::cout);`

### DIFF
--- a/src/benchmarklib/benchmark_runner.cpp
+++ b/src/benchmarklib/benchmark_runner.cpp
@@ -40,7 +40,7 @@ BenchmarkRunner::BenchmarkRunner(const BenchmarkConfig& config, std::unique_ptr<
 
     Topology::use_default_topology(config.cores);
     std::cout << "- Multi-threaded Topology:" << std::endl;
-    Topology::get().print(std::cout, 2);
+    std::cout << Topology::get();
 
     // Add NUMA topology information to the context, for processing in the benchmark_multithreaded.py script
     auto numa_cores_per_node = std::vector<size_t>();
--- a/src/benchmarklib/benchmark_runner.cpp
+++ b/src/benchmarklib/benchmark_runner.cpp
@@ -271,7 +271,7 @@ void BenchmarkRunner::_benchmark_individual_queries() {
     const auto duration_seconds = static_cast<float>(result.duration_ns) / 1'000'000'000;
     const auto items_per_second = static_cast<float>(result.num_iterations) / duration_seconds;
 
-    std::cout << "  -> Executed " << result.num_iterations << " times in " << duration_seconds << " seconds ("
+    std::cout << "  -> Executed " << result.num_iterations.load() << " times in " << duration_seconds << " seconds ("
               << items_per_second << " iter/s)" << std::endl;
 
     // Wait for the rest of the tasks that didn't make it in time - they will not count toward the results
--- a/src/lib/constant_mappings.cpp
+++ b/src/lib/constant_mappings.cpp
@@ -53,28 +53,23 @@ const boost::bimap<VectorCompressionType, std::string> vector_compression_type_t
     });
 
 std::ostream& operator<<(std::ostream& stream, AggregateFunction aggregate_function) {
-  stream << aggregate_function_to_string.left.at(aggregate_function);
-  return stream;
+  return stream << aggregate_function_to_string.left.at(aggregate_function);
 }
 
 std::ostream& operator<<(std::ostream& stream, FunctionType function_type) {
-  stream << function_type_to_string.left.at(function_type);
-  return stream;
+  return stream << function_type_to_string.left.at(function_type);
 }
 
 std::ostream& operator<<(std::ostream& stream, DataType data_type) {
-  stream << data_type_to_string.left.at(data_type);
-  return stream;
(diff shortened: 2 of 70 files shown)


src/lib/types.hpp
```diff
@@ -13,6 +8,13 @@
 #include <tuple>
 #include <vector>
 
+#include "tbb/concurrent_vector.h"
```
mrks: ?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -271,7 +271,7 @@ void BenchmarkRunner::_benchmark_individual_queries() {
     const auto duration_seconds = static_cast<float>(result.duration_ns) / 1'000'000'000;
     const auto items_per_second = static_cast<float>(result.num_iterations) / duration_seconds;
 
-    std::cout << "  -> Executed " << result.num_iterations << " times in " << duration_seconds << " seconds ("
+    std::cout << "  -> Executed " << result.num_iterations.load() << " times in " << duration_seconds << " seconds ("
```
mrks: Why did this become necessary?

src/lib/constant_mappings.cpp
```diff
@@ -97,7 +52,29 @@ const boost::bimap<VectorCompressionType, std::string> vector_compression_type_t
         {VectorCompressionType::SimdBp128, "SIMD-BP128"},
     });
 
-const boost::bimap<TableType, std::string> table_type_to_string =
-    make_bimap<TableType, std::string>({{TableType::Data, "Data"}, {TableType::References, "References"}});
+std::ostream& operator<<(std::ostream& stream, AggregateFunction aggregate_function) {
+  stream << aggregate_function_to_string.left.at(aggregate_function);
+  return stream;
```
mrks: Shouldn't `return stream << ` work, too?

src/lib/expression/between_expression.cpp
```diff
@@ -29,8 +29,7 @@ std::shared_ptr<AbstractExpression> BetweenExpression::deep_copy() const {
 
 std::string BetweenExpression::as_column_name() const {
   std::stringstream stream;
-  stream << _enclose_argument_as_column_name(*value()) << " "
-         << predicate_condition_to_string.left.at(predicate_condition) << " "
+  stream << _enclose_argument_as_column_name(*value()) << " " << predicate_condition << " "
```
mrks: Can we also `std::cout << *abstract_expression` now?

mrks:
I like this a lot :)

src/lib/types.hpp
mrzzzrm: File uses `tbb::concurrent_vector` - apparently it was transitively included before....

src/benchmarklib/benchmark_runner.cpp
mrzzzrm: There is no `operator<<(std::ostream&, std::atomic<int>)`, so this previously used `operator<<(..., int)` via [implicit conversion](https://en.cppreference.com/w/cpp/atomic/atomic/operator_T). With the introduction of operator<< overloads for STRONG_TYPEDEFs, there are multiple candidate `operator<<` now - those for the primitive data types and those for the strong typedefs. Calling load() removes the ambiguity.

src/lib/expression/between_expression.cpp
mrzzzrm: yes, but in this case `_enclose_argument_as_column_name` is necessary to ensure parentheses are printed to make the printed expression non-ambiguous

src/lib/constant_mappings.cpp
mrzzzrm: ehrm yes. good point!

mrzzzrm:
> I only scrolled over it. Did I miss anything in here that requires a proper review?

There shouldn't be hidden traps or anything. STRONG_TYPEDEF includes generating a operator<<, but that's the most significant thing that comes to mind.

src/lib/types.hpp
mrks: No, it was also included before. You switched it from <> to "".

mrzzzrm: probably because I like using <> for stdlib an "" for everything else... Want me to revert it?

mrks: "" looks in the local directory before it looks in the system files. Thus, all system headers should use <>

mrzzzrm: a right. i keep on forgetting we do not ship boost in-repository.

PR #1596, opened 2019-04-13
Enabled TPC-H table caching

mrks:
On my macbook, this reduces the TPC-H startup time for SF 10 from 8m14.176s to 3m41.794s.

### DIFF
--- a/src/benchmarklib/benchmark_runner.cpp
+++ b/src/benchmarklib/benchmark_runner.cpp
@@ -51,6 +51,12 @@ BenchmarkRunner::BenchmarkRunner(const BenchmarkConfig& config, std::unique_ptr<
 
     const auto scheduler = std::make_shared<NodeQueueScheduler>();
     CurrentScheduler::set(scheduler);
+
+    if (config.clients == 1) {
+      std::cout << "\n\n- WARNING: You are running in multi-threaded (MT) mode but have set --clients=1.";
+      std::cout << "           You will achieve better MT performance by executing multiple queries in parallel.\n";
+      std::cout << std::endl;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/benchmarklib/benchmark_runner.cpp
```diff
@@ -51,6 +51,12 @@ BenchmarkRunner::BenchmarkRunner(const BenchmarkConfig& config, std::unique_ptr<
 
     const auto scheduler = std::make_shared<NodeQueueScheduler>();
     CurrentScheduler::set(scheduler);
+
+    if (config.clients == 1) {
```
mrzzzrm: I'd put this into BenchmarkRunner::run() so that it prints when I - as a creator of the BenchmarkRunner - would probably more expect it to print.

src/benchmarklib/tpch/tpch_table_generator.cpp
```diff
@@ -343,6 +362,13 @@ std::unordered_map<std::string, BenchmarkTableInfo> TpchTableGenerator::generate
   table_info_by_name["nation"].table = nation_builder.finish_table();
   table_info_by_name["region"].table = region_builder.finish_table();
 
+  if (_benchmark_config->cache_binary_tables) {
+    std::filesystem::create_directories(cache_directory);
+    for (auto& [table_name, table_info] : table_info_by_name) {
+      table_info.binary_file_path = cache_directory + "/" + table_name + ".tbl";
```
mrzzzrm: wip? (doesn't seem to write out anything)

mrzzzrm:
Other than that, no objections.

src/benchmarklib/benchmark_runner.cpp
mrks: cli_config_parser.cpp is probably even better

mrks:
I am working on a big rewrite of the benchmark runner (so that it can deal with TPC-C) and will integrate it there.

PR #1597, opened 2019-04-15
BenchmarkRunner: Run items, not queries

### LINKED ISSUE #1302: BenchmarkRunner should visualize as soon as possible (fixed in #1597)
not at the very end of the benchmark. 

* while the benchmark is still running, you might already want to look at the plans of queries that completed. 

* If a query gets stuck and you have to kill the benchmark, e.g. because the plan is inefficient, you might want to look at the optimized LQP anyway, to get a clue why the query was inefficient.

### LINKED ISSUE #1383: Benchmark clients are all using the same thread (fixed in #1597)
...especially, this means that they only use a single thread to translate and optimize queries and this could be a bottleneck. Join Order Benchmark queries can take non-negligible time to optimize.

I am not sure what the right course would be here? Surely we cannot reserve a full core for each client. Maybe translation and optimization should be Tasks as well?

(Also, I am not sure about calling `std::this_thread::sleep_for(std::chrono::milliseconds(10));` when there are "sufficient" queries running at the moment. This halts the "client thread" for a considerable time and could be problematic for these OLTP workloads, maybe? At least use condition variables to block and unblock the client thread?)

### LINKED ISSUE #1572: Support --cache_binary_tables for TPC-H (fixed in #1597)
With SF 100, nemea takes almost an hour to generate and encode the data. Adding `--cache_binary_tables` should not be too difficult and should same quite some time.

mrks:
The item of work in the TPC-C benchmark is not a query, but a procedure, which consists of a number of queries interleaved with logic. The current benchmark runner only handles single SQL strings.

This PR refactors the BenchmarkRunner and associated classes and:
* Switches from queries to generalized "benchmark items"
* Enables table caching for TPC-H (from #1596, fixes #1572)
* **Removes the option of running benchmarks without MVCC. This means that the results of `./hyriseBenchmarkTPCH` will be lower than before.** If we claim to be a database, we can't produce numbers that are on read-only tables. Also, running with MVCC makes it easier to spot problems, for example the missing MVCC information of the tables that are loaded initially. (#1603)
* Moves the SQL parsing into a JobTask, fixes #1383 
* Cleans up the BenchmarkRunner quite a bit
* Stops the visualizer from spamming .dot files that no one ever looks at
* Makes TPC-H Q15 work with multi-threading
* Gets rid of the tpchTableGenerator binary. <s>If I hadn't mentioned it, I bet nobody would have even noticed it. ;)</s> Ok, it was handy for verifying the table generation (#1608), but that was a one time thing. The console can do the same.
* Makes the visualizer visualize immediately, fixes #1302. Currently, it visualizes every run, overwriting previous results. One could add a check if the query was already visualized, but that would add plumbing issues. Is it really a problem if we visualize more than once? Simply use `-r 1`.

The introduced AbstractBenchmarkItemRunner is basically the old AbstractQueryGenerator, only that it is now the job of that class to run SQL queries. These are not run directly through the SQLPipeline, but handled by the BenchmarkSQLExecutor, which deals with SQLite verification and visualization.

Performance change for MT on nemea with 100 clients (as image, because the comparison script doesn't want to compare old and new benchmarks):

Screenshot 2019-04-22 at 13 46 26


scripts/cpplint.py
```diff
@@ -367,6 +367,7 @@
     'regex',
     'scoped_allocator',
     'set',
+    'shared_mutex',
```
mrks: TODO fix compare_benchmarks.py

mrks: TODO check JSON output

mrks: TODO test file-based / JOB

mrks:
@julianmenzler @d-justen @engelfa 
I'd appreciate it if you could sit down with @mrzzzrm and review this and a couple of future PRs. Not necessarily all of you at the same time for the same PR, but we should make sure that @mrzzzrm's review skills are passed on.

mrzzzrm:
@d-justen, @engelfa and I agreed to do so Tue/Wed next week. I'd "save" this PR for that occasion and not look at it until then - unless you tell me otherwise

mrks:
Sounds good

julianmenzler:
> @d-justen, @engelfa and I agreed to do so Tue/Wed next week. I'd "save" this PR for that occasion and not look at it until then - unless you tell me otherwise

Great, then I will join you.

src/test/scheduler/scheduler_test.cpp
```diff
@@ -126,6 +126,31 @@ TEST_F(SchedulerTest, BasicTest) {
...
+
+  auto task = std::make_shared<JobTask>([&]() {
+    // Wait for test to increment the counter, then increment it again
+    while (counter == 0) {
+    }
+    ++counter;
+  });
+  task->schedule();
+
+  ASSERT_TRUE(CurrentScheduler::has_pending_tasks());
```
mrks: This test confuses pending tasks and running tasks. The change to the BenchmarkRunner should also check for both. Will rework this. Don't let this stop you from reviewing the rest.

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -116,346 +101,234 @@ void BenchmarkRunner::run() {
 
   // Create report
   if (_config.output_file_path) {
-    std::ofstream output_file(*_config.output_file_path);
-    _create_report(output_file);
+    if (!_config.verify && !_config.enable_visualization) {
```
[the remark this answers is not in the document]
mrks: With a bit more of plumbing, sure. However, I am more tempted to hide all execution times or at least add another big fat warning if visualization is enabled. You can't get meaningful execution times if you do significant work, write files, and spawn processes in between.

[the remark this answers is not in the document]
mrzzzrm: I think the big fat warning is fine, but I should get a report if I requested one. Having the vis/verify times as part of the "item" time seems like an unclean idea and should be avoided if reasonably possible.

src/benchmarklib/benchmark_sql_executor.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include "sql/sql_pipeline.hpp"
+#include "utils/sqlite_wrapper.hpp"
+
+namespace opossum {
+
+class TransactionContext;
+
+// This class provides SQL functionality to BenchmarkItemRunners. See AbstractBenchmarkItemRunner:_execute_item.
```
mrzzzrm: + "every query executed via this interface will also, if requested, be validated against sql and/or visualized"

src/benchmarklib/benchmark_sql_executor.hpp
```diff
@@ -0,0 +1,33 @@
...
+class BenchmarkSQLExecutor {
+ public:
+  // If sqlite_wrapper is set, all SQL queries executed in the benchmark will also be executed using SQLite and then
+  // validated.
+  BenchmarkSQLExecutor(bool enable_jit, const std::shared_ptr<SQLiteWrapper>& sqlite_wrapper,
+                       const std::optional<std::string>& visualize_prefix);
```
mrzzzrm: Please doc "visualize_prefix", I'm not perfectly sure what path (?) it contains

src/benchmarklib/benchmark_sql_executor.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+  std::shared_ptr<const Table> execute(const std::string& sql);
+  std::vector<SQLPipelineMetrics> metrics;
```
mrzzzrm: pls doc: one entry for each execute() call, i assume?
 also a newline before this would lead to better grouping of members

src/benchmarklib/benchmark_sql_executor.hpp
mrzzzrm: + "represents one transaction"?

src/benchmarklib/benchmark_sql_executor.cpp
```diff
@@ -0,0 +1,78 @@
...
+      std::cout << "- Verification failed: Hyrise returned a result, but SQLite didn't" << std::endl;
+    } else if (!check_table_equal(result_table, sqlite_result, OrderSensitivity::No, TypeCmpMode::Lenient,
+                                  FloatComparisonMode::RelativeDifference)) {
+      any_verification_failed = true;
+      std::cout << "- Verification failed: Tables are not equal" << std::endl;
+    }
+  } else {
+    if (sqlite_result && sqlite_result->row_count() > 0) {
+      any_verification_failed = true;
+      std::cout << "- Verification failed: SQLite returned a result, but Hyrise did not" << std::endl;
```
mrzzzrm: "did not" here and "didn't" above :P :P

src/benchmarklib/benchmark_sql_executor.cpp
```diff
@@ -0,0 +1,78 @@
...
+void BenchmarkSQLExecutor::_visualize(SQLPipeline& pipeline) const {
+  GraphvizConfig graphviz_config;
+  graphviz_config.format = "svg";
+
+  const auto& lqps = pipeline.get_optimized_logical_plans();
+  const auto& pqps = pipeline.get_physical_plans();
+
+  for (auto lqp_idx = size_t{0}; lqp_idx < lqps.size(); ++lqp_idx) {
+    const auto file_prefix = *_visualize_prefix + "-LQP-" + std::to_string(lqp_idx);
+    LQPVisualizer{graphviz_config, {}, {}, {}}.visualize({lqps[lqp_idx]}, file_prefix + ".svg");
```
mrzzzrm: minor: why one file per statement instead of all planviz of a query in a single file? doc? (e.g., I'd like it better to have all 3 plans of TPC-H 15 in one file instead of in 3)

src/benchmarklib/benchmark_sql_executor.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+// This class provides SQL functionality to BenchmarkItemRunners. See AbstractBenchmarkItemRunner:_execute_item.
+class BenchmarkSQLExecutor {
+ public:
+  // If sqlite_wrapper is set, all SQL queries executed in the benchmark will also be executed using SQLite and then
+  // validated.
+  BenchmarkSQLExecutor(bool enable_jit, const std::shared_ptr<SQLiteWrapper>& sqlite_wrapper,
+                       const std::optional<std::string>& visualize_prefix);
+
+  std::shared_ptr<const Table> execute(const std::string& sql);
```
mrzzzrm: doc? this doesn't only execute, right? what else does it do? what does it return if "sql" contains multiple statements? or is it only designed for "sql" being a single statement

src/benchmarklib/cli_config_parser.cpp
```diff
@@ -57,28 +50,36 @@ BenchmarkConfig CLIConfigParser::parse_basic_options_json_config(const nlohmann:
...
-  std::cout << "- " + std::to_string(clients) + " simulated clients are scheduling queries in parallel" << std::endl;
+  std::cout << "- " + std::to_string(clients) + " simulated clients are scheduling items in parallel" << std::endl;
 
   if (cores != default_config.cores || clients != default_config.clients) {
     if (!enable_scheduler) {
       PerformanceWarning("'--cores' or '--clients' specified but ignored, because '--scheduler' is false")
     }
   }
 
+  if (enable_scheduler && clients == 1) {
```
mrzzzrm: what happens if someone calls the benchmark with `--clients 0`? or even `--clients -1`?

src/benchmarklib/abstract_benchmark_item_runner.hpp
```diff
@@ -0,0 +1,57 @@
...
+  // Executes the benchmark item with the given ID. BenchmarkItemRunners should not use the SQL pipeline directly,
+  // but use the provided BenchmarkSQLExecutor. That class not only tracks the execution metrics and provides them
+  // back to the benchmark runner, but it also implements SQLite verification and plan visualization.
+  virtual void _execute_item(const BenchmarkItemID item_id, BenchmarkSQLExecutor& sql_executor) = 0;
+
+  std::shared_ptr<BenchmarkConfig> _config;
+
+  std::shared_ptr<SQLiteWrapper> _sqlite_wrapper;
+
+  std::vector<BenchmarkItemID> _selected_items;
```
mrzzzrm: Does the fact that we have `available_item_count` and `_selected_items` have any advantages? Why do we not just have a set "std::vector<BenchmarkItemID> _items"? The AbstractBenchmarkItemRunner supplies the items to be benchmarked, does the BenchmarkRunner have to care that this is just a "selected" subset?

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
```diff
@@ -27,40 +28,50 @@ std::string calculate_date(boost::gregorian::date date, int months, int days = 0
...
 }
 
-TPCHQueryGenerator::TPCHQueryGenerator(bool use_prepared_statements, float scale_factor,
-                                       const std::vector<QueryID>& selected_queries)
-    : _use_prepared_statements(use_prepared_statements), _scale_factor(scale_factor) {
-  _selected_queries = selected_queries;
+TPCHBenchmarkItemRunner::TPCHBenchmarkItemRunner(const std::shared_ptr<BenchmarkConfig>& config,bool use_prepared_statements, float scale_factor,
+                                                 const std::vector<BenchmarkItemID>& selected_items)
+    : AbstractBenchmarkItemRunner(config), _use_prepared_statements(use_prepared_statements), _scale_factor(scale_factor) {
+  _selected_items = selected_items;
```
mrzzzrm: Why is this not a virtual method call? The way it is now, nothing enforces that the derived *BenchmarkItemRunner actually supplies this information.

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
```diff
@@ -27,40 +28,50 @@ std::string calculate_date(boost::gregorian::date date, int months, int days = 0
...
+  _prepare_queries();
```
mrzzzrm: Running queries/doing a lot of work - including optimization (which might even be of benchmarking-interest) - in a constructor seems suboptimal. Why not run the preparation of item X in the first call to `execute_item(X)`?

src/lib/visualization/abstract_visualizer.hpp
```diff
@@ -87,16 +87,21 @@ class AbstractVisualizer {
 
   virtual ~AbstractVisualizer() = default;
 
-  void visualize(const GraphBase& graph_base, const std::string& graph_filename, const std::string& img_filename) {
+  void visualize(const GraphBase& graph_base, const std::string& img_filename) {
     _build_graph(graph_base);
-    std::ofstream file(graph_filename);
+
+    char* tmpname = strdup("/tmp/hyrise_viz_XXXXXX");
+    auto ret = mkstemp(tmpname);
```
mrzzzrm: abbreviation ;)

src/test/tpc/tpch_test.cpp
```diff
@@ -44,11 +44,16 @@ class TPCHTest : public BaseTestWithParam<TPCHTestParam> {
       {1, 0.01f},   {2, 0.004f},  {3, 0.01f},  {4, 0.005f},  {5, 0.01f},    {6, 0.01f},  {7, 0.01f},  {8, 0.01f},
       {9, 0.01f},   {10, 0.02f},  {11, 0.01f}, {12, 0.01f},  {13, 0.01f},   {14, 0.01f}, {15, 0.01f}, {16, 0.01f},
       {17, 0.013f}, {18, 0.005f}, {19, 0.01f}, {20, 0.008f}, {21, 0.0075f}, {22, 0.01f}};
+
+  // Helper method used to touch TPCHBenchmarkItemRunner's privates
+  std::string get_deterministic_query(TPCHBenchmarkItemRunner& runner, BenchmarkItemID item_id) {
+    return runner._build_deterministic_query(item_id);
```
mrzzzrm: What's the danger of making this public?

src/benchmarklib/abstract_benchmark_item_runner.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+  const bool _enable_visualization;
```
mrzzzrm: this info should also be in `_config`, right? why duplicate it here?

src/benchmarklib/benchmark_item_result.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+  std::atomic<size_t> num_iterations = 0;
+
+  // Used only for Ordered mode
+  uint64_t all_runs_duration_ns{0};
+
+  // Stores the runtime of the iterations of this item
+  tbb::concurrent_vector<Duration> durations;
+
+  // Holds one entry per execution of this item. The vector holds one entry per SQL pipeline executed as part of this
```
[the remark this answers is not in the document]
mrzzzrm: Comment might be incorrect: "One entry per Item"? Or "one entry per SQL pipeline"? those are two different things, right?

scripts/benchmark_multithreaded.py
```diff
@@ -29,7 +29,7 @@
```
mrzzzrm: (random line)
@mrks What about introducing a "report_format_version" attribute in the report context, given that we'll probably keep on changing the format of the JSON-report. With this, a BenchmarkCI could remain compatible with old commits as well, when newer have a new JSON-report version.

mrks: I thought about that for a second, but the time where we compare different versions is probably limited. Having to maintain multiple versions of something is generally a pain. I believe it would be easier to hack the compare_benchmark script accordingly if needed.

src/test/tpc/tpch_test.cpp
mrks: No danger in particular, but something that is only supposed to be used by tests seems like a candidate for a non-public interface. No strong feeling about this.

scripts/benchmark_multithreaded.py
mrzzzrm: If we ever want to re-benchmark older versions of hyrise we're pretty much lost if we do not have the report-version. Just add it, doesn't hurt if it's there?

mrzzzrm: Mainly thinking about a potential benchmark CI/speed center here

mrks: Ah, in the C++ code, you just want to add an identifier. Sure, we can do that. I first thought that you wanted to emit both versions.

mrzzzrm: correct, I want the speedcenter to deal with multiple JSON-report versions/layouts, if anyone has to.

mrks: Any idea how to enforce the correctness of this value? Until we have a user of this value, I foresee situations where the JSON is changed but the value is not. Some time later, there would be a "Oups, forgot it" commit; so instead of two versions, we now have three. If that happens, the JSON user would have to identify the version by the presence / absence of fields instead of the version ID.

mrzzzrm: Not really, tbh, we would just have to look out for it, I suppose... :( But given the infrequency that changes to the JSON format would happen, we would hopefully catch it in reviews?

mrks: Given the infrequency of those changes, I would expect people to have already forgotten about this (if they knew about it in the first place) ;)

scripts/compare_benchmarks.py
```diff
@@ -34,30 +34,24 @@ def get_iteration_durations(iterations):
     iteration_durations = []
     for iteration in iterations:
         iteration_duration = 0
-        for statement in iteration["statements"]:
-            iteration_duration += statement["sql_translation_duration"] + statement["optimization_duration"] + \
```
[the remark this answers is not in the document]
mrks: The JSON still contains it, but it is not used for the comparison, correct.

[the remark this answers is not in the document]
mrks: It even contains one entry per SQL query, so you can see the optimization time of the separate SQL steps of a TPC-C query.

src/benchmarklib/abstract_benchmark_item_runner.cpp
```diff
@@ -0,0 +1,31 @@
...
+    const BenchmarkItemID item_id) {
+  std::optional<std::string> visualize_prefix;
+  if (enable_visualization) {
+    auto name = item_name(item_id);
+    boost::replace_all(name, " ", "_");
+    visualize_prefix = std::move(name);
+  }
+
+  BenchmarkSQLExecutor sql_executor(enable_jit, _sqlite_wrapper, visualize_prefix);
+  _execute_item(item_id, sql_executor);
```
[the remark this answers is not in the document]
mrks: Can't come up with anything better than `_execute_item_impl()`

src/benchmarklib/abstract_benchmark_item_runner.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include "benchmark_item_result.hpp"
+#include "benchmark_sql_executor.hpp"
+#include "strong_typedef.hpp"
+
+STRONG_TYPEDEF(size_t, BenchmarkItemID);
+
+namespace opossum {
+
+// Item generators are used by benchmarks to generate SQL strings for the different benchmark items. In their
```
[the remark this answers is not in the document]
mrks: That comment was not fully updated yet. Better now?

src/benchmarklib/abstract_benchmark_item_runner.hpp
```diff
@@ -0,0 +1,54 @@
...
+  // Returns the number of items supported by the benchmark
+  virtual size_t available_item_count() const = 0;
+
+  // Returns the BenchmarkItemIDs of all selected items
+  const std::vector<BenchmarkItemID>& selected_items() const;
+
+  // Set the SQLite wrapper used for query verification. `nullptr` disables verification.
+  void set_sqlite_wrapper(std::shared_ptr<SQLiteWrapper> sqlite_wrapper);
+
+  bool enable_jit;
```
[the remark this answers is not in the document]
mrks: Is gone anyway.

src/benchmarklib/benchmark_item_result.hpp
```diff
@@ -0,0 +1,36 @@
...
+  BenchmarkItemResult();
+
+  // Need to explicitly implement move, because std::atomic implicitly deletes it...
+  BenchmarkItemResult(BenchmarkItemResult&& other) noexcept;
+  BenchmarkItemResult& operator=(BenchmarkItemResult&& other);
+
+  std::atomic<size_t> num_iterations = 0;
+
+  // Used only for Ordered mode
+  uint64_t all_runs_duration_ns{0};
```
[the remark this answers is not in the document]
mrks: changed

src/benchmarklib/benchmark_item_result.hpp
```diff
@@ -0,0 +1,36 @@
...
+  // Stores the runtime of the iterations of this item
+  tbb::concurrent_vector<Duration> durations;
+
+  // Holds one entry per execution of this item. The vector holds one entry per SQL pipeline executed as part of this
+  // item. For benchmarks like TPC-H, where each item corresponds to a single TPC-H query, this vector always has a
+  // size of 1. For others, like TPC-C, there are multiple SQL queries executed and thus multiple entries in the inner
+  // vector.
+  tbb::concurrent_vector<std::vector<SQLPipelineMetrics>> metrics;
+
+  // Set if verification was requested and this item was executed
```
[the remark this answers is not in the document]
mrks: The *optional* is set if the verification was executed; the *bool* is true if the verification succeeded.

src/benchmarklib/benchmark_item_result.hpp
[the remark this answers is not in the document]
mrks: Added: "The *inner* vector holds one entry per SQLPipeline [...]"

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -53,34 +49,25 @@ BenchmarkRunner::BenchmarkRunner(const BenchmarkConfig& config, std::unique_ptr<
...
+
   _table_generator->generate_and_store();
 
   if (_config.verify) {
     std::cout << "- Loading tables into SQLite for verification." << std::endl;
     Timer timer;
 
     // Load the data into SQLite
-    sqlite_wrapper = std::make_unique<SQLiteWrapper>();
+    sqlite_wrapper = std::make_shared<SQLiteWrapper>();
```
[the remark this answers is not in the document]
mrks: As we don't have modifying transactions right now, this is not a problem. For TPC-C, I will need to have another look.

src/benchmarklib/benchmark_runner.hpp
```diff
@@ -42,63 +43,45 @@ class BenchmarkRunner {
...
   nlohmann::json _context;
 
   std::optional<PerformanceWarningDisabler> _performance_warning_disabler;
 
   Duration _total_run_duration{};
+
+  // The atomic uints are modified by other threads when finishing an item, to keep track of when we can
+  // let a simulated client schedule the next item, as well as the total number of finished queries so far
+  std::atomic_uint _currently_running_clients{0};
+  std::atomic_uint _total_finished_runs{0};
```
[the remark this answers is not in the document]
mrks: Like this?

src/benchmarklib/benchmark_sql_executor.hpp
mrks: >     * "every query executed via this interface will also, if requested, be validated against sql and/or visualized"

Done, even though I don't see a significant difference to the previous comment on the constructor.

>     * "represents one transaction"?

"For now, we expect items to use a single transaction, which is why the BenchmarkSQLExecutor executes all queries in the same context."

src/benchmarklib/benchmark_sql_executor.hpp
mrks: Like this?

src/benchmarklib/benchmark_sql_executor.cpp
mrks: Legacy, fixed

src/benchmarklib/benchmark_sql_executor.cpp
mrks: created #1654

src/benchmarklib/benchmark_sql_executor.hpp
mrks: From the perspective of someone implementing a new benchmark, that's the key functionality of this class. The verification / visualization is nothing that I should care about when implementing a new benchmark.

src/benchmarklib/cli_config_parser.cpp
mrks: Added assert

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
mrks: I can't follow. If you don't supply it, the other constructor will be used.

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
mrks: Previously, the preparation was also outside of the benchmark. I'd rather keep execute_item simple.

src/lib/visualization/abstract_visualizer.hpp
mrks: Consistent with `console.cpp`. `return` is not an option.

src/benchmarklib/abstract_benchmark_item_runner.hpp
mrks: leftover

Bouncner:
To get a feeling for the MVCC impact: could you add a comparison of the single-threaded perf before (MVCC off by default) and after (on).

src/benchmarklib/abstract_benchmark_item_runner.cpp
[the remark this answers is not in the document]
mrzzzrm: sometimes we call them `_on_***` anything that differentiates the name from `execute_item` improves things.

src/benchmarklib/benchmark_runner.hpp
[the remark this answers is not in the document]
mrzzzrm: Hyrise uses the term "runs" to mean consecutive equal values (see RLE, JSM). Coming right out of the join operators, this confused me. Maybe "executions", "item_executions", or just "items"? 

comment still uses the term "queries" instead of "items"

src/benchmarklib/benchmark_sql_executor.hpp
mrzzzrm: (I first read class comments, not constructor comments.)

src/benchmarklib/benchmark_sql_executor.hpp
mrzzzrm: if you'd like to make me very happy, also doc (here) what the suffix is comprised of

src/benchmarklib/benchmark_sql_executor.hpp
mrzzzrm: all fine, but just a short comment about a long the lines of "this triggers everything to be done for a query, from execution to - optionally - verification and visualization"

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
mrzzzrm: I meant: Why is _selected_items a base class member variable (that a derived class needs to remember to set!!) instead of having `virtual AbstractBenchmarkItemRunner::selected_items() = 0` which forces deriving classes to implement it.

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
mrzzzrm: I won't let "previously" count as an argument - this PR is supposed to get us closer to a "proper" benchmarking suite. 


If a benchmark does what I'd regard as substantial work (optimizing those prepared queries) I expect it to report what it did. 
From my personal use case of benchmarking, monitoring optimization is just as important as monitoring execution.

src/lib/visualization/abstract_visualizer.hpp
mrzzzrm: kk, just wanted to be a pain.

scripts/benchmark_multithreaded.py
mrzzzrm: Basic testing for the JSON format could pretty easily be integrated into #1103. I'd add the `report_version`. Thank me in a year or so, when you can look at the shiny performance development of Hyrise, beautifully plotted over the course of months and years, not just the last few commit until the BenchmarkCI ran into a JSON that it didn't understand.

mrks: Wait a second - we already *have* a version identifier, which is guaranteed to be always up-to-date: Why don't we use the git hash?

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
mrks: >  From my personal use case of benchmarking, monitoring optimization is just as important as monitoring execution.

Which is why preparation is disabled by default.

> If a benchmark does what I'd regard as substantial work (optimizing those prepared queries) I expect it to report what it did.

Hiding it in the first execution of a query doesn't seem like the right way to do it either though. If the TPC-H was a real application, it would also register its prepared statements before they are actually used.

scripts/benchmark_multithreaded.py
mrzzzrm: Because we don't know all the git hashes that use a particular JSON format.

mrks: We don't need to. If we change the format at a certain point, we just need to tell the consumer that "from this hash on, you need to xyz".

src/benchmarklib/benchmark_sql_executor.hpp
mrks: That's the job of the caller and there is no real methodology. For TPC-H, it has a query number, for TPC-C, it is the name of the procedure, ...

src/benchmarklib/abstract_benchmark_item_runner.hpp
mrks: The results are stored in a vector at the position of their `item_id`. Thus, we need to know what size the vector should be initialized with.

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
mrks: Gotcha.

mrks:
```
+----------------+----------------+------+---------------+------+------------+---------------------------------+
| Benchmark      | prev. iter/s   | runs | new iter/s    | runs | change [%] | p-value (significant if <0.001) |
+----------------+----------------+------+---------------+------+------------+---------------------------------+
| TPC-H 1        | 0.358328700066 | 22   | 0.42716935277 | 26   | +19%       |                          0.0000 |
| TPC-H 2        | 7.59358501434  | 456  | 7.60669279099 | 457  | +0%        |                          0.2126 |
| TPC-H 3        | 5.65848922729  | 340  | 5.0229010582  | 302  | -11%       |                          0.0000 |
| TPC-H 4        | 2.02712988853  | 122  | 1.87886452675 | 113  | -7%        |                          0.0000 |
| TPC-H 5        | 4.0936832428   | 246  | 3.46917152405 | 209  | -15%       |                          0.0000 |
| TPC-H 6        | 5.03204631805  | 302  | 5.8973197937  | 354  | +17%       |                          0.0000 |
| TPC-H 7        | 1.18088662624  | 71   | 1.25977325439 | 76   | +7%        |                          0.0000 |
| TPC-H 8        | 4.79884815216  | 288  | 4.01583480835 | 241  | -16%       |                          0.0000 |
| TPC-H 9        | 1.42111325264  | 86   | 1.4606025219  | 88   | +3%        |                          0.0000 |
| TPC-H 10       | 2.39834499359  | 144  | 2.66411328316 | 160  | +11%       |                          0.0000 |
| TPC-H 11       | 28.6810150146  | 1721 | 23.5143489838 | 1411 | -18%       |                          0.0000 |
| TPC-H 12       | 6.79630470276  | 408  | 5.99238824844 | 360  | -12%       |                          0.0000 |
| TPC-H 13       | 1.95415484905  | 118  | 2.15863370895 | 130  | +10%       |                          0.0000 |
| TPC-H 14       | 20.0237598419  | 1202 | 13.404255867  | 805  | -33%       |                          0.0000 |
| TPC-H 15       | 17.2805080414  | 1037 | 7.97658443451 | 479  | -54%       |                          0.0000 |
| TPC-H 16       | 7.10105037689  | 427  | 6.72212982178 | 404  | -5%        |                          0.0000 |
| TPC-H 17       | 1.35094380379  | 82   | 1.1169500351  | 68   | -17%       |                          0.0000 |
| TPC-H 18       | 1.89737641811  | 114  | 1.30373120308 | 79   | -31%       |                          0.0000 |
| TPC-H 19       | 5.46647787094  | 328  | 5.13108873367 | 308  | -6%        |                          0.0000 |
| TPC-H 20       | 2.06289076805  | 124  | 1.83635878563 | 111  | -11%       |                          0.0000 |
| TPC-H 21       | 0.416004806757 | 25   | 0.34450879693 | 21   | -17%       |                          0.0000 |
| TPC-H 22       | 12.9294137955  | 776  | 10.2968244553 | 618  | -20%       |                          0.0000 |
| geometric mean |                |      |               |      | -11%       |                                 |
+----------------+----------------+------+---------------+------+------------+---------------------------------+
```

No idea why some of them improve.

scripts/benchmark_multithreaded.py
mrzzzrm: Mmh, I guess that would work....

Bouncner:
Strange. Locally on my mac, Q1 took a significant hit with MVCC enabled.

mrks:
Anything other than the scheduler thin blocking this? I not, I'd suggest to spin that one out and get done here quickly.

mrks:
Ah, wait. The selected items thing is also missing.

mrzzzrm:
I cannot promise I will be able to look into this (or anything else, really) until Monday.

mrks:
no worries. Just wanted to make sure I'mnot blocking anything.
________________________________

mrzzzrm:
> @mrzzzrm : any ideas why the first projection once is placed before the predicates and once after? 

After #1664 there should be less Projections, maybe that helps here.

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -93,20 +77,18 @@ BenchmarkRunner::~BenchmarkRunner() {
 void BenchmarkRunner::run() {
   std::cout << "- Starting Benchmark..." << std::endl;
 
-  const auto available_queries_count = _query_generator->available_query_count();
-  _query_plans.resize(available_queries_count);
-  _query_results.resize(available_queries_count);
+  const auto available_item_count = _benchmark_item_runner->available_item_count();
+  _results.resize(available_item_count);
```
mrzzzrm: Using a vector-as-a-map here doesn't work, right? Because `available_item_count` would be 3 for `selected_items()` `{1, 15, 16}` but would be accessed with an index of `16` later, right? I made `_results` an undered_map

mrzzzrm: ah, we allocated 22 slots and then only filled in those that are in selected_items(), I see.. Added a bit of commenting to the _results declaration in the header

src/benchmarklib/benchmark_sql_executor.hpp
mrzzzrm: well, that's not true, right? The suffix is just "LQP/PQP" + index. Will doc that

src/benchmarklib/benchmark_sql_executor.hpp
```diff
@@ -0,0 +1,40 @@
...
+// If sqlite_wrapper is set, all SQL queries executed in the benchmark will also be executed using SQLite and then
+// validated. For now, we expect items to use a single transaction, which is why the BenchmarkSQLExecutor executes
+// all queries in the same context.
+class BenchmarkSQLExecutor {
+ public:
+  // @param visualize_prefix    Prefix for the filename of the generated query plans (e.g., "TPC-H_6-")
+  BenchmarkSQLExecutor(bool enable_jit, const std::shared_ptr<SQLiteWrapper>& sqlite_wrapper,
+                       const std::optional<std::string>& visualize_prefix);
+
+  // This executes the given SQL query and returns a single table. If visualization and/or verification are enabled,
```
mrzzzrm: Added "records its metrics and returns a single table (the same as SQLPipeline::get_result_table() would)."

mrks:
@mrzzzrm Please have a look at c72238a. I removed the whole checks for pending tasks. First, it is unclear what "pending" actually means. It might have different results if a worker removes a task for execution in the same second. Second, it is not the benchmark's job to worry about this. A plugin might have spawned tasks that are unrelated to benchmark items, so we shouldn't mess with that.

I went to your changes and agree with them. Thanks for those!

With that, I think everything in here should be addressed.

mrzzzrm:
> @mrzzzrm Please have a look at c72238a.

Looks sensible to me. I'll look through this PR again, at the latest tomorrow morning

src/benchmarklib/benchmark_runner.hpp
```diff
@@ -31,9 +33,8 @@ class SQLiteWrapper;
 
 class BenchmarkRunner {
```
mrzzzrm: Optional, maybe it's time to give this class a one-liner doc comment?

src/benchmarklib/benchmark_sql_executor.cpp
```diff
@@ -0,0 +1,86 @@
...
+  if (_visualize_prefix) {
+    _visualize(pipeline);
+  }
+
+  return result_table;
+}
+
+void BenchmarkSQLExecutor::_verify_with_sqlite(SQLPipeline& pipeline) {
+  const auto sqlite_result = _sqlite_wrapper->execute_query(pipeline.get_sql());
+  const auto& result_table = pipeline.get_result_table();
```
mrzzzrm: why not `get_result_tables()`? the way it is done now, only the last statement in a query if verified against sqlite?

src/benchmarklib/cli_config_parser.cpp
```diff
@@ -40,15 +40,8 @@ BenchmarkConfig CLIConfigParser::parse_basic_options_json_config(const nlohmann:
   if (!output_file_string.empty()) {
     output_file_path = output_file_string;
     std::cout << "- Writing benchmark results to '" << *output_file_path << "'" << std::endl;
-  } else {
-    std::cout << "- Writing benchmark results to stdout" << std::endl;
```
mrzzzrm: I always liked it that each benchmark option printed a line at benchmark startup. What will this print if `-o` is not set? Nothing?

src/test/tpc/tpch_test.cpp
```diff
@@ -44,11 +44,16 @@ class TPCHTest : public BaseTestWithParam<TPCHTestParam> {
       {1, 0.01f},   {2, 0.004f},  {3, 0.01f},  {4, 0.005f},  {5, 0.01f},    {6, 0.01f},  {7, 0.01f},  {8, 0.01f},
       {9, 0.01f},   {10, 0.02f},  {11, 0.01f}, {12, 0.01f},  {13, 0.01f},   {14, 0.01f}, {15, 0.01f}, {16, 0.01f},
       {17, 0.013f}, {18, 0.005f}, {19, 0.01f}, {20, 0.008f}, {21, 0.0075f}, {22, 0.01f}};
+
+  // Helper method used to touch TPCHBenchmarkItemRunner's privates
```
mrzzzrm: I would... [rephrase that comment](https://www.dict.cc/englisch-deutsch/privates.html)

src/benchmarklib/benchmark_config.hpp
```diff
@@ -8,10 +8,10 @@
 namespace opossum {
 
 /**
- * IndividualQueries runs each query a number of times and then the next one
- * PermutedQuerySet runs the queries as set permuting their order after each run (this exercises caches)
+ * "Ordered" runs each item a number of times and then the next one
+ * "Shuffled" runs the items as set permuting their order after each run
```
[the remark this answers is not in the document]
mrks: I think the whole set part is too implementation-specific. Changed it to ` * "Shuffled" runs the items in a random order`

src/benchmarklib/benchmark_item_result.hpp
```diff
@@ -0,0 +1,36 @@
...
+  BenchmarkItemResult();
+
+  // Need to explicitly implement move, because std::atomic implicitly deletes it...
+  BenchmarkItemResult(BenchmarkItemResult&& other) noexcept;
+  BenchmarkItemResult& operator=(BenchmarkItemResult&& other);
+
+  std::atomic<size_t> num_iterations = 0;
+
+  // Used only for BenchmarkMode::Ordered mode
+  Duration all_runs_duration{0};
```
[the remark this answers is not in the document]
mrks: `duration_of_all_runs`?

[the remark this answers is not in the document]
mrzzzrm: okay

src/benchmarklib/benchmark_config.hpp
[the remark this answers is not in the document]
mrzzzrm: okay

src/benchmarklib/benchmark_sql_executor.cpp
mrks: As far as I know, the SQLite wrapper cannot handle multiple statements with multiple results yet. I'd rather require the pipeline to contain only a single statement.

src/benchmarklib/cli_config_parser.cpp
mrks: Correct. The previous output was wrong - we always print the readable output to the console and never print the JSON to stdout. I don't feel like arguing about this for too long, but I rather not add lines to the output that contain no information.

src/benchmarklib/benchmark_sql_executor.cpp
mrzzzrm: > I'd rather require the pipeline to contain only a single statement

I'm okay with that. Is there an Assert enforcing this?

src/benchmarklib/cli_config_parser.cpp
mrzzzrm: don't feel like arguing about this either. leave it as it is then

src/test/tpc/tpch_test.cpp
mrks: Oh. Oups. Yes, I would never make such an immature joke. ^^

src/benchmarklib/benchmark_sql_executor.cpp
mrks: Soon.

src/benchmarklib/benchmark_runner.hpp
mrks:
```
// The BenchmarkRunner is the main class for the benchmark framework. It gets initialized by the benchmark binaries
// (e.g., tpch_benchmark.cpp). They then hand over the control to the BenchmarkRunner (inversion of control), which
// calls the supplied table generator, runs and times the benchmark items, and reports the benchmark results.
```
?

mrzzzrm: great. dunno if "inversion of control" helps me, but maybe it helps others?

mrks: Add "(aka. Hollywood Principle)"?

mrzzzrm: nope, didn't ring a bell for me, prior to googling it.

PR #1599, opened 2019-04-15
[JIT] Remove necessity to access nullable information in JitAwareLQPTranslator

### LINKED ISSUE #1451: Remove (necessity to call) AbstractLQPNode::is_column_nullable() from JIT
As discussed, the only place where we currently need to figure out the nullability of  an expression in the LQP is the JITAwareLQPTranslator (Strg+F "is_column_nullable"). 
We should consider removing this necessity and make the JIT obtain nullability of an expression during operator execution, when the input table and with it proper nullability information is available.

(The reason we/I want "AbstractLQPNode::is_column_nullable" removed is because it is a rather awkward setup necessary because outer joins might unexpectedly introduce nullability into otherwise non-nullable columns/expression. It consists of altogether 27 overrides of `AbstractExpression::on_is_nullable_on_lqp` and `AbstractLQPNode::is_column_nullable()` that should be removed if it is feasible to refactor the JIT as described above.)

FabianWiebe:
This PR removes the necessity to access nullable information in the `JitAwareLQPTranslator`.

This requires that the nullable information in the `JitTupleEntry` instances, which are created during translation, must be updated before specialization. This is achieved by calling `before_specialization()` function on all JitOperators to allow them to update the nullable information in the `JitExpression` which they use.
The JitExpressions are updated recursively in `update_nullable_information`.

Fixes #1451


FabianWiebe:
I will close this pr until I fix the problem with `LLVM ERROR: Undefined temporary symbol`

FabianWiebe:
PR should work now.

src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
```diff
@@ -309,14 +309,15 @@ bool is_between_jit_expression_type_upper_inclusive(JitExpressionType jit_expres
...
+std::shared_ptr<JitExpression> JitAwareLQPTranslator::_try_translate_expression_to_jit_expression(
     const std::shared_ptr<AbstractExpression>& expression, JitReadTuples& jit_source,
     const std::shared_ptr<AbstractLQPNode>& input_node, const bool use_actual_value) const {
   const auto input_node_column_id = input_node->find_column_id(*expression);
   if (input_node_column_id) {
-    const auto tuple_entry = jit_source.add_input_column(
-        expression->data_type(), input_node->is_column_nullable(input_node->get_column_id(*expression)),
-        *input_node_column_id, use_actual_value);
+    // The correct nullable information is set when the JitOperatorWrapper is executed with access to the input table.
+    const bool is_nullabe = true;
```
mrks: typo. Also, it leaves us with wrong information for a while.`optional` or `bool guaranteed_non_null`?

src/lib/operators/jit_operator/operators/abstract_jittable.hpp
```diff
@@ -21,6 +21,11 @@ class AbstractJittable {
 
   std::shared_ptr<AbstractJittable> next_operator() { return _next_operator; }
 
+  /**
+   * Update the operator according to the input table before specialization
```
mrks: This doesn't tell me anything. Also, for a general "before" method, it takes quite specific arguments.

src/lib/operators/jit_operator/operators/jit_aggregate.cpp
```diff
@@ -21,6 +21,18 @@ std::string JitAggregate::description() const {
   return desc.str();
 }
 
+void JitAggregate::before_specialization(const Table& in_table, std::vector<bool>& tuple_nullable_information) {
+  // Update the tuple nullable information
```
mrks: then what does `update_nullable_information` do?

src/lib/operators/jit_operator/operators/jit_expression.hpp
```diff
@@ -73,10 +73,16 @@ class JitExpression {
   template <typename ResultValueType>
   std::optional<ResultValueType> compute(JitRuntimeContext& context) const;
 
-  const std::shared_ptr<const JitExpression> left_child;
-  const std::shared_ptr<const JitExpression> right_child;
+  /* Recursively update the nullable information of the result entries.
+   * The nullability information is not available during the creation of JitExpressions so that this information must
```
mrks: currently it is, isn't it?

src/lib/operators/jit_operator/operators/jit_validate.cpp
```diff
@@ -18,6 +18,10 @@ bool is_row_visible(const CommitID our_tid, const TransactionID row_tid, const C
 
 JitValidate::JitValidate(const TableType input_table_type) : input_table_type(input_table_type) {}
 
+void JitValidate::before_specialization(const Table& in_table, std::vector<bool>& tuple_nullable_information) {
+  input_table_type = in_table.type();
```
mrks: This shows a problem with the interface. Why do we pass in an unused parameter and do something different than in the other operators? Why do we overwrite the type that was explicitly set in the constructor?

src/lib/operators/jit_operator/operators/jit_aggregate.cpp
```diff
@@ -21,6 +21,18 @@ std::string JitAggregate::description() const {
...
 
+void JitAggregate::before_specialization(const Table& in_table, std::vector<bool>& tuple_nullable_information) {
+  // Update the groupby and aggregate column nullable information from the tuple_nullable_information
+  for (auto& groupby_column : _groupby_columns) {
+    groupby_column.tuple_entry.is_nullable = tuple_nullable_information[groupby_column.tuple_entry.tuple_index];
+    groupby_column.hashmap_entry.is_nullable = tuple_nullable_information[groupby_column.tuple_entry.tuple_index];
+  }
+
+  for (auto& aggregate_column : _aggregate_columns) {
+    aggregate_column.tuple_entry.is_nullable = tuple_nullable_information[aggregate_column.tuple_entry.tuple_index];
```
mrks: Wouldn’t this have to be updated? For example, `MIN(a)` could return NULL if `a` is a `NOT NULL` but empty column

src/lib/operators/jit_operator/operators/abstract_jittable.hpp
```diff
@@ -21,6 +21,18 @@ class AbstractJittable {
 
   std::shared_ptr<AbstractJittable> next_operator() { return _next_operator; }
 
+  /**
+   * Update the operator according to the encoding and nullable information of the input table columns before
```
mrks: How about this: “Before specialization, the Jittable is updated to include the encoding and nullability information of its input. As operators may change the nullability information, tuple_nullable_information is updated.”

src/test/operators/jit_operator/operators/jit_aggregate_test.cpp
```diff
@@ -341,4 +341,40 @@ TEST_F(JitAggregateTest, EmptyInputTableNoGroupbyColumns) {
...
+  jit_aggregate.add_groupby_column("c", tuple_entry_c);
+  jit_aggregate.add_groupby_column("d", tuple_entry_d);
+
+  // Update nullable information
+  // a and c are not nullable, b and d are nullable
+  std::vector<bool> tuple_nullable_information{false, true, false, true};
+  jit_aggregate.before_specialization(*input_table, tuple_nullable_information);
+
+  const auto& aggregate_columns = jit_aggregate.aggregate_columns();
+  EXPECT_FALSE(aggregate_columns[0].tuple_entry.is_nullable);
```
mrks: This looks incorrect, see above

src/lib/operators/jit_operator/operators/jit_validate.cpp
mrks: Brr..., yes. Also add a comment that the field will only be set correctly later.

src/lib/logical_query_plan/jit_aware_lqp_translator.cpp
mrks: If you don’t want an optional, having a flag "guaranteed to not be nullable” is different from “is_nullable” (or its opposite). For the former, `false` means “might or might not be nullable”, while it means “definitely not nullable” for the latter.

This would just be a naming change but would somewhat (but not completely) eliminate the problem with the incomplete state.

src/test/operators/jit_operator/operators/jit_aggregate_test.cpp
mrks: Add it to the test and explain which step (before or after the aggregate) you are testing?

src/lib/operators/jit_operator/operators/jit_aggregate.cpp
mrks: I'd add a comment saying explaining this.

FabianWiebe: Added a comment.

src/test/operators/jit_operator/operators/jit_aggregate_test.cpp
FabianWiebe: I added this to the test and added an explaining comment.

FabianWiebe:
I have now renamed `is_null` to `guaranteed_non_null` in `JitTupleEntry` and `JitHashmapEntry`.

Issue #1602, opened 2019-04-15
Get rid of mvcc_data()


mrks:
It seems unsafe to have an unlocked access method in addition to `get_scoped_mvcc_data_lock`. `Chunk::mvcc_data()` should go.

For example, this looks dangerous:
https://github.com/hyrise/hyrise/blob/2628fbfd09408563d3ca60f0e15f04b42e03a3da/src/lib/operators/projection.cpp#L100

While refactoring that part of the projection, we might also pass in the MVCC data directly instead of making the chunk resize the vectors first and replace them afterwards.

#749 might make the `mvcc_data()` part superfluous, but we should still look at the projection.

mrzzzrm:
This can be done very easily after #1647, I'm taking this.

mrks:
@mrzzzrm Still on this?

mrzzzrm:
Ah, forgot about this. I'm trying to step back a bit from tackling "small" issues. I think it's better if the new forces start working on these things, to gain some experience with Hyrise.

PR #1603, opened 2019-04-15
Fix MVCC performance regression, fix MVCC data for rows that are not inserted.

mrks:
#1585 changed the way MVCC vectors were extended. Resizing and writing a couple hundred atomics is quite expensive if we synchronize for each.

Also, the same PR changed new MVCC rows that they were not immediately visible. For the TPC-H with MVCC enabled, this meant that all tables were basically empty. This won't happen anymore with #1597, where we get rid of the option of running without MVCC altogether.

```
+-----------+----------------+------+----------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s   | runs | new iter/s     | runs | change | p-value (significant if <0.001) |
+-----------+----------------+------+----------------+------+--------+---------------------------------+
| TPC-H 1   | 0.533324956894 | 33   | 0.527172207832 | 32   | -1%    |                          0.1211 |
| TPC-H 2   | 7.16463375092  | 430  | 8.29684925079  | 498  | +16%   |                          0.0000 |
| TPC-H 3   | 6.57847356796  | 395  | 6.57278633118  | 395  | -0%    |                          0.6014 |
| TPC-H 4   | 1.90148293972  | 115  | 2.2281563282   | 134  | +17%   |                          0.0000 |
| TPC-H 5   | 3.23178911209  | 194  | 4.42444133759  | 266  | +37%   |                          0.0000 |
| TPC-H 6   | 29.7692642212  | 1787 | 29.3826828003  | 1763 | -1%    |                          0.0000 |
| TPC-H 7   | 1.2956045866   | 78   | 1.35758900642  | 82   | +5%    |                          0.0000 |
| TPC-H 8   | 4.04841804504  | 243  | 5.28122472763  | 317  | +30%   |                          0.0000 |
| TPC-H 9   | 1.37200808525  | 83   | 1.55264019966  | 94   | +13%   |                          0.0000 |
| TPC-H 10  | 2.75129842758  | 166  | 2.75398254395  | 166  | +0%    |                          0.7196 |
| TPC-H 11  | 20.1392726898  | 1209 | 31.9011650085  | 1915 | +58%   |                          0.0000 |
| TPC-H 12  | 7.31077384949  | 439  | 7.66930103302  | 461  | +5%    |                          0.0000 |
| TPC-H 13  | 2.22583818436  | 134  | 2.23531842232  | 135  | +0%    |                          0.5430 |
| TPC-H 14  | 23.4843826294  | 1410 | 24.4641456604  | 1468 | +4%    |                          0.0000 |
| TPC-H 15  | 19.5054988861  | 1171 | 19.4964313507  | 1170 | -0%    |                          0.4706 |
| TPC-H 16  | 7.12452030182  | 428  | 7.7613568306   | 466  | +9%    |                          0.0000 |
| TPC-H 17  | 1.17968988419  | 71   | 1.44373655319  | 87   | +22%   |                          0.0000 |
| TPC-H 18  | 1.50807476044  | 91   | 2.02401041985  | 122  | +34%   |                          0.0000 |
| TPC-H 19  | 5.29951524734  | 318  | 5.20063400269  | 313  | -2%    |                          0.0000 |
| TPC-H 20  | 2.00822305679  | 121  | 2.03985309601  | 123  | +2%    |                          0.0000 |
| TPC-H 21  | 0.425299465656 | 26   | 0.429752975702 | 26   | +1%    |                          0.1070 |
| TPC-H 22  | 10.0060491562  | 601  | 12.8247356415  | 770  | +28%   |                          0.0000 |
| average   |                |      |                |      | +13%   |                                 |
+-----------+----------------+------+----------------+------+--------+---------------------------------+
```

### DIFF
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -130,7 +130,7 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
       // new row count.
       {
         auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();
-        mvcc_data->grow_by(num_rows_for_target_chunk, false);
+        mvcc_data->grow_by(num_rows_for_target_chunk, MvccData::MAX_COMMIT_ID);
         for (auto chunk_offset = _target_chunk_ranges.back().begin_chunk_offset;
              chunk_offset < _target_chunk_ranges.back().end_chunk_offset; ++chunk_offset) {
           // This row is still invisible to everyone thanks to the begin set to infinity. We do this manually rather
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -130,15 +130,7 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
       // new row count.
       {
         auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();
-        mvcc_data->grow_by(num_rows_for_target_chunk, MvccData::MAX_COMMIT_ID);
-        for (auto chunk_offset = _target_chunk_ranges.back().begin_chunk_offset;
-             chunk_offset < _target_chunk_ranges.back().end_chunk_offset; ++chunk_offset) {
-          // This row is still invisible to everyone thanks to the begin set to infinity. We do this manually rather
-          // than passing context->transaction_id() because of performance reasons. No need for synchronization
-          // as long as we have a fence at the end.
-          mvcc_data->tids[chunk_offset].store(context->transaction_id(), std::memory_order_relaxed);
-        }
-        std::atomic_thread_fence(std::memory_order_seq_cst);
+        mvcc_data->grow_by(num_rows_for_target_chunk, context->transaction_id(), MvccData::MAX_COMMIT_ID);
       }
 
       // Grow data Segments.
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -130,7 +130,7 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
       // new row count.
       {
         auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();
-        mvcc_data->grow_by(num_rows_for_target_chunk);
+        mvcc_data->grow_by(num_rows_for_target_chunk, false);
         for (auto chunk_offset = _target_chunk_ranges.back().begin_chunk_offset;
              chunk_offset < _target_chunk_ranges.back().end_chunk_offset; ++chunk_offset) {
           // This row is still invisible to everyone thanks to the begin set to infinity. We do this manually rather
--- a/src/lib/operators/insert.cpp
+++ b/src/lib/operators/insert.cpp
@@ -128,7 +128,18 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
       // Grow MVCC vectors and mark new (but still empty) rows as being under modification by current transaction.
       // Do so before resizing the Segments, because the resize of `Chunk::_segments.front()` is what releases the
       // new row count.
-      target_chunk->get_scoped_mvcc_data_lock()->grow_by(num_rows_for_target_chunk, context->transaction_id());
+      {
+        auto mvcc_data = target_chunk->get_scoped_mvcc_data_lock();
+        mvcc_data->grow_by(num_rows_for_target_chunk);
+        for (auto chunk_offset = _target_chunk_ranges.back().begin_chunk_offset;
+             chunk_offset < _target_chunk_ranges.back().end_chunk_offset; ++chunk_offset) {
+          // This row is still invisible to everyone thanks to the begin set to infinity. We do this manually rather
+          // than passing context->transaction_id() because of performance reasons. No need for synchronization
+          // as long as we have a fence at the end.
+          mvcc_data->tids[chunk_offset].store(context->transaction_id(), std::memory_order_relaxed);
+        }
+        std::atomic_thread_fence(std::memory_order_seq_cst);
+      }
 
       // Grow data Segments.
       // Do so in REVERSE column order so that the resize of `Chunk::_segments.front()` happens last. It is this last
--- a/src/lib/operators/validate.hpp
+++ b/src/lib/operators/validate.hpp
@@ -23,7 +23,7 @@ class Validate : public AbstractReadOnlyOperator {
   const std::string name() const override;
 
   // MVCC evaluation logic is exposed so that JitValidate can also use it
-  static bool is_row_visible(CommitID our_tid, CommitID snapshot_commit_id, const TransactionID row_tid,
+  static bool is_row_visible(TransactionID our_tid, CommitID snapshot_commit_id, const TransactionID row_tid,
                              const CommitID begin_cid, const CommitID end_cid);
 
  protected:
--- a/src/lib/storage/chunk.cpp
+++ b/src/lib/storage/chunk.cpp
@@ -43,7 +43,7 @@ void Chunk::append(const std::vector<AllTypeVariant>& values) {
   DebugAssert(is_mutable(), "Can't append to immutable Chunk");
 
   // Do this first to ensure that the first thing to exist in a row is the MVCC data.
-  if (has_mvcc_data()) get_scoped_mvcc_data_lock()->grow_by(1u, TransactionID{0});
+  if (has_mvcc_data()) get_scoped_mvcc_data_lock()->grow_by(1u);
 
   // The added values, i.e., a new row, must have the same number of attributes as the table.
   DebugAssert((_segments.size() == values.size()),
--- a/src/lib/storage/chunk.cpp
+++ b/src/lib/storage/chunk.cpp
@@ -43,7 +43,7 @@ void Chunk::append(const std::vector<AllTypeVariant>& values) {
   DebugAssert(is_mutable(), "Can't append to immutable Chunk");
 
   // Do this first to ensure that the first thing to exist in a row is the MVCC data.
-  if (has_mvcc_data()) get_scoped_mvcc_data_lock()->grow_by(1u, CommitID{0});
+  if (has_mvcc_data()) get_scoped_mvcc_data_lock()->grow_by(1u, INVALID_TRANSACTION_ID, CommitID{0});
 
   // The added values, i.e., a new row, must have the same number of attributes as the table.
   DebugAssert((_segments.size() == values.size()),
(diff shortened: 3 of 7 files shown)


src/lib/storage/chunk.cpp
```diff
@@ -43,7 +43,7 @@ void Chunk::append(const std::vector<AllTypeVariant>& values) {
   DebugAssert(is_mutable(), "Can't append to immutable Chunk");
 
   // Do this first to ensure that the first thing to exist in a row is the MVCC data.
-  if (has_mvcc_data()) get_scoped_mvcc_data_lock()->grow_by(1u, TransactionID{0});
+  if (has_mvcc_data()) get_scoped_mvcc_data_lock()->grow_by(1u, true);
```
mrzzzrm: `MvccData::GrowMode::VisibleForAll/HiddenForAll`?

src/lib/operators/insert.cpp
```diff
@@ -128,7 +128,18 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
       // Grow MVCC vectors and mark new (but still empty) rows as being under modification by current transaction.
       // Do so before resizing the Segments, because the resize of `Chunk::_segments.front()` is what releases the
       // new row count.
-      target_chunk->get_scoped_mvcc_data_lock()->grow_by(num_rows_for_target_chunk, context->transaction_id());
+      {
```
mrzzzrm: This should be done outside of the whole-table lock then, right?

src/lib/storage/chunk.cpp
mrks: I get [the original idea](https://softwareengineering.stackexchange.com/questions/323554/are-all-boolean-arguments-flag-arguments-and-thus-a-code-smell) that bools that change the control flow might be a code flow, but this hate for bool values is going way too far. That is a single, simple parameter that does not significantly change any control flow. How is that different from `1u`?

src/lib/operators/validate.hpp
```diff
@@ -23,7 +23,7 @@ class Validate : public AbstractReadOnlyOperator {
   const std::string name() const override;
 
   // MVCC evaluation logic is exposed so that JitValidate can also use it
-  static bool is_row_visible(CommitID our_tid, CommitID snapshot_commit_id, const TransactionID row_tid,
+  static bool is_row_visible(TransactionID our_tid, CommitID snapshot_commit_id, const TransactionID row_tid,
```
mrzzzrm: STRONG_TYPEDEF for TransactionID?

mrks: I think that didn't work because of the atomic.

src/lib/operators/insert.cpp
mrks: Can't. If we shrink the vector, `mvcc_data->tids` might become invalid.

I'm thinking more and more that this whole locking thing is quite dangerous (see #1602).

src/lib/storage/chunk.cpp
mrzzzrm: grow_by(1) is intuitive ("add one element at the end.") - grow_by(1, false) could mean anything. I think it's similar for many places where I'd rather see a named bool.

libhyrise is 42k lines. A dozen named bools are imo not excessive. I am not being pedantic when I say I cannot read `grow_by(1u, true);` - I really cannot.  :(

mrzzzrm: In the worst case - specific to this situation - one might  read "grow_by(1, false)" as "add one false at the end"

mrks: Or bring back the CommitID parameter that was previously there? If not, I'd even split it into two methods, just to avoid yet another parameter. I'm not exaggerating when I'm saying that having to look up the name of those enums while writing code is one of the top things that annoy me the most.

mrzzzrm: another method would be fine. 
(Let's blame c++ for not having named parameters)

mrks: That doesn't work for the constructor. :/
I'll revert to the CommitID that was there previously.

mrzzzrm: > I'm not exaggerating when I'm saying that having to look up the name of those enums while writing code is one of the top things that annoy me the most.

I understand that will be the case if you know the function well, maybe even implemented it. I cannot imagine it would be like that if someone else wrote the function - or if someone else used a function you wrote.

src/lib/storage/mvcc_data.cpp
```diff
@@ -26,10 +26,10 @@ void MvccData::shrink() {
   end_cids.shrink_to_fit();
 }
 
-void MvccData::grow_by(size_t delta, TransactionID transaction_id) {
+void MvccData::grow_by(size_t delta, CommitID begin_commit_id) {
   _size += delta;
-  tids.grow_to_at_least(_size, transaction_id);
-  begin_cids.grow_to_at_least(_size, MAX_COMMIT_ID);
+  tids.grow_to_at_least(_size);
```
mrzzzrm: I don't get why this would be faster than `tids.grow_to_at_least(_size, TransactionID{0]);`....

mrzzzrm:
Just curious, where does the performance increase observed in the TPC-H come from? Did Validate get faster? Where do the, e.g., `+37%  ` for TPC-H 5 come from? What benchmark configuration was that?

mrks:
> Just curious, where does the performance increase observed in the TPC-H come from? Did Validate get faster? Where do the, e.g., +37% for TPC-H 5 come from? What benchmark configuration was that?

Even without `-mvcc`, the tables do have MVCC data. While that is never accessed, it is forwarded as part of the projection. That might be optimized in the future (#1602).

src/lib/storage/mvcc_data.cpp
mrks: [Here](https://github.com/01org/tbb/blob/a51a90bc609bb73db8ea13841b5cf7aa4344d4a9/include/tbb/concurrent_vector.h#L780), either `initialize_array` or `initialize_array_by` get called. Next, `init()` is called [with or without an argument](https://github.com/01org/tbb/blob/2ace525889b0c3de9c90da943fac9259220ef35f/include/tbb/concurrent_vector.h#L1110). Without an argument, it's a default initialization, with an argument, it's a copy construction.

src/lib/storage/table.cpp
```diff
@@ -165,7 +165,7 @@ void Table::append_chunk(const Segments& segments, const std::optional<Polymorph
   std::shared_ptr<MvccData> mvcc_data;
 
   if (_use_mvcc == UseMvcc::Yes) {
-    mvcc_data = std::make_shared<MvccData>(chunk_size);
+    mvcc_data = std::make_shared<MvccData>(chunk_size, CommitID{0});
```
mrzzzrm: would this ever be anything else than zero? If not remove the second constructor argument?

src/lib/storage/mvcc_data.cpp
mrks: Mmh, that might mean that the TIDs are uninitialized though (and were before). I tried moving the non-synchronized code here - let's see what the benchmarks say.

src/lib/storage/table.cpp
mrks: Rather make it a bit more explicit. Added a comment.

src/lib/storage/mvcc_data.cpp
mrks: Nevermind. Of course, `T()` is zero-initialized.

mrks: Not so sure anymore. `std::atomic` leaves the default-constructed object in an invalid state. With synchronization deactivated, initialization is not noticeable, so I rather set the tids manually in every case.

mrks:
Just for safety, I added non-atomic initialization to all calls of grow_by. Performance doesn't care too much and I can sleep better at night:

```
+-----------+----------------+------+----------------+------+--------+---------------------------------+
| Benchmark | prev. iter/s   | runs | new iter/s     | runs | change | p-value (significant if <0.001) |
+-----------+----------------+------+----------------+------+--------+---------------------------------+
| TPC-H 1   | 0.527172207832 | 32   | 0.526437997818 | 32   | -0%    |                          0.8627 |
| TPC-H 2   | 8.29684925079  | 498  | 8.25764942169  | 496  | -0%    |                          0.0472 |
| TPC-H 3   | 6.57278633118  | 395  | 6.58132410049  | 395  | +0%    |                          0.4691 |
| TPC-H 4   | 2.2281563282   | 134  | 2.27249026299  | 137  | +2%    |                          0.0000 |
| TPC-H 5   | 4.42444133759  | 266  | 4.18584394455  | 252  | -5%    |                          0.0000 |
| TPC-H 6   | 29.3826828003  | 1763 | 29.4725322723  | 1769 | +0%    |                          0.0004 |
| TPC-H 7   | 1.35758900642  | 82   | 1.32045865059  | 80   | -3%    |                          0.0000 |
| TPC-H 8   | 5.28122472763  | 317  | 5.07596683502  | 305  | -4%    |                          0.0000 |
| TPC-H 9   | 1.55264019966  | 94   | 1.58549284935  | 96   | +2%    |                          0.0000 |
| TPC-H 10  | 2.75398254395  | 166  | 2.70208239555  | 163  | -2%    |                          0.0000 |
| TPC-H 11  | 31.9011650085  | 1915 | 30.9003391266  | 1855 | -3%    |                          0.0000 |
| TPC-H 12  | 7.66930103302  | 461  | 7.78732919693  | 468  | +2%    |                          0.0000 |
| TPC-H 13  | 2.23531842232  | 135  | 2.23348283768  | 135  | -0%    |                          0.9046 |
| TPC-H 14  | 24.4641456604  | 1468 | 24.3536815643  | 1462 | -0%    |                          0.0003 |
| TPC-H 15  | 19.4964313507  | 1170 | 19.6669635773  | 1181 | +1%    |                          0.0000 |
| TPC-H 16  | 7.7613568306   | 466  | 7.65702676773  | 460  | -1%    |                          0.0000 |
| TPC-H 17  | 1.44373655319  | 87   | 1.40918648243  | 85   | -2%    |                          0.0000 |
| TPC-H 18  | 2.02401041985  | 122  | 1.96566224098  | 119  | -3%    |                          0.0002 |
| TPC-H 19  | 5.20063400269  | 313  | 5.24825954437  | 315  | +1%    |                          0.0096 |
| TPC-H 20  | 2.03985309601  | 123  | 2.0331120491   | 122  | -0%    |                          0.3275 |
| TPC-H 21  | 0.429752975702 | 26   | 0.425328433514 | 26   | -1%    |                          0.0994 |
| TPC-H 22  | 12.8247356415  | 770  | 13.0174331665  | 782  | +2%    |                          0.0000 |
| average   |                |      |                |      | -1%    |                                 |
+-----------+----------------+------+----------------+------+--------+---------------------------------+
```
 
This should be ready for a final review.

PR #1604, opened 2019-04-15
Cmake: add link to resources

### LINKED ISSUE #1583: cmake: symlink resources into build folder (fixed in 1604)
Having to `cd ..; ./build/hyriseTest` is annoying. Sometimes, I do a `ln -s ../resources` so that I can execute the test in the build folder. Should we make that the default and get cmake to automatically create that symlink?

mrks:
fixes #1583, makes it possible to call `./hyriseTest` without changing the directory.

Tested with ninja and OS X as well as make+ninja and Ubuntu.

### DIFF
--- a/src/test/CMakeLists.txt
+++ b/src/test/CMakeLists.txt
@@ -302,20 +302,13 @@ include_directories(../../third_party/googletest/googletest/)
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
-# For convenience, to be able to launch, e.g., hyriseTest, directly from the build directory,
-# add a link to the resources/ directory at the root of the build directory)
-add_custom_target(resourceLink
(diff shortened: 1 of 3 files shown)


src/test/CMakeLists.txt
```diff
@@ -302,13 +302,18 @@ include_directories(../../third_party/googletest/googletest/)
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
+add_custom_target(resourceLink
```
mrzzzrm: Doc with `# For convenience, to be able to launch, e.g., hyriseTest, directly from the build directory, add a link to the resources/ directory at the root of the build directory)

mrks: Good point.

d-justen:
an hyrise

PR #1605, opened 2019-04-15
Print file/line info for Fail()

### LINKED ISSUE #1507: Improve Fail() in assert.hpp (fixed in 1605)
As suggested by @mrks in https://github.com/hyrise/hyrise/pull/1462/files#r257565136:
> That was not what I meant. Have a look at assert.hpp. There, `Assert` builds the file/line/msg string and passes it to `Fail`. If we move that part to `Fail`, we would have line information for both asserts and fails. Only `assert.hpp` needs to change.

mrks:
fixes #1507

### DIFF
--- a/src/lib/utils/assert.hpp
+++ b/src/lib/utils/assert.hpp
@@ -42,7 +42,14 @@
 
 namespace opossum {
 
-[[noreturn]] inline void Fail(const std::string& msg) { throw std::logic_error(msg); }
+// We need this indirection so that we can throw exceptions from destructors without the compiler complaining. That is
+// generally forbidden and might lead to std::terminate, but since we don't want to handle most errors anyway,
+// that's fine.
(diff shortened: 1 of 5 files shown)


src/lib/utils/assert.hpp
```diff
@@ -42,18 +42,25 @@
 
 namespace opossum {
 
-[[noreturn]] inline void Fail(const std::string& msg) { throw std::logic_error(msg); }
+// We need this indirection so that we can throw exceptions from destructors without the compiler complaining. That is
+// generally forbidden and might lead to std::terminate, but since we don't want to handle most errors anyway,
+// that's fine.
+[[noreturn]] inline void fail(const std::string& msg) { throw std::logic_error(msg); }
```
mrzzzrm: minor, but can we put this into `namespace detail` so people are less inclined to use fail() directly over Fail()?

mrks: done

PR #1608, opened 2019-04-16
Fix TPC-H for larger scale factors 

### LINKED ISSUE #1574: TPC-H generator freezes with SF=150
> In the function mk_order in third_party/tpch-dbgen/build.c, the following loop never terminates:
> 
> ```
> while (o->custkey % CUST_MORTALITY == 0)
> {
> o->custkey += delta;
> o->custkey = MIN(o->custkey, O_CKEY_MAX);
> delta *= -1;
> }
> ```
> 
> The O_CKEY_MAX macro is defined as `#define O_CKEY_MAX (tdefs[CUST].base * scale)`.
> While there is global variable "scale", in this instance the code expanded from the macro will use the local function parameter "scale" of mk_order, which has type "float".
> 
> Now, `o->custkey` has integral type `DSS_HUGE` (which is "long long"), and thus the MIN macro expands to a ternary operator (? : ) where the second operand is integral and the third operand is float. According to C type promotion rules, this means that `o->custkey` is converted to float, and if it is smaller than `O_CKEY_MAX`, the resulting float value is converted back to long long.
> 
> And this is where the problem comes in. At the time where the endless loop occurs, `o->custkey` happens to have the value 18388620 for me. The loop tries to change that to either 18388619 or 18388621. Now, both values are smaller than `O_CKEY_MAX`. But converting either of those two values to float and then back to long long results in 18388620 again (due to limited precison of the 32-bit IEEE float data type).
> 
> Thus o->custkey stays forever at this value and the loop never terminates.

Here's a proposed patch:

> ```
> diff --git a/third_party/tpch-dbgen/build.c b/third_party/tpch-dbgen/build.c
> index 5c97b04c9..3c5e4a04d 100644
> --- a/third_party/tpch-dbgen/build.c
> +++ b/third_party/tpch-dbgen/build.c
> @@ -184,7 +184,7 @@ mk_order(DSS_HUGE index, order_t * o, long upd_num, float scale)
>  	while (o->custkey % CUST_MORTALITY == 0)
>  	{
>  		o->custkey += delta;
> -		o->custkey = MIN(o->custkey, O_CKEY_MAX);
> +		o->custkey = MIN(o->custkey, (DSS_HUGE) O_CKEY_MAX);
>  		delta *= -1;
>  	}
> ```

My worry is that there are similar float-to-long conversions somewhere in the modified code that have not caused any problems yet. Originally, dbgen only supported integral scale factors. We enabled floats because we couldn't even handle SF1 back then. In the debug build, some queries still need minutes to execute with SF1, but the release build runs the TPC-H tests in hyriseSystemTest in 

Thus, my suggestion is to get rid of non-integral scale factors and skip the TPC-H tests of hyriseSystemTest in the debug build. When we do that, we can generate the data once instead of 22 times, saving us some time there. 

With #1473, this would result in the following execution times for `./build-release/hyriseSystemTest --gtest_filter=TPC*`:

Screen Shot 2019-04-04 at 1 45 48 PM

Currently, this takes 12 seconds, so it would become 16x slower. However, as this is just system test, which is somewhat expected to take longer, I would be willing to live with that.

The alternative would be to work our way through the generator and check all uses of the float for correctness. I am not confident that I would get that right on the first try.

mrzzzrm:
Fix #1574


third_party/tpch-dbgen/tpch_dbgen.c
```diff
@@ -809,4 +812,51 @@ void dbgen_reset_seeds() {
...
+    }
+  }
+  else
+    scale = (long) flt_scale;
+  if (scale > MAX_SCALE)
+  {
+    fprintf (stderr, "%s %5.0f %s\n\t%s\n\n",
+             "NOTE: Data generation for scale factors >",
+             MAX_SCALE,
+             "GB is still in development,",
```
mrks: 100 TB? That should buy us some time...

src/benchmarklib/tpch/tpch_table_generator.cpp
```diff
@@ -13,6 +13,7 @@ extern "C" {
 #include "boost/hana/zip_with.hpp"
 
 #include "benchmark_config.hpp"
+#include "operators/print.hpp"
```
mrks: .

src/bin/tpch_table_generator.cpp
```diff
@@ -3,10 +3,16 @@
 #include "storage/storage_manager.hpp"
 #include "tpch/tpch_table_generator.hpp"
 
-int main() {
+#include "utils/assert.hpp"
```
mrks: Can we generate the CSV here and compare it to that of dbgen? Not as a test, just once?

src/test/tpc/tpch_db_generator_test.cpp
```diff
@@ -23,29 +23,53 @@ TEST(TpchDbGeneratorTest, RowCounts) {
   EXPECT_EQ(table_info_by_name.at("region").table->row_count(), std::floor(5));
 }
 
+TEST(TpchDbGeneratorTest, RowCountsMediumScaleFactor) {
+  /**
+   * Mostly intended to generate coverage and trigger potential leaks in third_party/tpch_dbgen
+   */
+  const auto scale_factor = 0.008f;
```
mrks: Is this really a "medium" scale factor?

src/benchmarklib/tpch/tpch_table_generator.cpp
```diff
@@ -220,10 +221,14 @@ TpchTableGenerator::TpchTableGenerator(float scale_factor, const std::shared_ptr
 std::unordered_map<std::string, BenchmarkTableInfo> TpchTableGenerator::generate() {
   Assert(!_benchmark_config->cache_binary_tables, "Caching binary Tables not supported by TpchTableGenerator, yet");
 
-  const auto customer_count = static_cast<size_t>(tdefs[CUST].base * _scale_factor);
-  const auto order_count = static_cast<size_t>(tdefs[ORDER].base * _scale_factor);
-  const auto part_count = static_cast<size_t>(tdefs[PART].base * _scale_factor);
-  const auto supplier_count = static_cast<size_t>(tdefs[SUPP].base * _scale_factor);
+  // Init tpch_dbgen - it is important this is done before any data structures from tpch_dbgen are read.
+  dbgen_reset_seeds();
+  dbgen_init_scale_factor(_scale_factor);
```
mrks: Do I understand it correctly that we undo the `scale` changes and use the functions closer to the way dbgen intended it?

third_party/tpch-dbgen/tpch_dbgen.h
```diff
@@ -50,4 +50,5 @@ extern distribution vp;
 extern distribution grammar;
 
 void dbgen_reset_seeds();
+void dbgen_init_scale_factor(float scale_factor);
```
mrks: Did you diff our dbgen and the original one? Otherwise, I'd do it and check if anything sticks out.

src/test/tpc/tpch_db_generator_test.cpp
mrzzzrm: Was WIP, is now 1.5GB

src/benchmarklib/tpch/tpch_table_generator.cpp
mrzzzrm: yes

third_party/tpch-dbgen/tpch_dbgen.h
mrzzzrm: I checked it manually, seems fine. Left more `// HYRISE: We changed this` comments

https://gist.github.com/mrzzzrm/82ac2721e87eb5a6645a5482c0ab7f7c

src/test/tpc/tpch_db_generator_test.cpp
```diff
@@ -28,24 +44,32 @@ TEST(TpchDbGeneratorTest, TableContents) {
...
+  EXPECT_TABLE_EQ_ORDERED(table_info_by_name.at("supplier").table, load_table(dir + "supplier.tbl", chunk_size));
+  EXPECT_TABLE_EQ_ORDERED(table_info_by_name.at("partsupp").table, load_table(dir + "partsupp.tbl", chunk_size));
+  EXPECT_TABLE_EQ_ORDERED(table_info_by_name.at("customer").table, load_table(dir + "customer.tbl", chunk_size));
+  EXPECT_TABLE_EQ_ORDERED(table_info_by_name.at("orders").table, load_table(dir + "orders.tbl", chunk_size));
+  EXPECT_TABLE_EQ_ORDERED(table_info_by_name.at("nation").table, load_table(dir + "nation.tbl", chunk_size));
+  EXPECT_TABLE_EQ_ORDERED(table_info_by_name.at("region").table, load_table(dir + "region.tbl", chunk_size));
+
+  // Run generation a second time to make sure no global state (of which tpch_dbgen has plenty :( ) from the first
+  // generation process carried over into the second
+  table_info_by_name = TpchTableGenerator(scale_factor, chunk_size).generate();
```
mrks: Run it with a different sf?

src/bin/tpch_table_generator.cpp
mrks:
```
#!/usr/bin/python3

import csv

for table in ['customer', 'lineitem', 'nation', 'orders', 'part', 'partsupp', 'region', 'supplier']:
    hyrise_filepath = table + '.csv'  
    dbgen_filepath = 'tpch-dbgen/' + table + '.tbl'
    with open(hyrise_filepath) as hyrise_file:
        hyrise_reader = csv.reader(hyrise_file)

        with open(dbgen_filepath) as dbgen_file:  
           dbgen_line = dbgen_file.readline().strip()
           cnt = 1
           
           for hyrise_values in hyrise_reader:
               dbgen_values = dbgen_line.split('|')
               for hyrise_value, dbgen_value in zip(hyrise_values, dbgen_values):
                   if(hyrise_value != dbgen_value and abs(float(hyrise_value) - float(dbgen_value))) > .1:
                       print("Table " + table + ", Line " + str(cnt) + ": Mismatch " + hyrise_value + " and " + dbgen_value)

               dbgen_line = dbgen_file.readline().strip()
               cnt += 1
           assert(not dbgen_line)
    print(table + " looks good")
```

SFs 0.1 and 1 are fine, running 200 over the night.

src/test/tpc/tpch_db_generator_test.cpp
mrzzzrm: done for 0.002 as well. Didn't choose a larger one as we have to ship the reference table with JIT

src/bin/tpch_table_generator.cpp
mrks:
```
customer looks good
lineitem looks good
nation looks good
orders looks good
part looks good
partsupp looks good
region looks good
supplier looks good
```

mrks:
Excluded unused-variable via third_party/CMakeLists.txt:2?

Issue #1189, opened 2018-10-15
Extend CREATE/DROP Table support


mrzzzrm:
#1178 introduces basic CREATE/DROP table support. A number of features is missing.

* [x] SQLParser is missing support for Hyrise's other types, apart from INT, DOUBLE and TEXT (in #1209)
* [x] SQLParser is missing support for NULL/NOT NULL qualifiers (in #1209)
* [x] Create Table from SELECT
* [x] DROP TABLE IF EXISTS
* [x] CREATE TABLE IF NOT EXISTS
* [x] Check for table existence in console before running plan containing CreateTable 
* [ ] UNIQUE / PRIMARY Constraints (same row and multi-column)

mrzzzrm:
@d-justen 
Regarding `Create Table from SELECT` (for completeness, the two syntax are shown [here](https://www.w3schools.com/sql/sql_create_table.asp)):
Let's call the syntax that Hyrise supports "CreateTableFromLayout" (`CREATE TABLE t (a INT, b FLOAT)`) and the new one "CreateTableFromQuery" (`CREATE TABLE t AS SELECT * FROM another_table WHERE a > 5`). I think it should be possible to support both syntaxes via the same `CreateTableNode` and `CreateTable`-operator. 
For CreateTableFromLayout the `CreateTable` operator would receive a Table with no rows and just the requested column layout from its input operator. For CreateTableFromQuery the input would be the query producing the Table. `CreateTable[Node]::column_definitions` would then be obsolete and could be removed. 

For the empty input table of CreateTableFromLayout we'd need a new concept in the LQP, one that we'd need for #1546 as well: Static/ConstantTableNodes - just a node wrapping a static table. For #1546 this would be a table containing the IN list, for CreateTableFromLayout this would be a table with no rows and just the column layout.

@mrks what do you think?

mrks:
To me, it sounds like you are forcing two things that don't have too much in common into one entity.

Trying to predict what the CreateTableNode will have to do in the future, I see indexes, default values and constraints. None of these apply to the `CREATE TABLE ... AS` path.

I'd suggest a StoreTable operator, similar to the GetTable operator, which deals with this new path. Keeping those separate should also make it easier to support `SELECT INTO`. After SQL translations, these should end up going down the same road.

[...]

PR #1613, opened 2019-04-18
Add IF EXISTS support for DROP TABLE and DROP VIEW

d-justen:
As suggested in #1189.

### DIFF
--- a/src/lib/operators/maintenance/drop_table.cpp
+++ b/src/lib/operators/maintenance/drop_table.cpp
@@ -14,6 +14,7 @@ const std::string DropTable::description(DescriptionMode description_mode) const
 }
 
 std::shared_ptr<const Table> DropTable::_on_execute() {
+  // If IF EXISTS is not set and the table is not found, StorageManager throws an exception
   if (!if_exists || StorageManager::get().has_table(table_name)) {
     StorageManager::get().drop_table(table_name);
   }
--- a/src/lib/operators/maintenance/drop_table.cpp
+++ b/src/lib/operators/maintenance/drop_table.cpp
@@ -4,8 +4,8 @@
 
 namespace opossum {
 
-DropTable::DropTable(const std::string& table_name)
-    : AbstractReadOnlyOperator(OperatorType::DropTable), table_name(table_name) {}
+DropTable::DropTable(const std::string& table_name, const bool if_exists)
+    : AbstractReadOnlyOperator(OperatorType::DropTable), table_name(table_name), if_exists(if_exists) {}
 
 const std::string DropTable::name() const { return "Drop Table"; }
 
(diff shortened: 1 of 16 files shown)


src/lib/operators/maintenance/drop_table.cpp
```diff
@@ -14,15 +14,17 @@ const std::string DropTable::description(DescriptionMode description_mode) const
 }
 
 std::shared_ptr<const Table> DropTable::_on_execute() {
-  StorageManager::get().drop_table(table_name);
+  if (!if_exists || StorageManager::get().has_table(table_name)) {
```
mrks: This could use a comment that we rely on the StorageManager to fail if the table does not exist but `IF EXISTS` is not set.

src/lib/operators/maintenance/drop_view.cpp
```diff
@@ -9,21 +9,23 @@
...
     const std::shared_ptr<AbstractOperator>& copied_input_right) const {
-  return std::make_shared<DropView>(_view_name);
+  return std::make_shared<DropView>(view_name, if_exists);
 }
 
 void DropView::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeVariant>& parameters) {}
 
 std::shared_ptr<const Table> DropView::_on_execute() {
-  StorageManager::get().drop_view(_view_name);
+  if (!if_exists || StorageManager::get().has_view(view_name)) {
```
mrks: same

src/test/operators/maintenance/drop_view_test.cpp
```diff
@@ -44,15 +44,32 @@ TEST_F(DropViewTest, DeepCopy) {
...
+  auto dv = std::make_shared<DropView>("view_name", false);
   dv->execute();
 
   EXPECT_EQ(dv->get_output()->row_count(), 0u) << "DropView returned non-empty table";
 
   EXPECT_FALSE(StorageManager::get().has_view("view_name")) << "View was not removed";
 }
 
+TEST_F(DropViewTest, ExecuteWithIfExists) {
+  EXPECT_TRUE(StorageManager::get().has_view("view_name")) << "View not found";
```
mrks: This looks like you are testing if the test was correctly setup. That is a bit confusing. Only test the things that the test is supposed to check.

src/test/operators/maintenance/drop_view_test.cpp
```diff
@@ -44,15 +44,28 @@ TEST_F(DropViewTest, DeepCopy) {
...
   EXPECT_EQ(dv->get_output()->row_count(), 0u) << "DropView returned non-empty table";
 
   EXPECT_FALSE(StorageManager::get().has_view("view_name")) << "View was not removed";
 }
 
+TEST_F(DropViewTest, ExecuteWithIfExists) {
+  auto dv_1 = std::make_shared<DropView>("view_name", true);
+  dv_1->execute();
+
+  EXPECT_EQ(dv_1->get_output()->row_count(), 0u) << "DropView returned non-empty table";
```
mrzzzrm: Admirable, but we usually do not supply additional messages in tests - `dv_1->get_output()->row_count() != 0` already tells us what the message says. Remember - comments and messages are essentially untested code and therefore error prone and better avoided if possible ;)

d-justen: I just wanted to keep consistency within in this test, but you're right of course. I'll just delete these additional messages

PR #1615, opened 2019-04-19
Make Logical and Physical Plan caches non-global

### LINKED ISSUE #1402: SQL{Logical,Physical}PlanCache should not be global (fixed in #1615)
Currently, the query caches are always used by the SQLPipelineStatement, the only option to influence what cache the SQLPipeline uses is to manipulate the global cache. The caches should be normal parameters of the SQLPipeline, just as the Optimizer etc are.

In benchmarks, e.g., I want to be able to disable query caching in order to be able to benchmark the optimizer as well. There should be a CLI option for this.

mrzzzrm:
Fix #1402 

Makes the Logical and the Physical plan caches optional. If their presence is desired for SQL execution they have to be manually added to the `SQLPipelineBuilder`:

```c++
auto pipeline_builder =  SQLPipelineBuilder{sql}.
  with_lqp_cache(_lqp_cache).
  with_pqp_cache(_pqp_cache).
  with_mvcc(_config.use_mvcc);
```


src/benchmarklib/benchmark_runner.cpp
```diff
@@ -483,7 +485,10 @@ void BenchmarkRunner::_create_report(std::ostream& stream) const {
 std::shared_ptr<SQLPipeline> BenchmarkRunner::_build_sql_pipeline(const QueryID query_id) const {
   // Create an SQLPipeline for this query
   const auto sql = _query_generator->build_query(query_id);
-  auto pipeline_builder = SQLPipelineBuilder{sql}.with_mvcc(_config.use_mvcc);
+  auto pipeline_builder = SQLPipelineBuilder{sql}
+                              .with_sql_lqp_cache(_sql_lqp_cache)
```
mrks: I don't like having to pass the cache in if we want to use it. The pipeline is supposed to be a single, one-stop solution for executing SQL queries. We shouldn't make it any more difficult than absolutely necessary. Even if the original issue called for non-global caches, I'd suggest to have a default cache in the pipeline (or the pipeline builder) and being able to overwrite/disable it using `with_sql_lqp_cache`.

mrzzzrm: And what cache would that be? A global singleton?

mrks: No, just a default cache stored as a static member in the pipeline (builder).

Maybe we should decide what we want to fix: Do we want the possibility of using different (or no) caches or do we want to get rid of the default cache?

For me, the first part is more important. Making the benchmark runner (and others) aware of the cache here means that we will have to touch multiple places when we want to disable the cache by default.

mrzzzrm: > Maybe we should decide what we want to fix: Do we want the possibility of using different (or no) caches or do we want to get rid of the default cache?

Both, see below.

> No, just a default cache stored as a static member in the pipeline (builder).

That *is* a global by definition, right? (I don't see how you could have a sensible default cache without it being a global.)

Caching can have unexpected effects like: the Optimizer is not invoked (e.g., chunk pruning info is not updated, the rule you're programming/testing not getting invoked), the SQLTranslator is not invoked, the LQPTranslator is not invoked, caching costs. I'd say caching should require a conscious decision on the libhyrise-user side. 

> Making the benchmark runner (and others) aware of the cache here means that we will have to touch multiple places when we want to disable the cache by default.

As shown by the diff of this PR, every explicit assignment of a cache was done for a place where an explicit cache was used anyway, except for the BenchmarkRunner (where, with the introduction of TPCC, we have to make a cache/no-cache decision in the future) and the Console. I don't see how this diff would be significantly shorter with a default cache. But it's likely I misunderstand your point,

mrzzzrm: Actually, the global cache is a pain: Every test using SQL then has to make sure this cache is empty before doing anything. There were a few seemingly random places where `SQLPhysicalPlanCache::get().clear()` was called that I could thankfully remove.

src/benchmark/operators/sql_benchmark.cpp
```diff
@@ -61,12 +58,13 @@ class SQLBenchmark : public MicroBenchmarkBasicFixture {
...
+    const auto sql_pqp_cache = std::make_shared<SQLPhysicalPlanCache>();
+
     // Enable query plan cache.
-    SQLPhysicalPlanCache::get().clear();
-    SQLPhysicalPlanCache::get().resize(16);
+    sql_pqp_cache->resize(16);
 
     for (auto _ : state) {
-      auto pipeline_statement = SQLPipelineBuilder{query}.create_pipeline_statement();
+      auto pipeline_statement = SQLPipelineBuilder{query}.with_sql_pqp_cache(sql_pqp_cache).create_pipeline_statement();
```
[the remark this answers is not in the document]
mrzzzrm: mmh, maybe I envisioned a LQP->PQP cache or something... I won't argue on this one, make a call to your preference

src/benchmarklib/benchmark_runner.cpp
mrks: Let's discuss this on Wednesday.

mrzzzrm:
@mrks we agreed that you are going to make your mind and make the call on this, right?

mrzzzrm:
@mrks we agreed that you are going to make your mind and make the call on this, right?

mrks:
Yes. If that was a rhetorical question: Hoping to get to it in the next days.

mrzzzrm:
Nope, not rhetorical, just wanted to check whether I remembered right.

mrks:
I'm still not really convinced, and have some more concerns about how we can invalidate caches when things like indexes being added, chunks being compressed, or statistics being updated happen. However, we would only have to change some tens of lines, so I don't think that it makes sense to pursue this discussion any further. Will change the `sql_lqp_cache` name and that's it.

Bouncner:
Could you add a short description?

Will it change the way Jan and me access & parse query plan caches?

mrzzzrm:
Added I comment. 
I do not know how you and Jan access and parse the caches but nothing has fundamentally changed. You just do `lqp_cache->something` instead of `LQPCache::get()->something`

mrks:
Wait a second - won't this make things like automatic index creation impossible? How is a plugin supposed to access the cache if it is hidden somewhere in the console, the benchmark runner, or the server?

mrzzzrm:
Because at some point there will be the concept of a `DatabaseInstance`s that wrap the StorageManager and Caches and will be passed to the Plugin or something. 

There will almost always be ways around globals that usually make things better in general. Probably the only real thing that cannot be made non-global is the PluginManager since it is global by nature, I suppose: A symbol cannot exist twice in a running process(?).

Everything else can either be made completely non-global or at least have an explicitly controlled lifetime in main(). Imho, everything else is just lazy and imo is the founding seed for things like https://github.com/hyrise/hyrise/issues/1635 happening over and over again with fixes being, no offence intended, "magic calls to ensure order, probably, trust me".

mrks:
Having to pass around all different types of instances is a huge pain. Singletons / globals have their problems, yes, but the indiscriminate bad rep that they get is undeserved. If the alternative is passing around a reference to the transaction manager (of which there will be only, ever, in every case, exactly one), you end up with bloated interfaces on every step of your path. I've been there.

I am not sure where @Bouncner and @Bensk1 access the query cache. From what I remember, it is some place that wouldn't have any reason to know about the cache if they wouldn't be *doing research* there.

mrzzzrm:
Let's have another f2f discussion, I suppose...

mrks:
I'm still willing to approve this, so let's wait for their answer.

Bouncner:
Cannot talk for @Bensk1, but I assume he does the same. I issue various queries from the playground and call the BenchmarkRunner from the playground. Afterwards and in between I analyze the plan cache by accessing both the LQP as well as the PQP cache.

I haven't really followed the complete discussion here, but that this would no longer be possible with the change to the global cache, right?

mrzzzrm:
You'd need to drop in `BenchmarkRunner::lqp/pqp_cache` instead of `SQLLogical/PhysicalPlanCache::get(), so no hindrance here

mrks:
BenchmarkRunner is currently not static, so instead of the "where do I get my cache instance from" problem, it becomes a "where do I get the benchmark runner from" problem.

mrzzzrm:
If they construct the BenchmarkRunner in the playground then this shouldn't be a problem.

mrks:
Makes sense. I guess that should work.

Bouncner:
Well, that gives me the cache of the benchmark runner, but not the ones I manually fired, right? I don't have running code right but, so I cannot really show you any code. But what I would like to do is not just simply use the benchmark runner but also to fire queries in parallel. Would that be possible? Or do I have to merge caches than?

I don't see how I can get a "single source of truth" of the current workload. After all, the means I'd like to implement won't create the workloads but rather just react to whatever happens within Hyrise.

mrzzzrm:
If you run queries from outside the BenchmarkRunner and want them to use the cache **of** the BenchmarkRunner, then you have to create the pipeline of those queries **with** the caches of the BenchmarkRunner: `SQLPipelineBuilder{"SELECT...."}.with_lqp_cache(my_benchmark_runner->lqp_cache).[...].create_pipeline()`

Bensk1:
Currently, I follow the same approach Martin takes. I see the problems with globals, Moritz, but I also agree with Markus that Singletons make sense for certain problems. 

How would a plugin that has nothing to do with actually executing the queries access the cache(s)? Neither does it create a BenchmarkRunner nor use the SQLPipelineBuilder.

mrzzzrm:
> How would a plugin that has nothing to do with actually executing the queries access the cache(s)? Neither does it create a BenchmarkRunner nor use the SQLPipelineBuilder.

I believe this is outside of the scope of this PR. One possibility would be hooks, e.g., for SQLPipelines being created, queries finishing, tables being loaded, encoding being changes, etc, etc, etc, that Plugins can listen to. Or you tie the PluginManager to a "HyriseInstance". 
There are options and we need a proper solution to the question "but, how would a plugin know about <thing>". Making everything plugins might care about global is not the solution.

mrzzzrm:
And I am not on a matter-of-principle crusade against globals here - I have admitted their usefulness, e.g., with regards to the PluginManager. I am however working against their documented (by various bugs, performance regression as seen in TPCC and "clear-everything-global-just-in-case"-tests) downsides.

Bouncner:
To come forward here: @mrzzzrm , are you around this week? Can we have a quick chat about this PR?

Bensk1:
I talked with @mrzzzrm last week. We both agreed that the current solution is not optimal but that the proposed fix will force us to apply small hacks for our experiments. A proper solution would be what Moritz described above with the database instance. I will speak to Markus in the next days about it and when you two have your quick chat we can probably finally merge or close that.

mrks:
Summary of discussion: `SQLPipeline::(set|get)DefaultCache` which is reset in `~BaseTest`.

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -32,6 +33,9 @@ BenchmarkRunner::BenchmarkRunner(const BenchmarkConfig& config,
       _benchmark_item_runner(std::move(benchmark_item_runner)),
       _table_generator(std::move(table_generator)),
       _context(context) {
+  SQLPipelineBuilder::default_pqp_cache = std::make_shared<SQLPhysicalPlanCache>();
+  SQLPipelineBuilder::default_lqp_cache = std::make_shared<SQLLogicalPlanCache>();
```
mrks: I'll need to move this once TPC-C comes in, but for the current master it's fine

src/lib/sql/sql_pipeline_builder.hpp
```diff
@@ -35,12 +36,21 @@ class Optimizer;
  */
 class SQLPipelineBuilder final {
  public:
+  // Plan caches used if the respective cache is nullptr in this builder. Both default caches can be nullptr themselves.
```
mrks: "if the respective cache is nullptr"? I think you mean the same, but "if with_(pqp|lqp)_cache are not used" would be clearer

mrzzzrm: I changed it, but that's slightly less correct because if those are called with nullptrs the default cache is still used. So for a builder where you explicitly stated that you don't want to use caches, you still get a cache. Globals are great!

Do you want me to make this `std::optional<std::shared_ptr<SQLPhysicalPlanCache>>` to work around this?

mrks: Why do we need an optional to work around this? In the constructor of the builder, set `_pqp_cache(default_pqp_cache)`. If `with_pqp_cache(nullptr)` is called, that is overwritten.

mrzzzrm: ah, clever.

mrzzzrm:
@mrks that looks like a macmini hickup? or is that just jenkins?

mrks:
```
fatal: unable to access 'https://github.com/lz4/lz4/': Could not resolve host: github.com
```

Same as with the "normal" Jenkins. -.-

src/benchmark/operators/sql_benchmark.cpp
```diff
@@ -61,12 +58,13 @@ class SQLBenchmark : public MicroBenchmarkBasicFixture {
 
   // Run a benchmark that plans the query operator with the given query with enabled query plan caching.
   void BM_QueryPlanCache(benchmark::State& state) {
+    const auto pqp_cache = std::make_shared<SQLPhysicalPlanCache>();
+
     // Enable query plan cache.
```
Bensk1: Does this comment make sense?

mrks: Especially with the line below:

```
     pqp_cache->resize(16); 
```

mrzzzrm:
@mrks one thing that makes the idea of "just" resetting the default caches/globals in general in the BaseTest destructor a bit unclean is the fact that many tests do not actually inherit from BaseTest:

image

If one of these tests touches global state, we currently rely on another subsequent test, which inherits from BaseTest, to reset it. Not sure whether that happens, but it's unclean. #makeglobalslocalsagain

mrks:
This will remain a problem even with #1677 as these tests also don't reset the StorageManager, PluginManager, etc.

They'll have to inherit from `BaseTest` and we might want to introduce a linter rule that prohibits directly inheriting from gtest.

Bouncner:
I don't know if you guys will be noticed here. But now that I merged the master in one of our plugins: how can I obtain the plan cache of the benchmark runner? I am just seeing setting it a default one but no way to access it.

mrks:
> I am now accessing the plan cache the following way

Yepp, should work, doesn't it?

> Also, there are TODOs referencing the changes with the global Hyrise singleton.

I don't see any advantage in moving the default caches there. A disadvantage would be the risk of overburdening the Hyrise class with everything that didn't find its own, proper place. If you ask me, I'd just remove the TODO.

Bouncner:
Removed the comment with #1932.

PR #1617, opened 2019-04-20
Use indexes for SQLite verification of join order benchmark

mrks:
part of #1409

### DIFF
--- a/src/benchmark/join_order_benchmark.cpp
+++ b/src/benchmark/join_order_benchmark.cpp
@@ -119,13 +119,14 @@ int main(int argc, char* argv[]) {
       BenchmarkRunner{*benchmark_config, std::move(query_generator), std::move(table_generator), context};
 
   if (benchmark_config->verify) {
-    // Add indexes to SQLite. This is a hack until we support CREATE INDEX ourselves and pass that on to SQLite
+    // Add indexes to SQLite. This is a hack until we support CREATE INDEX ourselves and pass that on to SQLite.
+    // Without this, SQLite would never finish.
 
     std::cout << "- Adding indexes to SQLite" << std::endl;
     Timer timer;
 
     // SQLite does not support adding primary keys, so we rename the table, create an empty one from the provided
-    // schema and copy the data. Without this, SQLite would never finish.
+    // schema and copy the data.
     for (const auto& table_name : StorageManager::get().table_names()) {
       benchmark_runner.sqlite_wrapper->raw_execute_query(std::string{"ALTER TABLE "} + table_name +  // NOLINT
                                                          " RENAME TO " + table_name + "_unindexed");
--- a/src/benchmark/join_order_benchmark.cpp
+++ b/src/benchmark/join_order_benchmark.cpp
@@ -17,6 +17,7 @@
 #include "utils/filesystem.hpp"
 #include "utils/load_table.hpp"
 #include "utils/performance_warning.hpp"
+#include "utils/sqlite_wrapper.hpp"
 
 /**
  * The Join Order Benchmark was introduced by Leis et al. "How good are query optimizers, really?".
@@ -116,16 +117,25 @@ int main(int argc, char* argv[]) {
   auto benchmark_runner =
       BenchmarkRunner{*benchmark_config, std::move(query_generator), std::move(table_generator), context};
 
-  // Add indexes to SQLite. This is a hack until we support CREATE INDEX ourselves and pass that on in the SQLiteWrapper
-  for (const auto& table_name : StorageManager::get().table_names()) {
-    _sqlite_wrapper._sqlite_wrapper->execute_query(std::string{"ALTER TABLE "} + table_name + " ADD PRIMARY KEY (id)");
+  std::cout << "- Adding indexes to SQLite... " << std::flush;
+
+  if (benchmark_config->verify) {
+    // Add indexes to SQLite. This is a hack until we support CREATE INDEX ourselves and pass that on in the SQLiteWrapper
+    for (const auto& table_name : StorageManager::get().table_names()) {
+      // SQLite does not support adding primary keys, so we just add a regular index
+      std::cout << (std::string{"CREATE INDEX "} + table_name + "_primary ON " + table_name + "(id)") << std::endl;;
+      benchmark_runner.sqlite_wrapper->raw_execute_query(std::string{"CREATE INDEX "} + table_name + "_primary ON " + table_name + "(id)");
+    }
+
+    std::ifstream foreign_key_file(std::string{DEFAULT_QUERY_PATH} + "/fkindexes.sql");
+    std::string foreign_key_definition;
+    while (getline(foreign_key_file, foreign_key_definition)) {
+      std::cout << (foreign_key_definition) << std::endl;;
+      benchmark_runner.sqlite_wrapper->raw_execute_query(foreign_key_definition);
+    }
   }
 
-  std::ifstream foreign_key_file(std::string{DEFAULT_QUERY_PATH} + "/fkindexes.sql");
-  std::string foreign_key_definition;
-  while (getline(foreign_key_file, foreign_key_definition)) {
-    _sqlite_wrapper._sqlite_wrapper->execute_query(foreign_key_definition);
-  }
+  std::cout << "done." << std::endl;
 
   benchmark_runner.run();
 }
--- a/src/benchmark/join_order_benchmark.cpp
+++ b/src/benchmark/join_order_benchmark.cpp
@@ -1,4 +1,4 @@
-#include <stdlib.h>
+#include <fstream>
 
 #include <boost/algorithm/string.hpp>
 #include <cxxopts.hpp>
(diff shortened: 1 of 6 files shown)


mrks:
Force push as this was erroneously based on the ubuntu 19.04 branch.

src/benchmark/join_order_benchmark.cpp
```diff
@@ -113,5 +114,31 @@ int main(int argc, char* argv[]) {
...
+  auto benchmark_runner =
+      BenchmarkRunner{*benchmark_config, std::move(query_generator), std::move(table_generator), context};
+
+  std::cout << "- Adding indexes to SQLite... " << std::flush;
+
+  if (benchmark_config->verify) {
+    // Add indexes to SQLite. This is a hack until we support CREATE INDEX ourselves and pass that on in the SQLiteWrapper
+    for (const auto& table_name : StorageManager::get().table_names()) {
+      // SQLite does not support adding primary keys, so we just add a regular index
+      std::cout << (std::string{"CREATE INDEX "} + table_name + "_primary ON " + table_name + "(id)") << std::endl;
```
mrzzzrm: `+ --> <<`

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -90,8 +82,15 @@ void BenchmarkRunner::run() {
       pipeline.get_result_table();
     }
   }
+}
+
+BenchmarkRunner::~BenchmarkRunner() {
+  if (CurrentScheduler::is_set()) {
```
mrzzzrm: BenchmarkRunner::run() isn't guaranteed to wait until all queries finished?

src/benchmark/join_order_benchmark.cpp
mrks: Ah. Copy and paste from the actual string for debugging. Do we even want it in here? I'd drop it.

mrks:
```
- Benchmarking Query 10a
- Running query with SQLite (5 s 179 ms).
- Comparing Hyrise and SQLite result tables
[PERF] operator[] used at src/lib/storage/value_segment.cpp:72

- Verification passed (1 rows; 19 µs 6 ns)
  -> Executed 1 times in 5.65266 seconds (0.176908 iter/s)
```

src/benchmarklib/benchmark_runner.cpp
mrks: I think this is more of a general precaution.

mrzzzrm: A general precaution that would potentially hide a possible bug in BenchmarkRunner::run(), right?

mrks: Addressed in #1597

src/benchmark/join_order_benchmark.cpp
```diff
@@ -113,5 +114,27 @@ int main(int argc, char* argv[]) {
...
+    }
+
+    std::ifstream foreign_key_file(std::string{DEFAULT_QUERY_PATH} + "/fkindexes.sql");  // NOLINT
+    std::string foreign_key_definition;
+    while (getline(foreign_key_file, foreign_key_definition)) {
+      benchmark_runner.sqlite_wrapper->raw_execute_query(foreign_key_definition);
+    }
+  }
+
+  std::cout << "done." << std::endl;
```
mrzzzrm: You could consider adding timing info (i.e. how long creating the indexes took) here. Most other setup steps in the benchmarks do the same

src/benchmark/join_order_benchmark.cpp
```diff
@@ -113,5 +114,27 @@ int main(int argc, char* argv[]) {
...
       std::make_unique<FileBasedQueryGenerator>(*benchmark_config, query_path, non_query_file_names, query_subset);
 
-  BenchmarkRunner{*benchmark_config, std::move(query_generator), std::move(table_generator), context}.run();
+  auto benchmark_runner =
+      BenchmarkRunner{*benchmark_config, std::move(query_generator), std::move(table_generator), context};
+
+  std::cout << "- Adding indexes to SQLite... " << std::flush;
+
+  if (benchmark_config->verify) {
+    // Add indexes to SQLite. This is a hack until we support CREATE INDEX ourselves and pass that on to SQLite
```
mrzzzrm: Comment that without this SQLite cannot even finish some queries and that we therefore cannot avoid adding the indices?

src/benchmarklib/benchmark_runner.cpp
mrzzzrm: just realised that this was there before. Alright, I rest my case

src/benchmarklib/benchmark_runner.hpp
```diff
@@ -41,6 +41,9 @@ class BenchmarkRunner {
 
   static nlohmann::json create_context(const BenchmarkConfig& config);
 
+  // If the query execution should be validated, this stores a pointer to the used SQLite instance
+  std::unique_ptr<SQLiteWrapper> sqlite_wrapper;
```
mrzzzrm: any reason this is public (and other members aren't)? I am all for public members, as long as there is consistency in a class.

mrzzzrm:
Just minor stuff. I am happy that this finally works.

mrks:
Not so fast. 13d is stuck since 04:20 while it only took 80 seconds without indexes. @Bouncner - *Indexes proven harmful*.

However, the master is stuck since 05:07 on 16b. I'll let it running until Tuesday or so and see what happens.

I'm tempted to say that sqlite is simply not cut out for the join-order benchmark. My suggestion is to store the log somewhere so that we know which queries to look at, fix those, run it once more, and forget it.

```
$ ./build-release/hyriseBenchmarkJoinOrder --verify
Oh boy. Please confirm that you have a lot of free memory and spare time [yN]: 
```

mrzzzrm:
Can you try running the JOB via the sqlite3 CLI? I seem to recall it worked reasonably well. Maybe the PRIMARY KEY decls from schema.sql are what is missing to make this run fast?

src/benchmarklib/benchmark_runner.hpp
mrks: Because there is no reason to touch them. We should only make members public if they would otherwise have an accessor.

mrks:
Yes, that actually made a difference. Loading the SQLite data now became even more ugly. ;)

This should be ready to go.

src/benchmark/join_order_benchmark.cpp
mrzzzrm: From the diff, the requested comment is still missing. Or do you think it isn't helpful? I think it'd help me...

src/benchmarklib/benchmark_runner.hpp
mrzzzrm: I think I was referring to sqlite_wrapper seemingly not being accessed from anyone else but the BenchmarkRunner, but this is a minor point, feel free to resolve.

src/benchmark/join_order_benchmark.cpp
mrks: I had "Without this, SQLite would never finish." in there, but a couple of lines below. Moved it up.

src/benchmarklib/benchmark_runner.hpp
mrks: I can't follow. Who can't access it?

mrzzzrm: Nevermind, I just saw that the index creation happens outside of the BenchmarkRunner, in main(). (I - erroneously - suspected the sqlite_wrapper was made public without reason.)

mrks:
Takes a bit more than an hour now on nemea.

mrzzzrm:
You could, if you feel curious, remove all the `MIN()` aggregations from the JOB-queries - that might give us better perspective on the errors.

src/benchmark/join_order_benchmark.cpp
```diff
@@ -113,5 +115,48 @@ int main(int argc, char* argv[]) {
...
+
+    // SQLite does not support adding primary keys, so we rename the table, create an empty one from the provided
+    // schema and copy the data.
+    for (const auto& table_name : StorageManager::get().table_names()) {
+      benchmark_runner.sqlite_wrapper->raw_execute_query(std::string{"ALTER TABLE "} + table_name +  // NOLINT
+                                                         " RENAME TO " + table_name + "_unindexed");
+    }
+
+    // Recreate tables from schema.sql
+    std::ifstream schema_file(std::string{DEFAULT_QUERY_PATH} + "/schema.sql");  // NOLINT
```
mrzzzrm: why not use `query_path` path here - `--query_path` is an option and should probably be respected. (useful, e.g., for running the JOB from the build dir)

mrks:
> You could, if you feel curious, remove all the MIN() aggregations from the JOB-queries - that might give us better perspective on the errors.

That would be a thing for #1409

src/benchmark/join_order_benchmark.cpp
```diff
@@ -113,5 +115,48 @@ int main(int argc, char* argv[]) {
...
+    std::string schema_sql((std::istreambuf_iterator<char>(schema_file)), std::istreambuf_iterator<char>());
+    benchmark_runner.sqlite_wrapper->raw_execute_query(schema_sql);
+
+    // Add foreign keys
+    std::ifstream foreign_key_file(std::string{DEFAULT_QUERY_PATH} + "/fkindexes.sql");  // NOLINT
```
mrzzzrm: Here, too, maybe use `query_path`

PR #1618, opened 2019-04-21
Enable TSan for NUMA builds

### LINKED ISSUE #1006: Enable TSAN for NUMA builds
Currently, TSAN reports a bunch of possible data races for the `PausableLoopThread` and the `MigrationPreparationTask`. As soon as the numa support has been reworked, we should activate thread sanitisation checks for numa builds.

mrks:
fixes #1006 as the problematic `MigrationPreparationTask` is long gone. If `PausableLoopThread` still causes issues, those need to be addressed, as it is used in #1493.


mrks:
Hoping to get the issue list down to 80 this weekend. ;)

mrks:
Shame. As soon as I have to change more than one thing and there is no compiler whining, I fail. :(

mrzzzrm:
was just about to throw my torch ;)

mrzzzrm:
> Hoping to get the issue list down to 80 this weekend. ;)

Merging #1550 would make us CRUSH that mark.

PR #1626, opened 2019-04-22
Clarify INVALID_VALUE_ID

### LINKED ISSUE #1471: ValueID inconsistencies and problems (fixed in #1626)
When encoding a ValueSegment with 2^32 unique values we should get into trouble because the resulting dictionary would have a size of 2^32 and the DictionaryEncoder does `const auto max_value = dictionary.size() + 1u;` which is problematic since `max_value` is passed to `compress_vector()` which expects a `uint32_t`. In addition, we would have no way to indicate NULL or `INVALID_VALUE_ID` without overflow.

In addition, `INVALID_VALUE_ID` (i.e. `constexpr ValueID INVALID_VALUE_ID{std::numeric_limits<ValueID::base_type>::max()};`) is used to indicate that a certain value is not part of the dictionary and returned by `lower/upper_bound()` when no corresponding `ValueID` can be found. The size of the dictionary (i.e. the largest used ValueID for this particular segment + 1) is used to indicate NULL. Not only could we see cases where NULL and `INVALID_VALUE_ID` are indicated by the same number, but also does this seem inconsistent in general and could lead to some confusion. Especially since it should not break anything because of NULL semantics.

mrks:
fixes #1471 if everyone is happy

### DIFF
--- a/src/lib/storage/dictionary_segment.cpp
+++ b/src/lib/storage/dictionary_segment.cpp
@@ -23,8 +23,8 @@ DictionarySegment<T>::DictionarySegment(const std::shared_ptr<const pmr_vector<T
   // ValueID::base_type (2^32 - 1), is needed to represent "value not found" in calls to lower_bound/upper_bound.
   // For a DictionarySegment of the max size Chunk::MAX_SIZE, those two values overlap.
 
-  DebugAssert(static_cast<ValueID>(_dictionary->size()) == _null_value_id, "Invalid NULL value id");
   Assert(_dictionary->size() < std::numeric_limits<ValueID::base_type>::max(), "Input segment too big");
+  DebugAssert(ValueID{static_cast<uint32_t>(_dictionary->size()}) == _null_value_id, "Invalid NULL value id");
 }
 
 template <typename T>
--- a/src/lib/storage/dictionary_segment.cpp
+++ b/src/lib/storage/dictionary_segment.cpp
@@ -19,8 +19,11 @@ DictionarySegment<T>::DictionarySegment(const std::shared_ptr<const pmr_vector<T
       _attribute_vector{attribute_vector},
       _null_value_id{null_value_id},
       _decompressor{_attribute_vector->create_base_decompressor()} {
-  // We need the last possible value ID to symbolize NULL in the attribute vector and "value not found" in
-  // lower_bound / upper_bound. Because of Chunk::MAX_SIZE, this should never happen anyway.
+  // NULL is represented by _dictionary.size(). INVALID_VALUE_ID, which is the highest possible number in
+  // ValueID::base_type (2^32 - 1), is needed to represent "value not found" in calls to lower_bound/upper_bound.
+  // For a DictionarySegment of the max size Chunk::MAX_SIZE, those two values overlap.
+
+  DebugAssert(_dictionary->size() == _null_value_id), "Invalid NULL value id");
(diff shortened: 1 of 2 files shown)


src/lib/storage/dictionary_segment.cpp
```diff
@@ -18,7 +18,11 @@ DictionarySegment<T>::DictionarySegment(const std::shared_ptr<const pmr_vector<T
       _dictionary{dictionary},
       _attribute_vector{attribute_vector},
       _null_value_id{null_value_id},
-      _decompressor{_attribute_vector->create_base_decompressor()} {}
+      _decompressor{_attribute_vector->create_base_decompressor()} {
+  // We need the last possible value ID to symbolize NULL in the attribute vector and "value not found" in
```
Bensk1: Regarding the representation of NULL: Let's think of a dictionary with 10 values from 1 to 10 again. The data type of ValueID would be `uint8_t`. I understand the last possible ValueID to symbolize NULL to be `std::numeric_limits<uint8_t>::max()` or `256`.

However, the comment below says that NULL is represented with the ValueID of `unique_values_count()` which should be `10`.

mrks: How about this:

> NULL is represented by `_dictionary.size()`. `INVALID_VALUE_ID`, which is the highest possible number in ValueID::base_type (`2^32 -1`) is needed to represent "value not found" in calls to `lower_bound/upper_bound`. For a DictionarySegment of the max size `Chunk::MAX_SIZE`, those two values overlap.

```
Assert(_dictionary->size() == _null_value_id), "Invalid NULL value id");
Assert(_dictionary->size() < std::numeric_limits<ValueID::base_type>::max(), "Input segment too big");
```

src/lib/storage/dictionary_segment.hpp
```diff
@@ -55,7 +55,17 @@ class DictionarySegment : public BaseDictionarySegment {
...
   EncodingType encoding_type() const final;
 
+  // Returns the first value ID that refers to a value >= the search value and INVALID_VALUE_ID if all values are
+  // smaller than the search value. Here, INVALID_VALUE_ID does not represent NULL (which isn't stored in the
+  // dictionary anyway). Imagine a segment with values from 1 to 10. A scan for `WHERE a < 12` would retrieve
+  // `lower_bound(12) == INVALID_VALUE_ID` and compare all values in the attribute vector to `< INVALID_VALUE_ID`.
+  // Thus, returning INVALID_VALUE_ID makes comparisons much easier. However, the caller has to make sure that
+  // NULL values stored in the attribute vector (stored with a value ID of unique_values_count()) are excluded.
+  // See #1471 for a deeper discussion.
   ValueID lower_bound(const AllTypeVariant& value) const final;
```
mrzzzrm: (Why) is it not possible to return a std::nullopt here, for "not found"? I think that would prevent half of the possible confusion around what this function returns. We have to check the return value anyway, right? So there would be no perf loss from returning a std::optional?

mrks: I started that, but that just moves the problem to the caller. A value id that is higher than everything else is exactly what we want for `< non_existing_value`.

mrks:
Jeez, finally

mrks:
@mrzzzrm just realized that you didn't yet comment on my answer.  Please don't let this preempt the discussion.

mrzzzrm:
If you say it wouldn't make anything easier, I am content with the current "solution" ;)

Issue #1534, opened 2019-03-08
Compile times are through the roof (again)


mrzzzrm:
Single threaded build times:

Commit | Clang-Debug | Clang-Release
------ | ----------- | -------------
9cdd14dde Compile time: Don't instantiate  (#1628) | 30m1,058s | 56m24,493s 
d65e68134 Update ExpressionReductionRule  | |
f9b8de0ad Make LIKE rewrite emit a  (#1638 | |
36910944e Naming: Add 'Segment' to FoR- and (#1639) | |
486f145f1 CI for Mac (#1637) | |
2c75839dd Fix multi-threaded multi-predicate join (#1625) | |
298cfa517 Clarify INVALID_VALUE_ID (#1626) | |
346ca70b6 Remove print in test (#1634) | |
30fb1df47 Print: Remove handling of empty chunks (#1629) | |
d36f9c424 Fix lossless tests on OS X (#1630) | |
eca9c67e1 Safer handling of cross-type operations with lossless | |
01bee9b17 TPC-H 6: Calculate discount only once (#1624) | |
e86add934 Remove scan alignment (#1621) | |
52c8a7f8e Enable TSan for NUMA builds (#1618) | |
560df6656 Use indexes for SQLite verification of join order ben | |
ac4e44549 Create resources link only once (#1620) | |
09ce0cff7 Remove unused return value of  | |
beb1e0adb Add new Between modes to ColumnStatistics (#1616) | |
0aeda11d6 Improve benchmark script (#1584) | 30m28,065s |  59m35,528s
771c5b4fe Add IF EXISTS support for DROP TABLE and DROP |  | 
5f048d3b3 Fix btree_index_test for libc++ (#1612) |  | 
89f2824a9 Fix TPC-H for larger scale factors  (#1608) |  | 
43d2b5460 Cmake: add link to resources (#1604) |  | 
8d5e2a487 Fix MVCC performance regression, fix MVCC data |  | 
5050ff20f Print file/line info for Fail() (#1605) |  | 
055b1fba4 Unified handling of operator<<(std::ostream, . |  | 
2628fbfd0 Compile time: More type erasure in ColumnVsCol |  | 
0222b0e3b JoinHash for semi/anti: Insert values just onc |  | 
0065b2530 Better documentation for --table_path option ( |  | 
6943c012c Fix make clean (#1594) |  | 
31aea916a Between: Make order of inserted predicates det |  | 
[...]

PR #1628, opened 2019-04-22
Compile time: Don't instantiate unused scans

mrks:
This reduces the compile time by 243.2 seconds by not instantiating unused scans. Refs #1534.

before:
```
         113.1 weighted s to build src/lib/CMakeFiles/hyrise.dir/operators/table_scan/column_between_table_scan_impl.cpp.o (113.1 s CPU time)
          32.0 weighted s to build src/lib/CMakeFiles/hyrise.dir/operators/table_scan/column_is_null_table_scan_impl.cpp.o (32.0 s CPU time)
          97.8 weighted s to build src/lib/CMakeFiles/hyrise.dir/operators/table_scan/column_like_table_scan_impl.cpp.o (97.8 s CPU time)
         221.1 weighted s to build src/lib/CMakeFiles/hyrise.dir/operators/table_scan/column_vs_value_table_scan_impl.cpp.o (221.1 s CPU time)
```

after:
```
          45.0 weighted s to build src/lib/CMakeFiles/hyrise.dir/operators/table_scan/column_between_table_scan_impl.cpp.o (45.0 s CPU time)
          29.8 weighted s to build src/lib/CMakeFiles/hyrise.dir/operators/table_scan/column_is_null_table_scan_impl.cpp.o (29.8 s CPU time)
          32.7 weighted s to build src/lib/CMakeFiles/hyrise.dir/operators/table_scan/column_like_table_scan_impl.cpp.o (32.7 s CPU time)
         113.3 weighted s to build src/lib/CMakeFiles/hyrise.dir/operators/table_scan/column_vs_value_table_scan_impl.cpp.o (113.3 s CPU time)
```


mrks:
Yepp, makes sense!

mrks:
I am changing some Jenkins parameters to make it fit for the mac build, which is why it's a bit lazy right now.

src/lib/operators/table_scan/column_is_null_table_scan_impl.cpp
```diff
@@ -42,11 +42,19 @@ std::shared_ptr<PosList> ColumnIsNullTableScanImpl::scan_chunk(const ChunkID chu
...
                                                       const std::shared_ptr<const PosList>& position_filter) const {
-  segment_with_iterators_filtered(segment, position_filter, [&](auto it, const auto end) {
-    const auto invert = _predicate_condition == PredicateCondition::IsNotNull;
-    const auto functor = [&](const auto& value) { return invert ^ value.is_null(); };
-
-    _scan_with_iterators<false>(functor, it, end, chunk_id, matches);
+  segment_with_iterators_filtered(segment, position_filter, [&](auto it, [[maybe_unused]] const auto end) {
+    // Don't instantiate this for this for DictionarySegments and ReferenceSegments to save compile time.
+    // ValueSegments are handled in _scan_value_segment()
+    // ReferenceSegments are handled via position_filter
```
mrks: No, they are not, because `ColumnIsNullTableScanImpl` is not a `AbstractSingleColumnTableScanImpl`. The latter is thus misnamed. Maybe `AbstractDereferencedTableScanImpl` or alike?

mrzzzrm: Why is it not a AbstractSingleColumnTableScanImpl?

mrzzzrm: I removed the apparently unused `position_filter` from the `ColumnIsNullTableScanImpl` because this one's existence was what lead me to believe ReferenceSegments are handled implicitly

mrks: > Why is it not a AbstractSingleColumnTableScanImpl?

```
  // We need to override scan_chunk because we do not want ReferenceSegments to be resolved (which would remove
  // NullValues from the referencing PosList)
  std::shared_ptr<PosList> scan_chunk(const ChunkID chunk_id) const override;
```

I'd like to rename the `AbstractSingleColumnTableScanImpl` though. The null scan certainly scans a single column, so it would only be logically for it to inherit from that. Options:

* AbstractDereferencedTableScanImpl
* AbstractResolvedReferenceTableScanImpl
* ...?

mrzzzrm: every name will suck, I suppose.... AbstractNonNullTableScan? I guess I could live with AbstractDereferencedTableScanImpl if the class doc makes  is unmistakably clear why this superclass exists and why it is named like that. ColumnIsNullTableScanImpl's docs should likewise be very clear as to why it does NOT inherit from AbstractDereferencedTableScanImpl

mrks: > There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

> ColumnIsNullTableScanImpl's docs should likewise be very clear as to why it does NOT inherit from AbstractDereferencedTableScanImpl

What are you missing in the comment quoted above?

mrzzzrm: Took me longer than a moment to understand why this comment implies that AbstractSingleColumnTableScanImpl is not the base class of ColumnIsNullTableScanImpl.

"// We need to override scan_chunk" implies "We cannot derive from ASCTSI" but that's not an obvious implication.

mrzzzrm: Also I'd really like this to be stated in the class comment of ColumnIsNullTableScanImpl. 

Just come up with something ;)

mrks: What do you think?

PR #1629, opened 2019-04-22
Print: Remove handling of empty chunks

### LINKED ISSUE #1627: Get rid of PrintIgnoreEmptyChunks (fixed in #1629)
Empty chunks (other than the first and only) have been illegal for a while now. The print operator should respect that.

mrks:
fixes #1627

### DIFF
--- a/src/lib/operators/print.cpp
+++ b/src/lib/operators/print.cpp
@@ -73,10 +73,6 @@ std::shared_ptr<const Table> Print::_on_execute() {
   // print each chunk
   for (ChunkID chunk_id{0}; chunk_id < input_table_left()->chunk_count(); ++chunk_id) {
     auto chunk = input_table_left()->get_chunk(chunk_id);
-    if (chunk->size() == 0 && (_flags & PrintIgnoreEmptyChunks)) {
-      continue;
-    }
-
     _out << "=== Chunk " << chunk_id << " ===" << std::endl;
 
     if (chunk->size() == 0) {
--- a/src/test/operators/print_test.cpp
+++ b/src/test/operators/print_test.cpp
@@ -300,7 +300,7 @@ TEST_F(OperatorsPrintTest, EmptyTable) {
       "|     int|  string|\n"
       "|not null|not null|\n"
       "=== Chunk 0 ===\n"
-      "Empty chunk.\n"; 
+      "Empty chunk.\n";
 
   EXPECT_EQ(output.str(), expected_output);
   EXPECT_FALSE(wrapper.is_printing_mvcc_information());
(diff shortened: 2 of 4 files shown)


src/lib/operators/print.cpp
```diff
@@ -73,10 +73,6 @@ std::shared_ptr<const Table> Print::_on_execute() {
   // print each chunk
   for (ChunkID chunk_id{0}; chunk_id < input_table_left()->chunk_count(); ++chunk_id) {
     auto chunk = input_table_left()->get_chunk(chunk_id);
-    if (chunk->size() == 0 && (_flags & PrintIgnoreEmptyChunks)) {
```
Bouncner: So the flag is removed which disables the printing of "empty chunk.".

Do I get this PR right? : we cannot have empty chunks except at the beginning of a table's lifetime, hence we always print the info as it helps us to recognize empty tables. As soon as one row has been added, the "empty chunk" info will never be display again anyhow.

Correct?

src/test/operators/print_test.cpp
```diff
@@ -238,15 +190,6 @@ TEST_F(OperatorsPrintTest, MVCCFlag) {
       "|not null|not null||      |      |      |\n";
 
   EXPECT_EQ(output.str(), expected_output);
-  EXPECT_TRUE(print_wrap.is_printing_empty_chunks());
-  EXPECT_TRUE(print_wrap.is_printing_mvcc_information());
```
Bouncner: Why remove?

Bouncner: Oh, never mind. Github broke up the diff.

src/test/operators/print_test.cpp
```diff
@@ -181,52 +179,6 @@ TEST_F(OperatorsPrintTest, TruncateLongValueInOutput) {
...
-  auto print_wrap_withempty = PrintWrapper(wrap, output_withempty, 0);
-  print_wrap_withempty.execute();
-
-  auto expected_output_withempty =
-      "=== Columns\n"
-      "|column_1|column_2|\n"
-      "|     int|  string|\n"
-      "|not null|not null|\n"
-      "=== Chunk 0 ===\n"
-      "Empty chunk.\n";
```
Bouncner: Might be related to my understanding question: we still print the empty information, but apparently no test does cover this situation?

src/lib/operators/print.cpp
mrks: No. The option of *ignoring* empty chunks is removed.

Bouncner: Yeah, that's what I tried to say.

PR #1633, opened 2019-04-23
RFC: Erase iterators/accessors depending on compile-time setting

mrks:
Proposed in #1534, this PR would add a parameter `-DERASE_SEGMENT_TYPES=RunLength,...` to cmake. With that set, iterators and accessors for the listed segment type are erased even for release builds. If they end up being used, a PerformanceWarning is shown.

Pro: This reduces the compile time (see below). Not significantly, but if all you're running is using dictionary encoding, this saves you some time over the course of a day.
Contra: It's ugly. Luckily only in two files, but still...
Contra: We won't be able to test it properly and in all configurations.

Any suggestions on how to make this less intrusive (especially how to use less macros) are welcome. Also, general criticism is appreciated. Not sure we actually want this.

```
Mac                         time ninja hyrise (from clean)
erase none:                 14m3.811s
erase all but Dictionary:   10m29.149s
erase all:                  9m18.317s

nemea                       time ninja hyrise (from clean)
erase none:                 5m30.759s
erase all but Dictionary:   4m2.030s
erase all:                  3m7.500s
```


src/lib/storage/create_iterable_from_segment.hpp
```diff
@@ -39,38 +39,58 @@ auto create_iterable_from_segment(const ValueSegment<T>& segment) {
...
     return create_any_segment_iterable<T>(segment);
   } else {
     return DictionarySegmentIterable<T, FixedStringVector>{segment};
   }
+#endif
 }
 
 template <typename T, bool EraseSegmentType = HYRISE_DEBUG>
 auto create_iterable_from_segment(const FrameOfReferenceSegment<T>& segment) {
+#ifdef HYRISE_ERASE_FRAMEOFREFERENCE
```
Bouncner: Ain't there some C++ magic to have a overload-all method `create_iterable_from_segment` that is enabled when `HYRISE_ERASE_FRAMEOFREFERENCE`?

Bouncner:
I like the flag, since most users won't be affected. But the code changes look like the style of code that I was super happy to have never seen in the Hyrise code base.

Is that runtime with or without your PR changes to the column scan?

mrzzzrm:
I am surprised this is so uninvasive. Still, the stuff in reference_segment_iterable.hpp will always be considered dark magic, no one will dare to touch that and probably break things/increase compile time without noticing if they do.

Hypothetical: Hyrise would have three (in release) unerased column types: Values and Dictionary and References to the former two. Everything else is always erased. Would anyone ever realistically need/profit from non-erased RLE or FoR?

mrks:
> Is that runtime with or without your PR changes to the column scan?

Without.

> But the code changes look like the style of code that I was super happy to have never seen in the Hyrise code base.

110% agreed. One idea that I just had is to add a header file that sets a constexpr bool depending on the value of the macro.

> Hypothetical: Hyrise would have three (in release) unerased column types: Values and Dictionary and References to the former two. Everything else is always erased. Would anyone ever realistically need/profit from non-erased RLE or FoR?

If you use those, you should not erase them. I'd rather not introduce a hierarchy of segment types. Instead, as a developer, you could decide that, for now, you won't use them, so you don't compile them.

mrzzzrm:
> 
> 
>     Hypothetical: Hyrise would have three (in release) unerased column types: Values and Dictionary and References to the former two. Everything else is always erased. Would anyone ever realistically need/profit from non-erased RLE or FoR?
> 
> If you use those, you should not erase them. I'd rather not introduce a hierarchy of segment types. Instead, as a developer, you could decide that, for now, you won't use them, so you don't compile them.

I don't have a hard opinion on this, so please don't feel like I am pressing on this, but I think we do have a de-facto hierarchy of segment types: The mentioned three are by far the most used and commonly (among hyrise devs) understood segment types. There are optimizations for DictionarySegments in, e.g., the scan operator, a privilege that the secondary segment types do not have and in the foreseeable future(?) will not have. 

Introducing the compile time flags breaks an explicit taboo - hard-coding a de-facto fact does not. And it should be relatively easy to switch from hard-coded erasure back to the current state, if the need ever arises.

mrks:
It feels wrong to ship something were half of the officially supported segment types are crippled. A different option would have been to specifically enable things like RLE in cmake, but to make this work everywhere and fix all tests would probably require too many code changes for too little result.

> Introducing the compile time flags breaks an explicit taboo

Which taboo exactly do you mean? We do not really enable / disable code completely, we just selectively turn on debug mode for certain types.

mrzzzrm:
> It feels wrong to ship something were half of the officially supported segment types are crippled

Are we sure that the overhead of a single virtual function call on RLE/FoR - which have to do their own computations in order to retrieve the value - justifies the term "crippled"?

> Which taboo exactly do you mean?

That of having a large set of compile flags. You have seen this on Hyrise1 and I have seen this on another project. A set of S compile flags creates 2^S different versions of Hyrise which cannot all be tested - thus this begs for a subset of them not working.

mrks:
> Are we sure that the overhead of a single virtual function call on RLE/FoR - which have to do their own computations in order to retrieve the value - justifies the term "crippled"?

[Figure 3b](https://openproceedings.org/2019/conf/edbt/EDBT19_paper_152.pdf) says so.

> That of having a large set of compile flags.

Just wanted to make sure that we are talking about the same thing. For me the difference here is that we are already erasing all of these in the debug build. It's not really code being added or removed. However that's not perfect (which is why this is an RFC).

mrzzzrm:
As I said, I am not pressing nor disagreeing for the sake of it, just offering my considerations. I think we have all been sufficiently burned by compile flags to not wade into those waters carelessly.

Just two last bits: As a Hyrise newbie (be it student or researcher from another uni) I would like code base as a black box to have as few as possible configuration knobs. NUMA on/off; JIT on/off is already irritating enough (My newbie thoughts would go like this: "The configuration claimed that JIT is enabled, does that mean JIT is ALWAYS on for every query, will the normal TableScan operator even be executed if I start the TPC-H? Is a numa-enabled build even designed to run on my laptop? Hyrise is a beast written by magicans, I better only touch it veeery carefull") - more compile flags just increase this. I also want other researchers to be able to compile Hyrise and test their DBMS against it without having to worry whether they compiled the "real" version of Hyrise.

Aaand: Are RLE and FoR even designed for "performant" access? Wouldn't it be possible to just declare them as compression schemes that just come with a certain performance impact? The figure in the paper shows a grave difference between erased/non-erased, so I know this is a weak point.

mrks:
> As I said, I am not pressing nor disagreeing for the sake of it, just offering my considerations. I think we have all been sufficiently burned by compile flags to not wade into those waters carelessly.

That's how I took it. Don't get me wrong - I'm not trying to convince you; I'm not convinced m

> I also want other researchers to be able to compile Hyrise and test their DBMS against it without having to worry whether they compiled the "real" version of Hyrise.

Isn't that an argument in favor of NOT crippling the iterators? If only we know how to run a benchmark where all segments run at their best, that's hardly fair.

> Just two last bits: As a Hyrise newbie (be it student or researcher from another uni) I would like code base as a black box to have as few as possible configuration knobs.

I wouldn't openly tell people about it. If you just compile Hyrise once, you shouldn't have to know about it. Only once you compile it tens of times a day, you might want to exclude some segments. Maybe add a link to a wiki page with "hacks for people who know what they are doing".

mrzzzrm:
> Isn't that an argument in favor of NOT crippling the iterators? If only we know how to run a benchmark where all segments run at their best, that's hardly fair.

Then we would need to play fair and never publish benchmark results with non-erased RLE/FoR, I guess.

mrks:
Nobody said we would. All benchmark runs (maybe except for the internal ones that we post in PRs) must not use this setting.

mrks:
I was originally thinking about moving the macros to the segment's header:
```
#ifdef HYRISE_ERASE_FRAMEOFREFERENCE
static constexpr bool EraseType = true;
#else 
static constexpr bool EraseType = false;
```

We could then change

```
+#ifdef HYRISE_ERASE_FRAMEOFREFERENCE
+          if constexpr (std::is_integral_v<T>) {
+            if constexpr (std::is_same_v<SegmentType, FrameOfReferenceSegment<T>>) return;
+          }
+#endif
```

to

```
+          if constexpr (std::is_integral_v<T>) {
+            if constexpr (std::is_same_v<SegmentType, FrameOfReferenceSegment<T>> && SegmentType::EraseType) return;
+          }
```

On second thought, however, I wonder if this is really cleaner or is just hiding the mess somewhere else. The current implementation makes it (painfully) clear what we are doing.

For those who agree, could you review the current version, please?

src/lib/CMakeLists.txt
```diff
@@ -673,6 +673,16 @@ if (${ENABLE_NUMA_SUPPORT})
     set(LIBRARIES ${LIBRARIES} ${NUMA_LIBRARY} hpinuma_msource_s)
 endif()
 
+set(ERASE_SEGMENT_TYPES "" CACHE STRING "Erase iterators and accessors for these types (e.g., RunLength,FrameOfReference). Good for compile time, bad for run time (if these types end up being used)")
```
mrzzzrm: Does checking for the valid segment types make sense? Otherwise I'll probably accidentally do something like `-DERASE_SEGMENT_TYPE="FrameOfReferrence"`

Should be just a call to `list(SEARCH <valid_types> my_type)` https://cmake.org/cmake/help/v3.12/command/list.html

mrks: I had the same thought, but at the same time I didn’t feel good about adding redundant information yet again. #1660 shows that we will forget it.

In the end, I figured that (a) developers will only type this once in a while, and (b) nothing goes really wrong if it’s mistyped.

An alternative would be to pass the comma-separated string to the compiler and use compile-time string magic to split it there. That way, we could check for the existence of all type, but it feels like overkill.

mrzzzrm: If you don't like it, that's fine. Just wanted to mention it.

Bouncner:
Wer zündet den Knaller?

mrzzzrm:
Soll Markus machen

mrks:
bumm

PR #1636, opened 2019-04-24
Support IF NOT EXISTS in CREATE TABLE and CREATE VIEW

d-justen:
...as suggested in #1189


src/lib/operators/maintenance/create_table.cpp
```diff
@@ -41,18 +43,19 @@ const std::string CreateTable::description(DescriptionMode description_mode) con
 }
 
 std::shared_ptr<const Table> CreateTable::_on_execute() {
-  // TODO(anybody) chunk size and mvcc not yet specifiable
-  const auto table = std::make_shared<Table>(column_definitions, TableType::Data, Chunk::DEFAULT_SIZE, UseMvcc::Yes);
-
-  StorageManager::get().add_table(table_name, table);
-
+  // If IF NOT EXISTS is not set and the table already exists, StorageManager throws an exception
```
mrks: What happens if I do it in the console? Optimally, typical user input errors shouldn't kill the entire system.

d-justen: The system is killed in fact if a user does that. This happens because `CreateTable` is strangely enough an `AbstractReadOnlyOperator`. It could become an `AbstractReadWriteOperator`, which wraps the execution in a try-catch block and comes with a state, but then it would also need a transaction context. What do you think?

mrzzzrm: We agreed on a fail-load policy for Hyrise, meaning that every failing `Assert` winds down the system. We weakened this with `AssertInput`, but this is to be used sparingly and probably only in the `SQLTranslator`. Definitely not in Operators. 

The Console is the only component of Hyrise that we want to survive user input errors. The Console, however, is not important enough to warrant weakening the policy that _we do not handle exceptions in operators (in fact, we almost never handle them)_. If the console wants to survive invalid input, it needs to validate the input itself. In this case this means that the console should check generated LQPs/PQPs for Create* operators and validate whether they can be executed.

Another point: The `AbstractReadOnlyOperator` and `AbstractReadWriteOperator` classes are easy to misunderstand in their naming. The "Read" and the "Write" refer **only** to row/MVCC data. 

(Regarding exceptions: I could write a looong rant here about how using them almost always leads to badly tested code, badly readable code, unexpected behaviour and swallowed errors. I could dig up an old version of the `SQLPipeline/Statement` - which used to handle exceptions - to prove this point. It was just terrible.)

src/lib/logical_query_plan/create_table_node.cpp
```diff
@@ -6,13 +6,19 @@
...
+      column_definitions(column_definitions),
+      if_not_exists(if_not_exists) {}
 
 std::string CreateTableNode::description() const {
   std::ostringstream stream;
 
-  stream << "[CreateTable] Name: '" << table_name << "' (";
+  const std::string if_not_exists_output = if_not_exists ? "IfNotExists " : "";
+
+  stream << "[CreateTable] " + if_not_exists_output + "Name: '" << table_name << "' (";
```
mrzzzrm: In streams, use operator<< over operator+

src/lib/logical_query_plan/create_view_node.cpp
```diff
@@ -7,28 +7,27 @@
...
+    : BaseNonQueryNode(LQPNodeType::CreateView), view_name(view_name), view(view), if_not_exists(if_not_exists) {}
 
 std::string CreateViewNode::description() const {
   std::stringstream stream;
-  stream << *_view->lqp;
+  stream << *view->lqp;
+  const std::string if_not_exists_output = if_not_exists ? "IfNotExists " : "";
 
-  return "[CreateView] Name: '" + _view_name + "' (\n" + stream.str() + ")";
+  return "[CreateView] " + if_not_exists_output + "Name: '" + view_name + "' (\n" + stream.str() + ")";
```
mrzzzrm: minor: use the stream to build your return value - and use operator<< over operator+ there.

src/test/logical_query_plan/create_table_node_test.cpp
```diff
@@ -11,7 +11,7 @@ class CreateTableNodeTest : public ::testing::Test {
   void SetUp() override {
     column_definitions.emplace_back("a", DataType::Int, false);
     column_definitions.emplace_back("b", DataType::Float, true);
-    create_table_node = CreateTableNode::make("some_table", column_definitions);
+    create_table_node = CreateTableNode::make("some_table", column_definitions, false);
```
mrzzzrm: @mrks I insist that having a bool parameter here is NOT readable.
(@d-justen ignore this comment)

src/test/sql/sql_translator_test.cpp
```diff
@@ -1609,11 +1609,39 @@ TEST_F(SQLTranslatorTest, CreateAliasView) {
...
+TEST_F(SQLTranslatorTest, CreateViewIfNotExists) {
+  const auto query =
+      "CREATE VIEW IF NOT EXISTS my_first_view AS SELECT a, b, a + b, a*b AS t FROM int_float WHERE a = 'b';";
+  const auto result_node = compile_query(query);
+
+  // clang-format off
+  const auto select_list_expressions = expression_vector(int_float_a, int_float_b, add_(int_float_a, int_float_b), mul_(int_float_a, int_float_b));  // NOLINT
+
+  const auto view_lqp =
+          AliasNode::make(select_list_expressions, std::vector<std::string>({"a", "b", "a + b", "t"}),
```
mrzzzrm: please indent with two spaces

src/test/sql/sql_translator_test.cpp
```diff
@@ -1609,11 +1609,39 @@ TEST_F(SQLTranslatorTest, CreateAliasView) {
...
 
-  const auto expected_lqp = CreateViewNode::make("my_second_view", view);
+  const auto expected_lqp = CreateViewNode::make("my_second_view", view, false);
 
   EXPECT_LQP_EQ(actual_lqp, expected_lqp);
 }
 
+TEST_F(SQLTranslatorTest, CreateViewIfNotExists) {
+  const auto query =
+      "CREATE VIEW IF NOT EXISTS my_first_view AS SELECT a, b, a + b, a*b AS t FROM int_float WHERE a = 'b';";
```
mrzzzrm: why such a complicated select list? Just keep it simple: `SELECT b, a FROM int_float WHERE a = 'b';";` and your LQP below gets smaller as well ;) "CreateView*" tests should test CreateView, not expression parsing

src/test/sql/sql_translator_test.cpp
```diff
@@ -1609,11 +1609,39 @@ TEST_F(SQLTranslatorTest, CreateAliasView) {
...
+  const auto view_columns = std::unordered_map<ColumnID, std::string>({
+                                                                              {ColumnID{0}, "a"},
+                                                                              {ColumnID{1}, "b"},
+                                                                              {ColumnID{3}, "t"},
+                                                                      });
+  // clang-format on
+
+  const auto view = std::make_shared<LQPView>(view_lqp, view_columns);
+
+  const auto lqp = CreateViewNode::make("my_first_view", view, true);
```
mrzzzrm: "expected_lqp" for consistency

src/test/sql/sql_translator_test.cpp
```diff
@@ -1609,11 +1609,39 @@ TEST_F(SQLTranslatorTest, CreateAliasView) {
...
+  // clang-format off
+  const auto select_list_expressions = expression_vector(int_float_a, int_float_b, add_(int_float_a, int_float_b), mul_(int_float_a, int_float_b));  // NOLINT
+
+  const auto view_lqp =
+          AliasNode::make(select_list_expressions, std::vector<std::string>({"a", "b", "a + b", "t"}),
+                          ProjectionNode::make(select_list_expressions,
+                                               PredicateNode::make(equals_(int_float_a, "b"),
+                                                                   stored_table_node_int_float)));
+
+  const auto view_columns = std::unordered_map<ColumnID, std::string>({
```
mrzzzrm: formatting seems weird here, I think whatever clang-format would pick would look better

src/test/sql/sql_translator_test.cpp
```diff
@@ -1635,7 +1663,24 @@ TEST_F(SQLTranslatorTest, CreateTable) {
...
+
+  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+}
+
+TEST_F(SQLTranslatorTest, CreateTableIfNotExists) {
+  const auto actual_lqp = compile_query(
+      "CREATE TABLE IF NOT EXISTS a_table (a_int INTEGER, a_long LONG, a_float FLOAT, a_double DOUBLE NULL, a_string "
+      "VARCHAR(10) NOT NULL)");
+
+  auto column_definitions = TableColumnDefinitions{};
```
mrzzzrm: minor

```c++
const auto column_definitions = TableColumnDefinitions{
  {"a_int", DataType::Int, false},
  ....
};

would be a lot less chars ;)
```

src/lib/operators/maintenance/create_table.cpp
d-justen: Thanks for the explanation, @mrzzzrm! 
So, if we want the console to check this, we would need it to do so at least for all 7 maintenance operators (CreatePreparedPlan, CreateTable, CreateView, DropTable, DropView, ShowColumns, ShowTables... There is no DropPreparedPlan-Operator implemented yet, is it?) if not more. 

Should we open a new issue for this?

mrzzzrm: > So, if we want the console to check this, we would need it to do so at least for all 7 maintenance operators.

Yes. Not that hard, though:

```c++
bool verify_plan(lqp) {
  visit_lqp(lqp, [&](const auto& node) {
    switch (node.type) {
      case LQPNodeType::CreateTable:
        return // check whether table already exists
    }
  });
}
```

Covered in "Check for table existence in console before running plan containing CreateTable" as part of https://github.com/hyrise/hyrise/issues/1189 so it doesn't need a new issue.

mrzzzrm:
Fixed some minor, minor things myself. Please note they formatting of LQPs in code as applied by https://github.com/hyrise/hyrise/pull/1636/commits/a9f948e3c50550ed405177a89a42d1257524c32e which makes LQPs, especially larger ones, much nicer to read.

PR #1433, opened 2019-01-25
Upgrade to LLVM7 and Clang7

FabianWiebe:
### Main Changes
This pr updates LLVM6 and Clang6 to LLVM7 and Clang7.
This requires the replacement of the LLVM6 files `CloneFunction.cpp` and `InlineFunction.cpp` with their LLVM7 pendant required for JIT.
The interface to pass a module to the JIT compiler slightly changed so that modules are now stored in a unique_ptr instead of a shared_ptr.

### Support for LLVM7 and Clang7
MacOS - Brew will automatically install the latest version of LLVM with `brew install llvm`.
Ubuntu 18.10 supports LLVM7 and Clang7.
Ubuntu 18.04 does not support LLVM7 and Clang7 out of the box.
To install the required versions in Ubuntu 18.04, add the following lines to `/etc/apt/sources.list`.
````
deb http://apt.llvm.org/stretch/ llvm-toolchain-stretch-7 main
# deb-src http://apt.llvm.org/stretch/ llvm-toolchain-stretch-7 main
````
PR removes one issue from #1077.
#### Open issues:
- remove workaround to remove LLVM lib files from coverage, see #1074
- remove experimental/filesystem (reopen #1151)

PR #1637, opened 2019-04-25
CI for Mac

### LINKED ISSUE #722: CI for mac
To prevent things like #721

mrks:
I did not find a nice way to run the traditional FullCI steps and this in parallel. It is possible, but Blue Ocean cannot visualize it properly. In the end, all steps are flying around and it's hard to find something. Thus, it is its own phase right now. Other systems (e.g., Power, Raspberry, ...) can be added there without too much hassle (and those would run in parallel).

fixes #722


install.sh
```diff
@@ -36,7 +33,7 @@ if echo $REPLY | grep -E '^[Yy]$' > /dev/null; then
                 fi
             done
 
-            if ! brew install llvm --with-toolchain; then
+            if ! brew install llvm; then
```
mrks: Not necessary/supported anymore

src/test/lossless_cast_test.cpp
```diff
@@ -113,9 +113,9 @@ TEST_F(LosslessCastTest, IntegralToIntegral) {
   EXPECT_EQ(lossless_cast<int32_t>(int64_t(3)), 3);
 
   EXPECT_EQ(lossless_cast<int32_t>(2'147'483'647), 2'147'483'647);
-  EXPECT_EQ(lossless_cast<int32_t>(int64_t{2'147'483'648}), std::nullopt);
-  EXPECT_EQ(lossless_cast<int32_t>(int64_t{-2'147'483'648}), -2'147'483'648);
-  EXPECT_EQ(lossless_cast<int32_t>(int64_t{-2'147'483'649}), std::nullopt);
+  EXPECT_EQ(lossless_cast<int32_t>(2'147'483'648), std::nullopt);
```
mrks: Broke this on purpose to make the CI red

install.sh
```diff
@@ -15,9 +15,6 @@ if echo $REPLY | grep -E '^[Yy]$' > /dev/null; then
             exit 1
         fi
 
-        # Needed for proper building under macOS
-        xcode-select --install
```
mrks: We require xcode to be installed prior to executing this, but the command line tools come automatically

Jenkinsfile
```diff
@@ -220,31 +223,51 @@ node {
...
+node('mac') {
+  stage("clangDebugMac") {
+    if (env.BRANCH_NAME == 'master' || full_ci) {
+      try {
+        checkout scm
+
+        // We do not use install.sh here as there is no way to run OS X in a Docker container
+        sh "git submodule update --init --recursive --jobs 4"
+
+        sh "mkdir clang-debug && cd clang-debug && /usr/local/bin/cmake -DCMAKE_CXX_COMPILER_LAUNCHER=/usr/local/bin/ccache -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=/usr/local/Cellar/llvm\\@7/7.0.1/bin/clang -DCMAKE_CXX_COMPILER=/usr/local/Cellar/llvm\\@7/7.0.1/bin/clang++ .."
```
mrks: Forced to llvm7 as 6 does not have std::filesystem on Mac. With #1433, this will be the default anyway.

Jenkinsfile
```diff
@@ -173,7 +176,7 @@ node {
             sh "TSAN_OPTIONS=suppressions=resources/.tsan-ignore.txt ./clang-release-thread-sanitizer/hyriseTest clang-release-thread-sanitizer"
             sh "TSAN_OPTIONS=suppressions=resources/.tsan-ignore.txt ./clang-release-thread-sanitizer/hyriseSystemTest clang-release-thread-sanitizer"
           } else {
-            Utils.markStageSkippedForConditional("clangReleaseThreadSanitizerNoNuma")
+            Utils.markStageSkippedForConditional("clangReleaseThreadSanitizer")
```
mrks: Missed that in #1618

Jenkinsfile
Bensk1: Just for clarification what is the `\\@7` necessary for?

Jenkinsfile
```diff
@@ -220,31 +223,51 @@ node {
...
+      try {
+        checkout scm
+
+        // We do not use install.sh here as there is no way to run OS X in a Docker container
+        sh "git submodule update --init --recursive --jobs 4"
+
+        sh "mkdir clang-debug && cd clang-debug && /usr/local/bin/cmake -DCMAKE_CXX_COMPILER_LAUNCHER=/usr/local/bin/ccache -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=/usr/local/Cellar/llvm\\@7/7.0.1/bin/clang -DCMAKE_CXX_COMPILER=/usr/local/Cellar/llvm\\@7/7.0.1/bin/clang++ .."
+        sh "cd clang-debug && CCACHE_CPP2=yes CCACHE_SLOPPINESS=file_macro PATH=/usr/local/bin:$PATH make -j libjemalloc-build"
+        sh "cd clang-debug && CCACHE_CPP2=yes CCACHE_SLOPPINESS=file_macro make -j4"
+        sh "cd clang-debug && ./hyriseTest"
```
Bensk1: Any specific reason for not using the typical `./clang-debug/hyriseTest`?

Jenkinsfile
```diff
@@ -1,6 +1,7 @@
 import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 
-node {
+try {  // TODO fix indentation - change for easier reviewing
+node('master') {
```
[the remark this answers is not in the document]
mrks: It's not a node, but a node label. "Run this on any node that is tagged 'master'".

Jenkinsfile
mrks: `brew install llvm` will give you clang-8. To install clang-7, we need `brew install llvm@7`, which creates `/usr/local/Cellar/llvm@7`. The `@` needs to be escaped for some reason, and as we are in a string, the escape character needs to be escaped, too.

Jenkinsfile
mrks: None in particular. Will change.

mrks:
> I am not an expert in manually parsing Jenkinsfiles

And I'm not an expert in writing them. If it works and provably runs on the Mac Mini, it can't be too bad, I guess... ;)

Jenkinsfile
Bensk1: Understood.

PR #1647, opened 2019-04-28
Projection performance: Don't allocate MvccData to then immediately overwrite it

### LINKED ISSUE #1645: Column pruning in Projection takes unexpectedly long
image

image

\> 1'000'000ns to select three segments from a chunk? This is a release build with sf=5

mrzzzrm:
Fix #1645 

Also, avoid redundant calls to `input_table_left()` for each `Chunk` to save a shared_ptr copy.
Also, remove `Chunk::set_mvcc_data()` because it isn't needed anymore since `Table::append_chunk()` now takes the new Chunk's MvccData as well. One entry point less for potential concurrency problems.

TPC-H with sf=5 
```
+----------------+-----------------+------+-----------------+------+------------+---------------------------------+
| Benchmark      | prev. iter/s    | runs | new iter/s      | runs | change [%] | p-value (significant if <0.001) |
+----------------+-----------------+------+-----------------+------+------------+---------------------------------+
| TPC-H 1        | 0.0490739420056 | 10   | 0.0489541068673 | 10   | -0%        |                          0.4264 |
| TPC-H 2        | 0.671977341175  | 60   | 0.730327427387  | 60   | +9%        |                          0.0000 |
| TPC-H 3        | 0.448871999979  | 60   | 0.448813527822  | 60   | -0%        |                          0.9500 |
| TPC-H 4        | 0.203322425485  | 41   | 0.208486706018  | 42   | +3%        |                          0.0162 |
| TPC-H 5        | 0.277782648802  | 56   | 0.319226652384  | 60   | +15%       |                          0.0000 |
| TPC-H 6        | 0.775415897369  | 60   | 0.792824447155  | 60   | +2%        |                          0.0666 |
| TPC-H 7        | 0.0973886549473 | 20   | 0.0974526852369 | 20   | +0%        |                          0.9659 |
| TPC-H 8        | 0.429121643305  | 60   | 0.514973104     | 60   | +20%       |                          0.0000 |
| TPC-H 9        | 0.113636538386  | 23   | 0.121970623732  | 25   | +7%        |                          0.0000 |
| TPC-H 10       | 0.216463521123  | 44   | 0.217899993062  | 44   | +1%        |                          0.0378 |
| TPC-H 11       | 2.28297924995   | 60   | 3.03103637695   | 60   | +33%       |     (run time too short) 0.0000 |
| TPC-H 12       | 0.699244916439  | 60   | 0.761795580387  | 60   | +9%        |                          0.0000 |
| TPC-H 13       | 0.157237693667  | 32   | 0.158344879746  | 32   | +1%        |                          0.3682 |
| TPC-H 14       | 1.7351616621    | 60   | 1.76626610756   | 60   | +2%        |     (run time too short) 0.0002 |
| TPC-H 15       | 1.3223618269    | 60   | 1.32369995117   | 60   | +0%        |     (run time too short) 0.6888 |
| TPC-H 16       | 0.699989378452  | 60   | 0.726635456085  | 60   | +4%        |                          0.0000 |
| TPC-H 17       | 0.071922019124  | 15   | 0.0755068510771 | 16   | +5%        |                          0.0000 |
| TPC-H 18       | 0.147115588188  | 30   | 0.171241939068  | 35   | +16%       |                          0.0000 |
| TPC-H 19       | 0.483642876148  | 60   | 0.48078635335   | 60   | -1%        |                          0.5958 |
| TPC-H 20       | 0.146263256669  | 30   | 0.147148549557  | 30   | +1%        |                          0.0003 |
| TPC-H 21       | 0.0334587022662 | 7    | 0.0333742797375 | 7    | -0%        |        (not enough runs) 0.5848 |
| TPC-H 22       | 0.900039494038  | 60   | 0.896938681602  | 60   | -0%        |                          0.2104 |
| geometric mean |                 |      |                 |      | +5%        |                                 |
+----------------+-----------------+------+-----------------+------+------------+---------------------------------+
```

### DIFF
--- a/src/benchmarklib/table_generator.cpp
+++ b/src/benchmarklib/table_generator.cpp
@@ -184,7 +184,7 @@ std::shared_ptr<Table> TableGenerator::generate_table(
 
       // add full chunk to table
       if (column_index == num_columns - 1) {
-        table->append_chunk(segments, allocator_chunk);
+        table->append_chunk(segments, nullptr, allocator_chunk);
       }
     }
   }
--- a/src/benchmarklib/tpcc/tpcc_table_generator.cpp
+++ b/src/benchmarklib/tpcc/tpcc_table_generator.cpp
@@ -61,7 +61,8 @@ std::shared_ptr<Table> TpccTableGenerator::generate_items_table() {
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
   for (const auto& segments : segments_by_chunk) {
-    table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
+    const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+    table->append_chunk(segments, mvcc_data);
   }
 
   _encode_table("ITEM", table);
@@ -100,7 +101,8 @@ std::shared_ptr<Table> TpccTableGenerator::generate_warehouse_table() {
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
   for (const auto& segments : segments_by_chunk) {
-    table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
+    const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+    table->append_chunk(segments, mvcc_data);
   }
 
   _encode_table("WAREHOUSE", table);
@@ -152,7 +154,8 @@ std::shared_ptr<Table> TpccTableGenerator::generate_stock_table() {
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
   for (const auto& segments : segments_by_chunk) {
-    table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
+    const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+    table->append_chunk(segments, mvcc_data);
   }
 
   _encode_table("STOCK", table);
@@ -196,7 +199,8 @@ std::shared_ptr<Table> TpccTableGenerator::generate_district_table() {
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
   for (const auto& segments : segments_by_chunk) {
-    table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
+    const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+    table->append_chunk(segments, mvcc_data);
   }
 
   _encode_table("DISTRICT", table);
@@ -265,7 +269,8 @@ std::shared_ptr<Table> TpccTableGenerator::generate_customer_table() {
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
   for (const auto& segments : segments_by_chunk) {
-    table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
+    const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+    table->append_chunk(segments, mvcc_data);
   }
 
   _encode_table("CUSTOMER", table);
@@ -300,7 +305,8 @@ std::shared_ptr<Table> TpccTableGenerator::generate_history_table() {
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
   for (const auto& segments : segments_by_chunk) {
-    table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
+    const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+    table->append_chunk(segments, mvcc_data);
   }
 
   _encode_table("HISTORY", table);
@@ -347,7 +353,8 @@ std::shared_ptr<Table> TpccTableGenerator::generate_order_table(
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
   for (const auto& segments : segments_by_chunk) {
-    table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
+    const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+    table->append_chunk(segments, mvcc_data);
   }
 
   _encode_table("ORDER", table);
@@ -452,7 +459,8 @@ std::shared_ptr<Table> TpccTableGenerator::generate_order_line_table(
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
   for (const auto& segments : segments_by_chunk) {
-    table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
+    const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+    table->append_chunk(segments, mvcc_data);
   }
 
   _encode_table("ORDER_LINE", table);
@@ -480,7 +488,8 @@ std::shared_ptr<Table> TpccTableGenerator::generate_new_order_table() {
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
   for (const auto& segments : segments_by_chunk) {
-    table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
+    const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+    table->append_chunk(segments, mvcc_data);
   }
 
   _encode_table("NEW_ORDER", table);
--- a/src/benchmarklib/tpcc/tpcc_table_generator.cpp
+++ b/src/benchmarklib/tpcc/tpcc_table_generator.cpp
@@ -99,7 +99,9 @@ std::shared_ptr<Table> TpccTableGenerator::generate_warehouse_table() {
   });
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
-  for (const auto& segments : segments_by_chunk) table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
+  for (const auto& segments : segments_by_chunk) {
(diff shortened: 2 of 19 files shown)


Bouncner:
Any ideas what’s the issue with Q14?

mrzzzrm:
Yeah, https://github.com/hyrise/hyrise/pull/1647/commits/e9cfa7b722cf58a5a62c7ea7f8d97cf8985a2cdd fixed that. Re-running TPC-H right now.

mrzzzrm:
Benchmark results in OP updated. Solid improvements for such a small deed.

src/lib/storage/table.cpp
```diff
@@ -144,7 +151,14 @@ void Table::remove_chunk(ChunkID chunk_id) {
...
 }
 
-void Table::append_chunk(const Segments& segments, const std::optional<PolymorphicAllocator<Chunk>>& alloc) {
+void Table::append_chunk(const Segments& segments, std::shared_ptr<MvccData> mvcc_data,
+                         const std::optional<PolymorphicAllocator<Chunk>>& alloc) {
+  DebugAssert(_type != TableType::References || !mvcc_data,
+              "Setting explicit MvccData on a reference Table makes no sense. Reference Tables should use MvccData of "
+              "referenced Table, if any");
+  DebugAssert(_type != TableType::Data || static_cast<bool>(mvcc_data) == (_use_mvcc == UseMvcc::Yes),
+              "Supply MvccData to data Tables iff MVCC is enabled");
```
FabianWiebe:
```suggestion
              "Supply MvccData to data Tables if MVCC is enabled");
```

src/lib/storage/table.hpp
```diff
@@ -95,11 +95,13 @@ class Table : private Noncopyable {
...
    * Makes sure the @param segments match with the TableType (only ReferenceSegments or only data containing segments)
-   * En/Disables MVCC for the Chunk depending on whether MVCC is enabled for the table (has_mvcc())
    * This is a convenience method to enable automatically creating a chunk with correct settings given a set of segments.
-   * @param alloc
+   *
+   * @param mvcc_data   If the Table uses MVCC, this can optionally be passed in. If the Table uses MVCC and no
+   *                    mvcc_data is supplied, MvccData will be allocated.
    */
-  void append_chunk(const Segments& segments, const std::optional<PolymorphicAllocator<Chunk>>& alloc = std::nullopt);
+  void append_chunk(const Segments& segments, std::shared_ptr<MvccData> mvcc_data = {},
```
FabianWiebe:
```suggestion
  void append_chunk(const Segments& segments, std::shared_ptr<MvccData> mvcc_data = nullptr,
```
I think it is more intuitive

src/lib/storage/table.cpp
```diff
@@ -160,12 +174,11 @@ void Table::append_chunk(const Segments& segments, const std::optional<Polymorph
...
-#endif
 
-  std::shared_ptr<MvccData> mvcc_data;
+  DebugAssert(!mvcc_data || mvcc_data->size() == chunk_size, "Invalid MvccData size");
+#endif
 
-  if (_use_mvcc == UseMvcc::Yes) {
-    // append_chunk is a helper method and rows are made visible immediately
+  if (_use_mvcc == UseMvcc::Yes && !mvcc_data) {
     mvcc_data = std::make_shared<MvccData>(chunk_size, CommitID{0});
```
FabianWiebe: This line will only be executed in release as the 2nd `DebugAssert` will fail in debug.
I would either remove the 2nd `DebugAssert` or this line.

src/lib/storage/table.hpp
```diff
@@ -95,11 +95,13 @@ class Table : private Noncopyable {
   /**
    * Creates a new Chunk and appends it to this table.
    * Makes sure the @param segments match with the TableType (only ReferenceSegments or only data containing segments)
-   * En/Disables MVCC for the Chunk depending on whether MVCC is enabled for the table (has_mvcc())
    * This is a convenience method to enable automatically creating a chunk with correct settings given a set of segments.
-   * @param alloc
+   *
+   * @param mvcc_data   If the Table uses MVCC, this can optionally be passed in. If the Table uses MVCC and no
+   *                    mvcc_data is supplied, MvccData will be allocated.
```
FabianWiebe: > If the Table uses MVCC and no mvcc_data is supplied, MvccData will be allocated.

No, it will not in Debug, it will fail. See `table.cpp`:
````
DebugAssert(_type != TableType::Data || static_cast<bool>(mvcc_data) == (_use_mvcc == UseMvcc::Yes)
````
I would prefer to get rid of the `DebugAssert`.

File: src/lib/storage/table.cpp [hunk 5]
```diff
@@ -12,6 +12,7 @@
 #include "statistics/table_statistics.hpp"
 #include "types.hpp"
 #include "utils/assert.hpp"
+#include "utils/timer.hpp"
```
FabianWiebe: The include should not be necessary for the changed lines.

src/lib/operators/abstract_operator.cpp
```diff
@@ -60,22 +60,7 @@ void AbstractOperator::execute() {
...
-        if (!_output) return true;
-        if (_output->chunk_count() <= ChunkID{1}) return true;
-        for (auto chunk_id = ChunkID{0}; chunk_id < _output->chunk_count(); ++chunk_id) {
-          if (_output->get_chunk(chunk_id)->size() < 1) return true;
-        }
-        return true;
-      }(),
-      "Empty chunk returned from operator " + description());
-
-  DebugAssert(!_output || _output->column_count() > 0, "Operator " + description() + " did not output any columns");
```
FabianWiebe: The first `DebugAssert` is always true and can be removed.
But the second `DebugAssert` could be false and should also be removed?

src/lib/storage/table.cpp
mrzzzrm: [That](https://en.wikipedia.org/wiki/If_and_only_if)'s actually intended:
Data Table with MVCC enabled -> You **must** supply MvccData
Data Table without MVCC enabed -> You **cannot** supply MvccData

src/lib/storage/table.hpp
mrzzzrm: It's actually the comment that was wrong. Updated to `@param mvcc_data   Has to be passed in iff the Table is a data Table which uses MVCC`

src/lib/operators/abstract_operator.cpp
mrzzzrm: I removed these DebugAsserts because a) they were showing up in Debug build perf and b) because in #1631 @mrks and I seemed to agree that a Table without Columns or Rows isn't actually an invalid state. 

To keep this PR on-topic (i.e. release build perf) I put those two DebugAsserts back in for now.

src/lib/storage/table.cpp
mrzzzrm: this line it is, implicit allocations of significant amounts of memory should not happen and were what lead to the perf impact fixed by this PR.

src/benchmarklib/table_generator.cpp
```diff
@@ -184,7 +184,7 @@ std::shared_ptr<Table> TableGenerator::generate_table(
 
       // add full chunk to table
       if (column_index == num_columns - 1) {
-        table->append_chunk(segments, allocator_chunk);
+        table->append_chunk(segments, nullptr, allocator_chunk);
```
mrks: Admittedly spiteful comment incoming:
How is nullptr here any better than true or false? Right now, I have no idea what I am looking at.

src/benchmarklib/tpcc/tpcc_table_generator.cpp
```diff
@@ -60,7 +60,9 @@ std::shared_ptr<Table> TpccTableGenerator::generate_items_table() {
       });
 
   auto table = std::make_shared<Table>(column_definitions, TableType::Data, _chunk_size, UseMvcc::Yes);
-  for (const auto& segment : segments_by_chunk) table->append_chunk(segment);
+  for (const auto& segments : segments_by_chunk) {
+    table->append_chunk(segments, std::make_shared<MvccData>(segments.front()->size(), CommitID{0}));
```
mrks: Split this into two rows? For me, it's two different things happening.

src/lib/import_export/csv_parser.cpp
```diff
@@ -84,7 +84,9 @@ std::shared_ptr<Table> CsvParser::parse(const std::string& filename, const std::
   CurrentScheduler::wait_for_tasks(tasks);
 
   for (auto& segments : segments_by_chunks) {
-    table->append_chunk(segments);
+    DebugAssert(!segments.empty(), "Empty chunks shouldn't occur when importing CSV");
```
mrks: What happens if I import an empty CSV?

src/lib/operators/projection.cpp
```diff
@@ -37,16 +38,20 @@ void Projection::_on_set_transaction_context(const std::weak_ptr<TransactionCont
 }
 
 std::shared_ptr<const Table> Projection::_on_execute() {
+  Timer t;
```
mrks: ?

src/lib/storage/table.hpp
```diff
@@ -95,11 +95,12 @@ class Table : private Noncopyable {
   /**
    * Creates a new Chunk and appends it to this table.
    * Makes sure the @param segments match with the TableType (only ReferenceSegments or only data containing segments)
-   * En/Disables MVCC for the Chunk depending on whether MVCC is enabled for the table (has_mvcc())
    * This is a convenience method to enable automatically creating a chunk with correct settings given a set of segments.
-   * @param alloc
+   *
+   * @param mvcc_data   Has to be passed in iff the Table is a data Table which uses MVCC
```
mrks: that

src/lib/import_export/csv_parser.cpp
FabianWiebe: This only checks whether a `Chunk` has any columns. So with this it is not possible to import a table without columns.

src/lib/operators/abstract_operator.cpp
FabianWiebe: If it makes sense, keep it in. I was just surprised.

src/lib/storage/table.cpp
FabianWiebe: Sounds good. Always learning something new. ;)

src/benchmarklib/table_generator.cpp
mrzzzrm: > Right now, I have no idea what I am looking at.

Exactly, you do not - and I will quote this during the next bool-vs-enum debate. ;)

nullptr is not categorically better than an unnamed bool, and I wouldn't claim it is. I had to use it here, because there are no real alternatives. `std::shared_ptr<MvccData>{}`, imo, doesn't read as "don't supply MvccData". Changing the parameter order of the allocator and the MvccData wouldn't help, because there are places where MvccData is supplied, but not an allocator, which would result in a call like `...->append_chunk(segments, std::nullopt, mvcc_data` - which isn't cleaner than the current state.

Two different function signature `append_chunk_with_mvcc_data()` and `append_chunk_without_mvcc_data()` would be a possibility, but then `Projection` would need a branching to decide which of these it calls. Also not better, imo.

src/lib/import_export/csv_parser.cpp
mrzzzrm: `TEST_F(OperatorsImportCsvTest, NoRows)` verifies that it is valid to import a CSV without rows - it just won't have any chunks.

src/lib/operators/abstract_operator.cpp
mrzzzrm: Removed them again to be able to import a truly empty CSV

PR #1650, opened 2019-04-28
Use ccache for jemalloc

### LINKED ISSUE #1641: jemalloc build does not use ccache
jemalloc is built using `./configure; make` (see third_party/CMakeLists.txt). We should make it use ccache if `CMAKE_CXX_COMPILER_LAUNCHER` is set. That saves us around three minutes per CI build.

mrks:
fixes #1641, reduces the gcc-debug build time to 55s if the cache is fully filled.


Jenkinsfile
```diff
@@ -240,8 +240,8 @@ try {
 
           // We do not use install.sh here as there is no way to run OS X in a Docker container
           sh "git submodule update --init --recursive --jobs 4"
-
-          sh "mkdir clang-debug && cd clang-debug && /usr/local/bin/cmake -DCMAKE_CXX_COMPILER_LAUNCHER=/usr/local/bin/ccache -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=/usr/local/Cellar/llvm\\@7/7.0.1/bin/clang -DCMAKE_CXX_COMPILER=/usr/local/Cellar/llvm\\@7/7.0.1/bin/clang++ .."
+          sh "env"
```
lanice: Just out of curiosity, is this just a print for debugging, or does this have any other purpose?

Jenkinsfile
```diff
@@ -263,6 +263,9 @@ try {
         }
       } else {
         githubNotify context: 'CI Pipeline', status: 'FAILURE'
+        if (env.BRANCH_NAME == 'master' || full_ci) {
+          githubNotify context: 'Full CI', status: 'FAILURE'
+        }
```
lanice: This has nothing directly to do with the jemalloc change I assume, it's just added in here for convenience?

mrks: Not for convenience. Previously, the "Full CI passed" status was never reset. Not a problem, because the other one failed, but still confusing.

Jenkinsfile
mrks: Will remove.

PR #1651, opened 2019-04-29
Hyrise Console Test

### LINKED ISSUE #1103: Tests for the console

engelfa:
This is a minimal PR for Issue #1103.

Testing the visualize command seems to be more complex because the emulated terminal has difficulties executing is_iterm2.sh and gets stuck.
I am still looking for a solution.


mrks:
On first sight it looks good. What happens if one of the assertions is not met? With the original expect, it would not fail properly.

engelfa:
An exception is thrown and 1 is returned. You said that's enough for Jenkins.
Could you install pexpect on the build server? I updated the requirements.txt.

mrks:
Simply add it to install.sh and Dockerfile. Once this PR is merged, we can update the cached docker image.

mrks:
Also, what's the general idea of this PR. You say that it "is a minimal PR", but at the same time you plan to close #1103?

engelfa:
The test is minimal right now but will grow with the next commits and the PR can be merged when the test is exhaustive enough.

scripts/test/console_output_test.py
```diff
@@ -0,0 +1,42 @@
...
+	if len(sys.argv) == 1:
+		print ("Usage: ./scripts/test/console_output_test.py <console_path>")
+		sys.exit(1)
+
+	if not os.path.isdir("resources/test_data/tbl"):
+		print ("Cannot find resources/test_data/tbl. Are you running the test suite from the main folder of the Hyrise repository?")
+		sys.exit(1)
+
+	console_path = sys.argv[1]	
+	console = pexpect.spawn(console_path + "/hyriseConsole", timeout=15, dimensions=(1920,1080))
```
mrzzzrm: "dimensions" refers to columns/rows - 1920x1080 is a bit much here, my terminal has a default size of 200x64, e.g.. I'd change this to something more realistic.

scripts/test/console_output_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+	# Test print command
+	console.sendline("print test")
+	console.expect("Exception thrown while loading table:")
+	
+	# Test load command
+	console.sendline("load resources/test_data/tbl/10_ints.tbl test")
+	console.expect('Loading .*tbl/10_ints.tbl into table "test"')
+	console.expect('Encoding "test" using Unencoded')
```
mrzzzrm: assume the next line that the console would write would be "Error applying encoding"  - or something to that extend. Would the test script still succeed, or would it fail?

scripts/test/console_output_test.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/python
```
mrzzzrm: I'd just call this the "console_test.py" or "hyriseConsole_test.py" - every "test" tests "output", that info is redundant or even wrong when you consider that you also supply "input" to the console and check whether that works.

scripts/test/console_output_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+import os
+import sys
+import pexpect
+
+def main():
+	if len(sys.argv) == 1:
+		print ("Usage: ./scripts/test/console_output_test.py <console_path>")
```
mrzzzrm: "<console_path>" makes me think "cmake-build-debug/hyriseTest" instead of just "cmake-build-debug".

Call this the "<build_dir>" to make this more clear.

scripts/test/console_output_test.py
```diff
@@ -0,0 +1,42 @@
...
+	console.sendline("select sum(a) from test")
+	console.expect("786")
+	console.expect("1 rows total")
+
+	# Test TPCH generation
+	console.sendline("generate_tpch 0.001")
+	console.expect("Generating tables done")
+
+	# Test TPCH tables
+	console.sendline("select * from nation")
```
mrzzzrm: what if the console would crash here, would this make the script fail as well?

scripts/test/console_output_test.py
engelfa: No, there is no method, which throws an exception, when a pattern is matched (something like not_expect()). If the encoding fails, a different message instead of ''Encoding "test" using Unencoded' has to be returned.

mrks:
> Testing the visualize command seems to be more complex because the emulated terminal has difficulties executing is_iterm2.sh and gets stuck.

Good chance that the shell hacking done there doesn't work with expect. Just set an environment variable HYRISE_DISABLE_ITERM_CHECK (naming tbd) and immediately return false in the`.sh` if the variable is set.

scripts/test/hyriseConsole_test.py
```diff
@@ -0,0 +1,86 @@
...
+	console.sendline("visualize")
+
+	# The emulated terminal freezes because of is_iterm2.sh in visualize command.
+	# Therefore, we don't exit cleanly.
+	close_console(console)
+
+	while not os.path.exists('.pqp.dot'):
+		time.sleep(1)
+
+	with open('.pqp.dot') as f: s = f.read()
```
mrks: Can we check that the dot file is not empty?

mrks: ah, nevermind.

mrzzzrm:
What are the blockers here? If it is just is_term2.sh problems, I wouldn't bother with testing `visualize` just now.
Let's get this PR in, quickly followed up by basic tests for the benchmarks and keep the hard challenges for later.

PR #1653, opened 2019-04-30
Remove Optimizer::remove_rules_of_type() + SubqueryToJoinBenchmark

### LINKED ISSUE #1652: What is Optimizer::remove_rules_of_type good for?
As there is no proper way of re-adding them at the correct place in the chain this feels like a unusable stub.

mrzzzrm:
Fix #1652 - I'm a bit embarrassed that I didn't flag this during review.

Also remove the SubqueryToJoinBenchmark (which used this function) since I don't feel we need a benchmark for every optimizer rule. That's just dead weight to carry around


mrzzzrm:
It wasn't used anywhere except for that benchmark. If removing rules every becomes a necessity, we can just make `Optimizer::_rules` public.

> (I have to admit that I am surprise that it has made it into the master)
I didn't want to flag benchmarks during reviews as to not to appear too pedantic. But I think it's fine to remove them/store them outside the main repo for later resurrection

PR #1656, opened 2019-05-02
Verify query plans in console

d-justen:
...as suggested in #1189 and #1636
Tests could be provided in or after #1651


mrks:
Nice. Not a review yet, but "verify" is already used for sqlite verification. No better idea yet.

mrzzzrm:
just "sanity check"?

mrks:
Also, will `create table t  ...; create view v as select * from t` still work?

src/bin/console/console.cpp
```diff
@@ -1078,6 +1086,58 @@ bool Console::_handle_rollback() {
   return false;
 }
 
+bool Console::_verify_plan(const std::vector<std::shared_ptr<AbstractLQPNode>>& lqp) {
```
mrks: Why is this in the console? I'd expect all SQL interfaces to behave alike.

Moving this to the pipeline should also make it possible to do the check on a per-statement basis.

mrzzzrm: And if the function were in the Pipeline, how would Hyrise behave then? What would the pipeine do differently than what it does now, which would be: Shut Hyrise down if the check failed? 

This is rightly placed in the console imo, because the console is the only place that we want to survive such input errors in, right?

src/bin/console/console.cpp
```diff
@@ -1078,6 +1086,58 @@ bool Console::_handle_rollback() {
...
+  const auto& sm = StorageManager::get();
```
mrzzzrm: please don't use abbreviations in newly written code. I know some old tests do this but newly written code shouldn't

src/bin/console/console.cpp
mrzzzrm: `lqp` -> `lqps/plans`

src/bin/console/console.cpp
```diff
@@ -1078,6 +1086,58 @@ bool Console::_handle_rollback() {
...
+
+  for (const auto& node : lqp) {
```
mrzzzrm: Please add a short comment why you're only looking at the root nodes here ("Create/Drop can only appear at the root")

src/bin/console/console.hpp
```diff
@@ -143,6 +143,9 @@ class Console : public Singleton<Console> {
   static char* _command_generator_setting(const char* text, int state);
   static char* _command_generator_setting_scheduler(const char* text, int state);
 
+  // Checks whether a query plan can be executed without failures
+  bool _verify_plan(const std::vector<std::shared_ptr<AbstractLQPNode>>& lqp);
```
mrzzzrm: can be `bool _verify_plan(const std::vector<std::shared_ptr<AbstractLQPNode>>& lqp) const?;` or static?

src/bin/console/console.cpp
mrks: Use AssertInput and throw the exception that we are allowed to catch - same as syntactically incorrect SQL. The server should be able to handle failed AssertInputs just as well.

mrzzzrm: Ah right, the server. Good, @d-justen, then we introduce `SQLPipelineStatement::_verify_lqp(...)` which performs AssertInput() checks - and call it from `get_unoptimized_plan()`

mrks: Please keep the other open question (multi-statement queries) in mind. Those are ugly, but we shouldn’t break them.

d-justen: @mrks By moving the method to `SQLPipelineStatement` it doesn't break multi-statement queries since every statement is verified one after another

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -186,6 +192,8 @@ const std::shared_ptr<const Table>& SQLPipelineStatement::get_result_table() {
     return _result_table;
   }
 
+  _verify_lqp();
```
mrzzzrm: why do this here? why not do this exactly where the LQP was created, i.e., `get_unoptimized_logical_plan()`. If there is a good reason for this being called here, please doc it.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -217,4 +225,47 @@ const std::shared_ptr<TransactionContext>& SQLPipelineStatement::transaction_con
 }
 
 const std::shared_ptr<SQLPipelineStatementMetrics>& SQLPipelineStatement::metrics() const { return _metrics; }
+
+void SQLPipelineStatement::_verify_lqp() {
+  if (!_unoptimized_logical_plan) {
```
mrzzzrm: a "verification" function should never issue significant work like SQL parsing/translation is. Calling `get_unoptimized_logical_plan();` here should not be necessary if _verify_lqp() was called from `get_unoptimized_logical_plan()` itself

src/lib/sql/sql_pipeline_statement.hpp
```diff
@@ -72,6 +72,9 @@ class SQLPipelineStatement : public Noncopyable {
   const std::shared_ptr<SQLPipelineStatementMetrics>& metrics() const;
 
  private:
+  // Checks whether a query plan can be executed without failures
```
mrzzzrm: That wording is too vague, imo. What this function actually does is basic sanity checking before DCL (http://www.datenbanken-verstehen.de/datenbank-grundlagen/dbms/dml-ddl-dcl-kommandos/) operators are executed that would predictably fail - and give the user the option to catch InvalidInputExceptions.

src/test/sql/sql_pipeline_statement_test.cpp
```diff
@@ -91,6 +91,10 @@ class SQLPipelineStatementTest : public BaseTest {
   const std::string _multi_statement_query = "INSERT INTO table_a VALUES (11, 11.11); SELECT * FROM table_a";
   const std::string _multi_statement_dependant = "CREATE VIEW foo AS SELECT * FROM table_a; SELECT * FROM foo;";
 
+  const std::string _create_table = "CREATE TABLE t (a_int INTEGER)";
```
mrzzzrm: these SQL strings "belong" to `TEST_F(SQLPipelineStatementTest, VerifyPlan) ` and the readability of that test case would increase if they were just inlined there. No need to save those strings in variables.

src/lib/sql/sql_pipeline_statement.hpp
d-justen: I get your point, what about this:

> Performs a sanity check in order to prevent an execution of a predictably failing DDL operator. Throws an InvalidInputException if an invalid LQP is detected.

mrzzzrm: +"DDL operator (e.g. creating a table that already exists")

then, marvelous

src/lib/sql/sql_pipeline_statement.cpp
d-justen: If this is verified in `get_unoptimized_logical_plan()` (like it is now) the verification doesn't take place when sending `CREATE TABLE t (...)` via the console twice. Then there is no new unoptimized LQP generated, because the PQP is already cached from the first execution.

mrzzzrm: Yep, that sounds like a valid reason to put "_verify_lqp()" here. Please add a clarifying comment.

mrzzzrm:
I changed the functions name to `_precheck_ddl_operators` and made it look at operators instead of lqp nodes - so now the check can be done in `get_physical_plan` and no redundant lqp needs to be created if the pqp is already cached.

Bouncner:
Does that mean you fixed all your mentioned issues, @mrzzzrm?

mrzzzrm:
yes, but I wouldn't approve what are essentially my own changes. @d-justen could you confirm everything is alright here?

mrks:
@d-justen Feel free to press the button if you agree with the changes.

Issue #1286, opened 2018-11-13
Unify CompressedVectorType and VectorCompressionType


Bouncner:
We use both and they apparently mean the same thing.

Bouncner:
Oh, they actually differ (see comment in `compressed_vector_type.hpp`). But maybe we can come up with better names for both that can be better distinguished and transfer more information.

Bouncner:
@mrks just recommended `VectorCompressionAlgorithm` and `VectorCompressionType`. Since they are exactly the same for SimdBp128 and just differ in the actual byte-width for FSBA, I am not sure if those names are ideal. Anyhow, they are much much better than the current ones.

mrks:
Another suggestion: *Type and *Subtype

Bensk1:
As soon as this gets resolved, information about vector compression should be included in our meta tables (`MetaTableManager ::generate_segments_table()`).

PR #1657, opened 2019-05-03
Enable re-encoding of segments

### LINKED ISSUE #1368: Support Segment Re-Encoding
After #1356 it will become annoying that we cannot create a, e.g., FrameOfReference encoded segment from a DictionarySegment. Should be simple to implement, BaseSegment should probably receive a virtual getter for the memory resource it is associated with and that's it.

Bouncner:
This PR introduces re-encoding to chunks/segments. Instead of passing a `ValueSegment` to the encoders, an `AnySegmentIterable` is passed.
Statistics are only built once.

__What has been enabled by this PR:__
* Re-encoding from an encoded segment to an `Unencoded` one is now possible.
* Loading binary files (e.g., from dictionary segment exports) and re-encoding them  (e.g., `-e LZ4`) is now possible.

__Performance:__
Performance got slightly faster (except for std::strings), but that's barely noticeable for most users. Fixed string encoding is much faster as one loop less over the data is performed.
```
----------------------------------------------------------------------
Benchmark                               Time           CPU Iterations
----------------------------------------------------------------------
BM_EncodingPerf_Integer_old      15627849 ns   15627568 ns        896
BM_EncodingPerf_Integer_new      14100860 ns   14100774 ns        992
BM_EncodingPerf_String_old      227411133 ns  227389640 ns         47
BM_EncodingPerf_String_new      254255025 ns  254225617 ns         60
BM_EncodingPerf_FSString_old    725246251 ns  725228876 ns         20
BM_EncodingPerf_FSString_new    247519962 ns  247511163 ns         58
BM_EncodingPerf_LZ4Integer_old   15705026 ns   15704350 ns        888
BM_EncodingPerf_LZ4Integer_new   14179989 ns   14179962 ns        988
BM_EncodingPerf_FoRInteger_old   15624515 ns   15624161 ns        897
BM_EncodingPerf_FoRInteger_new   14124000 ns   14123530 ns        994
```

Fixes #1368.


src/lib/storage/lz4_segment/lz4_encoder.hpp
```diff
@@ -168,35 +165,38 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
-    auto sample_sizes = pmr_vector<size_t>{alloc};
-    sample_sizes.resize(num_elements);
+    auto sample_sizes = pmr_vector<size_t>{allocator};
+
+    segment_iterable.with_iterators([&](auto it, auto end) {
+      const auto segment_size = std::distance(it, end);
+
+      null_values.resize(segment_size);
+      offsets.resize(segment_size);
+      sample_sizes.resize(segment_size);
```
mrks: Legacy, but maybe `string_length_samples` is better?

src/lib/storage/mvcc_data.cpp
```diff
@@ -26,8 +26,12 @@ void MvccData::shrink() {
   tids.shrink_to_fit();
   begin_cids.shrink_to_fit();
   end_cids.shrink_to_fit();
+
+  _is_shrunk = true;
 }
 
+bool MvccData::is_shrunk() { return _is_shrunk; }
```
mrks: Did you check if shrinking has any cost on a vector that is already shrunk?

src/lib/storage/segment_encoding_utils.cpp
```diff
@@ -52,4 +51,22 @@ std::shared_ptr<BaseEncodedSegment> encode_segment(EncodingType encoding_type, D
...
+ * @brief Returns the vector compression type for a given compressed vector type.
+ *
+ * For the difference of the two, please take a look at compressed_vector_type.hpp.
+ */
+VectorCompressionType parent_vector_compression_type(const CompressedVectorType compressed_vector_type) {
+  switch (compressed_vector_type) {
+    case CompressedVectorType::FixedSize4ByteAligned:
+    case CompressedVectorType::FixedSize2ByteAligned:
+    case CompressedVectorType::FixedSize1ByteAligned:
+      return VectorCompressionType::FixedSizeByteAligned;
```
mrks: Uargh. Introducing "parent types" makes #1286 even more necessary. `VectorCompressionAlgorithm` and `VectorCompressionType`?

src/test/storage/chunk_encoder_test.cpp
```diff
@@ -146,4 +151,26 @@ TEST_F(ChunkEncoderTest, EncodeMultipleChunksUsingSameEncoding) {
   verify_encoding(_table->get_chunk(ChunkID{1u}), unencoded_chunk_spec);
 }
 
+TEST_F(ChunkEncoderTest, ReencodingTable) {
+  const auto chunk_encoding_specs =
```
mrks: Comment on what this holds?

src/test/storage/encoded_segment_test.cpp
```diff
@@ -171,7 +185,7 @@ TEST_P(EncodedSegmentTest, SequentiallyReadNotNullableIntSegment) {
 
 TEST_P(EncodedSegmentTest, SequentiallyReadNullableIntSegment) {
   auto value_segment = create_int_with_null_value_segment();
-  auto base_encoded_segment = encode_value_segment(DataType::Int, value_segment);
+  auto base_encoded_segment = encode_base_segment(DataType::Int, value_segment);
```
mrks: Just "encode_segment"?

src/test/storage/encoded_segment_test.cpp
```diff
@@ -271,4 +285,55 @@ TEST_P(EncodedSegmentTest, SequentiallyReadEmptyIntSegment) {
...
+      auto value_segment_iterable = create_iterable_from_segment(*verification_segment);
+      auto encoded_segment_iterable = create_iterable_from_segment(encoded_segment);
+
+      value_segment_iterable.with_iterators([&](auto value_segment_it, auto value_segment_end) {
+        encoded_segment_iterable.with_iterators([&](auto encoded_segment_it, auto encoded_segment_end) {
+          for (; encoded_segment_it != encoded_segment_end; ++encoded_segment_it, ++value_segment_it) {
+            EXPECT_EQ(value_segment_it->is_null(), encoded_segment_it->is_null());
+
+            if (!value_segment_it->is_null()) {
+              EXPECT_EQ(value_segment_it->value(), encoded_segment_it->value());
```
mrks: Simply add the segments to tables and check for their equality?

src/test/storage/encoded_segment_test.cpp
```diff
@@ -271,4 +285,55 @@ TEST_P(EncodedSegmentTest, SequentiallyReadEmptyIntSegment) {
...
+            }
+          }
+        });
+      });
+    });
+  };
+
+  // Use the row_count used for frame of reference segments.
+  auto value_segment = create_int_with_null_value_segment(row_count(EncodingType::FrameOfReference));
```
mrks: Please use a paremeterized test instead.

src/test/storage/encoded_string_segment_test.cpp
```diff
@@ -294,4 +302,54 @@ TEST_P(EncodedStringSegmentTest, SequentiallyReadNullableStringSegmentWithShuffl
   });
 }
 
+TEST_F(EncodedStringSegmentTest, SegmentReencoding) {
```
mrks: Same as the non-strings

src/lib/storage/base_segment_encoder.hpp
```diff
@@ -131,14 +131,17 @@ class SegmentEncoder : public BaseSegmentEncoder {
...
   template <typename ColumnDataType>
-  std::shared_ptr<BaseEncodedSegment> encode(const std::shared_ptr<const BaseValueSegment>& base_value_segment,
+  std::shared_ptr<BaseEncodedSegment> encode(const std::shared_ptr<const BaseSegment>& base_segment,
                                              hana::basic_type<ColumnDataType> data_type_c) {
     static_assert(decltype(supports(data_type_c))::value);
+    const auto iterable = create_any_segment_iterable<ColumnDataType>(*base_segment);
 
-    const auto value_segment = std::dynamic_pointer_cast<const ValueSegment<ColumnDataType>>(base_value_segment);
-    Assert(value_segment, "Value segment must have passed data type.");
+    // Obtain allocator in case of value segment.
```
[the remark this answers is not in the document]
Bouncner: Changed to
```
    // Obtain allocator of segment to encode. Currently, only ValueSegment store the used allocator.
    // TODO(anyone): obtain allocator for all types of segment, when available.
```

Ok?

src/lib/storage/chunk_encoder.cpp
```diff
@@ -12,41 +12,104 @@
...
+           encoding_spec.vector_compression_type ==
+               parent_vector_compression_type(*encoded_segment->compressed_vector_type()))) {
+        return_segment = segment;
+        return;
+      }
+    }
+
+    const auto unencoded_segment = std::dynamic_pointer_cast<const ValueSegment<ColumnDataType>>(segment);
+    if (unencoded_segment && encoding_spec.encoding_type == EncodingType::Unencoded) {
+      return_segment = segment;
```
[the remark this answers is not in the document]
Bouncner: Renamed to `result`.

src/lib/storage/chunk_encoder.cpp
```diff
@@ -12,41 +12,104 @@
...
+      if (!encoding_spec.vector_compression_type ||
+          (encoding_spec.vector_compression_type && encoded_segment->compressed_vector_type() &&
+           encoding_spec.vector_compression_type ==
+               parent_vector_compression_type(*encoded_segment->compressed_vector_type()))) {
+        return_segment = segment;
+        return;
+      }
+    }
+
+    const auto unencoded_segment = std::dynamic_pointer_cast<const ValueSegment<ColumnDataType>>(segment);
```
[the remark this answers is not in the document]
Bouncner: I am checking whether the given BaseSegment is of type ValueSegment. Is that wrong? Using `resolve_*` would unnecessarily increase compile times, wouldn't it?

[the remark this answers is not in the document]
mrks: Not this line in particular, but the block started here.

src/lib/storage/base_segment_encoder.hpp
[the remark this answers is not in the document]
mrks: yepp

src/test/storage/chunk_encoder_test.cpp
Bouncner: Added `// Encoding specifications which will be applied one after another to the chunk.`

src/test/storage/encoded_segment_test.cpp
Bouncner: Not sure what you mean. Do you mean using `GetParam()` and `TEST_P`?

mrks: yes

src/test/storage/encoded_segment_test.cpp
Bouncner: If I remember correctly, table equality checks are rather expensive.

src/test/storage/encoded_segment_test.cpp
Bouncner: Did that, but for some reason clang does no longer find the eventual `encode_segment` method in the utils, even though the signature differ. Might be related to the other problem where I use `::opossum::`?

Any ideas what I am doing wrong?

src/lib/storage/chunk_encoder.cpp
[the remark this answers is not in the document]
Bouncner: Will add a comment. But this cast to check for the type shouldn't create/copy any data, right?

src/test/storage/encoded_segment_test.cpp
Bouncner: The tests are parameterized to take a encoding configuration and test this configuration. But the purpose of the reencoding test is to reencode between multiple configurations and check that actual data remains the same.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
```diff
@@ -28,20 +29,46 @@ class DictionaryEncoder : public SegmentEncoder<DictionaryEncoder<Encoding>> {
...
+    segment_iterable.with_iterators([&](auto segment_it, const auto segment_end) {
+      const auto segment_size = std::distance(segment_it, segment_end);
+      values.reserve(segment_size);
+      null_values.resize(segment_size);
+
+      for (auto current_position = size_t{0}; segment_it != segment_end; ++segment_it, ++current_position) {
+        const auto segment_item = *segment_it;
+        if (!segment_item.is_null()) {
+          const auto segment_value = segment_item.value();
+          values.push_back(segment_value);
```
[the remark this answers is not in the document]
Bouncner: Added comments.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
```diff
@@ -28,20 +29,46 @@ class DictionaryEncoder : public SegmentEncoder<DictionaryEncoder<Encoding>> {
...
-    // See: https://goo.gl/MCM5rr
-    // Create dictionary (enforce uniqueness and sorting)
-    const auto& values = value_segment->values();
+  std::shared_ptr<BaseEncodedSegment> _on_encode(const AnySegmentIterable<T> segment_iterable,
+                                                 const PolymorphicAllocator<T>& allocator) {
+    std::vector<T> values;
+    std::vector<bool> null_values;
+
+    auto max_string_length = size_t{0};
+    std::set<T> unique_values;
```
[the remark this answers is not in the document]
Bouncner: I am currently looking why clang fails (it does not on my Mac or nemea). After that, I will reevaluate with GCC.

But my current assumption is that obtaining values via an erased iterator hides the overhead of adding the values to the map. After that, obtaining unique elements in a sorted order is fast using the map.
That's very different from previous measurements were the input was a std::vector directly holding the values in a consecutive chunk of memory.

src/lib/storage/lz4_segment/lz4_encoder.hpp
```diff
@@ -168,35 +165,38 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
      * The offsets are stored as 32 bit unsigned integer as opposed to 64 bit (size_t) so that they can later be
      * compressed via vector compression.
      */
-    auto offsets = pmr_vector<uint32_t>{alloc};
-    offsets.resize(num_elements);
+    auto offsets = pmr_vector<uint32_t>{allocator};
```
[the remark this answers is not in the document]
Bouncner: Shouldn't change anything. Binary caching still only works for values and dictionary segments, since the current benchmark do only support an initial encoding configuration, re-encoding doesn't happen anyways.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
```diff
@@ -53,87 +80,41 @@ class DictionaryEncoder : public SegmentEncoder<DictionaryEncoder<Encoding>> {
...
     // We need to increment the dictionary size here because of possible null values.
     const auto max_value = dictionary.size() + 1u;
 
-    auto encoded_attribute_vector = compress_vector(
-        attribute_vector, SegmentEncoder<DictionaryEncoder<Encoding>>::vector_compression_type(), alloc, {max_value});
-    auto dictionary_sptr = std::allocate_shared<U>(alloc, std::move(dictionary));
+    auto encoded_attribute_vector =
+        compress_vector(attribute_vector, SegmentEncoder<DictionaryEncoder<Encoding>>::vector_compression_type(),
+                        allocator, {max_value});
+    auto dictionary_sptr = std::allocate_shared<U>(allocator, std::move(dictionary));
```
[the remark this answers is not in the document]
Bouncner: Any recommendation.

src/test/storage/encoded_segment_test.cpp
mrzzzrm: just use check_table_equals, it produces much more readable test failures.

src/lib/storage/chunk_encoder.cpp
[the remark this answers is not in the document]
mrks: No.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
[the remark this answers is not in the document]
mrks: Why can’t we allocate with the correct allocator in the first place?

src/lib/storage/lz4_segment/lz4_encoder.hpp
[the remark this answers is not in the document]
mrks: What if the stored binary table is different from the requested one?

src/test/storage/encoded_segment_test.cpp
mrks: Hard to check without access to a compiler. Not a.big thing anyway.

src/test/storage/encoded_segment_test.cpp
mrks: Ok

src/lib/storage/chunk_encoder.cpp
```diff
@@ -12,41 +12,104 @@
 #include "statistics/chunk_statistics/segment_statistics.hpp"
 #include "storage/base_encoded_segment.hpp"
 #include "storage/segment_encoding_utils.hpp"
+#include "storage/segment_iterables/any_segment_iterable.hpp"
+#include "storage/value_segment.hpp"
 #include "utils/assert.hpp"
 
 namespace opossum {
 
+std::shared_ptr<BaseSegment> ChunkEncoder::encode_segment(const std::shared_ptr<BaseSegment>& segment,
```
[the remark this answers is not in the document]
Bouncner: Added a check and a fail.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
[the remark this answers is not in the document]
Bouncner: I never really tried to understand the the meaning of this allocate_shared and thought is allocation magic. I will take a look at it.

src/lib/storage/lz4_segment/lz4_encoder.hpp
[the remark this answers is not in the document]
Bouncner: Ahh, I see your point. That should work just fine now. I will check if there are some quirks.

src/lib/storage/lz4_segment/lz4_encoder.hpp
Bouncner: Changed to `string_samples_lengths`.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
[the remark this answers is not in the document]
Bouncner: Apparently the reason is that shared_pointers are passed to the dictionary segment's constructor. I do not see why shared_pointers are passed to be honest. Do you know?

src/lib/storage/mvcc_data.cpp
Bouncner: On a 100'000 size_t vector, shrink() takes ~2000ns and the following shrink()'s take <200ns.
So there is an overhead, but probably not enough. Will remove it.

Code:
```
BENCHMARK_F(MicroBenchmarkBasicFixture, TBBShrink)(benchmark::State& state) {
  tbb::concurrent_vector<size_t> values;

  for (auto i = size_t{0}; i < 100'000; ++i) {
    values.push_back(i);
  }

  auto start_time = std::chrono::high_resolution_clock::now();
  values.shrink_to_fit();
  auto end_time = std::chrono::high_resolution_clock::now();
  std::cout << std::chrono::duration_cast<std::chrono::nanoseconds>(end_time - start_time).count() << std::endl;;

  for (auto i = size_t{0}; i < 5; ++i) {
    start_time = std::chrono::high_resolution_clock::now();
    values.shrink_to_fit();
    end_time = std::chrono::high_resolution_clock::now();
    std::cout << std::chrono::duration_cast<std::chrono::nanoseconds>(end_time - start_time).count() << std::endl;;
  }
}
```

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
```diff
@@ -28,20 +29,46 @@ class DictionaryEncoder : public SegmentEncoder<DictionaryEncoder<Encoding>> {
...
+          if constexpr (Encoding == EncodingType::FixedStringDictionary) {
+            if (segment_value.size() > max_string_length) max_string_length = segment_value.size();
+          }
+        } else {
+          null_values[current_position] = true;
+        }
+      }
+    });
+
+    pmr_vector<T> dictionary(std::make_move_iterator(unique_values.cbegin()),
```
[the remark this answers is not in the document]
Bouncner: I don't get your comment. We do use the correct allocator and we don't have the unique (sorted) values before this point.

src/lib/storage/segment_encoding_utils.cpp
Bouncner: Copied your recommendation to the issue.

src/test/storage/encoded_segment_test.cpp
Bouncner: Haven't found a check for segment equality. So I need to construct to tables, append a chunk with that segment and then check for equality. Right?

mrks: Yes, but a function `check_segments_equal` would be the same effort while being more reusable.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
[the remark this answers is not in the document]
mrks: Ah, the comment was under the assumption that `unique_values` is still a vector. Nevermind.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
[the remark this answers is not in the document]
mrks: Did you test inserting into a vector, sorting it, and making it unique? If it doesn‘t make a difference, it doesn‘t matter.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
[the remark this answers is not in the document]
mrks: `allocate_shared` is like `make_shared`, but with an additional allocator passed in. Here, we are creating the dictionary with the default allocator and then copy (not move) it to memory allocated with the correct allocator. Does that copy make sense?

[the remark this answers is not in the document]
mrks: Also, in case there was any confusion: No need to get rid of the `shared_ptr`, the original comment was about the „sptr“ prefix

src/lib/storage/chunk_encoder.cpp
```diff
@@ -12,41 +12,114 @@
...
+      auto iterable = create_any_segment_iterable<ColumnDataType>(*segment);
+      iterable.with_iterators([&](auto it, auto end) {
+        const auto segment_size = std::distance(it, end);
+        values.reserve(segment_size);
+        null_values.reserve(segment_size);
+
+        for (; it != end; ++it) {
+          const auto segment_item = *it;
+          const auto is_null = segment_item.is_null();
+          null_values.push_back(is_null);
```
mrks: Why do you use reserve+push_back if you know the final size?

src/lib/storage/chunk_encoder.cpp
```diff
@@ -12,41 +12,114 @@
...
+          if (!is_null) {
+            values.push_back(segment_item.value());
+          } else {
+            values.push_back({});
+          }
+        }
+      });
+      std::cout << "WTF? " << values.size() << " & " << null_values.size()
+                << std::endl;  // testing what's wrong with the CI
```
mrks: Just making a comment so that this does not get lost

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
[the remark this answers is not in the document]
Bouncner: Yes, did test that. Was slower in most cases (for strings clearly slower).

src/lib/storage/lz4_segment/lz4_encoder.hpp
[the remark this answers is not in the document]
Bouncner: Works like a charm: https://gist.github.com/Bouncner/2d134821662626047cb8d629660e504c

src/lib/storage/chunk_encoder.cpp
Bouncner: Would you use `reserve()` + `[i]`? I thought we measured that both are equally fast? If not, I'll change it.

src/test/storage/encoded_segment_test.cpp
Bouncner: What do you think about #e71342f? What bothers me a little bit is that the result prints "Tables not equal" in case of a failure. But I'd like to avoid bloating up the .cpp file just to have a special version of the comparison method for segments.

src/lib/storage/chunk_encoder.cpp
mrks: That would be UB. `resize() + []` however, is significantly faster, because it does neither capacity checks nor has to update the actual size of the vector.

src/test/storage/encoded_segment_test.cpp
mrks: No # in front of commits. e71342f (so that I have something to click on)

mrks: That’s the general idea, yes. Just two things: First, you are ignore TypeCmpMode by requiring the types to be the same in any case (have a look at how that parameter is used later). Second, you would store ReferenceSegments in a data table.

src/lib/storage/chunk_encoder.cpp
Bouncner: Sorry, that's what I meant. We need an additional run variable though. Will change.

Bouncner: Changed.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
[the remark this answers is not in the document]
Bouncner: Thanks for bringing that up! The initially better performance was using std::unsorted_set, which does not provide a sorted order. With sorting it is slower than using std::set. And std::set is slower than the initial approach. I also tested using `std::unique_copy` but it did not show any performance improvements.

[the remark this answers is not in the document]
Bouncner: I updated the PR information with the current numbers.

src/test/storage/encoded_segment_test.cpp
Bouncner: Changed both issues (there is a "full version" now and two specializations, similar to `EXPECT_TABLE_EQ`). Still the "issue" with "Tables not equal".

mrzzzrm:
Please don't forget the "Fix #..." in you opening post, so that the issue gets automatically closed when this PR is merged.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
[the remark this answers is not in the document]
Bouncner: Not a big fan of all these shared_pointers. But I changed the code accordingly with d3ef430.

I allocate directly a shared object, to be able to use the allocator. From my understanding, first creating the object unshared and then making it shared cannot be done with a custom allocator, right?

Bouncner:
I addressed most issues (the `check_table_equal` prints bother me and two smaller issues are open), but I struggle to the get the memcheck passed. I did a full leak check on Pella and anything went through just fine.
Has anybody an idea here?

Bouncner:
I addressed most issues (the `check_table_equal` prints bother me, the somewhat necessary `::opossum::`, and the shared pointer handling in the dictionary segment encoder). Any help and comments are welcome.

src/test/storage/encoded_segment_test.cpp
Bouncner: @mrzzzrm : I'd like to close most issues here. Apart from the compiler issue (`::opossum::`) the issue with the `check_segments_equal` remains. When it fails, it prints something like "Tables are not equal". Any ideas or is it not worth changing it?

Bouncner:
WIP'ed the PR again. I am currently working on the calibration framework and recognized some unnecessary passing of encoding types and vector compression types (and partially the lack of vector compression types). As I saw this issue in other parts as well, I'll try to clean up the "encoding handling" a little bit.

Bouncner:
Ready to review again. Two small issues remain: I am not certain about the shared_pointer handling/allocation as discussed in an issue with @mrks and I am not certain about the output of `check_segments_equal()`.

src/lib/storage/base_segment_encoder.hpp
```diff
@@ -131,14 +131,18 @@ class SegmentEncoder : public BaseSegmentEncoder {
...
   template <typename ColumnDataType>
-  std::shared_ptr<BaseEncodedSegment> encode(const std::shared_ptr<const BaseValueSegment>& base_value_segment,
+  std::shared_ptr<BaseEncodedSegment> encode(const std::shared_ptr<const BaseSegment>& base_segment,
                                              hana::basic_type<ColumnDataType> data_type_c) {
     static_assert(decltype(supports(data_type_c))::value);
+    const auto iterable = create_any_segment_iterable<ColumnDataType>(*base_segment);
 
-    const auto value_segment = std::dynamic_pointer_cast<const ValueSegment<ColumnDataType>>(base_value_segment);
-    Assert(value_segment, "Value segment must have passed data type.");
+    // Obtain allocator of segment to encode. Currently, only ValueSegment store the used allocator.
```
mrks: One `s` missing here - you decide which one ;)

src/lib/storage/chunk_encoder.cpp
```diff
@@ -12,42 +12,118 @@
...
+      if (!encoding_spec.vector_compression_type ||
+          (encoding_spec.vector_compression_type && encoded_segment->compressed_vector_type() &&
+           *encoding_spec.vector_compression_type ==
+               parent_vector_compression_type(*encoded_segment->compressed_vector_type()))) {
+        result = segment;
+        return;
+      }
+    }
+
+    // In case of unencoded re-encoding, an any segment iterable is used to manually setup
```
mrks: Use the actual class name (`AnySegmentIterable`) so that it is easier to find should we ever want to rename it

src/lib/storage/chunk_encoder.hpp
```diff
@@ -39,23 +28,21 @@ using ChunkEncodingSpec = std::vector<SegmentEncodingSpec>;
...
-   * Note: In some cases, it might be beneficial to
-   *       leave certain segments of a chunk unencoded.
-   *       Use EncodingType::Unencoded in this case.
    */
   static void encode_chunk(const std::shared_ptr<Chunk>& chunk, const std::vector<DataType>& column_data_types,
                            const ChunkEncodingSpec& chunk_encoding_spec);
 
   /**
-   * @brief Encodes a chunk using the same segment-encoding spec
+   * @brief Encodes a chunk using the same segment encoding spec
```
mrks: Use the class name

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
```diff
@@ -28,112 +29,92 @@ class DictionaryEncoder : public SegmentEncoder<DictionaryEncoder<Encoding>> {
...
   template <typename T>
-  std::shared_ptr<BaseEncodedSegment> _on_encode(const std::shared_ptr<const ValueSegment<T>>& value_segment) {
-    // See: https://goo.gl/MCM5rr
-    // Create dictionary (enforce uniqueness and sorting)
-    const auto& values = value_segment->values();
+  std::shared_ptr<BaseEncodedSegment> _on_encode(const AnySegmentIterable<T> segment_iterable,
+                                                 const PolymorphicAllocator<T>& allocator) {
+    // Vectors to gather the input segment's data. This data is used in a later step to
+    // construct the actual dictionary and attribute vector.
+    std::vector<T> values;          // does contain the actual values (no NULLs)
```
mrks: Simply “contains”? Also, just to make sure no one misses it “This is different from other definitions of `values`”?

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
```diff
@@ -28,112 +29,92 @@ class DictionaryEncoder : public SegmentEncoder<DictionaryEncoder<Encoding>> {
...
-
-        const auto value_id = _get_value_id(dictionary, *value_it);
-        attribute_vector.push_back(value_id);
-      }
-    } else {
-      auto value_it = values.cbegin();
-      for (; value_it != values.cend(); ++value_it) {
-        const auto value_id = _get_value_id(dictionary, *value_it);
-        attribute_vector.push_back(value_id);
+  std::shared_ptr<BaseEncodedSegment> _encode_dictionary_segment(const U& dictionary, const std::vector<T>& values,
```
mrks: This only creates the attribute vector, but it does not create a dictionary segment from scratch. In itself, it seems useless. Move it into a lambda `create_attribute_vector` that is part of `_on_encode`?

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
```diff
@@ -28,112 +29,92 @@ class DictionaryEncoder : public SegmentEncoder<DictionaryEncoder<Encoding>> {
...
+        attribute_vector[current_position] = value_id;
+        ++values_iter;
+      } else {
+        attribute_vector[current_position] = null_value_id;
       }
     }
 
     // We need to increment the dictionary size here because of possible null values.
-    const auto max_value = dictionary.size() + 1u;
+    const auto max_value = dictionary->size() + 1u;
```
mrks: max_value_id?

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
```diff
@@ -28,112 +29,92 @@ class DictionaryEncoder : public SegmentEncoder<DictionaryEncoder<Encoding>> {
...
-        attribute_vector, SegmentEncoder<DictionaryEncoder<Encoding>>::vector_compression_type(), alloc, {max_value});
-    auto dictionary_sptr = std::allocate_shared<U>(alloc, std::move(dictionary));
-    auto attribute_vector_sptr = std::shared_ptr<const BaseCompressedVector>(std::move(encoded_attribute_vector));
+    auto encoded_attribute_vector = std::shared_ptr<const BaseCompressedVector>(
+        compress_vector(attribute_vector, SegmentEncoder<DictionaryEncoder<Encoding>>::vector_compression_type(),
+                        allocator, {max_value}));
 
     if constexpr (Encoding == EncodingType::FixedStringDictionary) {
-      return std::allocate_shared<FixedStringDictionarySegment<T>>(alloc, dictionary_sptr, attribute_vector_sptr,
+      return std::allocate_shared<FixedStringDictionarySegment<T>>(allocator, dictionary, encoded_attribute_vector,
```
mrks: Encoded vs compressed

src/lib/storage/lz4_segment/lz4_encoder.hpp
```diff
@@ -63,32 +63,31 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
-      // values vectors
-      for (auto row_index = size_t{0u}; it != end; ++it, ++row_index) {
-        auto segment_value = *it;
+    segment_iterable.with_iterators([&](auto segment_it, auto segment_end) {
+      const auto segment_size = std::distance(segment_it, segment_end);
+      values.resize(segment_size);
+      null_values.resize(segment_size);
+
+      // iterate over the segment to access the values and increment the row index to copy values and null flags
+      for (auto row_index = size_t{0u}; segment_it != segment_end; ++segment_it, ++row_index) {
```
mrks: Not a fan of non-classical for-loops. I find it too easy to miss something

src/lib/storage/lz4_segment/lz4_encoder.hpp
```diff
@@ -63,32 +63,31 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
+        const auto segment_value = *segment_it;
+        const auto is_null = segment_value.is_null();
         values[row_index] = segment_value.value();
-        null_values[row_index] = segment_value.is_null();
-        contains_null_value = contains_null_value || segment_value.is_null();
+        null_values[row_index] = is_null;
+        contains_null_value = contains_null_value || is_null;
```
mrks: Simply `contains_null`?

src/lib/storage/lz4_segment/lz4_encoder.hpp
```diff
@@ -168,37 +165,40 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
...
-    auto offsets = pmr_vector<uint32_t>{alloc};
-    offsets.resize(num_elements);
+    auto offsets = pmr_vector<uint32_t>{allocator};
 
     /**
      * These are the lengths of each string. They are needed to train the zstd dictionary.
      */
-    auto sample_sizes = pmr_vector<size_t>{alloc};
-    sample_sizes.resize(num_elements);
+    auto string_samples_lengths = pmr_vector<size_t>{allocator};
```
mrks: I probably asked this when we originally brought this is, but how are these “samples” if we put in every single value?

src/lib/storage/segment_encoding_utils.cpp
```diff
@@ -40,16 +39,34 @@ std::unique_ptr<BaseSegmentEncoder> create_encoder(EncodingType encoding_type) {
...
-                                                   std::optional<VectorCompressionType> zero_suppression_type) {
-  auto encoder = create_encoder(encoding_type);
+std::shared_ptr<BaseEncodedSegment> encode_and_compress_segment(const std::shared_ptr<const BaseSegment>& segment,
+                                                                const DataType data_type,
+                                                                const SegmentEncodingSpec& encoding_spec) {
+  auto encoder = create_encoder(encoding_spec.encoding_type);
 
-  if (zero_suppression_type) {
-    encoder->set_vector_compression(*zero_suppression_type);
+  if (encoder->uses_vector_compression() && encoding_spec.vector_compression_type) {
```
mrks: In which case would one of these be true and the other false?

src/lib/utils/check_table_equal.cpp
```diff
@@ -95,6 +95,33 @@ bool almost_equals(T left_val, T right_val, FloatComparisonMode float_comparison
...
+      std::vector<TableColumnDefinition>{TableColumnDefinition("single_column", segment_to_test->data_type(), true)};
+
+  auto table_type = [&](const std::shared_ptr<BaseSegment> segment) {
+    if (const auto reference_segment = std::dynamic_pointer_cast<const ReferenceSegment>(segment)) {
+      return TableType::References;
+    }
+    return TableType::Data;
+  };
+
+  auto table_to_test = std::make_shared<Table>(definitions, table_type(segment_to_test));
```
mrks: I believe we call the inputs actual and <s>h...</s> expected

src/lib/utils/check_table_equal.cpp
```diff
@@ -95,6 +95,33 @@ bool almost_equals(T left_val, T right_val, FloatComparisonMode float_comparison
...
+  table_to_test->append_chunk(pmr_vector<std::shared_ptr<BaseSegment>>{segment_to_test});
+  auto expected_table = std::make_shared<Table>(definitions, table_type(expected_segment));
+  expected_table->append_chunk(pmr_vector<std::shared_ptr<BaseSegment>>{expected_segment});
+
+  // The optional returned by check_table_equal is set when the function failed to determine equality.
```
mrks: Took me three times to understand. Maybe “If `check_table_equal` returns something other than std::nullopt, a difference has been found.”

src/lib/visualization/pqp_visualizer.cpp
```diff
@@ -7,6 +7,7 @@
 #include "operators/limit.hpp"
 #include "operators/projection.hpp"
 #include "operators/table_scan.hpp"
+#include "utils/format_bytes.hpp"
```
mrks: ?

src/test/storage/any_segment_iterable_test.cpp
```diff
@@ -83,6 +78,8 @@ INSTANTIATE_TEST_CASE_P(
     AnySegmentIterableTestInstances, AnySegmentIterableTest,
     ::testing::Values(SegmentEncodingSpec{EncodingType::Unencoded}, SegmentEncodingSpec{EncodingType::Dictionary},
                       SegmentEncodingSpec{EncodingType::Dictionary, VectorCompressionType::FixedSizeByteAligned},
-                      SegmentEncodingSpec{EncodingType::Dictionary, VectorCompressionType::SimdBp128}), );  // NOLINT
+                      SegmentEncodingSpec{EncodingType::Dictionary, VectorCompressionType::SimdBp128},
+                      SegmentEncodingSpec{EncodingType::FrameOfReference}, SegmentEncodingSpec{EncodingType::RunLength},
+                      SegmentEncodingSpec{EncodingType::LZ4}), );  // NOLINT
```
mrks: Argh. We have too many places where these definitions have to be added. Can we centralize this? Maybe as part of a different PR?

src/lib/utils/check_table_equal.hpp
```diff
@@ -25,6 +25,10 @@ enum class TypeCmpMode { Strict, Lenient };
  */
 enum class FloatComparisonMode { RelativeDifference, AbsoluteDifference };
 
+bool check_segment_equal(const std::shared_ptr<BaseSegment>& segment_to_test,
```
mrks: Doc!

src/test/testing_assert.hpp
```diff
@@ -64,6 +64,27 @@ bool contained_in_query_plan(const std::shared_ptr<const AbstractOperator>& node
   EXPECT_TABLE_EQ(opossum_table, expected_table, OrderSensitivity::Yes, TypeCmpMode::Strict, \
                   FloatComparisonMode::AbsoluteDifference)
 
+/**
+ * Compare two segments with respect to OrderSensitivity, TypeCmpMode and FloatComparisonMode
+ */
+#define EXPECT_SEGMENT_EQ(segment_to_test, expected_segment, order_sensitivity, type_cmp_mode, float_comparison_mode) \
+  EXPECT_TRUE(segment_to_test&& expected_segment&& check_segment_equal(                                               \
```
mrks: Looks like the formatter went berserk here

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
Bouncner: Changed to "contains". I guess you mean that other times `values` stores NULL values as well? I think using `vector<T>` makes that clear enough. Or do you mean something else?

src/lib/visualization/pqp_visualizer.cpp
Bouncner: Removed it in another file which did not use any byte formatting. And the PQP visualizer uses format_bytes.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
mrks: (Almost?) everywhere else, this vector stores `T{}` for NULL values. Otherwise, we wouldn't be able to access values by index.

Here, it is fine, because we never access by index. I'd even go as far as to call it `dense_values`

src/test/storage/any_segment_iterable_test.cpp
Bouncner: I do not really know what you're meaning. For testing, I rather like to explicitly list all variants and especially test the non-default vector compressions.

Do you mean you'd like a function that yields all possible combinations?

mrks: LZ4 should have been added here when it was originally introduced. The fact that it wasn't shows that we have too many places where new segment types have to be added. I'll open an issue.

src/lib/utils/check_table_equal.cpp
Bouncner: Isn't actual the same as expected?

src/test/testing_assert.hpp
Bouncner: It wouldn't bother me. Shall I add the two lines for clang-format to disable it or leave it as is?

mrks: It took me some seconds to understand what happens here. I'd add those two lines.

src/lib/utils/check_table_equal.cpp
mrks: No. See the ten of `actual_lqp / expected_lqp` in the tests. Also:

https://github.com/hyrise/hyrise/blob/0046cfdb7a6ad4937294cbca64f4e39f68558e0e/src/lib/utils/check_table_equal.cpp#L148-L151

src/lib/storage/segment_encoding_utils.cpp
Bouncner: To be honest, I thought the check is not necessary, but valgrind complained and I was not able to reproduce it locally. The check fixed it for the CI setup. I will take another look at it.

mrks: I'd be interested in the valgrind message

src/test/storage/any_segment_iterable_test.cpp
Bouncner: Ah yes, got you. Makes sense.

src/test/testing_assert.hpp
Bouncner: Wait? You are not talking about the missing 10 spaces for the third line but about something else, right? What did confuse you exactly?

mrks: The `&&` that looks like a universal reference but is an "and"

src/lib/storage/lz4_segment/lz4_encoder.hpp
Bouncner: The values are(/might be) later "sampled" (taking the `n` first values) within `_train_dictionary(const pmr_vector<T>& values, const pmr_vector<size_t>& sample_sizes)`.

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
Bouncner: I adapted the function in a way that is is only responsible for the creation of compressed attribute vectors.
However, I am too stupid to move it into a simple lambda since I fail to get around the error, that dictionary template type cannot be deduced (_on_encode had two template parameters, T and S, and S was the parameter of the lambda function, S could not be deduced).

src/lib/storage/segment_encoding_utils.cpp
Bouncner: Cannot reproduce. `encoder->uses_vector_compression()` has been removed again.

src/lib/storage/lz4_segment/lz4_encoder.hpp
Bouncner: There are several issues with the current LZ4 segment implementation, which I plan to rewrite. Shall I open an issue with an overview or do you trust me to address these issues in an upcoming PR?

Amongst these things are the removal of offsets, larger blocks (we basically never fill up the 4 KB), "real" sampling (tracking unique values and using the most often occuring values when samples too large), and more.

src/lib/utils/check_table_equal.cpp
```diff
@@ -113,12 +113,12 @@ bool check_segment_equal(const std::shared_ptr<BaseSegment>& segment_to_test,
     return TableType::Data;
   };
 
-  auto table_to_test = std::make_shared<Table>(definitions, table_type(segment_to_test));
-  table_to_test->append_chunk(pmr_vector<std::shared_ptr<BaseSegment>>{segment_to_test});
+  auto table_to_test = std::make_shared<Table>(definitions, table_type(actual_segment));
```
mrks: actual_table

src/lib/storage/dictionary_segment/dictionary_encoder.hpp
```diff
@@ -58,19 +58,48 @@ class DictionaryEncoder : public SegmentEncoder<DictionaryEncoder<Encoding>> {
...
 
-    auto dictionary = std::make_shared<pmr_vector<T>>(values.cbegin(), values.cend(), allocator);
+    auto dictionary = std::make_shared<pmr_vector<T>>(dense_values.cbegin(), dense_values.cend(), allocator);
     std::sort(dictionary->begin(), dictionary->end());
     dictionary->erase(std::unique(dictionary->begin(), dictionary->end()), dictionary->cend());
     dictionary->shrink_to_fit();
 
+    const auto null_value_id = static_cast<uint32_t>(dictionary->size());
+
+    auto create_compressed_attribute_vector = [&](const auto& local_dictionary) {
```
mrks: Rename the dictionary from line 61 to `temp_dictionary` and use `dictionary` here?

PR #1665, opened 2019-05-07
Fix issue with empty string vectors and FixedStringVector

### LINKED ISSUE #1663: FixedString Vectors cannot handle empty vectors
The tests fail as soon as the FixedStringVectors are really tested with empty columns. Unfortunately, neither the encoding tests include FixedStringVectors nor do the FixedStringVectors tests test for this case.

#1657 added FixedStringDictionaries to the encoding tests and caused sanitizer issues.

Bouncner:
#1657 added a test to the string encoding test suite for the `FixedStringDictionaries` which use the `FixedStringVector`. These tests tested completely empty string vectors (i.e., every string has a length of 0) which failed with ASAN builds. The reason was that a check for the last char is done (string length - 1) which accesses an invalid address.

This does not sound super interesting, but in real-world data sets, empty string segments can occur frequently and FixedStringVectors save us a tremendous amount of memory here.

Fixes #1663 and #1666.


src/test/storage/fixed_string_vector_test.cpp
```diff
@@ -168,6 +168,8 @@ TEST_F(FixedStringVectorTest, StringLengthZero) {
   fixed_string_vector1.push_back("");
   EXPECT_EQ(fixed_string_vector1.size(), 1u);
   EXPECT_EQ(fixed_string_vector1[0u], "");
+  EXPECT_EQ(fixed_string_vector1[1u], "");
+  EXPECT_EQ(fixed_string_vector1[3u], "");
```
mweisgut: Why are these two assertions added which do not execute the fixed `get_string_at(…)`-function? Adding the SegmentEncodingSpecs in EncodedStringSegmentTest is sufficient, isn't it?

Bouncner: Argh, mixed up two things things here. Thanks!

PR #1667, opened 2019-05-10
Use JoinOperator::supports() to find a matching Join implementation for a JoinNode

### LINKED ISSUE #1529: Sort-Merge Join cannot compare float with double
**Query**
```sql
SELECT *
FROM orders
JOIN (
  SELECT SUM(o_totalprice) s
  FROM orders
) order_price_sum
ON o_totalprice <= order_price_sum.s;
```

**Actual behavior**
It crashes:
`src/lib/operators/join_sort_merge.cpp:64 Left and right column types do not match. The sort merge join requires matching column types`

**Expected behavior**
The query should return all orders.

**Explanation**
`o_totalprice` is a `float`, while `SUM(o_totalprice)` is a `double`. The sort-merge join should be able to handle this.

mrzzzrm:
Fix #1529 

Fall back from JoinHash to JoinSortMerge to JoinNestedLoop to find a matching Join operator for JoinNode.

### DIFF
--- a/src/test/logical_query_plan/lqp_translator_test.cpp
+++ b/src/test/logical_query_plan/lqp_translator_test.cpp
@@ -523,7 +523,7 @@ TEST_F(LQPTranslatorTest, JoinNodeToJoinHash) {
   const auto op = LQPTranslator{}.translate_node(join_node);
 
   /**
-   * Check PQP
+   * Check PQP - for a inner-equi join, JoinHash should be used.
    */
   const auto join_op = std::dynamic_pointer_cast<JoinHash>(op);
   ASSERT_TRUE(join_op);
@@ -541,7 +541,7 @@ TEST_F(LQPTranslatorTest, JoinNodeToJoinSortMerge) {
   const auto op = LQPTranslator{}.translate_node(join_node);
 
   /**
-   * Check PQP
+   * Check PQP - JoinHash doesn't support non-equi joins, thus we fall back to JoinSortMerge
    */
   const auto join_op = std::dynamic_pointer_cast<JoinSortMerge>(op);
   ASSERT_TRUE(join_op);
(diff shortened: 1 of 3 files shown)


src/test/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -546,22 +515,57 @@ TEST_F(LQPTranslatorTest, ProjectionNode) {
...
+   * Check PQP
+   */
+  const auto join_op = std::dynamic_pointer_cast<JoinHash>(op);
+  ASSERT_TRUE(join_op);
+  EXPECT_EQ(join_op->primary_predicate().column_ids, ColumnIDPair(ColumnID{1}, ColumnID{1}));
+  EXPECT_EQ(join_op->primary_predicate().predicate_condition, PredicateCondition::Equals);
+  EXPECT_EQ(join_op->mode(), JoinMode::Inner);
+}
+
+TEST_F(LQPTranslatorTest, JoinNodeToJoinSortMerge) {
```
Bouncner: Short note, why in this test no HashJoin is yielded?

Bouncner: Or maybe on top of the three tests a short overview, that the tests get "increasingly hard" and thus fall back to "slower" implementations?

mweisgut: The JoinHash cannot be used here because we have LessThan as predicate condition.

Bouncner: Yes, I know, but most don’t and thus I’d like to see a short comment with the reason.
Otherwise it’s manually searching the joins and their `supports()` methods to understand this “order”.

mrzzzrm: comment added to each test

mrzzzrm:
> I was surprised the LQP translation part needed that much "magic" for such a simple test. But I assume you know what you are doing. :)

Much magic? That's literally just iterating over a type-tuple. I refuse to call this magic, otherwise I wouldn't have used it.

> One last thing: I would vote for a constexpr list of join operators, which reflects the assume "performance order" (a little bit more explicit separation between performance and applicability test). Because that's the single point where a later cost model needs to touch the current code.

Extracted the list into a variable like so:

```c++
  constexpr auto join_operator_preference_order = hana::to_tuple(hana::tuple_t<JoinHash, JoinSortMerge, JoinNestedLoop>);
```

Bouncner:
> Much magic? That's literally just iterating over a type-tuple. I refuse to call this magic, otherwise I wouldn't have used it.

Anything that uses `boost::hana` is black magic to me.

However, everything looks fine to me.

PR #1668, opened 2019-05-13
Add AggregateTestRunner to thoroughly test aggregate operators 

mrzzzrm:
WIP, as always I just want a colorful diff


Bouncner:
I have no idea what to do with the following note but I wanted to bring it up nonetheless. There is a ton of aggregation functions that exist, especially when looking into OLAP/analytical workloads. Currently, we have a very restricted set of a few function.
It might make sense to already think about grouping the functions into distributive/(algebraic)/holistic functions and test one/a few functions of each set (see §2.2,  https://research.tableau.com/sites/default/files/p1221-wesley.pdf). Afaik, we do not support any holistic functions, so that might be an empty set.

PR #1675, opened 2019-05-20
Adapt supported encodings of FoR

### LINKED ISSUE #1662: FrameOfReference encoding (or vector compression?) cannot handle > UINT_MAX
Wanted to add FoR to the JoinTestRunner, which uses values > UINT_MAX, ran into this Assert in the `FrameOfReferenceEncoder`:

```c++
        // Make sure that the largest offset fits into uint32_t (required for vector compression.)
        Assert(static_cast<std::make_unsigned_t<T>>(*max_it - *min_it) <= std::numeric_limits<uint32_t>::max(),
               "Value range in block must fit into uint32_t.");
```

Just disable vector compression if the values are too large for it? 
Why is this Assert in the FoR-Encoder and not in compress_vector()?

Bouncner:
This PR disables the support of `int64_t` for FrameOfReverence segments.

During table generation for the calibration framework, I stumbled over the issue that Moritz found during the work on the join test runner as well. FrameOfReference segments do not support `uint64_t` values since the underlying vector compression only handles `uint32t` values (see discussion in #1662).

I would vote for extending the vector compression to default to `uint64_t` values, but that has several disadvantages (one of them compile time). And since SIMDBP adaptions do not appear to be too easy, I don't think it will be fixed within the next couple of weeks.

This change simply limits the usage of FrameOfReference segments. No benchmark I know of uses anything larger int and most users won't use FoR at all. So I think this limitation of FoR does not hurt.

Fix #1662.

### DIFF
--- a/src/lib/storage/frame_of_reference_segment.cpp
+++ b/src/lib/storage/frame_of_reference_segment.cpp
@@ -88,6 +88,7 @@ std::optional<CompressedVectorType> FrameOfReferenceSegment<T, U>::compressed_ve
 }
 
 template class FrameOfReferenceSegment<int32_t>;
-template class FrameOfReferenceSegment<int64_t>;
+// int64_t disabled for now, as vector compression cannot handle 64 bit values
+// template class FrameOfReferenceSegment<int64_t>;
 
 }  // namespace opossum
--- a/src/lib/storage/encoding_type.hpp
+++ b/src/lib/storage/encoding_type.hpp
@@ -37,7 +37,7 @@ constexpr auto supported_data_types_for_encoding_type = hana::make_map(
     hana::make_pair(enum_c<EncodingType, EncodingType::Dictionary>, data_types),
     hana::make_pair(enum_c<EncodingType, EncodingType::RunLength>, data_types),
     hana::make_pair(enum_c<EncodingType, EncodingType::FixedStringDictionary>, hana::tuple_t<pmr_string>),
-    hana::make_pair(enum_c<EncodingType, EncodingType::FrameOfReference>, hana::tuple_t<int32_t, int64_t>),
+    hana::make_pair(enum_c<EncodingType, EncodingType::FrameOfReference>, hana::tuple_t<int32_t>),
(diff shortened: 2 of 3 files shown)


src/lib/storage/frame_of_reference_segment.cpp
```diff
@@ -88,6 +88,7 @@ std::optional<CompressedVectorType> FrameOfReferenceSegment<T, U>::compressed_ve
 }
 
 template class FrameOfReferenceSegment<int32_t>;
-template class FrameOfReferenceSegment<int64_t>;
+// int64_t disabled for now, as vector compression cannot handle 64 bit values
```
mrks: Vector compression in itself can handle 64-bit, but FoR only makes sense with bit packing, correct? That should be clarified here.

Bouncner: No, vector compression does not support larger integers than 4 byte integers. Since we compress dictionary offsets in dictionary-encoded segments, that was never a problem.
But FoR (more or less) compresses the actual values.

And yes, FoR does not have an advantage when the deltas cannot be compressed via null surpression (~bit packing).

mrks:
Ah, you're right.  I thought at least fsba did.

src/lib/storage/frame_of_reference_segment.cpp
Bouncner: Would change to:
```
// Support for int64_t is disabled for now, because the current vector compression used by
// the FrameOfReverence encoding can only handle vectors of uint32_t. As the delta values for
// long segments can be larger than uint32_t, larger than 4-byte-integers are not supported.
// template class FrameOfReferenceSegment<int64_t>;
```

Would that address all your issues?

Bouncner: Whoopsie ... already merged. :)

There will be another PR today (table generator does not accept SegmentEncodingSpecs). I would add this comment there, if you find it helpful.

mrks: Sounds good

PR #1678, opened 2019-05-22
Binaries tests

engelfa:
This PR adds tests for the binaries hyriseBenchmarkFileBased, hyriseBenchmarkJoinOrder and hyriseBenchmarkTPCH as requested in #1103.


engelfa:
We can move the tests to the FullCI step now.
Is it a good idea to create a new stage for all four binary tests?

scripts/test/hyriseBenchmarkTPCH_test.py
```diff
@@ -0,0 +1,71 @@
...
+
+arguments1 = {}
+arguments1["--scale"] = ".01"
+arguments1["--use_prepared_statements"] = "true"
+arguments1["--queries"] = "'1,13,19'"
+arguments1["--time"] = "1"
+arguments1["--runs"] = "100"
+arguments1["--mode"] = "'PermutedQuerySet'"
+arguments1["--encoding"] = "'Unencoded'"
+arguments1["--compression"] = "'SIMD-BP128'"
```
Bouncner: I have no idea what you guys are doing here, but just looking at the config: compression does only affect a few encodings (dict, lz4, for FoR also but that one does only handle ints).
So for Unencoded there would be no effect.

Bouncner: No specified encoding should default to dict in arguments2. So here the compression does have an effect.

scripts/test/hyriseBenchmarkFileBased_test.py
```diff
@@ -0,0 +1,83 @@
...
+	
+	os.system("rm -rf imdb_data/*.bin")
+
+	benchmark = initialize(arguments1)
+	close_benchmark(benchmark)
+	check_exit_status(benchmark)
+
+	os.system("rm -rf imdb_data/*.bin")
+
+	benchmark = initialize(arguments2)
```
mrks: This is a bit hard to read. As the arguments are only used here, it is not helpful to put them at the top. Also, is there any reason we can't list them as a string?

scripts/test/hyriseBenchmarkFileBased_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+def close_benchmark(benchmark):
+	benchmark.expect(pexpect.EOF, timeout=None)
+	benchmark.close()
+
+def check_exit_status(benchmark):
+	if benchmark.exitstatus == None:
+		sys.exit(benchmark.signalstatus)
+
+def main():
```
mrks: A bit of documentation on what this does would be great

scripts/test/hyriseBenchmarkJoinOrder_test.py
```diff
@@ -0,0 +1,77 @@
...
+	benchmark.close()
+
+def check_exit_status(benchmark):
+	if benchmark.exitstatus == None:
+		sys.exit(benchmark.signalstatus)
+
+def main():
+	return_error = False
+	
+	os.system("rm -rf imdb_data/*.bin")
```
mrks: Check that bins are created?

scripts/test/hyriseBenchmarkJoinOrder_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+	build_dir = sys.argv[1]
+
+	concat_arguments = ' '.join(['='.join(map(str, x)) for x in arguments.items()])
+
+	benchmark = pexpect.spawn(build_dir + "/hyriseBenchmarkJoinOrder " + concat_arguments, maxread=1000000,timeout=None, dimensions=(200, 64))
+	benchmark.logfile = sys.stdout
+	return benchmark
+
+def close_benchmark(benchmark):
```
mrks: There is a lot of duplication here. Move to a common place?

scripts/test/hyriseBenchmarkTPCH_test.py
```diff
@@ -0,0 +1,71 @@
...
+
+def main():
+	return_error = False
+	
+	benchmark = initialize(arguments1)
+	close_benchmark(benchmark)
+	check_exit_status(benchmark)
+
+	benchmark = initialize(arguments2)
+	benchmark.expect("-> Executed")
```
mrks: I think we should test a bit more (e.g., creation of JSON files, ...). Right now, there is only a single expect per benchmark.

src/benchmark/file_based_benchmark.cpp
```diff
@@ -25,21 +25,24 @@ int main(int argc, char* argv[]) {
   cli_options.add_options()
       ("table_path", "Directory containing the Tables as csv, tbl or binary files. CSV files require meta-files, see csv_meta.hpp or any *.csv.json file.", cxxopts::value<std::string>()->default_value(".")) // NOLINT
       ("query_path", "A specific .sql file or a directory containing .sql files", cxxopts::value<std::string>()->default_value(".")) // NOLINT
-      ("q,queries", "Subset of queries to run as a comma separated list", cxxopts::value<std::string>()->default_value("all")); // NOLINT
+      ("q,queries", "Subset of queries to run as a comma separated list", cxxopts::value<std::string>()->default_value("all")) // NOLINT
+      ("excluded_queries", "Non-queries to exlude as a comma separated list", cxxopts::value<std::string>()->default_value("none")); // NOLINT
```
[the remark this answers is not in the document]
engelfa: We could delete the files beforehand.

scripts/test/hyriseBenchmarkFileBased_test.py
engelfa: We can list them as a string. I just thought it's easier to identify which parameters are tested than in a long string.

scripts/sample_imdb.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
mrks: (Just to clarify: What is this script file good for, not the shebang)

src/benchmark/file_based_benchmark.cpp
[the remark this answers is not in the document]
mrks: Or use --queries?

scripts/test/hyriseBenchmarkFileBased_test.py
mrks: Personally, it makes no difference for me and a string takes up less space. Your call, but I'd move it here in any case.

src/benchmark/file_based_benchmark.cpp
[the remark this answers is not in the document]
engelfa: That doesn't work right now because in the constructor of the FileBasedQueryGenerator all SQL files in the given directory are checked for valid SQL in line 65 and the hsql parser doesn't support indexes right now. Therefore, an error is thrown.

scripts/test/hyriseBenchmarkTPCH_test.py
engelfa: Now, we test the binary tables, JSON files, and sqlite verification.

mrzzzrm:
This passes a lot of important options to the benchmark (--queries, --clients, --encoding, ...) and almost does not test the stdout of the benchmark. I'd expect [sic] a `benchmark.expect("...")` line for each options passed to the benchmark.

The setup with `arguments1` and `arguments2` seems overly strict to me. What if I wanted to add a third test to the hyriseFileBasedBenchmark? Currently I'd need to extend the other tests, too.

Deleting imdb_data/*.bin when running these tests seems invasive to me, as this is probably where everyone keeps their default setup of JoinOrderBenchmark files. The tests should use a different --table_path.

Overall, I'd expect these tests to be as simple as 
```python
b = start_benchmark("./hyriseFileBasedBenchmark --table_path xyz --query_path uvw --queries 'select_statement' -t 20 -r 10 --visualize)
b.expect("Starting File Based Benchmark)
b.expect("Loading tables from xyz")
b.expect("Loading queries from uvw")
b.expect("Running a subset of queries: select_statement")
b.expect("Running for a maximum of 20 seconds")
b.expect("Performing a maximum of 10 runs")
b.expect("Visualization is enabled. Benchmark results will be unrelyable")
b.expect("Using dictionary encoding);
b.expect("Executing select_statement");
b.not_expect("Executing some_other_statement")
b.expect("Visualization done")
b.wait() # should throw exception if return code != 0
```

mrks:
Agreeing with all but the last point. For me, the value of checking the stdout is very small. Since those lines are printed right after taking in the argument, it does not prove much. It does not mean that they are properly used in the actual benchmark. Instead, they lead to a false sense of security.

mrks:
We could look at those parameters whose *effects* can be observed from the outside. For example, when using `--queries 1,4,5`, there should be three items listed in the JSON and the sum of their execution times should roughly be `3 * value_of_t`.

mrzzzrm:
> For me, the value of checking the stdout is very small.

It tests the option is acknowledged and would have saved us trouble in the past. I'd include it. Also, for example, this would test that `--queries "a, b, c"` splits up "a,b,c" correctly.

> JSON and the sum of their execution times

checking the JSON, yes. Summing up execution times I wouldn't do. Or maybe only to check for != 0, every other reference value is fragile. I, for example, couldn't run hyriseTest in release for a while because someone thought it would be smart to test for "execution_time_in_milliseconds > 0". Not everyone has slow apple hardware :(

mrks:
> It tests the option is acknowledged and would have saved us trouble in the past.

But not necessarily that it is correctly passed through - see #1699

mrzzzrm:
I agree with you in principle and I think we both agree the "penetration" of the benchmarks via these tests need to be improved. I would not suggest checking this output as the only method of testing, but merely as a complementary tool. And I do think this output in itself is important to be correct, especially when command lines get so long that people will not read them anymore (as I often do) and rely on the stdout instead (as I often do).

> In the quoted example, the test would give us that false sense of security. We have the output, but it is wrong.

(Not sure which "quoted example" you mean - but we don't have to press this to far. )

scripts/sample_imdb.sh
```diff
@@ -0,0 +1,31 @@
...
+grep -r -m 1 3292787.*\"Bulgaria\"    imdb_data/movie_info.csv        > imdb_sample/movie_info.csv
+tail -n 1                             imdb_data/movie_info_idx.csv    > imdb_sample/movie_info_idx.csv
+grep -w 3292787,1136                  imdb_data/movie_keyword.csv     > imdb_sample/movie_keyword.csv
+tail -n 1                             imdb_data/movie_link.csv        > imdb_sample/movie_link.csv
+tail -n 1                             imdb_data/name.csv              > imdb_sample/name.csv
+tail -n 1                             imdb_data/person_info.csv       > imdb_sample/person_info.csv
+tail -n 1                             imdb_data/role_type.csv         > imdb_sample/role_type.csv
+grep -w 3292787                       imdb_data/title.csv             > imdb_sample/title.csv
+
+cp imdb_data/*.csv.json          			imdb_sample/
```
mrks: The additional space here are confusing

scripts/sample_imdb.sh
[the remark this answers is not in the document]
mrks: Since we already have the data that this script generates as part of the commit, why do we need both?

Dockerfile
```diff
@@ -16,6 +16,7 @@ RUN apt-get update \
         gcovr \
         gcc-8 \
         g++-8 \
+        graphviz \
```
mrks: Random line:

```
-  Loading 'lineitem' into SQLite Traceback (most recent call last):
  File "./scripts/test/hyriseBenchmarkTPCH_test.py", line 134, in <module>
    main()
  File "./scripts/test/hyriseBenchmarkTPCH_test.py", line 80, in main
    benchmark.expect("-> Executed")
  File "/usr/local/lib/python2.7/site-packages/pexpect/spawnbase.py", line 341, in expect
    timeout, searchwindowsize, async_)
  File "/usr/local/lib/python2.7/site-packages/pexpect/spawnbase.py", line 369, in expect_list
    return exp.expect_loop(timeout)
  File "/usr/local/lib/python2.7/site-packages/pexpect/expect.py", line 119, in expect_loop
    return self.timeout(e)
  File "/usr/local/lib/python2.7/site-packages/pexpect/expect.py", line 82, in timeout
    raise TIMEOUT(msg)
pexpect.exceptions.TIMEOUT: Timeout exceeded.
<pexpect.pty_spawn.spawn object at 0x10b0a98d0>
command: build/hyriseBenchmarkTPCH
args: ['build/hyriseBenchmarkTPCH', '--runs=100', '--scale=.005', '--warmup=1', '--queries=2,4,6', '--compression=SIMD-BP128', '--time=1', '--verify=true', '--encoding=LZ4']
```

engelfa: I've increased the timeout.

scripts/test/hyriseBenchmarkFileBased_test.py
```diff
@@ -0,0 +1,139 @@
...
+# During the first run, the shell output is validated using pexpect.
+# After the first run, this test checks if an output file was created and if it matches the arguments.
+# During the second run, the shell output is validated using pexpect
+# and the test checks if all queries were successfully verified with sqlite.
+def main():
+
+  return_error = False
+
+  arguments1 = {}
+  arguments1["--table_path"] = "'resources/test_data/tbl/file_based/'"
```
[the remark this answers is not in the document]
engelfa: For consistency, I just added them to all strings.

scripts/sample_imdb.sh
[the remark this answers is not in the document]
engelfa: You are right, we don't really need it. I just added it for the sake of completeness.

engelfa:
Combined, all three tests take about 2:30 minutes.
Is that still short enough or shall I move them to the FullCI stage?

mrks:
I'd run it when hyriseSystemTest is run (but not with sanitizers etc). If I remember correctly, that would place it in FullCI territory.

engelfa:
#1717

PR #1679, opened 2019-05-24
File-Based TPC-DS Benchmark

mweisgut:
This PR contains the current file-based implementation of the master project's TPC-DS benchchmark.
Both the data and the queries are currently generated file-based.
The corresponding query files can be found in `third_party/tpcds-result-reproduction/query_qualification/`, the table files in `resources/benchmark/tpcds/tables`.

To significantly speed up the speed of SQLite verification, similar to the JoinOrderingBenchmark, the TPC-DS benchmark implementation creates indices on the foreign keys for the SQLite verification.

The benchmark uses the official tpc-ds toolkit's directory `tools` to generate the binary `dsdgen`. This binary is used to generate the table data (csv) files.
These files are stored in `resources/benchmark/tpcds/tables`.


resources/benchmark/tpcds/all_queries/query_01.sql
```diff
@@ -0,0 +1,22 @@
+with customer_total_return as
```
mrks: These queries don't have a random component, correct?

resources/benchmark/tpcds/all_queries/query_01.sql
```diff
@@ -0,0 +1,22 @@
...
+(select sr_customer_sk as ctr_customer_sk
```
mrks: Could you run this through some SQL formatter? This is hard to read.

resources/benchmark/tpcds/tables/call_center.csv.json
```diff
@@ -0,0 +1,168 @@
+{
```
mrks: What are the jsons used for?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,127 @@
...
+  Assert(std::filesystem::is_directory(query_path), "Query path (" + query_path + ") has to be a directory.");
+  Assert(std::filesystem::is_directory(table_path), "Table path (" + table_path + ") has to be a directory.");
+
+  auto query_generator =
+      std::make_unique<FileBasedQueryGenerator>(*config, query_path, query_filename_blacklist, query_subset);
+  auto table_generator = std::make_unique<FileBasedTableGenerator>(config, table_path);
+  auto benchmark_runner = BenchmarkRunner{*config, std::move(query_generator), std::move(table_generator), context};
+
+  if (config->verify) {
+    // TODO(anyone) encapsulate this code somewhere else since it us used for the TPC-DS and JOIN ORDER BENCHMARK
```
mrks: How much effort do you expect this to be?

resources/benchmark/tpcds/schema.sql
```diff
@@ -0,0 +1,552 @@
+create table dbgen_version
```
mrks: The filename should clarify that this is only used for sqlite. Same for the foreign keys.

resources/benchmark/tpcds/all_queries/query_01.sql
mweisgut: Correct.

resources/benchmark/tpcds/tables/call_center.csv.json
mrks: Ah, sorry. I just re-read the PR description.

We will need the tpcds tools package anyway, correct? Should we include it as a submodule now and call dbgen from cpp (similar to what happens with the JOB)?

resources/benchmark/tpcds/all_queries/query_01.sql
mweisgut: Done.

resources/benchmark/tpcds/schema.sql
mweisgut: This table has actually no relevant purpose for us.

mrzzzrm:
https://github.com/hyrise-mp/hyrise-mp19/wiki/How-to-use-the-TPC-DS-tools should probably be migrated to this repo?

mrzzzrm:
FullCI?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,127 @@
...
+      return 0;
+    }
+    scale_factor = cli_parse_result["scale"].as<int>();
+
+    config =
+        std::make_shared<opossum::BenchmarkConfig>(opossum::CLIConfigParser::parse_basic_cli_options(cli_parse_result));
+  }
+
+  Assert(scale_factor == 1, "For now, TPC-DS benchmark only supports scale factor 1");
+  Assert(!config->enable_scheduler, "Multithreaded benchmark execution is currently not supported for TPC-DS");
```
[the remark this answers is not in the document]
mrks: Same for scale_factor. If a limitation is not obvious, please add a comment explaining why it is necessary.

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+  auto context = opossum::BenchmarkRunner::create_context(*config);
+
+  std::cout << "- TPCDS scale factor is " << scale_factor << std::endl;
```
mrks: TPC-DS.

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+  // TPC-DS FileBasedQueryGenerator specification
+  std::optional<std::unordered_set<std::string>> query_subset;
+  const auto query_filename_blacklist = std::unordered_set<std::string>{};
+  std::string query_path = "resources/benchmark/tpcds/queries";
+  std::string table_path = "resources/benchmark/tpcds/tables";
+
+  Assert(std::filesystem::is_directory(query_path), "Query path (" + query_path + ") has to be a directory.");
+  Assert(std::filesystem::is_directory(table_path), "Table path (" + table_path + ") has to be a directory.");
```
mrks: Assert that the expected files are there (use one file per dir as a probe)?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+  auto query_generator =
```
mrks: Stopping the review for now, as this seems quite out-of-date.

src/benchmark/tpcds_benchmark.cpp
mweisgut: It's done now.

resources/benchmark/tpcds/tables/call_center.csv.json
mweisgut: done

mweisgut:
> https://github.com/hyrise-mp/hyrise-mp19/wiki/How-to-use-the-TPC-DS-tools should probably be migrated to this repo?

The updated version c7d06cc29bc83d9b83f6f8c0bedf44ccb6555de6 uses the tpc-ds toolkit to generate the table data files automatically. Also there's no need to do this step manually.

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  std::cout << "done." << std::endl;
+
+  benchmark_runner.run();
+}
+
+bool data_files_available(std::string table_path) {
+  // clang-format off
+  return std::filesystem::exists(table_path + "/call_center.csv") &&
```
mrks:
```
for (const auto& table : [“call_center”, “...”]) {
  if (!std::filesystem::exists(table_path + “/“ + table + “.csv”) return false;
}
return true
```
?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+  Assert(std::filesystem::is_directory(query_path), "Query path (" + query_path + ") has to be a directory.");
+  Assert(std::filesystem::is_directory(table_path), "Table path (" + table_path + ") has to be a directory.");
+  Assert(std::filesystem::exists(std::filesystem::path{query_path + "/query_07.sql"}), "Queries have to be available.");
+  Assert(std::filesystem::exists(std::filesystem::path{table_path + "/call_center.csv.json"}),
+         "Table schemes have to be available.");
+
+  if (!data_files_available(table_path)) {
+    if (std::filesystem::exists(std::filesystem::path{"third_party/tpcds-dsdgen/dsdgen"})) {
+      const auto setup_table_data_files_return_code = system(
```
mrks: Well, I guess I can’t blame you for verbose names. `ret` would have been fine, too ;)

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,142 @@
...
+          "cd ./third_party/tpcds-dsdgen/ &&"
+          "./dsdgen -scale 1 -dir ../../resources/benchmark/tpcds/tables -terminate n -verbose -f &&"
```
mrks: Aren’t you ignoring the scale parameter her?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,142 @@
...
+          "cd ../../resources/benchmark/tpcds/tables &&"
+          "for x in *.dat; do mv $x ${x%.dat}.csv; done &&"
+          "cd ../../../../");
+      Assert(setup_table_data_files_return_code == 0, "Generating table data files failed.");
+    } else {
+      Fail("Could not find 'third_party/tpcds-dsdgen/dsdgen'. Did you run the benchmark from the project root dir?");
+    }
+  }
+
+  Assert(data_files_available(table_path), "Table data files are missing.");
```
mrks: “Missing” sounds like they should have been there. In this case, they failed to be generated

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,142 @@
...
+  // This toolkit does not provide the option to generate data using a scale factor less than 1.
+  Assert(scale_factor >= 1, "For now, TPC-DS benchmark only supports scale factor 1");
+
+  auto context = opossum::BenchmarkRunner::create_context(*config);
+
+  std::cout << "- TPC-DS scale factor is " << scale_factor << std::endl;
+
+  // TPC-DS FileBasedQueryGenerator specification
+  std::optional<std::unordered_set<std::string>> query_subset;
+  const auto query_filename_blacklist = std::unordered_set<std::string>{};
```
mrks: Where is this used?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,142 @@
...
+        opossum::CLIConfigParser::parse_basic_options_json_config(json_config));
+
+  } else {
+    // Parse regular command line args
+    const auto cli_parse_result = cli_options.parse(argc, argv);
+
+    if (CLIConfigParser::print_help_if_requested(cli_options, cli_parse_result)) {
+      return 0;
+    }
+    scale_factor = cli_parse_result["scale"].as<int>();
```
mrks: What happens if I pass in `0.9`? Does this get erroneously rounded to `1` and accepted?

mweisgut: This happens:
```
terminate called after throwing an instance of 'cxxopts::argument_incorrect_type'
  what():  Argument ‘0.9’ failed to parse
```

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -35,11 +35,11 @@ int main(int argc, char* argv[]) {
 
   // clang-format off
   cli_options.add_options()
-    ("s,scale", "Database scale factor (1 ~ 1GB)", cxxopts::value<int>()->default_value("1"));
+    ("s,scale", "Database scale factor (1 ~ 1GB)", cxxopts::value<int32_t>()->default_value("1"));
```
mrks: Could you explain why you changed this to int32_t? `int` means "I just want a reasonably sized integer", `int32_t` says that you actually thought about the width of the int.

mweisgut: Hyrise's [contribution guidelines](https://github.com/hyrise/hyrise/blob/master/CONTRIBUTING.md#c-guidelines) led me to this change: `Use [u]int(8|16|32|64)_t instead of int, long, uint etc.`

mrks: Whoopsie. We might want to revisit that.

mrks: No need to change it here

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,146 @@
...
+int main(int argc, char* argv[]) {
+  auto cli_options = opossum::BenchmarkRunner::get_basic_cli_options("TPC-DS Benchmark");
+
+  // clang-format off
+  cli_options.add_options()
+    ("s,scale", "Database scale factor (1 ~ 1GB)", cxxopts::value<int32_t>()->default_value("1"));
+  // clang-format on
+
+  std::shared_ptr<opossum::BenchmarkConfig> config;
+  int32_t scale_factor;
```
mrzzzrm: in general, prefer the `auto scale_factor = int32_t{}` - this makes sure `scale_factor` always has a defined value (`int32_t{}` is `0`) which, among other things, lessens noise when debugging.

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,146 @@
...
+    add_indices_to_sqlite("resources/benchmark/tpcds/schema.sql", "resources/benchmark/tpcds/create_indices.sql",
+                          benchmark_runner);
+  }
+
+  std::cout << "done." << std::endl;
+
+  benchmark_runner.run();
+}
+
+bool data_files_available(const std::string table_path) {
```
mrzzzrm: * parameter as constref
* should be in anonymous namespace to avoid collisions

src/benchmarklib/sqlite_add_indices.hpp
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include <string>
+
+namespace opossum {
+
+class BenchmarkRunner;
+
+/**
+ * Add indexes to SQLite. This is a hack until we support CREATE INDEX ourselves and pass that on to SQLite.
```
mrzzzrm: pls revise comment, maybe doc params briefly, one might not know what a "create_indices_file" is.

 "Without this, SQLite would never finish" - never finish doing what?
"This is a hack until we support CREATE INDEX ourselves and pass that on to SQLite" - even if we'd support it, it's not a given that we'd use it, as Hyrise functions just as well without it.

mrks:
> We haven't shipped binary executable files alongside Hyrise up until now. Are we okay with this @mrks? Are we sure this binary works on Mac and Ubuntu?

No, just discussed this in the MP meeting

mweisgut:
The added third_party directory tpcds-tools is going to be removed, https://github.com/gregrahn/tpcds-kit will be used instead.

mweisgut:
@mrks I removed `gccDebug` as you suggested.
Check out the [CI log](https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-1679/36/pipeline). It's still the same error.

Do you have any idea how to fix this? Building and executing this on pella or sidon is successfull.

mrks:
Just to rule more things out: Could you create a new PR from this branch, closing this one?

mweisgut:
Done. #1716

PR #1684, opened 2019-05-24
Aggregate Function STDDEV_SAMP (Standard Deviation Sample)

mweisgut:
This PR contains the implementation of the aggregate function Standard Deviation Sample.
For this, the Welford's online algorithm is used.

Excerpt from [1]:
```
variance(samples):
  M := 0
  S := 0
  for k from 1 to N:
    x := samples[k]
    oldM := M
    M := M + (x-M)/k
    S := S + (x-M)*(x-oldM)
  return S/(N-1)
```

[1] Joni Salonen. 2013. _Welford’s method for computing variance_. http://jonisalonen.com/2013/deriving-welfords-method-for-computing-variance/


src/lib/operators/abstract_aggregate_operator.hpp
```diff
@@ -90,19 +93,47 @@ class AggregateFunctionBuilder<ColumnDataType, AggregateType, AggregateFunction:
...
+        } else {
+          current_primary_aggregate = new_value;
+        }
+        // add new value to sum of x²
+        if (current_secondary_aggregate) {
+          *current_secondary_aggregate += new_value * new_value;
+        } else {
+          current_secondary_aggregate = new_value * new_value;
+        }
+      } else {
```
Bensk1: Can we ever reach this branch? This [check](https://github.com/hyrise/hyrise/pull/1684/files#diff-d12958a0e19777d00bbb31d51731b7a4R68) should make it impossible, right?

src/lib/operators/aggregate_sort.cpp
```diff
@@ -186,46 +191,63 @@ void AggregateSort::_aggregate_values(const std::set<RowID>& group_boundaries, c
...
+    if (value_count <= 1) {
+      current_primary_aggregate = std::optional<AggregateType>();
+      current_secondary_aggregate = std::optional<AggregateType>();
+    } else {
+      // normal StdDevSamp calculation
+      const auto sum = *current_primary_aggregate;
+      const auto sum_square = *current_secondary_aggregate;
+      const double n = value_count;
+      const auto variance = (1 / (n - 1)) * (sum_square - (1 / n) * sum * sum);
+      current_primary_aggregate = std::sqrt(variance);
```
Bensk1: This is the first aggregate functions that is a bit more complex which makes the redundancy of the computations of the aggregate functions in both aggregate operators quite obvious now.

src/lib/operators/abstract_aggregate_operator.hpp
mweisgut: That's right. But we still have to do the `is_arithmetic` check here, since `ColumnDataType` could be string. So if the type is string, `new_value * new_value` would be invalid.

Is there a way to restrict the type on the template-type level?

src/lib/operators/jit_operator/operators/jit_aggregate.cpp
```diff
@@ -219,6 +219,8 @@ void JitAggregate::add_aggregate_column(const std::string& column_name, const Ji
       break;
     case AggregateFunction::CountDistinct:
       Fail("Aggregate function count distinct not supported");
+    case AggregateFunction::StdDevSamp:
+      Fail("Aggregate function standard deviation sample is not supported.");
```
mrzzzrm: Having this here means that the `JitAwareLQPTranslator` needs to be adjusted to not abort if this AggregateFunction is encountered. 

`JitAwareLQPTranslator::_expression_is_jittable()` should use a whitelist of supported AggregateFunctions, instead of a blacklist.

mweisgut:
> The formula used in the OP is slightly different from the one used in code.

What's OP?

mrzzzrm:
"OP. n. abrv. Original Post - The first post in a forum or imageboard thread "

mweisgut:
> The formula used in the OP is slightly different from the one used in code.
> `sqrt(1 / (n - 1)) * (sum_square - (1 / n) * sum * sum)` vs `sqrt((1 / (n - 1)) * (sum_square - (1 / n) * sum * sum))` afaict.
> 
> Also, I tried for a bit and I couldn't follow how you got from
> image
> to the formula used.
> 
> Can you boil the transformations for this down to a short comment, in code?
> 
> The duplication of the StdDevSample is unfortunate, and though it isn't you fault it's there it should be avoided to some degree. Can we have a function in `AbstractAggregateOperator` doing the math?

You were right, I forgot the outer brackets in the OP. After you mentioned the obscurity of the formula, I've been researched again and found the Welford's online algorithm [1][2] which is more accurate than the formula I used before. Now, I don't have to modify the Epsilon:

https://github.com/hyrise/hyrise/blob/814beafdc9480deb5050d4dc40fc3d8525337204/src/lib/utils/check_table_equal.cpp#L17

[1] http://jonisalonen.com/2013/deriving-welfords-method-for-computing-variance/
[2] https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm

src/lib/operators/jit_operator/operators/jit_aggregate.cpp
mweisgut: https://github.com/hyrise/hyrise/blob/3403d5f29d1e3f9d0670e3adb93c8cc39caaae53/src/lib/logical_query_plan/jit_aware_lqp_translator.cpp#L433-L464

src/lib/expression/aggregate_expression.hpp
```diff
@@ -4,7 +4,7 @@
 
 namespace opossum {
 
-enum class AggregateFunction { Min, Max, Sum, Avg, Count, CountDistinct };
+enum class AggregateFunction { Min, Max, Sum, Avg, Count, CountDistinct, StdDevSamp };
```
mrzzzrm: I'd spell this out, for consistency. Also "Dev" makes me think of "Developer".

src/lib/expression/expression_functional.hpp
```diff
@@ -121,6 +121,7 @@ inline detail::unary<AggregateFunction::Min, AggregateExpression> min_;
 inline detail::unary<AggregateFunction::Avg, AggregateExpression> avg_;
 inline detail::unary<AggregateFunction::Count, AggregateExpression> count_;
 inline detail::unary<AggregateFunction::CountDistinct, AggregateExpression> count_distinct_;
+inline detail::unary<AggregateFunction::StdDevSamp, AggregateExpression> stddev_samp_;
```
mrzzzrm: here, too. I'd spell this out (yeah, I know "avg_" isn't either...) or at least make it "std_dev_samp_"

src/lib/operators/abstract_aggregate_operator.hpp
```diff
@@ -90,19 +93,73 @@ class AggregateFunctionBuilder<ColumnDataType, AggregateType, AggregateFunction:
...
+        // The sample standard deviation is stored as current_primary_aggregate.
+        if (current_secondary_aggregates.empty()) {
+          current_secondary_aggregates.resize(3);
+          // current_secondary_aggregates[0]:   storage for count
+          // current_secondary_aggregates[1]:   storage for mean
+          // current_secondary_aggregates[2]:   storage for squared_distance_from_mean
+        }
+
+        // get values
+        double count = current_secondary_aggregates[0];
```
mrzzzrm: auto. Here and below.

src/lib/operators/abstract_aggregate_operator.hpp
mrzzzrm: `if constexpr(...) {....} else { Fail("this shouldn't be reached"); }` is how I'd do it as well. Done like this in a couple of places.

src/lib/operators/abstract_aggregate_operator.hpp
```diff
@@ -53,10 +54,11 @@ template <typename ColumnDataType, typename AggregateType>
 class AggregateFunctionBuilder<ColumnDataType, AggregateType, AggregateFunction::Max> {
  public:
   auto get_aggregate_function() {
-    return [](const ColumnDataType& new_value, std::optional<AggregateType>& current_aggregate) {
-      if (!current_aggregate || value_greater(new_value, *current_aggregate)) {
+    return [](const ColumnDataType& new_value, std::optional<AggregateType>& current_primary_aggregate,
+              std::vector<AggregateType>& current_secondary_aggregates) {
```
mrzzzrm: urgs, a to-heap pointer for each aggregate requiring more than one value is potentially slow, just wanted to point this out.

src/lib/operators/abstract_aggregate_operator.hpp
```diff
@@ -90,19 +93,73 @@ class AggregateFunctionBuilder<ColumnDataType, AggregateType, AggregateFunction:
...
+        if (current_secondary_aggregates.empty()) {
+          current_secondary_aggregates.resize(3);
+          // current_secondary_aggregates[0]:   storage for count
+          // current_secondary_aggregates[1]:   storage for mean
+          // current_secondary_aggregates[2]:   storage for squared_distance_from_mean
+        }
+
+        // get values
+        double count = current_secondary_aggregates[0];
+        double mean = current_secondary_aggregates[1];
```
mrzzzrm: how about making this a reference so you don't need the "store values" step below.

src/lib/operators/aggregate_hash.cpp
```diff
@@ -569,6 +574,40 @@ std::enable_if_t<func == AggregateFunction::Avg && !std::is_arithmetic_v<Aggrega
...
+  auto& null_values = segment->null_values();
+
+  values.resize(results.size());
+  null_values.resize(results.size());
+
+  size_t i = 0;
+  for (const auto& result : results) {
+    const auto count = static_cast<AggregateType>(result.aggregate_count);
+
+    if (result.current_primary_aggregate && count > 1) {
```
mrzzzrm: so the standard deviation sample of a single sample is undefined?

src/lib/operators/aggregate_hash.cpp
```diff
@@ -569,6 +574,40 @@ std::enable_if_t<func == AggregateFunction::Avg && !std::is_arithmetic_v<Aggrega
...
+                       const AggregateResults<ColumnDataType, AggregateType>& results) {
+  DebugAssert(segment->is_nullable(), "Aggregate: Output segment needs to be nullable");
+
+  auto& values = segment->values();
+  auto& null_values = segment->null_values();
+
+  values.resize(results.size());
+  null_values.resize(results.size());
+
+  size_t i = 0;
```
mrzzzrm: I know it is pedantic, but we normally don't use "i" as a variable name and try to come up with something more speaking. in this case "chunk_offset"?

src/lib/operators/aggregate_hash.cpp
mweisgut: Correctly. Since the formula contains a devision by `N - 1`, using only one sample (`N = 1`) leads to a division by zero.

image

mweisgut:
> LGTM. One request tough: Could you run the TPC-H benchmark with a reasonably high SF (I tend to use 5) to see what the performance impact of the increased memory footprint of AggregateResult is, if any?

Benchmark comparison, SF 5.  
Benchmark 1 version: `cefb5fb70dfc0a6e43f35d7dd098d7ac703d4602`
Benchmark 2 version: `1ab72aacfa03b0f1bff8d00579036ea40431f2a1`

```
+----------------+-----------------+------+-----------------+------+------------+---------------------------------+
| Benchmark      | prev. iter/s    | runs | new iter/s      | runs | change [%] | p-value (significant if <0.001) |
+----------------+-----------------+------+-----------------+------+------------+---------------------------------+
| TPC-H 1        | 0.0399382896721 | 3    | 0.0410519689322 | 3    | +3%        |        (not enough runs) 0.2671 |
| TPC-H 2        | 0.495163530111  | 30   | 0.486336022615  | 30   | -2%        |                          0.0678 |
| TPC-H 3        | 0.252941161394  | 16   | 0.250755548477  | 16   | -1%        |                          0.5180 |
| TPC-H 4        | 0.143066257238  | 9    | 0.148081675172  | 9    | +4%        |        (not enough runs) 0.0000 |
| TPC-H 5        | 0.181870356202  | 11   | 0.182223290205  | 11   | +0%        |                          0.8916 |
| TPC-H 6        | 0.652892291546  | 40   | 0.680475354195  | 41   | +4%        |                          0.0029 |
| TPC-H 7        | 0.0767786800861 | 5    | 0.0783066004515 | 5    | +2%        |        (not enough runs) 0.2026 |
| TPC-H 8        | 0.236355766654  | 15   | 0.240312665701  | 15   | +2%        |                          0.3286 |
| TPC-H 9        | 0.0908612310886 | 6    | 0.0941350385547 | 6    | +4%        |        (not enough runs) 0.0092 |
| TPC-H 10       | 0.180481538177  | 11   | 0.183996617794  | 12   | +2%        |                          0.1360 |
| TPC-H 11       | 1.3590015173    | 82   | 1.36057126522   | 82   | +0%        |                          0.7167 |
| TPC-H 12       | 0.32452622056   | 20   | 0.329913437366  | 20   | +2%        |                          0.0000 |
| TPC-H 13       | 0.157454803586  | 10   | 0.15432164073   | 10   | -2%        |                          0.1714 |
| TPC-H 14       | 0.51120197773   | 31   | 0.521532893181  | 32   | +2%        |                          0.0505 |
| TPC-H 15       | 0.290382087231  | 18   | 0.28532359004   | 18   | -2%        |                          0.3071 |
| TPC-H 16       | 0.583827912807  | 36   | 0.604969263077  | 37   | +4%        |                          0.0007 |
| TPC-H 17       | 0.0586458072066 | 4    | 0.0584707669914 | 4    | -0%        |        (not enough runs) 0.8203 |
| TPC-H 18       | 0.100078396499  | 7    | 0.0975055396557 | 6    | -3%        |        (not enough runs) 0.0659 |
| TPC-H 19       | 0.409562379122  | 25   | 0.413717091084  | 25   | +1%        |                          0.2556 |
| TPC-H 20       | 0.116328492761  | 7    | 0.115614049137  | 7    | -1%        |        (not enough runs) 0.4198 |
| TPC-H 21       | 0.0260674729943 | 2    | 0.0273434575647 | 2    | +5%        |        (not enough runs) 0.0773 |
| TPC-H 22       | 0.628851771355  | 38   | 0.63364470005   | 39   | +1%        |                          0.0244 |
| geometric mean |                 |      |                 |      | +1%        |                                 |
+----------------+-----------------+------+-----------------+------+------------+---------------------------------+
```

mrzzzrm:
A performance *increase*? Mmh, how would that come to be?

Then again, `git log --oneline | grep cefb5fb70dfc0a6e43f35d7dd098d7ac703d4602` is empty for me - did you really measure against the master?

mrks:
If you left the \` away, you’d get a link for free: cefb5fb70dfc0a6e43f35d7dd098d7ac703d4602
I guess this is the MP master that this is based on.

At least, it did not affect the performance negatively. A change in the range of a couple of percents might be everything, including measurement tolerances or changed alignment. I’ve almost given up on trying to understand that.

I’d go ahead and merge this if you don’t have any objections.

PR #1686, opened 2019-05-26
Fix handling of nullptrs in calls of Table::get_chunk()

julianmenzler:
In `TableType::Data` tables, the implementation and use of the MvccDeletePlugin (#1493) leads to null pointers in `_chunks`. This PR focuses on null-pointer-handling, thread safety and code consistency regarding chunk-objects.

In detail:

1. Read-/Write access to `shared_ptr<Chunk>` objects of `_chunks` is now handled via `std::atomic_load` and `std::atomic_store` to prevent data races. (in case of `TableType::Data` tables)

2. At critical points in code, Asserts are implemented to check for null pointers returned by `get_chunk()`. Usually, they should not become a problem since they're filtered out by the `GetTable` operator.
Sometimes we do not use `GetTable` (when using the `Print` operator for example). In this case, we now check for null pointers and skip them if present.

3. Declarations of `Chunk` in conjunction with `get_chunk()` were streamlined. We should always use `auto chunk` (no reference!) to claim ownership of chunk-objects. Also we use`const auto chunk` whenever possible.

4. Some calls of `Table::chunk_count()` have been drawn out of for-loops for better performance.


src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -1539,7 +1541,9 @@ std::vector<std::shared_ptr<ExpressionResult<Result>>> ExpressionEvaluator::_pru
       }
     } else {
       for (auto chunk_id = ChunkID{0}; chunk_id < table->chunk_count(); ++chunk_id) {
-        const auto& result_segment = *table->get_chunk(chunk_id)->get_segment(ColumnID{0});
+        const auto& chunk = table->get_chunk(chunk_id);
+        if (!chunk) continue;
+        const auto& result_segment = *chunk->get_segment(ColumnID{0});
```
mrks: Throw in a newline? Same in 1531?

src/lib/operators/aggregate_hash.cpp
```diff
@@ -228,7 +230,8 @@ void AggregateHash::_aggregate() {
         AggregateKeyEntry id_counter = 1u;
 
         for (ChunkID chunk_id{0}; chunk_id < input_table->chunk_count(); ++chunk_id) {
-          const auto chunk_in = input_table->get_chunk(chunk_id);
+          const auto& chunk_in = input_table->get_chunk(chunk_id);
+          if (!chunk_in) continue;
```
mrks: newline?

src/test/operators/join_test_runner.cpp
```diff
@@ -167,9 +167,9 @@ class JoinTestRunner : public BaseTestWithParam<JoinTestConfiguration> {
     const auto all_input_table_types =
         std::vector{InputTableType::Data, InputTableType::IndividualPosLists, InputTableType::SharedPosList};
 
-    const auto all_encoding_types =
-        std::vector{EncodingType::Unencoded, EncodingType::Dictionary, EncodingType::FrameOfReference,
-                    EncodingType::FixedStringDictionary, EncodingType::RunLength, EncodingType::LZ4};
+    const auto all_encoding_types = std::vector{EncodingType::Unencoded,        EncodingType::Dictionary,
+                                                EncodingType::FrameOfReference, EncodingType::FixedStringDictionary,
+                                                EncodingType::RunLength,        EncodingType::LZ4};
```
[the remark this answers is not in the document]
julianmenzler: You're right, my version is too recent.
```
clang-format version 8.0.0 (tags/RELEASE_800/final)
```

src/test/tasks/chunk_compression_task_test.cpp
```diff
@@ -63,7 +63,7 @@ TEST_F(ChunkCompressionTaskTest, DictionarySize) {
   auto dictionary_sizes = std::array<std::vector<size_t>, chunk_count>{{{3u, 3u}, {2u, 3u}}};
 
   for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
-    auto chunk = table_dict->get_chunk(chunk_id);
+    const auto& chunk = table_dict->get_chunk(chunk_id);
```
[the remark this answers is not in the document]
julianmenzler: Thanks for the article! Improved my understanding. I was aware of `get_chunk` returning a temporary. But somehow I believed we were inducing copy by value. 
At least we' re more consistent now.

src/lib/storage/storage_manager.cpp
```diff
@@ -21,7 +21,10 @@ void StorageManager::add_table(const std::string& name, std::shared_ptr<Table> t
   Assert(_views.find(name) == _views.end(), "Cannot add table " + name + " - a view with the same name already exists");
 
   for (ChunkID chunk_id{0}; chunk_id < table->chunk_count(); chunk_id++) {
-    Assert(table->get_chunk(chunk_id)->has_mvcc_data(), "Table must have MVCC data.");
+    const auto& chunk = table->get_chunk(chunk_id);
+    if (chunk) {
```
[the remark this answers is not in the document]
julianmenzler: True.

src/lib/statistics/table_statistics.cpp
```diff
@@ -287,7 +287,9 @@ TableStatistics TableStatistics::estimate_predicated_join(const TableStatistics&
       apply_right_outer_join();
       break;
     }
-    default: { Fail("Join mode not implemented."); }
+    default: {
+      Fail("Join mode not implemented.");
```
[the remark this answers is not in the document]
julianmenzler: Ok, I'll respect that in the future.

src/lib/operators/table_scan.cpp
```diff
@@ -93,9 +93,13 @@ std::shared_ptr<const Table> TableScan::_on_execute() {
 
   for (ChunkID chunk_id{0u}; chunk_id < in_table->chunk_count(); ++chunk_id) {
     if (excluded_chunk_set.count(chunk_id)) continue;
+    const auto& chunk_in = in_table->get_chunk(chunk_id);
+    if (!chunk_in) continue;
 
     auto job_task = std::make_shared<JobTask>([=, &output_mutex, &output_chunks]() {
-      const auto chunk_guard = in_table->get_chunk(chunk_id);
+      const auto& chunk_guard = in_table->get_chunk(chunk_id);
```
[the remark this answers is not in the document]
julianmenzler: I don't know if this is possible, but following would be a worst-case scenario: 

1. Chunk available → `job_task` gets created
2. Chunk is physically deleted
3. `job_task` starts with nullpointer.

Therefore, it makes sense to place a chunk-guard

* before the construction of a `job_task` to avoid "empty" jobs
* at the beginning of a `job_task` to ensure null-pointer safety.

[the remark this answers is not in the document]
mrks: But isn't this the same as `chunk_in`? Why don't we simply **copy** that into the lambda?

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -113,11 +117,17 @@ RadixContainer<T> materialize_input(const std::shared_ptr<const Table>& in_table
   jobs.reserve(in_table->chunk_count());
 
   for (ChunkID chunk_id{0}; chunk_id < in_table->chunk_count(); ++chunk_id) {
-    jobs.emplace_back(std::make_shared<JobTask>([&, chunk_id]() {
+    const auto& chunk = in_table->get_chunk(chunk_id);
+    if (!chunk) continue;
+
+    jobs.emplace_back(std::make_shared<JobTask>([&, in_table, chunk_id]() {
+      const auto& chunk_in = in_table->get_chunk(chunk_id);
+      if (!chunk) return;
```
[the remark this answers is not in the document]
julianmenzler: Correct!

src/lib/operators/delete.cpp
```diff
@@ -49,7 +49,11 @@ std::shared_ptr<const Table> Delete::_on_execute(std::shared_ptr<TransactionCont
                 "All segments of a Chunk in _referencing_table must have the same PosList");
 
     for (auto row_id : *pos_list) {
-      auto referenced_chunk = first_segment->referenced_table()->get_chunk(row_id.chunk_id);
+      const auto& referenced_chunk = first_segment->referenced_table()->get_chunk(row_id.chunk_id);
+      if (!referenced_chunk) {
+        _mark_as_failed();
```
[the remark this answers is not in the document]
julianmenzler: Well, bad things happen. But I rethought this:
Assuming our `MvccDeletePlugin` works as expected this is not going to happen: Our physical delete operations starts only when there are no other conflicting transactions.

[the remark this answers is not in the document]
mrks: In those places where we found a `get_chunk` but don't expect one, we should have an Assert. I'd argue it's cheap enough to be a full assert.

src/lib/operators/table_scan.cpp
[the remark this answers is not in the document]
julianmenzler: Definitely a possibility. In consequnce, we would claim ownership of the chunk in the time between job creation and execution. 
Our MvccDeletePlugin could be blocked by this in terms of physical delete. Otherwise, `TableScan` works with a TransactionContext. So conflicts are prevented anyhow.

src/lib/operators/delete.cpp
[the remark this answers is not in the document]
julianmenzler: Ok 👍

src/lib/operators/table_scan.cpp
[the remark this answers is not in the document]
mrks: Not sure if I understand what this finally means. Are you waiting for a response or are you planning to change this. If so, how?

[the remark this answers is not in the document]
julianmenzler: Sorry, I just wrote down my thoughts. Nevermind.
I'll remove the check within the `job_task`-lambda. One check is enough. Our MvccDeletePlugin won't cause any conflicts.

src/test/operators/join_test_runner.cpp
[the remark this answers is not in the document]
julianmenzler: I ran 
```
clang-format version 7.0.1 (tags/RELEASE_701/final)
```
but this does not change back.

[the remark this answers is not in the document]
mrks: That's fine

src/lib/operators/union_positions.cpp
```diff
@@ -225,7 +225,7 @@ std::shared_ptr<const Table> UnionPositions::_prepare_operator() {
    */
   const auto add = [&](const auto& table) {
     auto current_pos_list = std::shared_ptr<const PosList>();
-    const auto first_chunk = table->get_chunk(ChunkID{0});
+    const auto& first_chunk = table->get_chunk(ChunkID{0});
```
mrks: Why don't you check for nullptr in this file?

src/lib/operators/join_nested_loop.cpp
```diff
@@ -206,7 +214,9 @@ std::shared_ptr<const Table> JoinNestedLoop::_on_execute() {
     const auto invert = _mode == JoinMode::AntiNullAsFalse || _mode == JoinMode::AntiNullAsTrue;
 
     for (auto chunk_id = ChunkID{0}; chunk_id < left_table->chunk_count(); ++chunk_id) {
-      const auto chunk_size = left_table->get_chunk(chunk_id)->size();
+      const auto& chunk_left = left_table->get_chunk(chunk_id);
+      if (!chunk_left) continue;
```
mrks: I'd throw in some new lines in the changes to this file, too

src/lib/operators/index_scan.cpp
```diff
@@ -42,12 +42,16 @@ std::shared_ptr<const Table> IndexScan::_on_execute() {
   if (_included_chunk_ids.empty()) {
     jobs.reserve(_in_table->chunk_count());
     for (auto chunk_id = ChunkID{0u}; chunk_id < _in_table->chunk_count(); ++chunk_id) {
-      jobs.push_back(_create_job_and_schedule(chunk_id, output_mutex));
+      if (_in_table->get_chunk(chunk_id)) {
+        jobs.push_back(_create_job_and_schedule(chunk_id, output_mutex));
```
mrks: This still looks unsafe. Just because the pointer exists right now, it is not guaranteed to exist later.

src/lib/operators/delete.cpp
```diff
@@ -49,7 +49,8 @@ std::shared_ptr<const Table> Delete::_on_execute(std::shared_ptr<TransactionCont
                 "All segments of a Chunk in _referencing_table must have the same PosList");
 
     for (auto row_id : *pos_list) {
-      auto referenced_chunk = first_segment->referenced_table()->get_chunk(row_id.chunk_id);
+      const auto& referenced_chunk = first_segment->referenced_table()->get_chunk(row_id.chunk_id);
+      Assert(referenced_chunk, "Referenced chunks are not allowed to be nullpointers");
```
mrks: null pointers

src/lib/operators/delete.cpp
```diff
@@ -89,13 +90,13 @@ std::shared_ptr<const Table> Delete::_on_execute(std::shared_ptr<TransactionCont
 void Delete::_on_commit_records(const CommitID cid) {
   for (ChunkID referencing_chunk_id{0}; referencing_chunk_id < _referencing_table->chunk_count();
        ++referencing_chunk_id) {
-    const auto referencing_chunk = _referencing_table->get_chunk(referencing_chunk_id);
+    const auto& referencing_chunk = _referencing_table->get_chunk(referencing_chunk_id);
```
mrks: I just realized that `const auto&` is wrong here. Imagine `get_chunk` returned a reference. With this change, we would not copy the `shared_ptr` anymore, so we have no guarantee that the object is not suddenly deleted.

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -113,11 +117,17 @@ RadixContainer<T> materialize_input(const std::shared_ptr<const Table>& in_table
   jobs.reserve(in_table->chunk_count());
 
   for (ChunkID chunk_id{0}; chunk_id < in_table->chunk_count(); ++chunk_id) {
-    jobs.emplace_back(std::make_shared<JobTask>([&, chunk_id]() {
+    const auto& chunk_guard = in_table->get_chunk(chunk_id);
```
mrks: ?

julianmenzler: > Unfortunately, I found some more points. This looks really hard to get right. Luckily, a failure to do so is quite noticeable when we dereference a nullptr

Ok. I'll look into this later. For now, time is up 😅

julianmenzler:
> Unfortunately, I found some more points. This looks really hard to get right. Luckily, a failure to do so is quite noticeable when we dereference a nullptr

Ok. I'll look into this later. For now, time is up 😅

src/lib/operators/union_positions.cpp
julianmenzler: Because of this Assert:
```cpp
/**
   * Both tables must contain only ReferenceSegments
   */
  Assert(input_table_left()->type() == TableType::References && input_table_right()->type() == TableType::References,
         "UnionPositions doesn't support non-reference tables yet");
```

src/lib/operators/index_scan.cpp
julianmenzler: Correct. There should be an additional check in the job itself.

src/lib/operators/join_hash/join_hash_steps.hpp
julianmenzler: Creating jobs for null-pointer-chunks does not make sense in my opinion.

Bouncner: chunk_guard is a strange name and why not pass it to the job? And then again, why the same check in line 125?

mrzzzrm:
Also, `std::shared_ptr` is, to the best of my knowledge, not thread safe. Multiple threads reading from it to do `if (!chunk) continue;` is a data race with the thread doing the `chunk.reset()` call.

src/benchmark/operators/join_benchmark.cpp
```diff
@@ -44,7 +44,8 @@ std::shared_ptr<TableWrapper> generate_table(const size_t number_of_rows) {
                                                EncodingType::Dictionary);
 
   for (ChunkID chunk_id{0}; chunk_id < table->chunk_count(); ++chunk_id) {
-    auto chunk = table->get_chunk(chunk_id);
+    const auto& chunk = table->get_chunk(chunk_id);
```
mrzzzrm: @julianmenzler are you aware you're using one very obscure c++ feature, here? (https://abseil.io/tips/107)

One of @mrks favourites, though.

mrks:
> The only Operator that should have to deal with Tables from the StorageManager is GetTable. Wouldn't it be better if GetTable always builds a new Table with all "living" Chunks so that all other operators can ignore the possibility of a Chunk being nullptr?

For operators, that's a good idea and we should do it. Especially now that GetTable also does chunk and column pruning. Adding this to GetTable, however, does not solve the problem entirely. Example: https://github.com/hyrise/hyrise/blob/438b52e1dffcf3360b634c6a1eb138aa35132462/src/lib/optimizer/strategy/chunk_pruning_rule.cpp#L63

Also, std::shared_ptr is, to the best of my knowledge, not thread safe. Multiple threads reading from it to do if (!chunk) continue; is a data race with the thread doing the chunk.reset() call.

Not true, unless I misunderstood you:

> All member functions (including copy constructor and copy assignment) can be called by multiple threads on different instances of shared_ptr without additional synchronization even if these instances are copies and share ownership of the same object. If multiple threads of execution access the same shared_ptr without synchronization and any of those accesses uses a non-const member function of shared_ptr then a data race will occur; the shared_ptr overloads of atomic functions can be used to prevent the data race.

src/benchmark/operators/join_benchmark.cpp
mrks: We talked about this in person. This is a bad example for lifetime extension, as we actually do want a copy of the shared_ptr. If `get_chunk` for some reason returned a reference, we could not guarantee that the original shared_ptr does not get destroyed somewhere in between.

mrzzzrm:
The data race I am suspecting - not being entirely sure - is the following: T1 (MvccPlugin) calls "Table::chunk[x].reset()", right? T2 calls "Table::get_chunk(x)", which reads from *the same instance* as T1 writes to.

mrks:
Bloody hell.

julianmenzler:
Too bad. So I guess we have to use the following construct then?

```cpp
template <class T> struct std::atomic<std::shared_ptr<T>>; // (since C++20)
```

> If multiple threads of execution access the same std::shared_ptr object without synchronization and any of those accesses uses a non-const member function of shared_ptr then a data race will occur **unless all such access is performed through an instance of std::atomic<std::shared_ptr>>** (or, deprecated as of C++20, through the standalone functions for atomic access to std::shared_ptr).

mrzzzrm:
I don't think we could afford storing all Chunks in all Tables in atomic shared ptrs, judging from the gist that I remember about @lanice's MT defense. Parallelizability will probably suffer (can be tested by running benchmarks with multiple clients (--clients)). 

Don't have a solution at hand, just here to point out potential problems :(

mrks:
In `Chunk::get_segment` we use this:

```
std::shared_ptr<BaseSegment> Chunk::get_segment(ColumnID column_id) const {
  return std::atomic_load(&_segments.at(column_id));
}
```

I wonder what the cost of that is. Maybe we should do three measurements:

1) Status quo
2) Remove the atomic load
3) Add an atomic load to `Table::get_chunk`.

Setting up a benchmark for this should take five minutes, so that's worth a look. Remember to use a machine with multiple NUMA nodes and set `--scheduler` and `--clients`.

julianmenzler:
> In `Chunk::get_segment` we use this:
> 
> ```
> std::shared_ptr<BaseSegment> Chunk::get_segment(ColumnID column_id) const {
>   return std::atomic_load(&_segments.at(column_id));
> }
> ```
> 
> I wonder what the cost of that is. Maybe we should do three measurements:
> 
> 1. Status quo
> 2. Remove the atomic load
> 3. Add an atomic load to `Table::get_chunk`.
> 
> Setting up a benchmark for this should take five minutes, so that's worth a look. Remember to use a machine with multiple NUMA nodes and set `--scheduler` and `--clients`.

@mrks Is this a job for me? I'd like to try this, but I do not have access to a NUMA system yet.

mrks:
For things like this, a more modern CPU (like nemea) is preferred.

mrks:
(If you have already started, no need to migrate. Just letting you know for the future)

julianmenzler:
I used the following benchmarks parameters on nemea: `--scheduler --clients 100 -s 1 -t 120`

1. Removing `atomic_load` in `Chunk::get_segment` results in:
```
| TPC-H 1        | 0.858296692371 | 103  | 0.908296525478 | 109  | +6%        |                          0.5683 |
| TPC-H 2        | 19.1402301788  | 2297 | 18.332742691   | 2200 | -4%        |                          0.3958 |
| TPC-H 3        | 13.9078998566  | 1669 | 14.1155529022  | 1694 | +1%        |                          0.5458 |
| TPC-H 4        | 9.82484912872  | 1179 | 9.82434082031  | 1179 | -0%        |                          0.2374 |
| TPC-H 5        | 9.94986057281  | 1194 | 10.2242794037  | 1227 | +3%        |                          0.1578 |
| TPC-H 6        | 21.6318454742  | 2596 | 21.6654472351  | 2600 | +0%        |                          0.9392 |
| TPC-H 7        | 6.46631193161  | 776  | 6.49988126755  | 780  | +1%        |                          0.5219 |
| TPC-H 8        | 13.2490205765  | 1590 | 12.9830303192  | 1558 | -2%        |                          0.2157 |
| TPC-H 9        | 7.49945545197  | 900  | 7.60800647736  | 913  | +1%        |                          0.5740 |
| TPC-H 10       | 16.057756424   | 1927 | 15.8486776352  | 1902 | -1%        |                          0.4039 |
| TPC-H 11       | 37.9158821106  | 4550 | 37.9313011169  | 4552 | +0%        |                          0.8468 |
| TPC-H 12       | 16.6070804596  | 1993 | 16.8824996948  | 2026 | +2%        |                          0.2106 |
| TPC-H 13       | 14.6908845901  | 1763 | 14.8159914017  | 1778 | +1%        |                          0.5150 |
| TPC-H 14       | 22.1069393158  | 2653 | 21.4736251831  | 2577 | -3%        |                          0.0272 |
| TPC-H 15       | 13.4999160767  | 1620 | 13.9240694046  | 1671 | +3%        |                          0.4061 |
| TPC-H 16       | 23.1158752441  | 2774 | 22.4231586456  | 2691 | -3%        |                          0.1176 |
| TPC-H 17       | 4.29965162277  | 516  | 4.4080414772   | 529  | +3%        |                          0.7827 |
| TPC-H 18       | 3.83302474022  | 460  | 3.84977722168  | 462  | +0%        |                          0.5319 |
| TPC-H 19       | 23.1580734253  | 2779 | 23.3488559723  | 2802 | +1%        |                          0.6246 |
| TPC-H 20       | 8.74137020111  | 1049 | 8.88276863098  | 1066 | +2%        |                          0.7041 |
| TPC-H 21       | 0.833301603794 | 100  | 0.816630959511 | 98   | -2%        |                          0.0000 |
| TPC-H 22       | 19.9903736115  | 2399 | 41.1309547424  | 4936 | +106%      |                          0.0000 |
| geometric mean |                |      |                |      | +4%        |                                 |
```

2. Adding `atomic load` in `Table::get_chunk` results in:

```
| TPC-H 1        | 0.858296692371 | 103  | 0.808322310448 | 97   | -6%        |                          0.0856 |
| TPC-H 2        | 19.1402301788  | 2297 | 19.3332633972  | 2320 | +1%        |                          0.7970 |
| TPC-H 3        | 13.9078998566  | 1669 | 13.5491933823  | 1626 | -3%        |                          0.1416 |
| TPC-H 4        | 9.82484912872  | 1179 | 9.90793800354  | 1189 | +1%        |                          0.4243 |
| TPC-H 5        | 9.94986057281  | 1194 | 9.7415304184   | 1169 | -2%        |                          0.6964 |
| TPC-H 6        | 21.6318454742  | 2596 | 21.5580558777  | 2587 | -0%        |                          0.9979 |
| TPC-H 7        | 6.46631193161  | 776  | 6.54990148544  | 786  | +1%        |                          0.8948 |
| TPC-H 8        | 13.2490205765  | 1590 | 12.6496677399  | 1518 | -5%        |                          0.0193 |
| TPC-H 9        | 7.49945545197  | 900  | 7.57495927811  | 909  | +1%        |                          0.5690 |
| TPC-H 10       | 16.057756424   | 1927 | 15.6248311996  | 1875 | -3%        |                          0.1843 |
| TPC-H 11       | 37.9158821106  | 4550 | 37.2890739441  | 4475 | -2%        |                          0.5170 |
| TPC-H 12       | 16.6070804596  | 1993 | 16.7652759552  | 2012 | +1%        |                          0.7375 |
| TPC-H 13       | 14.6908845901  | 1763 | 14.8242292404  | 1779 | +1%        |                          0.5410 |
| TPC-H 14       | 22.1069393158  | 2653 | 21.4824848175  | 2578 | -3%        |                          0.0142 |
| TPC-H 15       | 13.4999160767  | 1620 | 13.7324571609  | 1648 | +2%        |                          0.4930 |
| TPC-H 16       | 23.1158752441  | 2774 | 23.0833034515  | 2770 | -0%        |                          0.8151 |
| TPC-H 17       | 4.29965162277  | 516  | 4.40828132629  | 529  | +3%        |                          0.3837 |
| TPC-H 18       | 3.83302474022  | 460  | 3.61651492119  | 434  | -6%        |                          0.7772 |
| TPC-H 19       | 23.1580734253  | 2779 | 23.0482254028  | 2766 | -0%        |                          0.7883 |
| TPC-H 20       | 8.74137020111  | 1049 | 8.8164396286   | 1058 | +1%        |                          0.7699 |
| TPC-H 21       | 0.833301603794 | 100  | 0.766653478146 | 92   | -8%        |                          0.4552 |
| TPC-H 22       | 19.9903736115  | 2399 | 20.2400455475  | 2429 | +1%        |                          0.5799 |
| geometric mean |                |      |                |      | -1%        |                                 |
```

mrks:
That's surprisingly affordable. Should we go down that road?

Also, what's happening with TPC-H 22 in the first benchmark? Opened #1709.

mrks:
Just realized that we can easily check if we're in a stored table and do the atomic check only then.  Reference tables should have no nullptrs.

julianmenzler:
> That's surprisingly affordable. Should we go down that road?
> 
> Also, what's happening with TPC-H 22 in the first benchmark? Opened #1709.

If it solves our MvccDelete multithreading issues, it is absolutely feasible I would say.

Regarding TPC-H 22: The performance gain is real. I was quite irritated as well, so I doublechecked it with another benchmark run.

src/lib/operators/join_hash/join_hash_steps.hpp
julianmenzler: Addressed. Now it is `chunk_in`

julianmenzler:
I tried to implement `std::atomic<std::shared_ptr<Chunk>>` as an element type for `_chunks` in class `Table`. It is the C++20 way of dealing with our problem. Unfortunately, this does not seem to be implemented yet.

Therefore, I adopted the atomics functions `std::atomic_load()` and `std::atomic_store()`. We have to use these functions (`std::atomic_load(&chunk)`) each time we access or modify a shared pointer from Table's `_chunks` vector to preserve thread-safety. It is a very error-prone approach since we might forget about using atomics functions and the compiler won't complain. This is why they become deprecated as of C++20.

So far, I have implemented `std::atomic_load` and ``std::atomic_store` in our `Table`-class. The following function extends the problem scope though:

```
const tbb::concurrent_vector<std::shared_ptr<Chunk>>& chunks() const;
```

It makes the `_chunks` vector available to the following classes:
Untitled

These classes also have to adopt `atomic_load` for thread-safety. I stopped here to discuss two possibilities with you:

1. Adopt latter classes to also use `std::atomic_load` for handling chunk-shared-pointers.
2. Remove the `chunks()` function from Table to limit the use of `std::atomic_load` to Table. Other classes will have to use chunk-ids and `get_chunk()` to get their thread-safe shared pointers to chunks.

So what do you suggest?

julianmenzler:
I talked to @mrks today. We are going to remove the `chunks()` function.

julianmenzler:
For now, I'm done with changes. Please have another look at this PR.

julianmenzler:
> I do not follow.
> Only tables from the StorageManager could have nullptr-Chunks, yet this PR checks Tables that are not from the StorageManager. (` src/benchmark/operators/join_benchmark.cpp`).
> 
> The only Operator that **should** have to deal with Tables from the `StorageManager` is `GetTable`. Wouldn't it be better if `GetTable` always builds a new Table with all "living" Chunks so that all other operators can ignore the possibility of a Chunk being nullptr?
> 
> We have `Validate` handle all the MVCC so that the other operators can ignore it, after all. Garbage collection shouldn't be imposed on all operators, either.

I have to admit, @mrzzzrm has its point here. However, now that the checks are in, does it make sense to remove them?

Btw: I merged the changes of this PR with the MvccDeletePlugin in #1734. Looks good, the FullCI has passed successfully. No more segmentation faults and data races! :)

src/lib/storage/table.cpp
```diff
@@ -151,19 +159,32 @@ uint32_t Table::max_chunk_size() const { return _max_chunk_size; }
...
+    return _chunks[chunk_id];
+  } else {
+    return std::atomic_load(&_chunks[chunk_id]);
+  }
 }
 
 std::shared_ptr<const Chunk> Table::get_chunk(ChunkID chunk_id) const {
   DebugAssert(chunk_id < _chunks.size(), "ChunkID " + std::to_string(chunk_id) + " out of range");
-  return _chunks[chunk_id];
+  if (_type == TableType::References) {
```
mrks: Please add a comment to the declaration of `_chunks` on why (1) we need to access it atomically and (2) why we cannot use `atomic<shared_ptr>`

mrks: Add: "see comment in non-const version"

src/lib/storage/table.cpp
```diff
@@ -151,19 +159,32 @@ uint32_t Table::max_chunk_size() const { return _max_chunk_size; }
 
 std::shared_ptr<Chunk> Table::get_chunk(ChunkID chunk_id) {
   DebugAssert(chunk_id < _chunks.size(), "ChunkID " + std::to_string(chunk_id) + " out of range");
-  return _chunks[chunk_id];
+  if (_type == TableType::References) {
+    // Locking for reference tables is wasted performance since they are not accessed by multiple concurring threads
```
mrks: Not *written* concurrently.

src/lib/tasks/chunk_compression_task.cpp
```diff
@@ -26,8 +26,8 @@ void ChunkCompressionTask::_on_execute() {
 
   for (auto chunk_id : _chunk_ids) {
     Assert(chunk_id < table->chunk_count(), "Chunk with given ID does not exist.");
-
-    auto chunk = table->get_chunk(chunk_id);
+    const auto chunk = table->get_chunk(chunk_id);
+    if (!chunk) continue;
```
mrks: assert

src/lib/storage/table.hpp
```diff
@@ -165,7 +169,11 @@ class Table : private Noncopyable {
   void create_index(const std::vector<ColumnID>& column_ids, const std::string& name = "") {
     SegmentIndexType index_type = get_index_type_of<Index>();
 
-    for (auto& chunk : _chunks) {
+    const auto chunk_count = _chunks.size();
+    for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+      auto chunk = std::atomic_load(&_chunks[chunk_id]);
+      if (!chunk) continue;
```
mrks: assert

src/lib/storage/chunk_encoder.cpp
```diff
@@ -134,10 +134,10 @@ void ChunkEncoder::encode_chunks(const std::shared_ptr<Table>& table, const std:
 
   for (auto chunk_id : chunk_ids) {
     Assert(chunk_id < table->chunk_count(), "Chunk with given ID does not exist.");
+    const auto chunk = table->get_chunk(chunk_id);
+    if (!chunk) continue;
```
mrks: Whole file: assert

src/lib/operators/validate.cpp
```diff
@@ -63,8 +63,10 @@ std::shared_ptr<const Table> Validate::_on_execute(std::shared_ptr<TransactionCo
   const auto our_tid = transaction_context->transaction_id();
   const auto snapshot_commit_id = transaction_context->snapshot_commit_id();
 
-  for (ChunkID chunk_id{0}; chunk_id < in_table->chunk_count(); ++chunk_id) {
+  const auto chunk_count = in_table->chunk_count();
+  for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
     const auto chunk_in = in_table->get_chunk(chunk_id);
+    if (!chunk_in) continue;
```
mrks: assert

src/lib/operators/union_all.cpp
```diff
@@ -28,13 +28,17 @@ std::shared_ptr<const Table> UnionAll::_on_execute() {
   // add positions to output by iterating over both input tables
   for (const auto& input : {input_table_left(), input_table_right()}) {
     // iterating over all chunks of table input
-    for (ChunkID in_chunk_id{0}; in_chunk_id < input->chunk_count(); in_chunk_id++) {
+    const auto chunk_count = input->chunk_count();
+    for (ChunkID in_chunk_id{0}; in_chunk_id < chunk_count; in_chunk_id++) {
+      const auto chunk = input->get_chunk(in_chunk_id);
+      if (!chunk) continue;
```
mrks: assert

src/lib/operators/table_scan.cpp
```diff
@@ -91,11 +91,14 @@ std::shared_ptr<const Table> TableScan::_on_execute() {
   auto jobs = std::vector<std::shared_ptr<AbstractTask>>{};
   jobs.reserve(in_table->chunk_count() - excluded_chunk_set.size());
 
-  for (ChunkID chunk_id{0u}; chunk_id < in_table->chunk_count(); ++chunk_id) {
+  const auto chunk_count = in_table->chunk_count();
+  for (ChunkID chunk_id{0u}; chunk_id < chunk_count; ++chunk_id) {
     if (excluded_chunk_set.count(chunk_id)) continue;
+    const auto chunk_in = in_table->get_chunk(chunk_id);
+    if (!chunk_in) continue;
```
mrks: assert

src/lib/operators/sort.cpp
```diff
@@ -218,8 +218,10 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
 
     auto& null_value_rows = *_null_value_rows;
 
-    for (ChunkID chunk_id{0}; chunk_id < _table_in->chunk_count(); ++chunk_id) {
-      auto chunk = _table_in->get_chunk(chunk_id);
+    const auto chunk_count = _table_in->chunk_count();
+    for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
+      const auto chunk = _table_in->get_chunk(chunk_id);
+      if (!chunk) continue;
```
mrks: assert

src/lib/operators/projection.cpp
```diff
@@ -60,10 +60,12 @@ std::shared_ptr<const Table> Projection::_on_execute() {
    */
   auto output_chunk_segments = std::vector<Segments>(input_table.chunk_count());
 
-  for (auto chunk_id = ChunkID{0}; chunk_id < input_table.chunk_count(); ++chunk_id) {
-    auto output_segments = Segments{expressions.size()};
-
+  const auto chunk_count_input_table = input_table.chunk_count();
+  for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count_input_table; ++chunk_id) {
     const auto input_chunk = input_table.get_chunk(chunk_id);
+    if (!input_chunk) continue;
```
mrks: twice in this file: assert

src/lib/operators/product.cpp
```diff
@@ -29,9 +29,17 @@ std::shared_ptr<const Table> Product::_on_execute() {
   }
 
   auto output = std::make_shared<Table>(column_definitions, TableType::References);
+  auto chunk_count_left_table = input_table_left()->chunk_count();
+  auto chunk_count_right_table = input_table_right()->chunk_count();
+
+  for (ChunkID chunk_id_left = ChunkID{0}; chunk_id_left < chunk_count_left_table; ++chunk_id_left) {
+    const auto chunk_left = input_table_left()->get_chunk(chunk_id_left);
+    if (!chunk_left) continue;
```
mrks: 2x assert

src/lib/operators/print.cpp
```diff
@@ -85,8 +85,10 @@ std::shared_ptr<const Table> Print::_on_execute() {
   _out << "|" << std::endl;
 
   // print each chunk
-  for (ChunkID chunk_id{0}; chunk_id < input_table_left()->chunk_count(); ++chunk_id) {
-    auto chunk = input_table_left()->get_chunk(chunk_id);
+  const auto chunk_count = input_table_left()->chunk_count();
+  for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
+    const auto chunk = input_table_left()->get_chunk(chunk_id);
+    if (!chunk) continue;
```
mrks: Fine in this file!

src/lib/operators/multi_predicate_join/multi_predicate_join_evaluator.cpp
```diff
@@ -63,14 +63,13 @@ std::vector<std::unique_ptr<AbstractSegmentAccessor<T>>> MultiPredicateJoinEvalu
...
-    // get_chunk might return nullptr if we iterate over chunk_count manually
-    // https://github.com/hyrise/hyrise/pull/1482#discussion_r261860186
-    if (chunk) {
-      const auto& segment = table.get_chunk(chunk_id)->get_segment(column_id);
-      accessors[chunk_id] = create_segment_accessor<T>(segment);
-    }
+  const auto chunk_count = table.chunk_count();
+  for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
+    const auto chunk = table.get_chunk(chunk_id);
+    if (!chunk) continue;
```
mrks: assert

src/lib/operators/limit.cpp
```diff
@@ -59,8 +59,11 @@ std::shared_ptr<const Table> Limit::_on_execute() {
   auto output_chunks = std::vector<std::shared_ptr<Chunk>>{};
 
   ChunkID chunk_id{0};
-  for (size_t i = 0; i < num_rows && chunk_id < input_table->chunk_count(); chunk_id++) {
+  const auto chunk_count = input_table->chunk_count();
+  for (size_t i = 0; i < num_rows && chunk_id < chunk_count; chunk_id++) {
     const auto input_chunk = input_table->get_chunk(chunk_id);
+    if (!input_chunk) continue;
```
mrks: assert

src/lib/operators/join_sort_merge/column_materializer.hpp
```diff
@@ -73,7 +73,10 @@ class ColumnMaterializer {
 
     std::vector<std::shared_ptr<AbstractTask>> jobs;
     for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
-      const auto samples_to_write = std::min(samples_per_chunk, input->get_chunk(chunk_id)->size());
+      const auto chunk = input->get_chunk(chunk_id);
+      if (!chunk) continue;
```
mrks: assert

src/lib/operators/join_nested_loop.cpp
```diff
@@ -139,25 +139,36 @@ std::shared_ptr<const Table> JoinNestedLoop::_on_execute() {
   auto left_matches_by_chunk = std::vector<std::vector<bool>>(left_table->chunk_count());
 
   auto right_matches_by_chunk = std::vector<std::vector<bool>>(right_table->chunk_count());
-  for (ChunkID chunk_id_right = ChunkID{0}; chunk_id_right < right_table->chunk_count(); ++chunk_id_right) {
-    right_matches_by_chunk[chunk_id_right].resize(right_table->get_chunk(chunk_id_right)->size());
+  const auto chunk_count_right = right_table->chunk_count();
+  for (ChunkID chunk_id_right = ChunkID{0}; chunk_id_right < chunk_count_right; ++chunk_id_right) {
+    const auto chunk_right = right_table->get_chunk(chunk_id_right);
+    if (!chunk_right) continue;
```
mrks: use assert in this file

src/lib/operators/join_mpsm/column_materializer_numa.hpp
```diff
@@ -84,9 +84,13 @@ class ColumnMaterializerNUMA {
     auto null_rows = std::make_unique<PosList>();
 
     auto jobs = std::vector<std::shared_ptr<AbstractTask>>();
-    for (auto chunk_id = ChunkID{0}; chunk_id < input->chunk_count(); ++chunk_id) {
+    const auto chunk_count = input->chunk_count();
+    for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+      const auto chunk = input->get_chunk(chunk_id);
+      if (!chunk) continue;
```
mrks: assert

src/lib/operators/join_index.cpp
```diff
@@ -83,14 +83,22 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
   const auto track_index_matches = _mode == JoinMode::FullOuter || _mode == JoinMode::Right;
 
   if (track_probe_matches) {
-    for (ChunkID probe_chunk_id = ChunkID{0}; probe_chunk_id < probe_input_table->chunk_count(); ++probe_chunk_id) {
-      _probe_matches[probe_chunk_id].resize(probe_input_table->get_chunk(probe_chunk_id)->size());
+    const auto chunk_count = probe_input_table->chunk_count();
+    for (ChunkID probe_chunk_id = ChunkID{0}; probe_chunk_id < chunk_count; ++probe_chunk_id) {
+      const auto chunk = probe_input_table->get_chunk(probe_chunk_id);
+      if (!chunk) continue;
```
mrks: assert in this file

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -137,7 +137,11 @@ inline std::vector<size_t> determine_chunk_offsets(const std::shared_ptr<const T
   size_t offset = 0;
   for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
     chunk_offsets[chunk_id] = offset;
-    offset += table->get_chunk(chunk_id)->size();
+
+    const auto chunk = table->get_chunk(chunk_id);
+    if (!chunk) continue;
```
mrks: assert in this file

src/lib/operators/jit_operator_wrapper.cpp
```diff
@@ -64,7 +64,11 @@ std::shared_ptr<const Table> JitOperatorWrapper::_on_execute() {
   _source()->before_query(*in_table, _input_parameter_values, context);
   _sink()->before_query(*out_table, context);
 
-  for (ChunkID chunk_id{0}; chunk_id < in_table->chunk_count() && context.limit_rows; ++chunk_id) {
+  const auto chunk_count = in_table->chunk_count();
+  for (ChunkID chunk_id{0}; chunk_id < chunk_count && context.limit_rows; ++chunk_id) {
+    const auto chunk = in_table->get_chunk(chunk_id);
+    if (!chunk) continue;
```
mrks: assert in this file

src/lib/operators/jit_operator/operators/jit_read_tuples.cpp
```diff
@@ -107,13 +107,15 @@ void JitReadTuples::before_specialization(const Table& in_table, std::vector<boo
...
+  // TODO(all) Should we create a get_first_valid_chunk() function
+  //  returning the first non-nullpointer-chunk of the table?
+  const auto chunk = in_table.get_chunk(ChunkID{0});
   // Remove expressions that use a column where the first segment is not dictionary-encoded
   _value_id_expressions.erase(
       std::remove_if(_value_id_expressions.begin(), _value_id_expressions.end(),
                      [&](const JitValueIdExpression& value_id_expression) {
                        const auto column_id = _input_columns[value_id_expression.input_column_index].column_id;
-                       const auto segment = chunk.get_segment(column_id);
+                       const auto segment = chunk->get_segment(column_id);
```
mrks: You added a todo, but this looks like it's currently broken if the first chunk is deleted.

src/lib/operators/index_scan.cpp
```diff
@@ -41,13 +41,19 @@ std::shared_ptr<const Table> IndexScan::_on_execute() {
   auto jobs = std::vector<std::shared_ptr<AbstractTask>>{};
   if (_included_chunk_ids.empty()) {
     jobs.reserve(_in_table->chunk_count());
-    for (auto chunk_id = ChunkID{0u}; chunk_id < _in_table->chunk_count(); ++chunk_id) {
+    const auto chunk_count = _in_table->chunk_count();
+    for (auto chunk_id = ChunkID{0u}; chunk_id < chunk_count; ++chunk_id) {
+      const auto chunk = _in_table->get_chunk(chunk_id);
+      if (!chunk) continue;
```
mrks: assert in this file

src/lib/operators/export_csv.cpp
```diff
@@ -78,8 +78,10 @@ void ExportCsv::_generate_content_file(const std::shared_ptr<const Table>& table
    * This is a lot of iterating, but to convert a column-based table to
    * a row-based representation takes some effort.
    */
-  for (ChunkID chunk_id{0}; chunk_id < table->chunk_count(); ++chunk_id) {
+  const auto chunk_count = table->chunk_count();
+  for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
     const auto chunk = table->get_chunk(chunk_id);
+    if (!chunk) continue;
```
mrks: assert

src/lib/operators/export_binary.cpp
```diff
@@ -167,6 +167,7 @@ void ExportBinary::_write_header(const Table& table, std::ofstream& ofstream) {
 
 void ExportBinary::_write_chunk(const Table& table, std::ofstream& ofstream, const ChunkID& chunk_id) {
   const auto chunk = table.get_chunk(chunk_id);
+  if (!chunk) return;
```
mrks: assert

src/lib/operators/difference.cpp
```diff
@@ -65,8 +68,10 @@ std::shared_ptr<const Table> Difference::_on_execute() {
   output_chunks.reserve(input_table_left()->chunk_count());
 
   // Iterating over all chunks and for each chunk over all segment
-  for (ChunkID chunk_id{0}; chunk_id < input_table_left()->chunk_count(); chunk_id++) {
+  const auto chunk_count_left = input_table_left()->chunk_count();
+  for (ChunkID chunk_id{0}; chunk_id < chunk_count_left; chunk_id++) {
     const auto in_chunk = input_table_left()->get_chunk(chunk_id);
+    if (!in_chunk) continue;
```
mrks: assert

src/lib/operators/difference.cpp
```diff
@@ -37,8 +37,11 @@ std::shared_ptr<const Table> Difference::_on_execute() {
   auto right_input_row_set = std::unordered_set<std::string>(input_table_right()->row_count());
 
   // Iterating over all chunks and for each chunk over all segments
-  for (ChunkID chunk_id{0}; chunk_id < input_table_right()->chunk_count(); chunk_id++) {
-    auto chunk = input_table_right()->get_chunk(chunk_id);
+  const auto chunk_count_right = input_table_right()->chunk_count();
+  for (ChunkID chunk_id{0}; chunk_id < chunk_count_right; chunk_id++) {
+    const auto chunk = input_table_right()->get_chunk(chunk_id);
+    if (!chunk) continue;
```
mrks: assert

src/lib/operators/alias_operator.cpp
```diff
@@ -54,8 +54,10 @@ std::shared_ptr<const Table> AliasOperator::_on_execute() {
    */
   auto output_chunks = std::vector<std::shared_ptr<Chunk>>{input_table.chunk_count()};
 
-  for (auto chunk_id = ChunkID{0}; chunk_id < input_table.chunk_count(); ++chunk_id) {
+  const auto chunk_count = input_table.chunk_count();
+  for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
     const auto input_chunk = input_table.get_chunk(chunk_id);
+    if (!input_chunk) continue;
```
mrks: assert

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -1526,8 +1526,12 @@ std::vector<std::shared_ptr<ExpressionResult<Result>>> ExpressionEvaluator::_pru
     if (table->column_is_nullable(ColumnID{0})) {
       result_nulls.resize(table->row_count());
 
-      for (auto chunk_id = ChunkID{0}; chunk_id < table->chunk_count(); ++chunk_id) {
-        const auto& result_segment = *table->get_chunk(chunk_id)->get_segment(ColumnID{0});
+      const auto chunk_count = table->chunk_count();
+      for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+        const auto chunk = table->get_chunk(chunk_id);
+        if (!chunk) continue;
```
mrks: assert in this file

src/benchmarklib/benchmark_table_encoder.cpp
```diff
@@ -147,8 +147,8 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
         auto my_chunk = next_chunk++;
         if (my_chunk >= table->chunk_count()) return;
 
-        const auto& chunk = table->get_chunk(ChunkID{my_chunk});
-        if (!is_chunk_encoding_spec_satisfied(chunk_encoding_spec, get_chunk_encoding_spec(*chunk))) {
+        const auto chunk = table->get_chunk(ChunkID{my_chunk});
+        if (chunk && !is_chunk_encoding_spec_satisfied(chunk_encoding_spec, get_chunk_encoding_spec(*chunk))) {
```
mrks: assert

src/benchmark/operators/join_benchmark.cpp
```diff
@@ -43,8 +43,10 @@ std::shared_ptr<TableWrapper> generate_table(const size_t number_of_rows) {
   auto table = table_generator->generate_table(std::vector<ColumnDataDistribution>{config}, number_of_rows, chunk_size,
                                                EncodingType::Dictionary);
 
-  for (ChunkID chunk_id{0}; chunk_id < table->chunk_count(); ++chunk_id) {
-    auto chunk = table->get_chunk(chunk_id);
+  const auto chunk_count = table->chunk_count();
+  for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
+    const auto chunk = table->get_chunk(chunk_id);
+    if (!chunk) continue;
```
mrks: assert

mrks:
P.S.: I might be wrong about some of these - happy to discuss them.

src/lib/operators/jit_operator/operators/jit_read_tuples.cpp
julianmenzler: Ok, I removed the todo. Is there anything I should change?

julianmenzler:
I'm done. Please have another look.

src/benchmark/operators/join_benchmark.cpp
```diff
@@ -46,7 +46,7 @@ std::shared_ptr<TableWrapper> generate_table(const size_t number_of_rows) {
   const auto chunk_count = table->chunk_count();
   for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
     const auto chunk = table->get_chunk(chunk_id);
-    if (!chunk) continue;
+    Assert(chunk, "Unexpected nullpointer-chunk.");
```
mrks: Suggestion for the text: `Did not expect deleted chunks here.` Maybe add a comment `// see (wherever we explain the nullptr thing)` if it fits into the same line?

src/lib/storage/table.cpp
```diff
@@ -160,7 +160,7 @@ uint32_t Table::max_chunk_size() const { return _max_chunk_size; }
 std::shared_ptr<Chunk> Table::get_chunk(ChunkID chunk_id) {
   DebugAssert(chunk_id < _chunks.size(), "ChunkID " + std::to_string(chunk_id) + " out of range");
   if (_type == TableType::References) {
-    // Locking for reference tables is wasted performance since they are not accessed by multiple concurring threads
+    // Not written concurrently, since reference tables are not accessed by more than one thread at a time.
```
mrks: Mmmh, I’d rather write “Since reference tables are not modified anymore once they are written.”

src/lib/storage/table.hpp
```diff
@@ -190,7 +190,16 @@ class Table : private Noncopyable {
   const TableType _type;
   const UseMvcc _use_mvcc;
   const uint32_t _max_chunk_size;
+
+  /**
+   * To prevent data races for TableType::Data tables, we must always access vector elements via atomics functions.
```
mrks: `we must access _chunks atomically`?

src/lib/storage/table.hpp
```diff
@@ -190,7 +190,16 @@ class Table : private Noncopyable {
...
+   * This is due to the existance of the MvccDeletePlugin, which might modify shared pointers from a separate thread.
```
mrks: Typo: existence

src/lib/operators/jit_operator/operators/jit_read_tuples.cpp
```diff
@@ -107,8 +107,6 @@ void JitReadTuples::before_specialization(const Table& in_table, std::vector<boo
 
   if (in_table.chunk_count() == 0) return;
 
-  // TODO(all) Should we create a get_first_valid_chunk() function
-  //  returning the first non-nullpointer-chunk of the table?
```
mrks: Apparently, we don’t need that helper function anywhere else. We do, however, need the check. Currently, this would fail if the first chunk is non-existent.

julianmenzler:
Ready to go.

julianmenzler:
Ready.

Issue #1186, opened 2018-10-12
RemoveUselessOperationsRule


mrzzzrm:
Name to be determined, obviously.

There is a number of operations that we can safely remove from LQPs without having to work hard to detect them. Imo it is best if there is one rule to handle them all, otherwise we'll have a lot of very small rules, but I am happy to split this into multiple rules as well.

* [ ] Secondary AliasNodes
Here, e.g., in TPC-H 9
![image](https://user-images.githubusercontent.com/2679018/46893493-5b1c2e80-ce71-11e8-9175-bd20b1712520.png)

* [ ] Reordering Projections in suqueries can always go

* [ ] Reordering Projections can go if there is another Projection/Aggregate/AliasNode above them

* [ ] Sorts 
SortNodes in EXISTS(...)/IN(...) subqueries can go
People may be silly enough to write `ORDER by ..., x, x`, hyrise shouldn't look silly, too, by actually sorting twice.

* [ ] IS NOT NULL and another scan: #1494

mrks:
While working on the TPC-H paper, I did not see any of these useless operators become a problem. They might exist, but we still have bigger fish to fry than removing a 9 μs operation.

I'm tempted to move this to the Trello list of long-term ideas. In cases where these useless operations actually have a performance impact, I'd expect people to come up with the idea of removing them without this issue being open.

Any objections?

PR #1692, opened 2019-05-28
Allow "SELECT a AS a1, a AS a2..."

janetzki:
Close hyrise-mp/hyrise#38
- Allow different aliases for same columns/aggregates
- Introduce a new `struct SQLTranslator::SelectListElement` that stores an `AbstractExpression` and its `SQLIdentifiers`
- Use `SelectListElement` to store aliases of the `SELECT` list elements that cannot be kept in the `SQLIdentifierResolver` 
- Keep `SQLIdentifierResolver::get_expression_identifiers()` because not every expression is part of the `SELECT` list, such as aggregates


mrks:
What happened here? The diff only has tests remaining.

janetzki:
> What happened here? The diff only has tests remaining.

Yes, I removed the code changes to write them from scratch again. The previous implementation made changes in `AbstractExpression`. This one is much cleaner.

mrzzzrm:
Cool, will hopefully be able to give it a full review later today. 

Had a quick look at the tests, though, and they seem pretty thin to me. What about "multi-aliased" columns in subqueries? Can they be correctly used in the enclosing query, e.g.? Can a correlated subquery correctly reference them?

Please stress this new system a bit and try to come up with queries that might be challenging/ambiguous or close-to-ambiguous and test them in the SQLTranslator, but also in the SQLiteTestrunner.

src/lib/sql/sql_translator.hpp
```diff
@@ -51,27 +51,37 @@ class SQLTranslator final {
   static std::shared_ptr<AbstractExpression> translate_hsql_expr(const hsql::Expr& hsql_expr, const UseMvcc use_mvcc);
 
  private:
+  // An expression and its identifiers. This is partly redundant to the SQLIdentifierResolver, but allows similar
+  // expressions with different identifiers (e.g., SELECT COUNT(*) AS cnt1, COUNT(*) AS cnt2 FROM ...).
+  struct NamedExpression {
+    explicit NamedExpression(const std::shared_ptr<AbstractExpression>& expression);
```
mrks: This reads as if I could create a "NamedExpression" that does not have a name?

src/lib/sql/sql_translator.cpp
```diff
@@ -1520,10 +1530,26 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_inverse_predicate(const Abst
   Fail("GCC thinks this is reachable");
 }
 
+std::vector<std::shared_ptr<AbstractExpression>> SQLTranslator::_retrieve_expressions(
```
mrks: The name `_retrieve_expressions` doesn't tell me too much. Maybe `_unwrap_named_expressions`?

src/lib/sql/sql_translator.cpp
```diff
@@ -775,7 +787,7 @@ std::vector<std::shared_ptr<AbstractExpression>> SQLTranslator::_translate_selec
 }
 
 void SQLTranslator::_translate_select_groupby_having(
-    const hsql::SelectStatement& select, const std::vector<std::shared_ptr<AbstractExpression>>& select_list_elements) {
+    const hsql::SelectStatement& select, const std::vector<SQLTranslator::NamedExpression>& select_list_elements) {
```
mrks: Why do we need `SQLTranslator::` here (and above too)?

mrzzzrm:
I guess the continuous income of commits means this is WIP and not ready for review?

resources/test_data/sqlite_testrunner_queries.sql
```diff
@@ -321,19 +329,17 @@ SELECT * FROM id_int_int_int_100 WHERE a * 10 IN (SELECT b FROM mixed)
...
 -- Correlated parameter in WHERE statement
 SELECT * FROM id_int_int_int_100 WHERE a < (SELECT MAX(b) FROM mixed WHERE mixed.b > id_int_int_int_100.b)
+SELECT * FROM id_int_int_int_100 t1 WHERE (SELECT MIN(t2.id + 10) FROM id_int_int_int_100 t2 WHERE t2.id = t1.id) > 20;
 
 -- Subqueries in FROM statement
 SELECT * FROM (SELECT t1.id FROM id_int_int_int_100 t1 JOIN id_int_int_int_100 t2 ON t1.id + 1 = t2.id) AS s1, id_int_int_int_100 t3 WHERE s1.id + 5 = t3.id;
 SELECT * FROM id_int_int_int_100 t1 WHERE id < 9 AND (SELECT MIN(t2.id + 10) FROM (SELECT * FROM id_int_int_int_100 t3 WHERE t3.id > t1.id + 90) AS s1, id_int_int_int_100 t2 WHERE t2.id = t1.id + 90) > 5;
 
--- Correlated parameter (t2.id) in FROM clause of subselect
-SELECT * FROM id_int_int_int_100 t1 WHERE (SELECT MIN(t2.id + 10) FROM id_int_int_int_100 t2 WHERE t2.id = t1.id) > 20;
```
janetzki: I moved this up because the comment does not match the query.

resources/test_data/sqlite_testrunner_queries.sql
```diff
@@ -298,6 +304,7 @@ INSERT INTO mixed_null (b, c, a, d) SELECT b, c, a, d FROM mixed WHERE id < 13;
 CREATE VIEW count_view1 AS SELECT a, COUNT(DISTINCT b) AS cd FROM id_int_int_int_100 GROUP BY a; SELECT * FROM count_view1;
 CREATE VIEW count_view2 AS SELECT a, COUNT(DISTINCT b) AS cd FROM id_int_int_int_100 GROUP BY a; SELECT * FROM count_view2 WHERE a > 10;
 CREATE VIEW count_view3 (foo, bar) AS SELECT a, COUNT(DISTINCT b) AS cd FROM id_int_int_int_100 GROUP BY a; SELECT * FROM count_view3 WHERE foo > 10;
+CREATE VIEW alias_view AS SELECT a AS a1, a AS a2 FROM id_int_int_int_100 WHERE a > 10; SELECT a1, a2 FROM alias_view;
```
janetzki: Unlike SQLite, we even support
```sql
CREATE VIEW alias_view (a3, a4) AS SELECT a AS a1, a AS a2 FROM id_int_int_int_100; SELECT a1, a4 FROM alias_view;
```

janetzki:
As @mrzzzrm recommended, I added complex tests cases for `CREATE VIEW`, `JOIN`, `FROM` Column Aliasing, and correlated subqueries. @mrks @julianmenzler Could you please have a (second) look?

src/lib/sql/sql_translator.hpp
```diff
@@ -51,27 +51,37 @@ class SQLTranslator final {
   static std::shared_ptr<AbstractExpression> translate_hsql_expr(const hsql::Expr& hsql_expr, const UseMvcc use_mvcc);
 
  private:
+  // An expression and its identifiers. This is partly redundant to the SQLIdentifierResolver, but allows similar
+  // expressions with different identifiers (e.g., SELECT COUNT(*) AS cnt1, COUNT(*) AS cnt2 FROM ...).
```
mrks: Can you describe how these "similar" expressions are similar?

src/test/sql/sql_translator_test.cpp
```diff
@@ -367,18 +367,267 @@ TEST_F(SQLTranslatorTest, SelectListAliasUsedInJoin) {
...
+  // clang-format on
+
+  EXPECT_LQP_EQ(actual_lqp_a, expected_lqp_a);
+  EXPECT_LQP_EQ(actual_lqp_b, expected_lqp_b);
+}
+
+TEST_F(SQLTranslatorTest, SelectListAliasesDifferentForSimilarColumns) {
+  const auto actual_lqp = compile_query("SELECT a AS a1, a AS a2, a AS a3, b AS b1, b AS b2, b AS b3 FROM int_float");
+
+  const auto aliases = std::vector<std::string>({"a1", "a2", "a3", "b1", "b2", "b3"});
```
mrks: Not a problem in this case, but generally I'd avoid making test data sorted (unless that is the input you want to test). Otherwise, you might have an algorithm that only works with sorted inputs.

src/test/sql/sql_translator_test.cpp
```diff
@@ -367,18 +367,267 @@ TEST_F(SQLTranslatorTest, SelectListAliasUsedInJoin) {
...
+
+TEST_F(SQLTranslatorTest, SelectListAliasesDifferentForSimilarAggregatesInSubquery) {
+  const auto actual_lqp =
+      compile_query("SELECT * FROM (SELECT COUNT(*) AS cnt1, COUNT(*) AS cnt2, COUNT(*) AS cnt3 FROM int_float) AS R");
+
+  const auto aliases = std::vector<std::string>({"cnt1", "cnt2", "cnt3"});
+  const auto aggregates = expression_vector(count_star_(), count_star_(), count_star_());
+
+  // clang-format off
+  // Redundant AliasNode due to the SQLTranslator architecture. Doesn't look nice, but not really an issue.
```
mrks: #1186

src/lib/sql/sql_translator.cpp
```diff
@@ -502,17 +509,24 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_origin(const hsq
...
-      for (const auto& subquery_expression : lqp->column_expressions()) {
-        const auto identifiers =
-            subquery_translator._sql_identifier_resolver->get_expression_identifiers(subquery_expression);
+      std::vector<std::vector<SQLIdentifier>> identifiers;
+      for (const auto& element : subquery_translator._inflated_select_list_elements) {
+        identifiers.emplace_back(element.identifiers);
+      }
+      Assert(identifiers.size() == lqp->column_expressions().size(),
+             "There have to be as many identifier lists as column expressions");
+      for (auto idx = size_t{0}; idx < lqp->column_expressions().size(); ++idx) {
```
mrks: `idx` is no better than `i`. Index of what?

src/lib/operators/aggregate_hash.hpp
```diff
@@ -69,7 +69,7 @@ using AggregateResultIdMapAllocator = PolymorphicAllocator<std::pair<const Aggre
 template <typename AggregateKey>
 using AggregateResultIdMap =
     ska::bytell_hash_map<AggregateKey, AggregateResultId, std::hash<AggregateKey>, std::equal_to<AggregateKey>,
-                       AggregateResultIdMapAllocator<AggregateKey>>;
+                         AggregateResultIdMapAllocator<AggregateKey>>;
```
janetzki: Is it a bad thing when the formatter changes files that I did not touch?

src/test/statistics/statistics_objects/generic_histogram_test.cpp
```diff
@@ -768,8 +768,7 @@ TEST_F(GenericHistogramTest, SplitAtEmptyBinBounds) {
   const auto expected_minima = std::vector<int32_t>{};
   const auto expected_maxima = std::vector<int32_t>{};
   const auto expected_heights = std::vector<HistogramCountType>{};
-  const auto expected_distinct_counts =
-      std::vector<HistogramCountType>{};
+  const auto expected_distinct_counts = std::vector<HistogramCountType>{};
```
janetzki: same here

src/lib/operators/aggregate_hash.hpp
mrks: Generally no, unless it adds tens of lines to your PR.

src/lib/sql/sql_translator.hpp
mrks: Not addressed yet. If you say something is similar, you should explain how it is not equal.

janetzki:
Thanks for the points, @mrks. I have addressed them.

mrks:
The "similar" thing is still open. Other than that, this should be good to go.

src/lib/sql/sql_translator.hpp
janetzki: Now I (hopefully) understood you. I actually meant "equal" instead of "similar". Good catch.

mrks: Yes. One more thing that I haven't seen before (and have changed myself): An SQL statement is the entire SELECT thing. In this case, we are talking about equal expressions within the select list of the statement.

src/lib/sql/sql_translator.hpp
```diff
@@ -149,6 +159,9 @@ class SQLTranslator final {
 
   std::shared_ptr<AbstractExpression> _inverse_predicate(const AbstractExpression& expression) const;
 
+  std::vector<std::shared_ptr<AbstractExpression>> _unwrap_elements(
+      const std::vector<SelectListElement>& select_list_elements) const;
```
julianmenzler: I would add a comment here answering the following two questions:
1.  What does this function return? 
2. What is this function used for?

janetzki: Good idea, I also do not like the function name very much. ;)
My suggestion is:
```
// Convenience function that extracts and returns the expressions in select_list_elements
```
Now that you point on it, it does not even have to be a method of the `SQLTranslator` but can be a helper function outside of the class. If you want, you can refactor this.

PR #1693, opened 2019-05-28
Allow reference to original column name with alias

janetzki:
Close hyrise-mp/hyrise#36


README.md
```diff
@@ -1,6 +1,5 @@
-[![Build Status](https://hyrise-ci.epic-hpi.de/buildStatus/icon?job=Hyrise/hyrise/master)](https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/activity/)
-[![Coverage Status](https://hyrise-coverage-badge.herokuapp.com/coverage_badge.svg)](https://hyrise-ci.epic-hpi.de/job/Hyrise/job/hyrise/job/master/lastStableBuild/Llvm-cov_5fReport/)
-[![CodeFactor](https://www.codefactor.io/repository/github/hyrise/hyrise/badge)](https://www.codefactor.io/repository/github/hyrise/hyrise)
+[![Build Status](https://hyrise-ci.epic-hpi.de/buildStatus/icon?job=Hyrise/hyrise-mp19/master)](https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/activity/)
```
Bouncner: ??

resources/benchmark/tpcds/tables/reason.csv.json
```diff
@@ -0,0 +1,29 @@
+{
```
Bouncner: Why are these JSON files part of this PR?

resources/test_data/sqlite_testrunner_queries.sql
```diff
@@ -191,6 +191,7 @@ SELECT a, SUM(b) FROM mixed GROUP BY a;
 SELECT a, SUM(b), AVG(c) FROM mixed GROUP BY a;
 SELECT a, b, MAX(c), AVG(b) FROM mixed GROUP BY a, b;
 SELECT a AS whatever, SUM(b) FROM mixed GROUP BY whatever;
+SELECT a AS whatever, SUM(b) FROM mixed GROUP BY a;
```
Bouncner: Maybe something like 'a_alias' instead of 'whatever'?

src/lib/sql/sql_identifier_resolver.cpp
```diff
@@ -48,19 +53,24 @@ std::shared_ptr<AbstractExpression> SQLIdentifierResolver::resolve_identifier_re
   return matching_expressions[0];
 }
 
-const std::optional<SQLIdentifier> SQLIdentifierResolver::get_expression_identifier(
-    const std::shared_ptr<AbstractExpression>& expression) const {
+const std::vector<SQLIdentifier> SQLIdentifierResolver::get_expression_identifiers(
+    const std::shared_ptr<AbstractExpression> &expression) const {
```
Bouncner: The position of the ampersand looks like the linter/formatter haven't been executed.

src/lib/operators/jit_operator/jit_types.cpp
```diff
@@ -38,8 +38,7 @@ namespace opossum {
   }
 
 #define JIT_VARIANT_VECTOR_GET_VECTOR(r, d, type)                                                                 \
-  template <>                                                                                                     \
-  /* NOLINTNEXTLINE(bugprone-macro-parentheses) - clang-tidy does not like & for reference */                     \
+  template <> /* NOLINTNEXTLINE(bugprone-macro-parentheses) - clang-tidy does not like & for reference */         \
```
janetzki: Is this auto-formatting problematic?

src/lib/sql/sql_translator.cpp
```diff
@@ -228,8 +228,8 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_select_statement(cons
   if (select.limit) _translate_limit(*select.limit);
 
   /**
-   * Name, select and arrange the Columns as specified in the SELECT clause
-   */
+     * Name, select and arrange the Columns as specified in the SELECT clause
```
janetzki: Why does the auto-formatter do this?

Bensk1: Not sure, which version did you use?

resources/test_data/sqlite_testrunner_queries.sql
```diff
@@ -191,7 +191,8 @@ SELECT * FROM id_int_int_int_100 AS r WHERE a < (SELECT SUM(min_a) FROM (SELECT
 SELECT a, SUM(b) FROM mixed GROUP BY a;
 SELECT a, SUM(b), AVG(c) FROM mixed GROUP BY a;
 SELECT a, b, MAX(c), AVG(b) FROM mixed GROUP BY a, b;
-SELECT a AS whatever, SUM(b) FROM mixed GROUP BY whatever;
+SELECT a AS a_alias, SUM(b) FROM mixed GROUP BY a_alias;
```
mrzzzrm: Some more queries would be justified here, imo. What about the aliases being used in WHERE? in HAVING? is it legal to use them there?

src/lib/sql/sql_identifier_resolver.cpp
```diff
@@ -48,19 +53,24 @@ std::shared_ptr<AbstractExpression> SQLIdentifierResolver::resolve_identifier_re
   return matching_expressions[0];
 }
 
-const std::optional<SQLIdentifier> SQLIdentifierResolver::get_expression_identifier(
+const std::vector<SQLIdentifier> SQLIdentifierResolver::get_expression_identifiers(
```
mrzzzrm: I know the `const` was there before, but pls remove it. const of return-by-value makes no sense.

src/lib/sql/sql_identifier_resolver.hpp
```diff
@@ -14,7 +14,7 @@ class AbstractExpression;
 
 struct SQLIdentifierContextEntry final {
   std::shared_ptr<AbstractExpression> expression;
-  std::optional<SQLIdentifier> identifier;
+  std::vector<SQLIdentifier> identifiers;
```
mrzzzrm: Can multiple identifiers of an expression have different `table_name`s? if not, wouldn't it be cleaner to make this 
```c++
struct SQLIdentifierContextEntry final {
  std::shared_ptr<AbstractExpression> expression;
  std::optional<std::string> table_name;
  std::vector<std::string> column_names;
}
```

if it is possible that there are multiple table names for one expression, please add SQLiteTestRunner and SQLTranslatorTest - queries for it.

src/lib/sql/sql_identifier_resolver.cpp
```diff
@@ -7,38 +7,43 @@ using namespace std::string_literals;  // NOLINT
...
-  if (entry.identifier) {
-    entry.identifier->column_name = column_name;
-  } else {
-    entry.identifier.emplace(column_name);
-  }
+  entry.identifiers.emplace_back(column_name);
+}
+
+void SQLIdentifierResolver::reset_column_names(const std::shared_ptr<opossum::AbstractExpression>& expression) {
+  auto& entry = _find_or_create_expression_entry(expression);
```
mrzzzrm: so, an Entry is potentially created if it didn't exist before. The function name doesn't reflect this. Do we even want this? If not, an Assert should be added, ensuring that an Entry for `expression` existed before. 
If we want it, the function name should reflect this behaviour and queries triggering this path should be added to the test suite.

src/test/sql/sql_translator_test.cpp
```diff
@@ -1080,6 +1095,23 @@ TEST_F(SQLTranslatorTest, FromColumnAliasingSimple) {
...
+  const auto expected_lqp =
+  AliasNode::make(expression_vector(int_float_a, int_float_b), std::vector<std::string>({"b", "a"}),
+    PredicateNode::make(equals_(int_float_a, int_float_b),
+      stored_table_node_int_float));
+  // clang-format on
+
+  EXPECT_LQP_EQ(actual_lqp_a, expected_lqp);
+  EXPECT_LQP_EQ(actual_lqp_b, expected_lqp);
+}
+
```
mrzzzrm: (random line)
`SELECT a AS b, b AS a FROM t WHERE a > 5` should now yield an InputError, right? If not, add tests for it. If so, add it to `TEST_F(SQLTranslatorTest, CatchInputErrors)`

src/test/sql/sql_translator_test.cpp
```diff
@@ -492,6 +492,21 @@ TEST_F(SQLTranslatorTest, AggregateWithGroupBy) {
   EXPECT_LQP_EQ(actual_lqp, expected_lqp);
 }
 
+TEST_F(SQLTranslatorTest, AliasWithGroupBy) {
+  const auto actual_lqp = compile_query("SELECT a AS x FROM int_float GROUP BY a");
```
mrks: Also test `GROUP BY x`?

src/lib/sql/sql_identifier_resolver.hpp
mrks: Adding on @mrzzzrm comment:

What is the result of `SELECT a AS x FROM t GROUP BY t.x`?

src/lib/sql/sql_identifier_resolver.hpp
```diff
@@ -25,9 +25,10 @@ class SQLIdentifierResolver final {
  public:
   /**
    * @{
-   * Set/Update the column/table name of an expression
+   * Set/Update/Delete the column/table names of an expression
    */
-  void set_column_name(const std::shared_ptr<AbstractExpression>& expression, const std::string& column_name);
+  void add_column_name(const std::shared_ptr<AbstractExpression>& expression, const std::string& column_name);
```
mrks: I'd add a comment explaining why an expression might have multiple column names

src/lib/sql/sql_identifier_resolver.cpp
mrks: Wouldn't it be cheaper to return a reference here?

src/lib/sql/sql_translator.cpp
```diff
@@ -522,8 +525,9 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_origin(const hsq
                 "Must specify a name for exactly each column");
 
     for (auto column_id = ColumnID{0}; column_id < hsql_table_ref.alias->columns->size(); ++column_id) {
-      sql_identifier_resolver->set_column_name(column_expressions[column_id],
-                                               (*hsql_table_ref.alias->columns)[column_id]);
+      const auto& expression = column_expressions[column_id];
+      sql_identifier_resolver->reset_column_names(expression);
```
mrks: Why does it have to be reset?

resources/test_data/sqlite_testrunner_queries.sql
mrks: I'd also add tests that make sure that the aliases properly survive joins etc.

src/lib/sql/sql_translator.cpp
janetzki: 7.0.1-8

src/lib/sql/sql_identifier_resolver.hpp
janetzki: Having multiple identifiers for the same expression with different `table_name`s could work for a query like:
```sql
SELECT R.a, S.a FROM mixed AS R, mixed AS S
```
However, the `SQLTranslator` does not support this behavior since it creates two `StoredTableNode`s in `_translate_stored_table()`. Therefore, `R.a` and `S.a` are identifiers for different expressions. Do you think that changing the SQLTranslator makes sense?
I also added tests for the mentioned query.

mrks: On a general note: As long as we don't break anything, we don't necessarily have to cover edge cases that we don't see in the TPC-DS

resources/test_data/sqlite_testrunner_queries.sql
janetzki: Do we want to support queries like `SELECT a AS x FROM mixed WHERE x > 10`? The `SQLTranslator` does not support this at the moment.

mrks: I believe that this is inconsistent across other DBMSs. My thoughts are:

1. If we need it for TPC-DS and it is reasonably doable, do it.
2. If it's not part of the TPC-DS syntax but can be tacked on without significant extra effort, do it.
3. If adding it adds significant effort, don't.
3a. If possible, add new assertions so that we, should it ever come around, get a more helpful message (maybe even with a comment outlining what would have to be done)

src/lib/sql/sql_identifier_resolver.hpp
mrzzzrm: > However, the SQLTranslator does not support this behavior since it creates two StoredTableNodes in _translate_stored_table().

That is fundamentally intended behavior. R.a and S.a are different columns/expressions in this case. They just happen to come from the same table. Think of `SELECT R.a, S.a FROM mixed AS R, mixed AS S WHERE S.a > 5` - it is obvious that R.a and S.a now contain different data - so they cannot possibly be the same expression/column

resources/test_data/sqlite_testrunner_queries.sql
mrzzzrm: I imagine this requires effort and agree with Markus' other points

src/lib/sql/sql_identifier_resolver.hpp
janetzki: Thanks for the clarification. Handling these columns identically would not help us implementing the TPC-DS. Thus, I won't change the `SQLTranslator`.

mrzzzrm:
@janetzki once you feel you have addressed the review, please ping me and I'll make another pass.

src/lib/sql/sql_identifier_resolver.hpp
janetzki: This crashes, as well as SQLite. I added a test for this.

src/lib/sql/sql_translator.cpp
janetzki: This is especially for queries like `SELECT * FROM int_float AS t (b, a) WHERE b = t.a`, in which the original column names should not be accessible anymore.

resources/test_data/sqlite_testrunner_queries.sql
janetzki: I was able to add alias support in the WHERE clause (https://github.com/hyrise-mp/hyrise/commit/6d9c7e8b7a642ba97a2cd845914ea773911f2bf5). However, it has some consequences:

janetzki: >     2. If it's not part of the TPC-DS syntax but can be tacked on without significant extra effort, do it.
Yes, it's not required for the TPC-DS.

mrzzzrm: I don't see a clear win here and it breaks even more with this comment, though I wouldn't see this as important unless the referenced blog post would actually reference the standard.

```
  // SQL Orders of Operations: http://www.bennadel.com/blog/70-sql-query-order-of-operations.htm
  // 1. FROM clause (incl. JOINs and sub-SELECTs that are part of this)
  // 2. WHERE clause
  // 3. GROUP BY clause
  // 4. HAVING clause
  // 5. SELECT clause (incl. DISTINCT)
  // 6. UNION clause
  // 7. ORDER BY clause
  // 8. LIMIT clause
```

We gain one thing and loose another - I'm fine with it. 

> This does not seem to be a problem from my perspective. I only wondered because there is an explicit test for this

Don't worry, the ParamaterIDs are arbitrary, they just have to be unique. `parameter_ids_of_value_placeholders` is what gives them order. Explicit values were tested to keep things simple.

Unless @mrks says something else, I'd say we can keep this. Please revise the naming of `_translate_select_list_groupby_having` as its behavior has  changed. Adjust that comment I mentioned above (probably removing the URL)

mrks: Sounds good.

> [...] not by Hyrise anymore because a becomes ambiguous.

So be it. Whoever writes those queries deserves to be punished. ;)

src/test/sql/sql_translator_test.cpp
janetzki: Yes, I added a test for the case that this query yields an 'InvalidInputException'.

janetzki:
@mrzzzrm I addressed all your comments. Please have another look.

src/lib/sql/sql_identifier_resolver.cpp
```diff
@@ -48,19 +53,25 @@ std::shared_ptr<AbstractExpression> SQLIdentifierResolver::resolve_identifier_re
...
 
-const std::optional<SQLIdentifier> SQLIdentifierResolver::get_expression_identifier(
+const std::vector<SQLIdentifier>& SQLIdentifierResolver::get_expression_identifiers(
     const std::shared_ptr<AbstractExpression>& expression) const {
   auto entry_iter = std::find_if(_entries.begin(), _entries.end(),
                                  [&](const auto& entry) { return *entry.expression == *expression; });
-  if (entry_iter == _entries.end()) return std::nullopt;
-  return entry_iter->identifier;
+  const auto empty_vector = std::make_shared<std::vector<SQLIdentifier>>();
+  if (entry_iter == _entries.end()) return *empty_vector;
```
mrzzzrm: This returns a reference to a local variable, which is subsequently destroyed. Did none of our sanitizers catch this?

mrzzzrm: I think I'd be fine with return-by-value.

src/lib/sql/sql_identifier_resolver.hpp
mrzzzrm: why is this resolved? I believe the answer to my original question ("Can multiple identifiers of an expression have different table_names?") was "No", so wouldn't the layout of `SQLIdentifierContextEntry` that I proposed be cleaner?

src/lib/sql/sql_identifier_resolver.hpp
```diff
@@ -25,9 +25,11 @@ class SQLIdentifierResolver final {
  public:
   /**
    * @{
-   * Set/Update the column/table name of an expression
+   * Set/Update/Delete the column/table names of an expression. There can be multiple column names because a new alias
```
mrzzzrm: "There can be multiple column names *referring to a single expression*"?

src/lib/sql/sql_translator.cpp
mrzzzrm: Please add this as a comment, I think that example helps.

src/lib/sql/sql_translator.cpp
```diff
@@ -739,7 +748,31 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_cross_product(const st
   return result_table_source_state;
 }
 
-void SQLTranslator::_translate_select_list_groupby_having(const hsql::SelectStatement& select) {
+std::vector<std::shared_ptr<AbstractExpression>> SQLTranslator::_translate_select_list(
+    const hsql::SelectStatement& select) {
```
mrzzzrm: Could take whatever the type of `select.selectList` is as a parameter?

src/test/sql/sql_translator_test.cpp
```diff
@@ -492,6 +510,81 @@ TEST_F(SQLTranslatorTest, AggregateWithGroupBy) {
   EXPECT_LQP_EQ(actual_lqp, expected_lqp);
 }
 
+TEST_F(SQLTranslatorTest, AliasWithGroupBy) {
```
mrzzzrm: I'd call these tests "SelectListAliasUsedIn...."

src/test/sql/sql_translator_test.cpp
```diff
@@ -1106,6 +1216,20 @@ TEST_F(SQLTranslatorTest, FromColumnAliasingTablesSwitchNames) {
   EXPECT_LQP_EQ(actual_lqp_b, expected_lqp);
 }
 
+TEST_F(SQLTranslatorTest, SameColumnForDifferentTableNames) {
+  const auto actual_lqp = compile_query("SELECT R.a, S.a FROM int_float AS R, int_float AS S");
+
+  // clang-format off
+  const auto expected_lqp =
+  ProjectionNode::make(expression_vector(int_float_a, int_float_a),
+      JoinNode::make(JoinMode::Cross,
```
mrzzzrm: indentation is off

mrzzzrm:
@mrks Once these points are addressed, I am fairly happy with this. Won't be available until Tuesday, feel free to merge without my approval.

src/lib/sql/sql_identifier_resolver.cpp
janetzki: Neither `lint.sh` not `format.sh` did something about this. Shouldn't the object be kept until there is no more reference to it because it comes from a shared pointer?

src/lib/sql/sql_translator.cpp
janetzki: I added a shorter version of this comment because there already is an explanation a few lines earlier.

src/lib/sql/sql_identifier_resolver.hpp
janetzki: Good catch. It should make more sense now.

src/lib/sql/sql_identifier_resolver.cpp
mrks: `*empty_vector` does not increase the counter, so there is exactly one `shared_ptr` pointing to `empty_vector`. When that `shared_ptr` goes out of scope, `*empty_vector` is broken.

My approach for things like this is to have a `static` empty object and return a const reference to that one.

src/lib/sql/sql_identifier_resolver.hpp
mrks: I'd expect it to be a suggestion

src/lib/sql/sql_identifier_resolver.cpp
mrzzzrm: > Neither lint.sh not format.sh did something about this

I was referring to valgrind, ASAN and UBSAN, but since this PR doesn't have FullCI enabled yet, those haven't actually run yet.

@mrks' solution is fine (though I always wondered whether there might be some performance hit when using it in multi-threaded scenarios)

mrks: You don't get any software level parallelization guarantees. If you want to access it from multiple threads, you will have to synchronize the access yourself. As the object is not modified, there should also be no cache line synchronization necessary. A `const` wouldn't hurt though.

mrzzzrm: Oh, but you do, right? 

https://stackoverflow.com/a/17713799

"...static storage duration for a function local means that only one instance of that local exists in the program" 

"The thread safety bits make it get a bit more complicated..."

But looking at the assembly in the SO answer, the primary guard variable is not synchronized as far as I can tell, so the overhead *should* be minimal

src/lib/sql/sql_identifier_resolver.hpp
mrks: @mrzzzrm I checked the other open points, which are addressed. With this one, I am not sure enough to discard your change request. Could you have a look at https://github.com/hyrise/hyrise/pull/1693/commits/67fb153dd8cd8cffdd105f062c2988eb0dcba1df

src/lib/sql/sql_identifier_resolver.cpp
janetzki: I chose to do it return-by-value because the `SQLIdentifier`s are not part of the resolver anymore, but they are created ad hoc.

src/lib/sql/sql_identifier_resolver.hpp
janetzki: Yes, this commit addresses your suggestion @mrzzzrm. (Sorry for resolving the discussion too soon, though.) As a consequence, we can now run queries, such as `SELECT a AS x FROM int_float GROUP BY int_float.x`, which does not look right in my eyes but is unambiguous.

janetzki:
Thanks a lot for your critical reviews @mrzzzrm! I addressed all your comments.

src/lib/sql/sql_identifier_resolver.hpp
mrzzzrm: Ok, I believe that's acceptable @mrks ?

mrks: Doesn’t feel completely right, but I don’t expect it to hurt anywhere.

mrks:
Could someone merge this once it’s green?

PR #1696, opened 2019-05-29
Add nullable column support to TableBuilder

pascalhpi:
- move TableBuilder to its own file so it can also be used for other tpc benchmarks
- add nullable column support
- add some tests


src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,240 @@
...
+                      auto is_nullable = is_optional<decltype(type)>();
+
+                      definitions.emplace_back(name, data_type, is_nullable);
+
+                      return definitions;
+                    });
+    _table = std::make_shared<Table>(column_definitions, TableType::Data, chunk_size, use_mvcc);
+
+    // Reserve some space in the vectors
+    hana::for_each(_data_vectors, [&](auto&& vector) { vector.reserve(_estimated_rows_per_chunk); });
```
mrzzzrm: I think everywhere else (e.g. ValueSegment) we have established to call these vector `values` and `null_values`

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  // Types == DataTypes, we deduce again to get universal references, so we can call with lvalues
+  template <typename... Types>
+  void append_row(Types&&... values) {
+    auto values_tuple = hana::make_tuple(std::forward<decltype(values)>(values)...);
+    BOOST_HANA_CONSTANT_ASSERT(hana::size(hana::tuple<DataTypes...>()) == hana::size(values_tuple));
+
+    // Create tuples ([&data_vector0, &is_null_vector0, value0], [&data_vector1, &is_null_vector1, value1], ...)
+    auto data_vectors_and_is_null_vectors_and_values = hana::zip_with(
```
mrzzzrm: why do you zip here and then do "hana::for_each" below in a different step? Why don't you do emplace_backs()s right here already?

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,240 @@
...
+
+  hana::tuple<std::vector<get_value_type<DataTypes>>...> _data_vectors;
+  hana::tuple<optional_constexpr<std::vector<bool>, (is_optional<DataTypes>())>...> _is_null_vectors;
+
+  size_t _current_chunk_row_count() const { return _data_vectors[hana::llong_c<0>].size(); }
+
+  void _emit_chunk() {
+    Segments segments;
+
+    auto _data_vectors_and_is_null_vectors = hana::zip_with(
```
mrzzzrm: same here, why are zip_with and for_each two distinct steps?

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,240 @@
...
+    }
+  }
+
+ private:
+  std::shared_ptr<Table> _table;
+  UseMvcc _use_mvcc;
+  size_t _estimated_rows_per_chunk;
+
+  hana::tuple<std::vector<get_value_type<DataTypes>>...> _data_vectors;
+  hana::tuple<optional_constexpr<std::vector<bool>, (is_optional<DataTypes>())>...> _is_null_vectors;
```
mrzzzrm: If this get's us rid of some template programming I'd be cool with leaving the optional_constexpr out here and just have an empty `std::vector<bool>` here.

src/test/benchmarklib/table_builder.cpp
```diff
@@ -0,0 +1,55 @@
+#include "gtest/gtest.h"
+
+#include "all_type_variant.hpp"
+#include "table_builder.hpp"
+
+namespace opossum {
+
+template <typename T>
+std::optional<T> get_optional_from_table(const Table& table, uint16_t x, int y) {
```
mrzzzrm: unless you have a very good (performance) reason to do so, please do not use uint16_t, nor "int", it makes things inconsistent and just sets us up for overflow-trouble.
`size_t` is the type to be used for indices

src/test/benchmarklib/table_builder.cpp
```diff
@@ -0,0 +1,55 @@
...
+  auto table = table_builder.finish_table();
+
+  EXPECT_EQ(table->row_count(), 2);
+  EXPECT_EQ(table->get_value<int32_t>(ColumnID{0}, 0), 42);
+  EXPECT_EQ(table->get_value<pmr_string>(ColumnID{2}, 0), pmr_string{"42"});
+
+  auto optional_value = get_optional_from_table<float>(*table, 1, 0);
+  EXPECT_TRUE(optional_value.has_value());
+  EXPECT_EQ(optional_value.value(), 42.0f);
+  EXPECT_FALSE(get_optional_from_table<float>(*table, 1, 1).has_value());
```
mrzzzrm: with EXPECT_TABLE_EQ_UNORDERED you shouldn't be needing this, but there is `Table::get_value()` for what you're doing here. 
`variant_is_null` for null value checks on AllTypeVariant.

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,240 @@
...
+#include "types.hpp"
+
+namespace hana = boost::hana;
+
+// helper structs and functions for compile time, the actual code is in namespace opossum
+namespace {
+// similar to std::optional but has_value is known at compile time, so "if constexpr" can be used
+// hana::optional does not allow moving and reinitializing its value (or I just did not find out how)
+template <typename T, bool _has_value, typename Enable = void>
+class optional_constexpr {
```
mrzzzrm: I'm not sure we're consistent with this (we're not, see, `pmr_string`), but generally, class names should be camel case.

src/benchmarklib/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ set(
     file_based_table_generator.hpp
     file_based_benchmark_item_runner.cpp
     file_based_benchmark_item_runner.hpp
+    table_builder.hpp
```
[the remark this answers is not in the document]
pascalhpi: done

src/test/benchmarklib/table_builder.cpp
```diff
@@ -0,0 +1,55 @@
...
+const auto types = boost::hana::tuple<int32_t, std::optional<float>, pmr_string>();
+const auto names = boost::hana::make_tuple("a", "b", "c");
+
+TEST(TableBuilderTest, CreateColumnsWithCorrectNamesAndTypesAndNullables) {
+  auto table_builder = TableBuilder(4, types, names, UseMvcc::No);
+  auto table = table_builder.finish_table();
+
+  EXPECT_EQ(table->column_count(), 3);
+
+  auto expected_names = std::vector<std::string>{"a", "b", "c"};
```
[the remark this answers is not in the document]
pascalhpi: Interesting, so the issue here is that the EXPECT_EQ macro does not handle the commas inside of curly braces correctly, but C++ language wise `std::vector<std::string>{"a", "b", "c"}` and `std::vector<std::string>({"a", "b", "c"})` are absolutely identical, right?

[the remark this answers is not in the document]
mrzzzrm: I'd say so, too. This is no fault of EXPECT_EQ in particular, the C preprocessor just can't match curly braces afaik

src/test/benchmarklib/table_builder.cpp
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+
+TEST(TableBuilderTest, AppendsRows) {
+  auto table_builder = TableBuilder(4, types, names, UseMvcc::No);
+  table_builder.append_row(42, 42.0f, "42");
+  table_builder.append_row(42, std::optional<float>{}, "42");
+  auto table = table_builder.finish_table();
+
+  EXPECT_EQ(table->row_count(), 2);
```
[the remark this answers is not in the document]
pascalhpi: I realized EXPECT_TABLE_EQ_UNORDERED does not check nullable flags - I added an issue for that (#1698). It definitely simplifies things though, thanks for the hint!

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,240 @@
...
+ * data vector for each column and appends values to them in append_row(). For nullable columns an additional
+ * is_null_vector is kept. Automatically creates chunks in accordance with the specified chunk size.
+ */
+template <typename... DataTypes>
+class TableBuilder {
+ public:
+  // names is a list of strings defining column names, types is a list of equal length defining respective column types
+  // types may contain std::optional<?>, which will result in a nullable column, otherwise columns are not nullable
+  template <typename... Names>
+  TableBuilder(size_t chunk_size, const hana::tuple<DataTypes...>& types, const hana::tuple<Names...>& names,
```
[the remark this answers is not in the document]
pascalhpi: Unfortunately no, DataTypes are used to declare members _data_vectors and _is_null_vectors. We need it to be a template parameter pack to get one std::vector for each type. I simplified Names though.

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,240 @@
...
+               opossum::UseMvcc use_mvcc, size_t estimated_rows = 0)
+      : _use_mvcc(use_mvcc), _estimated_rows_per_chunk(std::min(estimated_rows, chunk_size)) {
+    BOOST_HANA_CONSTANT_ASSERT(hana::size(names) == hana::size(types));
```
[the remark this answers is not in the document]
pascalhpi: Not really. It crashes at compile time, but with a cryptic error.

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,240 @@
...
+
+    // Iterate over the column types/names and create the columns.
+    auto column_definitions = TableColumnDefinitions{};
+    hana::fold_left(hana::zip(names, types), column_definitions,
+                    [](auto& definitions, const auto& name_and_type) -> decltype(definitions) {
```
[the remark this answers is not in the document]
pascalhpi: Yes.

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,240 @@
...
+  // types may contain std::optional<?>, which will result in a nullable column, otherwise columns are not nullable
+  template <typename... Names>
+  TableBuilder(size_t chunk_size, const hana::tuple<DataTypes...>& types, const hana::tuple<Names...>& names,
+               opossum::UseMvcc use_mvcc, size_t estimated_rows = 0)
+      : _use_mvcc(use_mvcc), _estimated_rows_per_chunk(std::min(estimated_rows, chunk_size)) {
+    BOOST_HANA_CONSTANT_ASSERT(hana::size(names) == hana::size(types));
+
+    // Iterate over the column types/names and create the columns.
+    auto column_definitions = TableColumnDefinitions{};
+    hana::fold_left(hana::zip(names, types), column_definitions,
```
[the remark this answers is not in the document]
pascalhpi: That worked after also adding a ` -> std::decay_t<decltype(definitions)> ` to the lambda's return type, but how does that help with the awkward duality? Do you prefer calling the lmbda's parameter `column_definitions` too? That would (kind of) shadow `column_definitions` from outside of the lambda and probably confuse readers in my opinion.

src/benchmarklib/table_builder.hpp
pascalhpi: As a reader I would probably not expect anything to happen in a zip_with lambda apart from zipping, also I'm pretty sure I just stuck to that pattern because it was already there when I started working on the table_builder. You're right though, it's easier to do it in one step, I will change it.

src/benchmarklib/table_builder.hpp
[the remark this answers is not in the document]
mrzzzrm: compiler/tidy probably won't even allow shadowing. sad that we cannot avoid it. I'd call it "column_definitions" and "column_definitions2" then, better not to hide these awkward things. But that's up to you.

src/benchmarklib/table_builder.hpp
mrzzzrm: ah, maybe zip_with does not guarantee order of execution, where fold_left does? in that case we should probably use two steps :(

src/benchmarklib/table_builder.hpp
pascalhpi: We would need an std::optional<std::vector>> to check whether a column is nullable. Basically all if constexprs would be normal ifs, which is especially bad for the one in append_row, since that's our "hot function". I will benchmark and see what the impact would be.

src/benchmarklib/table_builder.hpp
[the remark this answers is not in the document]
pascalhpi: Yep, because of [-Werror,-Wshadow-uncaptured-local] the commit can't be built right now by jenkins (on the sidon server that was not an issue). I didn't even consider that shadowing, since an uncaptured local variable is not even in the lambda's scope. Oh well, I will go with "column_definitions2" for the lambda's parameter.

src/benchmarklib/table_builder.hpp
pascalhpi: I found a comment in the ["old" table builder](https://github.com/hyrise/hyrise/blob/master/src/benchmarklib/tpch/tpch_table_generator.cpp): `Create a tuple ((column_name0, column_type0), (column_name1, column_type1), ...) so we can iterate over the columns. fold_left as below does this in order, I think boost::hana::zip_with() doesn't, which is why I'm doing two steps here.` However, I can't find that in the boost hana documentation. Do you have a link where I can read up on boost hana's order of execution?

src/benchmarklib/table_builder.hpp
pascalhpi: I can't explain why, but the test fails when I move the for_each part into zip_with. Maybe it's the order of execution?

src/benchmarklib/table_builder.hpp
pascalhpi: In this case the order of execution does not matter, so I would still do it in one step and add a comment about order of execution.

src/benchmarklib/table_builder.hpp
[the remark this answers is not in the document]
pascalhpi: I found another way: capturing column_definitions into the lambda and just using for_each. I guess we were looking for a complex solution, but there was a very simple one...

src/benchmarklib/table_builder.hpp
mrzzzrm: I think "Side effects and purity" [here](https://www.boost.org/doc/libs/1_65_0/libs/hana/doc/html/index.html) talks about this. "Except where specified otherwise, all functions used with Hana (i.e. used in higher order algorithms) should be pure. In particular, functions passed to higher order algorithms are not guaranteed to be called any specific number of times"

So I have to back away from my initial proposal and opt for staying with zip_with followed by fold_left

src/benchmarklib/table_builder.hpp
mrzzzrm: https://github.com/hyrise/hyrise/pull/1696#discussion_r290423335

src/benchmarklib/table_builder.hpp
mrzzzrm: ah, no need to bother, was just a thought.

mrzzzrm:
@pascalhpi ping me once you feel you've addressed my review an I'll give it another go.

src/benchmarklib/table_builder.hpp
pascalhpi: Alright, thanks a lot! It also says "However, some exceptional algorithms like for_each do expect impure functions, and they guarantee an order of evaluation." I prefer for_each over fold_left, to avoid that awkward duality we discussed before.

pascalhpi:
@mrzzzrm ping

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+  using value_type = typename T::value_type;
+};
+
+// get the value_type (lets call it V) of std::optional<V> and V using the same syntax - needed for templating
+template <typename T>
+using get_value_type = typename GetValueType<T>::value_type;
+
+// check if std::optional<V> or V is null using the same syntax - needed for templating
+template <typename T>
+constexpr bool is_null(const T& maybe_optional) {
```
mrzzzrm: minor: optional_or_value

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+    return maybe_optional.value();
+  } else {
+    return maybe_optional;
+  }
+}
+
+}  // namespace table_builder
+
+/**
+ * Helper to build a table with a static column layout, specified by constructor args types and names. Keeps a
```
mrzzzrm: "...specified by template parameters"?

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+ * null_values vector is kept. Automatically creates chunks in accordance with the specified chunk size.
+ */
+template <typename... DataTypes>
+class TableBuilder {
+ public:
+  // names is a boost::hana::tuple of strings defining column names, types is a list of equal length defining respective
+  // column types types may contain std::optional<?>, which will result in a nullable column, otherwise columns are not
+  // nullable
+  template <typename Names>
+  TableBuilder(size_t chunk_size, const boost::hana::tuple<DataTypes...>& types, const Names& names,
```
mrzzzrm: minor: const size_t - we tend to const qualify variables as much as possible.

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+    BOOST_HANA_CONSTANT_ASSERT(boost::hana::size(names) == boost::hana::size(types));
+
+    // Iterate over the column types/names and create the columns.
+    auto column_definitions = TableColumnDefinitions{};
+    boost::hana::for_each(boost::hana::zip(names, types), [&](const auto& name_and_type) {
+      auto name = name_and_type[boost::hana::llong_c<0>];
+      auto type = name_and_type[boost::hana::llong_c<1>];
+
+      auto data_type = data_type_from_type<table_builder::get_value_type<decltype(type)>>();
+      auto is_nullable = table_builder::is_optional_v<decltype(type)>;
```
mrzzzrm: if you need is_optional_v just here, maybe put its decl here to keep things compact?

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+
+  std::shared_ptr<Table> finish_table() {
+    if (_current_chunk_row_count() > 0) {
+      _emit_chunk();
+    }
+
+    return _table;
+  }
+
+  // Types == DataTypes, we deduce again to get universal references, so we can call with lvalues
```
mrzzzrm: "so we can call with **rvalues**", no? Don't remember exactly what the problem was, though.

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+
+    // Create tuples ([&values0, &null_values0, value0], [&values1, &null_values1, value1], ...)
+    auto value_vectors_and_null_value_vectors_and_values = boost::hana::zip_with(
+        [](auto& values, auto& null_values, auto&& value) {
+          return boost::hana::make_tuple(std::reference_wrapper(values), std::reference_wrapper(null_values),
+                                         std::forward<decltype(value)>(value));
+        },
+        _value_vectors, _null_value_vectors, values_tuple);
+
+    // The reason why we can't do the for_each stuff in zip_with is the following:
```
mrzzzrm: feel free to remove this comment, sprinkling hana docs about the libraries basic concepts into our code won't help much, imo

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+    // function f or even the number of executions of f. Therefore it should only be used with pure functions (no side
+    // effects). There are exceptions to that, for_each gives these guarantees and expects impure functions.
+
+    // Add the values to their respective value vector
+    boost::hana::for_each(value_vectors_and_null_value_vectors_and_values, [](auto& values_and_null_values_and_value) {
+      auto& values = values_and_null_values_and_value[boost::hana::llong_c<0>].get();
+      auto& null_values = values_and_null_values_and_value[boost::hana::llong_c<1>].get();
+      auto& maybe_optional_value = values_and_null_values_and_value[boost::hana::llong_c<2>];
+
+      constexpr bool column_is_nullable = std::decay_t<decltype(null_values)>::has_value;
```
mrzzzrm: minor: auto

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+      auto value_is_null = table_builder::is_null(maybe_optional_value);
+
+      DebugAssert(column_is_nullable || !value_is_null, "cannot insert null value into not-null-column");
+
+      if (value_is_null) {
+        values.emplace_back();
+      } else {
+        // on failure: make sure template pack Types is the same that was passed to table_builder constructor
```
mrzzzrm: Please revise comment, I don't see a failure nor something that "makes sure" anything here.

src/benchmarklib/table_builder.hpp
pascalhpi: I'd argue that the template parameters are not explicitly specified but rather deduced from the ctor args.

src/benchmarklib/table_builder.hpp
pascalhpi: It is needed elsewhere too.

src/benchmarklib/table_builder.hpp
pascalhpi: No, lvalues is correct. We could not call append_row like this:
`auto customer = call_dbgen_mk<customer_t>(row_idx + 1, mk_cust, TpchTable::Customer);
    customer_builder.append_row(customer.custkey, customer.name, customer.address, customer.nation_code, customer.phone,
                                convert_money(customer.acctbal), customer.mktsegment, customer.comment);`
for tpcds, all in32_t values ahd to be wrapped like `x.append_row(int32_t{customer.age}, ...);` (for tpch it was working for whatever reason).

src/benchmarklib/table_builder.hpp
pascalhpi: There was a comment about that in the original table_builder and it reduces confusion about why we first zip_with and then for_each, so I think it should stay.

src/benchmarklib/tpch/tpch_table_generator.cpp
```diff
@@ -260,17 +157,21 @@ std::unordered_map<std::string, BenchmarkTableInfo> TpchTableGenerator::generate
...
+                             order_count};
   TableBuilder lineitem_builder{_benchmark_config->chunk_size, lineitem_column_types, lineitem_column_names,
-                                order_count * 4};
-  TableBuilder part_builder{_benchmark_config->chunk_size, part_column_types, part_column_names, part_count};
+                                UseMvcc::Yes, order_count * 4};
+  TableBuilder part_builder{_benchmark_config->chunk_size, part_column_types, part_column_names, UseMvcc::Yes,
+                            part_count};
   TableBuilder partsupp_builder{_benchmark_config->chunk_size, partsupp_column_types, partsupp_column_names,
-                                part_count * 4};
+                                UseMvcc::Yes, part_count * 4};
```
mrks: Why are we putting UseMvcc back in?

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+#include "boost/hana/zip_with.hpp"
+
+#include "resolve_type.hpp"
+#include "types.hpp"
+
+namespace opossum {
+
+// helper structs and functions for compile time, the actual code is in namespace opossum
+namespace table_builder {
+// similar to std::optional but has_value is known at compile time, so "if constexpr" can be used
```
mrks: Did you write this code or is it based on a StackOverflow post (or alike)? If so, please add a comment.

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+/**
+ * Helper to build a table with a static column layout, specified by constructor arguments types and names. Keeps a
+ * value vector for each column and appends values to them in append_row(). For nullable columns an additional
+ * null_values vector is kept. Automatically creates chunks in accordance with the specified chunk size.
+ */
+template <typename... DataTypes>
+class TableBuilder {
+ public:
+  // names is a boost::hana::tuple of strings defining column names, types is a list of equal length defining respective
+  // column types types may contain std::optional<?>, which will result in a nullable column, otherwise columns are not
```
mrks: This sentence looks broken

src/benchmarklib/table_builder.hpp
```diff
@@ -0,0 +1,241 @@
...
+        // this will fail if append_row is not called with values of the types passed to the TableBuilder constructor
+        values.emplace_back(std::move(table_builder::get_value(optional_or_value_value)));
+      }
+
+      if constexpr (column_is_nullable) {
+        null_values.value().emplace_back(value_is_null);
+      }
+    });
+
+    // The reason why we can't do the for_each stuff in zip_with is the following:
```
mrks: Please to avoid an overly casual tone in comments. "The ... stuff" is borderline, but does not necessarily have to be changed.

src/benchmarklib/table_builder.hpp
pascalhpi: That's my code.

src/benchmarklib/tpch/tpch_table_generator.cpp
pascalhpi: Sorry, that was not intended. I'll remove it.

pascalhpi: Actually mvcc is always on now, StorageManager::add_table requires mvcc data. I removed the parameter.

src/benchmarklib/table_builder.hpp
mrks: sweet

mrks: As in "cool, I like it"

PR #1697, opened 2019-05-31
IndexScan improvements

Bensk1:
While working on the IndexScan, a few minor improvements came up:

1. IndexScans only work on data tables and scan the table chunk by chunk. Therefore, we can definitely guarantee that the output PosLists only reference a single chunk which speeds up following scans.
1. The output PosList list was previously `reserved` and written by `std::transform` using `std::back_inserter`. Replacing this with a simple for loop and `resize()` speeds up the IndexScan by a factor of two for large tables. For example, for the discount predicate of TPC-H query 6, the IndexScan was 5-6 times as fast as a normal TableScan and is now roughly 10x faster. 
1. ~~Also, the LQPTranslator always created an additional UnionPositions and TableScan for non-indexed chunks of a table. This is now only done if not all chunks have an index.~~

I don't know whether we want the `PosListGuarenteesSingleChunkReference`. It does test the behaviour for all index types which is not really necessary.

### DIFF
--- a/src/lib/logical_query_plan/lqp_translator.cpp
+++ b/src/lib/logical_query_plan/lqp_translator.cpp
@@ -212,10 +212,12 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_predicate_node_to_in
   // TableScan(s).
   auto index_scan = std::make_shared<IndexScan>(input_operator, SegmentIndexType::GroupKey, column_ids,
                                                 predicate->predicate_condition, right_values, right_values2);
+  index_scan->set_included_chunk_ids(indexed_chunks);
+
+  // If all chunks of a table are indexed, we don't need anything else except the IndexScan
+  if (indexed_chunks.size() == static_cast<size_t>(table->chunk_count())) return index_scan;
 
   const auto table_scan = _translate_predicate_node_to_table_scan(node, input_operator);
-
-  index_scan->set_included_chunk_ids(indexed_chunks);
   table_scan->set_excluded_chunk_ids(indexed_chunks);
 
   return std::make_shared<UnionPositions>(index_scan, table_scan);
--- a/src/lib/logical_query_plan/lqp_translator.cpp
+++ b/src/lib/logical_query_plan/lqp_translator.cpp
@@ -212,12 +212,10 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_predicate_node_to_in
   // TableScan(s).
   auto index_scan = std::make_shared<IndexScan>(input_operator, SegmentIndexType::GroupKey, column_ids,
                                                 predicate->predicate_condition, right_values, right_values2);
-  index_scan->set_included_chunk_ids(indexed_chunks);
-
-  // If all chunks of a table are indexed, we don't need anything else except the IndexScan
-  if (indexed_chunks.size() == static_cast<size_t>(table->chunk_count())) return index_scan;
 
   const auto table_scan = _translate_predicate_node_to_table_scan(node, input_operator);
+
+  index_scan->set_included_chunk_ids(indexed_chunks);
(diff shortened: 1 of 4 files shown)


src/lib/operators/index_scan.cpp
```diff
@@ -175,8 +175,18 @@ PosList IndexScan::_scan_chunk(const ChunkID chunk_id) {
       Fail("Unsupported comparison type encountered");
   }
 
-  matches_out.reserve(matches_out.size() + std::distance(range_begin, range_end));
-  std::transform(range_begin, range_end, std::back_inserter(matches_out), to_row_id);
+  // Since index scans only work on data tables, we can gurantee that the PosList references a single chunk without
```
mrks: Typo.

src/lib/operators/index_scan.cpp
```diff
@@ -175,8 +175,18 @@ PosList IndexScan::_scan_chunk(const ChunkID chunk_id) {
...
+  // checking incoming PosLists
+  matches_out.guarantee_single_chunk();
```
mrks: Assertions made in a comment are worthless. I'd add a DebugAssert right above the call that makes sure that the input is in fact a data table. It is too easy to add support for reference tables somewhere else and forget to change this.

src/lib/operators/index_scan.cpp
```diff
@@ -175,8 +175,18 @@ PosList IndexScan::_scan_chunk(const ChunkID chunk_id) {
...
+
+  const auto current_matches_size = matches_out.size();
+  const auto final_matches_size = current_matches_size + static_cast<size_t>(std::distance(range_begin, range_end));
+  matches_out.resize(final_matches_size);
+
+  for (size_t idx = current_matches_size; idx < final_matches_size; ++idx) {
```
mrks: `idx` is not better than `i`. Index of what?

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -212,10 +212,12 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_predicate_node_to_in
   // TableScan(s).
   auto index_scan = std::make_shared<IndexScan>(input_operator, SegmentIndexType::GroupKey, column_ids,
                                                 predicate->predicate_condition, right_values, right_values2);
+  index_scan->set_included_chunk_ids(indexed_chunks);
 
-  const auto table_scan = _translate_predicate_node_to_table_scan(node, input_operator);
+  // If all chunks of a table are indexed, we don't need anything else except the IndexScan
```
mrks: Negative. At translation time, we don't necessarily have a transaction id. Chunks might still be added before we execute the scan and those should not be excluded. In fact, we should add a test making sure that results from new chunks are included.

src/lib/operators/index_scan.cpp
Bensk1: I agree. The comment should not replace an assert, it's more meant like an explanation. There is an assert for the data table input in `IndexScan::_validate_input` already.

mrks: > It is too easy to add support for reference tables somewhere else and forget to change this.

How do you guarantee that whoever adds reference support does not forget about the assumption made **here**?

A different solution would be to simply check for a data table or an input poslist with the same guarantee before you call `guarantee_single_chunk`.

src/lib/logical_query_plan/lqp_translator.cpp
Bensk1: Right. In this case, the changes to the LQPTranslator and the added test are obsolete.

mrks:
Please update PR comment before merging

Bensk1:
Also, the test you proposed is missing.

mrks:
Did you decide against the test?

Bensk1:
Yes, I wanted to add it as a new issue.

PR #1702, opened 2019-06-06
Support Create Table As Select

d-justen:
...as suggested in #1189


mrks:
Is this reviewable now that the PR is merged?

d-justen:
Yes, now that the test executes and actually succeeds it should be ready to be reviewed.

src/lib/sql/sql_translator.cpp
```diff
@@ -995,44 +996,49 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_view(const hsq
...
+      }
 
-    column_definition.name = parser_column_definition->name;
-    column_definition.nullable = parser_column_definition->nullable;
+      column_definition.name = parser_column_definition->name;
+      column_definition.nullable = parser_column_definition->nullable;
+    }
+    input_node = StaticTableNode::make(Table::create_dummy_table(column_definitions));
+  } else {
+    input_node = _translate_select_statement(*create_statement.select);
```
mrks: Readability: I'd change the then and else paths. This branch is so short that it would be easier to read it first, so that you don't have to deal with it anymore.

src/lib/logical_query_plan/create_table_node.cpp
```diff
@@ -3,19 +3,18 @@
...
-      table_name(table_name),
-      column_definitions(column_definitions),
-      if_not_exists(if_not_exists) {}
+CreateTableNode::CreateTableNode(const std::string& table_name, const bool if_not_exists)
+    : BaseNonQueryNode(LQPNodeType::CreateTable), table_name(table_name), if_not_exists(if_not_exists) {}
 
 std::string CreateTableNode::description() const {
   std::ostringstream stream;
 
+  const auto column_definitions = std::dynamic_pointer_cast<StaticTableNode>(left_input())->table->column_definitions();
```
mrzzzrm: and what if left_input is no StaticTableNode? It could be any LQP, right?

src/lib/operators/maintenance/create_table.cpp
```diff
@@ -42,20 +43,33 @@ const std::string CreateTable::description(DescriptionMode description_mode) con
...
+  const auto column_definitions = _input_left->get_output()->column_definitions();
+
   // If IF NOT EXISTS is not set and the table already exists, StorageManager throws an exception
   if (!if_not_exists || !StorageManager::get().has_table(table_name)) {
     // TODO(anybody) chunk size and mvcc not yet specifiable
     const auto table = std::make_shared<Table>(column_definitions, TableType::Data, Chunk::DEFAULT_SIZE, UseMvcc::Yes);
     StorageManager::get().add_table(table_name, table);
   }
+
+  if (input_table_left()->row_count() > 0) {
```
mrzzzrm: Why this branch? Does Insert crash on empty input? otherwise I'd just let Insert handle this case.

mrzzzrm: If it does crash, please add a brief comment a la "// Insert cannot handle emtpy input"

src/lib/operators/maintenance/create_table.hpp
```diff
@@ -1,29 +1,38 @@
...
-#include "operators/abstract_read_only_operator.hpp"
+#include "operators/abstract_read_write_operator.hpp"
+#include "operators/insert.hpp"
 #include "storage/table_column_definition.hpp"
 
 namespace opossum {
 
 // maintenance operator for the "CREATE TABLE" sql statement
-class CreateTable : public AbstractReadOnlyOperator {
+class CreateTable : public AbstractReadWriteOperator {
```
mrzzzrm: Following from what @mrks said below, this can probably stay a AbstractReadOnlyOperator. It should force the internal Insert operator to initialize the begin commit ids to 0 - probably via a dummy TransactionContext

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -74,6 +74,7 @@ ExpressionUnorderedSet ColumnPruningRule::_collect_actually_used_columns(const s
       case LQPNodeType::ShowColumns:
       case LQPNodeType::ShowTables:
       case LQPNodeType::Sort:
+      case LQPNodeType::StaticTable:
```
mrzzzrm: @mrks happy with the name "StaticTableNode"? Does not feel overly intuitive to me, but I don't have better suggestions.

src/test/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -804,16 +807,31 @@ TEST_F(LQPTranslatorTest, CreateTable) {
...
+TEST_F(LQPTranslatorTest, StaticTable) {
+  auto column_definitions = TableColumnDefinitions{};
+  column_definitions.emplace_back("a", DataType::Int, false);
+  column_definitions.emplace_back("b", DataType::Float, true);
+
+  const auto lqp = StaticTableNode::make(Table::create_dummy_table(column_definitions));
+
+  const auto pqp = LQPTranslator{}.translate_node(lqp);
+
+  EXPECT_EQ(pqp->type(), OperatorType::TableWrapper);
```
mrzzzrm: at least test for EXPECT_EQ(pqp->wrapped_table, dummy_table) as well

src/lib/optimizer/strategy/column_pruning_rule.cpp
mrks: I started typing twice and deleted it again. It is basically the LQP equivalent of the table wrapper. At the same time, StoredTableNode/GetTable have different names, too.

src/lib/operators/maintenance/create_table.hpp
mrks: Mmh, I'd rather have proper commit ids here. After all, those are used to identify a snapshot of the database. It's bad enough that we cannot deal with schema modifications, but we should not allow rows to suddenly become visible to already running transactions.

src/lib/operators/maintenance/create_table.cpp
mrks: From a reader's perspective, CREATE TABLE has two cases (column definitions or SELECT as an input). If those two cannot be found in the code because `Insert` magically handles both, this is more confusing than helpful.

src/lib/operators/maintenance/create_table.hpp
mrzzzrm: Okay

src/lib/operators/maintenance/create_table.cpp
mrzzzrm: Worse, this `if` does not properly distinguish the two syntaxes of CREATE TABLE - CREATE TABLE ... FROM could return an empty table as well. In any case, doc'ing is warranted, I guess.

mrks: Good point. We should just doc that both cases can be handled the same.

mrzzzrm: > From a reader's perspective, CREATE TABLE has two cases

Regarding that, I really think "the reader" should be taught to forget about SQL as soon as he enters the LQP/PQP-realm. It's really not helpful to think in SQL terms once the optimizer comes into play. I think we should mentally decouple the CreateTable operator/node from the CREATE TABLE statement - and then it's just fine that there is one path handling both higher level cases.

src/lib/logical_query_plan/create_table_node.cpp
d-justen: Yes, you're right. So this means that the column_definitions may be still unknown. We could either leave out the column_definitions completely or insert the left_inputs description.

src/lib/operators/maintenance/create_table.hpp
```diff
@@ -1,29 +1,38 @@
...
       const std::shared_ptr<AbstractOperator>& copied_input_left,
       const std::shared_ptr<AbstractOperator>& copied_input_right) const override;
 
   void _on_set_parameters(const std::unordered_map<ParameterID, AllTypeVariant>& parameters) override;
+
+  // Commit happens in Insert operator
+  void _on_commit_records(const CommitID cid) override {}
+
+  // Rollback happens in Insert operator
+  void _on_rollback_records() override {}
```
mrks: Are the different MVCC cases tested? I.e., do we check that rows created this way are visible to future transaction but to already running transactions? Are the MVCC effects of a rollback tested?

d-justen: I just added an extra test for this and validate the tables in the other tests now

mrks:
I took the liberty to improve the way the tests check for the correctness of the result. It is preferable to use `EXPECT_TABLE_EQ` instead of some surrogates like the number of columns or the row count. Not only is it more comprehensive, but it is also better at conveying your intention.

d-justen:
Sure, this is way better! I just didn't know that `EXPECT_TABLE_EQ` actually exists.

PR #1704, opened 2019-06-07
Add filled query templates

janetzki:
Close hyrise-mp/hyrise#48
This PR adds the queries for verification against TPC-DS's standard answer set.


mrks:
You'll have to trigger Jenkins again if you set FullCI. I did not yet find a good way of doing that automatically.

Issue #1670, opened 2019-05-17
Extend Plugin Interface


lanice:
While fiddling with the [PCM](https://github.com/opcm/pcm) plugin, two issues came up:

1. The `AbstractPlugin` interface is very limited, as it provides only the two methods `void start()` and `void stop()`, both called during load/unload of the plugin. For something like additional benchmarking metrics, like with PCM, it would be useful if not necessary to have something more interactive, to start/stop performance counting, and to gather the results. One possibility would be to have one generic method, like `trigger`, `lap`, or `result`, that returns an (optional) variant that could contain, well, anything, or maybe it returns a JSON. This method could be used by various plugins for their specific purpose, and in the case of this PCM plugin to trigger the metric collection, returning the result of the last lap. Plugins that don't need that method would just not implement/override it.

2. Even with an extended plugin interface, there is (as far as I've seen) no way to access it from outside the `PluginManager`. I would suggest something like`PluginManager::plugin("PluginName")` that returns the PluginHandle.

Any thoughts on this, @mrks, @Bensk1?

mrks:
1. The whole point of plugins is that the core should not have to know about their presence. If we introduce `PCMPlugin::trigger()` at a defined point in the benchmark, this is not a plug-and-playable plugin anymore. If anything, we might have event listeners that plugins can subscribe to.

2. Same thing. Adding it would not be difficult, but having to access a specific plugin from the core seems to be a smell.

Maybe you could describe what you are trying to achieve?

lanice:
> Maybe you could describe what you are trying to achieve?

Two things: A way to sync the additional metric collection with the benchmark runs/items, and a way to return the results in the end. The results _could_ be stored in a separate files, although I feel like that wouldn't be the most elegant solution.

Event listeners you suggested could work too I guess. But wouldn't we have a similar problem there, that we put specified events into the benchmark, making it not 100% plug-and-playable anymore?

I do think for a plugin like that, enhancing benchmark metric collection, we have to make a compromise somewhere (i.e., give up some of the plug-and-playability).
The only way I see that does not need any change in the interface would be to start metric collection upon plugin `load`, and dump the results into a file upon `unload`. Start metric collection would mean start a separate thread, and read the performance counters e.g. every second.
But then the benchmark runs/items and the plugin results are completely disconnected, and we'd have to manually piece them together for evaluation.

mrks:
How about this (building on #1597 and making BenchmarkRunner inherit from some Listenable interface):

```
[...]

Issue #1677, opened 2019-05-21
Hyrise singleton instance


mrks:
@mrzzzrm and I just had a discussion about global state. The problems are
1. No one knows which variables hold state, as things like the singleton PluginManager are easily overseen
2. As a result, properly resetting the globally state in tests is difficult
3. The Static Initialization Order Fiasco

Having an instance that is constantly passed around bloats the interfaces unnecessarily. Also, we will likely never have more than one instance.

How about a single `Hyrise` class that holds the entire global state? Instead of `StorageManager::get()`, we would use `Hyrise::get().storage_manager`. Also, we could easily reset everything using `get() = Hyrise{}`.

julianmenzler:
Great idea!

Bensk1:
That sounds like a good solution to me. Does it have any downsides?

mrks:
Can't think of any

d-justen:
I'm working on this right now. Is there any reason why `StorageManager`, `TransactionManager` and `PluginManager` should stay Singletons or can this go away?

mrks:
As discussed: They should be regular classes, but constructible only by the Hyrise super class.

Issue #873, opened 2018-05-09
Global Command Line Options


mrks:
Some options should be shared among the binaries for the console, the server, and benchmarks. Maybe even tests.

* `--default-compression`
* `--scheduler`
* `--logger`
* ...?

Maybe even things like default chunk size.

also related to #741, #861, #839

mrzzzrm:
Same as #63?

mrks:
Hmm, similar, but #63 would also include things like "maximum number of workers to spawn per core" which I don't think should be exposed as a CLI option

PR #1705, opened 2019-06-10
Listenable interface that allows plugins to (e.g.) add additional metrics to benchmarks

lanice:
This is the first draft of the Listener interface discussed in #1670.
It is marked WIP and definitely not merge-ready, but I would appreciate any feedback on the concept in general, before I continue to work on this. Because if this version turns out to not be the way we want to handle interaction with plugins, then I'd rather know sooner than later, and change directions :)

The way this currently works:
- We have a new `Listenable` interface (`BenchmarkRunner` now inherits from that)
- The interface provides a method `add_listener` that takes an `Event` (from an enum) and a function to trigger for that event. It also provides the (protected) method `_notify_listeners` that takes an `Event` and an optional JSON payload. `_notify_listeners` basically calls each registered listener function for that event with the JSON payload as argument.
- `add_listener` is called from other objects (in this case: plugins) to listen to events, `_notify_listeners` is only called from the `Listenable` itself.
- The `BenchmarkRunner` now is a `Listenable` and, for now, notifies listeners when an item run is started, an item run is finished, and before the final report is created.
- The `BenchmarkRunner` additionally provides a new method `add_to_json_report` that can be used to add some JSON results from outside the runner itself (which is used by the plugin).
- The `PluginManager::load_plugin` has an additional optional `Listenable` parameter that gets passed on to the newly loaded plugin so that it can register itself and it's functions to the `Listenable`. That is something I'm NOT super happy with, and I spent some good time to figure out how to connect a plugin that's completely independent to the listenable, in our case the `BenchmarkRunner`. The other option I though about was to make the benchmark runner a Singleton and the `add_listener` functions static. I chose the non-singleton route for now, but like I said, anything is open to change if there is a better way.

The new `PcmPlugin` is independent of this repository, therefore not visible here in this PR. The `PcmPlugin::register_listenable` method (which is called by the `PluginManager` when the plugin is loaded) looks like this:
```
void PcmPlugin::register_listenable(std::shared_ptr<Listenable> listenable) {
  listenable->add_listener(Event::ItemRunStarted, [&](const nlohmann::json &payload){this->_start_counting(payload);});
  listenable->add_listener(Event::ItemRunFinished, [&](const nlohmann::json &payload){this->_stop_counting(payload);});

  Assert(std::dynamic_pointer_cast<BenchmarkRunner>(listenable), "Can register PcmPlugin only with BenchmarkRunner as Listenable.");
  listenable->add_listener(Event::CreateReport, [&](const nlohmann::json &payload){
    std::dynamic_pointer_cast<BenchmarkRunner>(listenable)->add_to_json_report("PcmPlugin", _results);
  });
}
```

- Another thing I'm not happy with is how the `Events` are stored. Right now, the `Listenable` interface holds the `Event` enum, and I added the appropriate events for the `BenchmarkRunner` there. However, that doesn't really make sense, basically each class that inherits from `Listenable` should have its own events. But enums don't have inheritance, and the calling `Listenable::add_listener` takes an `Event` as defined in `listenable.hpp`, so if each listenable implements their own event enums, they also would have to implement their own `add_listener`, and `_notify_listeners`...? I might have blinders on for this, since I wrapped my head so many times around these things.

Sooo, happy to hear feedback and suggestions for change, even if it is "You know what, scrap the whole PCM as plugin thing, we should really put this into the archive instead where nobody gets hurt" :sweat_smile:


src/benchmark/tpch_benchmark.cpp
```diff
@@ -49,13 +50,16 @@ int main(int argc, char* argv[]) {
   cli_options.add_options()
     ("s,scale", "Database scale factor (1.0 ~ 1GB)", cxxopts::value<float>()->default_value("1"))
     ("q,queries", "Specify queries to run (comma-separated query ids, e.g. \"--queries 1,3,19\"), default is all", cxxopts::value<std::string>()) // NOLINT
-    ("use_prepared_statements", "Use prepared statements instead of random SQL strings", cxxopts::value<bool>()->default_value("false")); // NOLINT
+    ("use_prepared_statements", "Use prepared statements instead of random SQL strings", cxxopts::value<bool>()->default_value("false")) // NOLINT
+    ("pcm", "Path to PcmPlugin.so", cxxopts::value<std::string>()->default_value("")); // NOLINT
```
mrks: We should rather have a (global) option `-p plugin` (#873).

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -100,6 +100,8 @@ void BenchmarkRunner::run() {
   if (_config.output_file_path) {
     if (!_config.verify && !_config.enable_visualization) {
       std::ofstream output_file(*_config.output_file_path);
+      _listener_reports = nlohmann::json{};
+      _notify_listeners(Event::CreateReport);
```
mrks: Why is `_listener_reports` a member but `listener_payload` a parameter? I like the second approach better.

src/benchmarklib/listenable.cpp
```diff
@@ -0,0 +1,22 @@
+#include "listenable.hpp"
+
+namespace opossum {
+
+void Listenable::add_listener(const Event event, EventListener listener) {
```
mrks: `listen(...)`?

src/benchmarklib/listenable.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace opossum {
+
+enum class Event {
+  ItemRunStarted,
+  ItemRunFinished,
+  CreateReport,
```
mrks: Having one global list of events seems to be a smell. How about letting each Listenable create its own `enum class` and use CRTP (`: public Listenable<BenchmarkRunner`)?

src/benchmarklib/listenable.cpp
```diff
@@ -0,0 +1,22 @@
...
+  if (_event_listeners.find(event) == _event_listeners.end()) {
```
mrks: If we were to use CRTP here, we could save us that map

src/benchmarklib/listenable.hpp
```diff
@@ -0,0 +1,32 @@
...
+};
+
+using EventListener = std::function<void(const nlohmann::json& payload)>;
```
mrks: JSON is a format for serializing something into a file and shouldn’t be abused for passing information around. We could define different message structs and accept a common super class here, but that might only lead to an illusion of type-safety. As the handler should know what type to expect, I’d even consider `std::any`.

src/lib/utils/plugin_manager.hpp
```diff
@@ -23,7 +25,7 @@ class PluginManager : public Singleton<PluginManager> {
   friend class SingletonTest;
 
  public:
-  void load_plugin(const std::filesystem::path& path);
+  void load_plugin(const std::filesystem::path& path, std::shared_ptr<Listenable> listenable = nullptr);
```
mrks: I don’t like this part. What if the plugin should attach to two things? No, the plugin needs to find whatever it wants to attach to on its own.

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -429,4 +435,8 @@ nlohmann::json BenchmarkRunner::create_context(const BenchmarkConfig& config) {
       {"GIT-HASH", GIT_HEAD_SHA1 + std::string(GIT_IS_DIRTY ? "-dirty" : "")}};
 }
 
+void BenchmarkRunner::add_to_json_report(const std::string& listener_name, const nlohmann::json& report) {
```
mrks: Naming: It is unclear on which level of the report this is added.

src/lib/utils/plugin_manager.hpp
mrks: > The other option I though about was to make the benchmark runner a Singleton and the add_listener functions static.

Maybe #1677 would help here?

src/benchmarklib/listenable.cpp
lanice: Isn't that the opposite? `Listenable::listen(...)` would mean that the `Listenable` is the one that listens for the event. But the object that is calling this method is actually the listener (or provides the listener), so it adds the listener to the Listenable with `add_listener`.

src/benchmarklib/listenable.hpp
lanice: I will change it to `std::any` then, good idea!

src/lib/utils/plugin_manager.hpp
lanice: Yes, I agree that this is not ideal. If the plugin should find what it wants to attach to on its own, I don't think we can get around making the benchmark runner a Singleton, which I'm okay with. Works with #1677 I'd say, although it doesn't really make a difference I think. We'd just have to decide that the benchmark runner is a single global instance then, either as Singleton or accessible through the Hyrise Singleton.

src/benchmarklib/listenable.hpp
lanice: I will look into that

src/benchmarklib/benchmark_runner.cpp
lanice: We now could add the listener reports to the payload for the `ItemFinished` event, as part of a tuple (using `std::any` as payload type). That would still mean passing it (the report only) as JSON, but since it will be used to serialize later that is okay?

src/benchmarklib/listenable.cpp
mrks: At least it's ambiguous. Never mind.

src/benchmarklib/benchmark_runner.cpp
mrks: I don't fully understand what would be passed as a JSON. Do you mean the data that is written into the result JSON? Sure, that's a different thing.

lanice: Yes. Just checking :)

src/lib/utils/plugin_manager.hpp
lanice: I'm having some trouble converting `BenchmarkRunner` into a Singleton. The problem is the `reset()` method, which assigns a newly created `BenchmarkRunner` to the static instance. `BenchmarkRunner` however has no copy-assignment constructor. They get implicitly deleted because of some of the fields, like the `const BenchmarkConfig`, or the `unique_ptr's`. It might be a bigger change than I thought at first.

mrks:
You'll ping me once this is ready for another review, right?

lanice:
@mrks, ready for another look. One change (BenchmarkRunner->Singleton) that is needed for this is still missing, however, I would like to do that separately.

src/benchmarklib/listenable.cpp
lanice: Hmm, not sure if that's right. We still need to map the specific Events to their listeners, right?

tobodner:
This seems to have been inactive for almost half a year. How shall we proceed?

mrks:
Part of the backlog that is way too long - feel free to have a look

tobodner:
> Part of the backlog that is way too long - feel free to have a look

This is tagged WIP. Also, I am not sure what the original motivation and stakeholders were. Shall we discuss it briefly in the next roundtable?

mrks:
Closing this for lack of resources. I still want memory bus information in, so hopefully, this will get picked up at a later time.

PR #1710, opened 2019-06-14
Add tools directory of the official TPC-DS toolkit

mweisgut:
This PR only adds the `tools` directory of the official TPC-DS toolkit to Hyrise's third_party directory.
Only file `VERSION` is added to the original files of the tools directory.


third_party/tpcds-tools/CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
+cmake_minimum_required(VERSION 3.10)
```
mrks: How is this different from the `tpcds-tools` repo?

mweisgut:
Added source files will be replaced by using a submodule.

PR #1715, opened 2019-06-19
Implement WITH queries

julianmenzler:
`WITH` is already supported by the `sql-parser`. This PR introduces support for it within the Hyrise SQL Translator.


src/lib/sql/sql_translator.cpp
```diff
@@ -266,6 +277,26 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_select_statement(cons
...
+  SQLTranslator with_translator{_use_mvcc, nullptr, _parameter_id_allocator, _with_descriptions};
+  auto lqp = with_translator._translate_select_statement(*desc.select);
+
+  // Save mappings: ColumnID -> ColumnName
+  std::unordered_map<ColumnID, std::string> column_names;
+  for (auto column_id = ColumnID{0}; column_id < lqp->column_expressions().size(); ++column_id) {
+    const auto identifiers =
+        with_translator._sql_identifier_resolver->get_expression_identifiers(lqp->column_expressions()[column_id]);
+    for (const auto& identifier : identifiers) {
+      column_names.insert_or_assign(column_id, identifier.column_name);
```
julianmenzler: This is something I want to discuss with you @mrzzzrm today.

mrzzzrm: Has to be tomorrow, sorry. Today I won't be in

src/lib/sql/sql_translator.cpp
```diff
@@ -130,7 +130,8 @@ bool is_trivial_join_predicate(const AbstractExpression& expression, const Abstr
 namespace opossum {
 
 SQLTranslator::SQLTranslator(const UseMvcc use_mvcc)
-    : SQLTranslator(use_mvcc, nullptr, std::make_shared<ParameterIDAllocator>()) {}
+    : SQLTranslator(use_mvcc, nullptr, std::make_shared<ParameterIDAllocator>(),
+                    std::unordered_map<std::string, std::shared_ptr<LQPView>>()) {}
```
mrzzzrm: `{}` would do

src/lib/sql/sql_translator.cpp
```diff
@@ -191,20 +194,28 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_statement(const hsql:
...
+  // 7. SELECT clause (incl. DISTINCT)
+  // 8. UNION clause
+  // 9. ORDER BY clause
+  // 10. LIMIT clause
 
   AssertInput(select.selectList, "SELECT list needs to exist");
   AssertInput(!select.selectList->empty(), "SELECT list needs to have entries");
   AssertInput(!select.unionSelect, "Set operations (UNION/INTERSECT/...) are not supported yet");
 
+  // Translate WITH clause
```
mrzzzrm: clause**s**, right?

src/lib/sql/sql_translator.cpp
```diff
@@ -191,20 +194,28 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_statement(const hsql:
...
+  if (select.withDescriptions) {
+    for (const auto& desc : *select.withDescriptions) {
+      _translate_hsql_with_description(*desc);
```
mrzzzrm: for consistent naming, I guess that should be called "_translate_hsql_with_description".

Since "desc" is a hyrise variable, it should have an unabbreviated name

src/lib/sql/sql_translator.cpp
```diff
@@ -266,6 +277,26 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_select_statement(cons
   return _current_lqp;
 }
 
+void SQLTranslator::_translate_hsql_with_description(hsql::WithDescription& desc) {
+  SQLTranslator with_translator{_use_mvcc, nullptr, _parameter_id_allocator, _with_descriptions};
```
mrzzzrm: did you test placeholders in WITH? `WITH a AS SELECT * FROM t WHERE a > ? SELECT ...`?

src/lib/sql/sql_translator.cpp
```diff
@@ -266,6 +277,26 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_select_statement(cons
...
+  auto lqp = with_translator._translate_select_statement(*desc.select);
```
mrzzzrm: can be const auto

src/lib/sql/sql_translator.cpp
mrzzzrm: So basically you're invalidating all but one name of the column, right? How did you handle this for views @janetzki ?

src/lib/sql/sql_translator.cpp
```diff
@@ -466,7 +497,27 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_origin(const hsq
 
   switch (hsql_table_ref.type) {
     case hsql::kTableName: {
-      if (StorageManager::get().has_table(hsql_table_ref.name)) {
+      // WITH descriptions or sub-queries are treated as though they were inline views or tables
+      // They mask existing tables or views with the same name.
+      if (_with_descriptions.count(hsql_table_ref.name)) {
+        const auto iter = _with_descriptions.find(hsql_table_ref.name);
```
mrzzzrm: count() searches, find() searches - a single find() should be sufficient. Also, names iterators I like better, e.g. "with_descriptions_iterator"

src/lib/sql/sql_translator.cpp
```diff
@@ -466,7 +497,27 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_origin(const hsq
...
+        AssertInput(iter != _with_descriptions.end(), "No such WITH query named '" + hsql_table_ref.name + "'");
```
mrzzzrm: How could that ever trigger if `_with_descriptions.count(hsql_table_ref.name) > 0`, as the if below ensures?

src/test/sql/sql_translator_test.cpp
```diff
@@ -1996,4 +1996,91 @@ TEST_F(SQLTranslatorTest, CatchInputErrors) {
...
+              stored_table_node_int_int_int));
+  // clang-format on
+  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+}
+
+TEST_F(SQLTranslatorTest, WithClauseSingleQueryAlias) {
+  const auto actual_lqp_alias = compile_query(
+      "WITH "
+      "with_query AS (SELECT a AS x FROM int_int_int) "
+      "SELECT * FROM with_query WHERE x > 123;");
```
mrzzzrm: `SELECT * FROM with_query WHERE x > 123` would throw an exception, currently, right? I think this should be fixed.

src/test/sql/sql_translator_test.cpp
```diff
@@ -1996,4 +1996,91 @@ TEST_F(SQLTranslatorTest, CatchInputErrors) {
...
+  const auto aliases = std::vector<std::string>({"x"});
+  const auto expressions = expression_vector(int_int_int_a);
+
+  // clang-format off
+  const auto expected_lqp_alias =
+      AliasNode::make(expressions, aliases,
```
mrzzzrm: indent of 2 as we do everywhere else should be preferred.

src/test/sql/sql_translator_test.cpp
```diff
@@ -1996,4 +1996,91 @@ TEST_F(SQLTranslatorTest, CatchInputErrors) {
...
+                      stored_table_node_int_int_int))));
+  // clang-format on
+  EXPECT_LQP_EQ(actual_lqp_alias, expected_lqp_alias);
+}
+
+TEST_F(SQLTranslatorTest, WithClauseDoubleQuery) {
+  // WITH queries build on top of each other
+  const auto actual_lqp = compile_query(
+      "WITH "
+      "with_query1 AS (SELECT a, b FROM int_int_int), "
```
mrzzzrm: a WHERE here and a WHERE in the line below would make this test a bit more interesting.

src/test/sql/sql_translator_test.cpp
```diff
@@ -1996,4 +1996,91 @@ TEST_F(SQLTranslatorTest, CatchInputErrors) {
...
+  // clang-format on
+  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+}
+
+TEST_F(SQLTranslatorTest, WithClauseDoubleQueryMasking) {
+  // Second WITH query masks first one
+  const auto actual_lqp = compile_query(
+      "WITH "
+      "with_query AS (SELECT a, b FROM int_int_int), "
+      "with_query AS (SELECT b FROM with_query) "
```
mrzzzrm: do other dbs allow this? if not, we shouldn't either, probably.

src/test/sql/sql_translator_test.cpp
```diff
@@ -1996,4 +1996,91 @@ TEST_F(SQLTranslatorTest, CatchInputErrors) {
...
+
+  // clang-format off
+  const auto expected_lqp =
+      ProjectionNode::make(expression_vector(int_int_int_b),
+           ProjectionNode::make(expression_vector(int_int_int_a, int_int_int_b),
+               stored_table_node_int_int_int));
+  // clang-format on
+  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+}
+
```
mrzzzrm: random line

why was submodule third_party/tpcds-result-reproduction  changed?

src/lib/sql/sql_translator.cpp
julianmenzler: I thought of the WITH clause as a whole with multiple statements or descriptions: 
```SQL
WITH
a AS (..), 
b AS (..), 
c AS (..)
```
But this is my interpretation.

src/lib/sql/sql_translator.cpp
janetzki: Good point. I did not make any changes in `_translate_create_view()` so far. I'll update you as soon as I have tested `CREATE VIEW` for hyrise-mp/hyrise#38.

src/lib/sql/sql_translator.cpp
mrzzzrm: Ah, I was mistaken about the syntax. If there is ever only one WITH, leave it at "clause"

src/test/sql/sql_translator_test.cpp
julianmenzler: Unfortunately I do not have any offical SQL reference. But I had success testing this query with `Oracle11g` (via [sqltest.net](https://sqltest.net). 
But, I just noticed that the results Oracle11g returns are not correct or as expected. Oracle11g seems to use the first with_description, ignoring the second re-declaration.
I guess we should prohibits this kind of query then.

src/test/sql/sql_translator_test.cpp
julianmenzler: > why was submodule third_party/tpcds-result-reproduction changed?
No reason. Slipped in, somehow.

src/lib/sql/sql_translator.cpp
julianmenzler: > for consistent naming, I guess that should be called "_translate_hsql_with_description".

Am I blind, I do not see any difference.

> Since "desc" is a hyrise variable, it should have an unabbreviated name

So `*with_description` instead of `*desc`?

src/lib/sql/sql_translator.cpp
julianmenzler: Oh these placeholders are for prepared statements, right? Not yet.

src/lib/sql/sql_translator.cpp
julianmenzler: Good catch!

src/lib/sql/sql_translator.cpp
julianmenzler: Correct 🙈

src/test/sql/sql_translator_test.cpp
julianmenzler: I do not quite understand.. Do you mean the `;` ?

src/lib/sql/sql_translator.cpp
mrks: > I may be blind, but I do not see the difference.

Neither do I

> So *with_description instead of *desc?

Yes.

src/lib/sql/sql_translator.cpp
mrks: ping?

src/lib/sql/sql_translator.cpp
mrks: This was marked as resolved but doesn't look like it was.

src/lib/sql/sql_translator.cpp
mrks: This PR looks like the comments have been seen, but not addressed yet. Skipping the rest of the comments and my own review for now.

src/lib/sql/sql_translator.cpp
julianmenzler: Did you update LQPView yet? @janetzki

src/lib/sql/sql_translator.cpp
julianmenzler: Now it is.

src/lib/sql/sql_translator.cpp
julianmenzler: Correct, I just started addressing them.

src/lib/sql/sql_translator.cpp
janetzki: Yes, in https://github.com/hyrise/hyrise/pull/1692/commits/a926f691958e8cd2b2ffca648843d8275dee2eb5.

mrks:
What's the status here?

src/lib/sql/sql_translator.cpp
```diff
@@ -466,7 +497,26 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_origin(const hsq
 
   switch (hsql_table_ref.type) {
     case hsql::kTableName: {
-      if (StorageManager::get().has_table(hsql_table_ref.name)) {
+
+      // WITH descriptions or sub-queries are treated as though they were inline views or tables
```
janetzki:
```suggestion
      // WITH descriptions or subqueries are treated as though they were inline views or tables.
```

src/lib/sql/sql_translator.cpp
```diff
@@ -266,6 +277,26 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_select_statement(cons
...
 
+void SQLTranslator::_translate_hsql_with_description(hsql::WithDescription& desc) {
+  SQLTranslator with_translator{_use_mvcc, nullptr, _parameter_id_allocator, _with_descriptions};
+  const auto lqp = with_translator._translate_select_statement(*desc.select);
+
+  // Save mappings: ColumnID -> ColumnName
+  std::unordered_map<ColumnID, std::string> column_names;
+  for (auto column_id = ColumnID{0}; column_id < lqp->column_expressions().size(); ++column_id) {
+    const auto identifiers =
+        with_translator._sql_identifier_resolver->get_expression_identifiers(lqp->column_expressions()[column_id]);
```
janetzki: This empty line seems odd.

src/test/sql/sql_translator_test.cpp
```diff
@@ -1996,4 +1996,91 @@ TEST_F(SQLTranslatorTest, CatchInputErrors) {
...
+TEST_F(SQLTranslatorTest, WithClauseSingleQuerySimple) {
+  const auto actual_lqp = compile_query(
+      "WITH "
+      "with_query AS (SELECT a, b FROM int_int_int) "
+      "SELECT * FROM with_query WHERE a > 123;");
+  // clang-format off
+  const auto expected_lqp =
+      PredicateNode::make(greater_than_(int_int_int_a, value_(123)),
+        ProjectionNode::make(expression_vector(int_int_int_a, int_int_int_b),
+          stored_table_node_int_int_int));
```
janetzki: Please fix the indentation.

src/lib/sql/sql_translator.cpp
janetzki: You might want to use `_inflated_select_list_elements[column_id].identifiers` instead of `get_expression_identifiers(...)` since `get_expression_identifiers(...)` would return `{"a", "a1", "a2"}` for both `ColumnID`s in `SELECT a AS a1, a AS a2`.

src/lib/sql/sql_translator.cpp
mrks: ping ping

src/test/sql/sql_translator_test.cpp
mrks: Can't reproduce what was meant either. Maybe he was thinking about `a > 123`, but that is addressed in a different issue.

src/lib/sql/sql_translator.cpp
julianmenzler: Currently, I'm try to understand the placeholder implementation to create a test afterwards.

src/test/sql/sql_translator_test.cpp
mrzzzrm: He probably meant `a >123` ;)

src/lib/sql/sql_translator.cpp
julianmenzler: How do you obtain identifiers from `_inflated_select_list_expressions`?
```
std::vector<std::shared_ptr<AbstractExpression>> _inflated_select_list_expressions
``` 

Unfortunately, I don't get your idea.

julianmenzler: Good point. 
For this change I have to merge the current master though. Before, we should merge #1753 to avoid migration efforts regarding `LQPView`.

julianmenzler:
Except for @janetzki's last suggestion, blocked by #1753, all change requests should be adressed.

src/lib/sql/sql_translator.cpp
janetzki: I have to trust you that you will address this because I will not be available until Aug. 6. ;)

mrks:
@julianmenzler Is this ready to go once the merge conflict is resolved?

mrks:
Ah, never mind. That was just the master merge.

julianmenzler:
Ready to go

mrks:
It sounds as if a5632c4 fixed a bug for which @janetzki provided an example. Does it make sense to add a test that covers this?

julianmenzler:
Correct. It is rather an edge case though. The test should be short, so I'll write one.
Another thought came to my mind regarding this. I have not tested aggregate functions yet.

julianmenzler:
I just added two more tests for WITH queries containing aggregate functions. 

> It sounds as if [a5632c4](https://github.com/hyrise/hyrise/commit/a5632c41883687efafce0fe819d747bb2b371d6e) fixed a bug for which @janetzki provided an example. Does it make sense to add a test that covers this?

I created a test and found a bug in the `SQLTranslator::_translate_table_origin`. The bug affects regular views as well (#1768). It does not make sense to address the bug in this PR. I will fix it separately and add the associated WITH test afterwards.

PR #1716, opened 2019-06-19
File-Based TPC-DS Benchmark

mweisgut:
This PR contains the current file-based implementation of the master project's TPC-DS benchchmark.
Both the data and the queries are currently generated file-based.
The corresponding query files can be found in `third_party/tpcds-result-reproduction/query_qualification/`, the table files in `resources/benchmark/tpcds/tables`.

To significantly speed up the speed of SQLite verification, similar to the JoinOrderingBenchmark, the TPC-DS benchmark implementation creates indices on the foreign keys for the SQLite verification.

The benchmark uses the official tpc-ds toolkit's directory `tools` to generate the binary `dsdgen`. This binary is used to generate the table data (csv) files.
These files are stored in `resources/benchmark/tpcds/tables`.


mrks:
Why does the build process generate data again?

```
cd /var/lib/jenkins/workspace/hyrise_hyrise_PR-1716/third_party/tpcds-kit/tools && /var/lib/jenkins/workspace/hyrise_hyrise_PR-1716/clang-debug/distcomp -i /var/lib/jenkins/workspace/hyrise_hyrise_PR-1716/third_party/tpcds-kit/tools/tpcds.dst -o /var/lib/jenkins/workspace/hyrise_hyrise_PR-1716/third_party/tpcds-kit/tools/tpcds.idx
```

mweisgut:
> Why does the build process generate data again?
> 
> ```
> cd /var/lib/jenkins/workspace/hyrise_hyrise_PR-1716/third_party/tpcds-kit/tools && /var/lib/jenkins/workspace/hyrise_hyrise_PR-1716/clang-debug/distcomp -i /var/lib/jenkins/workspace/hyrise_hyrise_PR-1716/third_party/tpcds-kit/tools/tpcds.dst -o /var/lib/jenkins/workspace/hyrise_hyrise_PR-1716/third_party/tpcds-kit/tools/tpcds.idx
> ```

The distcomp utility builds a tpcds.idx file which defines the data distribution.
This idx-file is used by the dsdgen binary and has to be in the same directory as the dsdgen binary.

mrks:
Independent of whether it crashes or not, this doesn't sound like something that should be part of the build process. Is there any chance to execute it when we actually need it?

mweisgut:
> Independent of whether it crashes or not, this doesn't sound like something that should be part of the build process. Is there any chance to execute it when we actually need it?

Sure, we could use a script which builds the dsdgen and executes it. Temporarily this was the solution of this branch, see https://github.com/hyrise/hyrise/pull/1716/commits/e628e337717b01905ba6c7ab0d4986f80064376a . Do you think this is the better approach?

mrks:
Can we continue to build the tools during the build phase? That should not be part of the script that is executed at runtime.

mweisgut:
We have to execute target `mkheader` first before building `distcomp`.
`mkheader` creates some files which are required for the build process of `distcomp`.
`distcomp` is required for creating the mentiones file `tpcds.idx`. Without this file, we are unable to execute `dsdgen`.

mweisgut:
We also have to execute `distcomp` to be able to build `dsdgen`. `distcomp` creates a `tpcds.idx.h` which is required of the `dsdgen` build preocess.

@pascalhpi this discussion could also be relevant for you since your setup for the TableGenerator is similar.

pascalhpi:
For the TpcdsTableGenerator we probably won't need all those targets (distcomp mkheader dsdgen) and header files. We can use the mk_<tpcds_tuple> functions with only the tpcds.idx file, which is a binary file, about 600 kB. To save us from building and executing distcomp and mkheader I'd suggest just commiting tpcds.idx

mweisgut:
The current CI report looks much better. I added column.h, table.h, stream.h, tpcds.idx and tpcs.idx.h to the submodule. Now we only have to repair the build process for Mac.

mrks:
`#include <values.h>` looks odd. Shouldn't it be `"`?

mweisgut:
Following [these steps](https://github.com/cartershanklin/hive-testbench/issues/3#issuecomment-307909263) fixed the issue.

Jenkinsfile
```diff
@@ -75,7 +75,7 @@ try {
 
         parallel clangDebug: {
           stage("clang-debug") {
-            sh "export CCACHE_BASEDIR=`pwd`; cd clang-debug && make all -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 3))"
+            sh "export CCACHE_BASEDIR=`pwd`; cd clang-debug && VERBOSE=1 make all -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 3))"
```
mrzzzrm: intended?

src/benchmarklib/sqlite_add_indices.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include "benchmark_runner.hpp"
+#include "storage/storage_manager.hpp"
+#include "utils/sqlite_wrapper.hpp"
+#include "utils/timer.hpp"
+
+namespace opossum {
+
+void add_indices_to_sqlite(const std::string& schema_file_path, const std::string& create_indices_file_path,
+                           BenchmarkRunner& benchmark_runner) {
```
mrzzzrm: should take "sqlite_wrapper" as a param, uses nothing else of the benchmark runner. is this tested?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include "storage/chunk_encoder.hpp"
+#include "utils/assert.hpp"
+#include "visualization/lqp_visualizer.hpp"
+#include "visualization/pqp_visualizer.hpp"
+
+using namespace opossum;  // NOLINT
+
+namespace {
+
+bool data_files_available(const std::string& table_path);
```
mrzzzrm: why don't you just but those functions here? It's quite common for main() to be at the bottom

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,172 @@
...
+const std::unordered_set<std::string> filename_blacklist();
+
+}  // namespace
+
+int main(int argc, char* argv[]) {
+  const std::string binary_path = argv[0];
```
mrzzzrm: const auto binary_path = std::string{argv[0]} for consistency, maybe?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,172 @@
...
+  const std::string binary_directory = binary_path.substr(0, binary_path.find_last_of("/"));
```
mrzzzrm: I am sure the filesystem lib has a "cleaner" version of this?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,172 @@
...
+    if (!std::filesystem::exists(table_path + "/" + table + ".csv")) {
+      return false;
+    }
+  }
+  return true;
+}
+
+const std::unordered_set<std::string> filename_blacklist() {
+  auto filename_blacklist = std::unordered_set<std::string>{};
+  const auto blacklist_file_path = "resources/benchmark/tpcds/query_blacklist.cfg";
```
mrzzzrm: assuming you wrote this query_blacklist.cfg yourself to exclude queries hyrise doesn't support (add a comment what filename_blacklist() does?) - we should use JSON instead of inventing our own "fileformat" (I see comment support below).

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,172 @@
...
+  Assert(std::filesystem::is_directory(table_path), "Table path (" + table_path + ") has to be a directory.");
+  Assert(std::filesystem::exists(std::filesystem::path{query_path + "/01.sql"}), "Queries have to be available.");
+  Assert(std::filesystem::exists(std::filesystem::path{table_path + "/call_center.csv.json"}),
+         "Table schemes have to be available.");
+
+  if (!data_files_available(table_path)) {
+    if (std::filesystem::exists(std::filesystem::path{binary_directory + "/dsdgen"})) {
+      const auto files_setup_return =
+          system(("cd " + binary_directory + " && ./dsdgen -scale " + std::to_string(scale_factor) +
+                  " -dir ../resources/benchmark/tpcds/tables -terminate n -verbose -suffix .csv -f")
```
mrzzzrm: So the hyriseBenchmarkTPCDS can only be run from one directory, right? Which is that? Please add an Assert with an informative message that makes sure that it is run from the right directory. Keep in mind that we have a "resources" symlink in every build dir.

Jenkinsfile
mweisgut: intended by @mrks

mrks: Just added for debugging, forgot to remove it.

src/benchmark/tpcds_benchmark.cpp
mrks: but = put?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -0,0 +1,172 @@
...
+                     .c_str());
+      Assert(files_setup_return == 0, "Generating table data files failed.");
+    } else {
+      Fail("Could not find 'dsdgen' in your build directory. Did you run the benchmark from the project root dir?");
```
mrks: Whenever you type `else Fail(...)`, there is a good chance you meant `Assert(...)`

src/benchmark/tpcds_benchmark.cpp
mrks: I find the current format more readable than `[{query: 15, executable: false}, ...]`. The "comment support" is for excluding queries (similar to the sqlparser tests). Not sure what we could gain from using JSON (and if it's worth spending time on it).

src/benchmarklib/sqlite_add_indices.cpp
mweisgut: Now it is.

mweisgut:
@mrks I added a test for the add_indices_to_sqlite after you approved this PR. Could you please check this (hopefully one last time)?

PR #1721, opened 2019-06-23
Make BenchmarkRunner a Singleton

lanice:
Discussed in (and needed by) #1705.


src/benchmarklib/benchmark_runner.cpp
```diff
@@ -26,7 +26,22 @@
...
+}
+
+BenchmarkRunner& BenchmarkRunner::operator=(BenchmarkRunner&& other) {
+  _config = std::move(other._config);
+  _benchmark_item_runner = std::move(other._benchmark_item_runner);
+  _table_generator = std::move(other._table_generator);
+  _results = std::move(other._results);
+  _context = std::move(other._context);
+  return *this;
+}
```
lanice: Not sure why I can't use the _default_'ed move-assignment operator here instead of providing one. The compiler then always chooses the _delete_'d copy-assignment operator.

mrks:
Waiting for #1677, which should make this easier.

mrks:
Superseded by #1812

PR #1723, opened 2019-06-24
Verification method: dedicated expected result files

mweisgut:
This PR is based on #1716 and integrates a new supplemental verification method. Executing TPC-DS queries using the SQLite verification showed that some queries led to a crash of the benchmark since some features which are present in the TPC-DS queries are not supported by SQLite.
This new verification method is based on provided dedicated expected results (.tbl files for benchmark items) and is currently only used for queries which cannot be verified using the previous (SQLite) technique. This method can only be executed for static benchmark items since the expected result is hard coded.

In this PR, the supplemental verification method is only effectively used for benchmarks which use the `FileBasedBenchmarkItemRunner`. For it's usage, the path to the dedicated expected result directory has to be passed to the `FileBasedBenchmarkItemRunner`. If the `--verify` option is used and the passed dedicated expected result directory has tbl-files which corresponds to the executed benchmark items, the actual item results are compared with the dedicated expected results. If an dedicated expected result is provided for a specific benchmark item, the SQLite verification is **not executed** for this specific item. To assign a dedicated expected result to a benchmark item, the expected result filename (without extension) has to equal the benchmark item filename. 

Example:
The query directory contains the files `01.sql`, `02.sql`, `03.sql`.
The dedicated expected results directory contains the files `01.tbl`, `03.tbl`.
If we execute the benchmark with verification option, the dedicated expected result verification is executed for the benchmark items `01.sql` and `03.sql`. The SQLite verification is only executed for the benchmark item `02.sql`.


src/benchmark/tpcds_benchmark.cpp
```diff
@@ -148,7 +148,8 @@ int main(int argc, char* argv[]) {
 
   Assert(data_files_available(table_path), "Generating table data files failed.");
 
-  auto query_generator = std::make_unique<FileBasedBenchmarkItemRunner>(config, query_path, filename_blacklist());
+  auto query_generator = std::make_unique<FileBasedBenchmarkItemRunner>(
+      config, query_path, filename_blacklist(), std::nullopt, "third_party/tpcds-result-reproduction/answer_sets_tbl");
```
mrks: For my taste, this thing is taking too many arguments that are only semi-related to its original job. Construction should create a complete object, but optional things such as the answer sets can be added later on.

src/benchmarklib/abstract_benchmark_item_runner.cpp
```diff
@@ -22,7 +23,12 @@ std::pair<std::vector<SQLPipelineMetrics>, bool> AbstractBenchmarkItemRunner::ex
     visualize_prefix = std::move(name);
   }
 
-  BenchmarkSQLExecutor sql_executor(_config->enable_jit, _sqlite_wrapper, visualize_prefix);
+  std::shared_ptr<const Table> expected_result_table = nullptr;
+  if (item_id < _dedicated_expected_results.size()) {
```
mrks: Is `_dedicated_expected_results.size() > 0` but `<= item_id` a valid case or would this be more speaking?
```
if (!_dedicated_expected_results.empty() {
  expected_result_table = _dedicated_expected_results.at(item_id);
```

src/benchmarklib/abstract_benchmark_item_runner.cpp
```diff
@@ -22,7 +23,12 @@ std::pair<std::vector<SQLPipelineMetrics>, bool> AbstractBenchmarkItemRunner::ex
...
+    expected_result_table = _dedicated_expected_results[item_id];
+  }
+
+  BenchmarkSQLExecutor sql_executor(_config->enable_jit, _sqlite_wrapper, visualize_prefix, expected_result_table);
```
mrks: The sql executor can be used for more than just a single SELECT (and will be in the TPC-C). This somewhat breaks its interface.

src/benchmarklib/abstract_benchmark_item_runner.hpp
```diff
@@ -35,6 +35,9 @@ class AbstractBenchmarkItemRunner {
   // Returns the BenchmarkItemIDs of all selected items
   virtual const std::vector<BenchmarkItemID>& items() const = 0;
 
+  // Loads the decidated expected restuls as tables into _dedicated_expected_results
```
mrks: typo

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
```diff
@@ -489,4 +489,8 @@ std::string TPCHBenchmarkItemRunner::_substitute_placeholders(const BenchmarkIte
   }
 }
 
+void TPCHBenchmarkItemRunner::load_dedicated_expected_results() {
+  // do nothing; dedicated expected result verification is not supported for the TPCHBenchmarkItemRunner
```
mrks: If it is not necessary to implement this method, simply don't make it pure virtual.

src/benchmarklib/file_based_benchmark_item_runner.cpp
```diff
@@ -5,19 +5,23 @@
 #include <fstream>
 
 #include "SQLParser.h"
+#include "operators/print.hpp"
```
mweisgut: Remove this

src/benchmarklib/abstract_benchmark_item_runner.cpp
mweisgut: In the current code base, your suggestion would also be possible since currently only the file based item runner implements behavior for `load_dedicated_expected_results`. But your suggested change would integrate a stronger restriction concerning the _dedicated_expected_results vector. This would then always have to have (max_item_id+1)-many elements.

mrks: I am not sure if you are describing a dense vector (holdings results for [0, 1, 3]) or a sparse vector (holding results for [0, 1, x, 3] with the third slot being empty). However, it should be one of these two, not a semi-sparse vector, where an item_id can either (a) have a dedicated result, (b) fall into the "x" slot, or (c) be bigger than the vector.

If this is unclear, maybe you could stop by. I'm having a hard time describing this.

mweisgut: Got it.

src/benchmarklib/abstract_benchmark_item_runner.cpp
mweisgut: I didn't get your point, yet. Could you explain the issue here more detailed? The _sqlite_wrapper is used in a similar way to the table newly integrated _expected_result_table: If the sqlite_wrapper is available, use the calculated expected output table and compare it with Hyrise's output table. Now, just don't do the calculation step via SQLite but use the provided expected result table for the comparison, if such a result table is provided.

mrks: https://github.com/hyrise/hyrise/blob/78788d952236ada05857f8e0b6a97628a70d005f/src/benchmarklib/tpcc/procedures/tpcc_new_order.cpp#L56-L65

Here, the SQL executor is reused. As such, there is more than one result table. Providing a single expected result breaks that.

src/benchmarklib/abstract_benchmark_item_runner.hpp
```diff
@@ -35,6 +35,9 @@ class AbstractBenchmarkItemRunner {
   // Returns the BenchmarkItemIDs of all selected items
   virtual const std::vector<BenchmarkItemID>& items() const = 0;
 
+  // Loads the dedicated expected restuls as tables into _dedicated_expected_results
+  virtual void load_dedicated_expected_results();
```
mrks: How does this differ between different implementation of the AbstractBenchmarkItemRunner, i.e., why is it virtual?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -71,6 +71,7 @@ BenchmarkRunner::BenchmarkRunner(const BenchmarkConfig& config,
     }
     std::cout << "- All tables loaded into SQLite (" << timer.lap_formatted() << ")" << std::endl;
     _benchmark_item_runner->set_sqlite_wrapper(sqlite_wrapper);
+    _benchmark_item_runner->load_dedicated_expected_results();
```
mrks: Why are setting and loading two different steps?

src/benchmarklib/benchmark_sql_executor.cpp
```diff
@@ -33,7 +35,10 @@ std::pair<SQLPipelineStatus, std::shared_ptr<const Table>> BenchmarkSQLExecutor:
 
   metrics.emplace_back(std::move(pipeline.metrics()));
 
-  if (_sqlite_wrapper) {
+  if (_expected_result_table) {
+    _compare_tables(_expected_result_table, result_table,
+                    "Using dedicated expected result table");
```
mrks: Probably need to re-run clang-format.

src/benchmarklib/abstract_benchmark_item_runner.cpp
mweisgut: Got it.

PR #1725, opened 2019-06-24
Show hash join's radix bits in PQP

Bouncner:
This PR prints the number of radix bits (used in the hash join operator) in the PQP visualization. Before, it has only been printed, when it was set with the creation of the hash join operator.

When somebody is not happy with public mutable members, I am happy to take suggestions.

image

### DIFF
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -140,14 +140,14 @@ std::shared_ptr<const Table> JoinHash::_on_execute() {
           !std::is_same_v<pmr_string, BuildColumnDataType> && !std::is_same_v<pmr_string, ProbeColumnDataType>;
 
       if constexpr (BOTH_ARE_STRING || NEITHER_IS_STRING) {
-        auto join_impl = JoinHashImpl<BuildColumnDataType, ProbeColumnDataType>(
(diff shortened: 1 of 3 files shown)


mrks:
On second thought: Is `_calculate_radix_bits` dependent on the types? If not, why is it part of the *Impl?

Bouncner:
Ouch ... almost embarassing. :)

Changed and simplified.

src/lib/operators/join_hash.cpp
```diff
@@ -64,6 +64,53 @@ std::shared_ptr<AbstractOperator> JoinHash::_on_deep_copy(
 
 void JoinHash::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeVariant>& parameters) {}
 
+template <typename T>
+size_t JoinHash::_calculate_radix_bits(const std::shared_ptr<const Table> build_table,
```
mrks: Const method?

Bouncner: Changed to public static with tests.

Bouncner: Sorry, forgot to WIP it.

But I think it's final now.

PR #1726, opened 2019-06-24
Index sensitive column pruning

mweisgut:
This PR refers to column pruning. If columns are pruned, the statistics about existing indices have to be adapted. Previously we could only get statistics about table's indices using tables themselves. This information is incorrect for pruned tables because the index ColumnIDs of the original table do not match the ColumnIDs for the pruned table. Therefore, this PR integrates the possibility to get index statistics from a StoredTableNode which consider the pruned columns. Additionally, the GetTable operator did not pass indices before. This is also addressed in this PR, so (single column) indices which are relevant after the column pruning are passed to the pruned table, i. e., the corresponding chunks contain these.


src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -89,6 +90,56 @@ bool StoredTableNode::is_column_nullable(const ColumnID column_id) const {
...
+
+  column_pruned_bitvector.resize(table->column_count());
+  column_left_shifts.resize(table->column_count());
+
+  // Fill the bitvector
+  for (const auto& pruned_column_id : _pruned_column_ids) {
+    column_pruned_bitvector[pruned_column_id] = true;
+  }
+
+  // Calculate left_shifts
```
mrks: (Not a full PR, just trying to get an overview of the open PRs)

This whole "adapt a vector to remove the pruned items" seems like something that is not needed only here. Can we make it more abstract?

mweisgut: It's rather "build a mapping structure which gives me the ColumnID of the pruned table as output for a provided original ColumnID as input" than "adapt a vector to remove the pruned items". I haven't found any similar easy to use procedure in the current code base yet. However, I agree with you that this procedure should be made reusable.

src/test/utils/pruning_column_id_mapping_test.cpp
```diff
@@ -0,0 +1,28 @@
...
+  auto expected_column_mapping_1 = std::vector<std::optional<ColumnID>>{std::nullopt, std::nullopt, std::nullopt,
+                                                                        ColumnID{0},  ColumnID{1},  ColumnID{2}};
+  auto expected_column_mapping_2 = std::vector<std::optional<ColumnID>>{ColumnID{0},  ColumnID{1}, std::nullopt,
+                                                                        std::nullopt, ColumnID{2}, ColumnID{3}};
+  auto expected_column_mapping_3 = std::vector<std::optional<ColumnID>>{ColumnID{0}, ColumnID{1}, ColumnID{2},
+                                                                        ColumnID{3}, ColumnID{4}, std::nullopt};
+  auto expected_column_mapping_4 = std::vector<std::optional<ColumnID>>{std::nullopt, ColumnID{0},  std::nullopt,
+                                                                        ColumnID{1},  std::nullopt, ColumnID{2}};
+
+  EXPECT_EQ(actual_column_mapping_1, expected_column_mapping_1);
```
mrks: This is hard to read. If line 22 fails, I have to jump to lines 8 and 13 to see what should happen. Test one thing, then the next.

src/lib/utils/pruning_column_id_mapping.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include "types.hpp"
+
+namespace opossum {
+
+const std::vector<std::optional<ColumnID>> column_ids_after_pruning(const size_t& original_table_column_count,
```
mrks: Unless there is something else in this file, it should have the same name.

src/lib/utils/pruning_column_id_mapping.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+  column_pruned_bitvector.resize(original_table_column_count);
+  column_id_mapping.resize(original_table_column_count);
+
+  // Fill the bitvector
+  for (const auto& pruned_column_id : pruned_column_ids) {
+    column_pruned_bitvector[pruned_column_id] = true;
+  }
+
+  // Calculate new colummn ids
```
mrks: typo

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -89,6 +91,39 @@ bool StoredTableNode::is_column_nullable(const ColumnID column_id) const {
...
+  if (_pruned_column_ids.empty()) {
+    return stored_indexes_statistics;
+  }
+
+  const auto column_id_mapping = column_ids_after_pruning(table->column_count(), _pruned_column_ids);
+
+  // update index statistics
+  for (auto stored_index_stats_iter = stored_indexes_statistics.begin();
+       stored_index_stats_iter != stored_indexes_statistics.end();) {
+    auto update_index_statistics = [&]() {
```
Bensk1: Why is this written as a lambda?

src/lib/operators/get_table.cpp
```diff
@@ -160,6 +160,7 @@ std::shared_ptr<const Table> GetTable::_on_execute() {
     } else {
       auto output_segments = Segments{stored_table->column_count() - _pruned_column_ids.size()};
       auto output_segments_iter = output_segments.begin();
+      auto output_indices = Indices{};
```
Bensk1: Above you use the plural `indexes`. Here it is `indices`. There might be further inconsistencies in the code base, but let's try to make it more consistent. When I heard native speakers talking about that topic, they always used `indexes`.

src/lib/storage/chunk.hpp
```diff
@@ -28,6 +28,7 @@ class BaseSegment;
 class BaseAttributeStatistics;
 
 using Segments = pmr_vector<std::shared_ptr<BaseSegment>>;
+using Indices = pmr_vector<std::shared_ptr<BaseIndex>>;
```
Bensk1: see above

src/lib/utils/column_ids_after_pruning.cpp
```diff
@@ -0,0 +1,33 @@
+#include "column_ids_after_pruning.hpp"
+
+#include <cstdlib>
+#include <optional>
+#include <vector>
+
+namespace opossum {
+
+const std::vector<std::optional<ColumnID>> column_ids_after_pruning(const size_t& original_table_column_count,
```
Bensk1: This should not be an issue, but I believe that we normally do not pass plain data types as `size_t` by reference. There is no overhead in passing it by value and it could only confuse the reader.

src/lib/utils/column_ids_after_pruning.cpp
```diff
@@ -0,0 +1,33 @@
...
+                                                                    const std::vector<ColumnID>& pruned_column_ids) {
+  std::vector<std::optional<ColumnID>> column_id_mapping;
+
+  auto column_pruned_bitvector = std::vector<bool>();
```
Bensk1: Is there any reason for not creating the vector as the one above? In addition, you could avoid the resizing by passing `original_table_column_count` to the constructor.

src/lib/operators/get_table.cpp
mrks: Make the linter complain about indices?

Bensk1: I believe for the plural of a loop/array/vector index, indices is fine. Maybe we should consult a native speaker.

src/lib/logical_query_plan/stored_table_node.cpp
mweisgut: Because I temporarily forgot about `break` and wanted to achieve an early exit.
I changed it to a nested loop.

src/lib/utils/column_ids_after_pruning.cpp
mweisgut: Good point, no specific reason.

Bensk1:
Was the last build really a "build everything without an error for 75 minutes and then crash while notifying because of network issues" thing?

mrzzzrm:
Had some thoughts about this that I didn't get a chance to write down until now:

`StoredTableNode::indexes_statistics() const` is only used in tests, right? Is there another PR coming that actually uses this? I don't think the StoredTableNode is a good place to put this function, as it extends the interface of a core class, and those should imo be as simple as possible. This is not to be taken lightly, imo. If two handful of specialized use cases all add their functions to the LQP data structures than the result is a complicated LQP interface for everyone writing optimizer rules.

Two further thoughts on this: 

* If this function is only to be used in one place in the upcoming PR, just put it in that place. Adding additional files to `utils/` for individual functions is something that admittedly happened before, but obviously doesn't scale well. 
* Why would retrieving index statistics be limited to StoredTableNodes, why is this not available for other node types? I.e., if `indexes_statistics` absolutely needs to be part of the LQP interface, why is not a member of AbstractLQPNode? If it has no place in AbstractLQPNode, than I'd take that as an indication that it isn't a good fit for StoredTableNode either

("Indexes" seems to be a valid plural, but I'd swear Hyrise uses "Indices" everywhere else)

mweisgut:
> Had some thoughts about this that I didn't get a chance to write down until now:
> 
> `StoredTableNode::indexes_statistics() const` is only used in tests, right? Is there another PR coming that actually uses this? I don't think the StoredTableNode is a good place to put this function, as it extends the interface of a core class, and those should imo be as simple as possible. This is not to be taken lightly, imo. If two handful of specialized use cases all add their functions to the LQP data structures than the result is a complicated LQP interface for everyone writing optimizer rules.
> 
> Two further thoughts on this:
> 
> * If this function is only to be used in one place in the upcoming PR, just put it in that place. Adding additional files to `utils/` for individual functions is something that admittedly happened before, but obviously doesn't scale well.
> * Why would retrieving index statistics be limited to StoredTableNodes, why is this not available for other node types? I.e., if `indexes_statistics` absolutely needs to be part of the LQP interface, why is not a member of AbstractLQPNode? If it has no place in AbstractLQPNode, than I'd take that as an indication that it isn't a good fit for StoredTableNode either
> 
> ("Indexes" seems to be a valid plural, but I'd swear Hyrise uses "Indices" everywhere else)

There is an upcoming PR which uses StoredTableNode::indexes_statistics(). It will be an optimizer rule which decides when and where to use an index join. I will come back to your mentioned points when I opened this specific PR at the latest.

Regarding to the utils/column_ids_after_pruning function: The mentioned optimizer rule will be the only place where StoredTableNode::indexes_statistics() is used, therefore column_ids_after_pruning is also used only in this context. The function was encapsulated for the purpose of reusability. I also think, that it should not be encapsulated as utility function when it's only used at one location. But the probability that this function will be encapsulated when it's needed at another location is pretty low IMO since it is not explicitly provided as reusable function and is therefore harder to find. @mrks do you have a strong opinion about that?

PR #1727, opened 2019-06-25
JoinIndex support index on both sides

mweisgut:
This PR integrates the possibility to use the JoinIndex with the mandatory index structures in the output table of the left input operator. Previously the index structures had to be on the right side.
Now, the JoinIndex constructor expects the specification of the join input operator which output table contains the index structures (`const JoinInputSide index_side`). By default, the index_side is `JoinInputSide::Right`.

### DIFF
--- a/src/lib/operators/join_index.cpp
+++ b/src/lib/operators/join_index.cpp
@@ -232,8 +232,8 @@ void JoinIndex::_append_matches(const ChunkID& probe_chunk_id, const ChunkOffset
 
 // join loop that joins two segments of two columns using an iterator for the probe side,
 // and an index for the index side
-template <typename LeftIterator>
-void JoinIndex::_join_two_segments_using_index(LeftIterator probe_iter, LeftIterator probe_end,
+template <typename ProbeIterator>
+void JoinIndex::_join_two_segments_using_index(ProbeIterator probe_iter, ProbeIterator probe_end,
                                                const ChunkID probe_chunk_id, const ChunkID index_chunk_id,
                                                const std::shared_ptr<BaseIndex>& index) {
   for (; probe_iter != probe_end; ++probe_iter) {
--- a/src/lib/operators/join_index.cpp
+++ b/src/lib/operators/join_index.cpp
@@ -21,7 +21,7 @@
 namespace opossum {
 
 /*
- * This is an index join implementation. It expects to find an index on the right column.
+ * This is an index join implementation. It expects to find an index on the index side column.
  * It can be used for all join modes except JoinMode::Cross.
  * For the remaining join types or if no index is found it falls back to a nested loop join.
  */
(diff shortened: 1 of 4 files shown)


src/lib/operators/join_index.cpp
```diff
@@ -55,8 +57,21 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
          "JoinHash doesn't support these parameters");
 
-  _right_matches.resize(input_table_right()->chunk_count());
-  _left_matches.resize(input_table_left()->chunk_count());
+  std::shared_ptr<const Table> left_input_table;
+  std::shared_ptr<const Table> right_input_table;
+  auto primary_predicate = _primary_predicate;
+
+  if (_index_side == JoinInputSide::Left) {
+    left_input_table = input_table_right();
```
mrks: Same as with the hash join (#1557) : Once we swap the tables, we should stop talking about "left" and "right" sides. Maybe "indexed" and "probed" side?

src/lib/operators/join_index.cpp
```diff
@@ -208,7 +238,13 @@ void JoinIndex::_join_two_segments_using_index(LeftIterator left_it, LeftIterato
     auto range_begin = BaseIndex::Iterator{};
     auto range_end = BaseIndex::Iterator{};
 
-    switch (_primary_predicate.predicate_condition) {
+    auto primary_predicate = _primary_predicate;
+
+    if (_index_side == JoinInputSide::Left) {
+      primary_predicate.flip();
```
mrks: The predicate should be flipped once and then stored. We shouldn't have flipped and non-flipped versions flying around.

src/lib/operators/join_index.cpp
```diff
@@ -55,45 +62,55 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
+    for (ChunkID chunk_id_left = ChunkID{0}; chunk_id_left < probe_input_table->chunk_count(); ++chunk_id_left) {
+      _probe_matches[chunk_id_left].resize(probe_input_table->get_chunk(chunk_id_left)->size());
     }
   }
 
-  if (track_right_matches) {
-    for (ChunkID chunk_id_right = ChunkID{0}; chunk_id_right < input_table_right()->chunk_count(); ++chunk_id_right) {
-      _right_matches[chunk_id_right].resize(input_table_right()->get_chunk(chunk_id_right)->size());
+  if (track_index_matches) {
+    for (ChunkID chunk_id_right = ChunkID{0}; chunk_id_right < index_input_table->chunk_count(); ++chunk_id_right) {
```
mrks: Still using "right" here. Please grep for left/right and make sure you catch them all.

src/lib/types.hpp
```diff
@@ -232,6 +232,8 @@ PredicateCondition inverse_predicate_condition(const PredicateCondition predicat
 //                      dropped. This behavior mirrors NOT EXISTS
 enum class JoinMode { Inner, Left, Right, FullOuter, Cross, Semi, AntiNullAsTrue, AntiNullAsFalse };
 
+enum class JoinInputSide { Left, Right };
```
mrks: Sorry I didn't bring this up before: aren't both sides inputs? Make it a class member of the join and name it IndexedSide?

PR #1728, opened 2019-06-27
Change Lambda captures to be compatible with C++20

julianmenzler:
With C++20 the implicit capture of `this` when the capture default is `=` [becomes deprecated](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0806r2.html). This PR adapts relevant Lambdas to be compatible with C++20.

* To capture `*this` by reference, explicit captures of `this` are added.
* `[=, this]` is an [extension of C++20](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0409r2.html) and not backward compatible with C++17. Therefore, some captures became more explicit.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -67,11 +67,11 @@ if (${ENABLE_COVERAGE})
 endif()
 
 # Global flags and include directories
-# When generating a dtrace header file, symbols containing dollar-signs are created This file needs to be compiled as well.
-# Hence, the `-Wno-dollar-in-identifier-extension` flag is required.
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_STANDARD 17)
 
+# When generating a dtrace header file, symbols containing dollar-signs are created This file needs to be compiled as well.
+# Hence, the `-Wno-dollar-in-identifier-extension` flag is required.
 add_compile_options(-pthread -Wall -Wextra -pedantic -Werror -Wno-unused-parameter -Wno-dollar-in-identifier-extension -Wno-unknown-pragmas)
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -70,7 +70,7 @@ endif()
 # When generating a dtrace header file, symbols containing dollar-signs are created This file needs to be compiled as well.
 # Hence, the `-Wno-dollar-in-identifier-extension` flag is required.
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD 17)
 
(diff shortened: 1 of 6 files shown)


src/CMakeLists.txt
```diff
@@ -69,7 +69,10 @@ endif()
 # Global flags and include directories
 # When generating a dtrace header file, symbols containing dollar-signs are created This file needs to be compiled as well.
 # Hence, the `-Wno-dollar-in-identifier-extension` flag is required.
-add_compile_options(-std=c++17 -pthread -Wall -Wextra -pedantic -Werror -Wno-unused-parameter -Wno-dollar-in-identifier-extension -Wno-unknown-pragmas)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
mrks: Especially with the new line below, the comment above `-Wno` does no longer match the code beneath it.

src/lib/operators/table_scan.cpp
```diff
@@ -94,7 +94,7 @@ std::shared_ptr<const Table> TableScan::_on_execute() {
   for (ChunkID chunk_id{0u}; chunk_id < in_table->chunk_count(); ++chunk_id) {
     if (excluded_chunk_set.count(chunk_id)) continue;
 
-    auto job_task = std::make_shared<JobTask>([=, &output_mutex, &output_chunks]() {
+    auto job_task = std::make_shared<JobTask>([this, in_table, chunk_id, &output_mutex, &output_chunks]() {
```
mrks: Why did you decide to copy the shared_ptr `in_table`?

src/CMakeLists.txt
julianmenzler: Ok. Is there a better place to put the `CXX_STANDARD` and `CXX_STANDARD_REQUIRED` declarations? At the top of the file or even in the root CMakeLists.txt file after line 2?
https://github.com/hyrise/hyrise/blob/6ce8213580fa87253ed628574f346c605e06a0b3/CMakeLists.txt#L1-L5

src/lib/operators/table_scan.cpp
julianmenzler: Before, `in_table` was passed by value via `=`.  Now, we could either 

* redeclare `const auto in_table = input_table_left();` or 
* use `input_table_left()` 

inside the Lambda. What do you suggest?

src/CMakeLists.txt
mrks: The place itself is not bad. You just need to move the last two lines of the comment to where they belong.

src/lib/operators/table_scan.cpp
mrks: Why can't we capture it by reference? That should have been done before, too, but now that we explicitly mention it, it looks like we gave it some thought.

src/CMakeLists.txt
julianmenzler: Ok 👍

src/lib/operators/table_scan.cpp
julianmenzler: Ok, I thought this was by done with intention. But since we have `CurrentScheduler::wait_for_tasks(jobs)` at the end of the function, it should also work with `&in_table`.

PR #1730, opened 2019-06-27
Remove function Table::chunks()

julianmenzler:
This PR has its root in #1686, which implements `std::atomic_load` and `std::atomic_store` for the vector `_chunks` in `Table`. Latter PR is in favor of the MvccDeletePlugin, which removes chunks from the `_chunks` vector.

Each read-/write-access on `_chunks` has to happen via atomics functions. People might forget about this and break thread-safety unwittingly.
Because of this, we remove the `chunks()` function from `Table`.  This limits all critical access to the `_chunks` vector to a single class (`Table`).

### DIFF
--- a/src/lib/operators/aggregate_sort.cpp
+++ b/src/lib/operators/aggregate_sort.cpp
@@ -400,7 +400,6 @@ std::shared_ptr<const Table> AggregateSort::_on_execute() {
     resolve_data_type(data_type, [&](auto type) {
       using ColumnDataType = typename decltype(type)::type;
       std::optional<ColumnDataType> previous_value;
-      ChunkID chunk_id_outer{0};
 
       /*
        * Initialize previous_value to the first value in the table, so we avoid considering it a value change.
@@ -417,12 +416,12 @@ std::shared_ptr<const Table> AggregateSort::_on_execute() {
       }
 
       // Iterate over all chunks and insert RowIDs when values change
-      for (auto chunk_id_inner = ChunkID{0}; chunk_id_inner < chunk_count; ++chunk_id_inner) {
-        const auto& segment = sorted_table->get_chunk(chunk_id_inner)->get_segment(column_id);
+      for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+        const auto& segment = sorted_table->get_chunk(chunk_id)->get_segment(column_id);
         segment_iterate<ColumnDataType>(*segment, [&](const auto& position) {
           if (previous_value.has_value() == position.is_null() ||
               (previous_value.has_value() && !position.is_null() && position.value() != *previous_value)) {
-            group_boundaries.insert(RowID{chunk_id_outer, position.chunk_offset()});
+            group_boundaries.insert(RowID{chunk_id, position.chunk_offset()});
             if (position.is_null()) {
               previous_value.reset();
             } else {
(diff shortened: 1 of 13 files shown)


src/lib/operators/aggregate_sort.cpp
```diff
@@ -392,21 +394,21 @@ std::shared_ptr<const Table> AggregateSort::_on_execute() {
...
   std::set<RowID> group_boundaries;
-  const auto& chunks = sorted_table->chunks();
+  const auto chunk_count = sorted_table->chunk_count();
   for (const auto& column_id : _groupby_column_ids) {
     auto data_type = input_table->column_data_type(column_id);
     resolve_data_type(data_type, [&](auto type) {
       using ColumnDataType = typename decltype(type)::type;
       std::optional<ColumnDataType> previous_value;
-      ChunkID chunk_id{0};
+      ChunkID chunk_id_outer{0};
```
mrks: What meaning does "outer/inner" have when it comes to an aggregate? Can we find a more descriptive name that does not reference the position in the code?

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -678,17 +678,21 @@ inline PosListsByChunk setup_pos_lists_by_chunk(const std::shared_ptr<const Tabl
   PosListsByChunk pos_lists_by_segment(input_table->column_count());
   auto pos_lists_by_segment_it = pos_lists_by_segment.begin();
 
-  const auto& input_chunks = input_table->chunks();
+  const auto input_chunks_count = input_table->chunk_count();
+  const auto input_columns_count = input_table->column_count();
 
   // For every column, for every chunk
-  for (ColumnID column_id{0}; column_id < input_table->column_count(); ++column_id) {
+  for (ColumnID column_id{0}; column_id < input_columns_count; ++column_id) {
```
mrks: Thanks for touching this, too :)

src/lib/operators/sort.cpp
```diff
@@ -202,8 +202,11 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
     auto materialization = std::make_shared<SortImplMaterializeOutput<SortColumnType>>(_table_in, _row_id_value_vector,
                                                                                        _output_chunk_size);
     auto output = materialization->execute();
-    for (auto& chunk : output->chunks()) {
-      chunk->set_ordered_by(std::make_pair(_column_id, _order_by_mode));
+
+    auto output_table = std::const_pointer_cast<Table>(output);  // cast away constness to call chunk->set_ordered_by
```
mrks: Argh. No, this shouldn't be necessary. **const cast is evil.** Is it necessary for `SortImplMaterializeOutput` to return a const?

src/lib/operators/aggregate_sort.cpp
julianmenzler: Ok, I went through the code: A single variable `chunk_id` is sufficient. I was quite irritated by this, but `chunk_id_outer` has no function. Therefore, the naming questing is solved.

src/lib/operators/sort.cpp
julianmenzler: Not really, it seems: I changed `SortImplMaterializeOutput` to return a non-const Table. Everything compiles and tests were fine.

PR #1737, opened 2019-07-02
Fix NUMA node binding

mrks:
Previously, `--scheduler` would take the physical configuration, ignoring everything that was set with `numactl -N x`. Now, the affinity is taken into account meaning that `numactl -N 1 -m 1 ./hyriseBenchmarkTPCH` will (a) correctly use only CPUs from node 1 and (b) spawn only 56 workers instead of 224.

### DIFF
--- a/src/lib/scheduler/topology.cpp
+++ b/src/lib/scheduler/topology.cpp
@@ -96,7 +96,7 @@ void Topology::_init_numa_topology(uint32_t max_num_cores) {
           }
           core_count++;
         }
-        if (!cpu_is_part_of_affinity) filtered_by_affinity = true;
+        if (!cpu_is_part_of_affinity) _filtered_by_affinity = true;
       }
 
       TopologyNode node(std::move(cpus));
@@ -194,8 +194,8 @@ void Topology::_create_memory_resources() {
 
 std::ostream& operator<<(std::ostream& stream, const Topology& topology) {
   stream << "Number of CPUs: " << topology.num_cpus() << std::endl;
-  if (topology.filtered_by_affinity()) {
-    stream << "Available CPUs / nodes filtered by externally set CPU affinity (e.g., numactl)." << std::endl;
+  if (topology._filtered_by_affinity) {
+    stream << "Available CPUs / nodes were filtered by externally set CPU affinity (e.g., numactl)." << std::endl;
   }
   for (size_t node_idx = 0; node_idx < topology.nodes().size(); ++node_idx) {
     stream << "Node #" << node_idx << " - ";
--- a/src/lib/scheduler/topology.cpp
+++ b/src/lib/scheduler/topology.cpp
@@ -96,6 +96,7 @@ void Topology::_init_numa_topology(uint32_t max_num_cores) {
           }
           core_count++;
         }
+        if (!cpu_is_part_of_affinity) filtered_by_affinity = true;
       }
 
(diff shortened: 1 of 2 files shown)


src/lib/scheduler/topology.cpp
```diff
@@ -71,23 +71,32 @@ void Topology::_init_numa_topology(uint32_t max_num_cores) {
...
+      numa_node_to_cpus(node_id, this_node_cpu_bitmask);
 
       for (CpuID cpu_id{0}; cpu_id < num_configured_cpus; ++cpu_id) {
-        if (numa_bitmask_isbitset(cpu_bitmask, cpu_id)) {
+        const auto cpu_is_part_of_node = numa_bitmask_isbitset(this_node_cpu_bitmask, cpu_id);
+        const auto cpu_is_part_of_affinity = numa_bitmask_isbitset(affinity_cpu_bitmask, cpu_id);
+        if (cpu_is_part_of_node && cpu_is_part_of_affinity) {
           if (max_num_cores == 0 || core_count < max_num_cores) {
             cpus.emplace_back(TopologyCpu(cpu_id));
             _num_cpus++;
```
Bouncner: `_num_cpus` does actually describe cores, right?

Bouncner: Ah, I just see that libnuma uses `cpu` for what I would call a core.
It's somewhat confusing that we talk about CPUs and cores and then increase `_num_cpus` if the core/cpu is used while `core_count` is always increased.

The check in line 84 ensure that we don't use too many cores, right? But `core_count` is increased no matter whether the cpu is actually used due to the affinity map.

src/lib/scheduler/topology.cpp
```diff
@@ -71,23 +71,32 @@ void Topology::_init_numa_topology(uint32_t max_num_cores) {
...
           }
           core_count++;
         }
+        if (!cpu_is_part_of_affinity) _filtered_by_affinity = true;
```
Bouncner: I don't get why we set the variable to true if the CPU is not part of the affinity map. From my (extremely limited) understanding, the value should be true when `affinity_cpu_bitmask` and `this_node_cpu_bitmask` differ.

src/lib/scheduler/topology.cpp
mrks: This is hard to do right. We like to talk about cores, libnuma calls them cpus, so does `lscpu`. The topology as at the boundary of the two. However, I don't expect this to really cause an issue in the future.

> The check in line 84 ensure that we don't use too many cores, right?

Correct

> But core_count is increased no matter whether the cpu is actually used due to the affinity map.

Incorrect.

src/lib/scheduler/topology.cpp
mrks: If no affinity is set, `affinity_cpu_bitmask` holds only 1s. If we find a value where it is 0, we know that the process had its affinity set from the outside.

`this_node_cpu_bitmask` is 1 only for those cores that are part of the node that we are looking at. As such, it is expected to be different from `affinity_cpu_bitmask`.

Bouncner: > If no affinity is set, `affinity_cpu_bitmask` holds only 1s. If we find a value where it is 0, we know that the process had its affinity set from the outside.

Add a comment?

PR #1485, opened 2019-02-08
Index join: change pos list size initially reserved

Bouncner:
The index join preallocated the result pos list, assuming the worst case output size of a join (n*m tuples). For large tables with hundreds of thousands of rows, the join easily allocated TBs.

This PR changes the size to reserve. It expects that each tuple of the smaller relation will have exactly one match. For very small tables, 100 RowIDs are reserved to avoid frequent resizes while not wasting memory.

Fixes #1390.

PR #1743, opened 2019-07-04
Using indexes for index joins on reference tables

mweisgut:
This PRs integrates the usage of indexes for index joins on reference tables. If the index side input operator of a JoinIndex is a reference table and its referenced original data table has indexes for the relevant join column, the available indexes are used for the calculation of the output table. In this case, the JoinIndex is executed chunk wise. Let `index_table` be the output table of the index side input operator and `probe_table` the output table of the probe side input operator.
For each ReferenceSegment `ref_seg` of the `index_table`'s join column which **(a) references a single chunk** and which **(b) referenced (data) chunk has an index for the relevant join column**, an index scan is executed for each value (`probe_val`) of the `probe_table`'s join column using the available index. Each index scan provides a position list which represents the matches of the referenced original data table for the corresponding scan with `probe_val` as search value. This position list is then intersected with the position list of `ref_seg`. The resulting position list represents the matches of the `index_table` for the corresponding scan with `probe_val` as search value.
If (a) and (b) are not given, a nested loop join is executed for the index side chunk as fallback solution.

**Edit**: This PR removes the support for the join setup "index side input table is a reference table & join mode is non-inner". Therefore, the `supports`-function of the joins were modified. We are now passing a join configuration to the supports function:

```c++
struct JoinConfiguration {
  JoinMode join_mode;
  PredicateCondition predicate_condition;
  DataType left_data_type;
  DataType right_data_type;
  bool secondary_predicates;
  // Only for JoinIndex
  std::optional<TableType> left_table_type{std::nullopt};
  std::optional<TableType> right_table_type{std::nullopt};
  std::optional<IndexSide> index_side{std::nullopt};
};
```

The supports function of the index join requires the left_table_type, right_table_type and index_side. All the other join operators do not. Since these parameters were not required before, the join test runner doesn't take these into account. Therefore, the test initialization for index joins will fail until the join test runner is adapted to these requirements. For now, the index join test initialization is just commented out. The adaptation of the runner has to be addressed during the processing of Issue #1755.


Bensk1:
Some high-level questions / quick braindump: should we integrate that in @mrzzzrm's join testing framework? We should also think of an elegant way to tests this end-to-end with our SQL tests.

mweisgut:
Currently, JoinIndex is actually integrated into the join testing framework, but the configuration for index joins is not sufficient yet. We should adapt this.

Bensk1:
Do we have an issue/task for that?

src/lib/operators/join_index.cpp
```diff
@@ -83,133 +83,130 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
-  }
-
-  _probe_pos_list->shrink_to_fit();
-  _index_pos_list->shrink_to_fit();
-
-  // Write PosLists for Semi/Anti Joins, which so far haven't written any results to the PosLists
-  // We use `probe_matches_by_chunk` to determine whether a tuple from the probe side found a match.
-  if (is_semi_or_anti_join) {
-    const auto invert = _mode == JoinMode::AntiNullAsFalse || _mode == JoinMode::AntiNullAsTrue;
+  } else {  // INDEX DATA JOIN
```
Bensk1: Is this branch also executed if the table type is reference but the join type is not inner? Is this a problem?

src/lib/operators/join_index.cpp
```diff
@@ -83,133 +83,130 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
+        // as we do not want to spend time on evaluating the best index inside of this join loop
+        index = indexes.front();
+      }
 
-    for (auto chunk_id = ChunkID{0}; chunk_id < probe_input_table->chunk_count(); ++chunk_id) {
-      const auto chunk_size = probe_input_table->get_chunk(chunk_id)->size();
-      for (auto chunk_offset = ChunkOffset{0}; chunk_offset < chunk_size; ++chunk_offset) {
-        if (_probe_matches[chunk_id][chunk_offset] ^ invert) {
-          _probe_pos_list->emplace_back(chunk_id, chunk_offset);
+      // Scan all chunks from the pruning side input
```
Bensk1: pruning again?

src/lib/operators/join_index.cpp
```diff
@@ -83,133 +83,130 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
+      if (index) {
```
Bensk1: see above

src/lib/operators/join_index.cpp
```diff
@@ -223,71 +220,132 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
-                                               const ChunkID probe_chunk_id, const ChunkID index_chunk_id,
-                                               const std::shared_ptr<BaseIndex>& index) {
+void JoinIndex::_data_join_two_segments_using_index(ProbeIterator probe_iter, ProbeIterator probe_end,
+                                                    const ChunkID probe_chunk_id, const ChunkID index_chunk_id,
+                                                    const std::shared_ptr<BaseIndex>& index) {
+  for (; probe_iter != probe_end; ++probe_iter) {
+    const auto probe_side_position = *probe_iter;
+    const auto index_ranges = _index_ranges_for_value(probe_side_position, index);
+    for (const auto& index_range : index_ranges) {
+      _append_matches(std::get<0>(index_range), std::get<1>(index_range), probe_side_position.chunk_offset(),
```
Bensk1: We should be able to use structured bindings here, right?

src/lib/operators/join_index.cpp
```diff
@@ -223,71 +220,132 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
+    const auto index_ranges = _index_ranges_for_value(probe_side_position, index);
+    for (const auto& index_range : index_ranges) {
+      std::transform(std::get<0>(index_range), std::get<1>(index_range), std::back_inserter(index_scan_pos_list),
+                     [index_chunk_id](ChunkOffset index_chunk_offset) {
+                       return RowID{index_chunk_id, index_chunk_offset};
+                     });
+    }
+
+    PosList sortable_ref_seg_pos_list;
+    sortable_ref_seg_pos_list.insert(sortable_ref_seg_pos_list.end(), reference_segment_pos_list->begin(),
```
Bensk1: Can't we just copy it or sort the pos list itself?

src/lib/operators/join_index.cpp
```diff
@@ -319,6 +377,59 @@ void JoinIndex::_append_matches(const BaseIndex::Iterator& range_begin, const Ba
...
+        }
+      }
+    }
+  }
+
+  _probe_pos_list->shrink_to_fit();
+  _index_pos_list->shrink_to_fit();
+
+  // Write PosLists for Semi/Anti Joins, which so far haven't written any results to the PosLists
+  // We use `probe_matches_by_chunk` to determine whether a tuple from the probe side found a match.
```
Bensk1: Is `probe_matches_by_chunk` used anywhere?

src/lib/operators/join_index.cpp
```diff
@@ -335,18 +446,23 @@ void JoinIndex::_write_output_segments(Segments& output_segments, const std::sha
             input_table->get_chunk(ChunkID{0})->get_segment(column_id));
 
         // de-reference to the correct RowID so the output can be used in a Multi Join
-        for (const auto& row : *pos_list) {
+        for (auto pos_list_offset = ChunkOffset{0}; pos_list_offset < pos_list->size(); ++pos_list_offset) {
+          const auto& row = (*pos_list)[pos_list_offset];
```
Bensk1: Dereferencing `pos_list` outside of the loop might save some performance.

mweisgut:
> Do we have an issue/task for that?

I am not yet very familiar with the join testing framework and would take a closer look at it to identify what exactly we need.

src/lib/operators/join_index.cpp
```diff
@@ -83,133 +83,130 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
+      _index_matches[index_chunk_id].resize(_index_input_table->get_chunk(index_chunk_id)->size());
     }
   }
 
   _probe_pos_list = std::make_shared<PosList>();
   _index_pos_list = std::make_shared<PosList>();
 
   auto pos_list_size_to_reserve =
-      std::max(uint64_t{100}, std::min(probe_input_table->row_count(), index_input_table->row_count()));
+      std::max(uint64_t{100}, std::min(_probe_input_table->row_count(), _index_input_table->row_count()));
```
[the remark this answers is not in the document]
mweisgut: #1485 says the following: "For very small tables, 100 RowIDs are reserved to avoid frequent resizes while not wasting memory."

src/lib/operators/join_index.cpp
```diff
@@ -83,133 +83,130 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
-        if (!_probe_matches[probe_chunk_id][chunk_offset]) {
-          _probe_pos_list->emplace_back(RowID{probe_chunk_id, chunk_offset});
-          _index_pos_list->emplace_back(NULL_ROW_ID);
+      const auto& reference_segment =
+          std::dynamic_pointer_cast<ReferenceSegment>(index_chunk->segments()[_primary_predicate.column_ids.second]);
+      Assert(reference_segment != nullptr,
+             "Non-empty index intput table (reference table) has to have only reference segments.");
+      auto index_data_table = reference_segment->referenced_table();
+      const std::vector<ColumnID> index_data_table_column_ids{reference_segment->referenced_column_id()};
+      Assert(index_data_table != nullptr, "ReferenceSegment has no reference table.");
```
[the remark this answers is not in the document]
mweisgut: The existence of the ref_table is ensured in the RefSeg constructor, so we can get rid of this line here.

src/lib/operators/join_index.cpp
```diff
@@ -83,133 +83,130 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
-  }
 
-  // For Full Outer and Right Join we need to add all unmatched rows for the index side.
-  if (_mode == JoinMode::FullOuter || _mode == JoinMode::Right) {
-    for (ChunkID chunk_id{0}; chunk_id < _index_matches.size(); ++chunk_id) {
-      for (ChunkOffset chunk_offset{0}; chunk_offset < _index_matches[chunk_id].size(); ++chunk_offset) {
-        if (!_index_matches[chunk_id][chunk_offset]) {
-          _index_pos_list->emplace_back(RowID{chunk_id, chunk_offset});
-          _probe_pos_list->emplace_back(NULL_ROW_ID);
+        // Scan all chunks from the pruning side input
```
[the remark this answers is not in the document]
mweisgut: Should be probe side.

src/lib/operators/join_index.cpp
mweisgut: That's the way it is supposed to be, so it's not a problem. The comment here may be misleading. This branch is for index data joins and for outer index reference joins.

src/lib/operators/join_index.cpp
mweisgut: I didn't know structured bindings in c++ before. Thanks for mentioning that. Yes, we can use them here.

src/lib/operators/join_index.cpp
mweisgut: _probe_matches

src/lib/operators/join_index.cpp
```diff
@@ -83,133 +83,130 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
+          std::dynamic_pointer_cast<ReferenceSegment>(index_chunk->segments()[_primary_predicate.column_ids.second]);
+      Assert(reference_segment != nullptr,
+             "Non-empty index intput table (reference table) has to have only reference segments.");
+      auto index_data_table = reference_segment->referenced_table();
+      const std::vector<ColumnID> index_data_table_column_ids{reference_segment->referenced_column_id()};
+      Assert(index_data_table != nullptr, "ReferenceSegment has no reference table.");
+
+      const auto& reference_segment_pos_list = reference_segment->pos_list();
+      if (reference_segment_pos_list->references_single_chunk()) {
+        if (reference_segment_pos_list->empty()) {
```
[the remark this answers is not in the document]
mweisgut: Sure.

src/lib/operators/join_index.cpp
mweisgut: We cannot sort the pos list itself since it is const. Additionally, the copy constructor of PosList is implicitly-deleted.
To avoid creating and filling an additional vector for performance purposes, we could use std::const_pointer_cast. But this isn't an elegant solution, either.
```c++
const auto& mutable_ref_seg_pos_list = std::const_pointer_cast<PosList>(reference_segment_pos_list);
    std::sort(mutable_ref_seg_pos_list->begin(), mutable_ref_seg_pos_list->end());
```

src/lib/operators/join_index.cpp
mweisgut: I changed this in ebf28ddb0671b708f6e605b44e8cbee97f61b4b1 but [this experiment](http://quick-bench.com/wL4aoNRhIBiiOClfd7Jqwofir14) leads me to the assumption that this won't improve the performance.

src/lib/operators/join_index.cpp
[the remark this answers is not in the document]
Bensk1: Perfect, thanks.

src/lib/operators/join_index.cpp
Bensk1: @mrks should we go with the current solution?

src/lib/operators/join_index.cpp
Bensk1: Thanks for setting up the experiment. Apparently, the compiler can optimize that. For GCC it actually is 10% faster. Markus pointed out to me that it the current solution in ebf28dd does not make any difference. Could you just use the previous version, dereferencing the PosList in the loop? Sorry!

src/lib/operators/join_index.cpp
mrks: Unless there is a significant performance reason to avoid copies, we should avoid modifying pos lists as they might be referenced by more than a single ReferenceSegment.

The PosList is non-copyable so that people don't accidentally do stupid stuff with it. I'd be more than fine with adding a `copy()` method.

mrks: (Implicit copies in C++ are... not so nice anyway)

https://pbs.twimg.com/media/Bhgkwl2IQAAQ5YG.jpg

src/lib/storage/pos_list.hpp
```diff
@@ -53,6 +53,14 @@ struct PosList final : private pmr_vector<RowID> {
 
   PosList& operator=(PosList&& other) = default;
 
+  PosList copy() const {
+    PosList copy(this->size());
+    for (uint32_t idx = 0; idx < this->size(); ++idx) {
```
Bensk1: Hm, could it make sense to use something like `std::copy` hear?

mweisgut: I just executed a quick [benchmark code](https://gist.github.com/mweisgut/108ac25a594eaeb9fdf77955c2939e9e) and got [these results](https://gist.github.com/mweisgut/53bf004c7130d68f368745e0eee9d8c5) (executed on pella, release build, using clang-7).
Therefore I would say that the access [] variant is slightly faster than the std::copy variant. However, the std::copy variant would be more compact.

mweisgut: Well, since the mentioned results show marginal differences between the access [] and std::copy variants and there are also iterations in which the std::copy variant is faster, we should probably use the standard lib.

Bensk1: Thank you looking into it that deeply

src/test/operators/join_index_test.cpp
```diff
@@ -102,14 +104,28 @@ class JoinIndexTest : public BaseTest {
...
+    } else {
+      index_side_input = right;
+    }
+
     const auto& performance_data = static_cast<const JoinIndex::PerformanceData&>(join->performance_data());
-    if (using_index && right->get_output()->type() == TableType::Data) {
-      // We can't execute the index join on referencing tables
-      EXPECT_EQ(performance_data.chunks_scanned_with_index, static_cast<size_t>(right->get_output()->chunk_count()));
+    // for index reference joins, only the join JoinMode::Inner is supported. Additionally, if the join segments of the
+    // reference table don't provde the guarantee of referencing one single chunk (of the original data table), then the
```
mrks: (1) typo
(2) Instead of doc'ing in the test, I'd put it in the hpp. Especially the limitation to inner joins 
(3) Is there a valid case in which we would only realize at runtime that the scan is 100% fallback? If not, there should at least be a performance warning.
(4) On a similar note, if we end up with a non-inner join on a reference input, didn't someone else before screw up? Supporting this makes the code more complicated and introduces additional potential for errors.

src/test/operators/join_index_test.cpp
```diff
@@ -443,7 +459,27 @@ TYPED_TEST(JoinIndexTest, JoinOnDictAndReferenceSegment) {
...
+  scan_b->execute();
+  auto scan_c = this->create_table_scan(this->_table_wrapper_h, ColumnID{0}, PredicateCondition::GreaterThanEquals, 0);
+  scan_c->execute();
+
+  auto join = std::make_shared<JoinIndex>(
+      scan_a, scan_b, JoinMode::Inner, OperatorJoinPredicate{{ColumnID{0}, ColumnID{0}}, PredicateCondition::Equals});
+  join->execute();
+
+  // Referencing single chunk guarantee is not given since the left input of the index join is also an index join
+  // and the IndexSide is left. The execution of the join index does not provide single chunk reference guarantee.
```
mrks: > The execution of the join index

Nitpicky: Do you mean the index join operator? We put the "join" first so that all joins are grouped. When talking about them, please either use, e.g., "JoinHash operator" or "the hash join", but not "the join hash".

src/lib/operators/join_index.cpp
```diff
@@ -319,6 +367,59 @@ void JoinIndex::_append_matches(const BaseIndex::Iterator& range_begin, const Ba
...
+    }
+  }
+
+  _probe_pos_list->shrink_to_fit();
+  _index_pos_list->shrink_to_fit();
+
+  // Write PosLists for Semi/Anti Joins, which so far haven't written any results to the PosLists
+  // We use `_probe_matches` to determine whether a tuple from the probe side found a match.
+  if (is_semi_or_anti_join) {
+    const auto invert = _mode == JoinMode::AntiNullAsFalse || _mode == JoinMode::AntiNullAsTrue;
```
mrks: Are the NULL modes handled/tested anywhere? I don't think it changed with this PR, so if they are not, feel free to make it a separate issue.

src/lib/operators/join_index.cpp
```diff
@@ -319,6 +367,59 @@ void JoinIndex::_append_matches(const BaseIndex::Iterator& range_begin, const Ba
   }
 }
 
+void JoinIndex::_append_matches_dereferenced(const ChunkID& probe_chunk_id, const ChunkOffset& probe_chunk_offset,
+                                             const PosList& index_table_matches) {
+  for (const auto& index_side_row_id : index_table_matches) {
+    _probe_pos_list->emplace_back(RowID{probe_chunk_id, probe_chunk_offset});
+    _index_pos_list->emplace_back((index_side_row_id));
```
mrks: Why `(())`?

src/lib/operators/join_index.cpp
```diff
@@ -319,6 +367,59 @@ void JoinIndex::_append_matches(const BaseIndex::Iterator& range_begin, const Ba
...
+    _index_pos_dereferenced.emplace_back(true);
+  }
+}
+
+void JoinIndex::_append_matches_non_inner(const bool is_semi_or_anti_join) {
+  // For Full Outer and Left Join we need to add all unmatched rows for the probe side
+  if (_mode == JoinMode::Left || _mode == JoinMode::FullOuter) {
+    for (ChunkID probe_chunk_id = ChunkID{0}; probe_chunk_id < _probe_input_table->chunk_count(); ++probe_chunk_id) {
+      for (ChunkOffset chunk_offset{0}; chunk_offset < _probe_matches[probe_chunk_id].size(); ++chunk_offset) {
```
mrks: No need to repeat the data type: `auto index_chunk_id = ChunkID{0}` (couple of times in this PR). Also, at least within one file, I'd try to keep the initialization style consistent.

src/lib/operators/join_index.cpp
```diff
@@ -83,133 +83,124 @@ std::shared_ptr<const Table> JoinIndex::_on_execute() {
...
+        // as we do not want to spend time on evaluating the best index inside of this join loop
+        const auto& index = indexes.front();
+
+        // Scan all chunks from the probe side input
+        for (ChunkID probe_chunk_id = ChunkID{0}; probe_chunk_id < _probe_input_table->chunk_count();
+             ++probe_chunk_id) {
+          const auto& probe_segment =
+              _probe_input_table->get_chunk(probe_chunk_id)->get_segment(_adjusted_primary_predicate.column_ids.first);
+          segment_with_iterators(*probe_segment, [&](auto probe_iter, const auto probe_end) {
+            _data_join_two_segments_using_index(probe_iter, probe_end, probe_chunk_id, index_chunk_id, index);
```
mrks: I don't get this part. The `else` block handles "DATA JOIN or OUTER REFERENCE JOIN", but here you unconditionally call `_data_join_two_segments_using_index`.

src/lib/operators/join_index.cpp
```diff
@@ -319,6 +367,59 @@ void JoinIndex::_append_matches(const BaseIndex::Iterator& range_begin, const Ba
...
+}
+
+void JoinIndex::_append_matches_non_inner(const bool is_semi_or_anti_join) {
+  // For Full Outer and Left Join we need to add all unmatched rows for the probe side
+  if (_mode == JoinMode::Left || _mode == JoinMode::FullOuter) {
+    for (ChunkID probe_chunk_id = ChunkID{0}; probe_chunk_id < _probe_input_table->chunk_count(); ++probe_chunk_id) {
+      for (ChunkOffset chunk_offset{0}; chunk_offset < _probe_matches[probe_chunk_id].size(); ++chunk_offset) {
+        if (!_probe_matches[probe_chunk_id][chunk_offset]) {
+          _probe_pos_list->emplace_back(RowID{probe_chunk_id, chunk_offset});
+          _index_pos_list->emplace_back(NULL_ROW_ID);
```
mrks: Aren't you violating the `_probe_pos_list->size()` == `_index_pos_list->size()` == `_index_pos_dereferenced.size()` invariant here?

mweisgut:
> Do we have an issue/task for that?

#1755

src/test/operators/join_index_test.cpp
mweisgut: (4) : since any outer index join on a reference input is effectively a nested loop join, we should not support this setup here, right.

mweisgut: Removing the setup "index side table is a ref. table and the join mode is non-inner" would require the static support function to be extended since we don't pass table data information.

Is there anything against this extension?

Current support function:
```c++
static bool supports(JoinMode join_mode, PredicateCondition predicate_condition, DataType left_data_type,
                       DataType right_data_type, bool secondary_predicates);
```

src/lib/operators/join_index.cpp
mweisgut: All join modes are tested by using the join test runner. Setting `invert = false;` and executing the test suite again resulted in 52 failed tests. In these tests either AntiNullAsFalse or AntiNullAsTrue was used.

src/lib/operators/join_index.cpp
mweisgut: Unconditionally is not exactly true: We only do the `_data_join_two_seg...` if `!indexes.empty()`. This is only the case for chunks which contain data segments. Therefore, a chunk which contains only ref. segments cannot contain indexes because of the nature of create_index.

Please correct me if I'm wrong.

Effectively, if this branch is entered and the index_input_table is a reference table, the nested loop fallback solution is executed. This will be changes if we remove the ref. table & non-inner join support.

src/lib/operators/join_index.cpp
Bouncner: Cannot see the discussion on a mobile. Do we prefer this style now?
I just removed a couple of these a few weeks ago because I think they are (especially in loops) not that easy to read.

mrks: I wouldn't go as far as to make it a rule, but if two consecutive for loops have different initializations, it forces people (well, at least me) to spend more time reading it than necessary.

src/test/operators/join_index_test.cpp
mrks: Feel free to pass `TableType`

src/lib/operators/join_index.cpp
mrks: Maybe add `// DATA JOIN` after checking for `!indexes.empty()`?

mweisgut: Even this wouldn't be exactly true: it is also a data join if the fallback solution is used. There are currently two ways of creating indexes: On the table level (add indexes to each chunk) or on the chunk level. If the latter is used, there could be chunks in a data table without an index for a specific column, even if the previous chunk had one.

mweisgut: [comment change](https://github.com/hyrise/hyrise/pull/1743/commits/72ffd8693d9a346fe84ff5ff61a1a5fd10ac73a4)

src/test/operators/join_test_runner.cpp
```diff
@@ -591,7 +591,7 @@ TEST_P(JoinTestRunner, TestJoin) {
 INSTANTIATE_TEST_CASE_P(JoinNestedLoop, JoinTestRunner, testing::ValuesIn(JoinTestRunner::create_configurations<JoinNestedLoop>()), );  // NOLINT
 INSTANTIATE_TEST_CASE_P(JoinHash, JoinTestRunner, testing::ValuesIn(JoinTestRunner::create_configurations<JoinHash>()), );  // NOLINT
 INSTANTIATE_TEST_CASE_P(JoinSortMerge, JoinTestRunner, testing::ValuesIn(JoinTestRunner::create_configurations<JoinSortMerge>()), );  // NOLINT
-INSTANTIATE_TEST_CASE_P(JoinIndex, JoinTestRunner, testing::ValuesIn(JoinTestRunner::create_configurations<JoinIndex>()), );  // NOLINT
+// INSTANTIATE_TEST_CASE_P(JoinIndex, JoinTestRunner, testing::ValuesIn(JoinTestRunner::create_configurations<JoinIndex>()), );  // NOLINT
```
mweisgut: I just want to highlight this.

PR #1745, opened 2019-07-09
TableScan: Use type-specific implementations in more cases

### LINKED ISSUE #1712: Cast for non-equi scans so that fast scan implementation is used
TPC-H 6 uses the slow ExpressionEvaluator for a `float between double and double` predicate, even though these doubles could be casted to the previous/next float without breaking correctness of the query if BETWEEN INCLUSIVE also gets rewritten to BETWEEN EXCLUSIVE

image

mrks:
fixes #1712

Previously, `l_discount BETWEEN .05 AND .07` in TPC-H query 6 was executed by the ExpressionEvaluator-based scan. This was because the sql-parser supplied .05 and .07 as doubles, while `l_discount` is a float. As .05 cannot be losslessly casted into a float, we did not use the ColumnBetweenTableScanImpl.

With this PR,  even if lossless_cast is not successful, we might be able to do a cast if we know the predicate that is used. For example, `float(a) > double(3.1)` can be reformulated as `float(a) >= float(3.099999904632568359375)`.

This is admittedly ugly and I wish we wouldn't need those completely unproductive lines just to handle ugly floats. :(

```
| TPC-H 1        | 0.606061816216 | 37   | 0.615257799625 | 37   | +2%        |                          0.0354 |
| TPC-H 2        | 7.43865060806  | 447  | 7.47933769226  | 449  | +1%        |                          0.0022 |
| TPC-H 3        | 6.50099563599  | 391  | 6.34496641159  | 381  | -2%        |                          0.0009 |
| TPC-H 4        | 2.19818210602  | 132  | 2.24267363548  | 135  | +2%        |                          0.0000 |
| TPC-H 5        | 3.89554905891  | 234  | 4.27952289581  | 257  | +10%       |                          0.0000 |
| TPC-H 6        | 21.6650161743  | 1300 | 42.5728569031  | 2555 | +97%       |                          0.0000 |
| TPC-H 7        | 1.62497806549  | 98   | 1.62532103062  | 98   | +0%        |                          0.9548 |
| TPC-H 8        | 5.73870420456  | 345  | 5.65294408798  | 340  | -1%        |                          0.0000 |
| TPC-H 9        | 1.64683425426  | 99   | 1.63609814644  | 99   | -1%        |                          0.0283 |
| TPC-H 10       | 3.01311159134  | 181  | 2.98998713493  | 180  | -1%        |                          0.0015 |
| TPC-H 11       | 24.0803031921  | 1445 | 24.4193706512  | 1466 | +1%        |                          0.0000 |
| TPC-H 12       | 7.3393073082   | 441  | 7.39103507996  | 444  | +1%        |                          0.0000 |
| TPC-H 13       | 2.29398775101  | 138  | 2.2624399662   | 136  | -1%        |                          0.0000 |
| TPC-H 14       | 23.727306366   | 1424 | 23.703950882   | 1423 | -0%        |                          0.2280 |
| TPC-H 15       | 14.7273988724  | 884  | 14.6538639069  | 880  | -0%        |                          0.0000 |
| TPC-H 16       | 6.91516304016  | 415  | 6.86940860748  | 413  | -1%        |                          0.0000 |
| TPC-H 17       | 1.44273531437  | 87   | 1.38560473919  | 84   | -4%        |                          0.0000 |
| TPC-H 18       | 1.29387676716  | 78   | 1.29598855972  | 78   | +0%        |                          0.7365 |
| TPC-H 19       | 5.56364631653  | 334  | 5.58018445969  | 335  | +0%        |                          0.0000 |
| TPC-H 20       | 2.46982288361  | 149  | 2.42373919487  | 146  | -2%        |                          0.0000 |
| TPC-H 21       | 0.729427814484 | 44   | 0.731076180935 | 44   | +0%        |                          0.6634 |
| TPC-H 22       | 12.7369623184  | 765  | 13.0812969208  | 785  | +3%        |                          0.0000 |
| geometric mean |                |      |                |      | +3%        |                                 |
```

TPC-H 5 was unaffected and is an outlier.


src/lib/utils/lossless_predicate_cast.hpp
```diff
@@ -0,0 +1,55 @@
+#include "lossless_cast.hpp"
+#include "types.hpp"
+
+namespace opossum {
+std::optional<float> next_float_towards(double value, double towards);
+
+// Even if lossless_cast is not successful, we might be able to do a cast if we know the predicate that is used.
+// As 3.1 has different float representations as a double and as a float (and neither is exact), lossless_cast will
+// fail. However, `float(a) > double(3.1)` can be reformulated as `float(a) >= float(3.099999904632568359375)`.
```
d-justen: It's supposed to be reformulated as the next-higher float, right?

mrks: Good catch.

src/lib/utils/lossless_predicate_cast.cpp
```diff
@@ -0,0 +1,63 @@
...
+    // this case. This is only for stupid queries like `WHERE x = NULL`, anyway.
+    return std::nullopt;
+  }
+
+  // Safe casting between NULL and non-NULL type is not possible. (Cannot be handled below as resolve_data_type()
+  // doesn't resolve NULL)
+  if ((source_data_type == DataType::Null) != (target_data_type == DataType::Null)) {
+    return std::nullopt;
+  }
+
```
d-justen: This could also be one if-clause with `source_data_type == DataType::Null || target_data_type == DataType::Null`

mrks: Do you mean combining this with the clause above? That might logically be the case, but my idea was to keep the two cases (that came from lossless_predicate) separate. I hoped this would make them easier to understand - what do you think?

Bensk1: I was about to submit that when you changed the code. I liked your previous version

src/lib/types.hpp
```diff
@@ -213,6 +213,9 @@ enum class PredicateCondition {
 // @return whether the PredicateCondition takes exactly two arguments
 bool is_binary_predicate_condition(const PredicateCondition predicate_condition);
 
+// @return whether the PredicateCondition takes exactly two arguments and is not one of LIKE or IN
```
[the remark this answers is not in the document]
mrks: Do you mean `x = NULL`? That's binary. `x IS NULL` is not. Not sure what needs to be clarified here.

src/lib/utils/lossless_predicate_cast.cpp
mrks: ... now what? :D 
Can someone come up with a compromise?

src/lib/types.hpp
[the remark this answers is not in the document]
Bensk1: Everything alright

src/lib/utils/lossless_predicate_cast.cpp
mrks: After discussing this with @Bensk1, we have agreed to go back to the original version. One reason being the similarity to lossless_cast, the other one being that these are two distinct cases. Nonetheless, I appreciate the feedback!

Issue #1700, opened 2019-06-04
Ensure that chunks added after LQP translation are handled


Bensk1:
In #1697 @mrks mentioned that we should add a test which ensures that chunks added after LQP translation of index scans are taken into account.

mrks:
@Bensk1 @mweisgut Can you explain why the LQPTranslator only cares about GroupKeyIndexes?

https://github.com/hyrise/hyrise/blob/e71169a1e18f73b03ac9054af0f5e3e1d04cf944/src/lib/logical_query_plan/lqp_translator.cpp#L208-L215

There should be a comment on that. Feel free to edit the upcoming PR to include it.

Bensk1:
I think that originates from a time where we only had the GroupKey index or where it was the only one properly and efficiently working.

mweisgut:
It was introduced with #661.

mrks:
Open question is part of #1790.

PR #1747, opened 2019-07-10
Ensure that chunks added after LQP translation are handled 

mrks:
refs #1700, but there is another question open over there.


src/lib/operators/index_scan.hpp
```diff
@@ -29,12 +29,8 @@ class IndexScan : public AbstractReadOnlyOperator {
 
   const std::string name() const final;
 
-  /**
```
mweisgut: random line.

https://github.com/hyrise/hyrise/blob/e03caf05b0cc48a044ef04593411553ef0013d17/src/lib/operators/index_scan.hpp#L23

This isn't necessary any longer, isn't it?

mweisgut:
The next few days will be very crowded. I probably won't get to a detailed review before Thursday.

PR #1752, opened 2019-07-17
Use UnionAll for combining Index- and TableScan results

### LINKED ISSUE #1711: LQPTranslator uses wrong Union for combining the results of Index- and TableScan
The LQPTranslator uses the expensive UnionPositions which is meant for other use cases. Instead, the cheapest possible unioning method (UnionAll) should be used.

d-justen:
fixes  #1711

### DIFF
--- a/src/lib/optimizer/join_ordering/join_graph_builder.cpp
+++ b/src/lib/optimizer/join_ordering/join_graph_builder.cpp
@@ -83,6 +83,7 @@ void JoinGraphBuilder::_traverse(const std::shared_ptr<AbstractLQPNode>& node) {
        */
 
       const auto union_node = std::static_pointer_cast<UnionNode>(node);
+
       if (union_node->union_mode == UnionMode::Positions) {
         const auto parse_result = _parse_union(union_node);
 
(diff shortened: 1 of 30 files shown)


src/lib/optimizer/join_ordering/join_graph_builder.cpp
```diff
@@ -83,15 +83,10 @@ void JoinGraphBuilder::_traverse(const std::shared_ptr<AbstractLQPNode>& node) {
        */
 
       const auto union_node = std::static_pointer_cast<UnionNode>(node);
+      const auto parse_result = _parse_union(union_node);
```
mrks: Could you please explain the changes in this file?

mrks: (Not in the code, but here)

d-justen: The original thought was that the distinction of UnionModes didn't make sense in the first place since there was only one. I basically just wanted to see if something bad happens if the UnionModes are treated the same there and then forgot about it.
Anyway, I don't really get yet why we should only parse and traverse over a UnionPositions, but not UnionAll.

mrks: Is this covered by any test at all?

d-justen: Actually, the UnionAll doesn't appear in any test (except for UnionAllTest of course)

mrks: Did I understand you right that you wanted to look into that? Otherwise, we should open an issue.

Bouncner:
@bensk1: does that have a measurable impact on your work together with Marcel?

Bensk1:
We were not working with partially indexed tables yet.

PR #1753, opened 2019-07-17
View columns can have only single names

### LINKED ISSUE #1748: Discuss whether columns in views can have multiple names
@janetzki and I discussed this part:

https://github.com/hyrise/hyrise/blob/6af75b3e402f3d95907d55e42b31a7bcf1562fa0/src/lib/storage/lqp_view.hpp#L27

For me, views should behave just like stored tables. Each column should have only one name.

janetzki:
- Close #1748 
- This partially reverts changes from #1692.
- The WITH feature (#1715) should use a similar solution.

### DIFF
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -482,9 +482,7 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_origin(const hsq
 
           const auto column_name_iter = view->column_names.find(column_id);
           if (column_name_iter != view->column_names.end()) {
-            for (const auto& column_name : column_name_iter->second) {
-              sql_identifier_resolver->add_column_name(column_expression, column_name);
-            }
+            sql_identifier_resolver->add_column_name(column_expression, column_name_iter->second);
           }
           sql_identifier_resolver->set_table_name(column_expression, hsql_table_ref.name);
         }
@@ -1029,11 +1027,11 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create(const hsql::Cr
 std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_view(const hsql::CreateStatement& create_statement) {
   auto lqp = _translate_select_statement(static_cast<const hsql::SelectStatement&>(*create_statement.select));
 
-  std::unordered_map<ColumnID, std::vector<std::string>> column_names;
+  std::unordered_map<ColumnID, std::string> column_names;
 
   for (auto column_id = ColumnID{0}; column_id < lqp->column_expressions().size(); ++column_id) {
     for (const auto& identifier : _inflated_select_list_elements[column_id].identifiers) {
-      column_names[column_id].emplace_back(identifier.column_name);
+      column_names.emplace(column_id, identifier.column_name);
     }
   }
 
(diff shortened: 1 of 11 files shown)


src/lib/sql/sql_translator.cpp
```diff
@@ -1043,7 +1041,7 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_view(const hsq
                 "Number of Columns in CREATE VIEW does not match SELECT statement");
 
     for (auto column_id = ColumnID{0}; column_id < create_statement.viewColumns->size(); ++column_id) {
-      column_names[column_id].emplace_back((*create_statement.viewColumns)[column_id]);
+      column_names.insert_or_assign(column_id, (*create_statement.viewColumns)[column_id]);
     }
   }
```
julianmenzler: Add an `else` clause here to move the two for-loops from above into it:

```cpp
if (create_statement.viewColumns) {
   ...
else {
    for (auto column_id = ColumnID{0}; column_id < lqp->column_expressions().size(); ++column_id) {
	    for (const auto& identifier : _inflated_select_list_elements[column_id].identifiers) {
	      column_names.insert_or_assign(column_id, identifier.column_name);
	    }
     }
}
```

src/test/logical_query_plan/create_view_node_test.cpp
```diff
@@ -53,8 +52,8 @@ TEST_F(CreateViewNodeTest, Equals) {
 
 TEST_F(CreateViewNodeTest, Copy) {
   const auto same_view_node = MockNode::make(MockNode::ColumnDefinitions({{DataType::Int, "a"}}));
-  const auto same_view = std::make_shared<LQPView>(
-      _view_node, std::unordered_map<ColumnID, std::vector<std::string>>{{ColumnID{0}, {"a"}}});
+  const auto same_view =
+      std::make_shared<LQPView>(_view_node, std::unordered_map<ColumnID, std::string>{{ColumnID{0}, {"a"}}});
```
julianmenzler:
```suggestion
      std::make_shared<LQPView>(_view_node, std::unordered_map<ColumnID, std::string>{{ColumnID{0}, "a"}});
```

src/lib/storage/lqp_view.hpp
```diff
@@ -15,16 +15,13 @@ class AbstractLQPNode;
  */
 class LQPView {
  public:
-  LQPView(const std::shared_ptr<AbstractLQPNode>& lqp,
-          const std::unordered_map<ColumnID, std::vector<std::string>>& column_names);
+  LQPView(const std::shared_ptr<AbstractLQPNode>& lqp, const std::unordered_map<ColumnID, std::string>& column_names);
```
mrks: Not really part of this PR, but should take no more than five minutes:
(1) Why is this a map anyway? Either we pass in nothing, or we give a list of the new column names. Partially renaming columns does not work, does it?
(2) As `column_names` will be copied anyway, it would be better to pass it in by value and `std::move` it in the initializer list. That would allow us to default it to `= {}`.

julianmenzler: - (1) It should also work with a plain `vector`. Indices would represent `ColumnID`s then. Renaming columns after `LQPView` was created does not make sense.
- (2) Great suggestion!

src/test/logical_query_plan/create_view_node_test.cpp
janetzki: Good catch!

src/lib/storage/lqp_view.hpp
janetzki: - (1) It seems that we use a map instead of a vector to keep Hyrise's architecture consistent. `CreateView` inherits corresponding function signatures from `AbstractOperator`.

janetzki: - (2) I agree and adjusted the initializer.

janetzki:
@julianmenzler I addressed your proposals.

PR #1759, opened 2019-07-21
TPC-C Preparation

### LINKED ISSUE #936: Same interface for TPCC and TPCH table generation
Currently, the `TpccTableGenerator` create the tables and takes care of encoding them (after PR #927  is merged, via an `EncodingConfig`). The `TpchDbGenerator` (maybe naming could be more consistent here) creates the table and takes care of storing them in the storage manager but does not take care of encoding them. 

So we have:

| TPC-? | Generate Tables | Stores Tables | Encodes Tables |
|:---------:|:-------------:|:-----:|:------:|
| TPC-**H** | yes | yes | no |
| TPC-**C** | yes |   no | yes |

Maybe this should be unified with the same interface. There is an `AbstractBenchmarkTableGenerator`, which only TPCC uses. We should/could use that and extend its interface to included the options we want.

Up for discussion:
------------------

Should a table generator:

- [ ] store tables in the storage manager
- [ ] encoded tables

mrks:
## TPC-C

- Harmonizes capitalization of TPCC/TPCH
- Adapts the TPC-C data generation to the AbstractTableGenerator interface used by the BenchmarkRunner, closes #936
- Fixes 1-indexing in TPC-C data generation
- Removes generation of TPC-C tables from the console. I haven’t used it in months and didn’t need it when developing the TPC-C implementation. If it turns out that it is helpful, we should re-write it in a way where we can pass the number of warehouses. The auto-completion for the table names (and the generation of single tables) does not seem to be required anymore.

I’ll add the TPC-C procedures and the actual benchmark in a follow-up PR to keep this consumable.

## BenchmarkRunner et al.

- Added start time of benchmarks to JSON output, needed for plotting
- Changed format slightly. Instead of one `duration` and one `metrics` list, have a list with exactly one entry per run
- Added the possibility of having a weighted distribution for shuffled benchmarks
- Makes storing the SQL metrics in the JSON optional
- Added the possibility of unsuccessful item runs (i.e., transaction failures)


mrks:
@Bensk1: Any chance you could have a look at that Monday morning?

src/benchmarklib/tpcc/tpcc_table_generator.hpp
```diff
@@ -38,40 +39,168 @@ class TpccTableGenerator : public opossum::AbstractBenchmarkTableGenerator {
...
   void _add_order_line_column(std::vector<Segments>& segments_by_chunk, TableColumnDefinitions& column_definitions,
                               std::string name, std::shared_ptr<std::vector<size_t>> cardinalities,
-                              order_line_counts_type order_line_counts,
+                              OrderLineCounts order_line_counts,
                               const std::function<T(std::vector<size_t>)>& generator_function);
 
-  TpccRandomGenerator _random_gen;
+  TPCCRandomGenerator _random_gen;
+
+  /**
```
mrks: From here until the end, this was moved from the AbstractBenchmarkTableGenerator

src/benchmarklib/tpcc/tpcc_table_generator.cpp
```diff
@@ -59,59 +55,59 @@ std::shared_ptr<Table> TpccTableGenerator::generate_items_table() {
...
-  add_column<pmr_string>(segments_by_chunk, column_definitions, "W_STATE", cardinalities,
-                         [&](std::vector<size_t>) { return pmr_string{_random_gen.astring(2, 2)}; });
-
-  add_column<pmr_string>(segments_by_chunk, column_definitions, "W_ZIP", cardinalities,
-                         [&](std::vector<size_t>) { return pmr_string{_random_gen.zip_code()}; });
-  add_column<float>(segments_by_chunk, column_definitions, "W_TAX", cardinalities,
-                    [&](std::vector<size_t>) { return _random_gen.random_number(0, 2000) / 10000.f; });
-  add_column<float>(segments_by_chunk, column_definitions, "W_YTD", cardinalities, [&](std::vector<size_t>) {
+  _add_column<int>(segments_by_chunk, column_definitions, "W_ID", cardinalities,
+                   [&](std::vector<size_t> indices) { return indices[0] + 1; });
```
mrks: The `+ 1` parts should be the only real change in this file.

Bensk1:
Sure!

src/benchmarklib/benchmark_sql_executor.hpp
```diff
@@ -24,23 +23,28 @@ class BenchmarkSQLExecutor {
...
                        const std::optional<const std::string>& description = std::nullopt);
   void _verify_with_sqlite(SQLPipeline& pipeline);
   void _visualize(SQLPipeline& pipeline) const;
 
-  const bool _enable_jit;
-  const std::shared_ptr<SQLiteWrapper> _sqlite_wrapper;
-  const std::optional<std::string> _visualize_prefix;
-
-  const std::shared_ptr<TransactionContext> _transaction_context;
+  bool _enable_jit;
```
Bensk1: Why was the const removed for these member variables?

src/bin/console/console.cpp
```diff
@@ -1003,14 +968,7 @@ char** Console::_command_completion(const char* text, int start, int end) {
   // Choose completion function depending on the input. If it starts with "generate",
```
Bensk1: The comment is a leftover?

src/benchmarklib/abstract_benchmark_item_runner.hpp
```diff
@@ -41,6 +41,11 @@ class AbstractBenchmarkItemRunner {
   // Set the SQLite wrapper used for query verification. `nullptr` disables verification. Default is disabled.
   void set_sqlite_wrapper(const std::shared_ptr<SQLiteWrapper>& sqlite_wrapper);
 
+  // Returns a mapping from item ID to its relative weight in the execution of the benchmark. Relevant for example in
+  // the TPC-C benchmark, where not all transactions are executed equally often. Returns an empty vector if all items
+  // are weighted the same
```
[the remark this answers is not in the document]
mrks: Would be easy to add, but we would either have to (a) implement it in each runner, even if no weights are used or (b) have it create the vector each time and return it by value instead of returning a reference to a static vector. Furthermore, I like the check in `benchmark_runner.cpp` where `.empty()` tells us if we are dealing with weights at all.

src/benchmarklib/benchmark_item_run_result.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include "sql/sql_pipeline.hpp"
+
+namespace opossum {
+
+// Stores the result of a SINGLE run of a single benchmark item (e.g., one execution of TPC-H query 5).
+struct BenchmarkItemRunResult {
+  BenchmarkItemRunResult(Duration begin, Duration duration, std::vector<SQLPipelineMetrics> metrics);
+
+  // Stores the begin timestamp of this run (measured as time since start of benchmark)
+  Duration begin;
```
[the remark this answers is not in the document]
mrks: Any suggestion? I would say it is not a timestamp (as it is relative).

src/bin/console/console.cpp
mrks: Whoopsie, thanks.

mrks:
> Just a few minor things. Most changes are renamings or moved code, right?

Yes. Maybe I should have asked for a review on the actual changes only and add the renaming later.

src/benchmarklib/benchmark_sql_executor.hpp
mrks: Leftover.

src/benchmarklib/abstract_benchmark_item_runner.hpp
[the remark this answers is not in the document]
Bensk1: Let's keep it simpel then.

Bensk1:
> > Just a few minor things. Most changes are renamings or moved code, right?
> 
> Yes. Maybe I should have asked for a review on the actual changes only and add the renaming later.

That was no criticism, more a way of justifying comments in only a few places.

src/benchmarklib/benchmark_item_run_result.hpp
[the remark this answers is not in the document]
Bensk1: Can we use our [`TimePoint`](https://github.com/hyrise/hyrise/blob/master/src/benchmarklib/benchmark_config.hpp#L17) data type?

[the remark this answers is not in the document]
mrks: No, `time_point` is the `chrono` way of saying "timestamp". If you subtract two timestamps (begin of this run - begin of the benchmark), you get a `Duration`.

mrks:
Anything missing?

Bensk1:
> Anything missing?

Only the timepoint and duration discussion.

mrks:
No, I just wanted to break the build. 8a583b2 is what I'd like you to look at. For the ordered mode, it basically reverts it to what we had previously.

PR #1760, opened 2019-07-22
New plot script: Run duration over time

mrks:
Adds a script that plots the duration of individual benchmark runs over time. Example:

tpcc

Detailed version:

tpcc_detailed

### DIFF
--- a/scripts/plot_run_duration_over_time.json
+++ b/scripts/plot_run_duration_over_time.json
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+
+# Takes a single benchmark output JSON and plots the duration of item executions over time
+
+# TODO add to requirements.txt
+import json
+import matplotlib.patches as mpatches
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import sys
+
+if(len(sys.argv) != 2):
+    print('Usage: ' + sys.argv[0] + ' benchmark.json')
+    exit()
+
+with open(sys.argv[1]) as json_file:
+    data_json = json.load(json_file)['benchmarks']
+
+# Build flat list of {name, begin, duration} entries for every benchmark item run
+data = []
+for benchmark_json in data_json:
+	name = benchmark_json['name']
+	for run_json in benchmark_json['runs']:
+		begin = run_json['begin']
+		duration = run_json['duration']
+		data.append({'name': name, 'begin': begin, 'duration': duration})
+
+fig, ax = plt.subplots()
+
+df = pd.DataFrame(data)
+colors = np.where(df['name'] == 'Delivery','r','-')
+colors[df['name'] == 'New-Order'] = 'g'
+colors[df['name'] == 'Order-Status'] = 'b'
+colors[df['name'] == 'Payment'] = 'c'
+colors[df['name'] == 'Stock-Level'] = 'm'
+df.reset_index().plot(kind='scatter', x='begin', y='duration', c=colors)
+
+handles = [
+	mpatches.Patch(color='r', label='Delivery'),
+	mpatches.Patch(color='g', label='NewOrder'),
+	mpatches.Patch(color='b', label='OrderStatus'),
+	mpatches.Patch(color='c', label='Payment'),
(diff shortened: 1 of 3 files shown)


scripts/plot_run_duration_over_time.json
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+
+# Takes a single benchmark output JSON and plots the duration of item executions over time
+
+# TODO add to requirements.txt
```
mrks: What is the best way to do that? I see that requirements.txt also holds the version numbers. Do I just take the current one?

scripts/plot_run_duration_over_time.py
```diff
@@ -0,0 +1,59 @@
...
+import sys
+
+if(len(sys.argv) != 2):
+    print('Usage: ' + sys.argv[0] + ' benchmark.json')
+    exit()
+
+with open(sys.argv[1]) as json_file:
+    data_json = json.load(json_file)['benchmarks']
+
+# Build flat list of {name, begin, duration} entries for every benchmark item run
```
mrks: As is evident, I am not the biggest Python hacker. I can make things work, but there are probably various things that could be improved here. Please let me know!

scripts/plot_run_duration_over_time.py
```diff
@@ -0,0 +1,60 @@
...
+
+name_to_color[benchmark_names[0]] = default_colors[0]
+colors = np.where(df['name'] == benchmark_names[0], default_colors[0], '-')
+for i in range(1, len(benchmark_names)):
+	color = default_colors[i % len(default_colors)]
+	name_to_color[benchmark_names[i]] = color
+	colors[df['name'] == benchmark_names[i]] = color
+
+# Plot
+df.reset_index().plot(kind='scatter', x='begin', y='duration', c=colors, s=.3)
```
mrks: Not really happy with the axis, but I haven't found a way of automatically formatting a duration with the best unit.

Bensk1: I guess that's fine, but we could add the unit `[ns]`?

scripts/plot_run_duration_over_time.json
Bensk1: `pip freeze`

Bensk1:
1. Why do new-order-prodcedures at the same point of time take from roughly 0.3 to roughly 1, do they add a different number of orders?
2. Does it make sense to draw everything into the same graph? I cannot really see anything for payment, stock-level, and order-status.

mrks:
> Why do new-order-prodcedures at the same point of time take from roughly 0.3 to roughly 1, do they add a different number of orders?

They add 5 to 15 order lines, which is why you should see 11 "lines" in there. Bear in mind that everything is still super inefficient.

> Does it make sense to draw everything into the same graph? I cannot really see anything for payment, stock-level, and order-status.

Added.

Bensk1:
That looks already much better. If we create a collection of these over time, I am sure we can sell it as modern art at some point.

mrks:
Anything else that needs to be done here?

mrks:
Closing this for now as #1766 will change the JSON format. Will re-open and change to new JSON format once that is done.

scripts/plot_run_duration_over_time.py
```diff
@@ -22,28 +22,34 @@
 for benchmark_json in data_json:
 	name = benchmark_json['name']
 	benchmark_names.append(name)
-	for run_json in benchmark_json['runs']:
-		begin = run_json['begin']
-		duration = run_json['duration']
-		data.append({'name': name, 'begin': begin / 1e9, 'duration': duration})
+	for run_json in benchmark_json['successful_runs']:
+		data.append({'name': name, 'begin': run_json['begin'] / 1e9, 'duration': run_json['duration'], 'success': True})
+	for run_json in benchmark_json['unsuccessful_runs']:
```
mrks: I'm not really happy with the code duplication here, but I have no idea how to avoid it without making everything more verbose.

Bensk1: I am sure we can handle these two duplicated lines

PR #1761, opened 2019-07-22
Hold state in Hyrise Singleton

### LINKED ISSUE #1677: Hyrise singleton instance

d-justen:
fixes #1677


src/test/operators/product_test.cpp
```diff
@@ -6,11 +6,11 @@
 #include "base_test.hpp"
 #include "gtest/gtest.h"
 
+#include "hyrise.hpp"
```
julianmenzler: required?

src/test/operators/projection_test.cpp
```diff
@@ -8,13 +8,13 @@
 
 #include "expression/expression_functional.hpp"
 #include "expression/pqp_column_expression.hpp"
+#include "hyrise.hpp"
```
julianmenzler: required?

src/test/operators/sort_test.cpp
```diff
@@ -5,6 +5,7 @@
 #include "base_test.hpp"
 #include "gtest/gtest.h"
 
+#include "hyrise.hpp"
```
julianmenzler: required?

src/test/operators/table_scan_string_test.cpp
```diff
@@ -9,13 +9,13 @@
 #include "gtest/gtest.h"
 
 #include "expression/evaluation/like_matcher.hpp"
+#include "hyrise.hpp"
```
julianmenzler: required?

src/test/operators/validate_test.cpp
```diff
@@ -8,12 +8,12 @@
 
 #include "concurrency/transaction_context.hpp"
 #include "expression/expression_functional.hpp"
+#include "hyrise.hpp"
```
julianmenzler: required?

src/test/optimizer/dp_ccp_test.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "cost_estimation/cost_estimator_logical.hpp"
 #include "expression/expression_functional.hpp"
+#include "hyrise.hpp"
```
julianmenzler: required?

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include "base_test.hpp"
 #include "expression/expression_functional.hpp"
+#include "hyrise.hpp"
```
julianmenzler: required?

src/test/optimizer/strategy/subquery_to_join_rule_test.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "expression/expression_functional.hpp"
 #include "expression/lqp_column_expression.hpp"
+#include "hyrise.hpp"
```
julianmenzler: required?

src/test/tpc/tpch_test.cpp
```diff
@@ -8,6 +8,7 @@
 #include "gtest/gtest.h"
 
 #include "constant_mappings.hpp"
+#include "hyrise.hpp"
```
julianmenzler: required?

src/test/lib/hyrise_test.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "types.hpp"
+#include "utils/plugin_manager.hpp"
+
+namespace opossum {
+
+class HyriseTest : public BaseTest {
+ protected:
+  void SetUp() override { Hyrise::reset(); }
+
+  std::unordered_map<PluginName, PluginHandleWrapper>& get_plugins() {
```
julianmenzler: You could add a comment, why we need this function. Private member access, I guess?

src/benchmark/micro_benchmark_main.cpp
```diff
@@ -30,7 +30,7 @@
  * before the end of main().
  *
  *
- * ---> But similar things could happen anytime you call, e.g. StorageManager::get() before calling
+ * ---> But similar things could happen anytime you call, e.g. Hyrise::get().storage_manager before calling
```
mrks: Please have a look at this whole comment and if it is still relevant. The static init order fiasco should be resolved by this PR. Maybe we should call `get_default_memory_resource()` before initializing any other class that was previously a singleton.

src/lib/storage/storage_manager.hpp
```diff
@@ -19,7 +18,7 @@ class AbstractLQPNode;
 
 // The StorageManager is a singleton that maintains all tables
 // by mapping table names to table instances.
-class StorageManager : public Singleton<StorageManager> {
+class StorageManager {
```
mrks: For all classes that were previously singleton, we should consider inheriting from Noncopyable. We don't want to end up with two StorageManagers by accident.

src/lib/hyrise.hpp
```diff
@@ -0,0 +1,19 @@
...
+namespace opossum {
+
+class Hyrise : public Singleton<Hyrise> {
+ public:
+  static void reset() { get() = Hyrise{}; }
+
+  PluginManager plugin_manager;
+  StorageManager storage_manager;
+  TransactionManager transaction_manager;
+};
```
julianmenzler: I just noticed that `CurrentScheduler` also implements a singleton. I guess we should include it here as well?

mrks: No, it just holds static methods. In fact, it should be changed so that it is not even initializable.

julianmenzler: True, that singleton stuff should be removed then.

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -1,6 +1,7 @@
 #include "stored_table_node.hpp"
 
 #include "expression/lqp_column_expression.hpp"
+#include "hyrise.hpp"
 #include "statistics/table_statistics.hpp"
 #include "storage/index/index_statistics.hpp"
 #include "storage/storage_manager.hpp"
```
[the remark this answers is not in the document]
d-justen: Actually not, `StoredTableNode` uses the Hyrise state quite often

src/lib/utils/plugin_manager.hpp
```diff
@@ -28,16 +28,10 @@ class PluginManager : public Singleton<PluginManager> {
...
-  // Deletes the entire PluginManager and creates a new one, used especially in tests.
-  // This can lead to a lot of issues if there are still running tasks / threads that
-  // want to access a resource. You should be very sure that this is what you want.
-  // Have a look at base_test.hpp to see the correct order of resetting things.
-  static void reset();
-
  protected:
-  friend class Singleton;
+  PluginManager() = default;
+  friend class Hyrise;
```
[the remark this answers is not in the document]
d-justen: There doesn't seem to be any convention, just wanted to keep it consistent with the other Singletons

src/lib/logical_query_plan/stored_table_node.cpp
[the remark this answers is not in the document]
julianmenzler: Oh sorry. I was mistaken here.

src/lib/storage/storage_manager.hpp
```diff
@@ -56,24 +55,12 @@ class StorageManager : public Singleton<StorageManager> {
...
 
-  StorageManager(StorageManager&&) = delete;
-
  protected:
-  StorageManager() {}
-
-  friend class Singleton;
-
-  const StorageManager& operator=(const StorageManager&) = delete;
-  StorageManager& operator=(StorageManager&&) = default;
```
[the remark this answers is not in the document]
d-justen: We don't, this is now solved by inheriting from Noncopyable

src/lib/utils/plugin_manager.hpp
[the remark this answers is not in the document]
mrks: No convention, no preference.

src/lib/hyrise.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include "concurrency/transaction_manager.hpp"
+#include "storage/storage_manager.hpp"
+#include "utils/plugin_manager.hpp"
+#include "utils/singleton.hpp"
+
+namespace opossum {
+
+class Hyrise : public Singleton<Hyrise> {
+ public:
+  // Resets the Hyrise state by deleting the Plugin-, Storage-, and TransactionManager
```
mrks: "by deleting its members (e.g., StorageManager)". Otherwise, that comment might become outdated when a member is added.

src/lib/hyrise.hpp
```diff
@@ -0,0 +1,39 @@
...
+  // and creating new ones. This is used especially in tests and can lead to a lot of
+  // issues if there are still running tasks / threads that want to access a resource.
+  // You should be very sure that this is what you want. Have a look at base_test.hpp
+  // to see the correct order of resetting things.
```
mrks: The "correct order" thing should be a thing of the past now, shouldn't it?

src/lib/hyrise.hpp
```diff
@@ -0,0 +1,39 @@
...
+  static void reset() { get() = Hyrise{}; }
+
+  PluginManager plugin_manager;
+  StorageManager storage_manager;
+  TransactionManager transaction_manager;
+
+ private:
+  Hyrise() {
+    // The default_memory_resource must be initialized before Plugin-, Storage-, and
+    // TransactionManager so that it is the last object to be destructed. This prevents
```
mrks: see above. No need to explicitly list them here.

src/lib/hyrise.hpp
```diff
@@ -0,0 +1,39 @@
...
+    // e.g. the StorageManager destructor to deallocate a table that is already
+    // destructed by the default_memory_resource destructor.
```
mrks: Not destructed. The destructor has never been called. The memory is gone, meaning when the destructor is called, its free() calls will fail.

src/lib/concurrency/transaction_manager.hpp
```diff
@@ -48,11 +47,11 @@ class TransactionContext;
  * which represents the current global visibility of records.
  * The TransactionManager is thread-safe.
  */
-class TransactionManager : public Singleton<TransactionManager> {
+class TransactionManager : public Noncopyable {
   friend class TransactionManagerTest;
 
  public:
-  static void reset();
+  TransactionManager& operator=(TransactionManager&) = delete;
```
mrks: What is this supposed to do? `operator=(TransactionManager&)` is not really a thing. `operator=(const TransactionManager&)` is. That one should be covered by Noncopyable. Am I missing something?

src/benchmark/micro_benchmark_main.cpp
```diff
@@ -1,49 +1,3 @@
...
-  // BENCHMARK_MAIN begin
-  ::benchmark::Initialize(&argc, argv);
-  if (::benchmark::ReportUnrecognizedArguments(argc, argv)) return 1;
-  ::benchmark::RunSpecifiedBenchmarks();
-  // BENCHMARK_MAIN end
-
-  // Destroy all benchmark::Fixture objects
-  ::benchmark::ClearRegisteredBenchmarks();
-}
+BENCHMARK_MAIN();
```
mrks: Have you tested the benchmarks? They are not part of the CI.

src/lib/hyrise.hpp
julianmenzler: Actually, `CurrentScheduler` does not really implement the Singleton pattern. It does not create any instances, which is a key property of the pattern. Instead, it just holds an instance of `AbstractScheduler`, not an instance of `CurrentScheduler`. The following comment is misleading:

https://github.com/hyrise/hyrise/blob/f452c8c71ebb200e9b1b224195e0e669ba279442/src/lib/scheduler/current_scheduler.hpp#L14-L20


Anyhow, in my opinion the current scheduler instance belongs to the *global state* of Hyrise. Or am I wrong?

mrks: Should have referenced #1786 here.

mrks: I'd make this a follow-up so that we can get this in somewhat quickly. Considering that this touches 137 files, I'd like to keep the number of conflicts low.

julianmenzler:
I found some more singletons: 

* Topology
* JitRepository
* Console

src/lib/hyrise.hpp
d-justen: I thought that "correct order" meant setting the CurrentScheduler to nullptr first to be sure that all tasks are done and then resetting Hyrise. But the sentence might still be misleading, so I'd be fine with just dropping it.

src/lib/hyrise.hpp
d-justen: I'm struggling a bit to find a good and brief explanation. What about:

> The default_memory_resource must be initialized before Hyrise's members in order to be torn down after them. If a member (e.g., the StorageManager) saves data that was allocated via the default_memory_resource, we want the member to deallocate the data first before the memory resource is destroyed.

mrks: Proposal:

The default_memory_resource must be initialized before Hyrise's members so that it is destructed after them and remains accessible during their deconstruction. For example, when the StorageManager is destructed, it causes its stored tables to be deconstructed, too. As these might call deallocate on the default_memory_resource, it is important that the resource has not been destructed before. As objects are destructed in the reverse order of their construction, explicitly initializing the resource first means that it is destructed last.

src/lib/hyrise.hpp
mrks: Ah, makes sense. Never mind then.

src/benchmark/micro_benchmark_main.cpp
d-justen: Yes, they still work properly. But I have noticed that the `BM_HashSemiProbeRelationLarger` benchmark throws an error:
```console
  what():  src/lib/operators/join_sort_merge.cpp:63 JoinSortMerge doesn't support these parameters
```
This doesn't seem to have something to do with my PR. Is there already an issue filed for?

mrks: I don't think there is. Could you please run a `git bisect` for it and open an issue once you found the offending commit?

mrks:
@julianmenzler I'd say we bring this in and move the three missing singletons to #1786.

src/benchmark/micro_benchmark_main.cpp
julianmenzler: Awesome, I did not know of `git bisect` before!

d-justen: Got it, it's this one from over 3 months ago https://github.com/hyrise/hyrise/commit/c2d90a17d1812366b2b7fd5f5d52de53f46edb07.
Issue is open: #1798

PR #1765, opened 2019-07-28
Index Placement Rule

mweisgut:
This PR integrates a rule which decides when index joins should be used. I'll provide a more detailed description when this PR isn't in WIP status anymore.


mweisgut:
First, the Index Reference Join has to be fixed.

PR #1766, opened 2019-07-29
TPC-C Preparation Part 2

mrks:
I realized that we will need support for unsuccessful transactions (i.e., transaction aborts). As #1759 will change the JSON result format and this will change it again, I'd rather have only a single master commit.

Also makes the inclusion of SQL metrics optional, greatly reducing the size of the resulting JSON.

To make the review easier, this is a PR into #1759.

Sorry for making this slightly confusing. Hopefully, this makes the final TPC-C PR a bit easier to read.

### DIFF
--- a/scripts/compare_benchmarks.py
+++ b/scripts/compare_benchmarks.py
@@ -103,7 +103,7 @@ for old, new in zip(old_data['benchmarks'], new_data['benchmarks']):
         old_unsuccessful_per_second = float(len(old['unsuccessful_runs'])) / (old['duration'] / 1e9)
         new_unsuccessful_per_second = float(len(new['unsuccessful_runs'])) / (new['duration'] / 1e9)
 
-        if len(old['unsuccessful_runs']) > 0.0:
+        if len(old['unsuccessful_runs']) > 0:
             diff_unsuccessful = float(new_unsuccessful_per_second / old_unsuccessful_per_second)
         else:
             diff_unsuccessful = float('nan')
--- a/scripts/compare_benchmarks.py
+++ b/scripts/compare_benchmarks.py
@@ -11,14 +11,16 @@ p_value_significance_threshold = 0.001
 min_iterations = 10
 min_runtime_ns = 59 * 1000 * 1000 * 1000
 
-def format_diff(diff):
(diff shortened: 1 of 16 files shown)


scripts/compare_benchmarks.py
```diff
@@ -95,7 +97,28 @@ def calculate_and_format_p_value(old, new):
...
     p_value_formatted = calculate_and_format_p_value(old, new)
 
-    table_data.append([name, str(old['items_per_second']), str(len(old['runs'])), str(new['items_per_second']), str(len(new['runs'])), diff_formatted, p_value_formatted])
+    table_data.append([name, str(old['items_per_second']), str(len(old['successful_runs'])), str(new['items_per_second']), str(len(new['successful_runs'])), diff_formatted, p_value_formatted])
+
+    if (len(old['unsuccessful_runs']) > 0 or len(new['unsuccessful_runs']) > 0):
+        old_unsuccessful_per_second = float(len(old['unsuccessful_runs'])) / (old['duration'] / 1e9)
+        new_unsuccessful_per_second = float(len(new['unsuccessful_runs'])) / (new['duration'] / 1e9)
+
+        if len(old['unsuccessful_runs']) > 0.0:
```
Bensk1: Is the `len` comparison with a float intended?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -177,6 +180,10 @@ void BenchmarkRunner::_benchmark_shuffled() {
   }
   _state.set_done();
 
+  for (auto& result : _results) {
```
Bensk1: Maybe another comment or a pointer to the comment in benchmark_item_result.hpp might be helpful for the reader.

scripts/compare_benchmarks.py
mrks: Nope, just copypasta.

src/benchmarklib/benchmark_runner.cpp
mrks: laik dis?

Bensk1: Naitz

Issue #670, opened 2018-01-31
AdaptiveRadixTreeIndex behaves differently from other CompositeGroupKeyIndex


Beta-Alf:
For #667 we test the index join with multiple Index implementations. One test (OuterJoinWithNull) is failed with the AdaptiveRadixTreeIndex implementation. When stepping through the code we can confirm that the index returns a wrong range (returning NULLs also when asked for elements equal to 12345).

Repro: check out the join_index branch and enable the test OuterJoinWithNull. Run the test and it will fail. The Index is queried in _join_two_columns in join_index.cpp.

Bensk1:
The indexes were implemented when NULL values where not supported at all in Hyrise.

Bouncner:
So disable it as a return of getIndexes?
@bensk , can you estimate how much work the fix would be? We probably need to add NULL-tests to all indices as well. Does it make sense for your index group to have a look into this?

mrzzzrm:
@mweisgut Have you seen this? If this is included on your current work on the indices, can you assign yourself? (trying to tidy up the issue list a bit)

mweisgut:
> @mweisgut Have you seen this? If this is included on your current work on the indices, can you assign yourself? (trying to tidy up the issue list a bit)

I have only used GroupKeyIndex and BTreeIndex, so I have not seen it yet.

PR #1772, opened 2019-07-31
Null-aware GroupKeyIndex

### LINKED ISSUE #1762: IndexScan crashes when columns are pruned
After adding indexes, the TPC-H benchmark crashes when an index scan is used and columns are pruned. Probably the column ids are not updated and [column_ids_after_pruning](https://github.com/mweisgut/hyrise/blob/d3f338e6f947df5239ee1be40dc2f6e974ac65b2/src/lib/utils/column_ids_after_pruning.cpp) might help.

Benchmark flags:
`./build/hyriseBenchmarkTPCH -s 0.001 -r 1`

- [Failing assertion](https://github.com/mweisgut/hyrise/blob/a0415b8d8ed5aeac97e2b7b668bbddebef4ec199/src/lib/operators/index_scan.cpp#L111)

- [How indexes were added](https://github.com/mweisgut/hyrise/blob/d5049f8190ac5efbf2e93e2fdee411a6fb014f10/src/benchmarklib/abstract_table_generator.cpp#L46-L96)

mweisgut:
Fixes #1762


src/lib/storage/index/group_key/group_key_index.cpp
```diff
@@ -20,9 +20,9 @@ GroupKeyIndex::GroupKeyIndex(const std::vector<std::shared_ptr<const BaseSegment
   Assert((segments_to_index.size() == 1), "GroupKeyIndex only works with a single segment.");
 
   // 1) Initialize the index structures
-  // 1a) Set the index_offset to size of the dictionary + 1 (plus one to mark the ending position)
+  // 1a) Set the index_offset to size of the dictionary + 2 (plus one for null and plus one to mark the ending position)
   //     and set all offsets to 0
-  _index_offsets = std::vector<size_t>(_indexed_segments->unique_values_count() + 1u, 0u);
+  _index_offsets = std::vector<size_t>(_indexed_segments->unique_values_count() + 2u, 0u);
```
mrks: Add a comment explaining that and how the rest of NULL value handling happens automagically?

src/test/storage/group_key_index_test.cpp
```diff
@@ -11,15 +11,32 @@
...
 
 namespace opossum {
 
 class GroupKeyIndexTest : public BaseTest {
  protected:
   void SetUp() override {
-    dict_segment = BaseTest::create_dict_segment_by_type<pmr_string>(
-        DataType::String, {"hotel", "delta", "frank", "delta", "apple", "charlie", "charlie", "inbox"});
+    auto value_segment_str = std::make_shared<ValueSegment<pmr_string>>(true);
+    //                                          corresponding ValueID
```
mweisgut: Position in corresponding attribute vector

src/lib/storage/index/group_key/group_key_index.cpp
Bensk1: This has nothing to do with the current PR but now that I see it: the `_index_offsets` points into the `_index_postings`. The `_index_postings` has the same size as the indexed segment which should be at maximum `ChunkOffset` which is a 4 byte instead an 8 byte (`size_t`) integer, right? Could we save memory here?

Bensk1: Maybe you could also add the explanation to the schematic explanation in the header file.

src/test/storage/group_key_index_test.cpp
```diff
@@ -11,15 +11,32 @@
...
+    value_segment_str->append("delta");     //  2
+    value_segment_str->append("frank");     //  3
+    value_segment_str->append("delta");     //  4
+    value_segment_str->append(NULL_VALUE);  //  5
+    value_segment_str->append(NULL_VALUE);  //  6
+    value_segment_str->append("apple");     //  7
+    value_segment_str->append("charlie");   //  8
+    value_segment_str->append("charlie");   //  9
+    value_segment_str->append("inbox");     // 10
+    value_segment_str->append(NULL_VALUE);  // 11
```
mrks: Let this be reflective of the example in the hpp? Also, the example is missing the offset to the end of the postings list. That was broken before.

src/test/operators/join_index_test.cpp
```diff
@@ -130,8 +130,7 @@ class JoinIndexTest : public BaseTest {
       _table_wrapper_m, _table_wrapper_n;
 };
 
-typedef ::testing::Types<AdaptiveRadixTreeIndex, CompositeGroupKeyIndex, BTreeIndex /* , GroupKeyIndex */>
-    DerivedIndices;
+typedef ::testing::Types<AdaptiveRadixTreeIndex, CompositeGroupKeyIndex, BTreeIndex, GroupKeyIndex> DerivedIndices;
```
mrks: I just realized that the index tests do not cover NULL values. Scrolling through the code, I worry that this PR is breaking more than it is fixing. For example, wouldn't a scan on `WHERE a > 5` also include all NULL values? Same for the join? If this is the case and it still survives the tests, the tests are insufficient.

Maybe we should discuss this in person together with @Bensk1?

src/test/storage/group_key_index_test.cpp
mweisgut: https://github.com/hyrise/hyrise/pull/1772/commits/068db8c262ffea70e14801ccdbfb89f00f5085b5

src/lib/storage/index/group_key/group_key_index.cpp
mweisgut: > Maybe you could also add the explanation to the schematic explanation in the header file.

https://github.com/hyrise/hyrise/commit/068db8c262ffea70e14801ccdbfb89f00f5085b5

> This has nothing to do with the current PR but now that I see it: the `_index_offsets` points into the `_index_postings`. The `_index_postings` has the same size as the indexed segment which should be at maximum `ChunkOffset` which is a 4 byte instead an 8 byte (`size_t`) integer, right? Could we save memory here?

In the worst-case scenario, each value of the segment is a distinct value. Therefore `_index_offsets` has the same size as `_index_postings`. So `ChunkOffset` should be sufficient, shouldn't it?

mweisgut: https://github.com/hyrise/hyrise/pull/1772/commits/a05d56d2a1df68c3dddd95fe71b776f5b9cdaf9d

src/test/operators/join_index_test.cpp
mweisgut: Your description sounds similar to this issue https://github.com/hyrise/hyrise/issues/670

I also checked  `index_scan_test` and `index_join_test` and couldn't find any null tests.
The only test which takes null values into account is the referenced one below, but it is disabled and even if we enable it, the used table `resources/test_data/tbl/join_operators/int_outer_join_null.tbl`  is not available anymore. https://github.com/hyrise/hyrise/blob/1a2397cdb9297231efa8b44b5693ee04feae7b80/src/test/operators/join_index_test.cpp#L262-L266

mweisgut:
Moved to #1783

PR #1774, opened 2019-08-02
Fix NULLable for aggregate, harden tests

### LINKED ISSUE #1698: check_table_equal does not check 'nullable'
- extend check_table_equal in check_table_equal.cpp to check whether the columns in the given table and the expected table are equal regarding their nullable flags
- add nullable flags to the output of check_table_equal
- fix all tests depending on check_table_equalto NOT check nullable flags (probably just have to change the expected tables TableColumnDefinitions)

### LINKED ISSUE #1769: Group by + Case + Null is broken
```
> load resources/test_data/tbl/int_float_with_null.tbl
> SELECT a, (CASE WHEN a IS NULL THEN 1 ELSE 2 END) FROM int_float_with_null GROUP BY a
=== Columns
|       a|CASE WHEN a IS NULL THEN 1 ELSE 2 END|
|     int|                                  int|
|not null|                                 null|
=== Chunk 0 ===
|<ValueS>|<ValueS>                             |
|   12345|                                    2|
|     123|                                    2|
|    NULL|                                    2|  // second column expected to be 1
|    1234|                                    2|
===
4 rows total
```

This does not occur if the `GROUP BY` is removed.

mrks:
AggregateHash always returned non-nullable columns, causing #1769.

This PR:
1) Fixes the nullability information returned by aggregates
2) Hardens the expression evaluator to DebugAssert that NULL values do not occur in non-nullable columns
3) Hardens the test suite by enforcing matching nullability information for everything but SQLite tests
4) Sneaks in a fix in the console so that `visualize` works from both the root and the build directory

fixes #1769
fixes #1698

### DIFF
--- a/src/lib/storage/index/group_key/group_key_index.hpp
+++ b/src/lib/storage/index/group_key/group_key_index.hpp
@@ -27,15 +27,19 @@ class GroupKeyIndexTest;
  *    |(i)| Attribute | Dictionary |  Index  | Index Postings |
  *    |   |  Vector   |            | Offsets |                |
  *    +---+-----------+------------+---------+----------------+
- *    | 0 |         4 | apple    ------->  0 ------------>  4 |  ie "apple" can be found at i = 4 in the AV
- *    | 1 |         2 | charlie  ------->  1 ------------>  5 |
- *    | 2 |         3 | delta    ------->  3 ---------|     6 |
- *    | 3 |         2 | frank    ------->  5 -------| |-->  1 |
- *    | 4 |         0 | hotel    ------->  6 -----| |       3 |  ie "delta" can be found at i = 1 and 3
- *    | 5 |         1 | inbox    ------->  7 ---| | |---->  2 |
- *    | 6 |         1 |            |         |  | |------>  0 |
- *    | 7 |         5 |            |         |  |-------->  7 |  ie "inbox" can be found at i = 7 in the AV
+ *    | 0 |         4 | apple    ------->  0 -------------> 4 |  ie "apple" can be found at i = 4 in the AV
+ *    | 1 |         2 | charlie  ------->  1 -------------> 5 |
+ *    | 2 |         3 | delta    ------->  3 ----------|    6 |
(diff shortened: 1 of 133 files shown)


src/lib/storage/index/group_key/group_key_index.hpp
```diff
@@ -27,16 +27,20 @@ class GroupKeyIndexTest;
  *    |(i)| Attribute | Dictionary |  Index  | Index Postings |
  *    |   |  Vector   |            | Offsets |                |
  *    +---+-----------+------------+---------+----------------+
- *    | 0 |         4 | apple    ------->  0 ------------>  4 |  ie "apple" can be found at i = 4 in the AV
```
Bensk1: Should that be part of #1772?

src/lib/utils/check_table_equal.cpp
```diff
@@ -206,6 +214,16 @@ std::optional<std::string> check_table_equal(const std::shared_ptr<const Table>&
...
       return stream.str();
     }
+
+    if (ignore_nullable == IgnoreNullable::No && left_column_is_nullable != right_column_is_nullable) {
+      const std::string error_type = "Column NULLable mismatch (column " + std::to_string(column_id) + ")";
+      const std::string error_msg = std::string{"Actual column is "} + (left_column_is_nullable ? "" : "NOT ") +
+                                    "NULL\n" + std::string{"Expected column is "} +
+                                    (right_column_is_nullable ? "" : "NOT ") + "NULL";
+
+      print_table_comparison(error_type, error_msg, {{2, column_id}});
```
Bensk1: Not sure, is the `2` a leftover and should also be replaced by `HEADER_SIZE`?

mrks: Negative. It is row #2 that should be compared. If you look in lines 204 and 214, you will see 0 and 1, respectively. If this was HEADER_SIZE, we would always compare the last header row, even if adding another row means that the nullable row is no longer the last one.

PR #1775, opened 2019-08-03
[JIT] Fix bug #1741 - no support for COUNT(*) without group by

### LINKED ISSUE #1741: COUNT(*) requires GROUP BY
While
```sql
SELECT COUNT(*) FROM mixed GROUP BY a;
```
works fine in `sqlite_testrunner_queries.sql`, executed by `hyriseSystemTest` (with clang), the query
```sql
SELECT COUNT(*) FROM mixed;
```
crashes in `JITAggregate::before_specialization()`.

FabianWiebe:
This pr fixes the issue that Aggregate in JIT could not handle COUNT(*) without a group by column.

`JitAggregate` always requires a value as a parameter for an aggregate function. Since COUNT(*) does not have a parameter, we just passed in the first value of the tuple as a dummy. This caused a problem if the tuple has no values and is now solved by adding a constant value to the tuple and using it as a parameter.

Closes #1741

### DIFF
--- a/resources/test_data/sqlite_testrunner_queries.sql
+++ b/resources/test_data/sqlite_testrunner_queries.sql
@@ -105,9 +105,9 @@ SELECT a AS x, SUM(b) FROM mixed GROUP BY x HAVING x > 10;
 SELECT a AS x, SUM(b) FROM mixed GROUP BY x HAVING x > 10;
 SELECT a AS a1, a AS a2 FROM mixed;
 SELECT a AS a1, b AS b2, b AS b3, a AS a3, b AS b1, a AS a2 FROM mixed;
-SELECT COUNT(*) AS cnt1, COUNT(*) AS cnt2, COUNT(*) AS cnt3 FROM mixed GROUP BY a;
+SELECT COUNT(*) AS cnt1, COUNT(*) AS cnt2, COUNT(*) AS cnt3 FROM mixed;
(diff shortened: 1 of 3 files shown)


resources/test_data/sqlite_testrunner_queries.sql
```diff
@@ -105,9 +105,9 @@ SELECT a AS x, SUM(b) FROM mixed GROUP BY x HAVING x > 10;
 SELECT a AS x, SUM(b) FROM mixed GROUP BY x HAVING x > 10;
 SELECT a AS a1, a AS a2 FROM mixed;
 SELECT a AS a1, b AS b2, b AS b3, a AS a3, b AS b1, a AS a2 FROM mixed;
-SELECT COUNT(*) AS cnt1, COUNT(*) AS cnt2, COUNT(*) AS cnt3 FROM mixed GROUP BY a;
```
mrks: Feel free to add this as an alternative. I'd still leave the old version intact here.

mrks:
Thank you for looking into this!

resources/test_data/sqlite_testrunner_queries.sql
FabianWiebe: The original reason for the change was to just revert all changes from the workaround: fb5f675be6b9e869fd2308b3f6199e51f5da63b5

mrks: I see. In any case, there is some value in checking that even that somewhat non-sensical query works.

PR #1781, opened 2019-08-09
ExpressionEvaluator: Cache expressions

### LINKED ISSUE #1470: Optimization Rule to avoid performing the same Projection multiple times
TPC-H 1 might benefit from this

```
      sum(l_extendedprice*(1-l_discount)) as sum_disc_price,
      sum(l_extendedprice*(1-l_discount)*(1+l_tax)) as sum_charge,
```

mrks:
This caches expressions that have already been evaluated, either in the same result column (SELECT (a+3)*(a+3)), or across columns (TPC-H Q1)

```
| TPC-H 1        | 0.606237113476 | 37   | 0.614059865475 | 37   | +1%        |                          0.0122 |
| TPC-H 2        | 7.50865459442  | 451  | 7.50186347961  | 451  | -0%        |                          0.6090 |
| TPC-H 3        | 6.30969047546  | 379  | 6.35418796539  | 382  | +1%        |                          0.3446 |
| TPC-H 4        | 1.8323905468   | 110  | 2.18110156059  | 131  | +19%       |                          0.0000 |
| TPC-H 5        | 4.26206302643  | 256  | 4.28413057327  | 258  | +1%        |                          0.0037 |
| TPC-H 6        | 41.7840576172  | 2508 | 41.8591270447  | 2512 | +0%        |                          0.0000 |
| TPC-H 7        | 1.5963242054   | 96   | 1.62044811249  | 98   | +2%        |                          0.0012 |
| TPC-H 8        | 4.73825788498  | 285  | 4.87165594101  | 293  | +3%        |                          0.0000 |
| TPC-H 9        | 1.6105440855   | 97   | 1.58452832699  | 96   | -2%        |                          0.0002 |
| TPC-H 10       | 3.02620267868  | 182  | 3.013651371    | 181  | -0%        |                          0.0331 |
| TPC-H 11       | 24.6073150635  | 1477 | 24.4388370514  | 1467 | -1%        |                          0.0000 |
| TPC-H 12       | 6.3315038681   | 380  | 7.27451658249  | 437  | +15%       |                          0.0000 |
| TPC-H 13       | 2.24641442299  | 135  | 2.23228502274  | 134  | -1%        |                          0.0067 |
| TPC-H 14       | 23.8029708862  | 1429 | 23.8495998383  | 1431 | +0%        |                          0.0265 |
| TPC-H 15       | 14.6416416168  | 879  | 14.5359859467  | 873  | -1%        |                          0.0000 |
| TPC-H 16       | 6.91219425201  | 415  | 6.88040733337  | 413  | -0%        |                          0.0000 |
| TPC-H 17       | 1.39224493504  | 84   | 1.42205262184  | 86   | +2%        |                          0.0000 |
| TPC-H 18       | 1.28179419041  | 77   | 1.30388629436  | 79   | +2%        |                          0.0001 |
| TPC-H 19       | 5.58016395569  | 335  | 5.56914710999  | 335  | -0%        |                          0.2077 |
| TPC-H 20       | 2.4787979126   | 149  | 2.40130209923  | 145  | -3%        |                          0.0000 |
| TPC-H 21       | 0.651131749153 | 40   | 0.727307200432 | 44   | +12%       |                          0.0000 |
| TPC-H 22       | 12.8345985413  | 771  | 12.3243761063  | 740  | -4%        |                          0.0000 |
```

However, it looks like those performance benefits are just weird outliers again. These are the expressions that are actually cached:

```
- Benchmarking TPC-H 1
reused l_extendedprice
reused l_extendedprice * (1 - l_discount)
reused 1
reused l_discount
- Benchmarking TPC-H 7
reused n_name
reused 'KENYA'
reused n_name
reused 'SAUDI ARABIA'
reused 1
- Benchmarking TPC-H 12
reused o_orderpriority
reused o_orderpriority
reused '1-URGENT'
reused o_orderpriority
reused '2-HIGH'
reused 1
reused 0
- Benchmarking TPC-H 14
reused l_extendedprice * (1 - l_discount)
- Benchmarking TPC-H 16
reused p_type
- Benchmarking TPC-H 19
reused p_size
reused l_quantity
reused 1
reused p_size
reused p_size
reused l_quantity
reused l_quantity
reused p_container
reused p_brand
reused 1
reused p_size
reused p_size
reused l_quantity
reused l_quantity
reused p_container
reused p_brand
```

Since it doesn't hurt and fixes #1470, I'd consider bringing it in anyway.


src/lib/expression/pqp_subquery_expression.cpp
```diff
@@ -43,11 +43,15 @@ std::string PQPSubqueryExpression::as_column_name() const {
 }
 
 bool PQPSubqueryExpression::_shallow_equals(const AbstractExpression& expression) const {
-  Fail("Can't compare PQPSubqueryExpressions");
+  const auto& other = static_cast<const PQPSubqueryExpression&>(expression);
+  return pqp == other.pqp && parameters == parameters;
 }
 
 size_t PQPSubqueryExpression::_on_hash() const {
-  Fail("PQPSubqueryExpressions can't, shouldn't and shouldn't need to be hashed");
```
mrks: Talked to mrzzzrm about this who said that the possibility of this PR was not considered back when this statement was made.

src/lib/expression/abstract_expression.hpp
```diff
@@ -127,21 +127,26 @@ inline std::ostream& operator<<(std::ostream& stream, const AbstractExpression&
 
 // Wrapper around expression->hash(), to enable hash based containers containing std::shared_ptr<AbstractExpression>
 struct ExpressionSharedPtrHash final {
-  size_t operator()(const std::shared_ptr<AbstractExpression>& expression) const { return expression->hash(); }
+  size_t operator()(const std::shared_ptr<const AbstractExpression>& expression) const { return expression->hash(); }
```
mrzzzrm: I'd provide both signatures, otherwise I would vouch for there being no copy when this is called with a non-const Expression. Same for ExpressionSharedPtrEqual

src/lib/expression/evaluation/expression_evaluator.hpp
```diff
@@ -197,7 +197,13 @@ class ExpressionEvaluator final {
   // One entry for each segment in the _chunk, may be nullptr if the segment hasn't been materialized
   std::vector<std::shared_ptr<BaseExpressionResult>> _segment_materializations;
 
+  // Optionally, uncorrelated selects can be evaluated by the caller and passed in to the evaluator. This way, they
+  // do not have to be executed multiple times by different evaluators
   const std::shared_ptr<const UncorrelatedSubqueryResults> _uncorrelated_subquery_results;
+
+  // Some expressions can be reused, either in the same result column (SELECT (a+3)*(a+3)), or across columns
+  // (TPC-H Q1)
+  ConstExpressionUnorderedMap<std::shared_ptr<BaseExpressionResult>> _evaluated_expressions;
```
mrzzzrm: I don't like it when caches are named by their key only - `_cached_expression_results` or `_expression_result_cache` is much more speaking, imo.

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -151,45 +151,66 @@ ExpressionEvaluator::ExpressionEvaluator(
 template <typename Result>
 std::shared_ptr<ExpressionResult<Result>> ExpressionEvaluator::evaluate_expression_to_result(
     const AbstractExpression& expression) {
+  // First, look in the cache
+  auto& evaluated_expression = _evaluated_expressions[expression.shared_from_this()];
```
mrzzzrm: Sorry to be that guy, but I'd prefer using find+iterators here as to not leave the cache in a temporarily invalid state.

mrzzzrm:
Any perf improvements visible in the plan visualization for Q1/Q14?

src/lib/expression/abstract_expression.hpp
mrks: Not sure if I can follow. If the compiler wouldn't find suitable `ExpressionSharedPtrHash` for `ExpressionUnorderedMap`, wouldn't it complain? This experiment shows that is does convert it to the matching form:
https://godbolt.org/z/E8-Ss7
Are you concerned about the use_count being increased?

mrks:
`8s 387ms` to `7s 706ms`.

Here is a generic one on TPC-H SF 10: `SELECT COUNT(*) FROM lineitem WHERE SUBSTR(l_shipmode, 1, 1) = 'T' OR SUBSTR(l_shipmode, 1, 1) = 'S' OR SUBSTR(l_shipmode, 1, 1) = 'R'`
`9 s 81 ms` to `6 s 304 ms` for the entire execution.

mrks:
Found some issues with `operator==` in the expressions. Most importantly, the ExistsExpression would always return true, making `EXISTS` and `NOT EXISTS` equal. Added what should be a comprehensive test for expression equality and hashes.

src/lib/expression/abstract_expression.hpp
mrzzzrm: > Are you concerned about the use_count being increased

Yep, that was my point. Having both a `const std::shared_ptr<AbstractExpression>& expression` as well as a `const std::shared_ptr<const AbstractExpression>& expression` would unequivocally avoid that, right?

mrks: done.

mrks:
@mrzzzrm Your comments should be addressed. I understand if you don't want to look at the expression changes - we can get the final review from someone else, but I don't want to overrun you.

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -151,45 +151,67 @@ ExpressionEvaluator::ExpressionEvaluator(
 template <typename Result>
 std::shared_ptr<ExpressionResult<Result>> ExpressionEvaluator::evaluate_expression_to_result(
     const AbstractExpression& expression) {
+  // First, look in the cache
+  const auto expression_ptr = expression.shared_from_this();
+  const auto cached_result_iter = _cached_expression_results.find(expression.shared_from_this());
```
mrzzzrm: y not reuse expression_ptr here?

src/lib/expression/pqp_subquery_expression.cpp
```diff
@@ -43,11 +43,20 @@ std::string PQPSubqueryExpression::as_column_name() const {
 }
 
 bool PQPSubqueryExpression::_shallow_equals(const AbstractExpression& expression) const {
-  Fail("Can't compare PQPSubqueryExpressions");
+  // For deep copies of this expression, the PQP is recreated. As such, the copy is considered unequal to its origin.
```
mrzzzrm: This is a bit cryptic. Prefix comment with "we're just comparing shared pointer addressees here, because it is unfeasable to deeply compare PQPs"?

src/test/expression/expression_test.cpp
```diff
@@ -276,4 +276,148 @@ TEST_F(ExpressionTest, IsNullable) {
...
+    if (!std::dynamic_pointer_cast<PQPSubqueryExpression>(first_expression)) {
+      deep_copy = first_expression->deep_copy();
+      EXPECT_EQ(*first_expression, *deep_copy);
+      EXPECT_EQ(first_expression->hash(), deep_copy->hash());
+    }
+
+    for (auto second_iter = first_iter + 1; second_iter != expressions.end(); ++second_iter) {
+      const auto& [second_line, second_expression] = *second_iter;
+      SCOPED_TRACE(std::string{"Second expression from line "} + std::to_string(second_line));
+      EXPECT_NE(*first_expression, *(second_expression));
```
mrzzzrm: need the extra parens?

src/test/expression/expression_test.cpp
```diff
@@ -276,4 +276,148 @@ TEST_F(ExpressionTest, IsNullable) {
...
+
+  for (auto first_iter = expressions.begin(); first_iter != expressions.end(); ++first_iter) {
+    const auto& [first_line, first_expression] = *first_iter;
+    SCOPED_TRACE(std::string{"First expression from line "} + std::to_string(first_line));
+
+    EXPECT_EQ(*first_expression, *first_expression);
+    EXPECT_EQ(first_expression->hash(), first_expression->hash());
+
+    std::shared_ptr<AbstractExpression> deep_copy;
+    if (!std::dynamic_pointer_cast<PQPSubqueryExpression>(first_expression)) {
```
mrzzzrm: one liner why PQPSubqueries dont need to fulfill this requirement?

src/test/expression/expression_test.cpp
```diff
@@ -276,4 +276,148 @@ TEST_F(ExpressionTest, IsNullable) {
...
+  expressions.emplace_back(__LINE__, unary_minus_(3));
+  expressions.emplace_back(__LINE__, unary_minus_(4));
+
+  // ValueExpression
+  expressions.emplace_back(__LINE__, value_(3));
+  expressions.emplace_back(__LINE__, value_(3.0f));
+  expressions.emplace_back(__LINE__, value_(3.25));
+  expressions.emplace_back(__LINE__, null_());
+
+  for (auto first_iter = expressions.begin(); first_iter != expressions.end(); ++first_iter) {
```
mrzzzrm: I dislike loops in tests and tests that do not yield a proper stacktrace, back to the input data, on failure. So I'd prefer if each `expressions.emplace_back(__LINE__, ...` line called a helper with the assertions and instead of a all-unequal-all-others comparison test you'd test just a few explicit cases. 

Just preferences.

src/test/sql/sql_translator_test.cpp
```diff
@@ -752,8 +752,8 @@ TEST_F(SQLTranslatorTest, WhereExists) {
 }
 
 TEST_F(SQLTranslatorTest, WhereNotExists) {
-  const auto actual_lqp =
-      compile_query("SELECT * FROM int_float WHERE EXISTS(SELECT * FROM int_float2 WHERE int_float.a = int_float2.a);");
+  const auto actual_lqp = compile_query(
+      "SELECT * FROM int_float WHERE NOT EXISTS(SELECT * FROM int_float2 WHERE int_float.a = int_float2.a);");
```
mrzzzrm: how come you could add the NOT without altering the `_exists()` below?

src/lib/expression/pqp_subquery_expression.cpp
mrks: "un-/infeasible" is cryptic itself. Is there a logical reason why one should not do it, a technical one, one that is related to the performance, or was the developer simply lazy? Suggestion:

```
  // For deep copies of this expression, the PQP is recreated. It is not only difficult to correctly compare PQPs, but
  // it also has unclear semantics. is an executed PQP equal to a non-executed one? How about two executed ones?
  // As such, we only report equality if two PQPSubqueryExpression refer to the very same PQP, meaning that a deep copy
  // is not equal to its source. If this ever becomes an issue, it should be easy to spot.
```

mrzzzrm: much bedda

src/test/expression/expression_test.cpp
mrks: Probably not, I'll try it without them

src/test/sql/sql_translator_test.cpp
mrks: The question should be "how come it did not fail before?" :D
https://github.com/hyrise/hyrise/pull/1781/files#diff-fba0ebca5459aafe9c003ae4c3640a4dL36

src/test/expression/expression_test.cpp
mrks: Have you seen the scoped_trace? I consider that better than helper methods:


```
../src/test/expression/expression_test.cpp:408: Failure
Expected equality of these values:
  *first_expression
    Which is: SUBQUERY (PQP, 0x11169bc38)
  *deep_copy
    Which is: SUBQUERY (PQP, 0x11169c098)
Google Test trace:
../src/test/expression/expression_test.cpp:399: First expression from line 380
../src/test/expression/expression_test.cpp:408: Failure
```

mrks: Pinging @mrzzzrm because I'm interested to hear if your loop aversion includes loops that have those traces

mrzzzrm: Yes, saw that. Makes is it a bit better, but not good. 

A failing test should be as easy as possible to understand by even a non-author. The best the test-author can do for this, imo, is make sure that all code/data involved in the test is in the stacktrace when the failing test is run in the debugger. If we make sure that's the case then any test is follows the same pattern and is "familiar" even to non-authors. If every test follows it's own pattern and requires the non-author to parse the log for line X and Y, then to parse the entire test function (which might well be the thing that's broken!), this is a significant burden. 
(There is a fascinating extent to which people complicate their tests that makes it very hard to figure out what is actually broken (test or code?) and what is actually being tested in a test case. I'm just trying to avoid stepping into that territory, here. Keeping things simple paranoidly is imo just good enough.)

mrzzzrm: But please take this as a philosophical rant about how unit tests should be written imho.

mrks: I agree with the general goal of making test failures easy to understand. At the same time, it should be easy to check that tests are complete (i.e., cover each variation in the input data). For me, it is easy to see that this list covers all EXIST cases:

```
  expressions.emplace_back(__LINE__, exists_(lqp_subquery_(int_float_node)));
  expressions.emplace_back(__LINE__, not_exists_(lqp_subquery_(int_float_node)));
  expressions.emplace_back(__LINE__, not_exists_(lqp_subquery_(int_float_node_nullable)));
```

A test that does the same without loops could look like this:

```
TEST_F(ExpressionTest, EqualsAndHash2) {
  const auto test_reflection = [](const auto& expression) {
    const auto deep_copy = expression->deep_copy();
    EXPECT_EQ(*expression, *deep_copy);
    EXPECT_EQ(expression->hash(), deep_copy->hash());
  };

  test_reflection(exists_(lqp_subquery_(int_float_node)));
  test_reflection(not_exists_(lqp_subquery_(int_float_node)));
  test_reflection(not_exists_(lqp_subquery_(int_float_node_nullable)));

  EXPECT_NE(*exists_(lqp_subquery_(int_float_node)), *not_exists_(lqp_subquery_(int_float_node)));
  EXPECT_NE(*exists_(lqp_subquery_(int_float_node)), *not_exists_(lqp_subquery_(int_float_node_nullable)));
}
```

To avoid adding the expressions twice, I could build a helper that first checks the reflective equality and the hash of the first argument and then compares the first with the second argument:

```
TEST_F(ExpressionTest, EqualsAndHash2) {
  const auto test = [](const auto& left, const auto& right) {
    const auto deep_copy = left->deep_copy();
    EXPECT_EQ(*left, *deep_copy);
    EXPECT_EQ(left->hash(), deep_copy->hash());

    EXPECT_NE(*left, *right);
  };

  test(*exists_(lqp_subquery_(int_float_node)), *not_exists_(lqp_subquery_(int_float_node)));
  test(*exists_(lqp_subquery_(int_float_node)), *not_exists_(lqp_subquery_(int_float_node_nullable)));
}
```

This makes it hard to see that `not_exists_` is never checked against its deep copy. Also, if I undo the fix for EqualsExpression, the error message looks like this:

```
[ RUN      ] ExpressionTest.EqualsAndHash2
../src/test/expression/expression_test.cpp:430: Failure
Expected: (*left) != (*right), actual: 64-byte object <D0-60 73-04 01-00 00-00 18-83 8E-0C 01-00 00-00 00-83 8E-0C 01-00 00-00 07-00 00-00 4E-75 6C-6C 80-01 BE-0C 01-00 00-00 90-01 BE-0C 01-00 00-00 90-01 BE-0C 01-00 00-00 00-00 00-00 69-67 6E-65> vs 64-byte object <D0-60 73-04 01-00 00-00 D8-D6 8C-0C 01-00 00-00 C0-D6 8C-0C 01-00 00-00 07-00 00-00 6E-6B 4F-66 90-01 BE-0C 01-00 00-00 A0-01 BE-0C 01-00 00-00 A0-01 BE-0C 01-00 00-00 01-00 00-00 69-67 6E-65>
```

That's hardly better. Even if we got gtest to print the column name, it would not be perfect, as it does not give us the absolute information on how the expression was created.

So while I agree with you that generated data and loops make it hard for the reader to understand what's going wrong (OperatorsIndexScanTest.SingleColumnScanOnDataTable, only to name a randomly picked one), I wouldn't consider it a black and white issue. In this case, my tendency is to stick to the version in the PR.

mrzzzrm: > I wouldn't consider it a black and white issue. In this case, my tendency is to stick to the version in the PR.

Yes, and alright.

mrzzzrm: (But it might be worth trying to figure out why googletest doesn't pretty-print these expressions so you could get rid of the weird capturing of __LINE__ there...)

src/lib/expression/pqp_column_expression.cpp
```diff
@@ -36,12 +36,14 @@ DataType PQPColumnExpression::data_type() const { return _data_type; }
 bool PQPColumnExpression::requires_computation() const { return false; }
 
 bool PQPColumnExpression::_shallow_equals(const AbstractExpression& expression) const {
-  const auto* pqp_column_expression = dynamic_cast<const PQPColumnExpression*>(&expression);
-  if (!pqp_column_expression) return false;
-  return column_id == pqp_column_expression->column_id;
+  DebugAssert(dynamic_cast<const PQPColumnExpression*>(&expression),
+              "Different expression type should have been caught by AbstractExpression::operator==");
+  const auto& pqp_column_expression = static_cast<const PQPColumnExpression&>(expression);
+  return column_id == pqp_column_expression.column_id && _data_type == pqp_column_expression._data_type &&
```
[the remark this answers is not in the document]
mrks: Who said it was? ;)

I can't come up with a case where deep equals is true, the column ids are the same, but the data types differ. In any case, I simply made sure that all members are part of the equality check.

src/test/expression/expression_test.cpp
mrks: > But it might be worth trying to figure out why googletest doesn't pretty-print these expressions

That is because `auto` in the test is not an AbstractExpression but an ExistsExpression. No idea why googletest doesn't use `operator<<` of the parent class while a normal print does.

> so you could get rid of the weird capturing of `__LINE__` there...

I'd still leave `__LINE__` in, as `as_column_name` is incomplete:

```
../src/test/expression/expression_test.cpp:415: Failure
Expected equality of these values:
  *first_expression
    Which is: a + b
  *second_expression
    Which is: a + b
Google Test trace:
../src/test/expression/expression_test.cpp:414: Second expression from line 374
../src/test/expression/expression_test.cpp:399: First expression from line 373
../src/test/expression/expression_test.cpp:415: Failure
```

for 

```
  expressions.emplace_back(__LINE__, pqp_column_(ColumnID{0}, DataType::Float, false, "a + b"));
  expressions.emplace_back(__LINE__, pqp_column_(ColumnID{1}, DataType::Float, false, "a + b"));
```

PR #1784, opened 2019-08-09
Fix sqlite comparison

### LINKED ISSUE #1409: hyriseBenchmarkJoinOrder --verify fails

mrks:
By default, LIKE is case-insensitive in sqlite. This PR changes that behavior to match Hyrise's.

Also, it fixes the comparison of all-NULL columns.

There is still a mismatch in the table sizes between the imdb_data set that we are using and the one linked in the benchmark. That leads to some results being NULL as described in #1409. I don't see this as a big issue, as we are only using the JOB for internal comparisons. As a result, I suggest that this closes #1409.

### DIFF
--- a/src/lib/utils/check_table_equal.cpp
+++ b/src/lib/utils/check_table_equal.cpp
@@ -196,16 +196,14 @@ std::optional<std::string> check_table_equal(const std::shared_ptr<const Table>&
         expected_column_type = DataType::Int;
       }
 
-      if (expected_column_type == DataType::Int) {
-        bool all_null = true;
-        for (auto row_id = size_t{HEADER_SIZE}; row_id < expected_matrix.size(); row_id++) {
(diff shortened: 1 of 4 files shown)


src/lib/utils/check_table_equal.cpp
```diff
@@ -164,62 +164,74 @@ std::optional<std::string> check_table_equal(const std::shared_ptr<const Table>&
...
-        right_column_type = DataType::Float;
-      } else if (right_column_type == DataType::Long) {
-        right_column_type = DataType::Int;
+      if (expected_column_type == DataType::Double) {
+        expected_column_type = DataType::Float;
+      } else if (expected_column_type == DataType::Long) {
+        expected_column_type = DataType::Int;
+      }
+
+      if (expected_column_type == DataType::Int) {
```
mrks: This is new, too.

PR #1785, opened 2019-08-12
Add warning about unused index

### LINKED ISSUE #1700: Ensure that chunks added after LQP translation are handled

mrks:
fixes #1700

### DIFF
--- a/src/lib/logical_query_plan/lqp_translator.cpp
+++ b/src/lib/logical_query_plan/lqp_translator.cpp
@@ -207,8 +207,12 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_predicate_node_to_in
   const auto chunk_count = table->chunk_count();
   for (ChunkID chunk_id{0u}; chunk_id < chunk_count; ++chunk_id) {
     const auto chunk = table->get_chunk(chunk_id);
-    if (chunk && chunk->get_index(SegmentIndexType::GroupKey, column_ids)) {
-      indexed_chunks.emplace_back(chunk_id);
+    if (chunk && !chunk->get_indexes(column_ids).empty()) {
+      if (chunk->get_index(SegmentIndexType::GroupKey, column_ids)) {
+        indexed_chunks.emplace_back(chunk_id);
+      } else {
+        PerformanceWarning("Found index, but LQPTranslator currently only supports GroupKeyIndexes");
+      }
     }
   }
 


src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -207,8 +207,12 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_predicate_node_to_in
   const auto chunk_count = table->chunk_count();
   for (ChunkID chunk_id{0u}; chunk_id < chunk_count; ++chunk_id) {
     const auto chunk = table->get_chunk(chunk_id);
-    if (chunk && chunk->get_index(SegmentIndexType::GroupKey, column_ids)) {
-      indexed_chunks.emplace_back(chunk_id);
+    if (chunk && !chunk->get_indexes(column_ids).empty()) {
```
mweisgut: What's the purpose of checking if the result of `get_indexes(…)` is empty in addition to checking if the result of `get_index(…)` is not nullopt? Calling get_indexes results in copying indexes.

https://github.com/hyrise/hyrise/blob/f452c8c71ebb200e9b1b224195e0e669ba279442/src/lib/storage/chunk.cpp#L97-L123

mrks: > What's the purpose of checking if the result of get_indexes(…) is empty in addition to checking if the result of get_index(…) is not nullopt?

`get_index` is only called with `SegmentIndexType::GroupKey`. Here, we want to explicitly check if there is any index that is not a group-key index. 

> Calling get_indexes results in copying indexes.

Now I am a bit nauseous. If it does work, it should not be called `get`.

Suggestion:
* Add `has_index()`, which does what we need here
* Rename `get_indexes` to `collect_indexes`

mrks: Thanks for pointing this out - that would have been unnecessarily expensive.

mrks: The more I look at this, the less I like it. Why does the LQPTranslator check the indexes of the table in the first place? This being the only place in the translator where `get_chunk` is called is a first sign that something is amiss. Shouldn't it be the job of the IndexScanRule to decide *which* index is used, not to just leave a mysterious "Use an index, but I won't tell you which" in the node?

mrks:
Superseded by #1790

PR #1788, opened 2019-08-13
Meta tables

mrks:
* Introduces meta_(tables|columns|chunks|segments).
* Retires Show(Columns|Tables)Node, which I never really liked. Instead, the SQLTranslator maps those special SQL commands to simple selects on the meta tables.

Example (slightly truncated):
```
(debug)> load resources/test_data/tbl/int.tbl t
(debug)> load resources/test_data/tbl/int.tbl t2 Dictionary
(debug)> SELECT * FROM meta_segments
|   table|chunk_id|column_id|column_name|column_type|  encoding|
|       t|       0|        0|          a|        int|      NULL|
|      t2|       0|        0|          a|        int|Dictionary|
```

Open points for discussion:
* Currently, we have a lot of name duplication. The name "meta_chunks" appears in one form or the other in the method signature and in table names, both in the code and in the tests. I am not sure if that is really an issue though.
* We also duplicate some information in the tables. Showing the column name in meta_segments is redundant, as it could be retrieved by joining with meta_columns. Do we want to keep it for ease of use?


Bouncner:
Everytime I work with such meta data, I usually end up joining a lot of tables to get such a flat view. Hence, while I don't think it's clean I am still voting for it.
Another commercial IMDB also does it that way. So we wouldn't be the only database to do that. And who is Codd anyways?

I work a lot with wordpress a couple of years ago, which has a couple of `meta` tables. Since we don't have schemas in Hyrise, maybe use a prefix like `__meta_` instead of just `meta_`? Could also be `$` if we agree that PHP is more awesome than Python.

Haven't looked at the code yet and will probably not do it for a couple of days.

resources/test_data/tbl/meta_tables/meta_columns.tbl
```diff
@@ -0,0 +1,7 @@
+table|name|type|nullable
```
Bensk1: I am wondering whether we should have column_id and data_type in meta_columns as well.

resources/test_data/tbl/meta_tables/meta_segments.tbl
```diff
@@ -0,0 +1,9 @@
+table|chunk_id|column_id|column_name|column_type|encoding
```
Bensk1: `data_type` instead of `column_type`?

src/lib/sql/sql_translator.cpp
```diff
@@ -304,7 +303,9 @@ std::shared_ptr<AbstractExpression> SQLTranslator::translate_hsql_expr(const hsq
 
 std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_insert(const hsql::InsertStatement& insert) {
   const auto table_name = std::string{insert.tableName};
+  AssertInput(!table_name.starts_with(MetaTableManager::META_PREFIX), "Cannot update meta tables");
```
Bensk1: I am fine with having the same error message for update, insert, and delete. Just double checking if it was intended or just copy-paste-forget?

src/lib/sql/sql_translator.cpp
```diff
@@ -1054,9 +1058,16 @@ void SQLTranslator::_translate_limit(const hsql::LimitDescription& limit) {
 std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_show(const hsql::ShowStatement& show_statement) {
   switch (show_statement.type) {
     case hsql::ShowType::kShowTables:
-      return ShowTablesNode::make();
-    case hsql::ShowType::kShowColumns:
-      return ShowColumnsNode::make(std::string(show_statement.name));
+      return StoredTableNode::make(std::string{MetaTableManager::META_PREFIX} + "tables");
+    case hsql::ShowType::kShowColumns: {
+      // TODO This hard-codes some information like the meta table name and that it stores the documented table's name
```
Bensk1: I think it should be fine to make that kind of decisions for meta tables.

src/lib/storage/storage_manager.cpp
```diff
@@ -34,13 +37,23 @@ void StorageManager::drop_table(const std::string& name) {
 }
 
 std::shared_ptr<Table> StorageManager::get_table(const std::string& name) const {
+  if (name.starts_with(MetaTableManager::META_PREFIX)) {
+    return MetaTableManager::get().generate_table(name.substr(5));
```
Bensk1: `name.substr(MetaTableManager::META_PREFIX.size())`? If yes, then see below as well.

src/lib/utils/meta_table_manager.cpp
```diff
@@ -0,0 +1,119 @@
...
+std::shared_ptr<Table> MetaTableManager::generate_table(const std::string& table_name) const {
+  return _methods.at(table_name)();
+}
+
+std::shared_ptr<Table> MetaTableManager::generate_tables_table() const {
+  const auto columns = TableColumnDefinitions{{"table", DataType::String, false},
+                                              {"column_count", DataType::Int, false},
+                                              {"row_count", DataType::Long, false},
+                                              {"chunk_count", DataType::Int, false},
+                                              {"max_chunk_size", DataType::Long, false}};
```
Bensk1: Since ChunkID is a `uint32_t`, could we make that `DataType::Int`?

src/test/utils/meta_table_manager_test.cpp
```diff
@@ -0,0 +1,53 @@
...
+class MetaTableManagerTest : public BaseTest {};
+
+TEST_F(MetaTableManagerTest, TableBasedMetaData) {
+  // This tests a bunch of meta tables that are somehow related to the tables stored in the StorageManager.
+  auto& storage_manager = StorageManager::get();
+
+  const auto prefix = std::string{MetaTableManager::META_PREFIX};
+  const auto path = std::string{"resources/test_data/tbl/meta_tables/meta_"};
+  for (const auto& meta_table_name : {"tables", "columns", "chunks", "segments"}) {
+    SCOPED_TRACE(meta_table_name);
```
Bensk1: I haven't seen this before, what is it?

src/lib/utils/meta_table_manager.cpp
```diff
@@ -0,0 +1,119 @@
...
+                            static_cast<int32_t>(table->column_is_nullable(column_id))});
+    }
+  }
+
+  return output_table;
+}
+
+std::shared_ptr<Table> MetaTableManager::generate_chunks_table() const {
+  const auto columns = TableColumnDefinitions{{"table", DataType::String, false},
+                                              {"chunk_id", DataType::Int, false},
```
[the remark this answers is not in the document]
Bensk1: I think I added that one last Wednesday, consider it outdated.

Bensk1:
I am also fine with duplicate information in the meta tables and I don't have a strong opinion on the exact meta prefix.

resources/test_data/tbl/meta_tables/meta_columns.tbl
mrks: Yes to column_id, but data_type is already there?!

resources/test_data/tbl/meta_tables/meta_segments.tbl
mrks: I like prefixing all the (duplicated) column information accordingly.

src/lib/sql/sql_translator.cpp
mrks: I took "update" as a synonym for "modify". Will change to the latter.

src/lib/storage/storage_manager.cpp
mrks: Oh, good point.

src/test/utils/meta_table_manager_test.cpp
mrks: If you have a test with expectations that are part of a loop, it gives you more context:

https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#adding-traces-to-assertions

resources/test_data/tbl/meta_tables/meta_columns.tbl
Bensk1: Right. Should we call it `data_type` instead of `type`?

resources/test_data/tbl/meta_tables/meta_segments.tbl
Bensk1: `column_data_type`?

src/test/utils/meta_table_manager_test.cpp
Bensk1: Ah, I see.

resources/test_data/tbl/meta_tables/meta_columns.tbl
mrks: I don't like it, but I don't care enough to discuss it.

mrks:
> maybe use a prefix like __meta_ instead of just meta_

The parser doesn't like that. Not sure how the SQL standard feels about it. I wouldn't touch it before it actually becomes an issue.

PR #1789, opened 2019-08-14
Disable TPC-DS Q85

### LINKED ISSUE #1782: TPC-DS Q24/Q85 fails / Test binary
```
- Benchmarking 85
  what():  src/lib/statistics/statistics_objects/abstract_histogram.cpp:621 Cannot split_at_bin_bounds() on string histogram
Aborted
```

**The TPC-DS should also become part of the binary tests.** If generation takes too long in CI, we should discuss ways of storing the table data outside the container.

mweisgut:
refs #1782

### DIFF
--- a/resources/benchmark/tpcds/query_blacklist.cfg
+++ b/resources/benchmark/tpcds/query_blacklist.cfg
@@ -86,7 +86,7 @@
 82.sql
 83.sql
 84.sql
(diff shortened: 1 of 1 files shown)


mrks:
Does not fix the issue, we are still missing the tests.

PR #1791, opened 2019-08-14
Reformulate OR to UNION

### LINKED ISSUE #1813: Resolve conflict between BetweenCompositionRule and DisjunctionToUnionRule
If the optimizer applies the `DisjunctionToUnionRule` before the BetweenCompositionRule, the execution of TPC-H 7 crashes with a segfault.

The actual problem is that the `BetweenCompositionRule` cannot handle expressions like `a = 1 AND b = 2` (equations in a conjunction). The crash happened because the `DisjunctionToUnionRule` created two of those expressions for TPC-H 7.

janetzki:
- Close hyrise-mp/hyrise/issues/51
- Fixes #1813 

- Extend **PredicateSplitUpRule**
  - Purpose: Split not only conjunctions (`AND`) but also disjunctions (`OR`) into PredicateNodes and UnionNodes
  - Reason: Other optimizer rules, such as the `SubqueryToJoinRule`, can run only on unnested predicates
- Introduce **PredicateMergeRule**
  - Purpose: Revert the changes of the `PredicateSplitUpRule` in some cases
  - Reason: For large predicate expressions (such as in TPC-DS 41), the `PredicateSplitUpRule` generates LQP subplans that are more expensive than having the ExpressionEvaluator run on the original monolithic expression. These LQPs are only merged if
    - no other optimizer rule benefited from them,
    - they contain at least 4 UnionNodes.
- Change **PredicateReorderingRule**
  - Purpose: Recognize optimizable predicate chains one node earlier
- Remove **JoinGraphBuilder::_parse_union()**
  - Reason: Code relic from when we did not handle disjunctions

# Benchmarks on Sidon
- compared to dd18d60344228a3312cb32537839cbc704416841
- single-threaded
## TPC-DS
```
+----------------+--------------------+------+-------------------+------+------------+---------------------------------+
| Benchmark      | prev. iter/s       | runs | new iter/s        | runs | change [%] | p-value (significant if <0.001) |
+----------------+--------------------+------+-------------------+------+------------+---------------------------------+
| 01             | 14.64333152771     | 879  | 14.7361736297607  | 885  | +1%        |                          0.0000 |
| 03             | 11.9376211166382   | 717  | 11.4723844528198  | 689  | -4%        |                          0.0000 |
| 06             | 10.3059186935425   | 619  | 10.175968170166   | 611  | -1%        |                          0.0043 |
| 07             | 7.0083703994751    | 421  | 7.39057731628418  | 444  | +5%        |                          0.0000 |
| 09             | 2.55413293838501   | 154  | 2.55693817138672  | 154  | +0%        |                          0.0424 |
| 10             | 4.75967741012573   | 286  | 5.03399467468262  | 303  | +6%        |                          0.0000 |
| 13             | 3.13416266441345   | 189  | 4.80863523483276  | 289  | +53%       |                          0.0000 |
| 15             | 18.6201267242432   | 1118 | 20.3155822753906  | 1219 | +9%        |                          0.0000 |
| 17             | 7.13787698745728   | 429  | 6.60065221786499  | 397  | -8%        |                          0.0000 |
| 19             | 9.37815284729004   | 563  | 9.82807540893555  | 590  | +5%        |                          0.0000 |
| 25             | 8.74565887451172   | 525  | 8.0463342666626   | 483  | -8%        |                          0.0000 |
| 26             | 14.8294544219971   | 890  | 15.1212739944458  | 908  | +2%        |                          0.0000 |
| 28             | 4.98067569732666   | 299  | 6.62102699279785  | 398  | +33%       |                          0.0000 |
| 29             | 8.68302631378174   | 521  | 8.3484411239624   | 502  | -4%        |                          0.0000 |
| 31             | 0.787197053432465  | 48   | 0.788376867771149 | 48   | +0%        |                          0.7364 |
| 34             | 9.16647052764893   | 550  | 9.21521854400635  | 553  | +1%        |                          0.1578 |
| 35             | 0.0749651566147804 | 5    | 3.43731117248535  | 207  | +4485%     |        (not enough runs) 0.0000 |
| 39a            | 0.413204073905945  | 25   | 0.418115675449371 | 26   | +1%        |                          0.0439 |
| 39b            | 0.417512655258179  | 26   | 0.417701035737991 | 26   | +0%        |                          0.8951 |
| 41             | 26.9418048858643   | 1617 | 26.6064319610596  | 1597 | -1%        |                          0.0000 |
| 42             | 10.8549146652222   | 652  | 10.2512683868408  | 616  | -6%        |                          0.0000 |
| 43             | 1.93547511100769   | 117  | 1.99247670173645  | 120  | +3%        |                          0.0000 |
| 45             | 27.3844528198242   | 1644 | 23.7936096191406  | 1428 | -13%       |                          0.0000 |
| 48             | 2.93025875091553   | 176  | 4.49252891540527  | 270  | +53%       |                          0.0000 |
| 50             | 11.5422496795654   | 693  | 12.9528903961182  | 778  | +12%       |                          0.0000 |
| 52             | 9.85784816741943   | 592  | 9.86636161804199  | 592  | +0%        |                          0.6956 |
| 55             | 10.8998289108276   | 654  | 10.9666376113892  | 659  | +1%        |                          0.2107 |
| 62             | 6.3749737739563    | 383  | 6.39965009689331  | 384  | +0%        |                          0.0002 |
| 65             | 5.30449342727661   | 319  | 5.67645215988159  | 341  | +7%        |                          0.0000 |
| 69             | 5.13577365875244   | 309  | 5.23223733901978  | 314  | +2%        |                          0.0000 |
| 73             | 12.3163614273071   | 739  | 12.753191947937   | 766  | +4%        |                          0.0000 |
| 79             | 4.01172208786011   | 241  | 4.33155870437622  | 260  | +8%        |                          0.0000 |
| 81             | 24.3882102966309   | 1464 | 24.430025100708   | 1466 | +0%        |                          0.0014 |
| 83             | 35.8681716918945   | 2153 | 35.93603515625    | 2157 | +0%        |                          0.0000 |
| 88             | 6.247230052948     | 375  | 6.21600818634033  | 374  | -0%        |                          0.2620 |
| 91             | 3.375483751297     | 203  | 20.5214519500732  | 1232 | +508%      |                          0.0000 |
| 93             | 2.41186714172363   | 145  | 2.40117478370667  | 145  | -0%        |                          0.0156 |
| 96             | 12.0982332229614   | 726  | 11.821587562561   | 710  | -2%        |                          0.0005 |
| 97             | 1.37546062469482   | 83   | 1.40054154396057  | 85   | +2%        |                          0.0002 |
| 99             | 3.07890439033508   | 185  | 3.174880027771    | 191  | +3%        |                          0.0000 |
| geometric mean |                    |      |                   |      | +19%       |                                 |
+----------------+--------------------+------+-------------------+------+------------+---------------------------------+
```
**Recent changes**
- **Query 35** used to show a performance increase of 20000%. It reduced to 4000% because the query became faster on the master.
- **Query 41** used to suffer a lot because it contains a depply nested expression with many `ANDs` and `ORs`. The unnesting in the `PredicateSplitUpRule` slowed down subsequent optimizer rules that called `apply_to()`, which has an exponential runtime. They now use `visit_lqp()`, which brought the query back to its former performance.
- **Query 45** runs slower because the `PredicateSplitUpRule` allows the `SubqueryToJoinRule` to insert an unnecessary "Reduction Semi Join". I added a comment to adjust the corresponding heuristic.

## TPC-H
```
+----------------+-------------------+------+-------------------+------+------------+---------------------------------+
| Benchmark      | prev. iter/s      | runs | new iter/s        | runs | change [%] | p-value (significant if <0.001) |
+----------------+-------------------+------+-------------------+------+------------+---------------------------------+
| TPC-H 01       | 0.411292910575867 | 25   | 0.410981088876724 | 25   | -0%        |                          0.9101 |
| TPC-H 02       | 9.24369621276855  | 555  | 9.43016052246094  | 566  | +2%        |                          0.0000 |
| TPC-H 03       | 4.50368309020996  | 271  | 4.6011004447937   | 277  | +2%        |                          0.0000 |
| TPC-H 04       | 4.94974946975708  | 297  | 5.34366941452026  | 321  | +8%        |                          0.0000 |
| TPC-H 05       | 2.94237899780273  | 177  | 2.85773730278015  | 172  | -3%        |                          0.0000 |
| TPC-H 06       | 67.0984878540039  | 4026 | 67.1580963134766  | 4030 | +0%        |                          0.7207 |
| TPC-H 07       | 1.00613844394684  | 61   | 1.55762565135956  | 94   | +55%       |                          0.0000 |
| TPC-H 08       | 3.24828314781189  | 195  | 3.31488060951233  | 199  | +2%        |                          0.0000 |
| TPC-H 09       | 1.08538055419922  | 66   | 1.07314026355743  | 65   | -1%        |                          0.0000 |
| TPC-H 10       | 2.14559745788574  | 129  | 2.10868406295776  | 127  | -2%        |                          0.0470 |
| TPC-H 11       | 18.3144416809082  | 1099 | 18.3226947784424  | 1100 | +0%        |                          0.7593 |
| TPC-H 12       | 5.20405054092407  | 313  | 5.20568895339966  | 313  | +0%        |                          0.6643 |
| TPC-H 13       | 1.60477733612061  | 97   | 1.57868003845215  | 95   | -2%        |                          0.0000 |
| TPC-H 14       | 32.0659370422363  | 1924 | 31.506477355957   | 1891 | -2%        |                          0.0000 |
| TPC-H 15       | 46.6904411315918  | 2802 | 46.1080131530762  | 2767 | -1%        |                          0.0000 |
| TPC-H 16       | 4.88137054443359  | 293  | 5.08951616287231  | 306  | +4%        |                          0.0000 |
| TPC-H 17       | 0.946663022041321 | 57   | 0.940404772758484 | 57   | -1%        |                          0.0002 |
| TPC-H 18       | 0.60135817527771  | 37   | 0.601025998592377 | 37   | -0%        |                          0.8322 |
| TPC-H 19       | 5.36013460159302  | 322  | 11.2822532653809  | 677  | +110%      |                          0.0000 |
| TPC-H 20       | 1.89766144752502  | 114  | 1.95619964599609  | 118  | +3%        |                          0.0000 |
| TPC-H 21       | 1.02758741378784  | 62   | 1.02957963943481  | 62   | +0%        |                          0.5805 |
| TPC-H 22       | 7.98162221908569  | 479  | 8.15192413330078  | 490  | +2%        |                          0.0000 |
| geometric mean |                   |      |                   |      | +6%        |                                 |
+----------------+-------------------+------+-------------------+------+------------+---------------------------------+
```


mrks:
Does this apply to TPC-H 15 by chance?

mrks:
In any case: nice numbers!

janetzki:
> Does this apply to TPC-H 15 by chance?

Do you mean TPC-H 19? I work on that one, too. It is more complicated than TPC-DS 35 because it has two ORs in the same PredicateNode instead of just one.
This rule cannot optimize TPC-H 15 because it contains no OR.

src/lib/optimizer/strategy/or_to_union_rule.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "logical_query_plan/predicate_node.hpp"
+#include "logical_query_plan/union_node.hpp"
+#include "or_to_union_rule.hpp"
+
+namespace opossum {
+
+std::string OrToUnionRule::name() const { return "Or to Union Rule"; }
+
+void OrToUnionRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node) const {
+  if (node->type == LQPNodeType::Predicate) {
```
mrzzzrm: Use visit_lqp to avoid redundant traversal of subplans under diamond shapes

mrzzzrm: (when using visit_lqp you need to only collect the nodes to be replaced in the lambda and then to the replacing outside of visit_lqp)

src/lib/optimizer/strategy/or_to_union_rule.cpp
```diff
@@ -0,0 +1,34 @@
...
+    const auto predicate_node = std::static_pointer_cast<PredicateNode>(node);
+    const auto predicate = predicate_node->predicate();
+
+    const auto flat_disjunction = flatten_logical_expressions(predicate, LogicalOperator::Or);
+    if (flat_disjunction.size() == 2) {
```
mrzzzrm: you shouldn't have to limit this to 2, disjunctive chains of any size >= 2 can be reformulated into a series of unions

src/test/optimizer/strategy/or_to_union_rule_test.cpp
```diff
@@ -0,0 +1,223 @@
...
+    UnionNode::make(UnionMode::Positions,
+      PredicateNode::make(value_(1),
+        node_a),
+      PredicateNode::make(greater_than_(value_(3), value_(2)),
+        node_a)));
+  // clang-format on
+
+  const auto actual_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
+
+  std::cout << "INPUT\n" << *input_lqp << "\n\n";
```
mrzzzrm: yo ;)

src/test/optimizer/strategy/or_to_union_rule_test.cpp
```diff
@@ -0,0 +1,223 @@
...
+    node_b = MockNode::make(MockNode::ColumnDefinitions{{DataType::Int, "a"}, {DataType::Int, "b"}}, "b");
+    b_a = node_b->get_column("a");
+    b_b = node_b->get_column("b");
+
+    node_c = MockNode::make(
+        MockNode::ColumnDefinitions{{DataType::Int, "a"}, {DataType::Int, "b"}, {DataType::Int, "c"}}, "c");
+    c_a = node_c->get_column("a");
+    c_b = node_c->get_column("b");
+
+    customer_demographics =
```
mrzzzrm: adding "real world tables" here just makes the unit test needlessly complicated imo. Identify  the complicated aspects of TPCDS35 and create properly named tests from them.

The OptimizeTPCDS35 tests is a lot of code, but in fact it doesn't exercise the code more than TwoExistsToUnion.

src/test/optimizer/strategy/or_to_union_rule_test.cpp
```diff
@@ -0,0 +1,223 @@
...
+        MockNode::make(MockNode::ColumnDefinitions{{DataType::Int, "c_customer_sk"}}, "customer_demographics");
```
mrzzzrm: Instead of OptimizeTPCDS35, how about testing an LQP where the PredicateNode has multiple outputs?

src/lib/optimizer/strategy/or_to_union_rule.hpp
```diff
@@ -0,0 +1,16 @@
+#pragma once
+
+#include "abstract_rule.hpp"
+#include "expression/abstract_expression.hpp"
+#include "types.hpp"
+
+namespace opossum {
+
+class OrToUnionRule : public AbstractRule {
```
mrzzzrm: class doc missing.

src/lib/optimizer/strategy/or_to_union_rule.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "logical_query_plan/lqp_utils.hpp"
+#include "logical_query_plan/predicate_node.hpp"
+#include "logical_query_plan/union_node.hpp"
+#include "or_to_union_rule.hpp"
+
+namespace opossum {
+
+std::string OrToUnionRule::name() const { return "Or to Union Rule"; }
+
+void OrToUnionRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node) const {
```
mrks: How do you feel about "disjunction_to_union_rule"?

src/lib/logical_query_plan/union_node.cpp
```diff
@@ -24,14 +24,7 @@ const std::vector<std::shared_ptr<AbstractExpression>>& UnionNode::column_expres
...
 
-  const auto left_input_column_count = left_input()->column_expressions().size();
-  if (static_cast<size_t>(column_id) >= left_input_column_count) {
-    return left_input()->is_column_nullable(column_id);
-  } else {
-    const auto right_column_id =
-        static_cast<ColumnID>(column_id - static_cast<ColumnID::base_type>(left_input_column_count));
-    return left_input()->is_column_nullable(right_column_id);
-  }
+  return left_input()->is_column_nullable(column_id) || right_input()->is_column_nullable(column_id);
```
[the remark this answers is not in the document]
mrks: Not sure if I agree. `SELECT * FROM t1 UNION ALL SELECT * FROM t2;`, which we currently do not support, could get a nullable column from t1 and a non-nullable column from t2:

http://sqlfiddle.com/#!9/72ac12/10

src/lib/optimizer/optimizer.cpp
```diff
@@ -105,6 +106,8 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {
   // for now we want the PredicateReorderingRule to have the final say on predicate positions
   optimizer->add_rule(std::make_unique<PredicatePlacementRule>());
 
+  optimizer->add_rule(std::make_unique<OrToUnionRule>());
```
mrks: Did you consider different positions for this rule?

src/lib/optimizer/strategy/or_to_union_rule.cpp
```diff
@@ -0,0 +1,34 @@
+#include <expression/expression_utils.hpp>
+
+#include "expression/logical_expression.hpp"
+#include "logical_query_plan/lqp_utils.hpp"
+#include "logical_query_plan/predicate_node.hpp"
+#include "logical_query_plan/union_node.hpp"
+#include "or_to_union_rule.hpp"
```
mrks: I thought the linter wouldn't let you do this. The corresponding hpp should be at the very top.

src/lib/optimizer/strategy/or_to_union_rule.cpp
mrks: You could also consider moving this to the top level of `ExpressionReductionRule::apply_to` where we are already visiting the LQP.

src/test/optimizer/strategy/or_to_union_rule_test.cpp
```diff
@@ -0,0 +1,219 @@
...
+    PredicateNode::make(exists_(subquery_b),
+      node_a));
+  // clang-format on
+
+  const auto actual_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
+
+  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+}
+
+TEST_F(OrToUnionRuleTest, NoRewriteThreeExistsToUnion) {
```
mrks: What are we exactly testing here? We should test that (a) the new code does what it advertises to do and (b) that it does not break anything, neither now, nor in what we can predict for the future. We should not test that it does not do more than advertised - that just leads to confusion when the feature is implemented.

src/lib/logical_query_plan/union_node.cpp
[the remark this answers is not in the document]
mrzzzrm: kk didn't know that was legal

src/lib/optimizer/strategy/or_to_union_rule.cpp
mrzzzrm: ExpressionReductionRule is for Expressions that get rewritten into expressions, not for expressions that are rewritten into operators imo

src/test/optimizer/strategy/or_to_union_rule_test.cpp
janetzki: Do you mean a diamond structure? Good idea, this would be an interesting test case.

src/lib/optimizer/strategy/or_to_union_rule.cpp
janetzki: @mrzzzrm is right. The ExpressionReductionRule does not change the nodes of the LQP.
(Another aspect is that the DisjunctionToUnionRule is the OR equivalent of the PredicateSplitUpRule (AND). If one rule becomes part of the ExpressionReductionRule, the other one should as well.)

src/lib/optimizer/optimizer.cpp
janetzki: Any position between the `ExpressionReductionRule` and the `SubqueryToJoinRule` works. Now, I moved it directly beneath the `PredicateSplitUpRule` to reflect the similarity of these rules.

src/test/optimizer/strategy/or_to_union_rule_test.cpp
janetzki: I added a test for it.

src/lib/optimizer/optimizer.cpp
janetzki: The downside of this order is that the introduced unions remain in the LQP even if no other rule benefits from them. @mrks If I remember correctly, you mentioned that there should be an additional clean-up rule at the end of the optimizer?

janetzki:
@mrks @mrzzzrm I addressed your comments and could reduce the runtime for TPC-H as well (see PR description).

mrks:
Sweeet

src/lib/optimizer/strategy/disjunction_to_union_rule.cpp
```diff
@@ -0,0 +1,63 @@
...
+namespace opossum {
+
+std::string DisjunctionToUnionRule::name() const { return "Disjunction to Union Rule"; }
+
+void DisjunctionToUnionRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root) const {
+  Assert(root->type == LQPNodeType::Root, "DisjunctionToUnionRule needs root to hold onto");
+
+  /**
+   * Step 1:
+   *    - Collect PredicateNodes that can be split up into multiple ones into `predicate_nodes_to_flat_disjunctions`
```
mrks: I understand why `ExpressionReductionRule` is not a good place for this.

As this is c/p from the PredicateSplitUpRule, have you considered joining those two rules? Not only would that reduce the code duplication and the number of LQP traversals, but it might also enable us to handle nested and/ors.

src/lib/optimizer/optimizer.cpp
mrks: Not sure if I can follow. Should we discuss the position of the rule and a possible merge of the rules in person?

mrks:
For me, TPC-H 7 segfaults. Also, moving the rule below the `BetweenCompositionRule` enables a potential optimization in TPC-H 16. Not benchmarked yet.

src/lib/optimizer/strategy/disjunction_to_union_rule.cpp
```diff
@@ -0,0 +1,63 @@
...
+   *    - Split up qualifying PredicateNodes into n-1 consecutive UnionNodes and n PredicateNodes. We have to do this in
+   *      a second pass because manipulating the LQP within `visit_lqp()`, while theoretically possible, is prone to
+   *      bugs.
+   */
+  for (const auto& [predicate_node, flat_disjunction] : predicate_nodes_to_flat_disjunctions) {
+    if (flat_disjunction.size() <= 1) {
+      break;
+    }
+
+    auto previous_union_node = UnionNode::make(UnionMode::Positions);
```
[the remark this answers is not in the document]
mrks: I gave this some more thought. Reliably identifying cases where the two (or more) sides are truly disjunct might take more than a handful of lines. Let's leave that for a future step. However, maybe you could benchmark how much of a difference it would make by simply replacing it and disregarding overlaps for the benchmark.

mrks:
I played around with this a bit, trying to understand why TPC-H 7 segfaults.

The way it looks, both the BetweenCompositionRule and the PredicatePlacementRule are not too good with handling diamonds. That's not super surprising, as we haven't used those too much yet. I have a good grip on the PredicatePlacementRule, as I am currently implementing something else there. Hopefully, I'll have a PR open before our meeting on Thursday.

If you could have a look at the BetweenCompositionRule that would be great.

src/lib/optimizer/strategy/disjunction_to_union_rule.cpp
janetzki: As discussed yesterday, this is not possible until #1813 is fixed.

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -407,4 +407,25 @@ TEST_F(BetweenCompositionTest, NonBoundaryPredicate) {
   EXPECT_LQP_EQ(result_lqp, expected_lqp);
 }
 
+TEST_F(BetweenCompositionTest, NoPullPastDiamond) {
```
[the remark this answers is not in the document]
janetzki: Unfortunately, I did not find the failing LQP yet because I cannot print TPC-H 7 when it crashes. To spent not too much time on the investigation now, I consider this failing test a part of #1813.

src/lib/optimizer/strategy/disjunction_to_union_rule.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include "abstract_rule.hpp"
+
+namespace opossum {
+
+/**
+ * This rule turns PredicateNodes with disjunctive chains (e.g., `PredicateNode(a OR b OR c)`) as their scan expression
+ *    into n-1 consecutive UnionNodes that combine the results of n PredicateNodes with simple expressions.
```
mrks: What is a "simple expression"? Those could be arbitrarily complex, too. Leave out the "simple"?

src/lib/optimizer/strategy/disjunction_to_union_rule.hpp
```diff
@@ -0,0 +1,25 @@
...
+ *    This rule is similar to the PredicateSplitUpRule, which handles ANDs instead of ORs. The DisjunctionToUnionRule
+ *    runs after the PredicateSplitUpRule, so that expressions that are not in conjunctive normal form (ANDs of ORs)
+ *    cannot be fully handled right now. Because the PredicateSplitUpRule already ran, the DisjunctionToUnionRule
+ *    assumes that no ANDs are left in the PredicateExpressions.
```
mrks: "assumes that no ANDs are left in the PredicateExpressions" - not necessarily true. We might have an AND somewhere deeper within the expressions.

Suggestion: 
> The DisjunctionToUnionRule runs after the PredicateSplitUpRule, so when flattening the top-level logical expression, we expect one or more top-level expressions that were connected using a disjunction ("or"). Conjunctions ("and") might occur nested within such a top-level expression. As the PredicateSplitUpRule already ran, they will not be split up into successive PredicateNodes. That might be a thing for the future, if it ever comes up.

src/lib/optimizer/strategy/disjunction_to_union_rule.hpp
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Doing so enables other Optimizer rules to process these PredicateNodes.
+ *
+ * EXAMPLE: TPC-DS query 35.
+ *   Without this rule `EXISTS (...) OR EXISTS (...)` would not be split up and therefore would not be reformulated to
+ *   two semi-joins.
```
mrks: Reducing the number of negations generally makes things easier to read:

> This rule splits up `EXISTS (...) OR EXISTS (...)` into two expressions that can later be rewritten into two semi-joins.

janetzki:
I have bad news: Unfortunately, TPC-H 19 does not benefit from the DisjunctionToUnionRule anymore, since the PredicatePlacementRule does (correctly) not push predicates under its Join anymore. (This means that the previous improvement of the runtime was only possible due to a wrong result.)
At least TPC-DS 35 did not suffer from this error.

janetzki:
TPC-H Benchmark (with `Union::Positions`)
image

src/lib/optimizer/strategy/disjunction_to_union_rule.cpp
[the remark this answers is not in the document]
janetzki: image
If we use `UnionAll` instead of `UnionPositions`, TPC-H 16 can be optimized further. It does not seem to be a large potential though.

src/lib/optimizer/strategy/disjunction_to_union_rule.cpp
janetzki: There also is another conflict with the `JoinOrderingRule`, which needs to run before the `DisjunctionToUnionRule`.

janetzki:
I work on

mrks:
If I see it correctly, the TPC-H suffers because we run the ExpressionEvaluator twice in Q7 (and thrice in Q19) instead of once. As its main cost is in materializing the inputs, that will hurt.

Both queries look like the problem would be fixed if we split up the conjunctions so that we can use proper scans. For testing, please add the PredicateSplitUpRule a second time after your rule. If that works, we should check (a) what happens if we merge them into one rule that runs recursively, and (b) if we need the split up at its current location at all.

That should solve the problems that we are seeing right now.

src/lib/optimizer/strategy/disjunction_to_union_rule.cpp
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  /**
+   * Step 2:
+   *    - Split up qualifying PredicateNodes into n-1 consecutive UnionNodes and n PredicateNodes. We have to do this in
+   *      a second pass because manipulating the LQP within `visit_lqp()`, while theoretically possible, is prone to
+   *      bugs.
+   */
+  for (const auto& [predicate_node, flat_disjunction] : predicate_nodes_to_flat_disjunctions) {
+    if (flat_disjunction.size() <= 1) {
```
mrzzzrm: This if makes no sense, right? You made sure that no vectors <= 1 go into predicate_nodes_to_flat_disjunctions a few lines above - and even if not, why would you want to break in that case?

src/test/optimizer/strategy/disjunction_to_union_rule_test.cpp
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  std::shared_ptr<DisjunctionToUnionRule> _rule;
+
+  std::shared_ptr<MockNode> node_a, node_b, node_c, node_d, node_e;
+  LQPColumnReference a_a, a_b, b_a, b_b, c_a, d_a, e_a;
+};
+
+TEST_F(DisjunctionToUnionRuleTest, TwoExistsToUnion) {
+  // SELECT * FROM a WHERE EXISTS (
```
mrzzzrm: why are you testing this with EXISTS predicates? just makes the tests hard to read, what's wrong with a dead simple a > 5?

.gitignore
```diff
@@ -35,6 +35,12 @@ provider.hpp
 /*.csv
 /*.csv.meta
 
+# Ignore generated and cached TPC-DS tables
+*.csv
+*.bin
```
mrks: Oooh no. That applies to all csvs and bins. What's with test data etc.?

Jenkinsfile
```diff
@@ -309,9 +309,6 @@ try {
       if (env.BRANCH_NAME == 'master' || full_ci) {
         githubNotify context: 'Full CI', status: 'FAILURE'
       }
-      if (env.BRANCH_NAME == 'master') {
```
mrks: This looks like one of the last commits was bigger than planned

.gitignore
janetzki: Sorry, I just mixed up the merge branch.

src/lib/optimizer/strategy/disjunction_to_union_rule.cpp
janetzki: Good catch. This was a redundancy.

janetzki:
> For testing, please add the PredicateSplitUpRule a second time after your rule.

image

Great idea! :ok_hand:

src/lib/optimizer/strategy/disjunction_to_union_rule.cpp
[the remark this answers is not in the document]
janetzki: image
This is the same comparison if we run the `PredicateSplitUpRule` after the `DisjunctionToUnionRule`. There still is no major advantage except for TPC-H 16.

mrks:
Please let me know when this is ready for another review

src/test/optimizer/strategy/disjunction_to_union_rule_test.cpp
janetzki: This was a relict because I assumed that only the `SubqueryToJoinRule` would benefit from the `DisjunctionToUnionRuleTest`.

janetzki:
>  If that works, we should check (a) what happens if we merge them into one rule that runs recursively, and (b) if we need the split up at its current location at all.

Merging the rules (a) did not change the good TPC-H benchmark results from above significantly, as well as executing the merged `PredicateSplitUpRule` only directly before and after the `JoinOrderingRule` (b).
So we still execute the `PredicateSplitUpRule` twice because the `JoinOrderingRule` depends on it but reverts the reformulation of disjunctions to unions when it calls `JoinGraphBuilder::_parse_union()`.

mrzzzrm:
JoinGraphBuilder::_parse_union() is a relic when we didn't have OR expressions. I think it can be removed entirely.

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
```diff
@@ -2,22 +2,20 @@
...
-#include "expression/lqp_subquery_expression.hpp"
 #include "logical_query_plan/lqp_utils.hpp"
+#include "logical_query_plan/union_node.hpp"
 
 namespace opossum {
 
-void PredicateSplitUpRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root) const {
-  Assert(root->type == LQPNodeType::Root, "PredicateSplitUpRule needs root to hold onto");
-
+bool PredicateSplitUpRule::splitAnd(const std::shared_ptr<AbstractLQPNode>& node) const {
```
mrzzzrm: splitConjunction

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
```diff
@@ -29,18 +27,81 @@ void PredicateSplitUpRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root
...
     for (const auto& predicate_expression : flat_conjunction) {
       lqp_insert_node(predicate_node, LQPInputSide::Left, PredicateNode::make(predicate_expression));
     }
     lqp_remove_node(predicate_node);
   }
+
+  return true;
+}
+
+bool PredicateSplitUpRule::splitOr(const std::shared_ptr<AbstractLQPNode>& node) const {
```
mrzzzrm: splitDisjunction

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
```diff
@@ -29,18 +27,81 @@ void PredicateSplitUpRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root
...
+    }
+  }
+
+  return true;
+}
+
+void PredicateSplitUpRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root) const {
+  Assert(root->type == LQPNodeType::Root, "PredicateSplitUpRule needs root to hold onto");
+
+  while (splitAnd(root) || splitOr(root)) {
```
mrzzzrm: Keep in mind that this is doing redundant re-traversals of the entire LQP. As soon as a single splitable predicate is there, this is doing four(!) traversals over the entire LQP.

janetzki:
> JoinGraphBuilder::_parse_union() is a relic when we didn't have OR expressions. I think it can be removed entirely.

Just commenting it out leads to a crash: `Code below relies on the JoinGraph having vertices` in `dp_ccp.cpp`. Do you spontaneously know how to fix this @mrzzzrm? Otherwise, I will dig deeper into that tomorrow.

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
janetzki: How much do we care about the optimizer's performance?

janetzki: A change here could come at the cost of clarity.

mrzzzrm: Ask the dude who does TPCC #1803 

It's not just performance, though. Code should just do what is necessary for the sake of the person trying to understand it while stepping/breaking through it in a debugger, or looking at it via a profiler capture.

mrzzzrm: Also, think of a person who's tasked to implement a similar rule and adapts this rule. They might be confused by the apparent redundancy and might deduce by implication that it is necessary for their rule as well, even if good sense tells them otherwise.

janetzki: Alright, I will change it.

mrzzzrm: If you find a good way. I just said "keep in mind..." ;)

mrzzzrm:
Union Node should be added to `_lqp_node_type_is_vertex` and removed from the switch in JoinGraphBuilder::_traverse, that should do it

janetzki:
> Union Node should be added to `_lqp_node_type_is_vertex` and removed from the switch in JoinGraphBuilder::_traverse, that should do it

image

It seems that this disables the JoinGraphBuilder to do any optimization below a Union. TPC-H 7 (picture) and 19 now contain only cross joins. @mrzzzrm Do you know what to do about this?

mrks:
Does the same happen if you run the PredicateSplitUpRule "light", i.e., only splitting up conjunctions?

janetzki:
> Does the same happen if the first time you run the PredicateSplitUpRule you only split up conjunctions?

No, this could be a workaround. This way, we would still apply the PredicateSplitUpRule twice (and with different settings). It would be nicer to only do this once before the JoinOrderingRule.

mrzzzrm:
The reason for this is that the join predicates are above the union and the joins below it, but the JoinGraph is build for nodes below and above vertices separately.

...and even if that problem can be overcome somehow, building a Join Graph from a non-tree DAG will require further adjustments. Is it possible to run the JoinOrderingRule before UnionPositions gets placed in the plan and before the memoization takes effect?

The JoinGraph/Builder and JoinOrderingRule where really only build with trees and diamonds-representing-disjunctions in mind, not with general case DAGs, I have to admit.

janetzki:
> Is it possible to run the JoinOrderingRule before UnionPositions gets placed in the plan and before the subplan memoization takes effect?

Yes, this is the workaround that also @mrks proposed. Although it means executing the PredicateSplitUpRule once before the JoinOrderingRule (without splitting up the disjunctions) and once after it (also splitting up disjunctions), I see no better way to solve this conflict.

janetzki:
@mrks @mrzzzrm Thank you for your ideas. You can take another look now.

janetzki:
41-LQP psur
This is the optimized query 41. Its performance dropped by 71% because the `PredicateSplitUpRule` unpacked a deeply nested PrediateNode to this huge diamond. Therefore, I am implementing a `PredicateMergeRule` at the very end of the optimizer that reverts as many changes of the `PredicateSplitUpRule` as possible.

mrzzzrm:
Why first "split up" only to then "merge" again? On which criteria do you base the decision to merge (i.e. union to OR?) predicates again? can we base this on cardinality estimation? pls make sure you're not creating a rule constellation specifically designed for Q41.

janetzki:
The merge can only happen where no other rule, such as the SubqueryToJoinRule, took advantage of the predicate split-up.

mrzzzrm:
So all UnionPositions are removed from the plan again because a OR evaluated by the Expression evaluator is always faster? That cannot be true, right?

janetzki:
> So all UnionPositions are removed from the plan again because a OR evaluated by the Expression evaluator is always faster? That cannot be true, right?

Another explanation for the long runtime would be that the PredicateReorderingRule does not pull the undermost PredicateNode below the ValidateNode, which would happen if the PredicateNode did not have multiple outputs.

janetzki:
The PredicateReorderingRule was indeed one reason why query 41 was so slow.
However, it still needs the PredicateMergeRule because the ExpressionEvaluator is faster if we condense large sub-plans (consisting of PredicateNodes and UnionNodes) to single PredicateNodes.
To differentiate between large and small sub-plans, I implemented a simple heuristic that counts the number of UnionNodes in a sub-plan. The threshold 4 is a value that leads to good results in the TPC-H and TPC-DS benchmark.

src/lib/optimizer/optimizer.cpp
```diff
@@ -91,15 +88,19 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {
...
 
-  // Run before SubqueryToJoinRule, since the Semi/Anti Joins it introduces are opaque to the JoinOrderingRule
+  // Run once before the JoinOrderingRule without disjunction split-up because the JoinOrderingRule cannot handle UNION
+  // (#1829)
+  optimizer->add_rule(std::make_unique<PredicateSplitUpRule>(false));
+
+  // Run before SubqueryToJoinRule since the Semi/Anti Joins it introduces are opaque to the JoinOrderingRule
   optimizer->add_rule(std::make_unique<JoinOrderingRule>());
 
+  optimizer->add_rule(std::make_unique<PredicateSplitUpRule>());
```
mrks: Not a full review, but did you test different positions for this? For #1819, it would be helpful if this rule was behind the PredicatePlacementRule. Would that break anything on your side?

janetzki: That works. Nothing breaks and the performance does not notably change.

src/lib/optimizer/strategy/predicate_merge_rule.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/**
+ * Merge an LQP that only consists of PredicateNodes and UnionNodes into a single PredicateNode. The
+ * subsequent_expression parameter passes the translated expressions to the translation of its children nodes, which
+ * enables to add the translated expression of child node before its parent node to the output expression.
+ */
+std::shared_ptr<AbstractExpression> PredicateMergeRule::_merge_subplan(
+    const std::shared_ptr<AbstractLQPNode>& begin,
+    const std::optional<const std::shared_ptr<AbstractExpression>>& subsequent_expression) const {
```
janetzki: This function is loosely based on `lqp_subplan_to_boolean_expression` in `lqp_utils`. The major difference it that the LQP is traversed and changed at the same time here.

src/lib/optimizer/strategy/predicate_reordering_rule.cpp
```diff
@@ -45,37 +45,40 @@ bool is_predicate_style_node(const std::shared_ptr<AbstractLQPNode>& node) {
...
-
-    // Gather adjacent PredicateNodes
-    auto current_node = node;
-    while (is_predicate_style_node(current_node)) {
-      // Once a node has multiple outputs, we're not talking about a Predicate chain anymore
-      if (current_node->outputs().size() > 1) {
-        break;
+  Assert(root->type == LQPNodeType::Root, "PredicateReorderingRule needs root to hold onto");
+
+  visit_lqp(root, [&](const auto& node) {
```
janetzki: I just put the function into `visit_lqp`, which the diff does not show nicely.

src/lib/optimizer/strategy/index_scan_rule.cpp
```diff
@@ -28,26 +28,29 @@ constexpr float INDEX_SCAN_SELECTIVITY_THRESHOLD = 0.01f;
...
 constexpr float INDEX_SCAN_ROW_COUNT_THRESHOLD = 1000.0f;
 
-void IndexScanRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node) const {
+void IndexScanRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root) const {
   DebugAssert(cost_estimator, "IndexScanRule requires cost estimator to be set");
+  Assert(root->type == LQPNodeType::Root, "ExpressionReductionRule needs root to hold onto");
 
-  if (node->type == LQPNodeType::Predicate) {
-    const auto& child = node->left_input();
+  visit_lqp(root, [&](const auto& node) {
```
janetzki: I just put the function into `visit_lqp`, which the diff does not show nicely.

janetzki:
@mrks The PR is ready for review now. I also updated the PR description. Could you please have another look?

src/test/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -118,7 +118,7 @@ TEST_F(BetweenCompositionTest, EmptyColumnRange) {
   EXPECT_LQP_EQ(result_lqp, expected_lqp);
 }
 
-TEST_F(BetweenCompositionTest, OrExpression) {
+TEST_F(BetweenCompositionTest, NoPullPastOrExpression) {
```
mrks: I don't see a change to the BetweenCompositionRule, so how was #1813 fixed? Isn't it still broken on certain diamonds?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -410,7 +410,7 @@ cxxopts::Options BenchmarkRunner::get_basic_cli_options(const std::string& bench
     ("visualize", "Create a visualization image of one LQP and PQP for each query, do not properly run the benchmark", cxxopts::value<bool>()->default_value("false")) // NOLINT
     ("verify", "Verify each query by comparing it with the SQLite result", cxxopts::value<bool>()->default_value("false")) // NOLINT
     ("cache_binary_tables", "Cache tables as binary files for faster loading on subsequent runs", cxxopts::value<bool>()->default_value("false")) // NOLINT
-    ("sql_metrics", "Track SQL metrics (parse time etc.) for each SQL query", cxxopts::value<bool>()->default_value("false")); // NOLINT
+    ("sql_metrics", "Track SQL metrics (parse time etc.) for each SQL query (needs output file to be set)", cxxopts::value<bool>()->default_value("false")); // NOLINT
```
mrks: `Track SQL metrics (parse time etc.) for each SQL query and adds it to the output JSON`?

src/lib/logical_query_plan/lqp_utils.hpp
```diff
@@ -100,7 +100,7 @@ enum class LQPVisitation { VisitInputs, DoNotVisitInputs };
 /**
  * Calls the passed @param visitor on @param lqp and recursively on its INPUTS. This will NOT visit subqueries.
  * The visitor returns `ExpressionVisitation`, indicating whether the current nodes's input should be visited
- * as well.
+ * as well. The algorithm is BFS.
```
mrks: ASS

src/lib/optimizer/optimizer.cpp
```diff
@@ -124,6 +125,8 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {
 
   optimizer->add_rule(std::make_unique<IndexScanRule>());
 
+  optimizer->add_rule(std::make_unique<PredicateMergeRule>(4));
```
mrks: Why is this an externally exposed parameter, no a parameter within the rule?

src/lib/optimizer/strategy/predicate_merge_rule.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <vector>
+
+#include "abstract_rule.hpp"
+#include "logical_query_plan/predicate_node.hpp"
+#include "logical_query_plan/union_node.hpp"
+
+namespace opossum {
+
+/**
+ * This rule reverts the changes of the PredicateSplitUpRule after other rules have run. It merges multiple
```
mrks: "This rule reverts the changes"
- "All changes?"

src/lib/optimizer/strategy/predicate_merge_rule.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+PredicateMergeRule::PredicateMergeRule(const size_t optimization_threshold)
+    : _optimization_threshold(optimization_threshold) {}
+
+void PredicateMergeRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root) const {
+  Assert(root->type == LQPNodeType::Root, "PredicateMergeRule needs root to hold onto");
+
+  visit_lqp(root, [&](const auto& node) {
+    size_t lqp_complexity = 0;
+    visit_lqp(node, [&](const auto& sub_node) {
```
mrks: Two nested traversal of DAGs are hard to understand / verify. Can you think of a way to do this in a single pass? Also, "complexity" sounds like something more sophisticated than what we are doing right now.

src/lib/optimizer/strategy/predicate_merge_rule.cpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    return LQPVisitation::VisitInputs;
+  });
+}
+
+/**
+ * Merge an LQP that only consists of PredicateNodes and UnionNodes into a single PredicateNode. The
+ * subsequent_expression parameter passes the translated expressions to the translation of its children nodes, which
+ * enables to add the translated expression of child node before its parent node to the output expression.
```
mrks: "enables" requires an object

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -591,6 +591,8 @@ void SubqueryToJoinRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node)
       // an example, see TPC-H query 21. That query is the main reason this part exists. The current thresholds are
       // somewhat arbitrary. If we start to see more of those cases, we can try and find a more complex heuristic. This
       // is also called "semi join reduction": http://www.db.in.tum.de/research/publications/conferences/semijoin.pdf
+      // However, TPC-DS 45 could be executed faster without an additional semi join reduction node.
+      // TODO(anybody): Adjust this heuristic so that both TPC-H 21 and TPC-DS 45 run fast.
```
mrks: Please remove this comment. I'm revamping the entire semi join part anyway.

src/lib/optimizer/strategy/predicate_merge_rule.cpp
mrks: I can't really follow these two methods. Could you give an example LQP?

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
```diff
@@ -2,44 +2,90 @@
...
-      }
+      predicate_nodes.emplace_back(predicate_node);
     }
-
     return LQPVisitation::VisitInputs;
   });
 
+  // _split_conjunction() and _split_disjunction() split up logical expressions by calling each other recursively
+  for (const auto& predicate_node : predicate_nodes) {
+    if (!_split_conjunction(predicate_node)) {
```
mrks: I'd rather check for `LogicalOperator::And` than call `_split_conjunction` and check the return value. That should be easier to follow.

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
```diff
@@ -2,44 +2,90 @@
...
+    lqp_insert_node(predicate_node, LQPInputSide::Left, new_predicate_node);
+    _split_disjunction(new_predicate_node);
+  }
+  lqp_remove_node(predicate_node);
+
+  return true;
+}
+
+void PredicateSplitUpRule::_split_disjunction(const std::shared_ptr<PredicateNode>& predicate_node) const {
+  if (!_should_split_disjunction) {
```
mrks: I'd call it simply `_split_disjunctions`.

mrks: This is not the right place for this check. It is not the job of `_split_disjunction` to decide whether it should do something or not - that is the job of the caller.

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
```diff
@@ -2,44 +2,90 @@
...
+  const auto flat_disjunction = flatten_logical_expressions(predicate_node->predicate(), LogicalOperator::Or);
+  if (flat_disjunction.size() <= 1) {
+    return;
+  }
+
+  /**
+   * Split up PredicateNode with disjunctive chain (e.g., `PredicateNode(a OR b OR c)`) as their scan expression into
+   * n-1 consecutive UnionNodes and n PredicateNodes.
+   */
+  auto previous_union_node = UnionNode::make(UnionMode::Positions);
```
mrks: Why `previous` and not `current`?

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
```diff
@@ -205,9 +205,12 @@ TEST_F(PredicateReorderingTest, PredicatesWithMultipleOutputs) {
      *    \           /
      *     Predicate_b
      *         |
+     *     Predicate_c
+     *         |
      *       Table
      *
-     * predicate_a should come before predicate_b - but since Predicate_b has two outputs, it can't be reordered
+     * Predicate_a should come before Predicate_b - but since Predicate_b has two outputs, it can't be reordered.
```
mrks: not your fault, but "should come" is misleading. No, it should not, because that would be wrong.

src/lib/optimizer/optimizer.cpp
janetzki: Because the parameter is set to 1 for testing.

janetzki: Because the parameter is set to 1 for testing.

mrks: You can also set it in the test without adding it as a parameter that is not used in regular execution:

https://github.com/hyrise/hyrise/blob/f79e2c99deb902bdc8632147ed6183c8cfcd015e/src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp#L40-L41

src/lib/logical_query_plan/lqp_utils.hpp
janetzki: What is the full name of this algorithm?

mrks: breadth-first search

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
janetzki: Good point. The variable names became misleading here. I changed them to make their roles in each step clearer.

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
janetzki: You're right. This way, `_split_conjunction` and `_split_disjunction` also get unified function signatures.

src/benchmarklib/benchmark_runner.cpp
janetzki: How about `Track SQL metrics (parse time etc.) for each SQL query and adds it to the output JSON (if set)` to make clear that this parameter depends on `-o`?

mrks: `Track SQL metrics (parse time etc.) and add it to the output JSON (see -o)`?

src/lib/logical_query_plan/lqp_utils.hpp
janetzki: What does `ASS` mean?

mrks: [ASS](https://gist.github.com/klaaspieter/12cd68f54bb71a3940eae5cdd4ea1764) ;)

janetzki: Alright, now I don't have to mess up my search history anymore. D:

src/test/optimizer/strategy/between_composition_rule_test.cpp
janetzki: Good that you mention it. I almost assumed that the problem was fixed because nothing crashed in the benchmarks again.
The actual problem was not that the `BetweenCompositionRule` could not handle diamond LQPs but that it could not handle expressions like `a = 1 AND b = 2`. The crash happened because the `DisjunctionToUnionRule` created two of those expressions for TPC-H 7.

janetzki:
@mrks Thank you for all your suggestions. I hope that these clarifications are helpful. I also updated the benchmarks. Could you please have another look?

src/test/optimizer/strategy/predicate_merge_rule_test.cpp
```diff
@@ -0,0 +1,278 @@
...
+    node_a = MockNode::make(MockNode::ColumnDefinitions{{DataType::Int, "a"}, {DataType::Int, "b"}});
+    a_a = node_a->get_column("a");
+    a_b = node_a->get_column("b");
+
+    node_b = MockNode::make(MockNode::ColumnDefinitions{{DataType::Int, "a"}, {DataType::Int, "b"}}, "b");
+    b_a = node_b->get_column("a");
+    b_b = node_b->get_column("b");
+
+    rule = std::make_shared<PredicateMergeRule>();
+    // A value of 1 means that the rule won't merge pure predicate chains without UNIONs.
```
mrks: Rather: "Reducing the optimization_threshold so that plans are merged earlier, making the test cases shorter"?

src/lib/optimizer/strategy/predicate_merge_rule.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <vector>
+
+#include "abstract_rule.hpp"
+#include "logical_query_plan/predicate_node.hpp"
+#include "logical_query_plan/union_node.hpp"
+
+namespace opossum {
+
+/**
+ * This rule cleans up any large subplan that the PredicateSplitUpRule created after other optimizer rules have run.
```
mrks: Suggestion "The PredicateSplitUpRule might have rewritten disjunctions into complex predicate/union chains. If no rule between the PredicateSplitUpRule and this rule has benefited from these chains, executing them as predicates and unions might be more expensive than having the ExpressionEvaluator run on the original monolithic expression. Controlled by [varname, see below], this rule reverts the PredicateSplitUpRule's changes by merging multiple PredicateNodes and UnionNodes into single a PredicateNode"?

src/lib/optimizer/strategy/predicate_merge_rule.hpp
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * EXAMPLE:
+ *   TPC-DS query 41 benefits from this rule because the PredicateSplitUpRule creates a huge LQP.
+ */
+class PredicateMergeRule : public AbstractRule {
+ public:
+  void apply_to(const std::shared_ptr<AbstractLQPNode>& root) const override;
+
+  // Simple heuristic: The PredicateMergeRule is more likely to improve the performance for complex LQPs.
+  size_t optimization_threshold{4};
```
mrks: `maximum_subsequent_unions`?

src/lib/optimizer/strategy/predicate_merge_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+        lqp_remove_node(begin->left_input());
+        predicate_node->node_expressions[0] = left_input_expression;
+        return left_input_expression;
+      } else {
+        return expression;
+      }
+    }
+
+    case LQPNodeType::Union: {
+      const auto union_node = std::dynamic_pointer_cast<UnionNode>(begin);
```
mrks: We should check the UnionMode here. Otherwise, this might conflict with, e.g., an actual union from SQL or a union created by the IndexScanRule

src/lib/optimizer/strategy/predicate_merge_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+      auto expression = predicate_node->predicate();
+      if (subsequent_expression && begin->output_count() == 1) {
+        // Only merge inside a chain
+        expression = and_(expression, *subsequent_expression);
+      }
+
+      const auto left_input_expression = _merge_subplan(begin->left_input(), expression);
+      if (begin->left_input()->output_count() == 1 && left_input_expression) {
+        // Only merge inside a chain. Nodes at the bottom of a diamond would unnecessarily inflate the resulting logical
+        // expression now.
```
mrks: Isn't this the opposite from what the ASCII art shows?

src/lib/optimizer/strategy/predicate_merge_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+        lqp_remove_node(begin->left_input());
+        predicate_node->node_expressions[0] = left_input_expression;
+        return left_input_expression;
+      } else {
+        return expression;
+      }
+    }
+
+    case LQPNodeType::Union: {
```
mrks: This is the case with which everything starts, correct? I'd place it first for better readability.

src/lib/optimizer/strategy/predicate_merge_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+        auto expression = or_(left_input_expression, right_input_expression);
+
+        if (subsequent_expression && begin->output_count() == 1) {
+          // Only merge inside a chain
+          expression = and_(expression, *subsequent_expression);
+        }
+
+        lqp_remove_node(begin->left_input());
+        lqp_remove_node(begin->right_input());
+        Assert(begin->left_input() == begin->right_input(), "The LQP should be an \"empty\" diamond now.");
```
mrks: "The diamond should now be collapsed"?

src/lib/optimizer/strategy/predicate_merge_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+    if (node_and_count.second >= optimization_threshold) {
+      _merge_subplan(node_and_count.first, std::nullopt);
+    }
+  }
+}
+
+/**
+ * Merge a subplan that only consists of PredicateNodes and UnionNodes into a single PredicateNode. The
+ * subsequent_expression parameter passes the translated expressions to the translation of its children nodes, which
+ * allows to add the translated expression of child node before its parent node to the output expression.
```
mrks: Still having trouble understanding the subsequent_expression part. Is this just to maintain the order within the resulting expression? If yes, why is this necessary?

mrks:
In any case: Great results!

src/test/optimizer/strategy/between_composition_rule_test.cpp
janetzki: Exactly.

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
```diff
@@ -2,44 +2,91 @@
...
+  if (flat_conjunction.size() <= 1) {
+    return;
+  }
+
   /**
-   * Step 2:
-   *    - Split up qualifying PredicateNodes into multiple consecutive PredicateNodes. We have to do this in a
-   *      second pass, because manipulating the LQP within `visit_lqp()`, while theoretically possible, is prone to
-   *      bugs.
+   * Split up PredicateNode with conjunctive chain (e.g., `PredicateNode(a AND b AND c)`) as its scan expression into
```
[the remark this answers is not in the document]
janetzki: The mix was actually on purpose in this comment. I hope it became clearer now.

src/lib/optimizer/strategy/predicate_split_up_rule.cpp
```diff
@@ -2,44 +2,91 @@
...
+      predicate_nodes.emplace_back(predicate_node);
     }
-
     return LQPVisitation::VisitInputs;
   });
 
+  // _split_conjunction() and _split_disjunction() split up logical expressions by calling each other recursively.
+  for (const auto& predicate_node : predicate_nodes) {
+    const auto flat_conjunction = flatten_logical_expressions(predicate_node->predicate(), LogicalOperator::And);
+    if (flat_conjunction.size() > 1) {
```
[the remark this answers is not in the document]
janetzki: You are right, this code is shorter and works as well. We now might have a superfluous call of `flatten_logical_expressions()` in `_split_disjunction()` because `_split_conjunction()` already unnested all expressions, but this does not increase the runtime compared to the previous solution.

janetzki:
> Getting there. I'll be out-of-office next week, but if you want, we can also sit down after that and work this through together.

Thanks again for all these specific remarks! I will continue to polish up this PR and hope that there won't be much left to discuss. ;)

src/lib/optimizer/strategy/predicate_merge_rule.hpp
```diff
@@ -25,7 +25,7 @@ class PredicateMergeRule : public AbstractRule {
   void apply_to(const std::shared_ptr<AbstractLQPNode>& root) const override;
 
   // Simple heuristic: The PredicateMergeRule is more likely to improve the performance for complex LQPs.
-  size_t optimization_threshold{4};
+  size_t optimization_threshold{10}; // > TPCDS-13
```
Bouncner: That name does really tell anything.

What does it do?

Bouncner:
For an outsider like me it’s very hard to follow why we have rules that are reverted at the end in some situations.
Some points that have been discussed here would help in the code as well as in the PR description.

src/lib/optimizer/strategy/predicate_merge_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+ *           |                        |                        |                              |
+ *      ___Union___              ___Union___           Predicate (A OR B)      Predicate ((D AND C) AND (A OR B))
+ *    /            \            /           \                  |                              |
+ * Predicate (A)   |         Predicate (A)  |                  |
+ *    |            |           |            |                  |
+ *    |       Predicate (B)    |      Predicate (B)            |
+ *    \           /            \          /                    |
+ *     Predicate (C)          Predicate (D AND C)     Predicate (D AND C)
+ *           |                        |                        |
+ *     Predicate (D)
```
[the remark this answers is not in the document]
janetzki: In general, it is even necessary because conjunctive predicates might be part of a disjunctive predicate (e.g., `(A AND B) OR C`). In this case, we can only merge the disjunction if we merged the conjunction before. However, I have not evaluated the case that a predicate chain is not inside a diamond, such as in the drawn example.

src/lib/optimizer/strategy/predicate_merge_rule.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include "logical_query_plan/union_node.hpp"
+
+using namespace opossum::expression_functional;  // NOLINT
+
+namespace opossum {
+
+void PredicateMergeRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root) const {
+  Assert(root->type == LQPNodeType::Root, "PredicateMergeRule needs root to hold onto");
+
+  // For every node, store a pointer to the root of the subplan if this subplan can be merged
```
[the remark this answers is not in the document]
janetzki: Good point. But the algorithm works differently: Instead of looking for disjunctive chains it counts the total number of `UnionNodes` (with `UnionMode::Positions`) in each subplan that only consists of `UnionNodes` (with `UnionMode::Positions`) and `PredicateNodes`. I changed the variable names accordingly.

src/lib/optimizer/strategy/predicate_merge_rule.hpp
janetzki: I renamed it to `minimum_union_count`.

janetzki:
Thanks again for your remarks @Bouncner @mrks! I have addressed them and hope that the comments and the PR description now clarify the idea.
Admittedly, the implementation of `_merge_subplan()` was very hard to understand. I splitted it up into `_merge_disjunction()` and `_merge_conjunction()`, which is a) more intuitive, b) corresponds to the implementation of the `PredicateSplitUpRule`, and is c) more reliable. The previous implementation had assumed that all UnionNodes came from the `PredicateSplitUpRule`, while the new one also works on more general cases. The concept of the new functions is to look for small, safe patterns in the subplan and only merge those.

mrks:
Thanks again. I added some last things that I found while looking at the changes, let me know if you agree.

mrks:
Since I have a PR that builds on this, I'll go ahead and merge this. Should you have any remarks about my last commit, please let me know.

PR #1793, opened 2019-08-15
Enable executable TPC-DS WITH-queries

mweisgut:
Since #1715 the sql WITH keyword is supported. This PR enables the executable WITH-queries of the TPC-DS benchmark which are currently blacklisted.


mrks:
Nais

PR #1795, opened 2019-08-15
Query Plan Optimization: Sub-Plan Memoization

### LINKED ISSUE #1816: Rename AbstractLQPNode::_on_hash() to _shallow_hash()
#1781 does this renaming for Expressions. Consistency.

mweisgut:
https://medium.com/hyrise/query-plan-optimization-sub-plan-memoization-25232491ddbc

#### TPC-H
```
+---------------------+----------------+------+----------------+------+------------+-----------------------------------------------+
| Benchmark           | prev. iter/s   | runs | new iter/s     | runs | change [%] |               p-value (significant if <0.001) |
+---------------------+----------------+------+----------------+------+------------+-----------------------------------------------+
| TPC-H 1 -> TPC-H 01 | 0.279449790716 | 6    | 0.2815528512   | 6    | +1%        | (run time too short) (not enough runs) 0.8993 |
| TPC-H 2 -> TPC-H 02 | 3.22970509529  | 65   | 3.67050719261  | 74   | +14%       |                   (run time too short) 0.0000 |
| TPC-H 3 -> TPC-H 03 | 2.14790701866  | 43   | 2.14673376083  | 43   | -0%        |                   (run time too short) 0.9485 |
| TPC-H 4 -> TPC-H 04 | 0.905623495579 | 19   | 0.877843558788 | 18   | -3%        |                   (run time too short) 0.0000 |
| TPC-H 5 -> TPC-H 05 | 1.48339641094  | 30   | 1.39016258717  | 28   | -6%        |                   (run time too short) 0.0000 |
| TPC-H 6 -> TPC-H 06 | 14.250289917   | 286  | 14.1953907013  | 284  | -0%        |                   (run time too short) 0.0610 |
| TPC-H 7 -> TPC-H 07 | 0.692684054375 | 14   | 0.694493353367 | 14   | +0%        |                   (run time too short) 0.7063 |
| TPC-H 8 -> TPC-H 08 | 1.53931510448  | 31   | 1.6660721302   | 34   | +8%        |                   (run time too short) 0.0000 |
| TPC-H 9 -> TPC-H 09 | 0.626673102379 | 13   | 0.653532266617 | 14   | +4%        |                   (run time too short) 0.0000 |
| TPC-H 10            | 1.30277228355  | 27   | 1.29520022869  | 26   | -1%        |                   (run time too short) 0.3691 |
| TPC-H 11            | 7.55953168869  | 152  | 7.43670082092  | 149  | -2%        |                   (run time too short) 0.0000 |
| TPC-H 12            | 2.97480916977  | 60   | 2.8850581646   | 58   | -3%        |                   (run time too short) 0.0000 |
| TPC-H 13            | 1.18233120441  | 24   | 1.19659924507  | 24   | +1%        |                   (run time too short) 0.2550 |
| TPC-H 14            | 9.02865028381  | 181  | 9.13107395172  | 183  | +1%        |                   (run time too short) 0.0001 |
| TPC-H 15            | 5.30843019485  | 107  | 10.6722774506  | 214  | +101%      |                   (run time too short) 0.0000 |
| TPC-H 16            | 3.14431834221  | 63   | 3.10523700714  | 63   | -1%        |                   (run time too short) 0.0001 |
| TPC-H 17            | 0.555964708328 | 12   | 0.568802654743 | 12   | +2%        |                   (run time too short) 0.0000 |
| TPC-H 18            | 0.488441616297 | 10   | 0.553546845913 | 12   | +13%       |                   (run time too short) 0.0000 |
| TPC-H 19            | 2.39886951447  | 48   | 2.36555552483  | 48   | -1%        |                   (run time too short) 0.0002 |
| TPC-H 20            | 1.05688273907  | 22   | 1.07021021843  | 22   | +1%        |                   (run time too short) 0.2082 |
| TPC-H 21            | 0.288463324308 | 6    | 0.316424250603 | 7    | +10%       | (run time too short) (not enough runs) 0.0028 |
| TPC-H 22            | 4.36559915543  | 88   | 4.33049583435  | 87   | -1%        |                   (run time too short) 0.0524 |
| geometric mean      |                |      |                |      | +5%        |                                               |
+---------------------+----------------+------+----------------+------+------------+-----------------------------------------------+
```

#### TPC-DS
```
+----------------+-----------------+------+-----------------+------+------------+-----------------------------------------------+
| Benchmark      | prev. iter/s    | runs | new iter/s      | runs | change [%] |               p-value (significant if <0.001) |
+----------------+-----------------+------+-----------------+------+------------+-----------------------------------------------+
| 01             | 5.64845609665   | 113  | 8.17224502563   | 164  | +45%       |                   (run time too short) 0.0000 |
| 07             | 3.35293197632   | 68   | 3.36980414391   | 68   | +1%        |                   (run time too short) 0.5139 |
| 09             | 1.2186870575    | 25   | 1.23807573318   | 25   | +2%        |                   (run time too short) 0.0000 |
| 10             | 0.928672552109  | 19   | 0.938642501831  | 19   | +1%        |                   (run time too short) 0.0141 |
| 13             | 1.67283332348   | 34   | 1.64862775803   | 33   | -1%        |                   (run time too short) 0.0009 |
| 15             | 7.61087417603   | 153  | 7.61201906204   | 153  | +0%        |                   (run time too short) 0.9762 |
| 17             | 2.60594034195   | 53   | 2.7865011692    | 56   | +7%        |                   (run time too short) 0.0000 |
| 25             | 2.92911601067   | 59   | 3.16035842896   | 64   | +8%        |                   (run time too short) 0.0000 |
| 26             | 6.69022464752   | 134  | 6.55376577377   | 132  | -2%        |                   (run time too short) 0.0003 |
| 28             | 2.87008357048   | 58   | 2.92549300194   | 59   | +2%        |                   (run time too short) 0.0000 |
| 29             | 2.965893507     | 60   | 2.92838096619   | 59   | -1%        |                   (run time too short) 0.3704 |
| 31             | 0.189053609967  | 4    | 0.563680052757  | 12   | +198%      | (run time too short) (not enough runs) 0.0000 |
| 34             | 3.79716157913   | 76   | 3.77345752716   | 76   | -1%        |                   (run time too short) 0.3845 |
| 35             | 0.0128922602162 | 1    | 0.0130517156795 | 1    | +1%        |                         (not enough runs) nan |
| 41             | 23.7274627686   | 475  | 23.3697738647   | 468  | -2%        |                   (run time too short) 0.0000 |
| 42             | 4.25412750244   | 86   | 4.21553421021   | 85   | -1%        |                   (run time too short) 0.0088 |
| 43             | 1.19385719299   | 24   | 1.22339653969   | 25   | +2%        |                   (run time too short) 0.0004 |
| 45             | 12.3740215302   | 248  | 11.9634418488   | 240  | -3%        |                   (run time too short) 0.0000 |
| 48             | 1.69309926033   | 34   | 1.66344356537   | 34   | -2%        |                   (run time too short) 0.0011 |
| 50             | 4.20877027512   | 85   | 4.17820978165   | 84   | -1%        |                   (run time too short) 0.2071 |
| 62             | 4.01882791519   | 81   | 4.0545873642    | 82   | +1%        |                   (run time too short) 0.0005 |
| 65             | 1.51340687275   | 31   | 2.91312098503   | 59   | +92%       |                   (run time too short) 0.0000 |
| 69             | 1.89709353447   | 38   | 1.85111570358   | 38   | -2%        |                   (run time too short) 0.0000 |
| 73             | 4.65311956406   | 94   | 4.59162950516   | 92   | -1%        |                   (run time too short) 0.0900 |
| 79             | 2.2418410778    | 45   | 2.27030873299   | 46   | +1%        |                   (run time too short) 0.0742 |
| 81             | 9.87025737762   | 198  | 14.8880281448   | 298  | +51%       |                   (run time too short) 0.0000 |
| 83             | 8.09310913086   | 162  | 15.363612175    | 308  | +90%       |                   (run time too short) 0.0000 |
| 88             | 0.585754513741  | 12   | 3.2152068615    | 65   | +449%      |                   (run time too short) 0.0000 |
| 93             | 0.209024578333  | 5    | 0.213119491935  | 5    | +2%        | (run time too short) (not enough runs) 0.0008 |
| 96             | 4.61264944077   | 93   | 4.57706022263   | 92   | -1%        |                   (run time too short) 0.3726 |
| 97             | 0.831402182579  | 17   | 0.839005768299  | 17   | +1%        |                   (run time too short) 0.3710 |
| 99             | 1.95268070698   | 40   | 1.97234606743   | 40   | +1%        |                   (run time too short) 0.0113 |
| geometric mean |                 |      |                 |      | +17%       |                                               |
+----------------+-----------------+------+-----------------+------+------------+-----------------------------------------------+
```

fixes #1816


mrks:
Nais

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -94,9 +94,10 @@ std::shared_ptr<AbstractOperator> LQPTranslator::translate_node(const std::share
    * would result in multiple operators created from predicate_c and thus in performance drops
    */
 
-  const auto operator_iter = _operator_by_lqp_node.find(node);
-  if (operator_iter != _operator_by_lqp_node.end()) {
-    return operator_iter->second;
+  for (const auto& [l_node, p_node] : _operator_by_lqp_node) {
```
mrks: Calling "deep" equals for an LQP Node is probably quite expensive. I'd add a hash on it that only includes the node type and a hash of its node expressions. That can be done in the abstract LQP node in <10 lines and should enable us to use a hashmap here.

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -121,6 +121,18 @@ std::vector<IndexStatistics> StoredTableNode::indexes_statistics() const {
   return stored_indexes_statistics;
 }
 
+size_t StoredTableNode::_on_hash() const {
```
mrks: Ouch. Good catch.

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -68,6 +68,29 @@ AbstractLQPNode::~AbstractLQPNode() {
   if (_inputs[1]) _inputs[1]->_remove_output_pointer(*this);
 }
 
+size_t AbstractLQPNode::hash() const {
```
mrks: Could you add a short comment on how you decided for this to be the hash function? Especially the special treatment of validate etc. should be explained.

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -11,6 +11,7 @@
 #include "abstract_lqp_node.hpp"
 #include "aggregate_node.hpp"
 #include "alias_node.hpp"
+#include "bytell_hash_map.hpp"
```
mrks: Did you see any performance benefits with that map? We usually only use it if it has a significant benefit over `unordered_map`.

mweisgut:
Even when the test suite passed for version 09f7312cc0436dddf80bf59871c33ad4705dec7e, some TPC-H and -DS query executions returned the wrong result. dea7f8b553de96b45d7daee060f4fd2c732a3ec2 returns correct results but now, the desired optimization is not achieved anymore. I haven't investigated the cause for this in detail now, but semantically correct nodes have different hash codes. Possibly different node expression hash codes may be the reason.

Implementing _on_hash for each concrete node instead of case distinction in the AbstractLQPNode seems to be a more elegant solution.

mrks:
> Implementing _on_hash for each concrete node instead of case distinction in the AbstractLQPNode seems to be a more elegant solution.

If you have problems with something not being found, I'd start by *reducing* the complexity of the hash, down to a degenerate hash (i.e., `return 0`).

mrks:
> Even when the test suite passed for version 09f7312, some TPC-H and -DS query executions returned the wrong result.

This means that the test suite is insufficient. Once you find the issue, please try and add an appropriate test.

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -68,6 +68,36 @@ AbstractLQPNode::~AbstractLQPNode() {
...
+  auto hash = boost::hash_value(static_cast<LQPNodeType>(type));
+  for (const auto& expression : node_expressions) {
+    boost::hash_combine(hash, expression->hash());
+  }
+  boost::hash_combine(hash, _on_hash());
+
+  switch (type) {
```
mrzzzrm: Abstract base classes shouldn't know about derived types.

```c++
if (left_input) hash_combine(hash, left_input->hash());
if (right_input) hash_combine(hash, right_input->hash());
```

Here too, visit_lqp would avoid redundant traversal of subplans below diamond shapes

src/lib/logical_query_plan/lqp_translator.hpp
```diff
@@ -68,9 +69,8 @@ class LQPTranslator {
       const std::vector<std::shared_ptr<AbstractExpression>>& lqp_expressions,
       const std::shared_ptr<AbstractLQPNode>& node) const;
 
-  // Cache operator subtrees by LQP node to avoid executing operators below a diamond shape multiple times
-  mutable std::unordered_map<std::shared_ptr<const AbstractLQPNode>, std::shared_ptr<AbstractOperator>>
-      _operator_by_lqp_node;
+  // Cache operator subtrees by LQP node hash to avoid executing identical operators redundantly
+  mutable ska::bytell_hash_map<size_t, std::shared_ptr<AbstractOperator>> _operator_by_lqp_node;
```
mrzzzrm: That's a bug, hashes can collide and then the wrong subplan gets inserted. LQP still need to be the key, use a custom hasher to extract the hash of the LQP from the shared ptr

src/lib/logical_query_plan/abstract_lqp_node.cpp
mweisgut: I'll address this after we got a working impl..

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -68,6 +68,36 @@ AbstractLQPNode::~AbstractLQPNode() {
   if (_inputs[1]) _inputs[1]->_remove_output_pointer(*this);
 }
 
+size_t AbstractLQPNode::hash() const {
+  auto hash = boost::hash_value(static_cast<LQPNodeType>(type));
```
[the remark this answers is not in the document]
mweisgut: It is not. Will be removed.

mweisgut:
Implementing the hash functions for LQP nodes revealed optimization potential for the _on_hash function for derived nodes. In general, including the hashes of all node expressions on the abstract lqp node level (`AbstractLQPNode::hash()`) is currently insufficient: Semantically identical nodes would have different hash values. Therefore, node expressions are currently not included into the node hash code. But this is leading to weak hash codes for some concrete node types.

mrks:
> Semantically identical nodes would have different hash values. 

Do you have an example for me?

mrzzzrm:
Would be interested to know what you mean by that as well. At least by design this shouldn't be the case.

mweisgut:
I'll investigate this in detail soon. For now, I just identified this issue for some predicate and join nodes in TPC-DS query 88. For a node to be translated (`node`), I iterate over all previously translated lqp nodes and check if the probe node (`lqp`) equals `node` and if the hash of `node` is different to the hash of `lqp`.

```
for (const auto& [lqp, pqp] : _operator_by_lqp_node) {
    if (*lqp == *node && lqp->hash() != node->hash()) {
      std::cout << "equal but different hash codes:\n";
      std::cout << *lqp << "\n";
      std::cout << *node << "\n";
    }
  }
```

I only find qualifying nodes for `*lqp == *node && lqp->hash() != node->hash()` in Q88 when I combine the hashes of node_expressions to the corresponding node hash.

An update will follow soon.

mweisgut:
https://github.com/hyrise/hyrise/pull/1795/commits/3c870998109f94b33fb851f09d15652caa89dcb5#diff-d0d013499d10dc0309a779411a835222R110-R121

mrks:
In that case `*expression_a == *expression_b` should also be false, because the `LQPColumnReference` references different original nodes (which, in the end, map to the same table). I can see how this becomes a problem and will need to think a bit more about that. Not sure what the expected behavior is. 

However, just from reading I can't map this to `*lqp == *node && lqp->hash() != node->hash()`.

mrks:
This is one of the culprits:

```
bool PredicateNode::_on_shallow_equals(const AbstractLQPNode& rhs, const LQPNodeMapping& node_mapping) const {
  const auto& predicate_node = static_cast<const PredicateNode&>(rhs);
  const auto equal =
      expression_equal_to_expression_in_different_lqp(*predicate(), *predicate_node.predicate(), node_mapping);

  return equal;
}
```

For `operator==`, we map expressions from one LQP to the other. For `hash`, we don't. This is a conceptual thing that I will need to think about for a while.

At least for this case, simplifying the LQPColumnReference would work around the issue. Not sure if it a solution:

```
size_t hash<opossum::LQPColumnReference>::operator()(const opossum::LQPColumnReference& column_reference) const {
  return static_cast<size_t>(column_reference.original_column_id());
}
```

src/test/logical_query_plan/create_table_node_test.cpp
```diff
@@ -64,7 +64,7 @@ TEST_F(CreateTableNodeTest, Hash) {
 
   EXPECT_NE(different_create_table_node_a->hash(), create_table_node->hash());
   EXPECT_NE(different_create_table_node_b->hash(), create_table_node->hash());
-  // TODO(anyone) StaticTableNode's hash code should take the table into account (at least table definitions) 
+  // TODO(anyone) StaticTableNode's hash code should take the table into account (at least table definitions)
```
mweisgut: include table ptr

mrzzzrm:
Okay, so apparently we (I :( ) included some pointers in the hash computations - fine, get them out of there :P I don't really see a deeper problem here? Should just be a case of going through all `_on_hash` implementations for expressions and LQP nodes and removing/replacing the usage of pointers?

And btw, @mweisgut, `boost::hash_combine()` will hash the second arg itself, there should be no need to call `hash_combine(hash, hash_value(value))`, can just be `hash_combine(hash, value)`

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -157,6 +165,11 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
   std::vector<std::shared_ptr<AbstractExpression>> node_expressions;
 
  protected:
+  /**
+   * Override to hash data fields in derived types. No override needed if derived expression has no
+   * data members.
+   */
+  virtual size_t _on_hash() const;
```
mrks: `_shallow_hash()`? See #1781, #1816

src/lib/logical_query_plan/lqp_column_reference.cpp
```diff
@@ -37,7 +37,7 @@ std::ostream& operator<<(std::ostream& os, const LQPColumnReference& column_refe
 namespace std {
 
 size_t hash<opossum::LQPColumnReference>::operator()(const opossum::LQPColumnReference& column_reference) const {
-  auto hash = boost::hash_value(column_reference.original_node().get());
+  auto hash = boost::hash_value(column_reference.original_node()->hash());
```
mrks: Given how long this took us, add a comment?

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -91,6 +91,14 @@ std::shared_ptr<AbstractOperator> LQPTranslator::translate_node(const std::share
    * would result in multiple operators created from predicate_c and thus in performance drops
    */
 
+  for (const auto& [lqp, pqp] : _operator_by_lqp_node) {
+    if (*lqp == *node && lqp->hash() != node->hash()) {
+      std::cout << "equal but different hash codes:\n";
```
mrks: ?

src/lib/logical_query_plan/lqp_translator.hpp
```diff
@@ -68,9 +68,8 @@ class LQPTranslator {
       const std::vector<std::shared_ptr<AbstractExpression>>& lqp_expressions,
       const std::shared_ptr<AbstractLQPNode>& node) const;
 
-  // Cache operator subtrees by LQP node to avoid executing operators below a diamond shape multiple times
-  mutable std::unordered_map<std::shared_ptr<const AbstractLQPNode>, std::shared_ptr<AbstractOperator>>
-      _operator_by_lqp_node;
+  // Cache operator subtrees by LQP node hash to avoid executing identical operators redundantly
```
mrks: ... Also, this takes care of diamond shapes.

src/lib/logical_query_plan/sort_node.cpp
```diff
@@ -31,6 +31,14 @@ std::string SortNode::description() const {
   return stream.str();
 }
 
+size_t SortNode::_on_hash() const {
+  size_t hash{0};
+  for (const auto& order_by_mode : order_by_modes) {
+    boost::hash_combine(hash, boost::hash_value(order_by_mode));
```
mrks: Did we need hash_value? If not, please check the diff if was left in somewhere else, too

src/test/logical_query_plan/alias_node_test.cpp
```diff
@@ -43,4 +47,47 @@ TEST_F(AliasNodeTest, ShallowEqualsAndCopy) {
...
+  const auto expr_a = std::make_shared<LQPColumnExpression>(LQPColumnReference{other_mock_node, ColumnID{0}});
+  const auto expr_b = std::make_shared<LQPColumnExpression>(LQPColumnReference{other_mock_node, ColumnID{1}});
+  const auto other_expressions = std::vector<std::shared_ptr<AbstractExpression>>{expr_a, expr_b};
+  const auto alias_node_other_expressions = AliasNode::make(other_expressions, aliases, mock_node);
+  EXPECT_NE(alias_node->hash(), alias_node_other_expressions->hash());
+  const auto alias_node_other_left_input = AliasNode::make(expressions, aliases, other_mock_node);
+  // alias_node == alias_node_other_left_input is true (see AliasNodeTest/Equals)
+  // but the hash codes of these nodes are not equal.
+  EXPECT_EQ(alias_node->hash(), alias_node_other_left_input->hash());
+  // the following assertions clearify why the nodes are not equal but their hashes are
```
mrks: Typo: clarify

Also, do they?

src/test/logical_query_plan/alias_node_test.cpp
```diff
@@ -43,4 +47,47 @@ TEST_F(AliasNodeTest, ShallowEqualsAndCopy) {
...
+  const auto other_mock_node =
+      MockNode::make(MockNode::ColumnDefinitions{{DataType::Int, "a"}, {DataType::Float, "b"}}, "named");
+  const auto expr_a = std::make_shared<LQPColumnExpression>(LQPColumnReference{other_mock_node, ColumnID{0}});
+  const auto expr_b = std::make_shared<LQPColumnExpression>(LQPColumnReference{other_mock_node, ColumnID{1}});
+  const auto other_expressions = std::vector<std::shared_ptr<AbstractExpression>>{expr_a, expr_b};
+  const auto alias_node_other_expressions = AliasNode::make(other_expressions, aliases, mock_node);
+  EXPECT_NE(alias_node->hash(), alias_node_other_expressions->hash());
+  const auto alias_node_other_left_input = AliasNode::make(expressions, aliases, other_mock_node);
+  // alias_node == alias_node_other_left_input is true (see AliasNodeTest/Equals)
+  // but the hash codes of these nodes are not equal.
```
mrks: Wait, that should never ever happen. Why is this still the case?

src/test/logical_query_plan/create_prepared_plan_node_test.cpp
```diff
@@ -44,6 +44,22 @@ TEST_F(CreatePreparedPlanNodeTest, Equals) {
...
+  const auto different_prepared_plan_node_a = CreatePreparedPlanNode::make("some_prepared_plan2", prepared_plan);
+
+  const auto different_lqp = MockNode::make(MockNode::ColumnDefinitions({{DataType::Int, "b"}}));
+  const auto different_prepared_plan =
+      std::make_shared<PreparedPlan>(different_lqp, std::vector<ParameterID>{ParameterID{1}});
+  const auto different_prepared_plan_node_b =
+      CreatePreparedPlanNode::make("some_prepared_plan", different_prepared_plan);
+
+  EXPECT_NE(different_prepared_plan_node_a->hash(), create_prepared_plan_node->hash());
+  EXPECT_EQ(different_prepared_plan_node_b->hash(), create_prepared_plan_node->hash());
```
mrks: These tests look very much like the equals tests. Wouldn't it be shorter to combine the two?

src/lib/logical_query_plan/lqp_translator.cpp
mweisgut: debug code, will be removed

src/lib/logical_query_plan/lqp_translator.hpp
mweisgut: As discussed in today's Fortnightly, term "identical" would cover this diamond shape case, wouldn't it?
But even if this is the case, the "equal case" (no diamond shape but semantically equal nodes)  is not covered. I will clarify the comment.

src/test/logical_query_plan/alias_node_test.cpp
mweisgut: I mistakenly switched the evaluation results of the expression.

// alias_node == alias_node_other_left_input **is false** (see AliasNodeTest/Equals)
// but the hash codes of these nodes **are equal**.

mrks: Maybe add a comment on why we are testing this. Having something in a test means that it is expected behavior. If someone sees this in the future, they might think that the hashes being equal is some magically necessary thing.

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -68,6 +69,33 @@ AbstractLQPNode::~AbstractLQPNode() {
...
+        boost::hash_combine(hash, expression->hash());
+      }
+      boost::hash_combine(hash, node->type);
+      boost::hash_combine(hash, node->_shallow_hash());
+      // since visit_lqp is used, the hash for an already visited node is not combined with
+      // the overall hash again, even if the same node is used as left and right input node (diamond structure).
+      // Therefore, a node that has only one (left) input node could have the same hash as a node with two
+      // (left and right) inputs which are in a diamond structure (same node pointer).
+      // To differentiate these, the boolean value (left input == right input) is combined with the overall hash.
+      boost::hash_combine(hash, node->left_input() == node->right_input());
```
mrzzzrm: Maybe I missed the discussion around this, but when would left_input == right_input ever be non-false? In my mind: never.

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -144,6 +144,14 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
   bool operator==(const AbstractLQPNode& rhs) const;
   bool operator!=(const AbstractLQPNode& rhs) const;
 
+  /**
+   *  Builds a hash code by hashing the node type combined with specific member variables of
+   *  derived classes (see _shallow_hash()) and the hash codes of non-empty input nodes recursively.
+   *  Node expressions are not taken into account since combining the hashes of the expressions
```
mrzzzrm: But... node expressions are hashed!?

mrzzzrm:
@mweisgut ?

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: I remember understanding this a while back. :/

mrzzzrm: I think I understand the *problem* that's trying to be addressed here: the hash is just comprised of the nodes, not of the graph structure. But this is, unless I am being stupid, not the solution. I have never seen a plan where the left input equals the right input and even if, that's not what we usually refer to when we say "diamond structure"

mrks:
> There doesn't seem to be a new optimizer rule in the diff of this PR?

No. This was initially planned to be a rule, but that turned out to be a huge PITA because of the way the column reference works. The short version is that a sub-LQP cannot produce two instances of the same data for a self-join.

This violates the "all optimizations happen on the LQP" to some degree. However, after spending significant time on it and realizing that changing the columns references would require tens of hours of work and have a unclear impact on the overall readability, we decided to do this in the LQPTranslator for now.

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: Absolutely correct. Should it be `node->right_input() == nulltptr`?

mrzzzrm:
ooookay. But there don't seem to be tests for this?

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrzzzrm: Don't think that would help. Whether inputs are set should be sufficiently encoded in the node type.

mrks:
I believe we had some before and that they might have gone missing over the course of changing this from a rule. :/

src/lib/logical_query_plan/abstract_lqp_node.hpp
mweisgut: Correct, I missed to update that comment.

mweisgut: Correct, I missed to update this comment. Addressed in #1836

src/lib/logical_query_plan/abstract_lqp_node.cpp
mweisgut: As clarified in #1836, this line was only added for an UpdateNode-Test with only one (left) input node.

PR #1800, opened 2019-08-16
Null-aware Single Segment Indexes

### LINKED ISSUE #670: AdaptiveRadixTreeIndex behaves differently from other CompositeGroupKeyIndex

### LINKED ISSUE #1764: Index join tests crash for GroupKeyIndexes
Since #667 the index join is implemented in Hyrise. Already since this PR the GroupKeyIndex is not used for the index join tests.

https://github.com/sleighsoft/hyrise/blob/e40ac125e3e2dcedd3f808f1f9e8c5570111dd0f/src/test/operators/join_index_test.cpp#L103

Even if all current tests (in this version 7087c18bd8e94b69b86d26f564a69d25e520d530) are passing after adding `GroupKeyIndex` to the testing types, a segmentation fault arises after executing the test suite.

```
[==========] 5545 tests from 209 test cases ran. (5404 ms total)
[  PASSED  ] 5522 tests.
[  SKIPPED ] 23 tests, listed below:
[...]

Speicherzugriffsfehler (Speicherabzug geschrieben)
```

Has anyone investigated this issue before and knows more about the cause? Maybe @mrks or @Bouncner since you were involved in reviewing the mentioned PR?

mweisgut:
This PR adds null handling for the single segment indexes `AdaptiveRadixTreeIndex`, `BTreeIndex` and `GroupKeyIndex`.

- added the OuterJoinWithNull test mentioned in #670 
- renaming `BaseIndex` to `AbstractIndex` in compliance with the contribution guidelines
- AbstractIndex interface extension: 
  - added null_cbegin(): Returns an Iterator to the position of the first element equal to null.
  - added null_cend(): Returns an Iterator past the position of the last element equal to null.
- added asserts for segments_to_index (index construction): at least one segment has to be available
- adapted memory consumption calculation
- test suite extension

fixes #670 
fixes #1764 
ref #1818


src/test/storage/btree_index_test.cpp
```diff
@@ -78,17 +97,108 @@ TEST_F(BTreeIndexTest, MemoryConsumptionVeryShortString) {
...
+  //    24 (reported by cpp_btree implementation)
+  // +  24 std::vector<ChunkOffset> object overhead (index non-null postings)
+  // +   0 number of non-null elements (0) * sizeof(ChunkOffset) (4)
+  // +  24 std::vector<ChunkOffset> object overhead (index null postings)
+  // +   8 number of null elements (2) * sizeof(ChunkOffset) (4)
+  // =  80
+  EXPECT_EQ(index->memory_consumption(), 80u);
+#else
+  // libc++:
+  // TODO(anyone) implelemt more accurate
```
mweisgut: Typo

src/test/storage/btree_index_test.cpp
```diff
@@ -78,17 +97,108 @@ TEST_F(BTreeIndexTest, MemoryConsumptionVeryShortString) {
...
+  //    24 (reported by cpp_btree implementation)
+  // +  24 std::vector<ChunkOffset> object overhead (index non-null postings)
+  // +   0 number of non-null elements (0) * sizeof(ChunkOffset) (4)
+  // +  24 std::vector<ChunkOffset> object overhead (index null postings)
+  // +   0 number of null elements (0) * sizeof(ChunkOffset) (4)
+  // =  72
+  EXPECT_EQ(index->memory_consumption(), 72u);
+#else
+  // libc++:
+  // TODO(anyone) implelemt more accurate
```
mweisgut: Typo

src/lib/storage/index/abstract_index.hpp
```diff
@@ -111,14 +113,32 @@ class BaseIndex : private Noncopyable {
...
+   * Calls _null_cbegin() of the most derived class.
+   *
+   * @return An Iterator on the position of the first element equal to null.
+   */
+  Iterator null_cbegin() const;
+
+  /**
+   * Returns an Iterator past the position of the last element equal to null.
+   * Iterating from null_cbegin() to null_cend() will result in a position list with all null values.
+   * Calls _null_cend() of the most derived class.
```
mrks: no (same above).

src/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index.cpp
```diff
@@ -22,8 +22,10 @@ size_t AdaptiveRadixTreeIndex::estimate_memory_consumption(ChunkOffset row_count
 }
 
 AdaptiveRadixTreeIndex::AdaptiveRadixTreeIndex(const std::vector<std::shared_ptr<const BaseSegment>>& segments_to_index)
-    : BaseIndex{get_index_type_of<AdaptiveRadixTreeIndex>()},
-      _indexed_segment(std::dynamic_pointer_cast<const BaseDictionarySegment>(segments_to_index.front())) {
+    : AbstractIndex{get_index_type_of<AdaptiveRadixTreeIndex>()},
+      _indexed_segment(segments_to_index.empty()
```
mrks: When would this be the case? (other indexes, too)

src/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index.cpp
```diff
@@ -32,50 +34,75 @@ AdaptiveRadixTreeIndex::AdaptiveRadixTreeIndex(const std::vector<std::shared_ptr
...
   _root = _bulk_insert(pairs_to_insert);
 }
 
-BaseIndex::Iterator AdaptiveRadixTreeIndex::_lower_bound(const std::vector<AllTypeVariant>& values) const {
-  assert(values.size() == 1);
+AbstractIndex::Iterator AdaptiveRadixTreeIndex::_lower_bound(const std::vector<AllTypeVariant>& values) const {
+  Assert((values.size() == 1), "Adaptive Radix Tree Index expects exactly one input value");
+
+  if (variant_is_null(values[0])) {
+    return null_cbegin();
```
mrks: This is most likely not the expected behavior. In the STL, `lower_bound` "[r]eturns an iterator pointing to the first element in the range [first, last) that is not less than (i.e. greater or equal to) value, or last if no such element is found". As NULL compares false to everything, this definition would lead us to always return `[null_]cend()`.

Maybe it is the callers job to deal with this and that `AbstractIndex::(lower|upper)_bound` should enforce that no NULL values are passed.

However, if we want to have multi column indexes that can deal with `WHERE a = 1 AND b IS NULL`, this behavior would be difficult. Maybe we need to discuss this in more detail.

src/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index.cpp
```diff
@@ -32,50 +34,75 @@ AdaptiveRadixTreeIndex::AdaptiveRadixTreeIndex(const std::vector<std::shared_ptr
...
-BaseIndex::Iterator AdaptiveRadixTreeIndex::_cend() const { return _chunk_offsets.cend(); }
+AbstractIndex::Iterator AdaptiveRadixTreeIndex::_cend() const { return _chunk_offsets.cend(); }
 
 std::shared_ptr<ARTNode> AdaptiveRadixTreeIndex::_bulk_insert(
     const std::vector<std::pair<BinaryComparable, ChunkOffset>>& values) {
-  DebugAssert(!(values.empty()), "Index on empty segment is not defined");
-  _chunk_offsets.reserve(values.size());
-  auto begin = _chunk_offsets.cbegin();
-  return _bulk_insert(values, static_cast<size_t>(0u), begin);
+  if (!values.empty()) {
```
mrks: guard clause?

src/lib/storage/index/b_tree/b_tree_index_impl.cpp
```diff
@@ -55,20 +56,27 @@ BaseBTreeIndexImpl::Iterator BTreeIndexImpl<DataType>::upper_bound(DataType valu
...
+  index_null_postings.reserve(values.size());
 
   // Materialize
   segment_iterate<DataType>(*segment, [&](const auto& position) {
-    if (position.is_null()) return;
-    values.push_back(std::make_pair(position.chunk_offset(), position.value()));
+    if (position.is_null()) {
+      index_null_postings.emplace_back(position.chunk_offset());
+    } else {
+      values.push_back(std::make_pair(position.chunk_offset(), position.value()));
```
mrks: emplace_back should get rid of the make_pair

src/lib/storage/index/b_tree/b_tree_index_impl.cpp
```diff
@@ -77,14 +85,16 @@ void BTreeIndexImpl<DataType>::_bulk_insert(const std::shared_ptr<const BaseSegm
...
   // Build index
-  DataType current_value = values[0].second;
-  _btree[current_value] = 0;
-  _add_to_heap_memory_usage(current_value);
-  for (size_t i = 0; i < values.size(); i++) {
-    if (values[i].second != current_value) {
-      current_value = values[i].second;
-      _btree[current_value] = i;
-      _add_to_heap_memory_usage(current_value);
+  if (!values.empty()) {
```
mrks: guard clause?

src/lib/storage/index/group_key/group_key_index.cpp
```diff
@@ -14,41 +15,60 @@ size_t GroupKeyIndex::estimate_memory_consumption(ChunkOffset row_count, ChunkOf
...
-  _index_postings = std::vector<ChunkOffset>(_indexed_segments->size());
+  // 1) Initialize the index_offset:
+  //    Set the index_offset to size of the dictionary + 2 (plus one for null and plus one to mark the ending position)
+  //    and set all offsets to 0.
+  //    Using `_index_offsets`, we want to count the occurences of each ValueID of the attribute vector (AV).
+  //    The ValueID for null in an AV is the highest available ValueID in the dictionary + 1.
+  //    `unique_values_count` returns the size of dictionary which does not store a ValueID for null.
+  //    Therefore we have `unique_values_count` + 1 (for null) ValueIDs for which we want to count the occurrences.
+  _index_offsets = std::vector<ChunkOffset>(
+      _indexed_segment->unique_values_count() + 1u /*for null*/ + 1u /*to mark the ending position */, 0u);
```
mrks: Add a comment explaining that albeit the NULL values are stored in a different data structure in the super class, we first collect them together with the other values and cut them off later?

src/lib/storage/index/group_key/group_key_index.cpp
```diff
@@ -14,41 +15,60 @@ size_t GroupKeyIndex::estimate_memory_consumption(ChunkOffset row_count, ChunkOf
...
     auto position = 0u;
     for (; value_id_it != attribute_vector.cend(); ++value_id_it, ++position) {
       const auto& value_id = static_cast<ValueID>(*value_id_it);
-      _index_postings[index_offset_copy[value_id]] = position;
+
+      if (value_id != null_value_id) {
+        _index_postings[index_offset_copy[value_id]] = position;
+      } else {
+        const auto null_postings_offset = index_offset_copy[value_id] - non_null_count;
+        _index_null_postings[null_postings_offset] = position;
```
mrks: I think this would be easier to read if you simply did held and incremented an iterator into `_index_null_postings`

src/lib/storage/index/abstract_index.hpp
```diff
@@ -29,11 +29,13 @@ class BaseSegment;
  * As each index has a different way of iterating over its data structures, it has to implement its iterator as well.
  * We might use the impl-pattern similar to the TableScan, but this will be in a future commit.
  *
- * Find more information about this in our wiki: https://github.com/hyrise/hyrise/wiki/BaseIndex and
+ * Find more information about this in our wiki: https://github.com/hyrise/hyrise/wiki/AbstractIndex and
```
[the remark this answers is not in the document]
mweisgut: Not yet. Since the wiki-page references base_index.hpp of the master branch, I would update it after this PR is merged. For now, I just reverted the wiki reference changes.

src/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index.cpp
mweisgut: I think an empty segments_to_index vector wouldn't be passed in practice. But since the constructor is part of the (public) interface, indexes should either be able to handle an empty vector or throw an exception that it's not supported.

mrks: Sounds like it should be an assertion. As this is our own interface, we could always remove it if there was a good reason to pass an empty vector. Until then, let's assume that whoever called it with an empty vector will be happy about being called out.

mweisgut: But currently we even have the assertion a few lines below:
```c++
GroupKeyIndex::GroupKeyIndex(const std::vector<std::shared_ptr<const BaseSegment>>& segments_to_index)
    : AbstractIndex{get_index_type_of<GroupKeyIndex>()},
      _indexed_segment(segments_to_index.empty()
                           ? nullptr
                           : std::dynamic_pointer_cast<const BaseDictionarySegment>(segments_to_index[0])) {
  Assert(static_cast<bool>(_indexed_segment), "GroupKeyIndex only works with dictionary segments_to_index.");
  Assert((segments_to_index.size() == 1), "GroupKeyIndex only works with a single segment.");
```

`Assert(static_cast<bool>(_indexed_segment), "GroupKeyIndex only works with dictionary segments_to_index.")` does it. Would you rather use an assertion in the initializer list? If this is the case, how is it done?

mrks: Ah, now I see the point of the ternary expression. Maybe add `// Illegal, but range check needed for accessing the first segment` after the `? nullptr`?

mweisgut: agreed

src/lib/storage/index/abstract_index.cpp
```diff
@@ -0,0 +1,77 @@
...
+  switch (type) {
+    case SegmentIndexType::GroupKey:
+      return GroupKeyIndex::estimate_memory_consumption(row_count, distinct_count, value_bytes);
+    case SegmentIndexType::CompositeGroupKey:
+      return CompositeGroupKeyIndex::estimate_memory_consumption(row_count, distinct_count, value_bytes);
+    case SegmentIndexType::AdaptiveRadixTree:
+      return AdaptiveRadixTreeIndex::estimate_memory_consumption(row_count, distinct_count, value_bytes);
+    case SegmentIndexType::BTree:
+      return BTreeIndex::estimate_memory_consumption(row_count, distinct_count, value_bytes);
+    default:
```
mrks: Is this the case for any index? If not, remove it the default. It's usually better to fail at compile-time.

src/lib/storage/index/abstract_index.hpp
```diff
@@ -111,14 +113,30 @@ class BaseIndex : private Noncopyable {
...
+   * Returns an Iterator past the position of the largest indexed non-null element. This is useful for open
    * end range queries.
    * Iterating from cbegin() to cend() will result in a position list with ordered values.
    * Calls _cend() of the most derived class.
    * @return an Iterator on the end of the index (one after the last element).
    */
   Iterator cend() const;
 
+  /**
+   * Returns an Iterator to the position of the first element equal to null.
```
mrks: Add a comment that NULL is supported for single-column indexes only and that we do not yet have an idea on how to handle multi-columns?

src/lib/storage/index/abstract_index.hpp
```diff
@@ -137,6 +155,7 @@ class BaseIndex : private Noncopyable {
   virtual Iterator _cend() const = 0;
   virtual std::vector<std::shared_ptr<const BaseSegment>> _get_indexed_segments() const = 0;
   virtual size_t _memory_consumption() const = 0;
+  std::vector<ChunkOffset> _index_null_positions;
```
mrks: "index" doesn't seem to add a lot here

src/lib/storage/index/group_key/group_key_index.cpp
```diff
@@ -14,41 +15,67 @@ size_t GroupKeyIndex::estimate_memory_consumption(ChunkOffset row_count, ChunkOf
...
-  // 1a) Set the index_offset to size of the dictionary + 1 (plus one to mark the ending position)
-  //     and set all offsets to 0
-  _index_offsets = std::vector<size_t>(_indexed_segments->unique_values_count() + 1u, 0u);
-  // 1b) Set the _index_postings to the size of the attribute vector
-  _index_postings = std::vector<ChunkOffset>(_indexed_segments->size());
+  // 1) Initialize the index_offset:
+  //    Set the index_offset to size of the dictionary + 1 (plus one to mark the ending position)
+  //    and set all offsets to 0.
+  //    Using `_index_offsets`, we want to count the occurences of each ValueID of the attribute vector (AV).
+  //    The ValueID for NULL in an AV is the highest available ValueID in the dictionary + 1.
```
mrks: Acronyms seriously suck

src/lib/storage/index/group_key/group_key_index.cpp
```diff
@@ -14,41 +15,67 @@ size_t GroupKeyIndex::estimate_memory_consumption(ChunkOffset row_count, ChunkOf
...
   Assert((segments_to_index.size() == 1), "GroupKeyIndex only works with a single segment.");
 
-  // 1) Initialize the index structures
-  // 1a) Set the index_offset to size of the dictionary + 1 (plus one to mark the ending position)
-  //     and set all offsets to 0
-  _index_offsets = std::vector<size_t>(_indexed_segments->unique_values_count() + 1u, 0u);
-  // 1b) Set the _index_postings to the size of the attribute vector
-  _index_postings = std::vector<ChunkOffset>(_indexed_segments->size());
+  // 1) Initialize the index_offset:
+  //    Set the index_offset to size of the dictionary + 1 (plus one to mark the ending position)
```
mrks: Not your fault, but "the index_offset"? Isn't it "Create _index_offsets with a size"? Same in the "title" of this block

src/lib/storage/index/group_key/group_key_index.cpp
```diff
@@ -14,41 +15,67 @@ size_t GroupKeyIndex::estimate_memory_consumption(ChunkOffset row_count, ChunkOf
...
   //    i.e. the position in the _index_offsets
-  resolve_compressed_vector_type(*_indexed_segments->attribute_vector(), [&](auto& attribute_vector) {
+  const auto null_value_id = _indexed_segment->null_value_id();
+  auto null_count = 0u;
+
+  resolve_compressed_vector_type(*_indexed_segment->attribute_vector(), [&](auto& attribute_vector) {
     for (const auto value_id : attribute_vector) {
-      _index_offsets[value_id + 1u]++;
+      if (static_cast<ValueID>(value_id) != null_value_id) {
+        _index_offsets[value_id + 1u]++;
```
mrks: This is a bit confusing, because until 3), _index_offsets does not actually hold any offsets. How about using `value_histogram` first, moving it into `_value_offset` (renamed from _index_offsets) and then running `std::partial_sum` on it?

src/lib/storage/index/group_key/group_key_index.cpp
```diff
@@ -14,41 +15,67 @@ size_t GroupKeyIndex::estimate_memory_consumption(ChunkOffset row_count, ChunkOf
...
+  _index_null_positions = std::vector<ChunkOffset>(null_count);
+
+  // 4) Create offsets for the postings in _index_offsets
   std::partial_sum(_index_offsets.begin(), _index_offsets.end(), _index_offsets.begin());
 
-  // 4) Create the postings
-  // 4a) Copy _index_offsets to use it as a write counter
-  auto index_offset_copy = std::vector<size_t>(_index_offsets);
+  // 5) Create the postings
+  // 5a) Copy _index_offsets to use it as a write counter
```
mrks: Why? I don't see _index_offsets being modified after this. What do we need the copy for?

src/lib/storage/index/group_key/group_key_index.hpp
```diff
@@ -18,34 +18,45 @@ class GroupKeyIndexTest;
 /**
  *
  * The GroupKeyIndex works on a single dictionary compressed segment.
- * It uses two structures, one being a postings list containing record positions (ie ChunkOffsets)
- * in the attribute vector. The other structure is an index offset, mapping value-ids to offsets
+ * It uses three structures, one being a postings list containing record positions (ie ChunkOffsets)
+ * for non-null values in the attribute vector, the second is a postings list containing record positions
+ * for null values and the third structure is an index offset, mapping non-null-value-ids to offsets
```
mrks: This reads as if all three were members of this class (not of the superclass)

src/lib/storage/index/group_key/group_key_index.hpp
```diff
@@ -18,34 +18,45 @@ class GroupKeyIndexTest;
...
- *    | 2 |         3 | delta    ------->  3 ---------|     6 |
- *    | 3 |         2 | frank    ------->  5 -------| |-->  1 |
- *    | 4 |         0 | hotel    ------->  6 -----| |       3 |  ie "delta" can be found at i = 1 and 3
- *    | 5 |         1 | inbox    ------->  7 ---| | |---->  2 |
- *    | 6 |         1 |            |         |  | |------>  0 |
- *    | 7 |         5 |            |         |  |-------->  7 |  ie "inbox" can be found at i = 7 in the AV
- *    +---+-----------+------------+---------+----------------+
+ *  +----+-----------+------------+---------+----------------+----------------+
+ *  | (i)| Attribute | Dictionary |  Index  | Index Postings | Index Postings |
+ *  |    |  Vector   |            | Offsets | (non-null)     | (null)         |
```
mrks: Clarify that NULLs are stored in AbstractIndex

src/test/storage/btree_index_test.cpp
```diff
@@ -69,7 +69,26 @@ TEST_F(BTreeIndexTest, IndexProbes) {
 // Short String Optimization (SSO) stores strings of a certain size in the pmr_string object itself.
 // Only strings exceeding this size (15 for libstdc++ and 22 for libc++) are stored on the heap.
 
-TEST_F(BTreeIndexTest, MemoryConsumptionVeryShortString) {
+/*
+  size_t memory_consumption() const;
```
mrks: ?

src/test/storage/btree_index_test.cpp
```diff
@@ -69,7 +69,26 @@ TEST_F(BTreeIndexTest, IndexProbes) {
...
+  
+  |    Characteristic               | Block 1 | Block 2 |
+  |---------------------------------|---------|---------|
+  |[A] index is empty               |    true |   false |
+  |[B] index has null positions     |    true |   false |
+  |[C] index has non-null positions |    true |   false |
```
mrks: How to read this table and the choices / combinations below?

src/test/storage/single_segment_index_test.cpp
```diff
@@ -40,34 +128,436 @@ class SingleSegmentIndexTest : public BaseTest {
     return result;
   }
 
-  std::shared_ptr<BaseIndex> index_int = nullptr;
-  std::shared_ptr<BaseSegment> dict_segment_int = nullptr;
-  std::shared_ptr<BaseIndex> index_str = nullptr;
-  std::shared_ptr<BaseSegment> dict_segment_str = nullptr;
+  // segments
+  std::shared_ptr<BaseSegment> dict_segment_int_no_nulls = nullptr;
```
mrks: Declare them in a list to save some space?

src/test/storage/single_segment_index_test.cpp
```diff
@@ -40,34 +128,436 @@ class SingleSegmentIndexTest : public BaseTest {
...
-    DerivedIndexes;
+typedef ::testing::Types<AdaptiveRadixTreeIndex, BTreeIndex, /*CompositeGroupKeyIndex,*/ GroupKeyIndex> DerivedIndexes;
 TYPED_TEST_CASE(SingleSegmentIndexTest, DerivedIndexes, );  // NOLINT(whitespace/parens)
 
-TYPED_TEST(SingleSegmentIndexTest, FullRange) {
-  auto begin_int = this->index_int->cbegin();
-  auto end_int = this->index_int->cend();
-  auto result_values_int = this->result_as_vector(this->dict_segment_int, begin_int, end_int);
+/*
+  bool is_index_for(const std::vector<std::shared_ptr<const BaseSegment>>& segments) const;
```
mrks: ?

src/test/storage/single_segment_index_test.cpp
```diff
@@ -40,34 +128,436 @@ class SingleSegmentIndexTest : public BaseTest {
...
+  // A1, B2, C2, D1, E1, F1
+  EXPECT_FALSE(this->index_int_empty->is_index_for({}));
+  // other
+  EXPECT_FALSE(this->index_string_no_nulls->is_index_for({this->dict_segment_int_no_nulls}));
+  EXPECT_FALSE(
+      this->index_string_no_nulls->is_index_for({this->dict_segment_string_no_nulls, this->dict_segment_int_no_nulls}));
+}
+
+/*
+  Test case suits:
```
mweisgut: test suits

src/test/storage/single_segment_index_test.cpp
```diff
@@ -40,34 +128,436 @@ class SingleSegmentIndexTest : public BaseTest {
...
+  // A3, B4, C1, D2
+  EXPECT_EQ(this->index_float_mixed->upper_bound({3.1f}), this->index_float_mixed->cbegin() + 3);
+  // A4, B4, C1, D2
+  EXPECT_EQ(this->index_double_mixed->upper_bound({3.1}), this->index_double_mixed->cbegin() + 3);
+  // A5, B4, C1, D2
+  EXPECT_EQ(this->index_string_mixed->upper_bound({"hello"}), this->index_string_mixed->cbegin() + 3);
+}
+
+/*
+  Test case suits:
```
mweisgut: test suits

src/lib/storage/index/group_key/group_key_index.cpp
mweisgut: I aggree, I noticed that too. That would definitely improve legibility and comprehensibility.

src/lib/storage/index/group_key/group_key_index.cpp
mweisgut: _index_offsets is not modified after this but the copy is.
We increment the offset of the ValueID for which position was added to _index_postings in the corresponding iteration (see Snippet 1).
Since _index_offsets is used in `_get_postings_iterator_at(ValueID value_id) const` (see Snippet 2), modifying a ValueID offset in the original _index_offsets instead of in the copy would lead to an incorrect return value of `_get_postings_iterator_at(ValueID value_id) const`.

Snippet 1:
https://github.com/hyrise/hyrise/blob/7143a3fea58cb05134f5e2806d64b0bfbb1063fd/src/lib/storage/index/group_key/group_key_index.cpp#L80-L82

Snippet 2:
https://github.com/hyrise/hyrise/blob/7143a3fea58cb05134f5e2806d64b0bfbb1063fd/src/lib/storage/index/group_key/group_key_index.cpp#L114-L126

Renaming might clarify this:
`index_offset_coy` -> `index_insert_offsets`
`_index_offsets` -> `_index_start_offsets` (not absolutely necessary but increases the informative value)

src/test/storage/single_segment_index_test.cpp
mweisgut: On the other hand, it would reduce the comprehensibility if the descriptive names are missing in the test cases.

src/test/storage/btree_index_test.cpp
mweisgut: Input Domain Modeling is used here as testing technique. The input domain is scoped by parameters, which are structured in terms of characteristics (here `index is empty`, for example).
These characteristics are then partitioned into blocks, which represent the different options for a characteristic and thus sets of values for the actual test case. After characteristics are defined and partitioned into blocks, combinations of these blocks have to be chosen. Choosing all combinations is usually infeasible. (Since only three characteristics are listed with only two blocks each, choosing all combinations would be possible here.) Instead of choosing all combinations, other choosing strategies such as _Each Choice_, _Pair-Wise Choice_ and _Base Choice_ exist.

The added tests are created by using the _Base Choice_:
>Base Choice Coverage (BCC): A base choice block is chosen for each characteristic, and a base test is formed by using the base choice for each characteristic. Subsequent tests are chosen by holding all but one base choice constant and using each non-base choice in each other characteristic.
[AmOf17]

Therefore, a specific base choice has to be chosen. I decided to use the most likely combination from an end-use point of view, which is (a) the index is not empty, (b) the index stores NULL-positions and (c) the index stores non-NULL-positions. With this base choice, the additional combinations to be tested can be calculated as mentioned in the quote.

Base choice:
  A2, B1, C1
Hold [A] and [B] constant but use all other [C]-blocks yields to:
  A2, B1, C2
Hold [A] and [C] constant but use all other [B]-blocks yields to:
  A2, B2, C1
Hold [B] and [C] constant but use all other [B]-blocks yields to:
  A1, B1, C1
Since this derived combination is infeasible (index is empty but contains postings), one of the blocks held constant is changed so that a feasible combination is achieved:
A1, B2, C2

[AmOf17] Introduction to Software Testing (2nd Edition); Paul Ammann, Jeff Offutt, Cambridge University Press, 2017

src/test/storage/adaptive_radix_tree_index_test.cpp
```diff
@@ -72,8 +72,12 @@ class AdaptiveRadixTreeIndexTest : public BaseTest {
         EXPECT_EQ(index->upper_bound({*search_value}), index->lower_bound({*search_value}));
       }
 
-      int32_t min = **distinct_values.begin();
-      int32_t max = **distinct_values.rbegin();
+      const auto begin = distinct_values.begin();
+      const auto rbegin = distinct_values.rbegin();
+
+      Assert(*begin && *rbegin, "Optional has no value.");
```
Bensk1: The reference says: "The behavior is undefined if *this does not contain a value." I think we have to use `has_value()` here.

Bensk1:
I think almost everything from the reviews was addressed. There are only two points that Markus has to resolve, right?

src/test/storage/adaptive_radix_tree_index_test.cpp
mweisgut: But in this line, we don't use `std::optional<T>::operator*`. We dereference the iterators `begin` and `rbegin` to get  the `optional<int32_t>` values they are pointing to.
Building the logical expression `*begin && *rbegin` converts the optional-values to bool-values.

cppreference:
>When an object of type optional<T> is contextually converted to bool, the conversion returns true if the object contains a value and false if it does not contain a value.

Pleas correct me if I'm wrong.

src/test/storage/btree_index_test.cpp
mrks: Never heard of it. Choosing one default configuration and varying one parameter after another is common, but I did not know that it has a name. For more complex tests, those naming schemes make sense. However, I expect that I am not the only one unfamiliar with them. Here, it might be sufficient to write `non-empty index covering both NULL and non-NULL position` instead of `A2, B1, C1`. That should make it easier for the reader.

mrks: Just saw that the combinations in the other test are more complex. I would still suggest auto-replacing them with the textual description (or at least adding a description). If you leave the scheme in, please add a reference to a short description of the testing pattern.

src/lib/storage/index/group_key/group_key_index.cpp
mrks: Sounds good.

src/lib/storage/index/abstract_index.cpp
```diff
@@ -21,9 +21,10 @@ size_t AbstractIndex::estimate_memory_consumption(SegmentIndexType type, ChunkOf
       return AdaptiveRadixTreeIndex::estimate_memory_consumption(row_count, distinct_count, value_bytes);
     case SegmentIndexType::BTree:
       return BTreeIndex::estimate_memory_consumption(row_count, distinct_count, value_bytes);
-    default:
-      Fail("estimate_memory_consumption() is not implemented for the given index type");
+    case SegmentIndexType::Invalid:
+      Fail("SegmentIndexType is invalid.");
   }
+  Fail("GCC claims this is reachable.");
```
mrks: Consistency: `s/claims/thinks/`

src/test/storage/group_key_index_test.cpp
```diff
@@ -52,24 +52,27 @@ class GroupKeyIndexTest : public BaseTest {
...
-  EXPECT_EQ(expected_offsets, *index_offsets);
+  EXPECT_EQ(expected_offsets, *value_start_offsets);
 }
 
 /*
-  size_t memory_consumption() const;
+  Test cases:
+    IndexMemoryConsumption
+  Tested functions:
+    size_t memory_consumption() const;
```
mrks: I like these descriptions.

PR #1801, opened 2019-08-17
 CI: Store query plans as artifacts 

mrks:
Always wanted an easy way of looking at TPC-H query plans.


Jenkinsfile
```diff
@@ -232,14 +232,36 @@ try {
...
+              sh "mkdir ./clang-release-memcheck-test"
+              // If this shows a leak, try --leak-check=full, which is slower but more precise
+              sh "valgrind --tool=memcheck --error-exitcode=1 --gen-suppressions=all --num-callers=25 --suppressions=resources/.valgrind-ignore.txt ./clang-release/hyriseTest clang-release-memcheck-test --gtest_filter=-NUMAMemoryResourceTest.BasicAllocate"
+            } else {
+              Utils.markStageSkippedForConditional("memcheckReleaseTest")
+            }
+          }
+        }, tpchQueryPlans: {
+          stage("tpchQueryPlans") {
+            // Query plan generation runs as part of this parallel block in order to avoid a load imbalance between the parallel blocks
```
Bensk1: I don't really understand this.

mrks: I thought I'd explain why it is not part of that big `parallel` block but runs towards the end. In the end, it probably doesn't even matter. Removed the comment so that it doesn't confuse anyone.

PR #1806, opened 2019-08-19
Semi Join Reduction

mrks:
Shamelessly copying the description from the comment within the rule:

> Semi/Anti joins are currently handled by the hash join, which performs badly if the right side is significantly bigger than the left side. For that case, we add a second semi join on the right side, which throws out all values that will not be found by the primary predicate of the later join, anyway. However, we can only throw away values on the build side if the first predicate is an equals predicate. For an example, see TPC-H query 21. [...] For a lack of a better name, we call this "reciprocal semi joins".

I'm probably not the first one who came up with this, so if you know of any reference, please let me know.

TPC-H Queries to which this applies: 2, 4, 17, 20, 21 (twice)

```
| TPC-H 1        | 0.617155075073 | 38   | 0.616368353367 | 37   | -0%        |                          0.8141 |
| TPC-H 2        | 7.49960327148  | 450  | 8.03324890137  | 482  | +7%        |                          0.0000 |
| TPC-H 3        | 6.31156110764  | 379  | 6.36647224426  | 382  | +1%        |                          0.2525 |
| TPC-H 4        | 2.21944689751  | 134  | 2.60383200645  | 157  | +17%       |                          0.0000 |
| TPC-H 5        | 4.03684329987  | 243  | 3.79352736473  | 228  | -6%        |                          0.0000 |
| TPC-H 6        | 41.9420509338  | 2517 | 42.1069831848  | 2527 | +0%        |                          0.0000 |
| TPC-H 7        | 1.6284481287   | 98   | 1.60848653316  | 97   | -1%        |                          0.0007 |
| TPC-H 8        | 4.88635730743  | 294  | 4.87220859528  | 293  | -0%        |                          0.5591 |
| TPC-H 9        | 1.62544131279  | 98   | 1.57623064518  | 95   | -3%        |                          0.0000 |
| TPC-H 10       | 3.01706027985  | 182  | 2.9529531002   | 178  | -2%        |                          0.0000 |
| TPC-H 11       | 24.3153686523  | 1459 | 24.1203250885  | 1448 | -1%        |                          0.0000 |
| TPC-H 12       | 7.34165716171  | 441  | 7.20224618912  | 433  | -2%        |                          0.0000 |
| TPC-H 13       | 2.2806186676   | 137  | 2.18179917336  | 131  | -4%        |                          0.0000 |
| TPC-H 14       | 23.7891483307  | 1428 | 23.5943088531  | 1416 | -1%        |                          0.0000 |
| TPC-H 15       | 14.7995157242  | 888  | 14.7773351669  | 887  | -0%        |                          0.0041 |
| TPC-H 16       | 6.90865135193  | 415  | 6.85388183594  | 412  | -1%        |                          0.0000 |
| TPC-H 17       | 1.3856202364   | 84   | 1.37110495567  | 83   | -1%        |                          0.0000 |
| TPC-H 18       | 1.28654658794  | 78   | 1.27460026741  | 77   | -1%        |                          0.0695 |
| TPC-H 19       | 5.55590867996  | 334  | 5.52691984177  | 332  | -1%        |                          0.0085 |
| TPC-H 20       | 2.2780623436   | 137  | 2.42103981972  | 146  | +6%        |                          0.0000 |
| TPC-H 21       | 0.728181242943 | 44   | 1.0785882473   | 65   | +48%       |                          0.0000 |
| TPC-H 22       | 12.9088258743  | 775  | 12.6806812286  | 761  | -2%        |                          0.0000 |
```

Unfortunately, this does not apply to any TPC-DS query - at least none of those that we already support.

Here is the juicy part of TPC-H 21's query plan:
Screenshot 2019-08-19 at 16 43 58


src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -561,10 +563,40 @@ void SubqueryToJoinRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node)
...
   lqp_replace_node(node, join_node);
   join_node->set_right_input(pull_up_result.adapted_lqp);
 
+  if (pull_up_result.adapted_lqp) {
+    const auto& estimator = cost_estimator->cardinality_estimator;
+    const auto left_side_cardinality = estimator->estimate_cardinality(join_node->left_input());
+    const auto right_side_cardinality = estimator->estimate_cardinality(join_node->right_input());
+    const auto& first_join_predicate = join_predicates[0];
+
+    bool first_join_predicate_is_equals = false;
```
Bouncner: Primary?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -561,10 +563,40 @@ void SubqueryToJoinRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node)
...
+    if (const auto predicate_expression = std::dynamic_pointer_cast<BinaryPredicateExpression>(first_join_predicate)) {
+      if (predicate_expression->predicate_condition == PredicateCondition::Equals) {
+        first_join_predicate_is_equals = true;
+      }
+    }
+
+    if ((join_mode == JoinMode::Semi || join_mode == JoinMode::AntiNullAsTrue ||
+         join_mode == JoinMode::AntiNullAsFalse) &&
+        right_side_cardinality > left_side_cardinality * 10 && first_join_predicate_is_equals) {
+      // Semi/Anti joins are currently handled by the hash join, which performs badly if the right side is much
```
Bouncner: For OLTP-type joins with very few rows, it's probably a good idea to avoid the added join? Maybe add a condition like `max(right_side_cardinality, left_side_cardinality) > 1000`?

Bouncner:
Not sure if I fully get the idea yet. 

Before I dive depper: What happens there in Q21? Why does the semi join on the right site add rows while semi-joining the same columns?

Bouncner:
Concerning the naming: this sounds like a "semi join reduction".

mrks:
All change requests make sense to me.

> Not sure if I fully get the idea yet.

I was already afraid that my description was not good enough. :/
Let me try again and tell me if this is better:

> Semi/Anti joins are currently handled by the hash join, which performs badly if the build side is significantly bigger than the probe side. In that case, many of the values on the build side will not be relevant for the join, as they are either not found on the probe side or are duplicates. Including these in a hash map makes both the build and the probe phase unnecessarily expensive.
With the following optimization, we first reduce the build side by throwing out all values that do not exist on the probe side and will thus not be found by the primary predicate of the join, anyway. We do this by adding a second semi join. However, we can only throw away values if the first predicate is an equals predicate. For an example, see TPC-H query 21. [...] For a lack of a better name, we call this "reciprocal semi joins".

> Before I dive depper: What happens there in Q21? Why does the semi join on the right site add rows while semi-joining the same columns?

It doesn't:

Screenshot 2019-08-20 at 09 50 03


> In general, I am not sure that this added trick is better than a properly implemented semi join. Smells like a performance trick that is not necessary when we would "fix" our semi join issues.

A good semi join implementation would do that internally, yes. If we get around to building one, we can easily delete those 34 lines again.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
mrks: We want the left_side_cardinality (renamed to probe_side_cardinality) to be big. I believe the additional criterion should simply be build_side_cardinality > 1000.

Bouncner: Ack.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -561,10 +563,42 @@ void SubqueryToJoinRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node)
...
+    if ((join_mode == JoinMode::Semi || join_mode == JoinMode::AntiNullAsTrue ||
+         join_mode == JoinMode::AntiNullAsFalse) &&
+        build_side_cardinality > probe_side_cardinality * 10 && build_side_cardinality > 1'000 &&
+        primary_join_predicate_is_equals) {
+      // Semi/Anti joins are currently handled by the hash join, which performs badly if the right side is much
+      // bigger than the left side. For that case, we add a second semi join on the right side, which throws out all
+      // values that will not be found by the primary (first) predicate of the later join, anyway. However, we can only
+      // throw away values on the build side if the primary predicate is an equals predicate. For an example, see TPC-H
+      // query 21. That query is the main reason this part exists. The current thresholds are somewhat arbitrary. If we
+      // start to see more of those cases, we can try and find a more complex heuristic. We call this "semi join
```
Bouncner: To make it clearer for anybody reading that in two years, maybe write something like "We add an additional join operation to the plan that behaves similar to a semi join reduction."?

mrks: I find "For that case, we add a second semi join on the build side, which throws out all values that will not be found by the primary (first) predicate of the later join, anyway." clearer that "semi join reduction", which I have not heard of before you mentioned the term. Added a link to a paper that explains it.

mrks:
> In general, I am not sure that this added trick is better than a properly implemented semi join. Smells like a performance trick that is not necessary when we would "fix" our semi join issues.

I gave this some more though and came up with an additional optimization: If we treat semi/anti joins as predicates for the purpose of the predicate reordering, we can push the reduction below the column-vs-column scan and the validate:

b

That gives us another 50% of performance in Q21 and 195% in Q4. Compared to the master, this results in +121% and +250%, respectively.

Is there anything that is keeping us from merging this one first?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -561,10 +563,42 @@ void SubqueryToJoinRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node)
...
+        build_side_cardinality > probe_side_cardinality * 10 && build_side_cardinality > 1'000 &&
+        primary_join_predicate_is_equals) {
+      // Semi/Anti joins are currently handled by the hash join, which performs badly if the right side is much
+      // bigger than the left side. For that case, we add a second semi join on the build side, which throws out all
+      // values that will not be found by the primary (first) predicate of the later join, anyway. However, we can only
+      // throw away values on the build side if the primary predicate is an equals predicate. For an example, see TPC-H
+      // query 21. That query is the main reason this part exists. The current thresholds are somewhat arbitrary. If we
+      // start to see more of those cases, we can try and find a more complex heuristic. This is also called "semi join
+      // reduction": http://www.db.in.tum.de/research/publications/conferences/semijoin.pdf
+      const auto pre_join_node = JoinNode::make(JoinMode::Semi, primary_join_predicate);
```
Bouncner: Took me some time to understand why we use a semi join here even when the initial join was an anti join. Can you add a short comment here to cover that?

mrks: Added

> This is the case no matter if the original join is a semi or anti join. In any case, we want the reducing join introduced here to limit the input to the join operator to those values that have a chance of being relevant for the semi/anti join.

ok?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -561,10 +563,42 @@ void SubqueryToJoinRule::apply_to(const std::shared_ptr<AbstractLQPNode>& node)
...
+      lqp_insert_node(join_node, LQPInputSide::Right, pre_join_node);
```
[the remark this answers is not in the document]
mrks: Added it to the LQP visualization. Teaching the LQPTranslator to "migrate" that comment feels like overkill.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
Bouncner: Ok.

Issue #1820, opened 2019-09-02
Cannot access aggregate with arithmetics outside subquery


janetzki:
The query
```sql
SELECT total_managers
FROM (SELECT i_item_sk,
             count(i_manager_id) + 3 total_managers
      FROM item
      GROUP BY i_item_sk) managers_by_item;
```
crashes with the message `Invalid input error: Failed to resolve Column 'i_manager_id', LQP is invalid` because we cannot access an aggregate that is changed by arithmetic operations from outside a subquery.
There is currently no pressure to fix this, although the error message is misleading.
(This used to be a problem with [query 39](https://github.com/hyrise-mp/tpcds-result-reproduction/commit/91989ede0a8886f8b2545a6c7e8354996e1a89ac#diff-cb0b896c1c63ee0c445ac5cd54b4f59c) in TPC-DS.)

mrks:
```
SELECT x FROM (SELECT c_nationkey, SUM(c_acctbal) + 1 AS x FROM customer GROUP BY c_nationkey) t

[0] [LogicalPlanRootNode]
 \_[1] [Alias] SUM(c_acctbal) + 1 AS x
    \_[2] [Projection] SUM(c_acctbal) + 1
       \_[3] [Aggregate] GroupBy: [] Aggregates: [SUM(c_acctbal)]  <<-- This should not be here
          \_[4] [Projection] c_acctbal
             \_[5] [Alias] c_nationkey, SUM(c_acctbal) + 1 AS x
                \_[6] [Projection] c_nationkey, SUM(c_acctbal) + 1
                   \_[7] [Aggregate] GroupBy: [c_nationkey] Aggregates: [SUM(c_acctbal)]
                      \_[8] [Validate]
                         \_[9] [StoredTable] Name: 'customer' pruned: 0/1 chunk(s), 0/8 column(s)
```

PR #1821, opened 2019-09-02
Execute TPC-DS query 39

janetzki:
Fix hyrise-mp/hyrise/issues/65
The trick was to find a [workaround](https://github.com/hyrise-mp/tpcds-result-reproduction/commit/91989ede0a8886f8b2545a6c7e8354996e1a89ac#diff-cb0b896c1c63ee0c445ac5cd54b4f59c) for #1820. 

Benchmark results on sidon (10 runs):
 - Query 39a: 4.86s
 - Query 39b: 4.92s

PQP for query 39a:
39a-PQP

### DIFF
--- a/resources/benchmark/tpcds/query_blacklist.cfg
+++ b/resources/benchmark/tpcds/query_blacklist.cfg
@@ -39,8 +39,8 @@
 36.sql
 37.sql
 38.sql
-39a.sql
-39b.sql
+#39a.sql
+#39b.sql
 40.sql
 #41.sql
 #42.sql


mweisgut:
Have you checked whether the changes to the query are allowed according to the specification?
If this is not allowed, we should document our changes.

janetzki:
The additional factor that I removed does neither appear in the specification nor in the template file of the TPC-DS kit. Apparently, the owner of our cloned verification repository made this change to make query 39 additionally challenging.

Bensk1:
Adding such a factor sounds weird, maybe there were some data type issues. Anyhow, as long as we are using the version from the TPC-DS specification now everything should be fine. Do we have other queries that are modified and that differ from the specification?

mweisgut:
> Do we have other queries that are modified and that differ from the specification?

At the beginning of the project, we did not use the query repository that we are currently using but used queries generated by the original tpc-ds toolkit. Using these toolkit generated queries, we evaluated which queries are supported by Hyrise. During the project we switched to the query repository that we are using now but did not evaluate the executability of the queries again.

Long story short: it might be that other queries that are currently classified as "not supported by Hyrise" differ from the officially provided query templates.

Bouncner:
Open an issue in the repository to clear that up?

janetzki:
> Open an issue in the repository to clear that up?
https://github.com/hyrise-mp/tpcds-result-reproduction/issues/4

Bouncner:
I meant the third party repo.

janetzki:
You mean in https://github.com/cwida/tpcds-result-reproduction? Why there?

Bouncner:
Just to avoid (semi-)manually comparing the SQL strings. Why not just ask them for the modifications?
Or did I get this issue wrong?

janetzki:
https://github.com/cwida/tpcds-result-reproduction/issues/6
Nevertheless, I reverted the changes from this repo in our clone for now.

PR #1825, opened 2019-09-03
Benchmark: Only add SQLite data if the dedicated result set is incomplete

mweisgut:
This PR integrates the possibility to avoid loading data into SQLite if only dedicated result tables are used for the verification.

Currently, if `--verify` is used, SQLite data is loaded, even if only dedicated results are used.

### DIFF
--- a/scripts/setup_imdb.py
+++ b/scripts/setup_imdb.py
@@ -84,7 +84,7 @@ print("- Validating integrity...")
 
 hash_dl = hash_md5.hexdigest()
 
-if hash_dl != "79e4c71f8ec0dae17d6aa9182fdab835":
+if hash_dl != "270d3b5aa4338c7a452083cf4525c1cd":
     print("  Aborting. MD5 checksum mismatch. Cleaning up.")
     clean_up()
     sys.exit(2)
--- a/scripts/setup_imdb.py
+++ b/scripts/setup_imdb.py
@@ -84,7 +84,7 @@ print("- Validating integrity...")
 
 hash_dl = hash_md5.hexdigest()
 
-if hash_dl != "270d3b5aa4338c7a452083cf4525c1cd":
+if hash_dl != "79e4c71f8ec0dae17d6aa9182fdab835":
     print("  Aborting. MD5 checksum mismatch. Cleaning up.")
     clean_up()
     sys.exit(2)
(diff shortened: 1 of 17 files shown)


Bensk1:
Can we discuss that tomorrow in person?

scripts/setup_imdb.py
```diff
@@ -84,7 +84,7 @@ def is_setup():
 
 hash_dl = hash_md5.hexdigest()
 
-if hash_dl != "79e4c71f8ec0dae17d6aa9182fdab835":
+if hash_dl != "270d3b5aa4338c7a452083cf4525c1cd":
```
Bensk1: This change should disappear when merging the master into the feature branch.

src/benchmarklib/abstract_benchmark_item_runner.cpp
```diff
@@ -38,6 +38,18 @@ void AbstractBenchmarkItemRunner::load_dedicated_expected_results(
   }
 }
 
+bool AbstractBenchmarkItemRunner::has_item_without_dedicated_result() {
+  if (items().size() > _dedicated_expected_results.size()) {
```
Bensk1: Regarding what we discussed in person yesterday: does that mean that there is not an element in `_dedicated_expected_results` per item?

mweisgut: `_dedicated_expected_results` is only a sparse vector if `load_dedicated_expected_results` was called. If it's not called, `load_dedicated_expected_results` is empty. The load-function is only called during the TPC-DS execution.

Bensk1: I understand. Does that mean, we could also check for `_dedicated_expected_results.empty()`?

mweisgut: The function name could be misleading then: if no items exists and thus no dedicated result, `has_item_without_dedicated_result` would return true.

Bensk1: Right. We could add a comment to the current version of the method or to the header file where `_dedicated_expected_results` is declared, saying that we either have an empty or a sparse vector with the same size as `items()` to make clear that `_dedicated_expected_results.size()` is either `0` or `items().size()`.

PR #1832, opened 2019-09-04
AbstractJoinOrderingAlgorithm does not use MPJoin - Clarify Comment

mweisgut:
The comment we talked about this morning is actually correct. The added note shall clarify it.

https://github.com/mweisgut/hyrise/blob/4052ddf18d36708fc459f0a9f1f3842af0446f92/src/lib/optimizer/join_ordering/abstract_join_ordering_algorithm.cpp#L83-L97


Bensk1:
Are there no cases where we use the MultiPredicateJoin?

PR #1836, opened 2019-09-07
Add const LQP support to visit_lqp() and clean up LQP hashing

mrzzzrm:
* Remove necessity to use const cast by making visit_lqp() support const LQPs
* HashEquals -> HashingAndEqualityCheck because "HashEquals" to me sound like "test whether hashes are equal", not "test whether hashing and equality checks work"
* Remove the left_input==right_input seed from the LQP hash - the only test that this broke was a IMO malformed test with an UpdateNode with only one input (should always be two, right?)

### DIFF
--- a/src/lib/logical_query_plan/abstract_lqp_node.hpp
+++ b/src/lib/logical_query_plan/abstract_lqp_node.hpp
@@ -145,10 +145,7 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
   bool operator!=(const AbstractLQPNode& rhs) const;
 
   /**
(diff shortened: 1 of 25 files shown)


src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -145,10 +145,7 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
   bool operator!=(const AbstractLQPNode& rhs) const;
 
   /**
-   *  Builds a hash code by hashing the node type combined with specific member variables of
```
mrzzzrm: API docs should doc the "what", not the "how"

PR #1841, opened 2019-09-13
Fix chunk size in benchmarks, add join side in viz

### LINKED ISSUE #1840: Chunk size ignored for sorted tables in benchmark
Fix:

```
--- a/src/benchmarklib/abstract_table_generator.cpp
+++ b/src/benchmarklib/abstract_table_generator.cpp
@@ -54,7 +54,7 @@ void AbstractTableGenerator::generate_and_store() {
       auto& table = table_info_by_name[table_name].table;
       auto table_wrapper = std::make_shared<TableWrapper>(table);
       table_wrapper->execute();
-      auto sort = std::make_shared<Sort>(table_wrapper, table->column_id_by_name(column_name));
+      auto sort = std::make_shared<Sort>(table_wrapper, table->column_id_by_name(column_name), OrderByMode::Ascending, _benchmark_config->chunk_size);
       sort->execute();
       const auto immutable_sorted_table = sort->get_output();
       table = std::make_shared<Table>(immutable_sorted_table->column_definitions(), TableType::Data,
```

Consider adding an end-to-end test.

mrks:
fixes #1840 

Also, as the query plans become more complex, this changes the arrowheads of multi-input nodes to identify the input sides. This is so that we don't mistake the visually left input of this semi join to be the probe side.

### DIFF
--- a/scripts/test/hyriseBenchmarkTPCH_test.py
+++ b/scripts/test/hyriseBenchmarkTPCH_test.py
@@ -72,6 +72,7 @@ def main():
 
   arguments = {}
   arguments["--scale"] = ".005"
+  arguments["--chunk_size"] = "10000"
   arguments["--queries"] = "'2,4,6'"
   arguments["--time"] = "10"
   arguments["--runs"] = "100"
@@ -91,7 +92,7 @@ def main():
   benchmark.expect("Running benchmark in 'Ordered' mode")
   benchmark.expect("Visualizing the plans into SVG files. This will make the performance numbers invalid.")
   benchmark.expect("Encoding is 'LZ4'")
-  benchmark.expect("Chunk size is 100000")
+  benchmark.expect("Chunk size is 10000")
   benchmark.expect("Max runs per item is 100")
   benchmark.expect("Max duration per item is 10 seconds")
   benchmark.expect("Warmup duration per item is 10 seconds")
@@ -102,6 +103,17 @@ def main():
 
   close_benchmark(benchmark)
   check_exit_status(benchmark)
+
+  visualization_file = 'TPC-H_06-PQP.svg'
+  if not os.path.isfile(visualization_file):
+    print ("ERROR: Cannot find visualization file " + visualization_file)
+    sys.exit(1)
+  with open(visualization_file) as f:
+    # Check whether the (a) the GetTable node exists and (b) the chunk count is correct for the given scale factor
+    if not '2/4 chunk(s)' in f:
+      print ("ERROR: Did not find expected pruning information in the visualization file")
+      sys.exit(1)
+
 
   if return_error:
     sys.exit(1)
--- a/scripts/test/hyriseBenchmarkTPCH_test.py
+++ b/scripts/test/hyriseBenchmarkTPCH_test.py
@@ -110,7 +110,7 @@ def main():
     sys.exit(1)
   with open(visualization_file) as f:
     # Check whether the (a) the GetTable node exists and (b) the chunk count is correct for the given scale factor
-    if not '/4 chunk(s)' in f:
+    if not '/4 chunk(s)' in f.read():
       print ("ERROR: Did not find expected pruning information in the visualization file")
       sys.exit(1)
 
--- a/scripts/test/hyriseBenchmarkTPCH_test.py
+++ b/scripts/test/hyriseBenchmarkTPCH_test.py
@@ -110,7 +110,7 @@ def main():
     sys.exit(1)
   with open(visualization_file) as f:
     # Check whether the (a) the GetTable node exists and (b) the chunk count is correct for the given scale factor
-    if not '2/4 chunk(s)' in f:
+    if not '/4 chunk(s)' in f:
       print ("ERROR: Did not find expected pruning information in the visualization file")
       sys.exit(1)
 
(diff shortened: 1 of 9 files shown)


scripts/test/hyriseBenchmarkTPCH_test.py
```diff
@@ -103,6 +104,17 @@ def main():
...
   check_exit_status(benchmark)
 
+  visualization_file = 'TPC-H_06-PQP.svg'
+  if not os.path.isfile(visualization_file):
+    print ("ERROR: Cannot find visualization file " + visualization_file)
+    sys.exit(1)
+  with open(visualization_file) as f:
+    # Check whether the (a) the GetTable node exists and (b) the chunk count is correct for the given scale factor
+    if not '/4 chunk(s)' in f:
+      print ("ERROR: Did not find expected pruning information in the visualization file")
```
Bouncner: The error does not really relate to the check.

src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -54,7 +54,8 @@ void AbstractTableGenerator::generate_and_store() {
       auto& table = table_info_by_name[table_name].table;
       auto table_wrapper = std::make_shared<TableWrapper>(table);
       table_wrapper->execute();
-      auto sort = std::make_shared<Sort>(table_wrapper, table->column_id_by_name(column_name));
+      auto sort = std::make_shared<Sort>(table_wrapper, table->column_id_by_name(column_name), OrderByMode::Ascending,
+                                         _benchmark_config->chunk_size);
```
Bouncner: The chunk size is not required, isn't it? Any reason to set it to this specific value (chunk_size * 10)?

src/lib/visualization/lqp_visualizer.cpp
```diff
@@ -119,6 +119,9 @@ void LQPVisualizer::_build_dataflow(const std::shared_ptr<AbstractLQPNode>& from
   VizEdgeInfo info = _default_edge;
   info.label = label_stream.str();
   info.pen_width = pen_width;
+  if (to->input_count() == 2) {
+    info.arrowhead = side == InputSide::Left ? "normallbox" : "normalrdiamond";
```
Bouncner: What about capped arrows? I know that I will have to lookup this port thing each and every time. :-/
image?

scripts/test/hyriseBenchmarkTPCH_test.py
mrks: Why not? We are looking for the text that is part of the node description.

src/benchmarklib/abstract_table_generator.cpp
mrks: Where do you get `chunk_size * 10` from?

src/lib/visualization/lqp_visualizer.cpp
mrks: My concern is that really weird plans might have the arrows come in from a different direction than below. In that case, left and right are hard to determine. However, the current layout also has the identifier on the left / right side of the arrow.

scripts/test/hyriseBenchmarkTPCH_test.py
Bouncner: Hmm .. ok

src/benchmarklib/abstract_table_generator.cpp
Bouncner: Never mind, I confused our default chunk size with that of 10k here. Nonetheless, why setting it?

mrks: Because Sort materializes the table and does not check for the size of the input table.

src/lib/visualization/lqp_visualizer.cpp
mrks: Never mind. I'm stupid.

Bouncner: Ah right. Ok ... let's see how it plays out over time.

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
```diff
@@ -297,7 +297,7 @@ std::string TPCHBenchmarkItemRunner::_build_query(const BenchmarkItemID item_id)
     }
 
     case 14 - 1: {
-      std::uniform_int_distribution<> date_diff_dist{0, 47};
+      std::uniform_int_distribution<> date_diff_dist{0, 5*12};
```
mrks: DATE is the first day of a month randomly selected from a random year within [1993 .. 1997].

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
```diff
@@ -196,8 +196,8 @@ std::string TPCHBenchmarkItemRunner::_build_query(const BenchmarkItemID item_id)
       const auto begin_date = calculate_date(boost::gregorian::date{1993, 01, 01}, diff * 12);
       const auto end_date = calculate_date(boost::gregorian::date{1993, 01, 01}, (diff + 1) * 12);
 
-      static std::uniform_real_distribution<> discount_dist{0.02f, 0.09f};
-      const auto discount = discount_dist(random_engine);
+      static std::uniform_int_distribution<> discount_dist{2, 9};
+      const auto discount = 0.01f * discount_dist(random_engine);
```
mrks: DISCOUNT is randomly selected within [0.02 .. 0.09];

PR #1842, opened 2019-09-14
Two expression resolution fixes in SQLTranslator

### LINKED ISSUE #1771: Changes to identifier resolution broke query
```
(release)> generate_tpcc
Generating TPCC tables (this might take a while) ...
(release)> SELECT C_W_ID, C_D_ID, C_ID, MAX(C_BALANCE), (CASE WHEN SUM_OL_AMOUNT IS NULL THEN 0 ELSE SUM_OL_AMOUNT END) AS SUM_OL_AMOUNT_NONNULL FROM CUSTOMER LEFT JOIN (SELECT O_W_ID, O_D_ID, O_C_ID, SUM(OL_AMOUNT) FROM "ORDER", ORDER_LINE WHERE OL_W_ID = O_W_ID AND OL_D_ID = O_D_ID AND OL_O_ID = O_ID AND OL_DELIVERY_D <> -1 GROUP BY O_W_ID, O_D_ID, O_C_ID) AS sub1(O_W_ID, O_D_ID, O_C_ID, SUM_OL_AMOUNT) ON O_W_ID = C_W_ID AND O_D_ID = C_D_ID AND O_C_ID = C_ID GROUP BY C_W_ID, C_D_ID, C_ID, SUM_OL_AMOUNT_NONNULL;
libc++abi.dylib: terminating with uncaught exception of type std::logic_error: src/lib/sql/sql_identifier_resolver.cpp:82 The expression does not exist.
```

"The expression" being `SUM(OL_AMOUNT)`.

Works on 7fa0d1bf68cf666f0460e2991a68272bc253223f, broken on e71169a1e18f73b03ac9054af0f5e3e1d04cf944. Will try to reduce the test case, but not today.

Mentioned query formatted:
```sql
SELECT
   C_W_ID,
   C_D_ID,
   C_ID,
   MAX(C_BALANCE),
   (
      CASE
         WHEN
            SUM_OL_AMOUNT IS NULL 
         THEN
            0 
         ELSE
            SUM_OL_AMOUNT 
      END
   )
   AS SUM_OL_AMOUNT_NONNULL 
FROM
   CUSTOMER 
   LEFT JOIN
      (
         SELECT
            O_W_ID,
            O_D_ID,
            O_C_ID,
            SUM(OL_AMOUNT) 
         FROM
            "ORDER",
            ORDER_LINE 
         WHERE
            OL_W_ID = O_W_ID 
            AND OL_D_ID = O_D_ID 
            AND OL_O_ID = O_ID 
            AND OL_DELIVERY_D <> - 1 
         GROUP BY
            O_W_ID,
            O_D_ID,
            O_C_ID
      )
      AS sub1(O_W_ID, O_D_ID, O_C_ID, SUM_OL_AMOUNT) 
      ON O_W_ID = C_W_ID 
      AND O_D_ID = C_D_ID 
      AND O_C_ID = C_ID 
GROUP BY
   C_W_ID,
   C_D_ID,
   C_ID,
   SUM_OL_AMOUNT_NONNULL
```

### LINKED ISSUE #1799: Division should always return a float/double
```
(debug)> SELECT 5/3
...
|       1|
```

No.

### LINKED ISSUE #1820: Cannot access aggregate with arithmetics outside subquery

mrks:
fixes #1771
fixes #1820
fixes #1799

### DIFF
--- a/src/lib/sql/sql_identifier_resolver.cpp
+++ b/src/lib/sql/sql_identifier_resolver.cpp
@@ -17,8 +17,10 @@ void SQLIdentifierResolver::add_column_name(const std::shared_ptr<AbstractExpres
 }
 
 void SQLIdentifierResolver::reset_column_names(const std::shared_ptr<opossum::AbstractExpression>& expression) {
-  auto& entry = _find_expression_entry(expression);
-  entry.column_names.clear();
+  auto entry_iter = std::find_if(_entries.begin(), _entries.end(),
+                                 [&](const auto& entry) { return *entry.expression == *expression; });
+  if (entry_iter == _entries.end()) return;
+  entry_iter->column_names.clear();
 }
 
 void SQLIdentifierResolver::set_table_name(const std::shared_ptr<AbstractExpression>& expression,
@@ -73,14 +75,6 @@ std::vector<std::shared_ptr<AbstractExpression>> SQLIdentifierResolver::resolve_
 
 void SQLIdentifierResolver::append(SQLIdentifierResolver&& rhs) {
   _entries.insert(_entries.end(), rhs._entries.begin(), rhs._entries.end());
-}
-
-SQLIdentifierContextEntry& SQLIdentifierResolver::_find_expression_entry(
-    const std::shared_ptr<AbstractExpression>& expression) {
-  auto entry_iter = std::find_if(_entries.begin(), _entries.end(),
-                                 [&](const auto& entry) { return *entry.expression == *expression; });
-  Assert(entry_iter != _entries.end(), "The expression does not exist.");
-  return *entry_iter;
 }
 
 SQLIdentifierContextEntry& SQLIdentifierResolver::_find_or_create_expression_entry(
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -867,10 +867,8 @@ void SQLTranslator::_translate_select_groupby_having(const hsql::SelectStatement
   auto aggregate_expression_set = ExpressionUnorderedSet{};
   auto aggregate_expressions = std::vector<std::shared_ptr<AbstractExpression>>{};
 
-  // Visitor that identifies AggregateExpressions and their arguments.
-  const auto find_aggregates_and_arguments = [&](auto& sub_expression) {
-    if (sub_expression->type != ExpressionType::Aggregate) return ExpressionVisitation::VisitArguments;
-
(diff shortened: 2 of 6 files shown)


src/lib/sql/sql_identifier_resolver.cpp
```diff
@@ -17,8 +17,10 @@ void SQLIdentifierResolver::add_column_name(const std::shared_ptr<AbstractExpres
 }
 
 void SQLIdentifierResolver::reset_column_names(const std::shared_ptr<opossum::AbstractExpression>& expression) {
-  auto& entry = _find_expression_entry(expression);
-  entry.column_names.clear();
+  auto entry_iter = std::find_if(_entries.begin(), _entries.end(),
+                                 [&](const auto& entry) { return *entry.expression == *expression; });
+  if (entry_iter == _entries.end()) return;
+  entry_iter->column_names.clear();
```
mrks: Taken from `_find_expression_entry`, which was not used anywhere else. Only change is that the assertion is removed. An expression might not have a name yet (because it is, e.g., an aggregate), but we want to call `reset_column_names` indiscriminately.

src/lib/sql/sql_translator.cpp
```diff
@@ -879,6 +877,8 @@ void SQLTranslator::_translate_select_groupby_having(const hsql::SelectStatement
      */
     if (_current_lqp->find_column_id(*sub_expression)) return ExpressionVisitation::DoNotVisitArguments;
 
+    if (sub_expression->type != ExpressionType::Aggregate) return ExpressionVisitation::VisitArguments;
```
mrks: Recurse into the arguments only if the entire expression was not already computed (line before)

mrks:
@janetzki Since you were the last one to look at the SQL identifiers, do you think you could spend five minutes reviewing this?

mrks:
Also sneaked in a one-line comment that clarifies #1799.

PR #1843, opened 2019-09-16
Update Google Test

Bouncner:
Updates the third party module of the Google Test framework.

With it goes the necessity to add an empty item at the list of some test instantiations and thus also some `NOLINT` are removed.

For `TYPED_TEST_SUITE` the problem remains (https://github.com/google/googletest/issues/2271).


Bouncner:
Number of executed tests are unchanged for clang (debug and release, hyriseTest, hyriseSystemsTest) and gcc.

PR #1844, opened 2019-09-16
Add DESCRIBE as an alias for SHOW COLUMNS

### LINKED ISSUE #1787: Describe a table
It seems `hyrise` doesn't support `DESCRIBE` a table yet, though I think this command is useful for people to play around in the database and not hard to implement.

mrks:
* fixes #1787 
* fixes predicates on meta tables
* fixes UB if `get_result_table` was called on xvalue pipeline:
```
      const auto result = SQLPipelineBuilder{"SELECT * FROM int_int"}
                             .create_pipeline()
                             .get_result_table();
```
As `get_result_table` would return a pair containing a reference to the result table, which is stored in the now expired pipeline, accessing the table was previously UB. We now copy the `shared_ptr<Table>` if necessary.

### DIFF
--- a/src/lib/sql/sql_pipeline.cpp
+++ b/src/lib/sql/sql_pipeline.cpp
@@ -204,6 +204,11 @@ std::pair<SQLPipelineStatus, const std::shared_ptr<const Table>&> SQLPipeline::g
   return {SQLPipelineStatus::Success, tables.back()};
 }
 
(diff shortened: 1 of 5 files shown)


Bouncner:
Is the actual work (the alias) completely hidden in parser commit?

mrks:
Yes. The parser translates it as if you wrote `SHOW COLUMNS`.

PR #1846, opened 2019-09-16
Clean up hyriseSystemTest and unify encoding enumerations

### LINKED ISSUE #1673: Reduce places where new segment types need to be added
LZ4 should have been added to `any_segment_iterable_test.cpp` when it was originally introduced. The fact that it wasn't shows that we have too many places where new segment types have to be added.

_Originally posted by @mrks in https://github.com/hyrise/hyrise/pull/1657/review_comment/create_

Bouncner:
This PR limits the testing of encoding types for sanitizer runs of hyriseSystemTest.

Further, it fixes #1673 by adding two fixed arrays of encoding types and all segment encoding specs to `encoding_types.hpp`. Tests now iterate over these lists (sometimes with explicit filtering in case some encodings to not support data types) instead of the cumbersome and error-prone listing of encodings (which lead to cases that encodings haven't been tested).

### DIFF
--- a/src/lib/storage/encoding_type.hpp
+++ b/src/lib/storage/encoding_type.hpp
@@ -8,6 +8,7 @@
 #include <boost/hana/tuple.hpp>
 #include <boost/hana/type.hpp>
 
+#include <array>
 #include <cstdint>
 
 #include "all_type_variant.hpp"
@@ -66,5 +67,22 @@ struct SegmentEncodingSpec {
 };
 
 using ChunkEncodingSpec = std::vector<SegmentEncodingSpec>;
+
+constexpr std::array<EncodingType, 6> all_encoding_types = {
+    EncodingType::Unencoded,        EncodingType::Dictionary,
+    EncodingType::FrameOfReference, EncodingType::FixedStringDictionary,
+    EncodingType::RunLength,        EncodingType::LZ4};
+
+constexpr std::array<SegmentEncodingSpec, 10> all_segment_encoding_specs = {{
+    {EncodingType::Unencoded},
+    {EncodingType::Dictionary, VectorCompressionType::FixedSizeByteAligned},
+    {EncodingType::Dictionary, VectorCompressionType::SimdBp128},
+    {EncodingType::FrameOfReference, VectorCompressionType::FixedSizeByteAligned},
+    {EncodingType::FrameOfReference, VectorCompressionType::SimdBp128},
+    {EncodingType::FixedStringDictionary, VectorCompressionType::FixedSizeByteAligned},
+    {EncodingType::FixedStringDictionary, VectorCompressionType::SimdBp128},
+    {EncodingType::LZ4, VectorCompressionType::FixedSizeByteAligned},
+    {EncodingType::LZ4, VectorCompressionType::SimdBp128},
+    {EncodingType::RunLength}}};
 
 }  // namespace opossum
--- a/src/lib/storage/encoding_type.hpp
+++ b/src/lib/storage/encoding_type.hpp
@@ -68,21 +68,19 @@ struct SegmentEncodingSpec {
 
 using ChunkEncodingSpec = std::vector<SegmentEncodingSpec>;
 
-constexpr std::array<EncodingType, 6> all_encoding_types = {
-    EncodingType::Unencoded,        EncodingType::Dictionary,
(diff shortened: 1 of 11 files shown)


mrks:
#1673 ...

Bouncner:
So the Full CI run took 2h 20min. The longest running tests (with hyriseSystemTest and sanitisers) ran 1h 20min. Before it took 50min.

The numbers fluctuate a lot, so it's hard to make a really good estimation. But is that something we are will to invest?

Bouncner:
Maybe as a side note: even though I might be the only one at the moment, I am working on  LZ4 segments and use them quite a lot. So it's definitely not "dead" code, even though their performance make them the last choice for unlimited-DRAM benchmarks.

mrks:
I wouldn't say that it's worth it. I'm even tempted to question whether we should test all encodings in the sanitized SQLite tests at all.

Bouncner:
For me testing only dictionary would be fine.

mrks:
Shall we exclude TPCHTestNoJITPreparedStatements, too?

Bouncner:
Your call. I don't understand anything JIT-related.

src/lib/storage/encoding_type.hpp
```diff
@@ -67,4 +68,21 @@ struct SegmentEncodingSpec {
 
 using ChunkEncodingSpec = std::vector<SegmentEncodingSpec>;
 
+constexpr std::array<EncodingType, 6> all_encoding_types = {
```
mrks: Shouldn't the size be deducible?

src/test/storage/any_segment_iterable_test.cpp
```diff
@@ -27,15 +28,17 @@ class AnySegmentIterableTest : public BaseTestWithParam<SegmentEncodingSpec> {
                            {ChunkID{0}, ChunkOffset{1}}, {ChunkID{0}, ChunkOffset{1}}, {ChunkID{0}, ChunkOffset{5}}};
     position_filter = std::make_shared<PosList>(std::move(row_ids));
     position_filter->guarantee_single_chunk();
+
+    std::copy_if(all_segment_encoding_specs.begin(), all_segment_encoding_specs.end(),
+                 std::back_inserter(int_supporting_segment_encodings), [](SegmentEncodingSpec spec) {
+                   return encoding_supports_data_type(spec.encoding_type, DataType::Int);
```
mrks: I find a simple loop easier to read.

Bouncner: Hmpf ... did not check the result of formatting. The one liner was much nicer than this. :(

Will change it back to a for loop.

src/lib/storage/encoding_type.hpp
Bouncner: Is there is nicer way to help the compiler deduce the type than that?
```
constexpr std::array all_segment_encoding_specs{std::initializer_list<SegmentEncodingSpec>{
    {EncodingType::Unencoded},
    {EncodingType::Dictionary, VectorCompressionType::FixedSizeByteAligned},
    {EncodingType::Dictionary, VectorCompressionType::SimdBp128},
    {EncodingType::FrameOfReference, VectorCompressionType::FixedSizeByteAligned},
    {EncodingType::FrameOfReference, VectorCompressionType::SimdBp128},
    {EncodingType::FixedStringDictionary, VectorCompressionType::FixedSizeByteAligned},
    {EncodingType::FixedStringDictionary, VectorCompressionType::SimdBp128},
    {EncodingType::LZ4, VectorCompressionType::FixedSizeByteAligned},
    {EncodingType::LZ4, VectorCompressionType::SimdBp128},
    {EncodingType::RunLength}}};
```

mrks: Uh, not sure. Works for me.

mrks:
>  Bouncner removed the FullCI label 11 hours ago

Did you mean to remove `WIP` by chance?

Bouncner:
Nope, found an issue with LZ4 that I’d like to address first.

Bouncner:
@mrks : feel free to review again. Cleaned up (fixed) the filtering of encodings in the tests.

We gained a few tests, but the coverage should not change since most parts have been implicitly covered already. I checked the number of tests and found no problems.

The LZ4 problem remains (it's not a bug, just an unexpected vector compression), but it is never triggered. I will take a lot at that at another point in time.

PR #1848, opened 2019-09-17
Fix performance regression in JOB

### LINKED ISSUE #1839: investigate performance decline in JoinOrderBenchmark
measures on sidon, hyrise/hyrise master a0878739f vs hyrise-mp/hyrise feature/final_presentation 02ac2aa42:
https://docs.google.com/spreadsheets/d/1fGscgkbUK2S3rYzqPhGL-YdV5pdj-I4Hv1ue_6SOyyw/edit?usp=sharing

mrks:
fixes #1839 - we still have some queries that run slower, but the -99% parts are gone. Given that we don't really optimize for the JOB, that should be good enough for now.

Splitting up `x LIKE '...' OR x LIKE '...'` into two predicates (#1791) enables the TableScan to use the ColumnLikeTableScanImpl. That implementation first builds a bitmap of the entire dictionary, then checks the value id against that bitmap. If the number of input rows is significantly lower than the size of the dictionary (because the input has already been filtered), that is a bad idea. Now, we only scan the entire dictionary if the filtered segment has no more unique values in the dictionary than rows in the input.

old:
Screenshot 2019-09-17 at 13 33 12

regression:
Screenshot 2019-09-17 at 13 35 03

new:
Screenshot 2019-09-17 at 13 54 02

### DIFF
--- a/src/lib/operators/table_scan/column_like_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_like_table_scan_impl.cpp
@@ -32,8 +32,9 @@ std::string ColumnLikeTableScanImpl::description() const { return "ColumnLike"; 
 void ColumnLikeTableScanImpl::_scan_non_reference_segment(const BaseSegment& segment, const ChunkID chunk_id,
                                                           PosList& matches,
                                                           const std::shared_ptr<const PosList>& position_filter) const {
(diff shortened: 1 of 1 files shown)


Bouncner:
So what did you do?

I see only the problem description and an immense performance boost.

PR #1853, opened 2019-09-17
Subquery in arithmetic expression to Join

pascalhpi:
The SubqueryToJoin rule could not deal with eg. `a > 1.2 * subquery`, because the subquery is part of a simple arithmetic expression. This PR fixes that. The idea is to push the arithmetic operation into the subquery (as a projection) to get to a case that the SubqueryToJoin rule can reformulate. 
TPC-DS improvements:
Screenshot from 2019-09-17 15-59-26


src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -223,10 +223,76 @@ std::optional<SubqueryToJoinRule::PredicateNodeInfo> SubqueryToJoinRule::is_pred
...
     result.join_mode = JoinMode::Semi;
 
+    /**
+     * Identify a subquery in an arithmetic expression and push the arithmetics into the subquery.
+     * e.g. SELECT * FROM a WHERE a.a > 3 * (SELECT SUM(b.a) FROM b WHERE b.b = a.b)
+     * becomes SELECT * FROM a WHERE a.a > (SELECT 3 * SUM(b.a) FROM b WHERE b.b = a.b)
+     * There are 4 cases that are essentially the same: The subquery could be in the left/right operand of the binary
+     * predicate and then it could be the left/right operand of that arithmetic expression.
+     */
+    if (const auto right_arithmetic_expression =
```
mrks: This should be easy to de-duplicate if instead of left/right_operand, you use arguments.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -223,10 +223,76 @@ std::optional<SubqueryToJoinRule::PredicateNodeInfo> SubqueryToJoinRule::is_pred
...
-                 std::dynamic_pointer_cast<BinaryPredicateExpression>(predicate_node.predicate())) {
+  } else if (auto binary_predicate = std::dynamic_pointer_cast<BinaryPredicateExpression>(predicate_node.predicate())) {
     result.join_mode = JoinMode::Semi;
 
+    /**
+     * Identify a subquery in an arithmetic expression and push the arithmetics into the subquery.
+     * e.g. SELECT * FROM a WHERE a.a > 3 * (SELECT SUM(b.a) FROM b WHERE b.b = a.b)
+     * becomes SELECT * FROM a WHERE a.a > (SELECT 3 * SUM(b.a) FROM b WHERE b.b = a.b)
+     * There are 4 cases that are essentially the same: The subquery could be in the left/right operand of the binary
+     * predicate and then it could be the left/right operand of that arithmetic expression.
```
mrks: What happens to `SELECT * FROM a WHERE a.a > 3 * (SELECT SUM(b.a) FROM b WHERE b.b = a.b) * 3`?

Not saying that we necessarily need to cover that, but there sure are more than four cases.

pascalhpi: We only cover simple cases for now. I updated the comment.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
pascalhpi: I moved the duplicate code into a helper function.

mrks: Much better

src/lib/optimizer/strategy/join_predicate_ordering_rule.hpp
```diff
@@ -9,6 +9,7 @@ namespace opossum {
 
 /**
  * A rule that brings join predicates of a multi-predicate-join into an efficient order.
+
```
mrks: missing `*`

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -199,6 +199,26 @@ std::pair<SubqueryToJoinRule::PredicatePullUpResult, bool> pull_up_correlated_pr
   return {result, false};
 }
 
+void push_arithmetic_expression_into_subquery(std::shared_ptr<BinaryPredicateExpression>& predicate_expression,
+                                              const std::shared_ptr<ArithmeticExpression>& arithmetic_expression,
+                                              std::shared_ptr<LQPSubqueryExpression>& subquery_expression,
+                                              size_t arithmetic_expression_position, bool subquery_expression_is_left) {
```
mrks: const parameters

mrks: rename `arithmetic_expression_position` to `arithmetic_expression_argument_idx`?

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -199,6 +199,26 @@ std::pair<SubqueryToJoinRule::PredicatePullUpResult, bool> pull_up_correlated_pr
...
+  // Create a copy of the arithmetic expression which has the subquery expression replaced with the subquery's result.
```
mrks: What is "the subquery's result"? Suggestion:

```
// (1) Create a new arithmetic expression that takes the result expression of the subquery and applie the arithmetic operation on it
// (2) Insert a ProjectionNode on top of the subquery's LQP that evaluates this arithmetic expression
// (3) In the PredicateExpression outside of the LQP, replace the old arithmetic expression with the updated subquery expression

File: src/lib/optimizer/strategy/subquery_to_join_rule.cpp [hunk 6]
```diff
@@ -199,6 +199,26 @@ std::pair<SubqueryToJoinRule::PredicatePullUpResult, bool> pull_up_correlated_pr
...
+  const auto new_arithmetic_expression = std::make_shared<ArithmeticExpression>(
+      arithmetic_expression->arithmetic_operator,
+      subquery_expression_is_left ? subquery_expression->lqp->node_expressions[0]
```
mrks: Even though a broken LQP should never make it this far, whenever you explicitly take the first element and ignore all others, you should (debug)assert that the vector's size is exactly 1.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -223,10 +243,44 @@ std::optional<SubqueryToJoinRule::PredicateNodeInfo> SubqueryToJoinRule::is_pred
...
-                 std::dynamic_pointer_cast<BinaryPredicateExpression>(predicate_node.predicate())) {
+  } else if (auto binary_predicate = std::dynamic_pointer_cast<BinaryPredicateExpression>(predicate_node.predicate())) {
     result.join_mode = JoinMode::Semi;
 
+    /**
+     * Identify a subquery in an arithmetic expression and push the arithmetics into the subquery.
+     * e.g. SELECT * FROM a WHERE a.a > 3 * (SELECT SUM(b.a) FROM b WHERE b.b = a.b)
+     * becomes SELECT * FROM a WHERE a.a > (SELECT 3 * SUM(b.a) FROM b WHERE b.b = a.b)
+     * We cover 4 cases that are essentially the same: The subquery could be in the left/right operand of the binary
+     * predicate and then it could be the left/right operand of that arithmetic expression.
```
mrks: + "More complex cases, such as `1 + (SELECT ... ) + 2` are not covered yet."

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
```diff
@@ -223,10 +243,44 @@ std::optional<SubqueryToJoinRule::PredicateNodeInfo> SubqueryToJoinRule::is_pred
...
+      if (auto subquery_expression =
+              std::dynamic_pointer_cast<LQPSubqueryExpression>(arithmetic_expression->right_operand())) {
+        push_arithmetic_expression_into_subquery(binary_predicate, arithmetic_expression, subquery_expression, 1,
+                                                 false);
+      }
+
+      if (auto subquery_expression =
+              std::dynamic_pointer_cast<LQPSubqueryExpression>(arithmetic_expression->left_operand())) {
+        push_arithmetic_expression_into_subquery(binary_predicate, arithmetic_expression, subquery_expression, 1, true);
+      }
```
mrks: If I see this correctly, both `if` cases can be true: `WHERE a > (SELECT ...) + (SELECT ...)`. Wouldn't this lead to the left subquery being pushed into the right one, the right one bein̨̓g̰̒ ̔ͅṕ͈u̞͆s̻͘ḧ̤e͓͠d̤͐ ̝̕i͙̎n̝̅t̯̅o̻̾ ͚̉t̩̍h̗͡e̾͜ ̞̈l̢̉e̮͘f͖͂t̬̉ ͟͠ò̯ñ͕e͕̚,̞͠ ͔̾a̠͠n̲͞d̟̀ ̳̆t̗͛h̢̾u̢͊s͖̈ ̳͆t̰͡h͍̀ê̡ ̜̎ō̮p̢̅e̖͂n͓͐i̗͆n̍ͅg̩̚ ̨̌o̬̊f̬̉ ̧͌t̲̀h̖͊e̗̍ ̣͞a̫̓b͈̏ÿ̖́s̨͛s͈̎?͍̔ We should fix this and test for that.

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
mrks: This doesn't seem to be addressed in the last commit

src/lib/optimizer/strategy/subquery_to_join_rule.cpp
pascalhpi: Yes, you are right, that would lead to a stack overflow. I changed it to `else if`, so now we push the left subquery into the right one and that's it. I also added a test for that.

PR #1570, opened 2019-03-30
[DYOD18/19] Unique Constraints

nilsthamm:
This PR is the final state of the unique constraint groups works. It summarizes the changes from these three PRs: #1416, #1495, #1565

All PRs already have been reviewed, this PR is to enable others an overview of all changes together (@mrzzzrm 😉 )

PR #1854, opened 2019-09-18
Refactor ColumnPruningRule, add basic constraints

### LINKED ISSUE #1898: Different LQPs are equal when compared
The test
```cpp
node_a = MockNode::make(MockNode::ColumnDefinitions{{DataType::Int, "a"}}, "a");
a_a = node_a->get_column("a");

node_b = MockNode::make(MockNode::ColumnDefinitions{{DataType::Int, "a"}}, "b");
b_a = node_b->get_column("a");

// clang-format off
const auto actual_lqp =
PredicateNode::make(equals_(a_a, 10),
  node_a);

const auto expected_lqp =
PredicateNode::make(equals_(b_a, 10),
  node_b);
// clang-format on

```
should fail but actually succeeds. It seems that `node_a` and `node_b` are assumed to be identical.

mrks:
There are some cases that the current ColumnPruningRule does not catch. Most importantly, if a column stops being useful at a certain point in the LQP, it is not pruned at the next opportunity.

This new rule:
* Performs a top-down search for actually required expressions and erases expressions from ProjectionNodes if they are unnecessarily forwarded
* Continues to prune columns from StoredTableNodes
* Identifies inner joins that are actually semi joins (because all columns from one side remain unused). This enables future optimization.

For the last point, we need information about constraints (UNIQUE, PRIMARY). I have partially integrated the constraint definitions from #1570; however, these constraints are for optimization only and are not yet enforced.

We have some good improvements and some minor regressions because the predicate ordering places a semi join with a slightly better selectivity in front of a predicate that does not throw out as many rows but gets executed faster.

```
| TPC-H 01       | 0.60651409626  | 37   | 0.602944254875 | 37   | -1%        |                          0.3481 |
| TPC-H 02       | 8.91521930695  | 535  | 8.77423858643  | 527  | -2%        |                          0.0000 |
| TPC-H 03       | 6.18643331528  | 372  | 5.72342729568  | 344  | -7%        |                          0.0000 |
| TPC-H 04       | 7.65977621078  | 460  | 7.66647338867  | 460  | +0%        |                          0.8283 |
| TPC-H 05       | 4.1903295517   | 252  | 4.04352617264  | 243  | -4%        |                          0.0000 |
| TPC-H 06       | 106.514526367  | 6391 | 105.80241394   | 6349 | -1%        |                          0.0000 |
| TPC-H 07       | 1.59403967857  | 96   | 1.55267488956  | 94   | -3%        |                          0.0000 |
| TPC-H 08       | 5.01469039917  | 301  | 7.2180223465   | 434  | +44%       |                          0.0000 |
| TPC-H 09       | 1.54667651653  | 93   | 1.71847867966  | 104  | +11%       |                          0.0000 |
| TPC-H 10       | 3.0907535553   | 186  | 3.06379699707  | 184  | -1%        |                          0.0140 |
| TPC-H 11       | 25.2146530151  | 1513 | 32.6817512512  | 1961 | +30%       |                          0.0000 |
| TPC-H 12       | 8.5431470871   | 513  | 8.69381046295  | 522  | +2%        |                          0.0000 |
| TPC-H 13       | 2.32613396645  | 140  | 2.28208470345  | 137  | -2%        |                          0.0000 |
| TPC-H 14       | 46.5960960388  | 2796 | 45.8843460083  | 2754 | -2%        |                          0.0000 |
| TPC-H 15       | 73.2578201294  | 4396 | 70.0796432495  | 4205 | -4%        |                          0.0000 |
| TPC-H 16       | 7.19012928009  | 432  | 7.12909936905  | 428  | -1%        |                          0.0000 |
| TPC-H 17       | 1.32355487347  | 80   | 1.41419005394  | 85   | +7%        |                          0.0000 |
| TPC-H 18       | 0.791804969311 | 48   | 0.783314466476 | 48   | -1%        |                          0.0246 |
| TPC-H 19       | 8.55182170868  | 514  | 8.44988918304  | 507  | -1%        |                          0.0000 |
| TPC-H 20       | 2.88092517853  | 173  | 2.63768291473  | 159  | -8%        |                          0.0000 |
| TPC-H 21       | 1.51750218868  | 92   | 1.53639674187  | 93   | +1%        |                          0.0000 |
| TPC-H 22       | 12.8655824661  | 772  | 13.3852567673  | 804  | +4%        |                          0.0000 |
```

Also fixes #1898 by taking the name of MockNodes into account when comparing them.


mrks:
Pinging @mrzzzrm, because he said that he was interested in it

src/benchmarklib/tpch/tpch_table_generator.cpp
```diff
@@ -255,14 +255,37 @@ std::unordered_map<std::string, BenchmarkTableInfo> TPCHTableGenerator::generate
...
-  table_info_by_name["customer"].table = customer_builder.finish_table();
-  table_info_by_name["orders"].table = order_builder.finish_table();
-  table_info_by_name["lineitem"].table = lineitem_builder.finish_table();
-  table_info_by_name["part"].table = part_builder.finish_table();
-  table_info_by_name["partsupp"].table = partsupp_builder.finish_table();
-  table_info_by_name["supplier"].table = supplier_builder.finish_table();
-  table_info_by_name["nation"].table = nation_builder.finish_table();
-  table_info_by_name["region"].table = region_builder.finish_table();
+  auto customer_table = customer_builder.finish_table();
+  customer_table->add_unique_constraint({ColumnID{0}}, true);
```
Bouncner: For readability, I'd prefer `column_id_by_name("c_custkey")` over `ColumnID{0}`.

src/lib/logical_query_plan/join_node.hpp
```diff
@@ -34,7 +34,7 @@ class JoinNode : public EnableMakeForLQPNode<JoinNode>, public AbstractLQPNode {
 
   const std::vector<std::shared_ptr<AbstractExpression>>& join_predicates() const;
 
-  const JoinMode join_mode;
+  JoinMode join_mode;
```
Bouncner: Why?

src/lib/storage/constraints/table_constraint_definition.hpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "types.hpp"
+
+namespace opossum {
+
+// Defines a constraint on a table. Can optionally be a PRIMARY KEY, requiring the column(s) to be non-NULL.
+// Constraints are currently NOT ENFORCED.
+
+struct TableConstraintDefinition final {
+  TableConstraintDefinition(std::vector<ColumnID> column_ids, bool is_primary_key)
```
Bouncner: const?

src/lib/storage/table.cpp
```diff
@@ -267,6 +268,40 @@ void Table::set_table_statistics(const std::shared_ptr<TableStatistics>& table_s
...
 
+const std::vector<TableConstraintDefinition>& Table::get_unique_constraints() const { return _constraint_definitions; }
+
+void Table::add_unique_constraint(const std::vector<ColumnID>& column_ids, bool primary) {
+  for (const auto& column_id : column_ids) {
+    Assert(column_id < column_count(), "ColumnID out of range");
+    Assert(!primary || !column_is_nullable(column_id), "Column must be not nullable for primary key constraint");
+  }
+
+  Assert(column_ids.size() == std::set<ColumnID>(column_ids.begin(), column_ids.end()).size(),
```
Bouncner: Use `std::unique` over `std::set`? https://stackoverflow.com/a/46477901

src/lib/storage/table.cpp
```diff
@@ -267,6 +268,40 @@ void Table::set_table_statistics(const std::shared_ptr<TableStatistics>& table_s
...
+         "Column IDs must be unique");
+
+  {
```
Bouncner: Does that scope really change anything? It's not running multiple times and the scope will not be left any earlier?

src/lib/storage/table.cpp
```diff
@@ -267,6 +268,40 @@ void Table::set_table_statistics(const std::shared_ptr<TableStatistics>& table_s
 
 std::vector<IndexStatistics> Table::indexes_statistics() const { return _indexes; }
 
+const std::vector<TableConstraintDefinition>& Table::get_unique_constraints() const { return _constraint_definitions; }
+
+void Table::add_unique_constraint(const std::vector<ColumnID>& column_ids, bool primary) {
```
Bouncner: const bool? I don't know how we handle const's that probably don't matter.

src/test/operators/insert_test.cpp
```diff
@@ -37,7 +37,6 @@ TEST_F(OperatorsInsertTest, SelfInsert) {
   auto ins = std::make_shared<Insert>(table_name, gt);
```
Bouncner: ASS? Well not really, but still.

src/test/storage/constraints_test.cpp
```diff
@@ -0,0 +1,94 @@
+#include <memory>
+
+#include "base_test.hpp"
+#include "gtest/gtest.h"
```
Bouncner: base test inherits from test.

src/lib/storage/constraints/table_constraint_definition.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
+
+#include <vector>
+
+#include "types.hpp"
+
+namespace opossum {
+
+// Defines a constraint on a table. Can optionally be a PRIMARY KEY, requiring the column(s) to be non-NULL.
+// Constraints are currently NOT ENFORCED.
```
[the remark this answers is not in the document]
mrks: No idea. Rename `Table::get_unique_constraints` to `Table::get_soft_unique_constraints`?

src/benchmarklib/tpch/tpch_table_generator.cpp
mrks: good point

src/lib/logical_query_plan/join_node.hpp
mrks: Because it is updated by the rule

src/lib/storage/table.cpp
mrks: No, but good practice requires an explicit scope for scoped locks. If someone were to add something to the method, chances are that they would inadvertently prolong the lifetime of the scoped_lock.

src/lib/storage/table.cpp
mrks: Absolutely.

src/lib/storage/constraints/table_constraint_definition.hpp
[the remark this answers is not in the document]
Bouncner: Sounds good to me. Soft unique constraints sound "strange enough" to make people wonder what they are and not blindly expect guarantees that are not given.

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,314 @@ using namespace opossum::expression_functional;  // NOLINT
...
+    recursive_gather_required_expressions(input, required_expressions_by_node, outputs_visited_by_node);
+  }
+}
+
+void prune_join_node(
+    const std::shared_ptr<AbstractLQPNode>& node,
+    std::unordered_map<std::shared_ptr<AbstractLQPNode>, ExpressionUnorderedSet>& required_expressions_by_node) {
+  // Sometimes, joins are not actually used to combine data but only to check the existence of a tuple in a second
+  // table. Example: SELECT c_name FROM customer, nation WHERE c_nationkey = n_nationkey AND n_name = 'GERMANY'
+  // These joins will be rewritten to semi joins. However, we can only do this if the join is on a unique/primary
```
Bouncner: The comment is a little bit confusing. What can we only do for the unique joins? Throwing out columns or using semi joins?

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,314 @@ using namespace opossum::expression_functional;  // NOLINT
...
+  DebugAssert(left_input_is_used || right_input_is_used, "Did not expect a useless join");
+  if (left_input_is_used && right_input_is_used) return;
+
+  // Check whether the join predicates operate on unique columns.
+  auto left_joins_on_unique_column = false;
+  auto right_joins_on_unique_column = false;
+
+  const auto& join_predicates = join_node->join_predicates();
+  for (const auto& join_predicate : join_predicates) {
+    const auto is_unique_column = [](const auto& expression) {
```
Bouncner: Why define the lambda in the loop?

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,314 @@ using namespace opossum::expression_functional;  // NOLINT
...
+            continue;
+          }
+
+          const auto column_expression = std::dynamic_pointer_cast<LQPColumnExpression>(expression);
+          pruned_column_ids.emplace_back(column_expression->column_reference.original_column_id());
+        }
+
+        auto stored_table_node = std::dynamic_pointer_cast<StoredTableNode>(node);
+
+        if (pruned_column_ids.size() == stored_table_node->column_expressions().size()) {
```
Bouncner: Can you give me a good example when this can happen? Looks rather random to leave the last row in the StoredTable node.

src/lib/optimizer/strategy/column_pruning_rule.hpp
```diff
@@ -9,26 +9,18 @@ namespace opossum {
...
- * Does NOT eliminate columns added as temporary columns later in the plan or columns that become useless after a
- * certain point in the LQP. *
- * E.g. `SELECT * FROM t WHERE a + 5 > b AND a + 6 > c`: Here `a + 5` and `a + 6` introduce temporary columns that will
- * NOT be removed by this Rule. But it `t` contains a column "d", which is obviously never used in this query, this
- * column "d" will be pruned.
- */
+// Removes expressions (i.e., columns) that are no longer used from the plan
+// - In StoredTableNodes, we can get rid of all columns that are not used anywhere in the LQP
+// - In ProjectionNodes, we can get rid of columns that were used before but are not used anymore. Example:
+//     SELECT SUM(a + 2) FROM (SELECT a, a + 1 FROM t1) t2
```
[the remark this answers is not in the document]
mrks: Should be "Removes expressions (i.e., columns) that are never or no longer used from the plan"

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,314 @@ using namespace opossum::expression_functional;  // NOLINT
...
-  actually_used_columns.insert(output_columns.begin(), output_columns.end());
+namespace {
+void gather_expressions_not_computed_by_expression_evaluator(
+    const std::shared_ptr<AbstractExpression>& expression,
+    const std::vector<std::shared_ptr<AbstractExpression>>& input_expressions,
+    ExpressionUnorderedSet& required_expressions, const bool top_level = true) {
+  if (std::find_if(input_expressions.begin(), input_expressions.end(),
+                   [&expression](const auto& other) { return *expression == *other; }) != input_expressions.end()) {
+    // This expression has already been computed and does not need to be computed again. Only add it if it is not a
+    // top-level (i.e., forwarded) expression
```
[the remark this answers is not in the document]
mrks: Top-level expressions are those that are (part of) the ExpressionEvaluator's final result. For example, for an ExpressionEvaluator producing `(a + b) + c`, the entire expression is a top-level expression. The ExpressionEvaluator does not need it to do its job and it is the consumers job to mark it as required. `(a + b)` however, is required by the ExpressionEvaluator and will be added to `required_expressions`, as it is not a top-level expression.

Does that help? I admit that this is a bit convoluted and that this is my third attempt at explaining it :/

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,314 @@ using namespace opossum::expression_functional;  // NOLINT
...
+    if (!top_level) required_expressions.emplace(expression);
+    return;
+  }
 
-  // Search for ProjectionNodes that forward the unused columns and prune them accordingly
-  _prune_columns_in_projections(lqp, actually_used_columns);
+  if (expression->type == ExpressionType::Aggregate || expression->type == ExpressionType::LQPColumn) {
+    // These cannot be computed by the expression evaluator and are expected to be provided by the input
```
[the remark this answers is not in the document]
mrks: Aggregates are calculated by the AggregateNode and LQPColumns are columns, which are not computed at all. Any idea how to clarify this?

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,314 @@ using namespace opossum::expression_functional;  // NOLINT
...
-          const auto& right_input_expressions = node->right_input()->column_expressions();
-          consumed_columns.insert(right_input_expressions.begin(), right_input_expressions.end());
+          has_at_least_one_expression = true;
         }
-      } break;
-    }
+      }
+
+      if (!has_at_least_one_expression) {
+        // Unsuccessful - add the first expression
```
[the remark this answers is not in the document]
mrks: It should only be the case for `COUNT(*)`, but I will add an assertion. Please don't resolve this yet.

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,314 @@ using namespace opossum::expression_functional;  // NOLINT
...
-          for (const auto& argument : expression->arguments) {
-            collect_consumed_columns_from_expression(argument);
+      // Tracks whether we have already added a column. For COUNT(*) without GROUP BY, we might otherwise end up
+      // without any expressions.
+      auto has_at_least_one_expression = false;
+
+      for (auto expression_idx = size_t{0}; expression_idx < node->node_expressions.size(); ++expression_idx) {
+        const auto& expression = node->node_expressions[expression_idx];
+
+        if (expression_idx < aggregate_node.aggregate_expressions_begin_idx) {
```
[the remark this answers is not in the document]
mrks: The AggregateNode's node_expressions contain both the group_by- and the aggregate_expressions (in that order). Will add a comment.

src/lib/optimizer/strategy/column_pruning_rule.cpp
mrks:
```
  // Sometimes, joins are not actually used to combine data but only to check the existence of a tuple in a second
  // table. Example: SELECT c_name FROM customer, nation WHERE c_nationkey = n_nationkey AND n_name = 'GERMANY'
  // If the join is on a unique/primary key column, we know that each value on the right side of the join will be
  // unique and can rewrite an inner join to a semi join.
```

Better?

src/lib/optimizer/strategy/column_pruning_rule.cpp
mrks: I like to define it as close to its use as possible both for readability and to limit its scope. Since that is entirely a compile time thing, it does not have any additional cost.

src/lib/optimizer/strategy/column_pruning_rule.cpp
mrks: > Can you give me a good example when this can happen?

`SELECT 1 FROM table` does not need any particular column, but it needs at least one column so that it knows how many 1s to produce. Will add this as a comment, please don't resolve.

> Looks rather random to leave the last row in the StoredTable node.

I can't follow. What do you mean by "last row"?

src/lib/optimizer/strategy/column_pruning_rule.hpp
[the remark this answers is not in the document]
Bouncner: Why can't we remove `a + 2` from the very beginning?

src/lib/optimizer/strategy/column_pruning_rule.cpp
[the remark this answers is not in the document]
Bouncner: "The ExpressionEvaluator does not need it to do its job" is unclear. Nonetheless, such a comment would help me a lot.

src/lib/optimizer/strategy/column_pruning_rule.cpp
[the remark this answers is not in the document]
Bouncner: A simple "AggregateNode and LQPColumns are not calculated by the expression eval" would be sufficient, I guess. I was not sure what "cannot be computed by the expression evaluator" is hinting at.

src/lib/optimizer/strategy/column_pruning_rule.cpp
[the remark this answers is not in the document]
Bouncner: Ok, so you collect the group by columns as they cannot be removed. Clear.
But the else part only checks for COUNT(*) and does nothing in this case. What about the actual column used in the expression (that would be column `a` in `SUM(a)`, right?)?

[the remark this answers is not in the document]
Bouncner: Makes sense. Single line comment should suffice and would help a lot.

src/lib/optimizer/strategy/column_pruning_rule.cpp
Bouncner: "combine data" > "combine relations"/tables ?
Maybe begin the second paragraph with "Furthermore, " to make clear, that this is an additional step, that might be added on top.

src/lib/optimizer/strategy/column_pruning_rule.cpp
Bouncner: Last row should be "last column". But I misread your code tbh. What are you actually doing here? In case we could potentially prune all columns (which might result in a failure), we remove just the last column?

And what if `stored_table_node->column_expressions().size()` is 1 and the all columns are prune. Then you would resize to 0, right? Which will fail?

src/lib/optimizer/strategy/column_pruning_rule.hpp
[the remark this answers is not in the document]
mrks: MD: Check if entirely unused subexpressions are pruned, too

src/lib/optimizer/strategy/column_pruning_rule.cpp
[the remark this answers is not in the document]
mrks: s/The ExpressionEvaluator does not need it to do its job and it is the consumers job to mark it as required./It is the consumers job to mark it as required./

src/lib/optimizer/strategy/column_pruning_rule.cpp
mrks: * Add comment from above
* s/Cannot prune all columns/Cannot prune all columns - remove one column from pruning list so that at least one column remains/

mrks:
Please have another look if the documentation is better now.

src/lib/optimizer/strategy/column_pruning_rule.hpp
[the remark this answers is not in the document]
Bouncner: Remove `#include <unordered_map>`.

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
+    const std::shared_ptr<AbstractLQPNode>& node,
+    std::unordered_map<std::shared_ptr<AbstractLQPNode>, ExpressionUnorderedSet>& required_expressions_by_node) {
+  // Sometimes, joins are not actually used to combine tables but only to check the existence of a tuple in a second
+  // table. Example: SELECT c_name FROM customer, nation WHERE c_nationkey = n_nationkey AND n_name = 'GERMANY'
+  // If the join is on a unique/primary key column, we can rewrite these joins into semi joins. If, however, the
+  // uniqueness is not guaranteed, we cannot perform the rewrite as non-unique joins could possibly emit a matching
+  // line more than once.
+
+  auto join_node = std::dynamic_pointer_cast<JoinNode>(node);
+  if (join_node->join_mode != JoinMode::Inner) return;
```
Bouncner: Why do you check for inner? I see that this check is required for the semi optimization, but also for the column pruning?

Bouncner: Ah ok, there is no column pruning taking place, right? The only thing we do in this method is the semi-join optimization. Was expecting something else from the name.

Apart from this optimization: could there by a case we actual column pruning could make sense in a join?

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
+      }
+      return false;
+    };
+
+    const auto& predicate = std::dynamic_pointer_cast<BinaryPredicateExpression>(join_predicate);
+    const auto& left_operand = predicate->left_operand();
+    const auto& right_operand = predicate->right_operand();
+
+    if (!left_joins_on_unique_column) {
+      left_joins_on_unique_column = is_unique_column(left_operand);
```
Bouncner: Can't follow this loop. You iterate over the join predicates and overwrite the `*joins_on_unique_column` bool every time. What if the primary join is an optimizable join, but the secondary predicate is not? That's probably the good case where nothing breaks.
I am wondering if there are cases where secondary predicates might introduce serious problems.

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
-    auto pruned_column_ids = std::vector<ColumnID>{};
-    for (const auto& expression : node->column_expressions()) {
-      if (referenced_columns.find(expression) != referenced_columns.end()) {
-        continue;
+  // Once all nodes that may require columns from this node (i.e., this node's outputs) have been visited, we can
+  // recurse into this node's inputs.
+  for (const auto& input : {node->left_input(), node->right_input()}) {
+    if (!input) continue;
+
+    // Make sure the entry exists, then insert all expressions that the current node needs
```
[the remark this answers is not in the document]
mrks: "Make sure the entry in required_expressions_by_node exists"

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
-      case LQPNodeType::StaticTable:
-      case LQPNodeType::StoredTable:
-      case LQPNodeType::Union:
-      case LQPNodeType::Validate:
-      case LQPNodeType::Mock: {
-        for (const auto& expression : node->node_expressions) {
-          collect_consumed_columns_from_expression(expression);
-        }
-      } break;
+    // For aggregate nodes, we need the group by columns and the arguments to the aggregate functions
```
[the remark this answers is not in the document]
mrks: The aggregate node might have more inputs than that

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
+        required_expressions.emplace(node->left_input()->column_expressions()[0]);
+      }
+    } break;
 
-    return LQPVisitation::VisitInputs;
-  });
+    // For ProjectionNodes, collect all expressions that
+    //   (1) were already computed and are re-used as arguments in this projection
+    //   (2) cannot be computed (i.e., Aggregate and LQPColumn inputs)
+    // As PredicateNodes use the ExpressionEvaluator, they have the same requirements.
```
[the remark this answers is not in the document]
mrks: PredicateNodes have the same requirements - if they have their own implementation, they require all columns to be already computed, if they use the ExpressionEvaluator the columns should at least be computable.

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
+    } break;
+
+    // For Joins, collect the expressions used on the left and right sides of the join expressions
+    case LQPNodeType::Join: {
+      const auto& join_node = static_cast<JoinNode&>(*node);
+      for (const auto& predicate : join_node.join_predicates()) {
+        DebugAssert(predicate->type == ExpressionType::Predicate && predicate->arguments.size() == 2,
+                    "Expected binary predicate for join");
+        required_expressions.emplace(predicate->arguments[0]);
+        required_expressions.emplace(predicate->arguments[1]);
```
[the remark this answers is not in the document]
mrks: The predicates of a join are always binary (asserted above). If we have multiple join predicates, they are another entry in `join_predicates`. We currently do not support OR in join conditions and handle those as cross joins

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
-  visit_lqp(lqp, [&](auto& node) {
-    // Early out for non-leaves
-    if (node->input_count() != 0) {
-      return LQPVisitation::VisitInputs;
-    }
+void recursive_gather_required_expressions(
+    const std::shared_ptr<AbstractLQPNode>& node,
+    std::unordered_map<std::shared_ptr<AbstractLQPNode>, ExpressionUnorderedSet>& required_expressions_by_node,
+    std::unordered_map<std::shared_ptr<AbstractLQPNode>, size_t>& outputs_visited_by_node) {
+  const auto additional_columns = gather_required_expressions(node);
```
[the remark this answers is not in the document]
mrks: locally_required_expressions

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
+  std::unordered_map<std::shared_ptr<AbstractLQPNode>, ExpressionUnorderedSet> required_expressions_by_node;
+
+  // Add top-level columns that need to be included as they are the actual output
+  required_expressions_by_node[lqp].insert(lqp->column_expressions().begin(), lqp->column_expressions().end());
+
+  // Recursively walk through the LQP. We cannot use visit_lqp as we explicitly need to take each path through the LQP.
+  // The right side of a diamond might require additional columns - if we only visited each node once, we might miss
+  // those. However, we track how many of a node's outputs we have already visited and recurse only once we have seen
+  // all of them. That way, the performance should be similar to that of visit_lqp.
+  std::unordered_map<std::shared_ptr<AbstractLQPNode>, size_t> outputs_visited_by_node;
```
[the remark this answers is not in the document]
mrks: resolve offline

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
-      if (referenced_columns.find(expression) != referenced_columns.end()) {
-        continue;
+  // Once all nodes that may require columns from this node (i.e., this node's outputs) have been visited, we can
+  // recurse into this node's inputs.
+  for (const auto& input : {node->left_input(), node->right_input()}) {
+    if (!input) continue;
+
+    // Make sure the entry exists, then insert all expressions that the current node needs
+    required_expressions_by_node[input];
+    for (const auto& required_expression : required_expressions) {
```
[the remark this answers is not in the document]
mrks: Add the columns needed here (and above) if they come from the input node. Reasons why this might NOT be the case are: (1) The expression is calculated in this node (and is thus not available in the input node), or (2) we have two input nodes (i.e., a join) and the expressions comes from the other side.

Better?

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
     }
 
-    // We cannot create a Table without columns - since Chunks rely on their first column to determine their row count
-    if (pruned_column_ids.size() == node->column_expressions().size() && !pruned_column_ids.empty()) {
-      pruned_column_ids.pop_back();
+    recursive_gather_required_expressions(input, required_expressions_by_node, outputs_visited_by_node);
+  }
+}
+
+void prune_join_node(
```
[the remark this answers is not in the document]
mrks: clang-format ¯\_(ツ)_/¯

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,168 +23,324 @@ using namespace opossum::expression_functional;  // NOLINT
...
-    }
+void recursive_gather_required_expressions(
+    const std::shared_ptr<AbstractLQPNode>& node,
+    std::unordered_map<std::shared_ptr<AbstractLQPNode>, ExpressionUnorderedSet>& required_expressions_by_node,
+    std::unordered_map<std::shared_ptr<AbstractLQPNode>, size_t>& outputs_visited_by_node) {
+  const auto additional_columns = gather_required_expressions(node);
+  auto& required_expressions = required_expressions_by_node[node];
+  required_expressions.insert(additional_columns.begin(), additional_columns.end());
+
+  ++outputs_visited_by_node[node];
```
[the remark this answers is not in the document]
mrks: We only continue with node's inputs once we have visited all paths above node. We check this by counting the number of the node's outputs that have already been visited. Once we reach the output count, we can continue.

src/lib/optimizer/strategy/column_pruning_rule.cpp
mrks: > Ah ok, there is no column pruning taking place, right? The only thing we do in this method is the semi-join optimization. Was expecting something else from the name.

Changed to `try_join_to_semi_rewrite`.

> Apart from this optimization: could there by a case we actual column pruning could make sense in a join?

Absolutely, but given that our join operators currently do not support pruning, there is nothing we can do. Adding it to this rule is trivial.

src/lib/optimizer/strategy/column_pruning_rule.cpp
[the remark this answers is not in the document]
Bouncner: Ack.

src/lib/optimizer/strategy/column_pruning_rule.cpp
mrks: > You iterate over the join predicates and overwrite the *joins_on_unique_column bool every time.

No, see the `left_joins_on_unique_column` above. But `|=` is clearer.

> What if the primary join is an optimizable join, but the secondary predicate is not? That's probably the good case where nothing breaks.

Added

mrks:
The issues in the tests should now be fixed. @Bouncner's feedback is included. There are some test fails remaining that come from a bug in the JoinOrderingRule. I'll address that bug in a separate PR that has to be merged first.

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,167 +23,344 @@ using namespace opossum::expression_functional;  // NOLINT
...
+  // Iterate over the ProjectionNode's expressions and add them to the required expressions if at least one output node
+  // requires them
+  auto projection_node = std::dynamic_pointer_cast<ProjectionNode>(node);
+
+  auto new_node_expressions = std::vector<std::shared_ptr<AbstractExpression>>{};
+  new_node_expressions.reserve(projection_node->node_expressions.size());
+
+  for (const auto& expression : projection_node->node_expressions) {
+    for (const auto& output : node->outputs()) {
+      const auto& required_expressions = required_expressions_by_node.at(output);
```
Bensk1: Are you using `at()` intentionally here to get an exception in case `output` is not in `required_expressions_by_node `?

src/lib/storage/table.cpp
```diff
@@ -282,6 +283,39 @@ void Table::set_table_statistics(const std::shared_ptr<TableStatistics>& table_s
 
 std::vector<IndexStatistics> Table::indexes_statistics() const { return _indexes; }
 
+const std::vector<TableConstraintDefinition>& Table::get_soft_unique_constraints() const {
+  return _constraint_definitions;
+}
+
+void Table::add_soft_unique_constraint(const std::vector<ColumnID>& column_ids, const bool primary) {
```
Bensk1: `primary_key` or `is_primary_key`

src/test/expression/expression_utils_test.cpp
```diff
@@ -37,6 +41,68 @@ TEST_F(ExpressionUtilsTest, ExpressionFlattenAndInflate) {
...
+
+  EXPECT_FALSE(expression_evaluable_on_lqp(lqp_column_(a_a), *input_lqp));
+  EXPECT_TRUE(expression_evaluable_on_lqp(lqp_column_(a_b), *input_lqp));
+  EXPECT_TRUE(expression_evaluable_on_lqp(sum_(lqp_column_(a_c)), *input_lqp));
+  EXPECT_FALSE(expression_evaluable_on_lqp(sum_(lqp_column_(a_a)), *input_lqp));
+  EXPECT_FALSE(expression_evaluable_on_lqp(lqp_column_(b_a), *input_lqp));
+}
+
+TEST_F(ExpressionUtilsTest, ExpressionEvaluableOnJoin) {
+  // SELECT b, SUM(c) FROM a WHERE a = 1
```
Bensk1: comment copy pasta

src/test/logical_query_plan/mock_node_test.cpp
```diff
@@ -61,7 +61,7 @@ TEST_F(MockNodeTest, HashingAndEqualityCheck) {
       MockNode::make(MockNode::ColumnDefinitions{{DataType::Int, "a"}, {DataType::Float, "b"}}, "other_name");
   EXPECT_EQ(*_mock_node_b, *_mock_node_b);
   EXPECT_NE(*_mock_node_b, *different_mock_node_1);
-  EXPECT_EQ(*_mock_node_b, *different_mock_node_2);
+  EXPECT_NE(*_mock_node_b, *different_mock_node_2);
```
Bensk1: That is because the name wasn't checked in the old version?

src/test/storage/constraints_test.cpp
```diff
@@ -0,0 +1,93 @@
...
+  EXPECT_THROW(table_nullable->add_soft_unique_constraint({ColumnID{0}, ColumnID{1}}, true), std::exception);
+
+  table->add_soft_unique_constraint({ColumnID{2}}, true);
+
+  // Invalid because another primary key already exists.
+  EXPECT_THROW(
+      {
+        try {
+          table->add_soft_unique_constraint({ColumnID{2}}, true);
+        } catch (const std::exception& e) {
```
Bensk1: wow

src/test/storage/constraints_test.cpp
```diff
@@ -0,0 +1,93 @@
...
+    {
+      TableColumnDefinitions column_definitions;
+      column_definitions.emplace_back("column0", DataType::Int, false);
+      column_definitions.emplace_back("column1", DataType::Int, true);
+      auto table = std::make_shared<Table>(column_definitions, TableType::Data, 2, UseMvcc::Yes);
+
+      auto& sm = Hyrise::get().storage_manager;
+      sm.add_table("table_nullable", table);
+
+      table->add_soft_unique_constraint({ColumnID{0}}, false);
```
Bensk1: Is this needed in the tests?

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -85,9 +85,23 @@ ExpressionUnorderedSet gather_locally_required_expressions(
     case LQPNodeType::Aggregate: {
       const auto& aggregate_node = static_cast<AggregateNode&>(*node);
 
-      // Tracks whether we have already added a column. For COUNT(*) without GROUP BY, we might otherwise end up
-      // without any expressions.
-      auto has_at_least_one_expression = false;
+      // Handling COUNT(*) (which is represented as an LQPColumnReference with a valid original_node and an
+      // INVALID_COLUMN_ID) is difficult, as we need to make sure that at least one expression from that
```
Bouncner: Why does at least one expression need to survive?

Bouncner: Further the note sounds like "we stop once we see star" while the code is "if we see count star, take all expressions".

src/lib/optimizer/strategy/column_pruning_rule.hpp
```diff
@@ -9,26 +9,18 @@ namespace opossum {
...
- * NOT be removed by this Rule. But it `t` contains a column "d", which is obviously never used in this query, this
- * column "d" will be pruned.
- */
+// Removes expressions (i.e., columns) that are never or no longer used from the plan
+// - In StoredTableNodes, we can get rid of all columns that are not used anywhere in the LQP
+// - In ProjectionNodes, we can get rid of columns that are not part of the result or not used anymore. Example:
+//     SELECT SUM(a + 2) FROM (SELECT a, a + 1 FROM t1) t2
+//   Here, `a + 1` is never actually used and should be pruned
+// - Joins that emit columns that are never used can be rewritten to semi joins if (a) the unused side has a unique
+//     constraint and (b) the join is an inner join. This is done in the ColumnPruningRule because it requires
```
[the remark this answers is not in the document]
mrks: That would require us to pass this information to a separate rule, either rule-to-rule or by storing it in the LQP. How would we make sure that does not become outdated?

src/lib/optimizer/strategy/column_pruning_rule.hpp
```diff
@@ -9,26 +9,18 @@ namespace opossum {
...
- * E.g. `SELECT * FROM t WHERE a + 5 > b AND a + 6 > c`: Here `a + 5` and `a + 6` introduce temporary columns that will
- * NOT be removed by this Rule. But it `t` contains a column "d", which is obviously never used in this query, this
- * column "d" will be pruned.
- */
+// Removes expressions (i.e., columns) that are never or no longer used from the plan
+// - In StoredTableNodes, we can get rid of all columns that are not used anywhere in the LQP
+// - In ProjectionNodes, we can get rid of columns that are not part of the result or not used anymore. Example:
+//     SELECT SUM(a + 2) FROM (SELECT a, a + 1 FROM t1) t2
+//   Here, `a + 1` is never actually used and should be pruned
+// - Joins that emit columns that are never used can be rewritten to semi joins if (a) the unused side has a unique
```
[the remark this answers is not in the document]
mrks: No

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,167 +23,344 @@ using namespace opossum::expression_functional;  // NOLINT
...
-    return LQPVisitation::VisitInputs;
-  });
+      const auto& column_reference = column->column_reference;
+      const auto& stored_table_node =
+          std::dynamic_pointer_cast<const StoredTableNode>(column_reference.original_node());
+      if (!stored_table_node) return false;
+
+      const auto& table = Hyrise::get().storage_manager.get_table(stored_table_node->table_name);
+      for (const auto& table_constraint : table->get_soft_unique_constraints()) {
+        // This currently does not handle multi-column constraints, but that should be easy to add once needed.
```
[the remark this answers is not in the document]
mrks: The emphasis is on the second part. Otherwise people might think that there is some god-given limitation that prevents multi-column constraints from being useful here.

src/lib/optimizer/strategy/column_pruning_rule.cpp
```diff
@@ -22,167 +23,344 @@ using namespace opossum::expression_functional;  // NOLINT
...
+    if (predicate->predicate_condition != PredicateCondition::Equals) continue;
+
+    const auto& left_operand = predicate->left_operand();
+    const auto& right_operand = predicate->right_operand();
+
+    join_is_unique_on_left_side |= is_unique_column(left_operand);
+    join_is_unique_on_right_side |= is_unique_column(right_operand);
+  }
+
+  // If one of the input sides is unused (i.e., its expressions are not needed in the output), check whether
```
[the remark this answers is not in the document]
mrks: d'oh

src/lib/optimizer/strategy/column_pruning_rule.cpp
mrks: Yes.

src/test/logical_query_plan/mock_node_test.cpp
mrks: Correct.

src/test/storage/constraints_test.cpp
mrks: Yeah, that's too much.

src/test/storage/constraints_test.cpp
mrks: Yes, for example in the exception case that you quoted above.

src/lib/optimizer/strategy/column_pruning_rule.cpp
mrks: > Handling COUNT(*) (which is represented as an LQPColumnReference with a valid original_node and an INVALID_COLUMN_ID) is difficult, as we need to make sure that at least one expression from that original node survives the pruning. Otherwise, we could not resolve that original_node later on. For now, we simply stop pruning (i.e., add all expressions as required) once we encounter COUNT(*).

better?

src/lib/storage/constraints/table_constraint_definition.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+struct TableConstraintDefinition final {
+  TableConstraintDefinition(std::vector<ColumnID> column_ids, const IsPrimaryKey is_primary_key)
+      : columns(std::move(column_ids)), is_primary_key(is_primary_key == IsPrimaryKey::Yes) {
+    DebugAssert(std::is_sorted(columns.begin(), columns.end()), "Expecting Column IDs to be sorted");
+    Assert(std::unique(columns.begin(), columns.end()) == columns.end(), "Expected Column IDs to be unique");
+  }
+
+  std::vector<ColumnID> columns;
+  bool is_primary_key;
```
Bouncner: Use `IsPrimaryKey`?

mrks: Thought about it, but when we access `x.is_primary_key`, we already know what the bool stands for. This is not one of the cases where we will see a bool without its meaning.

Bouncner: But adding a new constraint based on a constraint is now `add_constraint(con.column_id, con.is_primary_key ? IsPrimaryKey::Yes : IsPrimaryKey::No)`.

mrks:
> I think currently constraints are not set when the tables are loaded as binaries as the TPCH table generator is not used and the stored binary files do not export/import any constraints.

Good point. I'll fix that once I have @Bensk1's comments on the tests

mrks:
Ping @Bensk1

Bensk1:
> Ping

I have looked at it and didn't find anything to be missing.

mrks:
@Bouncner - I fixed `cache_binary_tables` and added asserts to ensure that we never again lose that information

src/benchmarklib/tpch/tpch_benchmark_item_runner.cpp
```diff
@@ -70,27 +71,36 @@ bool TPCHBenchmarkItemRunner::_on_execute_item(const BenchmarkItemID item_id, Be
...
+  const auto orders_table = Hyrise::get().storage_manager.get_table("orders");
+  const auto first_chunk = orders_table->get_chunk(ChunkID{0});
+  Assert(first_chunk->ordered_by(), "Ordering information was lost");
+  if (_config->indexes) {
+    Assert(!first_chunk->get_indexes({ColumnID{0}}).empty(), "Index was lost");
+  }
+  Assert(!orders_table->get_soft_unique_constraints().empty(), "Constraints were lost");
 
-  std::cout << " - Preparing queries" << std::endl;
+  if (_use_prepared_statements) {
```
mrks: From here until the end, nothing changed. I converted the guard clause into a regular if statement as it is not really a guard clause in the strictest sense.

PR #1856, opened 2019-09-19
Faster synthetic table creation

Bouncner:
This PR improves the runtime of the synthetic table generator. Mainly by using std::threads (similar to the benchmarking encoder). 

Time to create a large table for the model calibration went from `166s` to `56s` on a four core laptop.

It further does the following:
 * add tests for the synthetic table generator
 * add test to the chunk encoder
 * fixes LZ4 to SimdBp128 vector compression (which was the idea from the beginning since LZ4 is all about compression ratios) and adapts the encodings listings accordingly
 * fix a bug concerning MVCC data

### DIFF
--- a/src/benchmarklib/synthetic_table_generator.cpp
+++ b/src/benchmarklib/synthetic_table_generator.cpp
@@ -5,6 +5,7 @@
 #include <memory>
 #include <random>
 #include <string>
+#include <thread>
 #include <utility>
 #include <vector>
 
@@ -28,11 +29,11 @@ using namespace opossum;  // NOLINT
 
 template <typename T>
 pmr_concurrent_vector<T> create_typed_segment_values(const std::vector<int>& values) {
-  pmr_concurrent_vector<T> result;
-  result.reserve(values.size());
+  pmr_concurrent_vector<T> result(values.size());
 
+  auto insert_position = size_t{0};
   for (const auto& value : values) {
-    result.push_back(SyntheticTableGenerator::generate_value<T>(value));
+    result[insert_position++] = SyntheticTableGenerator::generate_value<T>(value);
   }
 
   return result;
@@ -57,7 +58,7 @@ std::shared_ptr<Table> SyntheticTableGenerator::generate_table(const size_t num_
 std::shared_ptr<Table> SyntheticTableGenerator::generate_table(
     const std::vector<ColumnDataDistribution>& column_data_distributions,
     const std::vector<DataType>& column_data_types, const size_t num_rows, const ChunkOffset chunk_size,
-    const std::optional<std::vector<SegmentEncodingSpec>>& segment_encoding_specs,
+    const std::optional<ChunkEncodingSpec>& segment_encoding_specs,
     const std::optional<std::vector<std::string>>& column_names, const UseMvcc use_mvcc) {
   Assert(chunk_size != 0ul, "cannot generate table with chunk size 0");
   Assert(column_data_distributions.size() == column_data_types.size(),
@@ -83,14 +84,13 @@ std::shared_ptr<Table> SyntheticTableGenerator::generate_table(
   std::shared_ptr<Table> table = std::make_shared<Table>(column_definitions, TableType::Data, chunk_size, use_mvcc);
 
   std::random_device random_device;
-  // Using std::mt19937 over std::default_random_engine since it provides better guarantees on the randomness
-  // of values being created (see discussion: https://stackoverflow.com/q/30240899/1147726)
   auto pseudorandom_engine = std::mt19937{};
 
   pseudorandom_engine.seed(random_device());
+  auto encoder_threads = std::vector<std::thread>{};
 
   for (auto chunk_index = ChunkOffset{0}; chunk_index < num_chunks; ++chunk_index) {
-    Segments segments;
+    Segments segments(num_columns);
     for (auto column_index = ColumnID{0}; column_index < num_columns; ++column_index) {
       resolve_data_type(column_data_types[column_index], [&](const auto column_data_type) {
         using ColumnDataType = typename decltype(column_data_type)::type;
@@ -149,20 +149,40 @@ std::shared_ptr<Table> SyntheticTableGenerator::generate_table(
           values.push_back(generate_value_by_distribution_type());
         }
 
-        segments.push_back(
-            std::make_shared<ValueSegment<ColumnDataType>>(create_typed_segment_values<ColumnDataType>(values)));
+        segments[column_index] = 
+            std::make_shared<ValueSegment<ColumnDataType>>(create_typed_segment_values<ColumnDataType>(values));
       });
+
       // add full chunk to table
       if (column_index == num_columns - 1) {
-        const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
-        table->append_chunk(segments, mvcc_data);
+        if (use_mvcc == UseMvcc::Yes) {
+          const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+          table->append_chunk(segments, mvcc_data);
+        } else {
+          table->append_chunk(segments);
+        }
+
+        if (segment_encoding_specs) {
+          /**
+          * For chunk sizes <1M, parallelizing the segment creation is not worth the overhead, because chunk encoding 
+          * dominates the runtime. Nonetheless, a single thread for the chunk encoder (which itself is parallelizing
+          * internally) ensures that the segment creation is not blocked by the encoding of the previous chunk. To
+          * avoid overparallelizing, calls to encode_chunk are not parallelized.
+          **/
+          if (!encoder_threads.empty()) {
+            encoder_threads.back().join();
+          }
+
+          const auto chunk_id = ChunkID{table->chunk_count() - 1};
+          encoder_threads.emplace_back([&, chunk_id] {
+            ChunkEncoder::encode_chunk(table->get_chunk(chunk_id), table->column_data_types(), *segment_encoding_specs, true);
+          });
+        }
       }
     }
   }
 
-  if (segment_encoding_specs) {
-    ChunkEncoder::encode_all_chunks(table, *segment_encoding_specs);
-  }
+  encoder_threads.back().join();
 
   return table;
 }
--- a/src/benchmarklib/synthetic_table_generator.cpp
+++ b/src/benchmarklib/synthetic_table_generator.cpp
@@ -33,7 +33,8 @@ pmr_concurrent_vector<T> create_typed_segment_values(const std::vector<int>& val
 
   auto insert_position = size_t{0};
   for (const auto& value : values) {
-    result[insert_position++] = SyntheticTableGenerator::generate_value<T>(value);
+    result[insert_position] = SyntheticTableGenerator::generate_value<T>(value);
+    ++insert_position;
   }
 
   return result;
@@ -164,8 +165,8 @@ std::shared_ptr<Table> SyntheticTableGenerator::generate_table(
 
         if (segment_encoding_specs) {
           /**
-          * For chunk sizes <1M, parallelizing the segment creation is not worth the overhead, because chunk encoding 
-          * dominates the runtime. Nonetheless, a single thread for the chunk encoder (which itself is parallelizing
+          * As the runtime of the table generation is dominated by the chunk encoding, parallelizing the actual data
+          * generation barely impact the runtime. Nonetheless, a single thread for the chunk encoder (which itself is parallelizing
           * internally) ensures that the segment creation is not blocked by the encoding of the previous chunk. To
           * avoid overparallelizing, there is at most one thread for encode_chunk() at any time.
           **/
--- a/src/benchmarklib/synthetic_table_generator.cpp
+++ b/src/benchmarklib/synthetic_table_generator.cpp
@@ -146,7 +146,13 @@ std::shared_ptr<Table> SyntheticTableGenerator::generate_table(
           if (chunk_index * chunk_size + (row_offset + 1) > num_rows - 2) {
             break;
           }
-          values.push_back(generate_value_by_distribution_type());
+
+          if constexpr (std::is_same_v<ColumnDataType, pmr_string> && column_data_distribution.distribution_type == DataDistributionType::NormalSkewed) {
+            // shift values and remove negative ones, because negative values cannot converted to strings as of now
+            values.push_back(abs(generate_value_by_distribution_type() + 10'000));
+          } else {
+            values.push_back(generate_value_by_distribution_type());
+          }
         }
 
         segments[column_index] =
--- a/src/benchmarklib/synthetic_table_generator.cpp
+++ b/src/benchmarklib/synthetic_table_generator.cpp
@@ -164,37 +164,34 @@ std::shared_ptr<Table> SyntheticTableGenerator::generate_table(
             values.push_back(generate_value_by_distribution_type());
           }
 
-          segments[column_index] =
+          auto value_segment =
               std::make_shared<ValueSegment<ColumnDataType>>(create_typed_segment_values<ColumnDataType>(values));
-        });
 
-        // add full chunk to table
-        if (column_index == num_columns - 1) {
-          if (use_mvcc == UseMvcc::Yes) {
-            const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
-            table->append_chunk(segments, mvcc_data);
+          if (!segment_encoding_specs) {
+            segments[column_index] = value_segment;
           } else {
-            table->append_chunk(segments);
+            segments[column_index] = ChunkEncoder::encode_segment(value_segment, column_data_types[column_index],
+                                         segment_encoding_specs->at(column_index));
           }
-
-          // get added chunk, mark it as immutable and add statistics
-          const auto& added_chunk = table->get_chunk(ChunkID{table->chunk_count() - 1});
-          added_chunk->mark_immutable();
-          if (!added_chunk->pruning_statistics()) {
-            generate_chunk_pruning_statistics(added_chunk);
-          }
-
-          if (segment_encoding_specs) {
-            for (auto column_id_to_encode = ColumnID{0}; column_id_to_encode < num_columns; ++column_id_to_encode)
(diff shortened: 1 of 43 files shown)


src/test/synthetic_table_generator_test.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+namespace opossum {
+
+TEST(SyntheticTableGeneratorTest, StringGeneration) {
+  EXPECT_EQ(SyntheticTableGenerator::generate_value<pmr_string>(0), "          ");
+  EXPECT_EQ(SyntheticTableGenerator::generate_value<pmr_string>(1), "         1");
+  EXPECT_EQ(SyntheticTableGenerator::generate_value<pmr_string>(2), "         2");
+  EXPECT_EQ(SyntheticTableGenerator::generate_value<pmr_string>(117), "        1t");
+  EXPECT_EQ(SyntheticTableGenerator::generate_value<pmr_string>(50'018), "       D0k");
+  EXPECT_EQ(SyntheticTableGenerator::generate_value<pmr_string>(3'433'820), "      EPIC");
```
mrks: Not 17? :(

src/benchmarklib/synthetic_table_generator.cpp
```diff
@@ -28,11 +29,11 @@ using namespace opossum;  // NOLINT
...
 template <typename T>
 pmr_concurrent_vector<T> create_typed_segment_values(const std::vector<int>& values) {
-  pmr_concurrent_vector<T> result;
-  result.reserve(values.size());
+  pmr_concurrent_vector<T> result(values.size());
 
+  auto insert_position = size_t{0};
   for (const auto& value : values) {
-    result.push_back(SyntheticTableGenerator::generate_value<T>(value));
+    result[insert_position++] = SyntheticTableGenerator::generate_value<T>(value);
```
mrks: Do one thing per line.

src/benchmarklib/synthetic_table_generator.cpp
```diff
@@ -149,20 +149,41 @@ std::shared_ptr<Table> SyntheticTableGenerator::generate_table(
...
+        if (use_mvcc == UseMvcc::Yes) {
+          const auto mvcc_data = std::make_shared<MvccData>(segments.front()->size(), CommitID{0});
+          table->append_chunk(segments, mvcc_data);
+        } else {
+          table->append_chunk(segments);
+        }
+
+        if (segment_encoding_specs) {
+          /**
+          * For chunk sizes <1M, parallelizing the segment creation is not worth the overhead, because chunk encoding
```
mrks: Where do I find the 1M?

src/benchmarklib/synthetic_table_generator.cpp
```diff
@@ -149,20 +149,41 @@ std::shared_ptr<Table> SyntheticTableGenerator::generate_table(
...
+
+        if (segment_encoding_specs) {
+          /**
+          * For chunk sizes <1M, parallelizing the segment creation is not worth the overhead, because chunk encoding 
+          * dominates the runtime. Nonetheless, a single thread for the chunk encoder (which itself is parallelizing
+          * internally) ensures that the segment creation is not blocked by the encoding of the previous chunk. To
+          * avoid overparallelizing, there is at most one thread for encode_chunk() at any time.
+          **/
+          if (!encoder_threads.empty()) {
+            encoder_threads.back().join();
```
mrks: I don't really get this part. In which case will there be more than one encoder_thread?

src/benchmarklib/synthetic_table_generator.hpp
```diff
@@ -97,18 +97,21 @@ class SyntheticTableGenerator {
...
 
-      Assert(input >= 0, "Integer values need to be positive in order to be converted to a pmr_string.");
-
       const std::vector<char> chars = {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F',
                                        'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V',
                                        'W', 'X', 'Y', 'Z', 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l',
                                        'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z'};
       const size_t chars_base = chars.size();
+
+      // For the current settings, integer valures are guaranteed to not exceed the value range. However, the assertion
```
mrks: typo

src/benchmarklib/synthetic_table_generator.hpp
```diff
@@ -97,18 +97,21 @@ class SyntheticTableGenerator {
...
+      // is kept as the generated string length might change or the interface (e.g., taking size_t input).
```
mrks: 1) `int` does not guarantee you anything. `int32_t` would.
2) What does this comment explain (read: what information would be missing without it)?

src/benchmarklib/synthetic_table_generator.hpp
```diff
@@ -125,6 +128,8 @@ class SyntheticTableGenerator {
       }
 
       return result;
+    } else {
+      Fail("Requested type not supported.");
```
mrks: What is this needed for? If you didn't have the else path and would try to initialize it with a type not covered above, the compiler would complain about a missing return value.

src/lib/storage/chunk_encoder.cpp
```diff
@@ -92,22 +93,41 @@ std::shared_ptr<BaseSegment> ChunkEncoder::encode_segment(const std::shared_ptr<
...
+  const auto column_count = chunk->column_count();
+  Assert((column_data_types.size() == column_count), "Number of column types must match the chunk’s column count.");
+  Assert((chunk_encoding_spec.size() == column_count),
          "Number of column encoding specs must match the chunk’s column count.");
 
-  for (ColumnID column_id{0}; column_id < chunk->column_count(); ++column_id) {
-    const auto spec = chunk_encoding_spec[column_id];
+  auto next_column_id = std::atomic_ushort{0};
+  auto threads = std::vector<std::thread>{};
+  const auto thread_count = std::min(column_count, static_cast<uint16_t>(std::thread::hardware_concurrency()));
```
mrks: I'm already unhappy about non-scheduler multithreading being in benchmarklib/benchmark_table_encoder.cpp, but I don't think that it should make its way into src/lib.

src/lib/storage/lz4_segment/lz4_encoder.hpp
```diff
@@ -224,7 +224,7 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
     }
 
     // Compress the offsets with a vector compression method to reduce the memory footprint of the LZ4 segment.
-    auto compressed_offsets = compress_vector(offsets, vector_compression_type(), allocator, {offsets.back()});
+    auto compressed_offsets = compress_vector(offsets, VectorCompressionType::SimdBp128, allocator, {offsets.back()});
```
mrks: Does this disable FSBA for LZ4?

src/lib/storage/lz4_segment.cpp
```diff
@@ -454,6 +455,16 @@ std::optional<CompressedVectorType> LZ4Segment<T>::compressed_vector_type() cons
   return std::nullopt;
 }
 
+template <>
+std::optional<CompressedVectorType> LZ4Segment<pmr_string>::compressed_vector_type() const {
```
mrks: Isn't this always `CompressedVectorType::SimdBp128` with your change below?

src/lib/storage/lz4_segment.cpp
```diff
@@ -454,6 +455,16 @@ std::optional<CompressedVectorType> LZ4Segment<T>::compressed_vector_type() cons
...
+  if (_string_offsets) {
+    CompressedVectorType type = CompressedVectorType::SimdBp128;  // initialize for GCC
```
mrks: You could also use an optional

src/test/base_test.hpp
```diff
@@ -163,6 +163,29 @@ class BaseTestWithParam
     return chunk_encoding_spec;
   }
 
+  static void verify_encoding(const std::shared_ptr<Chunk>& chunk, const ChunkEncodingSpec& spec) {
```
mrks: Maybe `bool chunk_matches_encoding`?

src/test/storage/chunk_encoder_test.cpp
```diff
@@ -19,72 +21,110 @@ namespace opossum {
...
+    return table;
   }
 
  protected:
   std::shared_ptr<Table> _table;
 };
 
+TEST_F(ChunkEncoderTest, ParentVectorCompressionType) {
+  ASSERT_EQ(parent_vector_compression_type(CompressedVectorType::FixedSize4ByteAligned),
+            VectorCompressionType::FixedSizeByteAligned);
```
mrks: This is not really a test

src/benchmarklib/synthetic_table_generator.hpp
Bouncner: It's supposed to say that the following assert should be kept, even though it can be never be wrong at the moment. We could remove it, but I am afraid the next user who changes the input to something larger will stumble upon this issue. So I'd keep the assertion.

src/benchmarklib/synthetic_table_generator.hpp
Bouncner: It would try to convert to a pmr_string. I thought being more explicit about the types here cannot be wrong.

src/lib/storage/chunk_encoder.cpp
Bouncner: Hmm ... good point. Any ideas how to get around this "performance issue" in case there is no scheduler?

src/lib/storage/lz4_segment/lz4_encoder.hpp
Bouncner: Yes.

src/lib/storage/lz4_segment.cpp
Bouncner: It is, but whenever somebody feels like changing to SimdBp128 or making the decision configurable, the method should still be good.

src/lib/storage/lz4_segment.cpp
Bouncner: True. Will do.

src/benchmarklib/synthetic_table_generator.hpp
mrks: Maybe just "Implicitly guaranteed by int32_t parameter, but asserting this just in case the parameter changes in the future"?

src/lib/storage/chunk_encoder.cpp
mrks: Set/unset the scheduler

src/lib/storage/lz4_segment/lz4_encoder.hpp
mrks: add a comment?

src/lib/storage/lz4_segment.cpp
mrks: add a comment

src/benchmarklib/synthetic_table_generator.cpp
Bouncner: There will never be, we just wait for the last thread (`.back()`) to finish.
Since the data generation is much faster than encoding, we would create many more encoders if we don't wait for the previous one.
I failed using a single std::thread variable and reusing it, thus the vector.

src/lib/storage/chunk_encoder.cpp
Bouncner: But then we have a component in the system that potentially changes the scheduler during runtime of Hyrise. Which sounds super bad to be.

src/lib/storage/lz4_segment.cpp
Bouncner: Added.

src/test/base_test.hpp
Bouncner: The current function does several checks internally which would then directly lead to the faulted assertion. Returning a single bool would require us to have additional means (like table printing) to show the actual problem. 
Any reason not to go the direct path here?

src/lib/storage/lz4_segment/lz4_encoder.hpp
Bouncner: Done.

src/test/synthetic_table_generator_test.cpp
Bouncner: How could I ...

src/benchmarklib/synthetic_table_generator.cpp
Bouncner: As I totally see your argument to avoid `std::thread` in `src/lib`, I think I'll call`encode_segment()` via some `std::threads` (duplicating 5 lines of marking the chunk as immutable).
This way, we don't have to set/reset the scheduler and we don't introduce any manual threads to core Hyrise. What do you think?

src/lib/storage/chunk_encoder.cpp
mrks: If you do it in the synthetic table generator, it is not really a component inside Hyrise.

src/test/base_test.hpp
mrks: We can do that as well, but then I would make the method's name more speaking. Maybe `assert_chunk_encoding`?

Bouncner: Sounds good. Will change.

src/lib/storage/chunk_encoder.cpp
Bouncner: Yeah, but when looking at the calibration, we’ll use the synth generator and can also set a scheduler or not. Having a component fiddling with the scheduler before the actual calibration sounds like something that can easily go wrong (or am I too pessimistic?).

mrks: If you make sure that you reset the scheduler once you are done, I don't see anything going wrong

src/test/base_test.hpp
Bouncner: Done.

src/lib/storage/lz4_segment/lz4_encoder.hpp
```diff
@@ -223,8 +223,10 @@ class LZ4Encoder : public SegmentEncoder<LZ4Encoder> {
                                                           nullptr, _block_size, 0u, 0u, null_values.size());
     }
 
-    // Compress the offsets with a vector compression method to reduce the memory footprint of the LZ4 segment.
-    auto compressed_offsets = compress_vector(offsets, vector_compression_type(), allocator, {offsets.back()});
+    // Compress the offsets with SimdBp128 vector compression to reduce the memory footprint of the LZ4 segment.
+    // SimdBp128 is chosen over FSBA, since it compresses better and the performance advantage of FSBA is neglectable,
```
mrks: Write out FSBA

src/benchmarklib/synthetic_table_generator.hpp
```diff
@@ -32,7 +32,7 @@ struct ColumnDataDistribution {
 
   static ColumnDataDistribution make_skewed_normal_config(const double skew_location = 0.0,
                                                           const double skew_scale = 1.0,
-                                                          const double skew_shape = 0.0) {
+                                                          const double skew_shape = 0.0001) {
```
mrks: still necessary?

src/lib/storage/chunk_encoder.cpp
```diff
@@ -92,22 +93,41 @@ std::shared_ptr<BaseSegment> ChunkEncoder::encode_segment(const std::shared_ptr<
...
+                                const ChunkEncodingSpec& chunk_encoding_spec, const bool encode_in_parallel) {
+  const auto column_count = chunk->column_count();
+  Assert((column_data_types.size() == column_count), "Number of column types must match the chunk’s column count.");
+  Assert((chunk_encoding_spec.size() == column_count),
          "Number of column encoding specs must match the chunk’s column count.");
 
-  for (ColumnID column_id{0}; column_id < chunk->column_count(); ++column_id) {
-    const auto spec = chunk_encoding_spec[column_id];
+  auto next_column_id = std::atomic_ushort{0};
+  auto threads = std::vector<std::thread>{};
```
mrks: You are still using threads here

src/lib/storage/chunk_encoder.cpp
```diff
@@ -92,22 +93,41 @@ std::shared_ptr<BaseSegment> ChunkEncoder::encode_segment(const std::shared_ptr<
 }
 
 void ChunkEncoder::encode_chunk(const std::shared_ptr<Chunk>& chunk, const std::vector<DataType>& column_data_types,
-                                const ChunkEncodingSpec& chunk_encoding_spec) {
-  Assert((column_data_types.size() == chunk->column_count()),
-         "Number of column types must match the chunk’s column count.");
-  Assert((chunk_encoding_spec.size() == chunk->column_count()),
+                                const ChunkEncodingSpec& chunk_encoding_spec, const bool encode_in_parallel) {
```
mrks: Is encode_in_parallel still needed?

src/lib/storage/chunk_encoder.hpp
```diff
@@ -35,11 +35,14 @@ class ChunkEncoder {
   /**
    * @brief Encodes a chunk
    *
-   * Encodes a chunk using the passed encoding specifications.
-   * Reduces also the fragmentation of the chunk’s MVCC data.
+   * Encodes a chunk using the passed encoding specifications. Reduces also the fragmentation of the chunk’s MVCC data.
+   * 
+   * thread_per_segment allows to parallelize the encoding. It defaults to false since encoding is expected to be a
+   * background process with a minimal overhead. The scheduler is not used in order to parallelize even when the
+   * scheduler is disabled.
```
mrks: update

src/benchmarklib/synthetic_table_generator.hpp
Bouncner: Not with the latest fixes to boost, but as long as we do not pull the latest master of boost, we have use this fix.

Bouncner: I had a few things not pushed yet, sorry for that. There is a comment with a link to the issue.

Bouncner:
I hope everything is addressed so far. Sorry for the confusion with half-committed code.

PR #1858, opened 2019-09-21
Improve doc wording in Optimizer and StorageManager

mrzzzrm:
Previously it sounded like the JOR was introducing semi joins

### DIFF
--- a/src/lib/optimizer/optimizer.cpp
+++ b/src/lib/optimizer/optimizer.cpp
@@ -96,7 +96,8 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {
 
   optimizer->add_rule(std::make_unique<ColumnPruningRule>());
 
-  // Run before SubqueryToJoinRule, since the Semi/Anti Joins it introduces are opaque to the JoinOrderingRule
+  // Run the JoinOrderingRule before the SubqueryToJoinRule, since the Semi/Anti Joins the SubqueryToJoinRule introduces
+  // are opaque to the JoinOrderingRule
   optimizer->add_rule(std::make_unique<JoinOrderingRule>());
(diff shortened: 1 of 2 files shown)


src/lib/optimizer/optimizer.cpp
```diff
@@ -96,7 +96,8 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {
 
   optimizer->add_rule(std::make_unique<ColumnPruningRule>());
 
-  // Run before SubqueryToJoinRule, since the Semi/Anti Joins it introduces are opaque to the JoinOrderingRule
+  // Run the JoinOrderingRule before the SubqueryToJoinRule, since the Semi/Anti Joins the SubqueryToJoinRule introduces
+  // are opaque to the JoinOrderingRule
```
mrks: Not sure if everyone has the same understanding of "opaque". Maybe "Run the JoinOrderingRule before the SubqueryToJoinRule as the join ordering cannot proceed past Semi/Anti Joins, which are introduced by the SubqueryToJoinRule."?

mrks: @mrzzzrm @Bouncner

Bouncner: Yep, a noob as I am one is confused by opaque. @mrks sentence sounds good.

mrks: Changed it yet another bit to clarify that semi/anti joins might ruin our day even without the SubqueryToJoinRule doing its thing. Please have another look and merge this if it makes sense.

mrks: Also added a comment that came up in a discussion with @krichly

PR #1861, opened 2019-09-22
Make DisjunctionStrategy test deterministic

### LINKED ISSUE #1857: InExpressionRewriteRuleTest.DisjunctionStrategy fails
clang 7.0.0, apparently using `__GLIBCXX__`:

```
Differing subtrees
-------------- Actual LQP --------------
[0] [Predicate] col_a = 4
 \_[1] [MockNode 'Unnamed'] pruned: 0/2 columns

------------- Expected LQP -------------
[0] [Predicate] col_a = 3
 \_[1] [MockNode 'Unnamed'] pruned: 0/2 columns
-------------..............-------------
/home/moritz/Coding/hyrise/src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp:110: Failure
Failed
```

If the output of something is by nature indeterministic, maybe it is best to not test strict order. Maybe use visit_lqp to collect the nodes and assert that it's four UnionAll nodes followed by the predicates in any order (or something to that effect).

mrks:
fixes #1857

### DIFF
--- a/src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp
+++ b/src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp
@@ -112,7 +112,8 @@ TEST_F(InExpressionRewriteRuleTest, DisjunctionStrategy) {
     verify_predicate_node(current_node);
 
     std::sort(values_found_in_predicates.begin(), values_found_in_predicates.end());
-    EXPECT_EQ(values_found_in_predicates, std::vector<int>{1, 2, 3, 4, 5});
+    const auto expected_values = std::vector<int>{1, 2, 3, 4, 5};
+    EXPECT_EQ(values_found_in_predicates, expected_values);
   }
 
   {
--- a/src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp
+++ b/src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp
@@ -80,34 +80,39 @@ TEST_F(InExpressionRewriteRuleTest, DisjunctionStrategy) {
     const auto input_lqp = PredicateNode::make(five_element_in_expression, node);
     const auto result_lqp = StrategyBaseTest::apply_rule(rule, input_lqp);
 
-    // clang-format off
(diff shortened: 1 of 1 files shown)


src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp
```diff
@@ -80,34 +80,39 @@ TEST_F(InExpressionRewriteRuleTest, DisjunctionStrategy) {
...
+    // Checks that a given node is a predicate of the form `col_a = x` where x is an int and will be added to
+    // values_found_in_predicates
+    const auto verify_predicate_node = [&](const auto& node) {
+      EXPECT_EQ(node->type, LQPNodeType::Predicate);
+      auto predicate_node = std::dynamic_pointer_cast<PredicateNode>(node);
+      EXPECT_TRUE(predicate_node);
+      auto predicate = std::dynamic_pointer_cast<BinaryPredicateExpression>(predicate_node->predicate());
+      EXPECT_TRUE(predicate);
+      EXPECT_EQ(predicate->left_operand(), col_a);
+      EXPECT_EQ(predicate->right_operand()->type, ExpressionType::Value);
```
mrzzzrm: ASSERT_EQ, otherwise the next line would be executed (and result in UB, likely?)

mrks: Yepp. I honestly don't see too much of a value in `EXPECT_` anyway.

src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp
```diff
@@ -80,34 +80,40 @@ TEST_F(InExpressionRewriteRuleTest, DisjunctionStrategy) {
     const auto input_lqp = PredicateNode::make(five_element_in_expression, node);
```
Bouncner: Not yours, but `five_element` misses an s.

mrks: I wouldn't say so. It is a five-element InExpression. Shouldn't matter, anyway.

PR #1862, opened 2019-09-23
Add new plotting scripts

mrks:


scripts/plot_performance_breakdown.py
```diff
@@ -0,0 +1,66 @@
...
+
+    sum_parse_duration = 0.0
+    sum_sql_translation_duration = 0.0
+    sum_optimization_duration = 0.0
+    sum_lqp_translation_duration = 0.0
+    sum_plan_execution_duration = 0.0
+
+    for run in benchmark_json['successful_runs']:
+        if len(run['metrics']) == 0:
+            print("No metrics found. Did you run the benchmark with --sql_metrics?")
```
Bouncner: `Exit("No metrics found. Did you run the benchmark with --sql_metrics?")` to return a non-success return code.

scripts/plot_performance_breakdown.py
```diff
@@ -0,0 +1,66 @@
...
+    benchmark.append(sum_lqp_translation_duration / len(benchmark_json['successful_runs']))
+    benchmark.append(sum_plan_execution_duration / len(benchmark_json['successful_runs']))
+
+    benchmarks.append(benchmark)
+
+df = pd.DataFrame(benchmarks, columns=['Benchmark', 'Parser', 'SQLTranslator', 'Optimizer', 'LQPTranslator', 'Execution'])
+
+# Normalize data from nanoseconds to percentage of total cost
+df.iloc[:,1:] = df.iloc[:,1:].apply(lambda x: x / x.sum(), axis=1)
+print(df)
```
Bouncner: Debugging?

scripts/plot_performance_breakdown.py
```diff
@@ -0,0 +1,66 @@
...
+
+ax = df.plot.bar(x='Benchmark', stacked=True)
```
Bouncner: It should be obvious, I would nonetheless add a legend to the y axis.

scripts/plot_performance_breakdown.py
```diff
@@ -0,0 +1,66 @@
...
+ax.set_yticklabels(['{:,.0%}'.format(x) for x in ax.get_yticks()])
+
+# Reverse legend so that it matches the stacked bars
+handles, labels = ax.get_legend_handles_labels()
+ax.legend(reversed(handles), reversed(labels))
```
Bouncner: In the email you sent around, the legend was floating on top of the bars which can potentially be annoying. Maybe place it right next to the plot?

scripts/plot_performance_breakdown.py
```diff
@@ -0,0 +1,66 @@
...
+
+basename = sys.argv[1].replace('.json', '')
+plt.tight_layout()
+plt.savefig(basename + '_breakdown.png')
```
Bouncner: Is there a reason to save a `png` and not a `pdf`?

scripts/plot_performance_over_commit_range.py
```diff
@@ -0,0 +1,62 @@
...
+		# If column headers are unset, set them
+		if len(column_names) == 1:
+			for name in re.findall(r'name": "(.*)"', file.read()):
+				column_names.append(name)
+
+			file.seek(0)
+
+		# Quick and dirty way to retrieve the throughput. This is "better" than parsing the JSON because while the JSON has
+		# changed over time, this field has been available for a long time.
+		for items_per_second in re.findall(r'items_per_second": ([0-9]+\.[0-9]*)', file.read()):
```
Bouncner: Is this something we should have a binary/systemTest (or whatever they are called) for?

scripts/plot_performance_over_commit_range.py
```diff
@@ -0,0 +1,62 @@
...
+
+df = pd.DataFrame(results)
+df.columns = column_names
+
+# Normalize dataframe so that all values are relative to the first throughput of that benchmark
+df.iloc[:,1:] = df.iloc[:,1:].apply(lambda x: x / x[0], axis=0)
+
+print(df)
+
+df.plot.line(x='Commit', figsize=(64, 18))
```
Bouncner: Same here for the y axis.

scripts/plot_performance_breakdown.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+
+# Given a benchmark result json that was generated with --sql_metrics, this script plots the cost of the different
+# steps (parsing, optimization, ...) as a normalized stacked bar chart.
```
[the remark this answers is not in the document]
mrks: I don't see any value in listing them here, but potential for confusion if we ever add a step and this is not updated.

scripts/plot_performance_breakdown.py
mrks: I like seeing the numeric values, too.

scripts/plot_performance_breakdown.py
mrks: pngs are easier to `imgcat`.

scripts/plot_performance_over_commit_range.py
mrks: I would consider that overkill. If this plot script doesn't work anymore, it is sufficient to fix it once it becomes a problem.

scripts/plot_performance_breakdown.py
Bouncner: Then let both plots be pngs?

scripts/plot_performance_over_commit_range.py
Bouncner: Yes, it sounds like an overkill. But isn't the overhead pretty big once we realize after 15 commits that some changes broke this script?

scripts/plot_performance_breakdown.py
mrks: The other one is 1.5 MB as a png, wouldn't want to imgcat that anyway.

scripts/plot_performance_over_commit_range.py
mrks: So what if it's broken? Fixing it will take the same amount of time no matter when we notice it.

PR #1866, opened 2019-09-25
Forward a chunk's ordered-by information to GetTable

Bouncner:
Currently, the order-by information is not carried through to the table scan. This PR adapts (i) the abstract table generator to add the sorted information to the chunks (in case sorting is requested for the benchmark) and (ii) the get table operator to forward and adapt (in case of column pruning) this information.

For the benchmarks, basically nothing changed. Only TPC-H Q3 has a qualifying scan, but the impact is not significant.
```
+----------------+-------------------+------+-------------------+------+------------+---------------------------------+
| Benchmark      | prev. iter/s      | runs | new iter/s        | runs | change [%] | p-value (significant if <0.001) |
+----------------+-------------------+------+-------------------+------+------------+---------------------------------+
| TPC-H 03       | 0.825697124004364 | 199  | 0.830946326255798 | 200  | +1%        |                          0.0075 |
| geometric mean |                   |      |                   |      | +1%        |                                 |
+----------------+-------------------+------+-------------------+------+------------+---------------------------------+
```

### DIFF
--- a/src/lib/operators/get_table.cpp
+++ b/src/lib/operators/get_table.cpp
@@ -155,6 +155,10 @@ std::shared_ptr<const Table> GetTable::_on_execute() {
     // The Chunk is to be included in the output Table, now we progress to excluding Columns
     const auto stored_chunk = stored_table->get_chunk(stored_chunk_id);
 
+    // Make a copy of the order-by information of the current chunk. This information is adapted when columns are
+    // pruned and will be set on the outputted chunk.
+    std::optional<std::pair<ColumnID, OrderByMode>> chunk_ordered_by = stored_chunk->ordered_by();
+
     if (_pruned_column_ids.empty()) {
       *output_chunks_iter = stored_chunk;
     } else {
@@ -170,6 +174,11 @@ std::shared_ptr<const Table> GetTable::_on_execute() {
           continue;
         }
 
+        if (chunk_ordered_by && chunk_ordered_by->first == stored_column_id) {
+          chunk_ordered_by->first = ColumnID{static_cast<uint16_t>(std::distance(_pruned_column_ids.begin(),
+                                                                                 pruned_column_ids_iter))};
+        }
+
         *output_segments_iter = stored_chunk->get_segment(stored_column_id);
         auto indexes = stored_chunk->get_indexes({*output_segments_iter});
         if (!indexes.empty()) {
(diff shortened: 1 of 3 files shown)


src/lib/operators/get_table.cpp
```diff
@@ -155,6 +155,10 @@ std::shared_ptr<const Table> GetTable::_on_execute() {
     // The Chunk is to be included in the output Table, now we progress to excluding Columns
     const auto stored_chunk = stored_table->get_chunk(stored_chunk_id);
 
+    // Make a copy of the order-by information of the current chunk. This information is adapted when columns are
+    // pruned and will be set on the outputted chunk.
+    std::optional<std::pair<ColumnID, OrderByMode>> chunk_ordered_by = stored_chunk->ordered_by();
```
mrks: auto

src/lib/operators/get_table.cpp
```diff
@@ -180,6 +189,9 @@ std::shared_ptr<const Table> GetTable::_on_execute() {
 
       *output_chunks_iter = std::make_shared<Chunk>(std::move(output_segments), stored_chunk->mvcc_data(),
                                                     stored_chunk->get_allocator(), std::move(output_indexes));
+      if (chunk_ordered_by) {
+        (*output_chunks_iter)->set_ordered_by(*chunk_ordered_by);
```
mrks: What if the column was pruned and the then clause in line 177 was never executed? Unless I am missing something, this is also missing a test.

Bouncner: Good catch. Should be solved. Test added which failed without the change.

src/test/operators/get_table_test.cpp
```diff
@@ -270,9 +271,22 @@ TEST_F(OperatorsGetTableTest, AdaptOrderByInformation) {
...
+  // pruning the columns on which chunks are sorted
+  {
+    auto get_table = std::make_shared<opossum::GetTable>("int_int_float", std::vector<ChunkID>{},
+                                                         std::vector{ColumnID{0}, ColumnID{2}});
+    get_table->execute();
+
+    auto get_table_output = get_table->get_output();
+    EXPECT_EQ(get_table_output->column_count(), 1);
+    EXPECT_FALSE(get_table_output->get_chunk(ChunkID{0})->ordered_by().has_value());
+    EXPECT_FALSE(get_table_output->get_chunk(ChunkID{0})->ordered_by().has_value());
```
mrks: No need for `has_value()` - optional automatically decays to bool.

Bouncner: I thought it’s more obvious and better readable. Will change.

PR #1871, opened 2019-09-29
Blacklist query 88

### LINKED ISSUE #1870: Blacklist TPC-DS query 88
Because query 88 cannot be verified until we resolve #1822, it should be blacklisted to avoid confusion.

janetzki:
Fix #1870


mrks:
Superseded by #1872. I had to add it too the blacklist as part of that PR so that I could add the TPC-DS to CI.

PR #1872, opened 2019-09-29
Fix TPC-DS 91, blacklist 88, add to CI

### LINKED ISSUE #1867: TPC-DS query 91 fails on current master
```
> hyriseBenchmarkTPCDS --verify -r 1
...
Using SQLite's result table as expected result table
- Verification failed: Expected result table is not empty, but Hyrise's actual result is (48 µs 903 ns)
```
9bc4938489dd284723b7e38157b6296539d2a330 works, while dd18d60344228a3312cb32537839cbc704416841 does not anymore. Investigation is needed here to find out which commit and what caused this bug.
Not applying the (first) `PredicateSplitUpRule` seems to fix at least the verification (worked on the branch for #1791).

### LINKED ISSUE #1870: Blacklist TPC-DS query 88

mrks:
closes #1867 
closes #1870

### DIFF
--- a/src/lib/statistics/statistics_objects/min_max_filter.cpp
+++ b/src/lib/statistics/statistics_objects/min_max_filter.cpp
@@ -96,7 +96,6 @@ template <typename T>
 bool MinMaxFilter<T>::does_not_contain(const PredicateCondition predicate_condition,
                                        const AllTypeVariant& variant_value,
                                        const std::optional<AllTypeVariant>& variant_value2) const {
-
   // Early exit for NULL variants.
   if (variant_is_null(variant_value)) {
     return false;
--- a/src/lib/statistics/statistics_objects/min_max_filter.cpp
+++ b/src/lib/statistics/statistics_objects/min_max_filter.cpp
@@ -55,19 +55,19 @@ std::shared_ptr<AbstractStatisticsObject> MinMaxFilter<T>::sliced(
       break;
 
     case PredicateCondition::BetweenInclusive: {
-      Assert(static_cast<bool>(variant_value2), "Between operator needs two values.");
+      Assert(variant_value2, "Between operator needs two values.");
       const auto value2 = boost::get<T>(*variant_value2);
       return sliced(PredicateCondition::GreaterThanEquals, value)->sliced(PredicateCondition::LessThanEquals, value2);
     }
 
     case PredicateCondition::BetweenLowerExclusive: {
-      Assert(static_cast<bool>(variant_value2), "Between operator needs two values.");
+      Assert(variant_value2, "Between operator needs two values.");
       const auto value2 = boost::get<T>(*variant_value2);
       return sliced(PredicateCondition::GreaterThan, value)->sliced(PredicateCondition::LessThanEquals, value2);
     }
 
     case PredicateCondition::BetweenUpperExclusive: {
-      Assert(static_cast<bool>(variant_value2), "Between operator needs two values.");
+      Assert(variant_value2, "Between operator needs two values.");
       const auto value2 = boost::get<T>(*variant_value2);
       return sliced(PredicateCondition::GreaterThanEquals, value)->sliced(PredicateCondition::LessThan, value2);
     }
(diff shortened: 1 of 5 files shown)


src/lib/statistics/statistics_objects/min_max_filter.cpp
```diff
@@ -127,17 +127,17 @@ bool MinMaxFilter<T>::does_not_contain(const PredicateCondition predicate_condit
     case PredicateCondition::BetweenLowerExclusive: {
       Assert(static_cast<bool>(variant_value2), "Between operator needs two values.");
       const auto value2 = boost::get<T>(*variant_value2);
```
Bouncner: Get rid of `boost::get`?

src/lib/statistics/statistics_objects/min_max_filter.cpp
```diff
@@ -127,17 +127,17 @@ bool MinMaxFilter<T>::does_not_contain(const PredicateCondition predicate_condit
...
+      return value >= max || value2 < min;
     }
     case PredicateCondition::BetweenUpperExclusive: {
       Assert(static_cast<bool>(variant_value2), "Between operator needs two values.");
       const auto value2 = boost::get<T>(*variant_value2);
-      return value >= max || value2 < min;
+      return value > max || value2 <= min;
     }
     case PredicateCondition::BetweenExclusive: {
       Assert(static_cast<bool>(variant_value2), "Between operator needs two values.");
```
Bouncner: I think `.has_value()` is clearer than such a cast on an optional.

src/lib/statistics/statistics_objects/min_max_filter.cpp
```diff
@@ -127,17 +127,17 @@ bool MinMaxFilter<T>::does_not_contain(const PredicateCondition predicate_condit
...
       const auto value2 = boost::get<T>(*variant_value2);
-      return value > max || value2 <= min;
+      return value >= max || value2 < min;
     }
     case PredicateCondition::BetweenUpperExclusive: {
       Assert(static_cast<bool>(variant_value2), "Between operator needs two values.");
       const auto value2 = boost::get<T>(*variant_value2);
-      return value >= max || value2 < min;
+      return value > max || value2 <= min;
```
Bouncner: Help me here. `BetweenUpperExclusive` would mean a search interval as `[5, 10)`.
If my minimum is 5, I don't see how this change makes sense.
Same for `BetweenLowerExclusive`.

src/lib/statistics/statistics_objects/min_max_filter.cpp
mrks: In favor of?

src/lib/statistics/statistics_objects/min_max_filter.cpp
mrks: Copied code. I'll just use operator bool.

src/lib/statistics/statistics_objects/min_max_filter.cpp
Bouncner: Never mind. Mixed something up.

src/lib/statistics/statistics_objects/min_max_filter.cpp
mrks: Took me some time to wrap my head around it, but didn't find a better way to explain it.

Which side of the disjunction are you asking about? As you mention min, probably the right side. That uses value2, not value. In your example 10 is not smaller or equal to 5, so we cannot prune. If value2 was 5, e.g., [2, 5), the 5 would be exclusive, so we could prune.

mrks: Added two examples, hopefully those help.

Bouncner: Shouldn't review late at night. I think it makes perfect sense. Will take another look on the comments/tests next week.

PR #1879, opened 2019-10-09
Format compare_benchmarks.py output for Github

mrks:
I always liked the colored version of this better. Now, we can have it on Github, too.

```diff
 +----------------+----------------+------+---------------+------+------------+-----------------------------------------------+
 | Benchmark      | prev. iter/s   | runs | new iter/s    | runs | change [%] |               p-value (significant if <0.001) |
 +----------------+----------------+------+---------------+------+------------+-----------------------------------------------+
+| 01             | 20.7994117737  | 208  | 21.9238319397 | 220  | +5%        |                   (run time too short) 0.0000 |
+| 03             | 16.0511054993  | 161  | 28.5994300842 | 287  | +78%       |                   (run time too short) 0.0000 |
+| 06             | 16.8967514038  | 169  | 29.4084148407 | 295  | +74%       |                   (run time too short) 0.0000 |
-| 07             | 11.2185573578  | 113  | 8.90366268158 | 90   | -21%       |                   (run time too short) 0.0000 |
...

### DIFF
--- a/scripts/compare_benchmarks.py
+++ b/scripts/compare_benchmarks.py
@@ -64,8 +64,11 @@ def calculate_and_format_p_value(old, new):
     return colored(notes + "{0:.4f}".format(p_value), color)
 
 
-if(len(sys.argv) != 3):
-    exit("Usage: " + sys.argv[0] + " benchmark1.json benchmark2.json")
+if(len(sys.argv) != 3 and len(sys.argv) != 4):
+    exit("Usage: " + sys.argv[0] + " benchmark1.json benchmark2.json [--github]")
+
+# Format the output as a diff (prepending - and +) so that Github shows colors
+github_format = bool(len(sys.argv) == 4 and sys.argv[3] == '--github')
 
 with open(sys.argv[1]) as old_file:
     old_data = json.load(old_file)
@@ -123,6 +126,32 @@ table_data.append(['geometric mean', '', '', '', '', format_diff(geometric_mean(
 table = AsciiTable(table_data)
 table.justify_columns[6] = 'right'
 
+# If github_format is set, format the output in the style of a diff file where added lines (starting with +) are
+# colored green, removed lines (starting with -) are red, and others (starting with an empty space) are black.
+# Because terminaltables (unsurprisingly) does not support this hack, we need to post-process the result string,
+# searching for the control codes that define lines formatted as green or red.
+
+result = str(table.table)
+if github_format:
+    new_result = ''
+    green_control_sequence = colored('', 'green')[0:5]
+    red_control_sequence = colored('', 'red')[0:5]
+
+    for line in result.splitlines():
+        if green_control_sequence in line:
+            new_result += '+'
+        elif red_control_sequence in line:
+            new_result += '-'
+        else:
+            new_result += ' '
+
+        new_result += line + '\n'
+    result = new_result
+
 print("")
-print(table.table)
+if github_format:
+    print("```diff")
+print(result)
+if github_format:
+    print("```")
 print("")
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/compare_benchmarks.py [hunk 1]
```diff
@@ -64,8 +64,11 @@ def calculate_and_format_p_value(old, new):
     return colored(notes + "{0:.4f}".format(p_value), color)
 
 
-if(len(sys.argv) != 3):
-    exit("Usage: " + sys.argv[0] + " benchmark1.json benchmark2.json")
+if(len(sys.argv) != 3 and len(sys.argv) != 4):
+    exit("Usage: " + sys.argv[0] + " benchmark1.json benchmark2.json [--github]")
+
+# Format the output as a diff (prepending - and +) so that Github shows colors
+github_format = bool(len(sys.argv) == 4 and sys.argv[3] == '--github')
```
mrks: Not really nice, but I did not want to add full argument parsing just for this. We could also simply make this the default - maybe people can live with ` ```diff ` prepended and `+`/`-` in front of the table.

Bouncner:
Count me as an opponent of having this as the default.
Nice change though.

scripts/compare_benchmarks.py
```diff
@@ -64,8 +64,11 @@ def calculate_and_format_p_value(old, new):
     return colored(notes + "{0:.4f}".format(p_value), color)
 
 
-if(len(sys.argv) != 3):
-    exit("Usage: " + sys.argv[0] + " benchmark1.json benchmark2.json")
+if(len(sys.argv) != 3 and len(sys.argv) != 4):
```
Bouncner: `in [3,4]`?

Bensk1:
> Count me as an opponent of having this as the default.
> Nice change though.

Why?

scripts/compare_benchmarks.py
```diff
@@ -123,6 +126,32 @@ def calculate_and_format_p_value(old, new):
...
+            new_result += '-'
+        else:
+            new_result += ' '
+
+        new_result += line + '\n'
+    result = new_result
+
 print("")
-print(table.table)
+if github_format:
```
Bouncner: Why not add those `diff` markers to `new_result` in the conditional above and do a simple `print(result)` here?

Bouncner:
> > Count me as an opponent of having this as the default.
> > Nice change though.
> 
> Why?

Opponent might be too strong of a word, but I simply wouldn't like the `diff` to be always printed. I wouldn't veto, but I also wouldn't vote for it.

tobodner:
Since this change is about looks: Can the benchmark and p-value columns be more narrow s.t. I do not have to scroll sideways to see the entire table on my 1920 pixel display?

Bouncner:
> Since this change is about looks: Can the benchmark and p-value columns be more narrow s.t. I do not have to scroll sideways to see the entire table on my 1920x1200 pixel display?

https://www.apple.com/de/pro-display-xdr/

mrks:
:D 

Good idea though - I run into the same problem when using split tabs.

scripts/compare_benchmarks.py
Bensk1: I would leave it as it is now.

scripts/compare_benchmarks.py
```diff
@@ -79,7 +81,7 @@ def calculate_and_format_p_value(old, new):
 diffs = []
 
 table_data = []
-table_data.append(["Benchmark", "prev. iter/s", "runs", "new iter/s", "runs", "change [%]", "p-value (significant if <" + str(p_value_significance_threshold) + ")"])
```
Bensk1: What's the reason for not showing the significance threshold anymore?

mrks: Makes the column shorter. Did you ever look at it?

tobodner:
Sorry - this was not by intention.

scripts/compare_benchmarks.py
Bensk1: Just asking.

Issue #1142, opened 2018-09-26
Return types of functions connected to custom types


timzimmermann:
We use types such as `ColumnID` and `ChunkOffset` often, but kind of lack a way to have consistent return types for functions related to them.

Here are two examples.

From [table.hpp](https://github.com/hyrise/hyrise/blob/master/src/lib/storage/table.hpp):

```c++
size_t column_count() const;

// return the maximum chunk size (cannot exceed ChunkOffset (uint32_t))
uint32_t max_chunk_size() const;
```

From [chunk.hpp](https://github.com/hyrise/hyrise/blob/master/src/lib/storage/chunk.hpp):

```c++
// returns the number of columns, which is equal to the number of segments (cannot exceed ColumnID (uint16_t))
uint16_t column_count() const;

// returns the number of rows (cannot exceed ChunkOffset (uint32_t))
uint32_t size() const;
```
From [chunk.cpp](https://github.com/hyrise/hyrise/blob/master/src/lib/storage/chunk.cpp):

```c++
uint32_t Chunk::size() const {
  if (_segments.empty()) return 0;
  auto first_segment = get_segment(ColumnID{0});
  return first_segment->size();
}
```

From [base_segment.hpp](https://github.com/hyrise/hyrise/blob/master/src/lib/storage/base_segment.hpp):

```c++
// returns the number of values
virtual size_t size() const = 0;
```

Column count returns either a `size_t` or `uint16_t`, depending on whether you call it on a chunk or table.
The `ChunkOffset`-related `size()` methods do return the same type across the different files, except for `BaseSegment`. That's pretty interesting, since there should be type narrowing in `Chunk::size()` (`size_t` to `uint32_t`) and I would have expected a compiler warning, or what am I missing?

In any case, I came across this same question of what return type to use for histograms.
The maximum number of elements in a bin is `ChunkOffset`, since histograms are per segment.
However, using that as a type in histogram files is confusing, so I decided to define an alias type (`using HistogramCountType = ChunkOffset`.

Similarly, we might want to think about defining aliases such as `using ChunkSize = ChunkOffset`, and using that as the return type.
I know that this adds more types and there might be a little confusion about the difference between `ChunkOffset` and `ChunkSize`, but the advantage is that changing types will not lead to overflow errors if we use those aliases rigidly.

Do you guys have any opinions on that?

timzimmermann:
Yep, duplicate.

I do vaguely remember those discussions now, but it keeps annoying me.
[...]

PR #1882, opened 2019-10-14
Optimizing Validation: validating entire chunks

Bensk1:
This is an optimization for the validate operator. The optimization is made possible by our chunk concept. Validate checks whether rows are visible for a certain transaction. Usually, the visibility is evaluated for each individual row in a chunk. Instead, we ca use a shortcut and evaluate the visibility on a chunk level with previously aggregated visibility information. To avoid complicated code paths and costly maintenance of this aggregate, we only take the shortcut for unmodified chunks. The performance numbers will look even better once the new PosList is implemented because right now, even with the optimization in place, we still lose a lot of time creating PosLists for entirely visible data tables.

This PR introduces the `Chunk::finalize()` method that makes a chunk immutable and determines the `max_begin_cid`. `Finalize()` should be called by actors that create and/or populate chunks. For now, we only finalize chunks in `load_table` and `Table::append()` and refrained from including finalization into the Insert operator for this PR because it is more complex than expected. It is not necessary that all actors support finalization for this optimization to work, but it might not be used in as many cases as it would be with the finalization being part of the Insert operator. Some code and tests for finalization as part of Inset can be found on the bensk1/insert_finalization branch.

The idea came up looking at TPC-DS execution plans where Validate took a significant part of the total execution time.

```
+----------------+----------------+------+----------------+------+------------+---------+
| Benchmark      | prev. iter/s   | runs | new iter/s     | runs | change [%] | p-value |
+----------------+----------------+------+----------------+------+------------+---------+
| 01             | 20.8229789734  | 1250 | 21.4769287109  | 1289 | +3%        |  0.0000 |
| 03             | 16.1746921539  | 971  | 21.316160202   | 1280 | +32%       |  0.0000 |
| 06             | 18.2594776154  | 1096 | 23.6341075897  | 1419 | +29%       |  0.0000 |
| 07             | 11.4016485214  | 685  | 13.6715450287  | 821  | +20%       |  0.0000 |
| 09             | 4.03387069702  | 243  | 6.08354091644  | 366  | +51%       |  0.0000 |
| 10             | 7.53198289871  | 452  | 9.40933704376  | 565  | +25%       |  0.0000 |
| 13             | 6.89585113525  | 414  | 7.60978603363  | 457  | +10%       |  0.0000 |
| 15             | 32.2875328064  | 1938 | 40.1408653259  | 2409 | +24%       |  0.0000 |
| 17             | 10.1142520905  | 607  | 13.6604118347  | 820  | +35%       |  0.0000 |
| 19             | 13.6174287796  | 818  | 18.6774997711  | 1121 | +37%       |  0.0000 |
| 25             | 11.8455982208  | 711  | 18.960193634   | 1138 | +60%       |  0.0000 |
| 26             | 23.657661438   | 1420 | 27.7607860565  | 1666 | +17%       |  0.0000 |
| 28             | 9.28837871552  | 558  | 11.6008443832  | 697  | +25%       |  0.0000 |
| 29             | 12.0817813873  | 725  | 17.6609573364  | 1060 | +46%       |  0.0000 |
| 31             | 1.16481041908  | 70   | 1.18550777435  | 72   | +2%        |  0.0000 |
| 34             | 14.2747545242  | 857  | 17.9322986603  | 1076 | +26%       |  0.0000 |
| 35             | 5.01705932617  | 302  | 5.81432914734  | 349  | +16%       |  0.0000 |
| 39a            | 0.557417094707 | 34   | 0.589927613735 | 36   | +6%        |  0.0000 |
| 39b            | 0.56584751606  | 34   | 0.589458405972 | 36   | +4%        |  0.0000 |
| 41             | 44.1935806274  | 2652 | 44.544467926   | 2673 | +1%        |  0.0000 |
| 42             | 14.5804862976  | 875  | 17.5901870728  | 1056 | +21%       |  0.0000 |
| 43             | 2.98624897003  | 180  | 3.05034589767  | 184  | +2%        |  0.0000 |
| 45             | 35.8152923584  | 2149 | 41.5248832703  | 2492 | +16%       |  0.0000 |
| 48             | 6.24047756195  | 375  | 7.21275091171  | 433  | +16%       |  0.0000 |
| 50             | 19.8338432312  | 1191 | 23.5320072174  | 1412 | +19%       |  0.0000 |
| 52             | 15.2885856628  | 918  | 17.8944892883  | 1074 | +17%       |  0.0000 |
| 55             | 15.9066343307  | 955  | 19.9012813568  | 1195 | +25%       |  0.0000 |
| 62             | 9.35123825073  | 562  | 9.47346591949  | 569  | +1%        |  0.0000 |
| 65             | 8.49001789093  | 510  | 9.11015415192  | 547  | +7%        |  0.0000 |
| 69             | 7.6755695343   | 461  | 9.50316333771  | 571  | +24%       |  0.0000 |
| 73             | 19.009098053   | 1141 | 23.1425285339  | 1389 | +22%       |  0.0000 |
| 79             | 4.8556060791   | 292  | 6.31667661667  | 380  | +30%       |  0.0000 |
| 81             | 34.8951911926  | 2094 | 36.2786521912  | 2177 | +4%        |  0.0000 |
| 83             | 50.7468833923  | 3045 | 58.0060958862  | 3481 | +14%       |  0.0000 |
| 91             | 28.234331131   | 1695 | 39.6882629395  | 2382 | +41%       |  0.0000 |
| 93             | 3.51358890533  | 211  | 3.62807941437  | 218  | +3%        |  0.0000 |
| 96             | 19.0695419312  | 1145 | 25.957447052   | 1558 | +36%       |  0.0000 |
| 97             | 1.97125303745  | 119  | 2.13553953171  | 129  | +8%        |  0.0000 |
| 99             | 4.67023420334  | 281  | 4.76261806488  | 286  | +2%        |  0.0000 |
| geometric mean |                |      |                |      | +19%       |         |
+----------------+----------------+------+----------------+------+------------+---------+
```

```
+----------------+----------------+------+----------------+-------+------------+----------------------+
| Benchmark      | prev. iter/s   | runs | new iter/s     | runs  | change [%] |              p-value |
+----------------+----------------+------+----------------+-------+------------+----------------------+
| TPC-H 01       | 0.602200984955 | 37   | 0.615667939186 | 37    | +2%        |               0.0009 |
| TPC-H 02       | 12.5324497223  | 752  | 13.022269249   | 782   | +4%        |               0.0000 |
| TPC-H 03       | 6.51425695419  | 391  | 7.37465000153  | 443   | +13%       |               0.0000 |
| TPC-H 04       | 7.79239416122  | 468  | 7.81475830078  | 469   | +0%        |               0.3783 |
| TPC-H 05       | 4.23157024384  | 254  | 4.69575595856  | 282   | +11%       |               0.0000 |
| TPC-H 06       | 147.493591309  | 8850 | 188.035568237  | 10000 | +27%       | (run time too short) |
| TPC-H 07       | 2.26774930954  | 137  | 2.32301664352  | 140   | +2%        |               0.0000 |
| TPC-H 08       | 5.12774562836  | 308  | 6.5526008606   | 394   | +28%       |               0.0000 |
| TPC-H 09       | 1.56087589264  | 94   | 1.61465251446  | 97    | +3%        |               0.0000 |
| TPC-H 10       | 2.95491504669  | 178  | 2.97246289253  | 179   | +1%        |               0.5161 |
| TPC-H 11       | 25.2029018402  | 1513 | 30.4436721802  | 1827  | +21%       |               0.0000 |
| TPC-H 12       | 8.39993095398  | 504  | 9.22503852844  | 554   | +10%       |               0.0000 |
| TPC-H 13       | 2.30922913551  | 139  | 2.34749341011  | 141   | +2%        |               0.0000 |
| TPC-H 14       | 46.5742950439  | 2795 | 48.7405929565  | 2925  | +5%        |               0.0000 |
| TPC-H 15       | 69.6107254028  | 4177 | 75.2728729248  | 4517  | +8%        |               0.0000 |
| TPC-H 16       | 7.31294155121  | 439  | 7.52288007736  | 452   | +3%        |               0.0000 |
| TPC-H 17       | 1.27624535561  | 77   | 1.3589631319   | 82    | +6%        |               0.0000 |
| TPC-H 18       | 0.771143615246 | 47   | 0.759111702442 | 46    | -2%        |               0.0191 |
| TPC-H 19       | 17.3327331543  | 1040 | 21.8195152283  | 1310  | +26%       |               0.0000 |
| TPC-H 20       | 2.82564306259  | 170  | 2.62062358856  | 158   | -7%        |               0.0000 |
| TPC-H 21       | 1.48859798908  | 90   | 1.54123795033  | 93    | +4%        |               0.0000 |
| TPC-H 22       | 13.2607707977  | 796  | 14.7057847977  | 883   | +11%       |               0.0000 |
| geometric mean |                |      |                |       | +8%        |                      |
+----------------+----------------+------+----------------+-------+------------+----------------------+
```


src/lib/operators/validate.cpp
```diff
@@ -130,21 +157,28 @@ void Validate::_validate_chunks(const std::shared_ptr<const Table>& in_table, co
...
-        for (auto row_id : pos_list_in) {
-          if (opossum::is_row_visible(our_tid, snapshot_commit_id, row_id.chunk_offset, *mvcc_data)) {
-            pos_list_out->emplace_back(row_id);
+        if (check_visibility_on_chunk_level &&
+            is_entire_chunk_visible(referenced_chunk, snapshot_commit_id, mvcc_data)) {
+          // We can simply copy the whole PosList since it is entirely visible. Copying is forbidden for PosLists.
+          // Because of that, we use memcpy here for now.
+          const auto pos_list_size = pos_list_in.size();
+          pos_list_out->resize(pos_list_size);
+          std::memcpy(pos_list_out->data(), pos_list_in.data(), pos_list_in.size() * sizeof(RowID));
```
Bensk1: At one point we discussed adding a copy method to the PosList for another issue. Maybe this could be a solution here.

mrks:
Also pinging @julianmenzler who, thanks to his work on the MVCC cleanup, has some understanding of the MVCC intricacies.

src/lib/storage/chunk.cpp
```diff
@@ -102,6 +102,17 @@ std::vector<std::shared_ptr<AbstractIndex>> Chunk::get_indexes(
...
 
+void Chunk::update_max_begin_cid() {
+  DebugAssert((has_mvcc_data()), "Chunk does not have mvcc data");
+  auto mvcc = get_scoped_mvcc_data_lock();
+
+  if (mvcc->begin_cids.empty()) {
+    return;
+  }
+
+  mvcc->max_begin_cid = *(std::max_element(mvcc->begin_cids.begin(), mvcc->begin_cids.end()));
```
mrks: Assert that it is not `MAX_COMMIT_ID`

src/lib/operators/validate.cpp
```diff
@@ -71,6 +80,23 @@ std::shared_ptr<const Table> Validate::_on_execute(std::shared_ptr<TransactionCo
   auto job_end_chunk_id = ChunkID{0};
   auto job_row_count = uint32_t{0};
 
+  // Checking visibility on chunk level is an optimization to reduce the work done during validation.
+  // Instead of checking each row for visibility, we determine the largest begin CommitID (max_begin_cid) for an entire
```
mrks: Start by explaining the ratio behind it and why it is possible, not by explaining what steps are done

src/lib/operators/validate.cpp
```diff
@@ -71,6 +80,23 @@ std::shared_ptr<const Table> Validate::_on_execute(std::shared_ptr<TransactionCo
...
+  // chunk and check this one for visibility. Determining max_begin_cid is not the responsibility of the Validate
+  // operator itself.
+  // There are more possible optimizations but right now, we are only applying this optimization if the chunk does not
```
mrks: "There are more possible optimizations but right now" - pointless

src/lib/operators/validate.cpp
```diff
@@ -71,6 +80,23 @@ std::shared_ptr<const Table> Validate::_on_execute(std::shared_ptr<TransactionCo
...
+  // contain any invalid rows and if no delete operators are registered for the same transaction.
+  bool check_visibility_on_chunk_level = true;
+
+  // Not allowed if a delete is registerd for the same transaction context
+  const auto& rw_operators = transaction_context->read_write_operators();
+  for (const auto& rw_operator : rw_operators) {
```
mrks: read_write

src/lib/operators/validate.cpp
```diff
@@ -71,6 +80,23 @@ std::shared_ptr<const Table> Validate::_on_execute(std::shared_ptr<TransactionCo
   auto job_end_chunk_id = ChunkID{0};
   auto job_row_count = uint32_t{0};
 
+  // Checking visibility on chunk level is an optimization to reduce the work done during validation.
+  // Instead of checking each row for visibility, we determine the largest begin CommitID (max_begin_cid) for an entire
+  // chunk and check this one for visibility. Determining max_begin_cid is not the responsibility of the Validate
+  // operator itself.
+  // There are more possible optimizations but right now, we are only applying this optimization if the chunk does not
+  // contain any invalid rows and if no delete operators are registered for the same transaction.
+  bool check_visibility_on_chunk_level = true;
```
mrks: Naming: can_use_chunk_shortcut?

src/lib/operators/validate.cpp
```diff
@@ -130,21 +157,28 @@ void Validate::_validate_chunks(const std::shared_ptr<const Table>& in_table, co
...
 
         const auto referenced_chunk = referenced_table->get_chunk(pos_list_in.common_chunk_id());
         auto mvcc_data = referenced_chunk->get_scoped_mvcc_data_lock();
 
-        for (auto row_id : pos_list_in) {
-          if (opossum::is_row_visible(our_tid, snapshot_commit_id, row_id.chunk_offset, *mvcc_data)) {
-            pos_list_out->emplace_back(row_id);
+        if (check_visibility_on_chunk_level &&
+            is_entire_chunk_visible(referenced_chunk, snapshot_commit_id, mvcc_data)) {
+          // We can simply copy the whole PosList since it is entirely visible. Copying is forbidden for PosLists.
```
mrks: But... why would you want to copy it in the first place instead of simply re-using it?

src/lib/operators/validate.cpp
```diff
@@ -165,18 +199,26 @@ void Validate::_validate_chunks(const std::shared_ptr<const Table>& in_table, co
         output_segments.push_back(ref_segment_out);
       }
 
-      // Otherwise we have a Value- or DictionarySegment and simply iterate over all rows to build a poslist.
+      // Otherwise we have a non-reference Segment and simply iterate over all rows to build a poslist.
```
mrks: The "otherwise" is now too far away from the initial comment to still make sense to a reader.

src/lib/storage/chunk.cpp
```diff
@@ -102,6 +102,17 @@ std::vector<std::shared_ptr<AbstractIndex>> Chunk::get_indexes(
   return result;
 }
 
+void Chunk::update_max_begin_cid() {
+  DebugAssert((has_mvcc_data()), "Chunk does not have mvcc data");
+  auto mvcc = get_scoped_mvcc_data_lock();
+
+  if (mvcc->begin_cids.empty()) {
+    return;
```
mrks: When would has_mvcc_data be true and begin_cids be empty?

src/lib/storage/chunk.cpp
```diff
@@ -102,6 +102,17 @@ std::vector<std::shared_ptr<AbstractIndex>> Chunk::get_indexes(
   return result;
 }
 
+void Chunk::update_max_begin_cid() {
+  DebugAssert((has_mvcc_data()), "Chunk does not have mvcc data");
```
mrks: extra parentheses?

src/lib/storage/table.cpp
```diff
@@ -200,7 +205,10 @@ void Table::append_chunk(const Segments& segments, std::shared_ptr<MvccData> mvc
   }
 #endif
 
-  _chunks.push_back(std::make_shared<Chunk>(segments, mvcc_data, alloc));
+  auto c = std::make_shared<Chunk>(segments, mvcc_data, alloc);
```
mrks: I wish we could teach clang-tidy to prohibit single character identifiers

src/lib/storage/table.cpp
```diff
@@ -200,7 +205,10 @@ void Table::append_chunk(const Segments& segments, std::shared_ptr<MvccData> mvc
...
+  _chunks.push_back(c);
+
+  if (mvcc_data) c->update_max_begin_cid();
```
mrks: This does not really feel like something that the table should do

src/lib/operators/validate.cpp
```diff
@@ -36,6 +37,14 @@ bool Validate::is_row_visible(TransactionID our_tid, CommitID snapshot_commit_id
   return snapshot_commit_id < end_cid && ((snapshot_commit_id >= begin_cid) != (row_tid == our_tid));
 }
 
+bool Validate::is_entire_chunk_visible(const std::shared_ptr<const Chunk>& chunk, CommitID snapshot_commit_id,
+                                       const SharedScopedLockingPtr<MvccData>& mvcc_data) {
```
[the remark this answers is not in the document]
Bensk1: Sure, I don't mind.

src/lib/operators/validate.cpp
Bensk1: I thought the following `else` adds enough context. What do you suggest, simply removing the "otherwise"?

src/lib/storage/chunk.cpp
Bensk1: I don't know. Could it be that the chunk points to MVCC data but does not have entries yet?

src/lib/storage/table.cpp
Bensk1: I could imagine a periodic task that does that. However, this is not going to happen soon. Just remove it here?

src/lib/storage/table.cpp
Bensk1: Right. I think I got inspired by one of the test cases.

src/lib/storage/table.cpp
```diff
@@ -117,6 +117,11 @@ ColumnID Table::column_id_by_name(const std::string& column_name) const {
 void Table::append(const std::vector<AllTypeVariant>& values) {
   auto last_chunk = !_chunks.empty() ? get_chunk(ChunkID{chunk_count() - 1}) : nullptr;
   if (!last_chunk || last_chunk->size() >= _max_chunk_size) {
+    // One chunk reached its capacity just now. We prepare this chunk for MVCC optimizations.
+    if (last_chunk && _use_mvcc == UseMvcc::Yes) {
```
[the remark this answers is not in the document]
Bensk1: Yes, because a chunk is either created for an empty table or for a non-empty one where the last chunk got filled up. Checking for `!_chunks.empty()` would just be a repetition of line 118. What do you suggest?

src/lib/storage/chunk.hpp
```diff
@@ -181,6 +181,8 @@ class Chunk : private Noncopyable {
 
   void set_cleanup_commit_id(CommitID cleanup_commit_id);
 
+  void update_max_begin_cid();
```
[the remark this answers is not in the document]
Bensk1: For now yes. There might be cases later where we update it. But I am happy to rename it to `set` or `determine`...

src/lib/operators/validate.cpp
Bensk1: What do you mean? Setting a flag and changing the part further down where the actual ReferenceSegment is constructed?

https://github.com/hyrise/hyrise/blob/master/src/lib/operators/validate.cpp#L164

src/lib/operators/validate.cpp
Bensk1: Better now.

src/lib/storage/chunk.cpp
Bensk1: Good catch

src/lib/storage/table.cpp
```diff
@@ -117,6 +117,11 @@ ColumnID Table::column_id_by_name(const std::string& column_name) const {
...
+      last_chunk->update_max_begin_cid();
```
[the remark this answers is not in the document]
Bensk1: Let's discuss this offline. I would also be happier to have this in a central place and have it called by an outside entity. E.g., at the same place where filters are generated and data is encoded for our benchmarks.

src/lib/operators/validate.cpp
mrks: I might be missing something, but why can't we pass the very same poslist into the newly constructed ReferenceSegment?

Bensk1: Let's discuss that offline tomorrow.

src/lib/storage/table.cpp
mrks: How about a `finalize_chunk` (name tbd) method that
* sets Chunk::_is_mutable (that should not be the job of the encoder)
* sets the max_begin_cid
* potentially encodes the chunk or schedules it for encoding
* potentially builds indexes, converts indexes, or schedules a job doing that
* ...

src/lib/storage/chunk.cpp
mrks: No, I'd assert for it to not be empty.

src/lib/operators/validate.cpp
mrks: never mind

src/lib/operators/validate.cpp
Bensk1: Done.

src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -74,7 +73,18 @@ void AbstractTableGenerator::generate_and_store() {
   }
 
   /**
-   * Encode the Tables
+   * Finalizing all chunks of all tables
```
Bensk1: In my opinion we don't need output and time measurements for finalizing, what do you think?

src/lib/operators/insert.cpp
```diff
@@ -240,6 +247,7 @@ void Insert::_on_rollback_records() {
     for (auto chunk_offset = target_chunk_range.begin_chunk_offset; chunk_offset < target_chunk_range.end_chunk_offset;
          ++chunk_offset) {
       mvcc_data->end_cids[chunk_offset] = 0u;
+      target_chunk->increase_invalid_row_count(1u);
```
Bensk1: Is there a reason why we didn't do that before?

src/lib/storage/chunk.hpp
```diff
@@ -181,6 +181,12 @@ class Chunk : private Noncopyable {
 
   void set_cleanup_commit_id(CommitID cleanup_commit_id);
 
+  /**
+     * Executes tasks that are connected with finalizing a chunk. Currently, chunks are made immutable and
+     * the MVCC max_begin_cid is set. Finalizing a chunk is the insert's responsibility.
```
Bensk1: `inserter's`

mrks:
You'll ping me once you want me to have a look at this, right?

Bensk1:
Ping!

Bensk1:
Anti ping! I found an issue when I was running the TPCC benchmark. I will come back to you soon.

Bensk1:
Ping!!

src/lib/operators/get_table.cpp
```diff
@@ -194,9 +194,12 @@ std::shared_ptr<const Table> GetTable::_on_execute() {
 
       *output_chunks_iter = std::make_shared<Chunk>(std::move(output_segments), stored_chunk->mvcc_data(),
                                                     stored_chunk->get_allocator(), std::move(output_indexes));
+
       if (adapted_chunk_order) {
         (*output_chunks_iter)->set_ordered_by(*adapted_chunk_order);
       }
+
+      (*output_chunks_iter)->increase_invalid_row_count(stored_chunk->invalid_row_count());
```
mrks: Add a comment?

src/lib/operators/insert.cpp
mrks: Probably because we forgot it :D

src/lib/operators/insert.cpp
```diff
@@ -106,13 +106,18 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
 
     while (remaining_rows > 0) {
       auto target_chunk_id = ChunkID{_target_table->chunk_count() - 1};
       auto target_chunk = _target_table->get_chunk(target_chunk_id);
 
       // If the last Chunk of the target Table is either immutable or full, append a new mutable Chunk
       if (!target_chunk->is_mutable() || target_chunk->size() == _target_table->max_chunk_size()) {
+        // Finalize the last Chunk only if it was not added by us.
+        if (target_chunk->is_mutable() && chunks_before_insert - 1 == target_chunk_id)
+          target_chunk->finalize();
```
mrks: Wait a second - how can we be sure that everyone who is writing on this table has set their begin CIDs?

mrks: Have a look at `ChunkCompressionTask::_chunk_is_completed`. Maybe we can (1) merge those and (2) make sure that once the last insert tasks finishes or aborts, it is responsible for finalizing the chunk. Good potential for race conditions though.

Also, this shows that we probably have a missing test.

mrks:
Did you run the TPCC multithreaded and with multiple clients?

Bensk1:
> Did you run the TPCC multithreaded and with multiple clients?

Not yet. But you are right, if we run a multithreaded TPCC, the finalization should crash.

src/lib/operators/insert.cpp
Bensk1: The content of `_chunk_is_completed` makes sense, but how do we find out which Insert is the last? A solution would be a periodic task that checks for chunks that reached their capacity before finalizing and encoding them.

Bensk1:
> > Did you run the TPCC multithreaded and with multiple clients?
> 
> Not yet. But you are right, if we run a multithreaded TPCC, the finalization should crash.

Even without this validation optimization, I get: `std::logic_error: src/lib/concurrency/transaction_context.cpp:183 Illegal phase transition detected.` when running `./cmake-build-release/hyriseBenchmarkTPCC --scheduler --clients 10` on 5b81e2fc7a94d2380440b74ded9ab9824576df1c on mrks/tpcc.

src/lib/operators/insert.cpp
mrks: I tried to come up with a way where we identify the last writing transaction, but it will be hard to prove thread safety for those. As of now, I would consider a separate task and doing this asynchronously more intuitive. Maybe this could be merged with the MVCC cleanup?

Bensk1: Let's discuss that offline.

Bensk1: We decided to remove the finalization from the Insert operator for this PR. The validation optimization can be used without it, too. It might not be applied for as many cases as it would be with the finalization in Insert. The code and test were moved to bensk1/insert_finalization

Bouncner: Could you add a summary of this discussion/outcome to the description?

I can’t follow.

Bouncner:
> For now, we only finalize chunks in `load_table` and `Table::append()`

What about the chunk encoder? We set the immutability flag there, so I'd guess that's also a point where we could make the chunk as finalized?

Bensk1:
The chunk encoder does not set the immutability flag anymore. It should not be the responsibility of chunk consumers.

Bouncner:
What is a chunk consumer?

Bensk1:
It should be the responsibility of the inserter/actor that appends a chunk.

src/test/operators/insert_test.cpp
```diff
@@ -44,14 +44,18 @@ TEST_F(OperatorsInsertTest, SelfInsert) {
...
   // Check that row has been inserted.
-  EXPECT_EQ(t->get_chunk(ChunkID{0})->size(), 6u);
+  EXPECT_EQ(t->row_count(), 6u);
+  EXPECT_EQ(t->get_chunk(ChunkID{0})->size(), 3u);
+  EXPECT_EQ(t->get_chunk(ChunkID{1})->size(), 3u);
   EXPECT_EQ((*t->get_chunk(ChunkID{0})->get_segment(ColumnID{1}))[0], AllTypeVariant(12345));
   EXPECT_EQ((*t->get_chunk(ChunkID{0})->get_segment(ColumnID{0}))[0], AllTypeVariant(458.7f));
-  EXPECT_EQ((*t->get_chunk(ChunkID{0})->get_segment(ColumnID{1}))[3], AllTypeVariant(12345));
-  EXPECT_EQ((*t->get_chunk(ChunkID{0})->get_segment(ColumnID{0}))[3], AllTypeVariant(458.7f));
+  EXPECT_EQ((*t->get_chunk(ChunkID{1})->get_segment(ColumnID{1}))[0], AllTypeVariant(12345));
```
mrks: This is because chunk 0 was previously not marked as immutable, correct?

src/test/operators/projection_test.cpp
```diff
@@ -53,6 +54,37 @@ TEST_F(OperatorsProjectionTest, ExecutedOnAllChunks) {
                             load_table("resources/test_data/tbl/projection/int_float_add.tbl"));
 }
 
+TEST_F(OperatorsProjectionTest, InvalidRowCount) {
```
mrks: PassThroughInvalidRowCount?

src/lib/operators/validate.cpp
```diff
@@ -36,6 +37,17 @@ bool Validate::is_row_visible(TransactionID our_tid, CommitID snapshot_commit_id
   return snapshot_commit_id < end_cid && ((snapshot_commit_id >= begin_cid) != (row_tid == our_tid));
 }
 
+bool Validate::is_entire_chunk_visible(const std::shared_ptr<const Chunk>& chunk, const CommitID snapshot_commit_id) {
+  DebugAssert(!std::dynamic_pointer_cast<const ReferenceSegment>(chunk->get_segment(ColumnID{0})),
+    "is_entire_chunk_visible cannot be called on reference chunks.");
+
+  const auto mvcc_data = chunk->mvcc_data();
```
mrks: Personally, I prefer `const auto&` for cases like this. If `mvcc_data()` ever returns a reference, we are on the safe side.

src/lib/operators/projection.cpp
```diff
@@ -107,8 +107,11 @@ std::shared_ptr<const Table> Projection::_on_execute() {
     const auto input_chunk = input_table.get_chunk(chunk_id);
     Assert(input_chunk, "Did not expect deleted chunk here.");  // see #1686
 
+    // The output chunk contains all the rows that are in the stored chunk, including invalid rows. We have to keep
+    // MVCC data and invalid_row_count because following operators might rely on that.
```
mrks: see above

src/lib/operators/insert.cpp
```diff
@@ -106,13 +106,18 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
     if (_target_table->chunk_count() == 0) {
       _target_table->append_mutable_chunk();
     }
+    const auto chunks_before_insert = _target_table->chunk_count();
```
mrks: chunk_count_...

src/lib/operators/insert.cpp
```diff
@@ -106,13 +106,18 @@ std::shared_ptr<const Table> Insert::_on_execute(std::shared_ptr<TransactionCont
...
 
     while (remaining_rows > 0) {
       auto target_chunk_id = ChunkID{_target_table->chunk_count() - 1};
       auto target_chunk = _target_table->get_chunk(target_chunk_id);
 
       // If the last Chunk of the target Table is either immutable or full, append a new mutable Chunk
       if (!target_chunk->is_mutable() || target_chunk->size() == _target_table->max_chunk_size()) {
+        // Finalize the last Chunk only if it was not added by us.
```
mrks: why?

src/test/operators/validate_test.cpp
```diff
@@ -52,7 +65,10 @@ void OperatorsValidateTest::set_all_records_visible(Table& table) {
 }
 
 void OperatorsValidateTest::set_record_invisible_for(Table& table, RowID row, CommitID end_cid) {
-  table.get_chunk(row.chunk_id)->get_scoped_mvcc_data_lock()->end_cids[row.chunk_offset] = end_cid;
+  auto chunk = table.get_chunk(row.chunk_id);
```
mrks: I don't like `set_record_invisible_for` - what does "for" refer to? Just `set_record_invisible` or, even better `invalidate_record`?

src/test/operators/validate_test.cpp
```diff
@@ -85,6 +101,91 @@ TEST_F(OperatorsValidateTest, ScanValidate) {
...
+  Segments empty_segment = {std::make_shared<ReferenceSegment>(_test_table, ColumnID{0}, pos_list)};
+  auto chunk = std::make_shared<Chunk>(empty_segment);
+
+  EXPECT_THROW(Validate::is_entire_chunk_visible(chunk, snapshot_cid), std::logic_error);
+}
+
+TEST_F(OperatorsValidateTest, ChunkNotEntirelyVisibleWithoutMaxBeginCid) {
+  auto snapshot_cid = CommitID{1};
+  Segments empty_segment = {std::make_shared<ValueSegment<int32_t>>()};
+  auto chunk = std::make_shared<Chunk>(empty_segment, std::make_shared<MvccData>(0, 0));
```
mrks: I'd feel more comfortable if we would not test on empty segments (other tests as well).

src/lib/operators/validate.hpp
```diff
@@ -26,11 +26,14 @@ class Validate : public AbstractReadOnlyOperator {
   static bool is_row_visible(TransactionID our_tid, CommitID snapshot_commit_id, const TransactionID row_tid,
                              const CommitID begin_cid, const CommitID end_cid);
 
+  static bool is_entire_chunk_visible(const std::shared_ptr<const Chunk>& chunk, const CommitID snapshot_commit_id);
```
mrks: Doc? Also explain that this is only valid after checking `can_use_chunk_shortcut`.

src/lib/operators/validate.cpp
```diff
@@ -74,6 +86,23 @@ std::shared_ptr<const Table> Validate::_on_execute(std::shared_ptr<TransactionCo
   auto job_end_chunk_id = ChunkID{0};
   auto job_row_count = uint32_t{0};
 
+  // In some cases, we can identify a chunk as being entirely visible for the current transaction. Simply said,
+  // if the youngest row in a chunk is visible, all other rows are older and hence visible, too. This
+  // (1) all rows in the chunk have been commited (i.e., their begin_cid has been set),
```
mrks: committed

src/lib/operators/validate.cpp
```diff
@@ -74,6 +86,23 @@ std::shared_ptr<const Table> Validate::_on_execute(std::shared_ptr<TransactionCo
...
+  // (2) the highest begin_cid in the chunk is lower/equal than the snapshot_cid of the transaction
+  //     (the max_begin_cid is stored in the chunk, not determined by the ValidateOperator),
+  // (3) no rows in the chunk have been deleted and committed,
```
mrks: no rows in the chunk have been invalidated before this transaction was started

src/lib/operators/validate.cpp
```diff
@@ -74,6 +86,23 @@ std::shared_ptr<const Table> Validate::_on_execute(std::shared_ptr<TransactionCo
   auto job_end_chunk_id = ChunkID{0};
   auto job_row_count = uint32_t{0};
 
+  // In some cases, we can identify a chunk as being entirely visible for the current transaction. Simply said,
+  // if the youngest row in a chunk is visible, all other rows are older and hence visible, too. This
+  // (1) all rows in the chunk have been commited (i.e., their begin_cid has been set),
+  // (2) the highest begin_cid in the chunk is lower/equal than the snapshot_cid of the transaction
```
mrks: (1) lower than / equal to
(2) meaning that all rows have been inserted and committed before this transaction was started

src/lib/operators/validate.cpp
mrks: (0) the chunk is immutable, i.e., no new rows can be added while this transaction is being executed

src/lib/operators/validate.cpp
```diff
@@ -111,13 +140,15 @@ std::shared_ptr<const Table> Validate::_on_execute(std::shared_ptr<TransactionCo
...
+                                std::vector<std::shared_ptr<Chunk>>& output_chunks, std::mutex& output_mutex,
+                                const bool can_use_chunk_shortcut) {
   for (auto chunk_id = chunk_id_start; chunk_id <= chunk_id_end; ++chunk_id) {
     const auto chunk_in = in_table->get_chunk(chunk_id);
     Assert(chunk_in, "Did not expect deleted chunk here.");  // see #1686
 
     Segments output_segments;
-    auto pos_list_out = std::make_shared<PosList>();
+    auto pos_list_out = std::make_shared<const PosList>();
+    PosList internal_mutable_pos_list;
```
mrks: temp_pos_list?

src/lib/storage/chunk.hpp
```diff
@@ -162,13 +162,13 @@ class Chunk : private Noncopyable {
   /**
    * Returns the count of deleted/invalidated rows within this chunk resulting from already committed transactions.
    */
-  uint64_t invalid_row_count() const { return _invalid_row_count.load(); }
+  uint32_t invalid_row_count() const { return _invalid_row_count.load(); }
```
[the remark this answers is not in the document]
Bensk1: I added this sentence to the header file.

src/lib/storage/mvcc_data.hpp
```diff
@@ -22,6 +22,8 @@ struct MvccData {
   pmr_concurrent_vector<CommitID> begin_cids;                  ///< commit id when record was added
   pmr_concurrent_vector<CommitID> end_cids;                    ///< commit id when record was deleted
 
+  std::optional<CommitID> max_begin_cid;
```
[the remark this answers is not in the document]
Bensk1: Oops. This should definitely be tracked.

src/lib/storage/table.cpp
```diff
@@ -64,7 +64,7 @@ TableType Table::type() const { return _type; }
 
 UseMvcc Table::has_mvcc() const { return _use_mvcc; }
 
-size_t Table::column_count() const { return _column_definitions.size(); }
+uint16_t Table::column_count() const { return static_cast<uint16_t>(_column_definitions.size()); }
```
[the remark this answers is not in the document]
Bensk1: Is a DebugAssert sufficient?

src/lib/storage/table.cpp
```diff
@@ -200,7 +205,8 @@ void Table::append_chunk(const Segments& segments, std::shared_ptr<MvccData> mvc
   }
 #endif
 
-  _chunks.push_back(std::make_shared<Chunk>(segments, mvcc_data, alloc));
+  auto chunk = std::make_shared<Chunk>(segments, mvcc_data, alloc);
+  _chunks.push_back(chunk);
```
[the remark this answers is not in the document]
Bensk1: Leftover from a different implementation.

src/lib/storage/table.cpp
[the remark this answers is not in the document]
mrks: I feel like that only adds to the confusion. If you find changing these types important, feel free to add a type `using ColumnCount = ColumnID::base_type` or a strong typedef and we will tackle #1142 one step at a time

src/test/storage/chunk_encoder_test.cpp
```diff
@@ -169,6 +184,7 @@ TEST_F(ChunkEncoderTest, EncodeMultipleChunksUsingSameEncoding) {
   const auto segment_encoding_spec = SegmentEncodingSpec{EncodingType::Dictionary};
   const auto chunk_encoding_spec = ChunkEncodingSpec{3u, segment_encoding_spec};
 
+  _table->get_chunk(static_cast<ChunkID>(ChunkID{2u}))->finalize();
```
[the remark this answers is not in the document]
mrks: > This could be used in many places

Could you rephrase that concern?

src/lib/storage/chunk.cpp
```diff
@@ -102,6 +102,19 @@ std::vector<std::shared_ptr<AbstractIndex>> Chunk::get_indexes(
   return result;
 }
 
+void Chunk::finalize() {
+  Assert(is_mutable(), "Only mutable chunks can be finalized. Chunks cannot be finalized twice.");
+  mark_immutable();
```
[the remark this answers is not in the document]
mrks: > didn't we talk about the ambiguity yet?

Please help me remember. Is that doc'ed somewhere?

src/test/operators/validate_test.cpp
Bensk1: Hm, I guess "for" should refer to the CommitID? But I wasn't the author of this method.

This does not proof anything, but there is a fair chance that you came up with it :D `c9fc17080f (Markus Dreseler 2017-03-23 21:59:28 +0100  67) void OperatorsValidateTest::set_record_invisible_for(Table& table, RowID row, CommitID end_cid) {`

src/lib/operators/insert.cpp
Bensk1: Chunks that we added can only be finalized after all rows have been committed or rolled back. But your are right, we cannot be sure that this is the case so we should not finalize here.

src/test/operators/validate_test.cpp
Bensk1: Fair. For the test above using an empty segment, it should not make a difference since it throws because of the reference segment.

src/lib/operators/validate.hpp
Bensk1: I could also imagine to make `is_entire_chunk_visible` a private non-static method and `can_use_chunk_shortcut` a member and have an assert on `can_use_chunk_shortcut`.

src/test/operators/validate_test.cpp
mrks: > it should not make a difference since it throws because of the reference segment

Sure, but when testing edge cases, we should have one edge case at a time.

src/test/operators/insert_test.cpp
Bensk1: Correct. Before completion load_table marks the last chunk as immutable. Therefor, Insert has to add a new chunk.

PR #1883, opened 2019-10-16
Retire JIT

### LINKED ISSUE #1569: JIT compilation fails on MacOS
The JIT query compilation of specialized code fails on MacOS with:
````
LLVM ERROR: Inline asm not supported by this streamer because we don't have an asm parser for this target
````
during `_compile_layer.addModule(module, resolver)` in [src/lib/operators/jit_operator/specialization/jit_compiler.cpp:44](https://github.com/hyrise/hyrise/blob/master/src/lib/operators/jit_operator/specialization/jit_compiler.cpp)

In the past this was only the case for optimized builds with optimization levels `-O2` and `-O3`. Now this is also the case for unoptimized debug builds.
I think that I first noticed this after a minor update of the MacOS operating system in late 2018.

### LINKED ISSUE #1606: JIT: Support scans with differing operand types
#1550 disables JITting of scans like `int_column < 5.5f`, because the JIT would cast `5.5f` to `5`, which obviously leads to incorrect results.

### LINKED ISSUE #1873: Warning while building jit_llvm_bundle.cpp
ubuntu, clang 7

```
In file included from /home/moritz/Coding/hyrise/cmake-build-debug-clang-7/src/lib/embed_llvm/jit_llvm_bundle.cpp:1:
/home/moritz/Coding/hyrise/src/lib/operators/validate.cpp:89:54: warning: explicit capture of 'this' with a capture default of '=' is a C++2a extension [-Wc++2a-extensions]
        jobs.push_back(std::make_shared<JobTask>([=, this, &output_chunks, &output_mutex] {
```

### PR DESCRIPTION
**Pull request by mrks**
This retires the JIT component. We make this step with a heavy heart as both students and ourselves have put a significant amount of time into it. The following reasons compel us to make this step:

1. With the dependencies to LLVM and the modified LLVM source files, upgrading to a newer LLVM version is difficult. We are still stuck with LLVM 7, while LLVM 8 has been out since March and LLVM 9 since September.

2. The performance of our JIT code is significantly worse than that of the non-JITted code (see below). With six months spent optimizing the JIT code, it became clear that the underlying architecture is a bottleneck. By now, even TPC-H query 1, which was the main show case, does not have an advantage anymore.

3. We are lacking the resources to work on this and improve it to a point where it delivers the performance generally expected from a JIT engine.

The following features have been removed as they are no longer used:
* `SQLPipelineBuilder::with_lqp_translator`, as we only have one LQPTranslator left
* `InsertLimitInExistsRule`, as the Limit operator is generally bad for performance in our non-JITted execution

`lqp_subplan_to_boolean_expression` is now unused, but since LQP utilities are always good to have, I would vote for leaving it.

closes #1873
closes #1606
closes #1569

```diff
 +----------------+----------------+------+----------------+------+------------+----------------------+
 | Benchmark      | prev. iter/s   | runs | new iter/s     | runs | change [%] |              p-value |
 +----------------+----------------+------+----------------+------+------------+----------------------+
 | TPC-H 01       | 0.580308973789 | 6    | 0.584861218929 | 6    | +1%        | (run time too short) |
-| TPC-H 02       | 8.2397146225   | 83   | 5.83198881149  | 59   | -29%       | (run time too short) |
 | TPC-H 03       | 2.9124212265   | 30   | 2.96221160889  | 30   | +2%        | (run time too short) |
-| TPC-H 04       | 6.17415618896  | 62   | 4.56229639053  | 46   | -26%       | (run time too short) |
-| TPC-H 05       | 1.8943977356   | 19   | 1.77639722824  | 18   | -6%        | (run time too short) |
 | TPC-H 06       | 97.7952957153  | 978  | 96.9379577637  | 970  | -1%        | (run time too short) |
-| TPC-H 07       | 1.66412949562  | 17   | 1.09279954433  | 11   | -34%       | (run time too short) |
-| TPC-H 08       | 2.14703297615  | 22   | 1.91438364983  | 20   | -11%       | (run time too short) |
-| TPC-H 09       | 0.992621421814 | 10   | 0.640674710274 | 7    | -35%       | (run time too short) |
 | TPC-H 10       | 2.18790173531  | 22   | 2.14246749878  | 22   | -2%        | (run time too short) |
 | TPC-H 11       | 8.97191524506  | 90   | 8.7804517746   | 88   | -2%        | (run time too short) |
-| TPC-H 12       | 4.59321498871  | 46   | 2.41232919693  | 25   | -47%       | (run time too short) |
-| TPC-H 13       | 2.096534729    | 21   | 0.169884979725 | 2    | -92%       | (run time too short) |
-| TPC-H 14       | 29.374671936   | 294  | 24.9425296783  | 250  | -15%       | (run time too short) |
-| TPC-H 15       | 41.5667572021  | 416  | 19.4448070526  | 195  | -53%       | (run time too short) |
-| TPC-H 16       | 5.42614173889  | 55   | 4.27060413361  | 43   | -21%       | (run time too short) |
-| TPC-H 17       | 0.756906867027 | 8    | 0.62147551775  | 7    | -18%       | (run time too short) |
-| TPC-H 18       | 0.571924626827 | 6    | 0.468188196421 | 5    | -18%       | (run time too short) |
-| TPC-H 19       | 7.9266409874   | 80   | 3.98039484024  | 40   | -50%       | (run time too short) |
-| TPC-H 20       | 2.27415323257  | 23   | 2.10240602493  | 22   | -8%        | (run time too short) |
-| TPC-H 21       | 0.993776023388 | 10   | 0.754742026329 | 8    | -24%       | (run time too short) |
-| TPC-H 22       | 6.5044093132   | 66   | 4.9977350235   | 50   | -23%       | (run time too short) |
-| geometric mean |                |      |                |      | -30%       |                      |
 +----------------+----------------+------+----------------+------+------------+----------------------+
```

Even with SF 10, this does not look much better, reducing the -30% performance loss to -21%.


Bensk1:
> Even with SF 10, this does not look much better, reducing the -30% performance loss to -21%.

Do you think that we would actually see a small improvement for SF 1000?

src/lib/utils/sqlite_wrapper.cpp
```diff
@@ -54,7 +54,6 @@ void SQLiteWrapper::create_table(const Table& table, const std::string& table_na
         column_types.emplace_back("TEXT");
         break;
       case DataType::Null:
-      case DataType::Bool:
```
Bensk1: Is bool a jit-related type?

mrks: Yes. See the deleted comment in types.hpp

mrks:
> Looks good, but it's hard to tell if we missed anything.

I grepped for JIT, so we should be good.
Did you have a look at the things mentioned in the PR description?

Bensk1:
> Did you have a look at the things mentioned in the PR description?

SQLPipelineBuilder::with_lqp_translator and InsertLimitInExistsRule? I saw that you removed them which makes sense.

mrks:
> > Even with SF 10, this does not look much better, reducing the -30% performance loss to -21%.
> 
> Do you think that we would actually see a small improvement for SF 1000?

With SF 100, JIT and non-JIT are much closer, however as larger scale factors put more and more pressure on the join, that is not unexpected.

mrks:
SF 100:

```diff
 +----------------+------------------+------+------------------+------+------------+-------------------+
 | Benchmark      | prev. iter/s     | runs | new iter/s       | runs | change [%] |           p-value |
 +----------------+------------------+------+------------------+------+------------+-------------------+
 | TPC-H 01       | 0.00478359917179 | 3    | 0.0049383896403  | 3    | +3%        | (not enough runs) |
+| TPC-H 02       | 0.0440288335085  | 27   | 0.046633567661   | 28   | +6%        |            0.0000 |
+| TPC-H 03       | 0.0206603966653  | 13   | 0.0217653047293  | 14   | +5%        |            0.0029 |
+| TPC-H 04       | 0.0312505029142  | 19   | 0.0294770915061  | 18   | -6%        |            0.0000 |
 | TPC-H 05       | 0.0117073049769  | 8    | 0.0114909810945  | 7    | -2%        | (not enough runs) |
 | TPC-H 06       | 0.854867219925   | 513  | 0.836318016052   | 502  | -2%        |            0.0038 |
-| TPC-H 07       | 0.00633114529774 | 4    | 0.00532157206908 | 4    | -16%       | (not enough runs) |
-| TPC-H 08       | 0.0156327728182  | 10   | 0.0148215228692  | 9    | -5%        | (not enough runs) |
-| TPC-H 09       | 0.00482178898528 | 3    | 0.00398001866415 | 3    | -17%       | (not enough runs) |
 | TPC-H 10       | 0.0096455514431  | 6    | 0.010112689808   | 7    | +5%        | (not enough runs) |
 | TPC-H 11       | 0.0703871920705  | 43   | 0.0708767846227  | 43   | +1%        |            0.0687 |
+| TPC-H 12       | 0.0383019000292  | 23   | 0.0244838446379  | 15   | -36%       |            0.0000 |
-| TPC-H 13       | 0.00943929702044 | 6    | 0.00150683685206 | 1    | -84%       | (not enough runs) |
+| TPC-H 14       | 0.16481795907    | 99   | 0.155446082354   | 94   | -6%        |            0.0000 |
+| TPC-H 15       | 0.140046909451   | 85   | 0.149029240012   | 90   | +6%        |            0.0000 |
+| TPC-H 16       | 0.0382567606866  | 23   | 0.0358681716025  | 22   | -6%        |            0.0000 |
-| TPC-H 17       | 0.00262302462943 | 2    | 0.00207191263326 | 2    | -21%       | (not enough runs) |
-| TPC-H 18       | 0.00260134320706 | 2    | 0.00211681216024 | 2    | -19%       | (not enough runs) |
+| TPC-H 19       | 0.0561725385487  | 34   | 0.0302615500987  | 19   | -46%       |            0.0000 |
 | TPC-H 20       | 0.00992708466947 | 6    | 0.00975871365517 | 6    | -2%        | (not enough runs) |
-| TPC-H 21       | 0.0060215331614  | 4    | 0.00512284459546 | 4    | -15%       | (not enough runs) |
 | TPC-H 22       | 0.0503090769053  | 31   | 0.0496789813042  | 30   | -1%        |            0.0020 |
-| geometric mean |                  |      |                  |      | -16%       |                   |
 +----------------+------------------+------+------------------+------+------------+-------------------+
```

PR #1885, opened 2019-10-17
CI: Replace TPC-H test with --verify

mrks:
When we started implementing the TPC-H, our performance was too bad to even execute SF 1 as part of CI. Thus, we had different scale factors that were used for the tests. I suspect that the time spent building those different TPC-H databases now surpasses that of the actual queries. Let's see how long `./hyriseBenchmarkTPCH --verify` takes on Jenkins.

As a bonus, we will now execute TPC-H on tsan runs with the scheduler, so that tsan actually makes some sense.

Also removes the tests for TPC-H/DS data generation from the sanitizer runs. As those won't change and are unlikely to suddenly develop asan/ubsan/tsan issues, it feels like we are wasting CI time there.


resources/test_data/tbl/tpch/sf-0.01/nation.tbl
```diff
@@ -0,0 +1,27 @@
+n_nationkey|n_name|n_regionkey|n_comment
```
mrks: I feel bad about this. If we keep the tests they are right now and want to continue verifying a against a tbl version that came from dbgen, we need scale factors that are actually legal. SF 0.001 (used before) was not.

src/lib/utils/sqlite_add_indices.cpp
```diff
@@ -31,10 +33,12 @@ void add_indices_to_sqlite(const std::string& schema_file_path, const std::strin
   sqlite_wrapper->raw_execute_query(schema_sql);
 
   // Add foreign keys
-  std::ifstream create_indices_file(create_indices_file_path);
-  std::string create_indices_sql((std::istreambuf_iterator<char>(create_indices_file)),
-                                 std::istreambuf_iterator<char>());
-  sqlite_wrapper->raw_execute_query(create_indices_sql);
+  if (!create_indices_file_path.empty()) {
```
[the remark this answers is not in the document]
Bensk1: Ah, now that I see the comments in the header, I believe I understand. The indexes could also be part of the schema.

src/test/tpc/tpch_db_generator_test.cpp
```diff
@@ -7,48 +7,16 @@
...
-  EXPECT_EQ(table_info_by_name.at("customer").table->row_count(), std::floor(150'000 * scale_factor));
-  EXPECT_EQ(table_info_by_name.at("orders").table->row_count(), std::floor(1'500'000 * scale_factor));
-  EXPECT_EQ(table_info_by_name.at("nation").table->row_count(), std::floor(25));
-  EXPECT_EQ(table_info_by_name.at("region").table->row_count(), std::floor(5));
-}
-
-TEST(TPCHTableGeneratorTest, TableContents) {
-  /**
-   * Check whether that data TPCHTableGenerator generates with a scale factor of 0.001 is the exact same that dbgen
+   * Check whether that data TPCHTableGenerator generates with a scale factor of 0.01 is the exact same that dbgen
```
Bensk1: whether that

src/test/tpc/tpch_db_generator_test.cpp
```diff
@@ -58,12 +26,20 @@ TEST(TPCHTableGeneratorTest, TableContents) {
   EXPECT_TABLE_EQ_ORDERED(table_info_by_name.at("nation").table, load_table(dir_001 + "nation.tbl", chunk_size));
   EXPECT_TABLE_EQ_ORDERED(table_info_by_name.at("region").table, load_table(dir_001 + "region.tbl", chunk_size));
 
+#if defined(__has_feature)
+#if (__has_feature(thread_sanitizer) || __has_feature(address_sanitizer))
+  // We verified thread and address safety above. As this is quite expensive to sanitize, don't perform the following
+  // check - double parantheses mark the code as explicitly dead.
+  if ((true)) return;
```
Bensk1: I don't understand. We want to break here in case sanitization is turned on? Why can't we just have a `return;`, why do we need explicitly dead code?

mrks: Any action required?

Bensk1: No.

PR #1889, opened 2019-10-19
Upgrade to clang-tidy-8

mrks:
* When returning a value, clang-tidy now correctly complains about that value being const. Those instances are now fixed.
* Changed some parameters to const references.
* Changed some defines to constexprs.
* Disabling some warnings that are too strict.


.clang-tidy-ignore
```diff
@@ -1,9 +1,6 @@
...
 
-src/lib/operators/jit_operator/specialization/llvm/InlineFunction.cpp  # File clone from LLVM - Hyrise code style does not apply
-src/lib/operators/jit_operator/specialization/llvm/CloneFunction.cpp   # dto.
-
-src/lib/server/client_connection.cpp                                   # clang-tidy doesn't like htons
-src/lib/server/postgres_wire_handler.cpp                               # dto.
-
-src/lib/utils/uninitialized_vector.hpp                                 # external source
\ No newline at end of file
+src/lib/server/client_connection.cpp            # clang-tidy doesn't like htons
```
Bouncner: htons?

.clang-tidy
```diff
@@ -1,40 +1,47 @@
...
+#   cppcoreguidelines-pro-bounds-pointer-arithmetic           Doesn't like DebugAssert
+#   google-runtime-references                                 Doesn't like mutable references
+#   cert-err58-cpp                                            We reeeeeally don't care about uncaught exceptions
+#   llvm-include-order                                        Handled by cpplint.py which is way faster
+#   clang-analyzer-cplusplus.NewDelete*                       False positives with shared_ptr::operator=
+#   cert-msc32-c, -cert-msc51-cpp                             Ok, so our generated tables are not cryptographically safe
+#   fuchsia-statically-constructed-objects                    We have too many static objects
+#   bugprone-exception-escape                                 We throw exceptions in many places (even destructors) and are completely fine with std::terminate
+#   hicpp-uppercase-literal-suffix                            Don't really care if it's 1.0f or 1.0F
+#   cert-dcl16-c                                              dto.
```
Bouncner: dto?

Jenkinsfile
```diff
@@ -140,7 +140,7 @@ try {
           stage("clang-debug:tidy") {
             if (env.BRANCH_NAME == 'master' || full_ci) {
               // We do not run tidy checks on the src/test folder, so there is no point in running the expensive clang-tidy for those files
-              sh "export CCACHE_BASEDIR=`pwd`; cd clang-debug-tidy && make hyrise hyriseBenchmarkFileBased hyriseBenchmarkTPCH hyriseBenchmarkJoinOrder hyriseConsole hyriseServer -l -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 3))"
+              sh "export CCACHE_BASEDIR=`pwd`; cd clang-debug-tidy && make hyrise hyriseBenchmarkFileBased hyriseBenchmarkTPCH hyriseBenchmarkJoinOrder hyriseConsole hyriseServer -k -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 3))"
```
Bouncner: Just curious? What does the change do?

scripts/clang_tidy_wrapper.sh
```diff
@@ -8,10 +8,11 @@ file=$1
 shift  # Remove the file from $@
 
 file_relative_to_source=${file//$cmake_source_dir/}
-
+file_relative_to_source=${file_relative_to_source/..\/src/src}
```
Bouncner: I'd guess there is some magic hidden with this double slash that I do not understand because bash.
I guess a simple comment would help here a lot.

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -218,7 +218,7 @@ void BenchmarkRunner::_benchmark_ordered() {
     result.duration = _state.benchmark_duration;
     const auto duration_of_all_runs_ns =
         static_cast<float>(std::chrono::duration_cast<std::chrono::nanoseconds>(_state.benchmark_duration).count());
-    const auto duration_seconds = duration_of_all_runs_ns / 1'000'000'000;
+    const auto duration_seconds = duration_of_all_runs_ns / 1e9f;
```
Bouncner: I hate the e notation and it doesn't add readability at all IMO.

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -352,7 +352,7 @@ void BenchmarkRunner::_create_report(std::ostream& stream) const {
         _config.benchmark_mode == BenchmarkMode::Shuffled ? _total_run_duration : result.duration;
     const auto reported_item_duration_ns =
         static_cast<float>(std::chrono::duration_cast<std::chrono::nanoseconds>(reported_item_duration).count());
-    const auto duration_seconds = reported_item_duration_ns / 1'000'000'000;
+    const auto duration_seconds = reported_item_duration_ns / 1e9f;
```
Bouncner: dto.

src/lib/memory/numa_memory_resource.cpp
```diff
@@ -3,7 +3,7 @@
 #include <string>
 
 #if HYRISE_NUMA_SUPPORT
-#define NUMA_MEMORY_RESOURCE_ARENA_SIZE 1llu << 30u
+constexpr auto NUMA_MEMORY_RESOURCE_ARENA_SIZE = 1llu << 30u;
```
Bouncner: Big fan of the various removals of `#define`.

Bouncner: But same as above: I really don't like those short cut definition of numbers. But probably to large here anyways.

src/lib/operators/export_binary.cpp
```diff
@@ -130,7 +130,10 @@ void ExportBinary::write_binary(const Table& table, const std::string& filename)
   }
 }
 
-const std::string ExportBinary::name() const { return "ExportBinary"; }
+const std::string& ExportBinary::name() const {
+  static const auto name = std::string{"ExportBinary"};
```
Bouncner: Never seen this. The `static` ensures that the string survives the stack cleanup?
I don't like that.

src/lib/operators/join_index.cpp
```diff
@@ -330,8 +333,8 @@ void JoinIndex::_reference_join_two_segments_using_index(
 }
 
 template <typename SegmentPosition>
-const std::vector<IndexRange> JoinIndex::_index_ranges_for_value(SegmentPosition probe_side_position,
-                                                                 const std::shared_ptr<AbstractIndex>& index) const {
+std::vector<IndexRange> JoinIndex::_index_ranges_for_value(SegmentPosition probe_side_position,
```
Bouncner: Can't `probe_side_position` not be const as well?

src/lib/operators/join_index.hpp
```diff
@@ -70,8 +70,8 @@ class JoinIndex : public AbstractJoinOperator {
                                                 const std::shared_ptr<const PosList>& reference_segment_pos_list);
 
   template <typename SegmentPosition>
-  const std::vector<IndexRange> _index_ranges_for_value(SegmentPosition probe_side_position,
-                                                        const std::shared_ptr<AbstractIndex>& index) const;
+  std::vector<IndexRange> _index_ranges_for_value(SegmentPosition probe_side_position,
```
Bouncner: dto.

src/lib/optimizer/strategy/expression_reduction_rule.cpp
```diff
@@ -232,7 +232,9 @@ void ExpressionReductionRule::remove_duplicate_aggregate(
     std::vector<std::shared_ptr<AbstractExpression>>& input_expressions,
     const std::shared_ptr<AbstractLQPNode>& aggregate_node, const std::shared_ptr<AbstractLQPNode>& root_node) {
   // Create a list of all sums, counts, and averages in the aggregate node.
-  std::vector<std::reference_wrapper<const std::shared_ptr<AbstractExpression>>> sums, counts, avgs;
+  std::vector<std::reference_wrapper<const std::shared_ptr<AbstractExpression>>> sums;
```
Bouncner: Looks somewhat ridiculous here.

src/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_nodes.cpp
```diff
@@ -205,8 +205,8 @@ AbstractIndex::Iterator ARTNode16::end() const {
 
 ARTNode48::ARTNode48(const std::vector<std::pair<uint8_t, std::shared_ptr<ARTNode>>>& children) {
   _index_to_child.fill(INVALID_INDEX);
-  for (uint8_t i = 0u; i < children.size(); ++i) {
-    _index_to_child[children[i].first] = i;
+  for (auto i = size_t{0}; i < children.size(); ++i) {
```
Bouncner: ASS. 

Well ... not really. But maybey `child_id`?

src/lib/storage/segment_accessor.cpp
```diff
@@ -13,7 +11,7 @@ std::unique_ptr<AbstractSegmentAccessor<T>> CreateSegmentAccessor<T>::create(
     using SegmentType = std::decay_t<decltype(typed_segment)>;
     if constexpr (std::is_same_v<SegmentType, ReferenceSegment>) {
       const auto& pos_list = *typed_segment.pos_list();
-      if (pos_list.references_single_chunk() && pos_list.size() > 0) {
+      if (pos_list.references_single_chunk() && !pos_list.empty()) {
```
Bouncner: Just curious: did clang-tidy bring that up?

.clang-tidy-ignore
mrks: https://linux.die.net/man/3/htons

Jenkinsfile
mrks: This should have always been `-k`, which means that the build should keep running even after encountering the first error. `-l` does nothing in this case.

scripts/clang_tidy_wrapper.sh
mrks: done

src/benchmarklib/benchmark_runner.cpp
mrks: Sublime doesn't like `1'000'000'000f;`, which breaks the formatting until the next `'`. Alternative would be `static_cast<float>(1'000'000'000)` - is that better?

src/lib/operators/export_binary.cpp
mrks: It would be nicer to make it `constexpr auto name = std::string{"ExportBinary"}`, but p0980, which allows `constexpr std::string` is [not supported by anyone](https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B2a_library_features) yet.

If we want to return a string reference, we don't have an alternative. The only other option would be to construct an std::string every time this method is called.

mrks: Actually, we could probably also do `constexpr auto name = "ExportBinary"; return std::string_view{name};` but I am not sure if that is better.

src/lib/operators/join_index.cpp
mrks: Absolutely. It's just that clang-tidy did not complain about it

src/lib/optimizer/strategy/expression_reduction_rule.cpp
mrks: In this place, yes. I still agree with the general rule. Is it worth making an exception here?

src/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_nodes.cpp
mrks: The whole ART is full of `i`s, some of which are child_ids, others are not. Once we start to use it, we should look at refactoring it.

src/lib/storage/segment_accessor.cpp
mrks: Yepp

src/lib/memory/numa_memory_resource.cpp
mrks: Here, we can express it as `1024 * ...`

.clang-tidy-ignore
Bouncner: Add a comment? I'd guess most people don't know.

src/benchmarklib/benchmark_runner.cpp
Bouncner: Not really better. But should Sublime's formatting be a reason for us?

src/lib/optimizer/strategy/expression_reduction_rule.cpp
Bouncner: Just looks stupid. But all the other places make sense.

So let's keep it this way.

src/lib/operators/export_binary.cpp
Bouncner: I still don't understand how the reference survives to an object that should be immediately freed?!?

Wouldn't it be better to store the string as part of the operator and return sth like `return _operator_name;`.

src/benchmarklib/benchmark_runner.cpp
mrks: Hmm, actually, Sublime is right:

```
floating-point-literal:
-> decimal-floating-point-literal
   (hexadecimal-floating-point-literal)

decimal-floating-point-literal:
    fractional-constant exponent-part(opt) floating-point-suffix(opt)
    digit-sequence exponent-part floating-point-suffix(opt)
```
(http://eel.is/c++draft/lex.fcon#nt:decimal-floating-point-literal)

Looks like to use a float literal, the number has to either be fractional or use an exponent. Unsure why the Lexer does not like `digit-sequence exponent-part(opt) floating-point-suffix(opt)`.

src/lib/operators/export_binary.cpp
mrks: > I still don't understand how the reference survives to an object that should be immediately freed?!?

https://en.cppreference.com/w/cpp/language/storage_duration#Static_local_variables

> Wouldn't it be better to store the string as part of the operator and return sth like return _operator_name;.

You wouldn't want each instance of the operator to have its own instance of the name. Thus, you would make it a static member. Other than breaking the encapsulation of this method, it wouldn't change much.

Bouncner: I just believe that you're right here.

PR #1890, opened 2019-10-19
Correct two words in LQP utils

Bouncner:
Working with LQP and stumbled upon this wording.

Not sure if we could merge `LQPUpwardVisitation` and `LQPVisitation`.


Bouncner:
Ah nice. Wasn't aware of this new functiion and also not aware that `visit_lqp` just traverses in a single direction.

Issue #1367, opened 2018-12-11
Binary import/export: Support all encoding types


mrzzzrm:
Currently only ValueSegments and DictionarySegments are supported, seemingly

mrks:
As far as I can see, only the FixedStringDictionarySegment is missing. Dumping the FixedStringVector (which is just a vector of characters) into a binary file should not be too difficult, so maybe we could get this done, too, and finally close this issue?

dey4ss:
If @j-tr is not quicker, I can do this after vacation.

PR #1896, opened 2019-10-23
Add Run-Length Segment binary import/export

dey4ss:
export and import RunLength encoded segments as requested in #1367


src/lib/import_export/binary.hpp
```diff
@@ -2,7 +2,7 @@
 
 namespace opossum {
 
-enum class BinarySegmentType : uint8_t { value_segment = 0, dictionary_segment = 1 };
+enum class BinarySegmentType : uint8_t { value_segment = 0, dictionary_segment = 1, run_length_segment = 2 };
```
Bouncner: Can't we use Hyrise's enums (types.hpp) directly here?

src/lib/operators/export_binary.cpp
```diff
@@ -77,6 +78,12 @@ void export_values(std::ofstream& ofstream, const std::vector<pmr_string>& value
 
 // specialized implementation for bool values
 template <>
+void export_values(std::ofstream& ofstream, const pmr_vector<bool>& values) {
+  // Cast to fixed-size format used in binary file
+  const auto writable_bools = std::vector<BoolAsByteType>(values.begin(), values.end());
+  export_values(ofstream, writable_bools);
```
Bouncner: Code duplication with void export_values(std::ofstream& ofstream, const std::vector<bool>& values).

src/test/operators/export_binary_test.cpp
```diff
@@ -215,6 +236,33 @@ TEST_F(OperatorsExportBinaryTest, AllTypesDictionarySegment) {
   EXPECT_TRUE(compare_files("resources/test_data/bin/AllTypesDictionarySegment.bin", filename));
 }
 
+TEST_F(OperatorsExportBinaryTest, AllTypesRunLengthSegment) {
```
Bouncner: We should probably use paramater templated tests here (see encoding tests). Or is there a good reason to not use them?
It's "only" a few lines right now, but will blow up once we have FoR, LZ4, ... tests for the binary export/import.

src/lib/storage/base_run_length_segment.hpp
```diff
@@ -0,0 +1,22 @@
+#pragma once
```
Bouncner: Not a big fan of that class. Is it only necessary because we want to import/export binary run length segments?

src/lib/operators/export_binary.cpp
mrks: Same for `void export_values(std::ofstream& ofstream, const pmr_vector<pmr_string>& values) {` (which is not your fault).

Suggestion: https://gist.github.com/mrks/6d6e9892850d694e396de857840c2c7a

Not sure what your C++ level is - if you want to discuss this approach, feel free to stop by.

src/test/operators/export_binary_test.cpp
```diff
@@ -164,6 +164,27 @@ TEST_F(OperatorsExportBinaryTest, FixedStringDictionarySegment) {
   EXPECT_TRUE(compare_files("resources/test_data/bin/StringDictionarySegment.bin", filename));
 }
 
+TEST_F(OperatorsExportBinaryTest, StringRunLengthSegment) {
+  TableColumnDefinitions column_definitions;
+  column_definitions.emplace_back("a", DataType::String, false);
+
+  auto table = std::make_shared<Table>(column_definitions, TableType::Data, 10);
+  table->append({"This"});
+  table->append({"is"});
```
mrks: For run-length segments, it makes sense to have run-lengths > 1 in the tested data

src/test/operators/export_binary_test.cpp
```diff
@@ -164,6 +164,27 @@ TEST_F(OperatorsExportBinaryTest, FixedStringDictionarySegment) {
   EXPECT_TRUE(compare_files("resources/test_data/bin/StringDictionarySegment.bin", filename));
 }
 
+TEST_F(OperatorsExportBinaryTest, StringRunLengthSegment) {
+  TableColumnDefinitions column_definitions;
+  column_definitions.emplace_back("a", DataType::String, false);
+
+  auto table = std::make_shared<Table>(column_definitions, TableType::Data, 10);
```
mrks: Choose chunk size so that we have more than one chunk?

src/lib/operators/export_binary.cpp
mrks: This just came to my mind: Why do we have `std::vector` here at all? All data (data in value segments, dictionaries, attribute vectors, run lengths, ...) should be stored in `pmr_vector`s. Unless I am missing something, the `std::vector` versions should be thrown out and any segment type that depends on it is wrong.

src/lib/storage/base_run_length_segment.hpp
dey4ss: Indeed the class is only necessary for exporting, as the concretion of the segment encoding is done via a visitor pattern. This would likely be omitted if every segment would encode itself, which on the other hand would probably mean a deeper design decision.

src/test/operators/export_binary_test.cpp
dey4ss: We now have a special multi-chunk test case with repeating values especially for RunLength-segments

src/lib/storage/base_run_length_segment.hpp
Bouncner: I think we should get rid of this class and all the visitors altogether. The visitor pattern has been removed from pretty much all other parts of Hyrise and I'd vote to remove it here as well.

I think a simple method that resolves the encoding type and forwards the segment accordingly (see `ChunkEncoder::encode_segment()`) should suffice.

mrks:
You'll remove WIP and ping us once you're ready

src/lib/operators/export_binary.cpp
Bouncner: Feel free to mark such issues as resolved once you have addressed them. At least when it's clear that there are no remaining open questions/issues.

src/lib/operators/aggregate_hash.hpp
```diff
@@ -19,14 +19,15 @@
...
 #include "resolve_type.hpp"
-#include "storage/abstract_segment_visitor.hpp"
 #include "storage/reference_segment.hpp"
 #include "storage/value_segment.hpp"
 #include "types.hpp"
 #include "utils/assert.hpp"
 
 namespace opossum {
 
+class SegmentVisitorContext {};
```
mrks: Is this a forward declaration? Those usually come without curly braces.

src/lib/operators/export_binary.cpp
```diff
@@ -171,90 +149,70 @@ void ExportBinary::_write_header(const Table& table, std::ofstream& ofstream) {
...
+  export_value(ofstream, EncodingType::Unencoded);
 
-  // Unfortunately, we have to iterate over all values of the reference segment
-  // to materialize its contents. Then we can write them to the file
-  for (ChunkOffset row = 0; row < ref_segment.size(); ++row) {
-    export_value(context->ofstream, boost::get<T>(ref_segment[row]));
-  }
-}
+  if (ref_segment.data_type() == DataType::String) {
+    if (ref_segment.size() == 0) return;
```
mrks: This guard clause is not unique to string types, is it? I'd move it outside.

mrks: Not your code, but unless we have a good reason not to, we should write out `reference_segment`

src/lib/operators/export_binary.cpp
```diff
@@ -171,90 +149,70 @@ void ExportBinary::_write_header(const Table& table, std::ofstream& ofstream) {
...
+    pmr_vector<size_t> string_lengths(ref_segment.size());
 
-  // We materialize reference segments and save them as value segments
-  export_value(context->ofstream, BinarySegmentType::value_segment);
-
-  // If there is no data, we can skip all of the coming steps.
-  if (ref_segment.size() == 0) return;
+    // We export the values materialized
+    for (ChunkOffset row = 0; row < ref_segment.size(); ++row) {
+      value = boost::get<pmr_string>(ref_segment[row]);
```
mrks: No need to change this as we are not yet looking at the performance of the export, but `operator[]` is quite slow because it converts everything into an AllTypeVariant. For new code, have a look at `segment_accessor.hpp`.

src/lib/operators/export_binary.hpp
```diff
@@ -73,7 +73,7 @@ class ExportBinary : public AbstractReadOnlyOperator {
    * Row count             | ChunkOffset                           |  4
    *
    * Next, it dumps the contents of the segments in the respective format (depending on the type
-   * of the segment, such as ReferenceSegment, DictionarySegment, ValueSegment).
+   * of the segment, such as ReferenceSegment, DictionarySegment, ValueSegment, RunLengthSegment).
```
mrks: I wouldn't expect the "such as" list to be complete - otherwise it would be also missing the FixedStringDictionarySegment.

src/lib/operators/export_binary.hpp
```diff
@@ -161,13 +152,29 @@ class ExportBinary::ExportBinaryVisitor : public AbstractSegmentVisitor {
...
+   * Values                 | T (int, float, double, long)          |   size * sizeof(T)
+   * NULL values            | vector<bool> (BoolAsByteType)         |   size * 1
+   * End Positions          | ChunkOffset                           |   size * 4
+   *
+   * Please note that the number of rows are written in the header of the chunk.
+   * The type of the column can be found in the global header of the file.
+   *
+   *
+   * @param base_segment The segment to export
+   * @param base_context A context in the form of an ExportContext. Contains a reference to the ofstream.
```
mrks: outdated

src/test/operators/export_binary_test.cpp
```diff
@@ -52,12 +52,30 @@ class OperatorsExportBinaryTest : public BaseTest {
   const std::string filename = test_data_path + "export_test.bin";
 };
 
+class OperatorsExportBinaryMultiEncodingTest : public OperatorsExportBinaryTest,
```
mrks: Haven't looked at the tests in detail yet. Now that we have a parameterized test, do we still need the old tests?

mrks: Just some cases that we should make sure to cover:
- single chunk
- multiple chunks
- All-NULL
- Repeated values (for RLE)
- Sorted / unsorted values

src/test/operators/export_binary_test.cpp
```diff
@@ -67,158 +85,157 @@ TEST_F(OperatorsExportBinaryTest, TwoColumnsNoValues) {
...
-  ChunkEncoder::encode_all_chunks(table, EncodingType::FixedStringDictionary);
+  ChunkEncoder::encode_all_chunks(table, GetParam());
 
   auto table_wrapper = std::make_shared<TableWrapper>(std::move(table));
   table_wrapper->execute();
   auto ex = std::make_shared<opossum::ExportBinary>(table_wrapper, filename);
   ex->execute();
 
+  std::map<EncodingType, std::string> reference_filenames{
+      {EncodingType::Unencoded, "resources/test_data/bin/SingleChunkSingleFloatColumnValueSegment.bin"},
```
mrks: Should be possible to make this generic with `encoding_type_to_string`

src/lib/operators/aggregate_hash.hpp
mrks: Or is this because the aggregate is the only one who still needs the visitor? In that case, maybe add a short comment

src/lib/operators/export_binary.hpp
```diff
@@ -161,13 +152,29 @@ class ExportBinary::ExportBinaryVisitor : public AbstractSegmentVisitor {
...
+  /**
+   * RunLength Segments are dumped with the following layout:
+   *
+   * Description            | Type                                  | Size in bytes
+   * -----------------------------------------------------------------------------------------
+   * Column Type            | ColumnType                            |   1
+   * Size                   | uint32_t                              |   4
+   * Values                 | T (int, float, double, long)          |   size * sizeof(T)
+   * NULL values            | vector<bool> (BoolAsByteType)         |   size * 1
+   * End Positions          | ChunkOffset                           |   size * 4
```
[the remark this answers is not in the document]
Bouncner: Run count?

src/test/operators/export_binary_test.cpp
j-tr: the old (OperatorsExportBinaryTests) tests are still used for tests that work with just one encoding type (see FixedStringDictionarySegment or RepeatedIntRunLengthSegment)

mrks: Got it. However, we should cover all cases (e.g., repeated ints) as part of the parameterized tests, too.

j-tr:
might fix #1920

src/lib/operators/export_binary.cpp
```diff
@@ -171,95 +149,75 @@ void ExportBinary::_write_header(const Table& table, std::ofstream& ofstream) {
...
-  Assert(base_segment.compressed_vector_type(),
+void ExportBinary::_write_segment(const BaseDictionarySegment& base_dictionary_segment, std::ofstream& ofstream) {
+  Assert(base_dictionary_segment.compressed_vector_type(),
          "Expected DictionarySegment to use vector compression for attribute vector");
-  Assert(is_fixed_size_byte_aligned(*base_segment.compressed_vector_type()),
+  Assert(is_fixed_size_byte_aligned(*base_dictionary_segment.compressed_vector_type()),
          "Does only support fixed-size byte-aligned compressed attribute vectors.");
-
-  export_value(context->ofstream, BinarySegmentType::dictionary_segment);
+  export_value(ofstream, EncodingType::Dictionary);
```
mrks: This means that we convert FixedString dictionaries to regular dictionaries. That is probably surprising for the caller. As we don't have an import for the former yet, I'd change the support from BaseDictionarySegment to DictionarySegment.

mrks:
> might fix #1920

Add to hyriseConsole_test.py?

src/benchmark/tpch_benchmark.cpp
```diff
@@ -94,7 +94,9 @@ int main(int argc, char* argv[]) {
   std::cout << "- Benchmarking Queries: [ ";
   for (const auto& item_id : item_ids) {
     std::cout << (item_id + 1);
-    if (item_id != item_ids.back()) { std::cout << ","; }
+    if (item_id != item_ids.back()) {
+      std::cout << ",";
```
Bouncner: Can't we append `", "` and then `std::cout << " ]" << std::endl;` (if we really want these nasty spaces next to the brackets)?

mrks: Did you mean to comment on this PR?

Bouncner: Yep ... it's not really their change, but if they are going to change for formatting anyways ...

mrks: I feel like we have already added more than enough to this PR

src/test/operators/export_binary_test.cpp
j-tr: Since RLE segments are usually created from value segments and value segments guarantee that stored values for NULL-attributed values are default-constructed `T{}`, it's rather complicated to construct such a table as in the example. I did so by directly assigning the values to the value vector of the segment, since all other access methods prevent the value vector from containing actual values if the null value property is set. Consequently someone must be very malicious to construct such a table. But in case it happens, it's tested now.

mrks: Doesn't that happen with the current encoder? To me, it looks like for NULL values, it continues the current run without looking at the value in the input segment:

https://github.com/hyrise/hyrise/blob/9d274347689191dc5faa619800c4475b7ce7c83e/src/lib/storage/run_length_segment/run_length_encoder.hpp#L47-L53

mrks: Ah, no, I misunderstood that code. In any case, that is what should happen. I'll open an issue.

mrks: Ok, I was entirely mistaken, likely confused by this code in run_length_segment.hpp:

```
As in value segments, null values are represented as an additional boolean vector.
```

I would remove that sentence and change a sentence a couple of lines above (including a typo fix):

```
Each run is represented by the value and the number of **occurrences. Runs can also be NULL, in which case `values` is undefined (but likely `T{}`) and `null_values` is set to true for that run.**

**Comment by mrks** [comment]
Sorry for all the comments coming in on this that are unrelated to the actual purpose of the PR. This touches code that many of us have not looked at for a while, so we spot things that should have been caught earlier. Had I known how many of those things we would find, I would have moved all of these to a different issue.

File: src/test/operators/export_binary_test.cpp (hunk 12 above)
Comment (mrks) [reply src/test/operators/export_binary_test.cpp]: Moving this to #1928 so that we can find an end here.

**Comment by Bouncner** [comment]
Ack, go for it.

**Comment by mrks** [comment]
@j-tr @dey4ss ?

File: src/test/operators/export_binary_test.cpp [hunk 17]
```diff
@@ -8,11 +8,11 @@
 #include "base_test.hpp"
 #include "gtest/gtest.h"
 
-#include "import_export/binary.hpp"
 #include "operators/export_binary.hpp"
 #include "operators/table_scan.hpp"
```
tobodner: Not used anymore. So is <vector>.

scripts/test/hyriseConsole_test.py
```diff
@@ -38,6 +38,10 @@ def main():
 	console.expect('Loading .*tbl/10_ints.tbl into table "test"')
 	console.expect('Encoding "test" using Unencoded')
 
+	console.sendline("load resources/test_data/bin/float.bin test_bin")
```
tobodner: Not exactly part of this PR, but related: I am wondering, whether this command should really be named import to mirror the export command. load is - internally - only considered for testing via .tbl files.

Bouncner: Ack.

Probably open a ticket? Now with the new server (which lacks means to load data), we should consider implement `COPY TABLE x FROM file`.

Bouncner:
@tobodner : do you zünd the knaller once your points are addressed?

tobodner:
> @tobodner : do you zünd the knaller once your points are addressed?

Will do. Any objections anyone?

scripts/test/hyriseConsole_test.py
tobodner: As you suggested, I opened a separate ticket: https://github.com/hyrise/hyrise/issues/1933

scripts/test/hyriseConsole_test.py
```diff
@@ -38,6 +38,10 @@ def main():
...
+	console.expect('Loading .*bin/float.bin into table "test_bin"')
+	console.expect('Encoding "test_bin" using Unencoded')
```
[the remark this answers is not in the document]
tobodner: Not part of this PR. May be fixed at a later point in time.

src/test/operators/export_binary_test.cpp
tobodner: I can do a separate cleanup pass here.

PR #1907, opened 2019-11-02
Fix LQP outputs

mrks:
With both visit_lqp_upwards and the new ColumnPruningRule, we depend on the output links of AbstractLQPNodes to be correct. We now make sure that the optimizer has exclusive ownership so that no remnants of the old plan remain.

This also adds a verify_lqp method to the optimizer (which we should have had a long time ago). While it is not perfect, it should help us catch issues as soon as a rule breaks something, not just once a following rule or the LQPTranslator stumbles upon the issue.

TODO
- [x] Check impact on CI runtime
- [x] Check impact on benchmark results
- [x] Tests for Optimizer::verify_lqp

### DIFF
--- a/src/lib/logical_query_plan/lqp_column_reference.cpp
+++ b/src/lib/logical_query_plan/lqp_column_reference.cpp
@@ -26,10 +26,10 @@ std::ostream& operator<<(std::ostream& os, const LQPColumnReference& column_refe
   const auto original_node = column_reference.original_node();
   Assert(original_node, "OriginalNode has expired");
 
-  const auto stored_table_node = std::static_pointer_cast<const StoredTableNode>(column_reference.original_node());
+  const auto stored_table_node = std::dynamic_pointer_cast<const StoredTableNode>(column_reference.original_node());
   const auto table = Hyrise::get().storage_manager.get_table(stored_table_node->table_name);
   Assert(column_reference.original_column_id() != INVALID_COLUMN_ID,
-    "Tried to print and uninitialized column or COUNT(*)");
+         "Tried to print an uninitialized column or COUNT(*)");
   os << table->column_name(column_reference.original_column_id());
 
   return os;
--- a/src/lib/logical_query_plan/lqp_column_reference.cpp
+++ b/src/lib/logical_query_plan/lqp_column_reference.cpp
@@ -4,6 +4,8 @@
 
 #include "abstract_lqp_node.hpp"
 #include "hyrise.hpp"
+#include "logical_query_plan/mock_node.hpp"
+#include "logical_query_plan/static_table_node.hpp"
 #include "logical_query_plan/stored_table_node.hpp"
 #include "storage/table.hpp"
 #include "utils/assert.hpp"
@@ -25,12 +27,27 @@ bool LQPColumnReference::operator==(const LQPColumnReference& rhs) const {
 std::ostream& operator<<(std::ostream& os, const LQPColumnReference& column_reference) {
   const auto original_node = column_reference.original_node();
   Assert(original_node, "OriginalNode has expired");
-
-  const auto stored_table_node = std::dynamic_pointer_cast<const StoredTableNode>(column_reference.original_node());
-  const auto table = Hyrise::get().storage_manager.get_table(stored_table_node->table_name);
   Assert(column_reference.original_column_id() != INVALID_COLUMN_ID,
          "Tried to print an uninitialized column or COUNT(*)");
-  os << table->column_name(column_reference.original_column_id());
+
+  switch (original_node->type) {
+    case LQPNodeType::StoredTable: {
+      const auto stored_table_node = std::static_pointer_cast<const StoredTableNode>(column_reference.original_node());
+      const auto table = Hyrise::get().storage_manager.get_table(stored_table_node->table_name);
+      os << table->column_name(column_reference.original_column_id());
+    } break;
+    case LQPNodeType::Mock: {
+      const auto mock_node = std::static_pointer_cast<const MockNode>(column_reference.original_node());
+      os << mock_node->column_definitions().at(column_reference.original_column_id()).second;
+    } break;
+    case LQPNodeType::StaticTable: {
+      const auto static_table_node = std::static_pointer_cast<const StaticTableNode>(column_reference.original_node());
+      const auto& table = static_table_node->table;
+      os << table->column_name(column_reference.original_column_id());
+    } break;
+    default:
+      Fail("Unexpected original_node for LQPColumnReference");
+  }
 
   return os;
 }
--- a/src/lib/logical_query_plan/lqp_utils.cpp
+++ b/src/lib/logical_query_plan/lqp_utils.cpp
@@ -296,31 +296,4 @@ std::vector<std::shared_ptr<AbstractLQPNode>> lqp_find_subplan_roots(const std::
   return root_nodes;
 }
 
-void collect_subquery_expressions_by_lqp(SubqueryExpressionsByLQP& subquery_expressions_by_lqp,
-                                         const std::shared_ptr<AbstractLQPNode>& node,
-                                         std::unordered_set<std::shared_ptr<AbstractLQPNode>>& visited_nodes) {
-  if (!node) return;
-  if (!visited_nodes.emplace(node).second) return;
-
-  for (const auto& expression : node->node_expressions) {
-    visit_expression(expression, [&](const auto& sub_expression) {
-      const auto subquery_expression = std::dynamic_pointer_cast<LQPSubqueryExpression>(sub_expression);
-      if (!subquery_expression) return ExpressionVisitation::VisitArguments;
-
-      for (auto& [lqp, subquery_expressions] : subquery_expressions_by_lqp) {
-        if (*lqp == *subquery_expression->lqp) {
-          subquery_expressions.emplace_back(subquery_expression);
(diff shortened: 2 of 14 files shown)


src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -296,4 +296,31 @@ std::vector<std::shared_ptr<AbstractLQPNode>> lqp_find_subplan_roots(const std::
   return root_nodes;
 }
 
+void collect_subquery_expressions_by_lqp(SubqueryExpressionsByLQP& subquery_expressions_by_lqp,
```
mrks: Moved here 1:1 from optimizer.cpp

src/lib/optimizer/optimizer.cpp
```diff
@@ -148,6 +110,16 @@ std::shared_ptr<AbstractLQPNode> Optimizer::optimize(const std::shared_ptr<Abstr
...
   for (const auto& rule : _rules) {
     _apply_rule(*rule, root_node);
+
+    // The SubqueryToJoinRule was written before we required all outputs of a node to be part of the same plan
+    // (see validate_lqp). Other than that, the rule works correctly and refactoring it only for this additional
+    // requirements is not worth the effort. Thus, we manually remove all outputs that become unused due to the
+    // rewrite. We trigger this here so that it is only called on the root node, not on any subquery LQPs.
+    if (const auto subquery_to_join_rule = dynamic_cast<SubqueryToJoinRule*>(&*rule)) {
+      subquery_to_join_rule->remove_unused_outputs(root_node);
+    }
```
mrks: Really unhappy about this part

mrzzzrm:
The LQP (both in the input setters and the node destructor) unlinks itself, iirc? Under which circumstances does this not work?

mrks:
Oh, good thing that you asked. I though I remembered that we decided against LQP nodes automatically untying themselves after destruction. That, however, does happen.

While nodes from within the LQP get correctly destructed, the caller might hold the original root node, which might not even be part of the LQP anymore.

Fixing that is MUCH cleaner than what I had before.

src/lib/logical_query_plan/lqp_column_reference.cpp
```diff
@@ -26,10 +26,10 @@ std::ostream& operator<<(std::ostream& os, const LQPColumnReference& column_refe
   const auto original_node = column_reference.original_node();
   Assert(original_node, "OriginalNode has expired");
 
-  const auto stored_table_node = std::static_pointer_cast<const StoredTableNode>(column_reference.original_node());
+  const auto stored_table_node = std::dynamic_pointer_cast<const StoredTableNode>(column_reference.original_node());
```
Bensk1: Why don't we need to check the result of the cast for a `nulltpr`?

src/lib/optimizer/optimizer.cpp
```diff
@@ -141,13 +141,23 @@ void Optimizer::add_rule(std::unique_ptr<AbstractRule> rule) {
   _rules.emplace_back(std::move(rule));
 }
 
-std::shared_ptr<AbstractLQPNode> Optimizer::optimize(const std::shared_ptr<AbstractLQPNode>& input) const {
+std::shared_ptr<AbstractLQPNode> Optimizer::optimize(std::shared_ptr<AbstractLQPNode> input) const {
+  // We cannot allow multiple owners of the LQP as one owner could decide to optimize the plan and others might hold a
+  // pointer to a node that is not even part of the plan anymore after optimization. Thus, callers of this method need
+  // to relinquish their ownership (i.e., move their shared_ptr into the method) and take ownership of the resulting
+  // optimized plan.
+  Assert(input.use_count() == 1, "Optimizer should have exclusive ownership of plan");
```
Bensk1: Hm, are there any other ways than checking the use_count of the shared pointer to achieve this? This does not seem very clean.

src/lib/optimizer/optimizer.cpp
```diff
@@ -157,6 +167,100 @@ std::shared_ptr<AbstractLQPNode> Optimizer::optimize(const std::shared_ptr<Abstr
...
+  // (1) Make sure that all outputs found anywhere in the entire plan are also part of the plan
+  // (2) Make sure each node has the number of inputs expected for that node type
+  // (3) Make sure that for all LQPColumnExpressions, the original_node is part of the LQP
+  for (const auto& lqp : lqps) {
+    visit_lqp(lqp, [&all_nodes](const auto& node) {
+      // Check that all outputs are part of the LQP
+      const auto outputs = node->outputs();
+      for (const auto& output : outputs) {
+        // LQP nodes should only be part of a single LQP and all of their outputs should be part of the same LQP. If you
+        // run into this assertion as part of a new test, there is a good chance that you are using part of the LQP
```
Bensk1: partS or a part?

src/lib/optimizer/optimizer.cpp
```diff
@@ -157,6 +167,100 @@ std::shared_ptr<AbstractLQPNode> Optimizer::optimize(const std::shared_ptr<Abstr
...
+        // outside of the LQP that is currently optimized. For example, if you create a bunch of LQP nodes in the test's
+        // SetUp method but these are not used in the current LQP, this can cause this assertion to fail. This is only
+        // enforced when rules are executed through the optimizer, not tests call the rule directly.
```
Bensk1: not when tests

src/lib/sql/sql_pipeline.cpp
```diff
@@ -145,16 +145,19 @@ const std::vector<std::shared_ptr<AbstractLQPNode>>& SQLPipeline::get_optimized_
          "One or more SQL statement is dependent on the execution of a previous one. "
          "Cannot translate all statements without executing, i.e. calling get_result_table()");
 
+  // The optimizer works on the original unoptimized LQP nodes. After optimizing, the unoptimized version is also
```
Bensk1: I think I understand what you want to say but I find it a bit confusing. Maybe it gets clearer if we just say that the unoptimized version is altered during the optimization process. 
A different question, why do we have to modify the unoptimized version at all, can't we just create a copy if we have `_optimized_logical_plans` anyways?

src/test/optimizer/optimizer_test.cpp
```diff
@@ -43,6 +43,90 @@ class OptimizerTest : public ::testing::Test {
...
+      if (!projection_node) return;
+      projection_node->node_expressions[0] = out_of_plan_expression;
+    }
+
+    std::shared_ptr<AbstractExpression> out_of_plan_expression;
+  };
+
+  optimizer.add_rule(std::make_unique<LQPBreakingRule>(lqp_column_(x)));
+
+  // Does not move the LQP into the optimizer
```
Bensk1: Comment copy pasta?

src/test/optimizer/optimizer_test.cpp
```diff
@@ -142,7 +228,8 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
   Optimizer optimizer{};
   optimizer.add_rule(std::move(rule));
 
-  optimizer.optimize(lqp);
+  const auto optimized_lqp = optimizer.optimize(std::move(lqp));
+  lqp = nullptr;
```
Bensk1: Why do we specifically set `lqp = nullptr` here and don't do that above?

src/lib/optimizer/optimizer.cpp
mrks: Could you try to explain your concerns?

src/lib/sql/sql_pipeline.cpp
mrks:
```
// The optimizer modifies the input LQP and requires exclusive ownership of that LQP. This means that we need to clear
// _unoptimized_logical_plans. This is not an issue as the unoptimized plans will no longer be needed. Note that this
// does not clear the unoptimized LQPs stored in the SQLPipelineStatement - those are cleared as part of
// SQLPipelineStatement::get_optimized_logical_plan. Calls to get_unoptimized_logical_plans are still allowed (e.g.,
// for visualization), in which case the unoptimized plan will be recreated.
```

better?

mrks: If yes, I'll adapt the version in sql_pipeline_statement.cpp accordingly

src/test/optimizer/optimizer_test.cpp
mrks: Mmh, no real reason other than that there were more than a couple of lines below. In the end, both should work, so I would just leave both versions.

src/lib/logical_query_plan/lqp_column_reference.cpp
mrks: I was surprised to see only StoredTableNode here. For testing, I changed it to check for segfaults. In regular execution, we do not print MockNode expressions, but I fixed it anyway.

src/lib/optimizer/optimizer.cpp
Bensk1: As discussed offline

mrks: Did we reach a conclusion?

src/lib/sql/sql_pipeline.cpp
Bensk1: Yes

src/lib/optimizer/optimizer.cpp
Bensk1: It seems to be fine.

PR #1909, opened 2019-11-05
Fix multi-threading in GetTable, add conflict test

### LINKED ISSUE #1908: Server: Test for transaction conflicts
I tried to add a test for the transaction conflict error code:

```
TEST_F(ServerTestRunner, TestTransactionConflicts) {
  // Similar to TestParallelConnections, but this time we modify the table
  const std::string sql = "UPDATE table_a SET b = b + 1 WHERE b = (SELECT MIN(b) FROM table_a);";

  const auto connection_run = [&]() {
    pqxx::connection connection{_connection_string};
    pqxx::nontransaction transaction{connection};
    const auto result = transaction.exec(sql);
  };

  const auto num_threads = 100u;
  std::vector<std::future<void>> thread_futures;
  thread_futures.reserve(num_threads);

  for (auto thread_num = 0u; thread_num < num_threads; ++thread_num) {
    // We want a future to the thread running, so we can kill it after a future.wait(timeout) or the test would freeze
    thread_futures.emplace_back(std::async(std::launch::async, connection_run));
  }

  for (auto& thread_future : thread_futures) {
    // We give this a lot of time, not because we need that long for 100 threads to finish, but because sanitizers and
    // other tools like valgrind sometimes bring a high overhead that exceeds 10 seconds.
    if (thread_future.wait_for(std::chrono::seconds(150)) == std::future_status::timeout) {
      ASSERT_TRUE(false) << "At least one thread got stuck and did not commit.";
    }
  }

  auto pipeline =
      SQLPipelineBuilder{std::string{"SELECT MIN(b) FROM table_a"}}
          .create_pipeline();
  const auto [_, table] = pipeline.get_result_table();
  std::cout << table->get_value<float>(ColumnID{0}, 0) << std::endl;
}
```

As the cout in the last line shows, we have transaction conflicts. I would have expected pqxx to throw an exception that I could use to distinguish between transaction conflicts and actual errors.

mrks:
- [x] Check remaining concurrent_vectors
- [x] Add insert test
- [x] Doc the concurrent_vector thing somewhere. But where?
- [x] check performance impact, mostly because of GetTable - no discernible difference

When accessing a table from the StorageManager, the chunk count might change somewhere in between. As the GetTable operator is the only operator that directly accesses the stored table, this (hopefully) does not happen anywhere else.

Also adds a similar test for the server, fixes #1908 

ST (with jemalloc fix):
```diff
 +----------------+----------------+-------+----------------+-------+------------+----------------------+
 | Benchmark      | prev. iter/s   | runs  | new iter/s     | runs  | change [%] |              p-value |
 +----------------+----------------+-------+----------------+-------+------------+----------------------+
 | TPC-H 01       | 0.601657509804 | 37    | 0.605851709843 | 37    | +1%        |               0.3627 |
 | TPC-H 02       | 11.9922161102  | 720   | 11.7451496124  | 705   | -2%        |               0.0000 |
 | TPC-H 03       | 6.63913393021  | 399   | 6.48139810562  | 389   | -2%        |               0.0000 |
 | TPC-H 04       | 6.91854810715  | 416   | 7.01023054123  | 421   | +1%        |               0.0047 |
 | TPC-H 05       | 4.2410030365   | 255   | 4.2090473175   | 253   | -1%        |               0.2747 |
 | TPC-H 06       | 201.229888916  | 10000 | 196.862518311  | 10000 | -2%        | (run time too short) |
 | TPC-H 07       | 7.82547521591  | 470   | 7.67237234116  | 461   | -2%        |               0.0000 |
 | TPC-H 08       | 6.52758312225  | 392   | 6.3308596611   | 380   | -3%        |               0.0000 |
 | TPC-H 09       | 1.5080062151   | 91    | 1.48862397671  | 90    | -1%        |               0.0008 |
 | TPC-H 10       | 2.90244078636  | 175   | 2.89184975624  | 174   | -0%        |               0.6659 |
 | TPC-H 11       | 28.5252933502  | 1712  | 28.0677642822  | 1685  | -2%        |               0.0000 |
 | TPC-H 12       | 8.98255348206  | 539   | 8.70111370087  | 523   | -3%        |               0.0000 |
 | TPC-H 13       | 2.1892645359   | 132   | 2.2159705162   | 133   | +1%        |               0.0003 |
 | TPC-H 14       | 47.0348320007  | 2823  | 46.9707260132  | 2819  | -0%        |               0.1711 |
 | TPC-H 15       | 76.2191238403  | 4574  | 76.6935501099  | 4602  | +1%        |               0.0000 |
 | TPC-H 16       | 7.04813575745  | 423   | 6.84563112259  | 411   | -3%        |               0.0000 |
 | TPC-H 17       | 1.52463304996  | 92    | 1.54568088055  | 93    | +1%        |               0.0189 |
 | TPC-H 18       | 0.773465931416 | 47    | 0.786594450474 | 48    | +2%        |               0.0631 |
-| TPC-H 19       | 35.2544746399  | 2116  | 33.2077941895  | 1993  | -6%        |               0.0000 |
 | TPC-H 20       | 2.74802684784  | 165   | 2.80631637573  | 169   | +2%        |               0.0000 |
 | TPC-H 21       | 1.42818105221  | 86    | 1.41221356392  | 85    | -1%        |               0.0022 |
-| TPC-H 22       | 14.4316520691  | 866   | 13.3441267014  | 801   | -8%        |               0.0000 |
 | geometric mean |                |       |                |       | -1%        |                      |
 +----------------+----------------+-------+----------------+-------+------------+----------------------+
```

MT (with jemalloc fix):
```diff
 +----------------+---------------+-------+---------------+-------+------------+---------+
 | Benchmark      | prev. iter/s  | runs  | new iter/s    | runs  | change [%] | p-value |
 +----------------+---------------+-------+---------------+-------+------------+---------+
 | TPC-H 01       | 12.7563943863 | 3827  | 12.8199882507 | 3846  | +0%        |  0.4559 |
 | TPC-H 02       | 286.546386719 | 10002 | 295.930328369 | 10002 | +3%        |  0.0011 |
 | TPC-H 03       | 126.921340942 | 10000 | 127.285835266 | 10000 | +0%        |  0.9459 |
 | TPC-H 04       | 36.5649261475 | 10000 | 35.3371200562 | 10001 | -3%        |  0.0296 |
 | TPC-H 05       | 75.5186538696 | 10004 | 75.6515426636 | 10000 | +0%        |  0.9870 |
 | TPC-H 06       | 3585.16992188 | 10068 | 3548.56567383 | 10073 | -1%        |  0.9688 |
 | TPC-H 07       | 171.930496216 | 10000 | 170.033584595 | 10001 | -1%        |  0.3120 |
 | TPC-H 08       | 146.09022522  | 10001 | 143.777389526 | 10001 | -2%        |  0.3562 |
 | TPC-H 09       | 34.8444328308 | 10000 | 34.5305366516 | 10000 | -1%        |  0.4077 |
 | TPC-H 10       | 73.7197647095 | 10001 | 73.7737426758 | 10000 | +0%        |  0.8032 |
 | TPC-H 11       | 557.274841309 | 10001 | 564.361022949 | 10005 | +1%        |  0.4040 |
 | TPC-H 12       | 366.447784424 | 10002 | 363.94744873  | 10000 | -1%        |  0.8001 |
 | TPC-H 13       | 55.4529914856 | 10000 | 54.4465141296 | 10000 | -2%        |  0.0705 |
 | TPC-H 14       | 923.971923828 | 10012 | 927.079711914 | 10001 | +0%        |  0.6661 |
 | TPC-H 15       | 1812.96533203 | 10004 | 1832.65856934 | 10000 | +1%        |  0.0842 |
 | TPC-H 16       | 210.811843872 | 10000 | 209.790924072 | 10002 | -0%        |  0.7594 |
 | TPC-H 17       | 25.90275383   | 7771  | 25.9124851227 | 7774  | +0%        |  0.8528 |
 | TPC-H 18       | 14.573302269  | 4372  | 14.7099609375 | 4413  | +1%        |  0.3200 |
 | TPC-H 19       | 613.185974121 | 10000 | 606.499633789 | 10033 | -1%        |  0.1398 |
 | TPC-H 20       | 53.5546226501 | 10000 | 53.1940002441 | 10000 | -1%        |  0.5046 |
 | TPC-H 21       | 23.413192749  | 7024  | 23.1194782257 | 6936  | -1%        |  0.1971 |
 | TPC-H 22       | 211.096969604 | 10003 | 211.436447144 | 10001 | +0%        |  0.8446 |
 | geometric mean |               |       |               |       | -0%        |         |
 +----------------+---------------+-------+---------------+-------+------------+---------+
```

No idea why this got faster. :/


Bouncner:
How long does that stress test run approx?

mrks:
Less than 1 second

tobodner:
> Less than 1 second
Confirmed. On both nemea and a MacBook.

mrks:
If you haven't started reviewing yet, I'll add the server test in a couple of minutes.

resources/.tsan-ignore.txt
```diff
@@ -12,12 +12,18 @@ race:^std::__1::unique_ptr
...
 # Boost is not built with tsan support, either
 race:boost::detail::thread_data_base
 race:thread_data
 
 # A DebugAssert makes sure that predecessors are not added to scheduled tasks, but it is missing in release builds
 race:^opossum::AbstractTask::set_as_predecessor_of
 
+# tsan does not understand parts of our MVCC concept - see mvcc_data.hpp for documentation
+race:&opossum::MvccData::set_begin_cid
+race:&opossum::MvccData::set_end_cid
```
mrks: @julianmenzler - this should be of interest for #1493

mrks:
Looks we still have some things to look at. @tobodner - feel free to have a look at what we already have, but I'll still need to add something. If only I knew what :D Of course, if you want to, you can try and track down the cause of that failure, too.

resources/.tsan-ignore.txt
tobodner: I have not used tsan before, but '&' does not seem to be a valid token: https://github.com/google/sanitizers/wiki/ThreadSanitizerSuppressions
If I locally replace it with '^', the listed methods get actually filtered.

tobodner:
> Looks we still have some things to look at. @tobodner - feel free to have a look at what we already have, but I'll still need to add something. If only I knew what :D Of course, if you want to, you can try and track down the cause of that failure, too.

I am not sure, whether I am running tsan correctly, but I got more races. Here three log files from example runs.

[log_file_1.txt](https://github.com/hyrise/hyrise/files/3811425/log_file_1.txt)
[log_file_2.txt](https://github.com/hyrise/hyrise/files/3811426/log_file_2.txt)
[log_file_3.txt](https://github.com/hyrise/hyrise/files/3811427/log_file_3.txt)

mrks:
Which machine is than on? `std::__1::__atomic_base` looks like it comes from std not being instrumented.

resources/.tsan-ignore.txt
mrks: But the keys are right next to each other. That should count for something, right?

tobodner:
> Which machine is than on? `std::__1::__atomic_base` looks like it comes from std not being instrumented.

This is from my MacBook with macOS 10.15.1/clang-9.

mrks:
Weird. I can't even find anyone else who accesses `_started`. Looks like another false positive :/

tobodner:
> > Which machine is than on? `std::__1::__atomic_base` looks like it comes from std not being instrumented.
> 
> This is from my MacBook with macOS 10.15.1/clang-9.

Please note that there are also others, in AbstractLQPNode and Optimizer (I think).

tobodner:
> Looks we still have some things to look at. @tobodner - feel free to have a look at what we already have, but I'll still need to add something. If only I knew what :D Of course, if you want to, you can try and track down the cause of that failure, too.

OK, I will look into this as well.

resources/.tsan-ignore.txt
```diff
@@ -12,12 +12,18 @@ race:^std::__1::unique_ptr
 race:^std::__1::__unique_if
 race:shared_count
 
+race:^std::atomic
```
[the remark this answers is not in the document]
julianmenzler: I also stumbled across the following [TSan flags](https://github.com/google/sanitizers/wiki/ThreadSanitizerFlags): 
* `force_seq_cst_atomics` (runtime)
* `-mllvm -tsan-instrument-atomics` (compiler)

Useful to us?

src/test/concurrency/stress_test.cpp
```diff
@@ -0,0 +1,85 @@
...
+  const std::string sql = "UPDATE table_a SET a = a + 1 WHERE a = (SELECT MIN(a) FROM table_a);";
+
+  std::atomic_int successful_increments{0};
+  std::atomic_int conflicted_increments{0};
+  const auto iterations_per_thread = 10;
+
+  const auto run = [&]() {
+    for (auto iteration = 0; iteration < iterations_per_thread; ++iteration) {
+      auto pipeline =
+          SQLPipelineBuilder{std::string{"UPDATE table_a SET a = a + 1 WHERE a = (SELECT MIN(a) FROM table_a);"}}
```
mrks: use sql from above

resources/.tsan-ignore.txt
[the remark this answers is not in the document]
mrks: Those are probably Mac false positives; let's discuss that in #1913

src/lib/storage/mvcc_data.hpp
```diff
@@ -41,6 +41,15 @@ struct MvccData {
...
   void grow_by(size_t delta, TransactionID transaction_id, CommitID begin_commit_id);
 
+  /**
+   * The thread sanitzer (tsan) complains about concurrent writes and reads to begin/end_cids. That is because it is
+   * unaware of their thread-safety being guaranteed by the update of the global last_cid. Furthermore, we exploit that
+   * writes up to eight bytes are atomic on x64, which C++ and tsan do not know about. These helper methods were added
+   * to .tsan-ignore.txt and can be used (carefully) to avoid those false positives.
+   */
+  void set_begin_cid(const ChunkOffset offset, const CommitID commit_id);
+  void set_end_cid(const ChunkOffset offset, const CommitID commit_id);
```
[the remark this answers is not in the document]
mrks: I have given up on trying to understand which calls have costs and which calls don't. I'm not a fan of inlining things without looking at the profiling data first. Let's wait for that to pop up.

resources/.tsan-ignore.txt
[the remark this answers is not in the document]
mrks: You were talking about Mac, weren't you?

mrks:
Fixed another issue with GetTable returning chunks that did not have any rows

mrks:
<pre>
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `./hyriseSystemTest --gtest_filter=*Stress* --gtest_repeat=-1 --gtest_break_on_f'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000841f0f in ?? ()
[Current thread is 1 (Thread 0x7fe18cfff700 (LWP 156830))]
(gdb) bt
#0  0x0000000000841f0f in ?? ()
#1  0x0000558e9fbd5277 <b>in opossum::Chunk::size()</b> const ()
</pre>

No bueno.

mrks:
https://software.intel.com/en-us/blogs/2009/04/09/delusion-of-tbbconcurrent_vectors-size-or-3-ways-to-traverse-in-parallel-correctly

Adding multiple threads to computers was a mistake...

mrks:
What a mess. I'd like more than one review on this, please.

mrks:
The key multi-threading fixes are in get_table.cpp and table.cpp. I'd probably start reviewing there.

src/lib/operators/get_table.hpp
```diff
@@ -10,7 +10,14 @@
 
 namespace opossum {
 
-// operator to retrieve a table from the StorageManager by specifying its name
+// Operator to retrieve a table from the StorageManager by specifying its name. Depending on how the operator was
+// constructed, chunks and columns may be pruned if they are irrelevant for the final result. The returned table is NOT
+// the same table as stored in the StorageManager. If that stored table is changed (most importantly: if a chunk is
+// added), this is not reflected in GetTable's result. This is by design to make sure than following operators do not
```
Bensk1: make sure thaT

src/lib/storage/mvcc_data.hpp
```diff
@@ -18,6 +18,13 @@ struct MvccData {
   // The last commit id is reserved for uncommitted changes
   static constexpr CommitID MAX_COMMIT_ID = std::numeric_limits<CommitID>::max() - 1;
 
+
+  // Entries that have just been appended might be uninitialized (i.e., have a random value):
+  // https://software.intel.com/en-us/blogs/2009/04/09/delusion-of-tbbconcurrent_vectors-size-or-3-ways-to-traverse-in-parallel-correctly  // NOLINT
+  // However, they are not accessed by any other transaction as long as only the MvccData but not the Chunk's size
+  // has been incremented. This is because Chunk::size looks at the size of the first segment. The Insert operator
```
Bensk1: `Chunk::size()`

src/lib/storage/mvcc_data.hpp
```diff
@@ -37,10 +44,19 @@ struct MvccData {
...
   /**
-   * Grows mvcc data by the given delta
+   * Grows mvcc data by the given delta, guarded by the table's append_mutex.
    */
   void grow_by(size_t delta, TransactionID transaction_id, CommitID begin_commit_id);
 
+  /**
+   * The thread sanitzer (tsan) complains about concurrent writes and reads to begin/end_cids. That is because it is
+   * unaware of their thread-safety being guaranteed by the update of the global last_cid. Furthermore, we exploit that
+   * writes up to eight bytes are atomic on x64, which C++ and tsan do not know about. These helper methods were added
```
Bensk1: Regarding the "writes up to eight bytes are atomic on x64" guarantee. Do you have any idea why they don't take  the current architecture into account for sanitization?

src/lib/storage/mvcc_data.hpp
```diff
@@ -37,10 +44,19 @@ struct MvccData {
   void shrink();
 
   /**
-   * Grows mvcc data by the given delta
+   * Grows mvcc data by the given delta, guarded by the table's append_mutex.
```
Bensk1: For me that sounds like `grow_by` is somehow ensuring that the table's mutex is guarding it. But the mutex is only guarding the operation because the Insert operator is acquiring it, right?

Bensk1:
> No idea why this got faster. :/

That really makes no sense, does it?

src/lib/storage/mvcc_data.hpp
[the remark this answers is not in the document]
julianmenzler: Alright, legit strategy.

src/lib/storage/mvcc_data.hpp
mrks: I guess because tsan is architecture-agnostic. C++ says "that's UB", so you're on your own

tobodner: How about our friends from IBM/s390?

src/lib/storage/table.cpp
```diff
@@ -208,14 +208,20 @@ void Table::append_chunk(const Segments& segments, std::shared_ptr<MvccData> mvc
...
 
-#if HYRISE_DEBUG
-  for (const auto& segment : segments) {
-    const auto is_reference_segment = std::dynamic_pointer_cast<ReferenceSegment>(segment) != nullptr;
-    DebugAssert(is_reference_segment == (_type == TableType::References), "Invalid Segment type");
+  if constexpr (HYRISE_DEBUG) {
+    for (const auto& segment : segments) {
+      const auto is_reference_segment = std::dynamic_pointer_cast<ReferenceSegment>(segment) != nullptr;
+      Assert(is_reference_segment == (_type == TableType::References), "Invalid Segment type");
+    }
```
tobodner: I suspect that the constexpr changes do not really relate to the rest of this PR but am OK of sneaking them in.

src/lib/storage/table.hpp
```diff
@@ -198,9 +198,11 @@ class Table : private Noncopyable {
    *
    * With C++20 we will get std::atomic<std::shared_ptr<T>>, which allows us to omit the std::atomic_load() and
    * std::atomic_store() function calls.
+   *
+   * For the zero_allocator, see the implementation of Table::append_chunk.
    */
 
-  tbb::concurrent_vector<std::shared_ptr<Chunk>> _chunks;
+  tbb::concurrent_vector<std::shared_ptr<Chunk>, tbb::zero_allocator<std::shared_ptr<Chunk>>> _chunks;
```
tobodner: How does this relate to our PMR vectors? We cannot use another memory resource here?

tobodner:
> First half of my review. The second follows tonight.

I will need some more time over the weekend for this. Looking good so far.

src/lib/storage/mvcc_data.hpp
mrks: I'll shoot them an email

src/lib/storage/mvcc_data.hpp
mrks: "The caller should guard this using the table's append_mutex." - better?

resources/.tsan-ignore.txt
```diff
@@ -1,5 +1,7 @@
-# There are multiple data races in libtbb (probably false positives)
-race:^tbb*
+# There are multiple data races identified in the uninstrumented libtbb (probably false positives)
+# However, be aware of concurrent_vector entries being potentially uninitialized:
+#   https://software.intel.com/en-us/blogs/2009/04/09/delusion-of-tbbconcurrent_vectors-size-or-3-ways-to-traverse-in-parallel-correctly
+race:^tbb::*
```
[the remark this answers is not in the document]
mrks: We aren't. As tsan does not like concurrent_vector in general, I am not sure how to fix that.

src/lib/storage/table.cpp
mrks: Nope, but changing an `#if` to a `if constexpr` hardly warrants its own PR. People may argue "but what if we need to revert this commit", but in the history of this repository, there has been only a single revert on master.

src/lib/storage/table.hpp
mrks: I can't follow. Could you please rephrase your question?

Bensk1:
> > No idea why this got faster. :/
> 
> That really makes no sense, does it?

That seems to be resolved.

src/lib/storage/mvcc_data.hpp
Bensk1: Much better!

julianmenzler:
I still have some comments to add. Try to finish this review now.

src/test/server/server_test_runner.cpp
```diff
@@ -221,13 +221,79 @@ TEST_F(ServerTestRunner, TestParallelConnections) {
...
   }
 
-  for (auto& thread_fut : thread_futures) {
+  for (auto& thread_future : thread_futures) {
     // We give this a lot of time, not because we need that long for 100 threads to finish, but because sanitizers and
     // other tools like valgrind sometimes bring a high overhead that exceeds 10 seconds.
-    if (thread_fut.wait_for(std::chrono::seconds(150)) == std::future_status::timeout) {
+    if (thread_future.wait_for(std::chrono::seconds(150)) == std::future_status::timeout) {
       ASSERT_TRUE(false) << "At least one thread got stuck and did not commit.";
+      thread_future.get();
```
julianmenzler: What's the purpose of this line?

src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -61,8 +61,13 @@ void AbstractTableGenerator::generate_and_store() {
       table = std::make_shared<Table>(immutable_sorted_table->column_definitions(), TableType::Data,
                                       Chunk::DEFAULT_SIZE, UseMvcc::Yes);
       for (auto chunk_id = ChunkID{0}; chunk_id < immutable_sorted_table->chunk_count(); ++chunk_id) {
-        auto mvcc_data = std::make_shared<MvccData>(immutable_sorted_table->get_chunk(chunk_id)->size(), CommitID{0});
-        table->append_chunk(immutable_sorted_table->get_chunk(chunk_id)->segments(), mvcc_data);
+        const auto chunk = immutable_sorted_table->get_chunk(chunk_id);
+        auto mvcc_data = std::make_shared<MvccData>(chunk->size(), CommitID{0});
+        Segments segments{};
+        for (auto column_id = ColumnID{0}; column_id < immutable_sorted_table->column_count(); ++column_id) {
```
julianmenzler: Function call in for-declaration

src/lib/server/result_serializer.cpp
```diff
@@ -61,15 +61,21 @@ void ResultSerializer::send_query_response(
 
   // Iterate over each chunk in result table
   for (ChunkID chunk_id{0}; chunk_id < chunk_count; chunk_id++) {
-    const auto& current_chunk = table->get_chunk(chunk_id);
-    const auto chunk_size = current_chunk->size();
-    const auto& segments = current_chunk->segments();
+    const auto& chunk = table->get_chunk(chunk_id);
```
julianmenzler:
```suggestion
    const auto chunk = table->get_chunk(chunk_id);
```

src/test/server/server_test_runner.cpp
```diff
@@ -221,13 +221,79 @@ TEST_F(ServerTestRunner, TestParallelConnections) {
...
+    auto pipeline = SQLPipelineBuilder{std::string{"SELECT SUM(a) FROM table_a"}}.create_pipeline();
+    const auto [_, table] = pipeline.get_result_table();
+    final_sum = table->get_value<int64_t>(ColumnID{0}, 0);
+  }
+
+  // Really pessimistic, but at least 2 statements should have made it
+  EXPECT_GT(successful_increments, 2);
+
+  // We also want to see at least one conflict so that we can be sure that conflict handling in the server works
+  EXPECT_GE(successful_increments, 1);
```
julianmenzler:
```suggestion
  EXPECT_GE(conflicted_increments, 1);
```

src/lib/operators/get_table.cpp
```diff
@@ -71,44 +71,52 @@ void GetTable::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeV
 std::shared_ptr<const Table> GetTable::_on_execute() {
   const auto stored_table = Hyrise::get().storage_manager.get_table(_name);
 
+  // The chunk count might change while we are in this method as other threads insert new data. Thanks to MVCC, we need
+  // (read: must) not be interested in these chunks
```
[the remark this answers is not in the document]
mrks:
```
  // The chunk count might change while we are in this method as other threads insert new data. MVCC guarantees that
  // rows that are inserted after this transactions was started (and thus after GetTable started to execute) are not
  // visible. Thus, we do not have to care about chunks added after this point. By retrieving chunk_count only once,
  // we avoid concurrency issues, for example when more chunks are added to output_chunks than entries were originally
  // allocated.
```

src/test/concurrency/stress_test.cpp
```diff
@@ -0,0 +1,147 @@
...
+  EXPECT_FLOAT_EQ(final_sum - initial_sum, successful_increments);
+}
+
+TEST_F(StressTest, TestTransactionInserts) {
+  // An insert-heavy load on a table with a ridiculously low max chunk size, creating many new chunks
+  TableColumnDefinitions column_definitions;
+  column_definitions.emplace_back("a", DataType::Int, false);
+  column_definitions.emplace_back("b", DataType::Int, false);
+  const auto table = std::make_shared<Table>(column_definitions, TableType::Data, 3, UseMvcc::Yes);
+  Hyrise::get().storage_manager.add_table("table_b", table);
```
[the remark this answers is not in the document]
mrks: As this is only used once, I'd rather keep it here. Not really for performance reasons, but because I don't like huge setup methods where it is unclear what is used for what purpose

src/test/concurrency/stress_test.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+  // Really pessimistic, but at least 2 statements should have made it
+  EXPECT_GT(successful_increments, 2);
+
+  EXPECT_EQ(successful_increments + conflicted_increments, num_threads * iterations_per_thread);
+  EXPECT_FLOAT_EQ(final_sum - initial_sum, successful_increments);
+}
+
+TEST_F(StressTest, TestTransactionInserts) {
+  // An insert-heavy load on a table with a ridiculously low max chunk size, creating many new chunks
```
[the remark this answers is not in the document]
mrks: added "This is different from TestTransactionConflicts, which uses updates, in that each iteration is guaranteed to create a row. In the other test, a failed "mark for deletion" (i.e., swap of the row's tid) would lead to no row being appended."

src/test/server/server_test_runner.cpp
mrks: Added "Retrieve the future so that exceptions stored in its state are thrown"

src/lib/storage/mvcc_data.hpp
mrks: Conceptually, that should not be necessary. Part of MVCCs appeal is that we do not need atomics here. Also, this would mean that each read would have to be atomic.

As this comes only into play when we have concurrent modifications, the TPC-H doesn't help. Here are numbers from a slightly modified stress test, using 200 iterations per thread instead of 20:

Without atomics:
```
[ RUN      ] StressTest.TestTransactionConflicts
[       OK ] StressTest.TestTransactionConflicts (4287 ms)
```

With atomics:
```
[ RUN      ] StressTest.TestTransactionConflicts
[       OK ] StressTest.TestTransactionConflicts (4514 ms)
```

I'll admit that this is not too significant, but I am pretty sure that the synchronization that comes with atomic access would have a performance cost.

resources/.tsan-ignore.txt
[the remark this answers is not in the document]
tobodner: Does it make sense to limit the regex to tbb::concurrent_vector?

src/lib/storage/table.hpp
tobodner: What if we wanted to introduce an own allocator? Inherit from zero_allocator for this particular place here?

mrks: `¯\_(ツ)_/¯`

As long as whatever we use zeros the memory, we will be just fine.

mrks: Also, this is not a place where we would care for PMR.

src/lib/operators/get_table.cpp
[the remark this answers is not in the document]
tobodner: I am OK with the initial comment, but it is always good to know what confuses others.

resources/.tsan-ignore.txt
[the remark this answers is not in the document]
mrks: At least the queue has similar issues. I am afraid that pretty much everything in an uninstrumented tbb will cause warnings.

src/test/server/server_test_runner.cpp
tobodner: Please refer to @toniSta's comment from https://github.com/hyrise/hyrise/issues/1908

mrks: I'd rather have the comments self-contained where possible. Which information are you missing?

src/test/server/server_test_runner.cpp
```diff
@@ -221,13 +223,85 @@ TEST_F(ServerTestRunner, TestParallelConnections) {
...
+    auto pipeline = SQLPipelineBuilder{std::string{"SELECT SUM(a) FROM table_a"}}.create_pipeline();
+    const auto [_, table] = pipeline.get_result_table();
+    final_sum = table->get_value<int64_t>(ColumnID{0}, 0);
+  }
+
+  // Really pessimistic, but at least 2 statements should have made it
+  EXPECT_GT(successful_increments, 2);
+
+  // We also want to see at least one conflict so that we can be sure that conflict handling in the server works
+  EXPECT_GE(conflicted_increments, 1);
```
tobodner: That's generally a good thing to do, maybe additionally suggest what to do if this fails - maybe increase the concurrency of this test.

tobodner: And, why don't we do this in the internal stress test?

src/lib/operators/get_table.cpp
```diff
@@ -71,44 +71,55 @@ void GetTable::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeV
 std::shared_ptr<const Table> GetTable::_on_execute() {
   const auto stored_table = Hyrise::get().storage_manager.get_table(_name);
 
+  // The chunk count might change while we are in this method as other threads insert new data. MVCC guarantees that
```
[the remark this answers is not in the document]
mrks: For me, "other threads" always implies "concurrently", but why not

src/test/concurrency/stress_test.cpp
```diff
@@ -0,0 +1,159 @@
+#include <future>
+#include <thread>
+
+#include "base_test.hpp"
```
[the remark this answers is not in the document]
mrks: It's not really a corresponding hpp, so neither I nor the linter will like that

src/test/server/server_test_runner.cpp
mrks: > That's generally a good thing to do, maybe additionally suggest what to do if this fails - maybe increase the concurrency of this test.

done

> And, why don't we do this in the internal stress test?

I'm not really happy about the existence of this. Internally, we have more detailed concurrency tests than this. We just need it for the server because we currently have no psql-compatible way of guaranteeing a conflict.

src/test/concurrency/stress_test.cpp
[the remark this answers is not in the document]
julianmenzler: Fair.

PR #1917, opened 2019-11-07
Make return types a little bit more consistent.

Bensk1:


src/lib/storage/chunk.hpp
```diff
@@ -161,13 +161,13 @@ class Chunk : private Noncopyable {
    * However, `size() - invalid_row_count()` does not necessarily tell you how many rows are visible for
    * the current transaction.
    */
-  uint32_t invalid_row_count() const { return _invalid_row_count.load(); }
+  ChunkOffset invalid_row_count() const { return _invalid_row_count.load(); }
```
Bensk1: This is not really what we want here but, as discussed yesterday, creating a strong typedef for ChunkOffset will take much more time and this might be a bit less wrong here.

src/lib/storage/chunk.cpp
```diff
@@ -200,7 +200,7 @@ std::vector<std::shared_ptr<const BaseSegment>> Chunk::_get_segments_for_ids(
     const std::vector<ColumnID>& column_ids) const {
   DebugAssert(([&]() {
                 for (auto column_id : column_ids)
-                  if (column_id >= column_count()) return false;
+                  if (static_cast<ColumnCount>(column_id) >= column_count()) return false;
```
mrks: For me, it makes more sense to convert the right side to a ColumnID

PR #1921, opened 2019-11-11
Reference LQP nodes in PQP nodes

Bouncner:
This is a discussion PR to reference LQP nodes in PQP nodes during translation.

This PR is a first attempt to allow better parsing of the plan cache and was initially "proposed"/envisioined by @Bensk1 and @mrzzzrm . One of the main motivations was to retrieve the actual physical column when parsing scans/join etc.

Parsing of PQPs is something that @Bensk1  and I already do often, and what will probably be done much more often in the future (NVM, bachelor's project, ...). This PR is one of several PRs coming which will lay the foundation for the best self-pimping database ever.

There is a just a simple test included and I have probably not covered all cases to consider, yet. But should be enough to discuss the issue.

### DIFF
--- a/src/lib/logical_query_plan/lqp_translator.cpp
+++ b/src/lib/logical_query_plan/lqp_translator.cpp
@@ -204,8 +204,9 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_predicate_node_to_in
 
   // All chunks that have an index on column_ids are handled by an IndexScan. All other chunks are handled by
   // TableScan(s).
-  auto index_scan = std::make_shared<IndexScan>(input_operator, SegmentIndexType::GroupKey, column_ids,
-                                                predicate->predicate_condition, right_values, right_values2);
+  auto index_scan =
+      std::make_shared<IndexScan>(input_operator, SegmentIndexType::GroupKey, column_ids,
+                                  predicate->predicate_condition, right_values, right_values2, node->deep_copy());
 
   const auto table_scan = _translate_predicate_node_to_table_scan(node, input_operator);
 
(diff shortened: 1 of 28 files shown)


src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -204,8 +204,9 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_predicate_node_to_in
 
   // All chunks that have an index on column_ids are handled by an IndexScan. All other chunks are handled by
   // TableScan(s).
-  auto index_scan = std::make_shared<IndexScan>(input_operator, SegmentIndexType::GroupKey, column_ids,
-                                                predicate->predicate_condition, right_values, right_values2);
+  auto index_scan =
+      std::make_shared<IndexScan>(input_operator, SegmentIndexType::GroupKey, column_ids,
+                                  predicate->predicate_condition, right_values, right_values2, node->deep_copy());
```
mrks: Does this have to be part of the constructor? It really has nothing to do with the purpose of the object being constructed.

As the LQPTranslator is the only one setting it, maybe a single virtual call in translate_node would even be easier?

mrks: Why do you deep_copy the LQP?

Bouncner: @Bensk1, do you remember the reason?

Bouncner: Good point. Probably opening a new PR and closing this one to handle that.

Bouncner:
Closing for now.

mrks:
When you bring this back, we should use the opportunity and assert that operators return whatever their LQP node's column_expressions promised.

Bouncner:
I throw away this approach and followed your recommendation of setting it in the translator (#1931, which is already merged).

mrks:
Ah, missed that one.

PR #1924, opened 2019-11-14
A potpourri of smaller fixes

mrks:
Some minor things that I had lying around. As none of these should have any effect on benchmarks etc., I hope that compressing them into one PR should be fine

* Make clang-format work independent of current clang version
* Make visualization work if benchmark item has multiple SQL pipelines (e.g., TPC-C)
* Make `ReadWriteOperatorState` printable
* Fix visualization of `x IN (...)` when the IN expression is rewritten to a join

### DIFF
--- a/src/benchmarklib/benchmark_sql_executor.cpp
+++ b/src/benchmarklib/benchmark_sql_executor.cpp
@@ -100,15 +100,28 @@ void BenchmarkSQLExecutor::_compare_tables(const std::shared_ptr<const Table>& e
   }
 }
 
-void BenchmarkSQLExecutor::_visualize(SQLPipeline& pipeline) const {
+void BenchmarkSQLExecutor::_visualize(SQLPipeline& pipeline) {
   GraphvizConfig graphviz_config;
   graphviz_config.format = "svg";
 
   const auto& lqps = pipeline.get_optimized_logical_plans();
   const auto& pqps = pipeline.get_physical_plans();
 
-  LQPVisualizer{graphviz_config, {}, {}, {}}.visualize(lqps, *_visualize_prefix + "-LQP.svg");
-  PQPVisualizer{graphviz_config, {}, {}, {}}.visualize(pqps, *_visualize_prefix + "-PQP.svg");
+  auto prefix = *_visualize_prefix;
+
+  if (_num_visualized_plans == 1) {
+    // We have multiple SQL pipelines in this benchmark item - rename the existing file
+    std::filesystem::rename(prefix + "-LQP.svg", prefix + "-0-LQP.svg");
+    std::filesystem::rename(prefix + "-PQP.svg", prefix + "-0-PQP.svg");
+  }
+  if (_num_visualized_plans > 0) {
+    prefix += "-" + std::to_string(_num_visualized_plans);
+  }
+
+  LQPVisualizer{graphviz_config, {}, {}, {}}.visualize(lqps, prefix + "-LQP.svg");
+  PQPVisualizer{graphviz_config, {}, {}, {}}.visualize(pqps, prefix + "-PQP.svg");
+
+  ++_num_visualized_plans;
 }
 
 }  // namespace opossum
--- a/src/benchmarklib/benchmark_sql_executor.cpp
+++ b/src/benchmarklib/benchmark_sql_executor.cpp
@@ -110,7 +110,7 @@ void BenchmarkSQLExecutor::_visualize(SQLPipeline& pipeline) {
   auto prefix = *_visualize_prefix;
 
   if (_num_visualized_plans == 1) {
-    // We have multiple SQL pipelines in this benchmark item - rename the existing file
+    // We have already visualized a prior SQL pipeline in this benchmark item - rename the existing file
     std::filesystem::rename(prefix + "-LQP.svg", prefix + "-0-LQP.svg");
     std::filesystem::rename(prefix + "-PQP.svg", prefix + "-0-PQP.svg");
   }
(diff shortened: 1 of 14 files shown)


src/benchmarklib/benchmark_sql_executor.cpp
```diff
@@ -100,15 +100,28 @@ void BenchmarkSQLExecutor::_compare_tables(const std::shared_ptr<const Table>& e
...
   const auto& lqps = pipeline.get_optimized_logical_plans();
   const auto& pqps = pipeline.get_physical_plans();
 
-  LQPVisualizer{graphviz_config, {}, {}, {}}.visualize(lqps, *_visualize_prefix + "-LQP.svg");
-  PQPVisualizer{graphviz_config, {}, {}, {}}.visualize(pqps, *_visualize_prefix + "-PQP.svg");
+  auto prefix = *_visualize_prefix;
+
+  if (_num_visualized_plans == 1) {
+    // We have multiple SQL pipelines in this benchmark item - rename the existing file
+    std::filesystem::rename(prefix + "-LQP.svg", prefix + "-0-LQP.svg");
```
Bouncner: `== 1` signals multiple pipelines?

Bouncner: Ah ... there's an increment. Ok ...

Bouncner: Looks pretty strange, but ok.

src/benchmarklib/benchmark_sql_executor.hpp
```diff
@@ -40,10 +40,11 @@ class BenchmarkSQLExecutor {
                        const std::shared_ptr<const Table>& actual_result_table,
                        const std::optional<const std::string>& description = std::nullopt);
   void _verify_with_sqlite(SQLPipeline& pipeline);
-  void _visualize(SQLPipeline& pipeline) const;
+  void _visualize(SQLPipeline& pipeline);
 
   const std::shared_ptr<SQLiteWrapper> _sqlite_wrapper;
   const std::optional<std::string> _visualize_prefix;
+  int _num_visualized_plans{0};
```
Bouncner: Why `int` and not something unsigned?

src/lib/optimizer/strategy/in_expression_rewrite_rule.cpp
```diff
@@ -43,6 +44,10 @@ void rewrite_to_join(const std::shared_ptr<AbstractLQPNode>& node,
     list_as_table->append_chunk({value_segment});
   });
 
+  // All regular tables have statistics. Even if the JoinOrderingRule already ran, add statistics to the dummy table so
+  // that following rules, which expect statistics to be present, do not run into problems.
+  list_as_table->set_table_statistics(TableStatistics::from_table(*list_as_table));
```
Bouncner: I don't get what this has to do with the JoinOrderingRule.

src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp
```diff
@@ -248,19 +253,25 @@ TEST_F(InExpressionRewriteRuleTest, JoinStrategy) {
...
+    EXPECT_FLOAT_EQ(cardinality_estimator.estimate_cardinality(result_lqp), 1000.f / 200 * 1);
   }
 
   {
     // ExpressionEvaluator for five elements
     const auto input_lqp = PredicateNode::make(five_element_in_expression, node);
     const auto result_lqp = StrategyBaseTest::apply_rule(rule, input_lqp);
     EXPECT_EQ(result_lqp, input_lqp);
+
+    // No cardinality check here, as unmodified IN expressions are currently not estimated by the CardinalityEstimator.
```
Bouncner: Do I have to read the rule to understand why 5 elements leads to an "unmodified" expression? I'd guess the threshold for a table is larger? Maybe comment?

src/benchmarklib/benchmark_sql_executor.cpp
mrks: Any suggestions?

src/benchmarklib/benchmark_sql_executor.hpp
mrks: Didn't really put too much thought into it, will change.

src/lib/optimizer/strategy/in_expression_rewrite_rule.cpp
mrks: Re-reading that I wonder if I didn't write too much. Maybe "Add statistics to the dummy table so that following rules, which expect statistics to be present, do not run into problems. " is sufficient?

src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp
mrks: Yes, rewrites to join only happen with at least `MIN_ELEMENTS_FOR_JOIN` elements. Someone who looks into this rule and its test (and not at a small PR that does not really touch the rule) should have already seen that a couple of times, so I am not sure if a comment is helpful.

src/benchmarklib/benchmark_sql_executor.cpp
mrks: Maybe "We have already visualized a prior SQL pipeline in this benchmark item"?

Bouncner: Would be better. Probably not worth thinking too much about it.

src/lib/optimizer/strategy/in_expression_rewrite_rule.cpp
Bouncner: Yep, better.

src/test/optimizer/strategy/in_expression_rewrite_rule_test.cpp
Bouncner: Hmm, since both the 100 and 5 are so arbitrary, I think a one liner that the 5 is below `MIN_ELEMENTS_FOR_JOIN` and 100 is above would help.

mrks: better?

PR #1927, opened 2019-11-14
Benchmark: make --runs default to infinity

### LINKED ISSUE #1919: Make benchmark --runs parameter default to infinity
The current max of 10000 runs means that a multi-threaded run of TPC-H 6 runs for less than there seconds on nemea. Not enough for reliable numbers. Instead of arbitrarily increasing that value, we should let it default to -1, meaning infinity.

I prefer -1 over 0 because I sometimes use `-r 0` to generate an empty JSON file for comparison.

mrks:
fixes #1919

### DIFF
--- a/scripts/test/hyriseBenchmarkTPCH_test.py
+++ b/scripts/test/hyriseBenchmarkTPCH_test.py
@@ -16,7 +16,7 @@ def main():
   arguments["--use_prepared_statements"] = "true"
   arguments["--queries"] = "'1,13,19'"
   arguments["--time"] = "10"
-  arguments["--runs"] = "100"
+  arguments["--runs"] = "-1"
   arguments["--output"] = "'json_output.txt'"
   arguments["--mode"] = "'Shuffled'"
   arguments["--encoding"] = "'Dictionary'"
@@ -34,7 +34,6 @@ def main():
   benchmark.expect("Running benchmark in 'Shuffled' mode")
   benchmark.expect("Encoding is 'Dictionary'")
   benchmark.expect("Chunk size is 100000")
-  benchmark.expect("Max runs per item is 100")
   benchmark.expect("Max duration per item is 10 seconds")
   benchmark.expect("No warmup runs are performed")
   benchmark.expect("Not caching tables as binary files")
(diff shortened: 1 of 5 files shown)


src/benchmarklib/benchmark_config.cpp
```diff
@@ -3,7 +3,7 @@
 namespace opossum {
 
 BenchmarkConfig::BenchmarkConfig(const BenchmarkMode benchmark_mode, const ChunkOffset chunk_size,
-                                 const EncodingConfig& encoding_config, const bool indexes, const size_t max_runs,
+                                 const EncodingConfig& encoding_config, const bool indexes, const int64_t max_runs,
```
tobodner: Do you think that we will ever exceed int32_t here?

scripts/test/hyriseBenchmarkTPCH_test.py
```diff
@@ -16,7 +16,7 @@ def main():
   arguments["--use_prepared_statements"] = "true"
   arguments["--queries"] = "'1,13,19'"
   arguments["--time"] = "10"
-  arguments["--runs"] = "100"
+  arguments["--runs"] = "-1"
```
tobodner: Remove, since this is now back to default of benchmark config?

src/benchmarklib/benchmark_config.cpp
Bouncner: I wouldn’t be surprised if we did. Especially some long running self-pimping tests (or super slow lambda functions 😁) might run TPC-H Q6 for hours.
And it basically doesn’t cost anything.

scripts/test/hyriseBenchmarkTPCH_test.py
mrks: The idea is to see what is tested in this test

src/benchmarklib/benchmark_config.cpp
tobodner: Alright.

scripts/test/hyriseBenchmarkTPCH_test.py
tobodner: I do not like redundancy, which is why I brought it up. I am OK with this as well.

mrks: I wouldn't call it redundant. If this was a defaulted c++ parameter, I would also want to explicitly test the parameter, not relying on whatever was set as a default

PR #1929, opened 2019-11-15
Optimizer rule: remove dependent group-by columns

Bouncner:
This PR partially implements the choke point 1.4 of Boncz et al.'s "TPC-H Analyzed: Hidden Messages and Lessons Learned from an Influential Benchmark".

The concept is simple. Whenever we have the primary key(s) and other columns (of the same table) in the group by list, we can remove the other columns are they are functionally dependent and thus do not have an effect on the grouping.
Often, columns are grouped to allow later access to them (see Q10). To cover this case, we added the `ANY()` aggregation function which is a sole internal function that returns "any" element of the group (it is thus only used when we know there is only a single distinct value).

We only "partially" implemented the rule, as Boncz et al. also list potential optimization for which we require foreign keys. Due to the current lack of these, we can only optimize TPC-H Q10 and Q18.
For Q18, the runtime of the aggregate only contributes a very small part of the runtime. Thus, there are no big gains to expect.
The regressions appear to be noise (tested multiple times).

TPC-H - SF 1:
```diff
 +----------------+----------------+-------+----------------+-------+------------+---------+
 | Benchmark      | prev. iter/s   | runs  | new iter/s     | runs  | change [%] | p-value |
 +----------------+----------------+-------+----------------+-------+------------+---------+
 | TPC-H 01       | 0.667014181614 | 41    | 0.672787129879 | 41    | +1%        |  0.1814 |
 | TPC-H 02       | 69.6656951904  | 4180  | 70.0439300537  | 4203  | +1%        |  0.4237 |
+| TPC-H 03       | 7.06638860703  | 424   | 7.62346124649  | 458   | +8%        |  0.0000 |
 | TPC-H 04       | 6.84692955017  | 411   | 6.82267856598  | 410   | -0%        |  0.4807 |
 | TPC-H 05       | 3.58185744286  | 215   | 3.59642386436  | 216   | +0%        |  0.7215 |
-| TPC-H 06       | 226.290054321  | 13578 | 211.505340576  | 12691 | -7%        |  0.0000 |
 | TPC-H 07       | 7.54159021378  | 453   | 7.54759073257  | 453   | +0%        |  0.8982 |
 | TPC-H 08       | 6.46811437607  | 389   | 6.42894029617  | 386   | -1%        |  0.1270 |
 | TPC-H 09       | 1.58041131496  | 95    | 1.59978067875  | 96    | +1%        |  0.0106 |
+| TPC-H 10       | 2.77212309837  | 167   | 2.9676861763   | 179   | +7%        |  0.0000 |
 | TPC-H 11       | 29.7536792755  | 1786  | 29.8762893677  | 1793  | +0%        |  0.0013 |
 | TPC-H 12       | 14.1946849823  | 852   | 14.3648328781  | 862   | +1%        |  0.0000 |
 | TPC-H 13       | 2.40185832977  | 145   | 2.44331884384  | 147   | +2%        |  0.0000 |
 | TPC-H 14       | 48.9279594421  | 2936  | 48.4898643494  | 2910  | -1%        |  0.0000 |
 | TPC-H 15       | 95.6900253296  | 5742  | 97.5214691162  | 5852  | +2%        |  0.0000 |
 | TPC-H 16       | 7.27398967743  | 437   | 7.30749702454  | 439   | +0%        |  0.0443 |
 | TPC-H 17       | 4.52551174164  | 272   | 4.47930812836  | 269   | -1%        |  0.1500 |
 | TPC-H 18       | 1.22573947906  | 74    | 1.23251497746  | 74    | +1%        |  0.7050 |
 | TPC-H 19       | 8.37693023682  | 503   | 7.98328590393  | 480   | -5%        |  0.0000 |
 | TPC-H 20       | 27.3785839081  | 1643  | 27.3152484894  | 1639  | -0%        |  0.2920 |
 | TPC-H 21       | 1.18774628639  | 72    | 1.17874157429  | 71    | -1%        |  0.2301 |
+| TPC-H 22       | 15.7339887619  | 945   | 16.5302753448  | 992   | +5%        |  0.0000 |
 | geometric mean |                |       |                |       | +1%        |         |
 +----------------+----------------+-------+----------------+-------+------------+---------+
```


Bouncner:
It is zum Mäuse melking!

mrks:
?

tobodner:
> It is zum Mäuse melking!

Any major roadblocks here?

Bouncner:
It’s my stupidness combined with a severe lack of LQP fiddling experience.

Just closing since I don’t like stale PRs. Will be reopened as soon as I have fixed the remaining issues.

mrks:
> lack of LQP fiddling experience

Let me know if you want a second pair of eyes

Bouncner:
And another shot. Thanks a lot @mrks . Your ideas significantly simplified the rule.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include "abstract_rule.hpp"
+#include "expression/abstract_expression.hpp"
+#include "expression/lqp_column_expression.hpp"
+
+namespace opossum {
+
+class AbstractLQPNode;
+
+// Removes columns from list of group by columns when they are functionally dependent from column(s) of the grouping.
```
mrks: An example would be a better introduction than many words

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
```diff
@@ -0,0 +1,27 @@
...
+// This is currently only the case, when both the primary key and other columns of the same table are grouped.
```
mrks: -,

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
```diff
@@ -0,0 +1,27 @@
...
+// This is regularly the case, when the columns are supposed to be later accessed (e.g., see usage of "c_acctbal" in
+// TPC-H query 10). As "c_acctbal" does not need to be in the grouping (since the primary key c_custkey is present),
```
mrks: The "this is regularly..." sentence is a bit confusing. Rather give an example and describe it?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
```diff
@@ -0,0 +1,27 @@
...
+// the column in added as `ANY(c_acctbal)` to the aggregation list. ANY() selects "any" value from the list of values
+// per group (since we group by the primary key in this case, the group is ensured to be of size one).
+// This rule implements choke point 1.4 of "TPC-H Analyzed: Hidden Messages and Lessons Learned from an Influential
+// Benchmark" (Boncz et al.). Due to the current lack of foreign key suppport, not all queries that are listed in the
+// paper can be optimized.
+// When the group by reduction changes the column order and this order is also the final column order of the plan, a
+// projection is appended as the new root to restore the column order.
```
mrks: Not sure if this is understood by others. I made my own attempt, maybe we can merge the two into something good:

> In SQL, the returned columns of an aggregate are either "group by columns" or "aggregate columns, e.g., sum(x)". A query to select the customer's id, their name, and the number of orders could look like this:
SELECT customer.id, customer.name, COUNT(*) FROM customer JOIN orders ON ... GROUP BY customer.id, customer.name
Note how customer.name is included as a group by column, even though the groups are already exhaustively defined by `GROUP BY customer.id`. The reason for that is the functional dependency `id => name`, or, more practically speaking, the fact that `customer.id` is a primary key.
This rule identifies cases where functionally dependent columns are given as group by columns while their primary key column(s) is a/are also group by column(s). It then replaces that column (e.g., customer.name) with a dummy aggregate function ANY(customer.name). This reduces the grouping cost in the aggregate operator.
As the aggregate operators first produce all group by column, followed by all aggregate columns (including ANY), the order might be changed. Unless a following Projection redefines the order anyway, a new Projection is inserted to restore the original column order.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,132 @@
...
+  // Store a copy of the root's column expressions.
+  const auto root_column_expressions = lqp->column_expressions();
+  visit_lqp(lqp, [&](const auto& node) {
+    if (node->type != LQPNodeType::Aggregate) {
+      return LQPVisitation::VisitInputs;
+    }
+    auto& aggregate_node = static_cast<AggregateNode&>(*node);
+
+    std::unordered_map<std::shared_ptr<const StoredTableNode>, std::set<ColumnID>> group_by_columns_per_table;
+    // Collect the group-by columns for each table in the aggregate node
```
mrks: Aggregate columns can be skipped, thus using an index-based for loop

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,132 @@
...
+    for (auto expression_idx = size_t{0}; expression_idx < aggregate_node.aggregate_expressions_begin_idx;
+         ++expression_idx) {
+      const auto& expression = aggregate_node.node_expressions[expression_idx];
+      const auto& column_expression = std::dynamic_pointer_cast<LQPColumnExpression>(expression);
+      if (!column_expression) continue;
```
mrks: Doc this. `a+1` could generally be relevant too, we just decided not to cover that case yet.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+      const auto& stored_table_node =
+          std::dynamic_pointer_cast<const StoredTableNode>(column_expression->column_reference.original_node());
+      // If column is not a physical column skip
+      if (!stored_table_node) continue;
```
mrks: StaticTableNodes also have physical columns

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+      const auto column_id = column_expression->column_reference.original_column_id();
+
+      auto inserted = group_by_columns_per_table.try_emplace(stored_table_node, std::set<ColumnID>{column_id});
+      if (!inserted.second) {
+        group_by_columns_per_table[stored_table_node].insert(column_id);
+      }
```
mrks:
```
group_by_columns_per_table[stored_table_node].insert(column_id);
```
would be sufficient, even though you should add a comment on why that is the case.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,132 @@
...
+    // In case the initial query plan root returned the same columns in the same column order and was not a projection,
+    // it is likely that the result of the current aggregate was either the root itself or only operators followed that
+    // do not modify the column order (e.g., sort or limit). In this case, we need to restore the initial column order
+    // by adding a projection with the initial column_references since we changed the column order by moving columns
+    // from the group-by list to the aggregations.
+    if (group_by_list_changed && initial_aggregate_column_expressions == root_column_expressions &&
+        lqp->type != LQPNodeType::Projection) {
+      const auto projection_node = std::make_shared<ProjectionNode>(root_column_expressions);
+      lqp_insert_node(lqp, LQPInputSide::Left, projection_node);
+    }
```
mrks: Why don't you compare the original and current column order in the root?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    // Store copy of aggregate expression to enable restoring the original order of columns via a projection
+    const auto initial_aggregate_column_expressions = aggregate_node.column_expressions();
+    bool group_by_list_changed = false;
+
+    // Main loop. Iterate over the tables and its group-by columns, gather primary keys and see if we can reduce.
+    for (const auto& [stored_table_node, group_by_columns] : group_by_columns_per_table) {
+      // Obtain column IDs of the primary key
+      auto unique_columns = std::set<ColumnID>();
```
mrks: primary key vs unique. Why can't we operate on unique?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,132 @@
...
+      std::set_intersection(unique_columns.begin(), unique_columns.end(), group_by_columns.begin(),
+                            group_by_columns.end(), std::back_inserter(intersection));
+
+      // Skip the current table as the primary key is not completely present.
+      if (intersection.size() != unique_columns.size()) {
+        continue;
+      }
+
+      for (const auto& group_by_column : group_by_columns) {
+        // Every column that is not part of the primary key is going to be removed.
```
mrks: * that is part of the table but not part of the primary key
* removed? not adapted?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,132 @@
...
+        if (unique_columns.find(group_by_column) == unique_columns.end()) {
```
mrks: Guard clause?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,132 @@
...
+          // Remove node if it is a column reference and references the correct stored table node. Further, decrement
+          // the aggregate's index which denotes the end of group-by expressions.
+          aggregate_node.node_expressions.erase(
+              std::remove_if(aggregate_node.node_expressions.begin(), aggregate_node.node_expressions.end(),
+                             [&, stored_table_node = stored_table_node](const auto expression) {
```
mrks: `stored_table_node = stored_table_node`?

src/test/operators/aggregate_test.cpp
```diff
@@ -177,6 +178,52 @@ TYPED_TEST(OperatorsAggregateTest, CannotStandardDeviationSampleStringColumns) {
   EXPECT_THROW(aggregate->execute(), std::logic_error);
 }
 
+// The ANY aggregation is a special case which is used to obtain "any value" of a group of which we know that each
+// value in this group is the same (for most cases, the group will have a size of one). This can be the case, when
+// the aggragated column is functionally dependent on the group-by columns. In debug mode, we DebugAssert that each
+// value is the same within the group.
+TYPED_TEST(OperatorsAggregateTest, CannotUseAnyOnNonDependentColumn) {
```
mrks: "CannotUseAnyOnNonDependentColumn" - this is not really enforced anywhere is it? Isn't this test more a "ANY complains if not all values are equal"?

src/test/operators/aggregate_test.cpp
```diff
@@ -177,6 +178,52 @@ TYPED_TEST(OperatorsAggregateTest, CannotStandardDeviationSampleStringColumns) {
...
+      filtered, std::vector<AggregateColumnDefinition>{{ColumnID{2}, AggregateFunction::Any}},
+      std::vector<ColumnID>{ColumnID{0}, ColumnID{1}});
+  aggregate->execute();
+
+  auto first_column = PQPColumnExpression::from_table(*aggregate->get_output(), "a");
+  auto second_column = PQPColumnExpression::from_table(*aggregate->get_output(), "b");
+  auto third_column = PQPColumnExpression::from_table(*aggregate->get_output(), "ANY(c)");
+  auto projection = std::make_shared<opossum::Projection>(
+      aggregate, expression_vector(first_column, second_column, add_(third_column, 17)));
+  projection->execute();
```
mrks: ANY should never leave the aggregate - is this still current?

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -0,0 +1,273 @@
...
+class DependentGroupByReductionRuleTest : public StrategyBaseTest {
+ public:
+  void SetUp() override {
+    auto& storage_manager = Hyrise::get().storage_manager;
+
+    TableColumnDefinitions column_definitions{
+        {"column0", DataType::Int, false}, {"column1", DataType::Int, false}, {"column2", DataType::Int, false}};
+
+    table_a = std::make_shared<Table>(column_definitions, TableType::Data, 2, UseMvcc::Yes);
+    table_a->add_soft_unique_constraint({ColumnID{0}}, IsPrimaryKey::Yes);
```
mrks: Have you tested non-primary constraints?

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -56,7 +57,19 @@ std::string AggregateNode::description() const {
 }
 
 const std::vector<std::shared_ptr<AbstractExpression>>& AggregateNode::column_expressions() const {
-  return node_expressions;
+  _column_expressions.resize(node_expressions.size());
+  std::copy(node_expressions.begin(), node_expressions.end(), _column_expressions.begin());
+
+  for (auto& column_expression : _column_expressions) {
+    if (column_expression->type == ExpressionType::Aggregate) {
```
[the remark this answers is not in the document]
Bouncner: `DebugAssert` should be sufficient, right?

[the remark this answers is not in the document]
mrks: yepp

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
Bouncner: To be honest I never cared about them since I only thought about primary keys initially. Is there any reason to not include unique columns the same way we currently do with primary keys?
As far as I see it, the main difference is NULL values which should not matter in the aggregate.

mrks: Not an answer, but as discussed: Test ANY on completely NULL column

Bouncner: Tests still to be done, but as I understand unique constraints and primary keys now, we are fine with any unique constraint. So I would remove the check for primary key. Your opinion?

Bouncner: I first thought I would understand NULL, but apparently I don't.

Bouncner: So my current understanding: NULL are groupable (which I thought would not make sense, as they would not equal each other). If they are groupable, each single row of a group-by column that has a unique constraint (which allows NULLs) does no longer identify exactly one single row. Hence, we can only use the rule for primary keys.

Your take on that?

mrks: https://academy.vertabelo.com/blog/null-values-group-clause/

> To explain why NULLs are grouped into one bucket, we need to review the SQL standard. SQL defines “any two values that are equal to one another, or any two NULLs”, as “not distinct”. This definition of “not distinct” allows SQL to group and sort NULLs when the GROUP BY clause (or other keywords that perform grouping) is used.

(1) How would limiting this to primary keys reduce the applicability to the TPC-H queries?
(2) I am not sure how well we track NULLable information. If we can show that there are no NULL values, we are fine, too.

Bouncner: > (1) How would limiting this to primary keys reduce the applicability to the TPC-H queries?
> (2) I am not sure how well we track NULLable information. If we can show that there are no NULL values, we are fine, too.

(1) Wouldn't change anything since we don't have unique columns which are not also primary keys. But I don't know if there are further constraints allowed in TPC-H which we might use at a later point in time.
(2) Good point. I will change the check for PK to PK or Unique with Non-Nullable columns. I thought about NULLs that are later on added (e.g., outer joins) but they should not impact the rule as it would be still be safe to remove columns here since they are all NULL anyways for added rows. Does that make sense to you?

mrks: NULLs that are added later on should result in the expression being NULLable, so I don't see how that is an issue. We are talking about `is_nullable_on_lqp`, not `column_is_nullable`, right?

src/lib/expression/aggregate_expression.hpp
```diff
@@ -4,7 +4,7 @@
 
 namespace opossum {
 
-enum class AggregateFunction { Min, Max, Sum, Avg, Count, CountDistinct, StandardDeviationSample };
+enum class AggregateFunction { Min, Max, Sum, Avg, Count, CountDistinct, StandardDeviationSample, Any };
```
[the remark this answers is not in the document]
Bouncner: Added a comment.

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -56,7 +57,19 @@ std::string AggregateNode::description() const {
 }
 
 const std::vector<std::shared_ptr<AbstractExpression>>& AggregateNode::column_expressions() const {
-  return node_expressions;
+  _column_expressions.resize(node_expressions.size());
```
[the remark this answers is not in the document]
Bouncner: Added a comment.

src/lib/logical_query_plan/aggregate_node.cpp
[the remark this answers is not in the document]
Bouncner: Done.

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -56,7 +57,19 @@ std::string AggregateNode::description() const {
...
+  std::copy(node_expressions.begin(), node_expressions.end(), _column_expressions.begin());
+
+  for (auto& column_expression : _column_expressions) {
+    if (column_expression->type == ExpressionType::Aggregate) {
+      const auto aggregate_expression = std::dynamic_pointer_cast<AggregateExpression>(column_expression);
```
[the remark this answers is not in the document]
Bouncner: Done.

src/lib/operators/abstract_aggregate_operator.hpp
```diff
@@ -140,6 +140,21 @@ class AggregateFunctionBuilder<ColumnDataType, AggregateType, AggregateFunction:
   }
 };
 
+template <typename ColumnDataType, typename AggregateType>
+class AggregateFunctionBuilder<ColumnDataType, AggregateType, AggregateFunction::Any> {
+ public:
+  auto get_aggregate_function() {
+    return [](const ColumnDataType& new_value, std::optional<AggregateType>& current_primary_aggregate,
+              std::vector<AggregateType>& current_secondary_aggregates) {
+      DebugAssert(!current_primary_aggregate || *current_primary_aggregate == new_value, "Any violation");
```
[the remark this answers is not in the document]
Bouncner: Done.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
Bouncner: Done.

src/lib/operators/abstract_aggregate_operator.hpp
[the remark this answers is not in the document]
Bouncner: Changed to "ANY() expects all values in the **group** to be equal".

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
Bouncner: The issue of `error: reference to local binding 'stored_table_node' declared in enclosing lambda expression`. The problem of structured bindings and lambdas. Is there a better/preferred way to deal with that?

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
Bouncner: Starting at line 66, I now check if we have a constraint (no matter what kind, thus can be unique constraint or PK) and that this column is not nullable (in that case, unique columns are equal to PK columns from the stand point of this rule).

src/test/operators/aggregate_test.cpp
Bouncner: Right. `FailAnyOnNonDependentColumn` ?

src/lib/operators/aggregate_hash.cpp
```diff
@@ -418,6 +418,9 @@ void AggregateHash::_aggregate() {
               _aggregate_segment<ColumnDataType, AggregateFunction::StandardDeviationSample, AggregateKey>(
                   chunk_id, column_index, *base_segment, keys_per_chunk);
               break;
+            case AggregateFunction::Any:
```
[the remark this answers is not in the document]
Bouncner: Added this to lines 132..
```
else if constexpr (function == AggregateFunction::Any) {
        // Gathering the group's first value for ANY() is sufficient
        return;
      }
```

Tests no longer run. The change for the sort-based aggregate is fine however. Need to better understand the hash aggregate.

src/lib/sql/sql_translator.cpp
```diff
@@ -1357,6 +1357,7 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_translate_hsql_expr(
           case AggregateFunction::Sum:
           case AggregateFunction::Avg:
           case AggregateFunction::StandardDeviationSample:
+          case AggregateFunction::Any:
```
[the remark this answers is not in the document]
Bouncner: Added `Fail("ANY() is a pure internal aggregation function.");`.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
Bouncner: Added an example.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
Bouncner: Added:
```
        // In case the group-by column is not a column expression (e.g., grouping by `a+1`), we take the safe route and
        // continue even though constraints often hold for such simple expression.
```

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
Bouncner: Argh ... never heard of them. Is there a standard way to handle this? Shall I check for Static and Stored and then store a `BaseNonQueryNode` ptr?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
Bouncner: Wow, did not know that. But is a comment here necessary?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
Bouncner: Leftover from previous version. Changed to `aggregate_node.column_expressions() == root_column_expressions`.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
Bouncner: We can. Cleared that up now in the comments as well as in the initial description in the header.

src/test/operators/aggregate_test.cpp
Bouncner: Nope.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
Bouncner: Sounds good. Merged your text with mine.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
Bouncner: It is actually done by intent. We take the optimised aggregate's columns and if they have been equal to the root, there is probably no modifying operator following (that means limit, sort can follow, but when aggregate follows and is the root, we don't need to project).

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
Bouncner: Now, ignoring all group-by columns that are nullable.

Bouncner: Further, we now process the constraints by size and continue after one reduction was successful (also tested).

src/lib/operators/aggregate_hash.cpp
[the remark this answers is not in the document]
Bouncner: I don't think we can break here, since (in contrast to the consecutive group values in the sorted case) the hash join receives the group-by keys in an arbitrary order. We might just skip the following groups if we simply break here. And in the aggregation itself, we already check if a value has been set already.

Bouncner:
@mrks: is it worth the work to temporarily merge your aggregate changes and check for the performance?

mrks:
Shouldn't cause any conflicts, but wouldn't it be easier to bring that one into the master first?

src/test/operators/aggregate_test.cpp
```diff
@@ -177,6 +182,38 @@ TYPED_TEST(OperatorsAggregateTest, CannotStandardDeviationSampleStringColumns) {
...
+      this->_table_wrapper_2_2, equals_(this->get_column_expression(this->_table_wrapper_2_2, ColumnID{0}), 123));
+  filtered->execute();
+
+  {
+    auto aggregate = std::make_shared<TypeParam>(
+        filtered, std::vector<AggregateColumnDefinition>{{ColumnID{2}, AggregateFunction::Any}},
+        std::vector<ColumnID>{ColumnID{0}, ColumnID{1}});
+    aggregate->execute();
+
+    EXPECT_EQ(aggregate->get_output()->template get_value<int>(ColumnID{2}, 0u), 20);
```
mrks: I'd argue that ANY on non-equal values is Hyrise-UB, so `20` is not a "right" result that we should test for.

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// Test that a removable column is removed when the single primary key column is present.
+// Check for the restored colum order.
+TEST_F(DependentGroupByReductionRuleTest, SingleKeyReduction) {
+  // clang-format off
+  auto lqp =
+  AggregateNode::make(expression_vector(column_a_0, column_a_1),
+                      expression_vector(sum_(column_a_0), sum_(column_a_1),
+                                        sum_(column_a_2)),
```
mrks: Formatting: Nodes should take a single line, the next node should be indented. (Applies to whole file)

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+                      stored_table_node_b);
+
+  const auto actual_lqp = apply_rule(rule, lqp);
+
+  const auto expected_lqp =
+  AggregateNode::make(expression_vector(column_b_0, column_b_1),
+                      expression_vector(sum_(column_b_0), sum_(column_b_1),
+                                        sum_(column_b_2)),
+                      stored_table_node_b);
+  // clang-format on
```
mrks: For LQPs that are not expected to change, compare against a deep_copy (Applies to whole file)

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+                      expression_vector(sum_(column_a_0)),
+    JoinNode::make(JoinMode::FullOuter, equals_(column_a_0, column_b_0),
+                   stored_table_node_a,
+                   stored_table_node_b));
+  // clang-format on
+
+  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+}
+
+// Check that we reduce using the shortest (in term of number of columns) constraints.s
```
mrks: -s

Bouncner:
> Shouldn't cause any conflicts, but wouldn't it be easier to bring that one into the master first?

I am fine with either way.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
```diff
@@ -0,0 +1,53 @@
...
+class StoredTableNode;
+
+/**
+ * In SQL, the returned columns of an aggregate are either "group-by columns" or "aggregate columns, e.g., SUM(a)".
+ * As an example, in TPC-H Q10, the aggregate on customer includes both c_custkey as well as c_name (amonst others) in
+ * the group-by columns.
+ * Note how c_name is included as a group-by column, even though the groups are already exhaustively defined by
+ * GROUP BY c_custkey. The reason for that is the functional dependency c_custkey => c_name, or, more practically
+ * speaking, the fact that c_custkey is the primary key of the customer table.
+ * This rule identifies cases where functionally dependent columns are given as group-by columns while their primary
```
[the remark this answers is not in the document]
Bensk1: Yes and no. I believe for Hyrise in its current state, this is correct but in general it could be more precise. A data dependency pedant would probably trip over it.

Smart ass excursus on FDs (you might want to skip reading this): A FD X -> Y, where X (determinant) and Y (dependent) are sets of columns is defined as: if two tuples are equal in X, they will also be equal in Y. Duplicates are possible, FDs do not say anything about uniqueness. Therefore, FDs themselves have nothing to do with primary keys.
There are also UniqueColumnCombinations (UCCs) also called candidate keys. UCCs are defined as: a set of columns K is unique iff there are no two tuples which are equal in K. Primary keys are UCCs. A UCC induces trivial FDs. The UCC columns are the determinant (c_custkey in above's case) and all other columns of the table are possible dependents. While UCCs induce FDs, FDs do not guarantee UCCs.

Now it becomes a bit more interesting again: this optimization works not only in the presence of primary keys but also for simple FDs. FDs are sufficient to remove dependent columns. We could rephrase the comment to make it more precise: "This rule identifies cases where functionally dependent columns are given as group-by columns while the FD's determinant column(s) is a/are also group by column(s)." However, currently we only support unique columns as soft constraints and cannot differentiate between weaker FDs and stronger UCCs, right?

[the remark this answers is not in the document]
Bouncner: I don’t have comment at hand on my mobile, but don’t we solely talk about FDs? Uniqueness is not required and would render the rule useless after most joins anyways.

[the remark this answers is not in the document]
Bensk1: Yes, we only need FDs, but the reader might think that primary keys are necessary or that FDs need primary keys even though that's not the case.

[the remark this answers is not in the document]
mrks: Thanks!

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
Bouncner: Is 0070034  ok?

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
Bouncner: Even though is breaks the line length limit we usually use?

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
Bouncner: I changed two tests.
In other tests, I am not comparing the original LQP against an optimised one but against a "new" lqp. That should be unproblematic, right?
But it's probably a dump idea to create the same LQP twice. So I'll change it to a `deep_copy()`'d one.

src/test/operators/aggregate_test.cpp
Bouncner: Not sure what you mean. The test first filters the table for the first column being `123`. For all aggregation values of the remaining tuples, the value is 20.

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
Bouncner: Done.

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
Bouncner: Added `  // NOLINT` as done in other tests.

src/test/operators/aggregate_test.cpp
```diff
@@ -177,6 +182,38 @@ TYPED_TEST(OperatorsAggregateTest, CannotStandardDeviationSampleStringColumns) {
   EXPECT_THROW(aggregate->execute(), std::logic_error);
 }
 
+// The ANY aggregation is a special case which is used to obtain "any value" of a group of which we know that each
+// value in this group is the same (for most cases, the group will have a size of one). This can be the case, when
+// the aggragated column is functionally dependent on the group-by columns.
```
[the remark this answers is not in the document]
Bouncner: Should be correct, @Bensk1?
Typo fixed.

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -57,7 +58,26 @@ std::string AggregateNode::description(const DescriptionMode mode) const {
...
+  _column_expressions.resize(node_expressions.size());
+  std::copy(node_expressions.begin(), node_expressions.end(), _column_expressions.begin());
+
+  for (auto expression_idx = aggregate_expressions_begin_idx; expression_idx < _column_expressions.size();
+       ++expression_idx) {
+    auto& column_expression = _column_expressions[expression_idx];
+    DebugAssert(column_expression->type == ExpressionType::Aggregate,
+                "Unexpected non-aggregate in list of aggregates.");
+    if (column_expression->type == ExpressionType::Aggregate) {
+      const auto aggregate_expression = std::static_pointer_cast<AggregateExpression>(column_expression);
```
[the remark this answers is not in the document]
Bouncner: Done.

src/lib/expression/aggregate_expression.hpp
```diff
@@ -4,7 +4,12 @@
 
 namespace opossum {
 
-enum class AggregateFunction { Min, Max, Sum, Avg, Count, CountDistinct, StandardDeviationSample };
+/**
+ * Supported aggregate functions. In addition to the default SQL functions (e.g., MIN(), MAX()), Hyrise internally uses
+ * the ANY() function, which return any single value. This function is only used by the optimizer in case that the
+ * group to aggregate is either of size one or all values of the group are known to be equal.
```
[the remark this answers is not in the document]
Bouncner: We'd put a reference to something that might be no longer existing at some point in time. Your call.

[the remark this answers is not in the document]
Bouncner: Yeah, that's rather implicit for PKs. Changed to "This function is only used by the optimizer in case that all values of the group are known to be equal."

src/lib/operators/abstract_aggregate_operator.hpp
```diff
@@ -140,6 +140,20 @@ class AggregateFunctionBuilder<ColumnDataType, AggregateType, AggregateFunction:
   }
 };
 
+template <typename ColumnDataType, typename AggregateType>
+class AggregateFunctionBuilder<ColumnDataType, AggregateType, AggregateFunction::Any> {
+ public:
+  auto get_aggregate_function() {
+    return [](const ColumnDataType& new_value, std::optional<AggregateType>& current_primary_aggregate,
+              std::vector<AggregateType>& current_secondary_aggregates) {
+      // ANY() is expected to be only executed on groups whose size is either 1 or whose values are all equal
```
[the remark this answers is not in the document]
Bouncner: Adapted (see comment about implicit PK).

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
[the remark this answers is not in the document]
Bouncner: @Bensk1 : thanks for the clear up. Is the current description ok?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,167 @@
...
+      std::vector<std::pair<size_t, size_t>> constraints_position_and_size;
+      constraints_position_and_size.reserve(table_constraints.size());
+      auto constraint_id = size_t{0};
+      for (const auto& table_constraint : table_constraints) {
+        constraints_position_and_size.emplace_back(constraint_id, table_constraint.columns.size());
+        ++constraint_id;
+      }
+      std::sort(
+          constraints_position_and_size.begin(), constraints_position_and_size.end(),
+          [](const auto& constraint_1, const auto& constraint_2) { return constraint_1.second < constraint_2.second; });
```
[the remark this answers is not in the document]
Bouncner: Changed.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+      // Try to reduce the group-by list one constraint at a time, starting with the shortest constraint. As soon as
+      // one reduction took place, we can ignore the remaining constraints.
```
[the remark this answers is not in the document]
Bouncner: Hmm. Yeah, I think I had a wrong assumption about our constraints here. Will need to think about it.

src/lib/expression/aggregate_expression.hpp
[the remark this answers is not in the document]
mrks: Mmh, I'd risk it.

src/lib/operators/abstract_aggregate_operator.hpp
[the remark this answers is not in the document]
mrks: Can't find that change in https://github.com/hyrise/hyrise/pull/1929/files/d6a32aae2d338b0c3655cf66ab7883c86e7a21e7..54771ff443158150132413f7c9dac45c4b561ee3

src/test/operators/aggregate_test.cpp
mrks: `column_2` is `ANY(input.column_2)`, isn't it?

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
mrks: > Even though is breaks the line length limit we usually use?

Yes

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
```diff
@@ -12,15 +12,15 @@ struct TableConstraintDefinition;
...
+ * In SQL, the returned columns of an aggregate are either "group-by columns" or "aggregate columns", e.g., SUM(a). As
+ * an example, in TPC-H Q10, the aggregate on customer includes both c_custkey as well as c_name (amonst others) in
  * the group-by columns.
  * Note how c_name is included as a group-by column, even though the groups are already exhaustively defined by
  * GROUP BY c_custkey. The reason for that is the functional dependency c_custkey => c_name, or, more practically
  * speaking, the fact that c_custkey is the primary key of the customer table.
- * This rule identifies cases where functionally dependent columns are given as group-by columns while their primary
- * key column(s) is a/are also group by column(s). It then replaces that column (e.g., c_name) with a dummy aggregate
- * function ANY(c_name). This reduces the grouping cost in the aggregate operator.
+ * This rule identifies cases where functionally dependent columns are given as group-by columns while the FD's
```
mrks: Either introduce FD as an abbreviation or write it out as the abbreviation is only used twice

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
```diff
@@ -33,21 +33,18 @@ class StoredTableNode;
  * 		Aggregate: Grouping = [ c_custkey, c_acctbal, c_phone, n_name, c_address, c_comment],
  * 				   Aggregates = [ SUM(l_extendedprice * (1 - l_discount)), ANY(c_name) ]
 
- * ANY() selects "any" value from the list of values per group (since we group by the primary key in this case, the
- * group is ensured to be of size one). This rule implements choke point 1.4 of "TPC-H Analyzed: Hidden Messages and
- * Lessons Learned from an Influential Benchmark" (Boncz et al.). Due to the current lack of foreign key suppport, not
- * all queries that are listed in the paper can be optimized.
+ * ANY() selects "any" value from the list of values per group (since we group by the FD's determinat column(s) in this
```
mrks: typo

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
```diff
@@ -33,21 +33,18 @@ class StoredTableNode;
...
+ * case, the group is ensured to be of size one). This rule implements choke point 1.4 of "TPC-H Analyzed: Hidden
+ * Messages and Lessons Learned from an Influential Benchmark" (Boncz et al.). Due to the current lack of foreign
+ * key suppport, not all queries that are listed in the paper can be optimized.
```
mrks: typo

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  // Every column that is part of the table but not of the primary key/unique constraint is going to be moved from
+  // the group-by list to the list of aggregates wrapped in an ANY().
+  for (const auto& group_by_column : group_by_columns) {
+    if (std::find(constraint_columns.begin(), constraint_columns.end(), group_by_column) != constraint_columns.end()) {
+      // Do not touch primary key/unique constraint columns.
+      continue;
+    }
+
+    // Remove node if it is a column reference and references the correct stored table node. Further, decrement
```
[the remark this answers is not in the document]
Bouncner: Using "given" stored table node as it is a parameter.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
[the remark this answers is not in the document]
Bensk1: yes

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.hpp
```diff
@@ -0,0 +1,50 @@
...
+namespace opossum {
+
+class AbstractLQPNode;
+class AggregateNode;
+struct TableConstraintDefinition;
+class StoredTableNode;
+
+/**
+ * In SQL, the returned columns of an aggregate are either "group-by columns" or "aggregate columns", e.g., SUM(a). As
+ * an example, in TPC-H Q10, the aggregate on customer includes both c_custkey as well as c_name (amonst others) in
```
Bensk1: typo `amonst`

src/test/operators/aggregate_test.cpp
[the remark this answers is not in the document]
Bensk1: Correct.

src/test/operators/aggregate_test.cpp
Bouncner: Yes.

Bouncner:
```
=== Columns
|       a|       b|  ANY(c)|
|     int|   float|     int|
|not null|not null|    null|
=== Chunk 0 ===
|<ValueS>|<ValueS>|<ValueS>|
|     123|   458.7|      20|
```

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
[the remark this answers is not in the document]
Bouncner: Just to be sure that I am not making a mistake here. Can there be a situation where we can gain an advantage when handling multiple constraints?
I thought that the shortest constraint should always be sufficient (not thinking about data types such as strings, which would actually make sense .. hmm) to minimise the group-by list. Then I thought that might not be the case, but then again I think it could still be the case.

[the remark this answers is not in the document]
mrks: I haven't given this too much thought. You are probably correct, but I suggest to add an explanation to the claim that "we can ignore the remaining constraints".

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -0,0 +1,167 @@
...
+        lqp->type != LQPNodeType::Projection) {
+      const auto projection_node = std::make_shared<ProjectionNode>(root_column_expressions);
+      lqp_insert_node(lqp, LQPInputSide::Left, projection_node);
+    }
+
+    return LQPVisitation::VisitInputs;
+  });
+}
+
+bool DependentGroupByReductionRule::reduce_for_constraint(
```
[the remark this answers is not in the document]
Bouncner: Changed to `reduce_group_by_columns_for_constraint` and added the following comment.

```
/**
 * This function reduces the group-by columns of @param aggregate_node for the inputs @param stored_table_node and
 * @param table_constraint. @param group_by_columns is passed as it already stores the ColumnID's which are used for
 * intersection instead of retrieving them from @param aggregate_node again.
 *
 * @returns   Boolean value denoting whether at the group-by list of @param aggregate_node changed.
 */
```

src/test/operators/aggregate_test.cpp
Bouncner: Split the tests.

Bouncner:
Only few open issues left (e.g., handling of multiple constraints).

Besides that, shall I block this PR until we merged #1957?

mrks:
> Besides that, shall I block this PR until we have merged #1957?

I don't expect too many conflicts, and the order should not matter.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -143,12 +150,11 @@ void DependentGroupByReductionRule::apply_to(const std::shared_ptr<AbstractLQPNo
       std::sort(constraints_position_and_size.begin(), constraints_position_and_size.end(),
                 [](const auto& left, const auto& right) { return left.second < right.second; });
 
-      // Try to reduce the group-by list one constraint at a time, starting with the shortest constraint. As soon as
-      // one reduction took place, we can ignore the remaining constraints.
+      // Try to reduce the group-by list one constraint at a time, starting with the shortest constraint.
```
mrks: Now the info that you stop after the first match is missing entirely...

Bouncner:
> > Besides that, shall I block this PR until we have merged #1957?
> 
> I don't expect too many conflicts, and the order should not matter.

I was rather interested in potential performance changes if we bring in #1957 before this one.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
Bouncner:
```
// Try to reduce the group-by list one constraint at a time, starting with the shortest constraint. Multiple
      // constraints (e.g., a primary key and a unique column constraint) might result in different group-by lists
      // (depending on the number of columns of the constraint), but we stop as soon as one constraint successfully
      // reduced the group-by list. The reason is that there is no advantage in a second reduction if the first
      // reduction was successful, because no further columns will be removed. Hence, as soon as one reduction took
      // place, we can ignore the remaining constraints.
```

Ok?

src/lib/operators/abstract_aggregate_operator.hpp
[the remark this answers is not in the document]
Bouncner: It's in now.

PR #1930, opened 2019-11-15
SQL keyword "table" as column name in meta manager

Bouncner:
The meta table manager currently uses `table` as a column name. A student (thanks Caterina) just informed me that queries such as `select table, chunk_id from meta_chunks` can thus not be executed. While a simple `select "table"` is sufficient, I'd say we should better not have SQL keywords as column names.

Further, `column_name` & `column`  and `row_count` & `rows` have been unified.

### DIFF
--- a/resources/test_data/tbl/meta_tables/meta_chunks.tbl
+++ b/resources/test_data/tbl/meta_tables/meta_chunks.tbl
@@ -1,4 +1,4 @@
-table|chunk_id|rows|invalid_rows|cleanup_commit_id
+table_name|chunk_id|row_count|invalid_rows|cleanup_commit_id
 string|int|long|long|long_null
 int_int|0|2|0|null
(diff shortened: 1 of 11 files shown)


mrks:
Technically, it's the sql-parser's fault, but this is certainly the easier way

Bouncner:
Last question: `row_count` over `rows`? I'd say yes. But what about `invalid_rows`. Sound good to be, but `invalid_row_count` is probably better. Any opinions?

mrks:
I don't see any potential for ambiguity. row_count matches the table interface, rows is shorter. Whatever. ¯\\\_(ツ)\_/¯

Bouncner:
Changed to `invalid_row_count` to keep it consistent with Hyrise. I'd rather have a speaking name than typing few charts less.

Issue #1790, opened 2019-08-14
Refactor index scan


mrks:
## Make IndexScan a TableScanImpl
If the optimizer decided to use an index, the TableScan should create two impls, where the first is the one that is currently generated and the second is the one that handles indexes.
- Pro: The type casting (#1688) comes for free
- Pro: The LQPTranslator [does not have to check which chunks have which indexes
](https://github.com/hyrise/hyrise/pull/1785/files#r313746357)
- Pro: The PQPs become nicer
- Pro: We can get rid of the inclusion/exclusion lists
- Pro: Should allow us to recreate indexes

- The TableScan should not decide which chunks use an index, that information should come in the form of a list of ChunkIDs from the rule
- The visualization should print for how many chunks an index was used
- If a chunk has multiple applicable indexes, throw an exception (for now)

# Chunk::get_indexes
- Should be renamed and reflect that work is being done. Maybe "collect"?

# Things to test
- Null handling (including `IS (NOT) NULL`)
- Recreation
- Handling of floats/doubles
- Correctness if a chunk is indexed/added after the PQP was created
- ...

mrks:
@Bensk1 @mweisgut Did I miss something?

Bensk1:
I cannot think of anything. But maybe there is something unnecessary:

> Correctness if a chunk is indexed/added after the PQP was created


Didn't you add a test for that lately?

mrks:
Yes. Still wanted to bring that up, as that is quite important.

Bensk1:
Right!

mweisgut:
> * The TableScan should not decide which chunks use an index, that information should come in the form of a list of ChunkIDs from the rule

In this respect we also have to consider that different indexes can be used for different chunks.

> ## Things to test
> * Null handling (including `IS (NOT) NULL`)
> * Recreation

Recreation of what?

mrks:
See AbstractOperator::deep_copy(). Updated the top post.

Bouncner:
What do we do about multiple predicates which can be answered with a single index scan using the composite group key index?
The whole idea of this issue make sense to me, but I cannot think of a clean way of handling that in the table scan.

Bensk1:
@mrks just came up with the idea to hand conjunctive predicates into the TableScan that then spawns multiple TableScanImpls

PR #1931, opened 2019-11-15
Reference LQP node in PQP

Bouncner:
This is the second discussion PR to reference LQP nodes in PQP nodes during translation.

This PR is am attempt to allow better parsing of the plan cache. One of the main motivations was to retrieve the actual physical column when parsing scans/join etc.

Parsing of PQPs is something that @Bensk1 and I already do often, and what will probably be done much more often in the future (NVM, bachelor's project, ...). This PR is one of several PRs coming which will lay the foundation for the best self-pimping database ever.


src/lib/operators/abstract_operator.hpp
```diff
@@ -161,6 +165,9 @@ class AbstractOperator : public std::enable_shared_from_this<AbstractOperator>,
   // Is nullptr until the operator is executed
   std::shared_ptr<const Table> _output;
 
+  // LQP node with which this operator has been created. Can be uninitialized.
+  std::shared_ptr<const AbstractLQPNode> _lqp_node;
```
mrks: Right now, it probably doesn't make any difference as we have nothing pointing from the LQP to the PQP, but I think I would feel better if this was a weak_ptr

src/lib/operators/abstract_operator.hpp
```diff
@@ -118,6 +118,10 @@ class AbstractOperator : public std::enable_shared_from_this<AbstractOperator>,
   std::shared_ptr<const Table> input_table_left() const;
   std::shared_ptr<const Table> input_table_right() const;
 
+  void set_lqp_node(const std::shared_ptr<const AbstractLQPNode> lqp_node);
+
+  std::shared_ptr<const AbstractLQPNode> get_lqp_node() const;
```
mrks: Any reason we need setters and getters?

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -206,6 +206,7 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_predicate_node_to_in
   // TableScan(s).
   auto index_scan = std::make_shared<IndexScan>(input_operator, SegmentIndexType::GroupKey, column_ids,
                                                 predicate->predicate_condition, right_values, right_values2);
+  index_scan->set_lqp_node(node);
```
mrks: The support in the LQPTranslator looks incomplete. Any reason not to make this a part of `LQPTranslator::translate_node` and add an exception for the index scan (which will be redundant with #1790)?

src/test/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -399,6 +400,58 @@ TEST_F(LQPTranslatorTest, PredicateNodeBetweenScan) {
...
+  // clang-format off
+
+  // Predicates
+  {
+    auto predicate_node = PredicateNode::make(between_inclusive_(int_float_a, 42, 1337), int_float_node);
+    const auto op = LQPTranslator{}.translate_node(predicate_node);
+
+    const auto lqp_node = op->get_lqp_node();
+    const auto recovered_node = std::dynamic_pointer_cast<const PredicateNode>(lqp_node);
+    // TODO(anyone): is this test to compare the input and recovered nodes sufficient?
```
mrks: Should be ok

src/lib/operators/abstract_operator.hpp
Bouncner: Let's discuss in person. I see what you mean, but I am not sure that it helps us here.

src/lib/logical_query_plan/lqp_translator.cpp
Bouncner: The reason is (and I don't know if that's really an issue though) that we do not want to have the overhead of keeping LQP nodes around for each and every single operator even though it is never accessed for later analyses.
Currently, only operators that we analyse for predictions are storing their LQP nodes.

mrks: Having support only for whatever you considered worthwhile for your current work seems inconsistent.

src/lib/operators/abstract_operator.hpp
mrks: For the most part, we have switched to directly modifying the field: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rh-get

src/lib/logical_query_plan/lqp_translator.cpp
Bouncner: Hmm ... looking at how much we "save" right now it not worth the differentiation anyways. Will change.

tobodner:
> This is the second discussion PR to reference LQP nodes in PQP nodes during translation.
> 
> This PR is am attempt to allow better parsing of the plan cache. One of the main motivations was to retrieve the actual physical column when parsing scans/join etc.
> 
> Parsing of PQPs is something that @Bensk1 and I already do often, and what will probably be done much more often in the future (NVM, bachelor's project, ...). This PR is one of several PRs coming which will lay the foundation for the best self-pimping database ever.

Did we also consider passing down the necessary information from LQP nodes to PQP nodes? IIRC, this is what traditional optimisers a la DB2 do.

Bouncner:
> Did we also consider passing down the necessary information from LQP nodes to PQP nodes? IIRC, this is what traditional optimizers a la DB2 do.

Lacking a standardized interface that means first finding one and then adapting all operators.

src/lib/operators/abstract_operator.hpp
Bouncner: Done.

src/lib/logical_query_plan/lqp_translator.cpp
Bouncner: Done.

src/test/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -399,6 +400,57 @@ TEST_F(LQPTranslatorTest, PredicateNodeBetweenScan) {
...
+    auto validate_node = ValidateNode::make(int_float_node);
+    const auto op = LQPTranslator{}.translate_node(validate_node);
+
+    const auto lqp_node = op->lqp_node;
+    const auto recovered_node = std::dynamic_pointer_cast<const ValidateNode>(lqp_node);
+    EXPECT_EQ(recovered_node, validate_node);
+  }
+
+  // clang-format off
+}
```
Bensk1: Hm, I don't have a strong opinion on that, but maybe a test with more than one node instead of a (not complete) list of different operators?

Bouncner: Will do.

src/lib/operators/abstract_operator.hpp
```diff
@@ -124,6 +124,9 @@ class AbstractOperator : public std::enable_shared_from_this<AbstractOperator>,
   // Set parameters (AllParameterVariants or CorrelatedParameterExpressions) to their respective values
   void set_parameters(const std::unordered_map<ParameterID, AllTypeVariant>& parameters);
 
+  // LQP node with which this operator has been created. Might be uninitialized.
+  std::shared_ptr<const AbstractLQPNode> lqp_node;
```
mrks: Ok, here's an idea on this whole circular dependency thing: How about a test where you create a pipeline, retrieve a weak_ptr (!) to one of the LQP nodes, let the pipeline expire, and assert that the weak_ptr is expired?

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -89,7 +89,8 @@ std::shared_ptr<AbstractOperator> LQPTranslator::translate_node(const std::share
     return operator_iter->second;
   }
 
-  const auto pqp = _translate_by_node_type(node->type, node);
+  auto pqp = _translate_by_node_type(node->type, node);
```
mrks: What about IndexScans? I believe those are the only ones without a 1:1 mapping to LQP nodes.

src/test/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -399,6 +400,57 @@ TEST_F(LQPTranslatorTest, PredicateNodeBetweenScan) {
...
+                              int_float_node,
+                              int_float2_node);
+    const auto op = LQPTranslator{}.translate_node(join_node);
+
+    const auto lqp_node = op->lqp_node;
+    const auto recovered_node = std::dynamic_pointer_cast<const JoinNode>(lqp_node);
+    EXPECT_EQ(recovered_node, join_node);
+  }
+
+  // Joins
```
mrks: copy pasta

src/lib/logical_query_plan/lqp_translator.cpp
Bouncner: Depending on the scan type, we create either a table scan or an index scan. But that shouldn't make a difference here, does it?

src/lib/operators/abstract_operator.hpp
Bouncner: I don't understand what this would actually test. :-/

mrks: If we had circular ownerships (leading to memory leaks), the use_count of the shared_ptrs would never reach 0. By holding a weak_ptr, we can check if the LQP and PQP nodes are properly destroyed once their pipeline and cache entries are gone.

src/lib/logical_query_plan/lqp_translator.cpp
mrks: We always create a table scan, have a look at `_translate_predicate_node_to_index_scan`.

src/test/logical_query_plan/lqp_translator_test.cpp
Bouncner: @Bensk1 : can you take another look?

src/lib/operators/abstract_operator.hpp
Bouncner: So I keep the abstract operator as is and just test, that the shared_ptr to the LQP is nicely cleaned up. With this test, which should expectedly break right now, we can later see if there are new references which keep the LQP nodes alive. And thus the expected throw no longer occurs.

Do I get this right?

src/test/logical_query_plan/lqp_translator_test.cpp
Bouncner: Please ignore the unnecessary brackets. Will be removed.

src/test/logical_query_plan/lqp_translator_test.cpp
Bouncner: Changed the test.

src/lib/operators/abstract_operator.hpp
mrks: Why would it break now? The PQP operator should be removed at some point at which the use_count of the LQP becomes 0, cleaning up the LQP node and making the weak_ptr expired. Maybe easier to discuss this in person tomorrow?

mrks: Yepp, that's the idea. I'd just run the entire SQLPipeline so that more potential leaks are covered.

Bouncner: Adapted the test in 21caa7e.

src/lib/logical_query_plan/lqp_translator.cpp
Bouncner: Added LQP node to both scans, added an explaining comment, and a test that checks that all three PQP nodes share the same LQP node.

src/lib/operators/abstract_operator.hpp
Bouncner: I hope to have addressed all your brought up issues.

PR #1932, opened 2019-11-19
Retrieve access frequency of cached elements

Bouncner:
This PR introduces the `frequency()` method to the cache. The GDFS cache (which is the only one used right now) has a nice `F` in the name and we should use that.

The function is not implemented for other cache. On the one hand, they aren't used anyways. On the other hand, as far as I can tell they do not store any frequency information.

### DIFF
--- a/src/lib/cache/abstract_cache_impl.hpp
+++ b/src/lib/cache/abstract_cache_impl.hpp
@@ -66,8 +66,6 @@ class AbstractCacheImpl {
   virtual ErasedIterator begin() = 0;
   virtual ErasedIterator end() = 0;
 
-  size_t frequency(const Key& key) { Fail("Cache implementation does not suppport frequency()."); }
-
   // Return the capacity of the cache.
   size_t capacity() const { return _capacity; }
 
--- a/src/lib/cache/abstract_cache_impl.hpp
+++ b/src/lib/cache/abstract_cache_impl.hpp
@@ -4,8 +4,6 @@
 
 #include <boost/iterator/iterator_facade.hpp>
 
-#include "utils/assert.hpp"
-
 namespace opossum {
 
 // Generic template for a cache implementation.
--- a/src/lib/cache/abstract_cache_impl.hpp
+++ b/src/lib/cache/abstract_cache_impl.hpp
@@ -4,6 +4,8 @@
 
 #include <boost/iterator/iterator_facade.hpp>
 
+#include "utils/assert.hpp"
+
 namespace opossum {
 
 // Generic template for a cache implementation.
@@ -63,6 +65,8 @@ class AbstractCacheImpl {
 
   virtual ErasedIterator begin() = 0;
   virtual ErasedIterator end() = 0;
+
+  size_t frequency(const Key& key) { Fail("Cache implementation does not suppport frequency()."); }
 
   // Return the capacity of the cache.
   size_t capacity() const { return _capacity; }
--- a/src/lib/cache/cache.hpp
+++ b/src/lib/cache/cache.hpp
@@ -78,6 +78,8 @@ class Cache {
 
   Iterator end() { return _impl->end(); }
 
+  size_t frequency(const Key& key) { return _impl->frequency(key); }
+
  protected:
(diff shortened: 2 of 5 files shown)


src/lib/cache/abstract_cache_impl.hpp
```diff
@@ -64,6 +66,8 @@ class AbstractCacheImpl {
   virtual ErasedIterator begin() = 0;
   virtual ErasedIterator end() = 0;
 
+  size_t frequency(const Key& key) { Fail("Cache implementation does not suppport frequency()."); }
```
mrks: If only one cache implementation supports this, why don't we add it just there?

src/lib/cache/cache.hpp
```diff
@@ -78,6 +78,8 @@ class Cache {
 
   Iterator end() { return _impl->end(); }
 
+  size_t frequency(const Key& key) { return _impl->frequency(key); }
```
mrks: doc?

src/lib/cache/abstract_cache_impl.hpp
Bouncner: You mean as a virtual method which is only implemented in GDFS?

src/lib/cache/gdfs_cache.hpp
```diff
@@ -132,6 +132,13 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
 
   ErasedIterator end() { return ErasedIterator{std::make_unique<Iterator>(_map.end())}; }
 
+  size_t frequency(const Key& key) {
+    const auto it = _map.find(key);
```
Bensk1: What if `key` is not part of the `_map`?

Bensk1: Actually, returning `0` would be fine in that case as well

src/lib/cache/cache.hpp
Bouncner: What about:
"Returns the frequency with which a cached item has been accessed (=1 for set(), +1 for each get()). Returns 0 for keys not cached." (second part is Jan's remark, working on it right now).

src/lib/cache/gdfs_cache.hpp
Bouncner: Returning 0 now, added a test.

src/lib/cache/cache.hpp
Bensk1: `access frequency` instead of `with which...`?

Bouncner: "Returns the access frequency of a cached item (=1 for set(), +1 for each get()). Returns 0 for keys not being cache-resident."

?

src/lib/cache/abstract_cache_impl.hpp
mrks: No, as a non-virtual method. Why should it be virtual if there is exactly one implementation.

src/lib/cache/cache.hpp
mrks: works for me

src/lib/cache/abstract_cache_impl.hpp
Bouncner: That would require us to always cast the PQP cache (`using SQLPhysicalPlanCache = Cache<std::shared_ptr<AbstractOperator>, std::string>;`) to GDFS, doesn't it?

mrks: Yes, but I consider that a good thing.

Bouncner: I don't think so, but ok. Will change.

src/lib/cache/cache.hpp
Bouncner: Added the comment.

src/lib/cache/abstract_cache_impl.hpp
Bouncner: Addressed.

Bouncner:
Please wait, I’ll remove a remaining todo.

PR #1934, opened 2019-11-21
SQLite test runner: Print line numbers

mrks:
This has happened one time too often:

```
[ RUN      ] SQLiteTestRunnerEncodings/SQLiteTestRunner.CompareToSQLite/876
script returned exit code 139
```

By making the line number within the .sql file part of the test names, it will be easier to trace this back to the SQL query that caused the segfault:

```
[ RUN      ] SQLiteTestRunnerEncodings/SQLiteTestRunner.CompareToSQLite/Line309WithUnencoded
[       OK ] SQLiteTestRunnerEncodings/SQLiteTestRunner.CompareToSQLite/Line309WithUnencoded (12 ms)
```

### DIFF
--- a/src/test/sql/sqlite_testrunner/sqlite_testrunner.cpp
+++ b/src/test/sql/sqlite_testrunner/sqlite_testrunner.cpp
@@ -46,8 +46,8 @@ void SQLiteTestRunner::SetUpTestCase() {
 
   _table_cache_per_encoding.emplace(EncodingType::Unencoded, unencoded_table_cache);
 
-  opossum::Hyrise::get().topology.use_numa_topology();
-  opossum::Hyrise::get().set_scheduler(std::make_shared<NodeQueueScheduler>());
+  Hyrise::get().topology.use_numa_topology();
+  Hyrise::get().set_scheduler(std::make_shared<NodeQueueScheduler>());
 }
 
 void SQLiteTestRunner::SetUp() {
@@ -137,7 +137,7 @@ std::vector<std::pair<size_t, std::string>> SQLiteTestRunner::queries() {
   std::ifstream file("resources/test_data/sqlite_testrunner_queries.sql");
   std::string query;
 
-  auto next_line = size_t{0};
+  auto next_line = size_t{0};  // Incremented before first use
   while (std::getline(file, query)) {
     ++next_line;
     if (query.empty() || query.substr(0, 2) == "--") continue;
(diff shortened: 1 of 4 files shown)


src/test/sql/sqlite_testrunner/sqlite_testrunner.hpp
```diff
@@ -53,11 +54,20 @@ class SQLiteTestRunner : public BaseTestWithParam<SQLiteTestRunnerParam> {
...
+  // Returns pair of the line in the sql file and the query itself
+  static std::vector<std::pair<size_t, std::string>> queries();
 
   inline static std::unique_ptr<SQLiteWrapper> _sqlite;
   inline static std::map<EncodingType, TableCache> _table_cache_per_encoding;
   inline static std::string _master_table_suffix = "_master_copy";
 };
 
+auto sqlite_testrunner_formatter = [](const ::testing::TestParamInfo<SQLiteTestRunnerParam>& info) {
+  const auto query_pair = std::get<0>(info.param);
```
tobodner: const reference instead of copy construction?

src/test/sql/sqlite_testrunner/sqlite_testrunner.cpp
```diff
@@ -129,26 +129,31 @@ void SQLiteTestRunner::SetUp() {
...
-  static std::vector<std::string> queries;
+std::vector<std::pair<size_t, std::string>> SQLiteTestRunner::queries() {
+  static std::vector<std::pair<size_t, std::string>> queries;
 
   if (!queries.empty()) return queries;
 
   std::ifstream file("resources/test_data/sqlite_testrunner_queries.sql");
   std::string query;
 
+  auto next_line = size_t{0};
```
tobodner: This should really be 1. The printouts are off by 1.

tobodner: > This should really be 1. The printouts are off by 1.

Just double checked, both my vi and IDE start counting at 1.

mrks: It's incremented before it is used. That way, we can get away with just one ++, even in the case of continue

mrks: Will add a comment

tobodner: In line 145, there is a -1 again..

mrks: I fixed that before you wrote this, didn't I?

tobodner: I am not sure, I still get confused by the GitHub PR tool.

PR #1935, opened 2019-11-21
jemalloc: Fix rebuild on change, fix performance regression

### LINKED ISSUE #1677: Hyrise singleton instance

### LINKED ISSUE #1925: jemalloc is not recompiled when the submodule is changed / fix huge_threshold
https://github.com/jemalloc/jemalloc/issues/1677

I have the fix ready, just adding this so that it doesn't get lost.

mrks:
fixes #1925


src/lib/hyrise.cpp
```diff
@@ -1,8 +1,16 @@
 #include "hyrise.hpp"
 
+#include <jemalloc/jemalloc.h>
+
 namespace opossum {
 
 Hyrise::Hyrise() {
+  // Work around for a performance regression in jemalloc:
```
tobodner: *Workaround

src/lib/hyrise.cpp
```diff
@@ -1,8 +1,16 @@
...
+  // https://github.com/jemalloc/jemalloc/issues/1677
+  // Other jemalloc tuning options can be added here as found necessary.
+  auto huge_threshold = 0;
```
tobodner: Did we play around with this value? In the referenced issue, the jemalloc developer states that 40M may be an appropriate value for the benchmark we showed them.

src/lib/hyrise.cpp
mrks: Still WIP, no point in reviewing this yet. This part will have to go, anyway.

src/lib/hyrise.cpp
mrks: Nope. For now, we don't really care about anyone else on the same machine, so giving back memory is not really an issue. I don't really want to open that box in addition to all the other open issues.

mrks:
Unfortunately, that config option is read-only at runtime. As far as I understand jemalloc options, we need to set it before the binary starts. I found no way of setting it during compile time, so I was left with two options:

* Modify the jemalloc header file during configuration time using sed
* fork jemalloc

After trying the first, I found the second to be slightly (!) less ugly

mrks:
Never mind - the old jemalloc was better anyway. I'll switch back to an old commit in the jemalloc repo.

mrks:
Now it's ready for review. I can't check it on my Mac right now, as all my repos are quite dirty. Could you please verify that switching between jemalloc commits rebuilds jemalloc for Hyrise?

tobodner:
> Never mind - the old jemalloc was better anyway. I'll switch back to an old commit in the jemalloc repo.

I too think that this is the cleanest way for now. Can you put a comment somewhere to remind us on the next try to bump the jemalloc version that we need to figure out a way to deal with huge_threshold?

tobodner:
Will give this a close look tomorrow.

mrks:
> I too think that this is the cleanest way for now. Can you put a comment somewhere to remind us on the next try to bump the jemalloc version that we need to figure out a way to deal with huge_threshold?

Not sure where/how to do that. In the end, either the new version is better or it is not. The whole problem this time came because when we looked at it, it was not rebuilt.

Bouncner:
Still, without the comment, the guy who does a well-intended update of all submodules won’t know.

Can we assert the version via CMake?

mrks:
> Still, without the comment, the guy who does a well-intended update of all submodules <s>won’t know.</s> without running the benchmarks should be punished.

I maintain that this was a one time issue caused by jemalloc not being rebuilt.

> Can we assert the version via CMake?

That would be the first and only place where actively guard code from being changed. As everywhere, if you touch something that could have performance implications (or you are unsure), you should run the numbers.

PR #1937, opened 2019-11-23
Add detailed descriptions for LQP nodes and expressions

mrks:
Problem: For **debug prints** (`std::cout << *lqp;`), we currently use `AbstractExpression:as_column_name`. For ambiguous column names, that is not really helpful:

```
(debug)> load resources/test_data/tbl/int_float.tbl t1
(debug)> load resources/test_data/tbl/int_float2.tbl t2
(debug)> SELECT * FROM t1, t2 WHERE t1.a = t2.a
[0] [Projection] a, b, a, b
 \_[1] [Join] Mode: Inner [a = a]
    \_[2] [Validate]
    |  \_[3] [StoredTable] Name: 't2' pruned: 0/1 chunk(s), 0/2 column(s)
    \_[4] [Validate]
       \_[5] [StoredTable] Name: 't1' pruned: 0/1 chunk(s), 0/2 column(s)
```

With this PR, we include the addresses of the source nodes:
<pre>
[0] [Projection] <b>0x12d6b2298</b>.a, 0x12d6b2298.b, 0x12d6b23d8.a, 0x12d6b23d8.b @ 0x12d6fe338
 \_[1] [Join] Mode: Inner [0x12d6b2298.a = 0x12d6b23d8.a] @ 0x12d6e0918
    \_[2] [Validate] @ 0x12d6fe158
    |  \_[3] [StoredTable] Name: 't2' pruned: 0/1 chunk(s), 0/2 column(s) @ 0x12d6b23d8
    \_[4] [Validate] @ 0x12d6fe018
       \_[5] [StoredTable] Name: 't1' pruned: 0/1 chunk(s), 0/2 column(s) @ <b>0x12d6b2298</b>
</pre>

This does not apply to the visualization, which is unchanged.

With this change, you can "easily" find the associated StoredTableNode. This will become even more important once my deduplication PR comes in where LQPColumnReferences (have to) contain disambiguation information:

<pre>
(debug)> load resources/test_data/tbl/int_float.tbl t
(debug)> SELECT t1.a + t2.a + t3.a FROM t t1, t t2, t t3 WHERE t1.a = 123 AND t2.a = 1234 AND t3.a = 12345
[0] [LogicalPlanRootNode] @ 0x125765c98
 \_[1] [Projection] ("a from 0x125666f58 via 0x1260a86d8(left)" + "<b>a from 0x125666f58 via 0x1260a8558(left) via 0x1260a86d8(right)</b>") + "a from 0x125666f58 via 0x1260a8558(right)" @ 0x1257683f8
    \_[2] [Join] Mode: Cross @ <b>0x1260a86d8</b>
       \_[3] [Validate] @ 0x125768358
       |  \_[4] [Predicate] "a from 0x125666f58" = 123 @ 0x1260a8318
       |     \_[5] [StoredTable] Name: 't' pruned: 0/1 chunk(s), 1/2 column(s) @ <b>0x125666f58</b>
       \_[6] [Join] Mode: Cross @ <b>0x1260a8558</b>
          \_[7] [Validate] @ 0x125768498
          |  \_[8] [Predicate] "a from 0x125666f58" = 1234 @ 0x1260a83d8
          |     \_Recurring Node --> [5]
          \_[9] [Validate] @ 0x125768538
             \_[10] [Predicate] "a from 0x125666f58" = 12345 @ 0x1260a8498
                \_Recurring Node --> [5]
</pre>

Also, harmonizes _on_shallow_copy, _on_shallow_equals and _on_shallow_hash (previously _shallow_hash).

### DIFF
--- a/src/lib/expression/abstract_expression.hpp
+++ b/src/lib/expression/abstract_expression.hpp
@@ -65,9 +65,19 @@ class AbstractExpression : public std::enable_shared_from_this<AbstractExpressio
   virtual std::shared_ptr<AbstractExpression> deep_copy() const = 0;
 
   /**
-   * @return a human readable string representing the Expression that can be used as a column name
+   * @return the expression's column name or, optionally, a more detailed description of the expression
    */
-  virtual std::string as_column_name() const = 0;
+  enum class DescriptionMode {
+    ColumnName,  // returns only the column name (e.g., name)
+    Detailed     // additionally includes the address of referenced nodes
+  };
+  virtual std::string description(const DescriptionMode mode = DescriptionMode::Detailed) const = 0;
+
+  /**
+   * @return a human readable string representing the Expression that can be used as a column name
+   *         (shortcut for description(DescriptionMode::ColumnName))
+   */
+  std::string as_column_name() const;
 
   /**
    * @return the DataType of the result of the expression
@@ -105,7 +115,7 @@ class AbstractExpression : public std::enable_shared_from_this<AbstractExpressio
   virtual bool _on_is_nullable_on_lqp(const AbstractLQPNode& lqp) const;
 
   /**
-   * Used internally in _enclose_argument_as_column_name() to put parentheses around expression arguments if they have a lower
+   * Used internally in _enclose_argument to put parentheses around expression arguments if they have a lower
    * precedence than the expression itself.
    * Lower precedence indicates tighter binding, compare https://en.cppreference.com/w/cpp/language/operator_precedence
    *
@@ -114,15 +124,15 @@ class AbstractExpression : public std::enable_shared_from_this<AbstractExpressio
   virtual ExpressionPrecedence _precedence() const;
 
   /**
-   * @return    argument.as_column_name(), enclosed by parentheses if the argument precedence is lower than
+   * @return    argument.description(mode), enclosed by parentheses if the argument precedence is lower than
    *            this->_precedence()
    */
-  std::string _enclose_argument_as_column_name(const AbstractExpression& argument) const;
+  std::string _enclose_argument(const AbstractExpression& argument, const DescriptionMode mode) const;
 };
 
 // So that google test, e.g., prints readable error messages
 inline std::ostream& operator<<(std::ostream& stream, const AbstractExpression& expression) {
-  return stream << expression.as_column_name();
+  return stream << expression.description();
 }
 
 // Wrapper around expression->hash(), to enable hash based containers containing std::shared_ptr<AbstractExpression>
--- a/src/lib/expression/abstract_expression.hpp
+++ b/src/lib/expression/abstract_expression.hpp
@@ -68,7 +68,7 @@ class AbstractExpression : public std::enable_shared_from_this<AbstractExpressio
    * @return the expression's column name or, optionally, a more detailed description of the expression
    */
   enum class DescriptionMode {
-    ColumnName,  // returns only the column name (e.g., name)
+    ColumnName,  // returns only the column name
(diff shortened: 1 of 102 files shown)


tobodner:
Just wondering: Instead of the inline table addresses, could we also inline table names and list the table to address mapping somehow else? Maybe, this would look less bloated.

mrks:
Doesn't help if a table is used twice

tobodner:
> Doesn't help if a table is used twice

Got it. Did not think of it looking at above example.

Bensk1:
Could you ping me again as soon as CI is green?

mrks:
I just have to add `static` in one line. Haven't done it yet because I am in the middle of something bigger and don't want to switch branches. Won't affect the review, though.

mrks:
(green now)

src/lib/expression/abstract_expression.hpp
```diff
@@ -64,10 +64,20 @@ class AbstractExpression : public std::enable_shared_from_this<AbstractExpressio
 
   virtual std::shared_ptr<AbstractExpression> deep_copy() const = 0;
 
+  /**
+   * @return the expression's column name or, optionally, a more detailed description of the expression
+   */
+  enum class DescriptionMode {
+    ColumnName,  // returns only the column name (e.g., name)
```
Bensk1: What's the meaning of the `e.g., name`?

src/lib/logical_query_plan/drop_view_node.cpp
```diff
@@ -14,9 +14,9 @@ namespace opossum {
 DropViewNode::DropViewNode(const std::string& view_name, const bool if_exists)
     : BaseNonQueryNode(LQPNodeType::DropView), view_name(view_name), if_exists(if_exists) {}
 
-std::string DropViewNode::description() const { return "[Drop] View: '"s + view_name + "'"; }
+std::string DropViewNode::description(const DescriptionMode mode) const { return "[Drop] View: '"s + view_name + "'"; }
```
Bensk1: Even though it's not your code, mixing this `"bla"s` with the `std::string("bla")` triggers my neuroses.

src/test/logical_query_plan/create_prepared_plan_node_test.cpp
```diff
@@ -21,10 +22,11 @@ class CreatePreparedPlanNodeTest : public ::testing::Test {
 };
 
 TEST_F(CreatePreparedPlanNodeTest, Description) {
-  EXPECT_EQ(create_prepared_plan_node->description(),
+  // Use short description mode as addresses are non-deterministic
```
Bensk1: This comment sounds to me like: let's use the short description mode here so that addresses are not part of the output because they are non-deterministic and we cannot test them. Then, I see the dummy address a few lines below and I am confused.

src/lib/expression/abstract_expression.hpp
mrks: Mmmh, "name" is a bad example for a column name. :D We don't really need an example here, anyway.

src/lib/logical_query_plan/drop_view_node.cpp
mrks: Doesn't get mixed in this file. `std::string_literals` is used in 23 files and I don't really feel like touching all of them.

src/test/logical_query_plan/create_prepared_plan_node_test.cpp
mrks: As `operator<<` for AbstractLQPNode always uses the detailed mode, this does not actually make a difference here. Fixed.

PR #1939, opened 2019-11-25
Add encoding tests and improve FoR encoding

### LINKED ISSUE #1928: Missing RunLengthSegment and FrameOfReferenceSegment tests
I can't find anything that actually tests those segments. For funsies, I removed the part in run_length_encoder.hpp that identifies a continuing run and nothing broke.

While the generic encoding tests make sure that data does not get lost, we should still make sure that the encoding serves its purpose.

Also, I find this part in run_length_segment.hpp confusing:

> As in value segments, null values are represented as an additional boolean vector.

I would remove that sentence and change a sentence a couple of lines above (including a typo fix):

> Each run is represented by the value and the number of **occurrences. Runs can also be NULL, in which case `values` is undefined (but likely `T{}`) and `null_values` is set to true for that run.**

Bouncner:
As brought up in #1928, certain encoding lack tests of their internal storage. While their usage is pretty well covered via various iterator tests, we do not explicitly test that RLE compressed runs of repeated values etc.

This PR adds a few tests to check the internal data structures of RLE and FoR segments.

Furthermore, the tests brought up that NULL values in FrameOfReference segments led to block_minima of 0, no matter how large the values are (because the value vectors stored `T{0u}` for NULLs). Now, we added an additional step that replaces NULL values with the block minimum and hence the block minima are set as expected (which potentially reduces the compressed offset vector).

Fixes #1928.


src/test/storage/encoded_segment_test.cpp
```diff
@@ -303,4 +303,187 @@ TEST_F(EncodedSegmentTest, SegmentReencoding) {
...
+  }
+
+  const auto value_segment = std::make_shared<ValueSegment<int32_t>>(std::move(values));
+  const auto encoded_segment =
+      this->encode_segment(value_segment, DataType::Int, SegmentEncodingSpec{EncodingType::RunLength});
+
+  const auto run_length_segment = std::dynamic_pointer_cast<const RunLengthSegment<int32_t>>(encoded_segment);
+
+  ASSERT_TRUE(run_length_segment);
+  EXPECT_EQ(run_length_segment->values()->size(), 10 + 8 + 10);
```
dey4ss: Maybe this sum calculation would be more self-containing when doing something like:
``` c++
const auto no_run_values = 10;
const auto run_values = 8; 
for(auto value = 0; value < num_run_values; ...) { 
... 
EXPECT_EQ(run_length_segment->values()->size(), 2 * no_run_values + run_values);
```

src/test/storage/encoded_segment_test.cpp
```diff
@@ -303,4 +303,187 @@ TEST_F(EncodedSegmentTest, SegmentReencoding) {
...
+      this->encode_segment(value_segment, DataType::Int, SegmentEncodingSpec{EncodingType::RunLength});
+
+  const auto run_length_segment = std::dynamic_pointer_cast<const RunLengthSegment<int32_t>>(encoded_segment);
+  ASSERT_TRUE(run_length_segment);
+
+  // The handling of NULL values is something unintuitive. When two runs of values are both NULL, they are merged into
+  // a single run, only keeping the first value of both runs in values() as a placeholder for the merged NULL run.
+  // That means the runs of values 3/4 and the single-value runs of 95/96 are stored as two value runs 3 and 95 in the
+  // values() vector (thus, we substract two from the number of expected runs).
+  EXPECT_EQ(run_length_segment->values()->size(), 9 + 10 - 2);
```
dey4ss: sum calculation as above

src/test/storage/encoded_segment_test.cpp
```diff
@@ -303,4 +303,187 @@ TEST_F(EncodedSegmentTest, SegmentReencoding) {
...
+  EXPECT_EQ(for_segment->null_values().size(), row_count);
+
+  EXPECT_EQ(for_segment->null_values()[0], false);
+  EXPECT_EQ(for_segment->null_values()[1], true);
+  EXPECT_EQ(for_segment->null_values()[7], true);
+  EXPECT_EQ(for_segment->null_values()[16], false);
+
+  // Block minium should be the smallest value: 0
+  EXPECT_EQ(for_segment->block_minima().front(), minimum);
+  resolve_compressed_vector_type(for_segment->offset_values(), [&](const auto& offset_values) {
```
dey4ss: Does this method fail if the vector is not compressed? I.e. do we make sure we use vector compression since FoR encoding would not give use benefits otherwise?

src/test/storage/encoded_segment_test.cpp
```diff
@@ -303,4 +303,187 @@ TEST_F(EncodedSegmentTest, SegmentReencoding) {
...
+  ASSERT_TRUE(run_length_segment);
+  EXPECT_EQ(run_length_segment->values()->size(), 10 + 8 + 10);
+  EXPECT_EQ(run_length_segment->end_positions()->size(), 10 + 8 + 10);
+  EXPECT_EQ(run_length_segment->end_positions()->front(), 0ul);
+  EXPECT_EQ(run_length_segment->end_positions()->back(), 99ul);
+
+  // first longer run has the value 10, starts at position 10, and ends at position 19
+  EXPECT_EQ(run_length_segment->values()->at(10), 0);
+  EXPECT_EQ(run_length_segment->end_positions()->at(9), 9);
+  EXPECT_EQ(run_length_segment->end_positions()->at(10), 19);
```
dey4ss: Just like above, sth. like `9 + num_repetitions` could be clearer. Anyways, I do not know if that's necessary to understand the test.

mrks:
> Should I implement them or would you like to

Generally, we try to keep implementer and reviewer separate. If you implemented something here, your review would not be independent any more and we would have to find a second reviewr.

dey4ss:
> Generally, we try to keep implementer and reviewer separate. If you implemented something here, your review would not be independent any more and we would have to find a second reviewr.

Yes, this absolutely makes sense.

src/test/storage/encoded_segment_test.cpp
```diff
@@ -303,4 +303,187 @@ TEST_F(EncodedSegmentTest, SegmentReencoding) {
...
+
+  const auto value_segment = std::make_shared<ValueSegment<int32_t>>(std::move(values), std::move(null_values));
+  const auto encoded_segment =
+      this->encode_segment(value_segment, DataType::Int, SegmentEncodingSpec{EncodingType::RunLength});
+
+  const auto run_length_segment = std::dynamic_pointer_cast<const RunLengthSegment<int32_t>>(encoded_segment);
+  ASSERT_TRUE(run_length_segment);
+
+  // The handling of NULL values is something unintuitive. When two runs of values are both NULL, they are merged into
+  // a single run, only keeping the first value of both runs in values() as a placeholder for the merged NULL run.
```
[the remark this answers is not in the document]
Bouncner: Added to both.

Can you check if they help?

src/test/storage/encoded_segment_test.cpp
Bouncner: Yes, FoR without vector compression would mean that the offsets are uncompressed and that would lead to a netto gain (due to the additional minima vector). So FoR requires vector compression.

Bouncner: The call should not fail since `offset_values` is of type `BaseCompressedVector`.

src/test/storage/encoded_segment_test.cpp
Bouncner: Changed. Please take a look.

Issue #1933, opened 2019-11-19
SQL COPY command


tobodner:
Our new server implementation lacks support for
- loading/importing data from files into tables
- unloading/exporting data from tables into files

In order to facilitate experiments and benchmarks, we should support the SQL COPY command:
- https://www.postgresql.org/docs/current/sql-copy.html
- We may want to support more file formats than vanilla PostgreSQL, like Snowflake (https://docs.snowflake.net/manuals/sql-reference/sql/copy-into-table.html)

Bensk1:
Because we need this functionality soon for the bachelor's project, @dey4ss is going to work on it. The SQL standard of 1992 does neither contain a COPY or IMPORT command. Currently, the SQL parser is able to handle IMPORT but being consistent with PostgreSQL wouldn't be too bad.

tobodner:
> Because we need this functionality soon for the bachelor's project, @dey4ss is going to work on it. The SQL standard of 1992 does neither contain a COPY or IMPORT command. Currently, the SQL parser is able to handle IMPORT but being consistent with PostgreSQL wouldn't be too bad.

Since the server supports Postgre's wire protocol, it seems most consistent to also align with Postgre's (derivative) SQL features.

dey4ss:
Supporting the `COPY` command would require its implementation in the SQLParser. Therefore, I suggest to first add the general file import first by adding an LQPNode for it. Afterward, the specific interface could be adapted. Anyhow, the parser would need a little work on it since it is only aware of `.tbl` and `.csv` files.

dey4ss:
The parser should be able to understand the COPY statement with my PR there (unfortunately I cannot request a review there). On the Hyrise side, I am currently after a (hopefully last) little bug, then imports will be supported. Afterward, I am going to add exporting functionality there.

Bensk1:
Nice! Would it be a lot of effort to move the branch to our repo and file a new PR?

dey4ss:
> 
> 
> Nice! Would it be a lot of effort to move the branch to our repo and file a new PR?

I think this would not be a problem I had write access to the parser repository :D

Bensk1:
Ah, I thought this was for the hyrise/hyrise repo and not for the parser. Let's keep it like this for now.

dey4ss:
[...]

PR #1941, opened 2019-11-27
Make import an operator

dey4ss:
As requested in #1933, the server should be able to receive a `COPY FROM ...` statement to load a table from a file. Without changing the SQLParser first, I suggest making use of its current `IMPORT` statement. The statement can then be added to the parser later on, without necessarily having to update code here.

The statement the parser accepts right now looks like this: `IMPORT FROM CSV FILE 'file_path/file.csv' INTO table_name;`

I suggest handling the `IMPORT` statement with a new LQPNode that gets translated to an import operator. Binary and CSV import have already been done with own operators, so I think that would make sense. This leads to some discussable points:

1. Currently, we have operators for binary and CSV import. Creating tables from `.tbl` files is located in `utils/load_table.*pp`. Maybe an abstract base import operator (and binary, CSV and tbl implementations) would structure this more nicely. Moving tbl import to an operator would anyhow be required. This would mean some refactoring.
2. At that point, we cannot make use of the `ImportStatement.type` provided by the parser, as it does not have a type for binary files yet. I would just distinguish the file type by its path, which later would also match with the Postgres command. In the meantime, this would make providing the file type in the `IMPORT` statement irrelevant but necessary to be valid (which is not cool).

If this is the way to go, implementing the import should hopefully be straightforward and we can continue with adding the `COPY` statement to the parser. Supporting all its options would require a bunch of features to be added to the import at that time, but should maybe be discussed later on [here](https://github.com/hyrise/hyrise/issues/1933).


mrks:
Haven't looked at the code yet.

> Currently, we have operators for binary and CSV import. Creating tables from .tbl files is located in utils/load_table.*pp. Maybe an abstract base import operator (and binary, CSV and tbl implementations) would structure this more nicely. Moving tbl import to an operator would anyhow be required. This would mean some refactoring.

load_table / tbl files being separate is by design. Those are used mainly in tests to provide the "expected" table. As such, the method to load them should be as easy as possible and not involve, e.g., the operator subsystem. The goal is to have a know-to-be-good implementation.

> At that point, we cannot make use of the ImportStatement.type provided by the parser, as it does not have a type for binary files yet. I would just distinguish the file type by its path, which later would also match with the Postgres command. In the meantime, this would make providing the file type in the IMPORT statement irrelevant but necessary to be valid (which is not cool).

This should not take more than half an hour. If you want, we can take a look at that together. Maybe Friday?

mrks:
Code looks good so far. I would expect the ImportNode to feed into a CreateTableNode.

dey4ss:
I haven't done a lot of coding yet as I wanted to see if got things right first.

> load_table / tbl files being separate is by design. Those are used mainly in tests to provide the "expected" table. As such, the method to load them should be as easy as possible and not involve, e.g., the operator subsystem. The goal is to have a know-to-be-good implementation.

Okay, this makes it even easier. Wrapping the load_table in an operator should be fine then, right?

> This should not take more than half an hour. If you want, we can take a look at that together. Maybe Friday?

Perfect 👍

mrks:
> Wrapping the load_table in an operator should be fine then

Yes. In the long term, I would like to see the (Export|Import)(Csv|Binary) operators to disappear in favor of a single export and an import operator that have separate backends. That feels cleaner than having the LQPTranslator decide which operator has to be used. Also pinging @tobodner who is thinking about adding more data types.

mrks:
Fun fact: This being an LQP node even allows us to do things like "IMPORT ... WHERE ..." if we ever want to.

dey4ss:
That's exactly what Postgres supports with the `COPY` statement :D

dey4ss:
A table name is currently optional for the import operators. Since both Postgres COPY and the current IMPORT require a table name, should it be non-optional?

mrks:
I wouldn't say so. In the console, `load t.tbl` is handy

mrks:
Before you request a review from someone, could you mark the code blocks that were moved 1:1?

src/test/lib/import_export/binary/binary_writer_test.cpp
```diff
@@ -1,471 +1,408 @@
-#include <cstdio>
```
dey4ss: 1:1 copy, only replaced operator execution with static method call

src/test/lib/import_export/csv/csv_parser_test.cpp
```diff
@@ -0,0 +1,275 @@
+#include "base_test.hpp"
```
dey4ss: 1:1 copy, only replaced operator execution with static method call and removed the test for the correct job execution (see CsvParser for details).

src/test/lib/import_export/csv/csv_writer_test.cpp
```diff
@@ -1,267 +1,233 @@
-#include <cstdio>
```
dey4ss: 1:1 copy, only replaced operator execution with static method call

src/test/operators/export_test.cpp
```diff
@@ -0,0 +1,162 @@
+#include <cstdio>
```
dey4ss: copied test class methods

src/test/operators/import_test.cpp
```diff
@@ -0,0 +1,151 @@
+#include <memory>
```
dey4ss: 1:1 copy, only removed scheduling test (see CsvParser for details).

src/lib/operators/import.cpp
```diff
@@ -0,0 +1,82 @@
...
+std::shared_ptr<Table> Import::_import() {
+  switch (_type) {
+    case FileType::Csv:
+      return CsvParser::parse(_filename, _chunk_size, _csv_meta);
+    case FileType::Tbl:
+      return load_table(_filename, _chunk_size);
+    case FileType::Binary:
+      return BinaryParser::parse(_filename);
+    case FileType::Auto:
+      return _import_any_type();
```
mrks: Same as with Export: `if (_type == FileType::Auto) _type = _determine_file_type(...)?` By moving that into `file_type.hpp`, we can deduplicate the code.

src/lib/operators/import.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+const std::string& Import::name() const {
+  static const auto name = std::string{"Import"};
+  return name;
+}
+
+std::shared_ptr<const Table> Import::_on_execute() {
+  if (_tablename && Hyrise::get().storage_manager.has_table(*_tablename)) {
+    return Hyrise::get().storage_manager.get_table(*_tablename);
+  }
```
mrks: Is this really what a user would expect?

src/lib/operators/import.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+  // Check if file exists before giving it to the parser
+  std::ifstream file(_filename);
+  Assert(file.is_open(), "Import: Could not find file " + _filename);
+  file.close();
+
+  const auto table = _import();
```
mrks: Is `_import` called anywhere else? Otherwise, I'd inline it here so that the reader does not have to jump around

src/lib/import_export/csv/csv_writer.hpp
```diff
@@ -0,0 +1,86 @@
...
+   *  "ColumnType","c","float"
+   *
+   *  which resembles the following table of meta data:
+   *
+   *  PropertyType  | Key | Value
+   *  ------------------------------
+   *  ChunkSize     |     | 100
+   *  ColumnType    |  a  | int_null
+   *  ColumnType    |  b  | string
+   *  ColumnType    |  c  | float
```
mrks: Not your fault, but this is horribly outdated, isn't it?

src/lib/import_export/csv/csv_writer.cpp
```diff
@@ -0,0 +1,126 @@
...
+  // Open file for writing
+
+  /**
+   * Multiple rows containing the values of each respective row are written.
+   * Therefore we first iterate through the chunks, then through the rows
+   * in the chunks and afterwards through the segments of the chunks.
+   *
+   * This is a lot of iterating, but to convert a column-based table to
+   * a row-based representation takes some effort.
+   */
```
mrks: Shouldn't this be below the next block?

src/lib/import_export/csv/csv_writer.cpp
```diff
@@ -0,0 +1,126 @@
...
+    _write_string_value(boost::get<pmr_string>(value), ofstream, config);
+    return;
+  }
+
+  ofstream << value;
+}
+
+void CsvWriter::_write_string_value(const pmr_string& value, std::ofstream& ofstream, const ParseConfig& config) {
+  /**
+   * We put an the quotechars around any string value by default
```
mrks: typo

src/lib/import_export/csv/csv_parser.cpp
```diff
@@ -71,32 +72,29 @@ std::shared_ptr<Table> CsvParser::parse(const std::string& filename, const std::
...
+  csvfile.close();
 
   return table;
 }
 
 std::shared_ptr<Table> CsvParser::create_table_from_meta_file(const std::string& filename,
                                                               const ChunkOffset chunk_size) {
-  _meta = process_csv_meta_file(filename);
-  return _create_table_from_meta(chunk_size);
+  auto meta = process_csv_meta_file(filename);
```
mrks: const?

src/lib/import_export/csv/csv_parser.cpp
```diff
@@ -71,32 +72,29 @@ std::shared_ptr<Table> CsvParser::parse(const std::string& filename, const std::
     content_view = content_view.substr(field_ends.back() + 1);
 
     // create and start parsing task to fill chunk
-    tasks.emplace_back(std::make_shared<JobTask>([this, relevant_content, field_ends, &table, &segments]() {
-      _parse_into_chunk(relevant_content, field_ends, *table, segments);
-    }));
-    tasks.back()->schedule();
-  }
 
-  Hyrise::get().scheduler()->wait_for_tasks(tasks);
```
mrks: I rather comment here so that the review is self-contained:

Could you clarify how this made testing difficult? The way I see it, the tasks do not leave the method, so how does multithreading have an externally visible effect?

src/lib/operators/import_csv.hpp
```diff
@@ -1,58 +0,0 @@
...
-/**
- * Creates a Table with values of the parsed csv file <filename>
- * and the corresponding meta file <filename>.meta
- * For the structure of the meta csv file see export_csv.hpp
- *
- * If the parameter `tablename` is provided, the imported table is stored in the StorageManager.
- * If a table with this name already exists, it is returned and no import is performed.
- *
- * TODO(mjendruk): is this still true?
- * Note: ImportCsv does not support null values at the moment
```
mrks: Wasn't this fixed a while ago? If yes, please make sure that the comment is no longer there.

src/lib/operators/import.cpp
dey4ss: It was surprising for me, but I kept this since I thought this was intended. I think as we now can get a table directly by calling a parser, we could make the table name non-optional and always add it to the storage manager if there is no real reason for not doing so.

src/lib/operators/import.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+  if (_tablename) {
+    Hyrise::get().storage_manager.add_table(*_tablename, table);
+  }
+
+  return table;
```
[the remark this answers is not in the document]
dey4ss: I guess this is kind of related to the adding to storage manager thing. By letting the operator have an output, one could just read a table without any side effects in the past. Since we now have parsers we can call instead, there is no real reason for returning the table anymore.

src/lib/operators/import.cpp
mrks: I like not having to type more than I have to and for the table name being optional. My question was more about: "Would a user be surprised if they submit an import statement to be given an old, potentially completely unrelated table?"

dey4ss: At least for me, this was surprising, but I thought this was intended. Anyways, if we do not return the table anymore to be in line with column_expressions, I would just add the table to the storage manager. The problem then would be that actual nothing will happen when no tablename is provided. As a tablename is always given from Hyrise interfaces (import statement needs a tablename, console uses filename if necessary), making it mandatory would not change anything there.

src/lib/import_export/file_type.cpp
```diff
@@ -4,15 +4,26 @@
...
+}
+
+FileType file_type_from_filename(const std::string& filename) {
+  const auto extension = std::string{std::filesystem::path{filename}.extension()};
+  if (extension == ".csv") {
+    return FileType::Csv;
+  } else if (extension == ".tbl") {
+    return FileType::Tbl;
+  } else if (extension == ".bin") {
+    return FileType::Binary;
```
mrks: use std::tolower?

mrks:
```
[ RUN      ] CsvParserTest.WithScheduler
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
/Users/hyrise-ci/.jenkins/workspace/hyrise_hyrise_PR-1941@tmp/durable-60fa07a7/script.sh: line 1: 89517 Abort trap: 6           ./clang-debug/hyriseTest
```

Haven't seen this one before; looks new.

dey4ss:
> 
> 
> ```
> [ RUN      ] CsvParserTest.WithScheduler
> libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
> /Users/hyrise-ci/.jenkins/workspace/hyrise_hyrise_PR-1941@tmp/durable-60fa07a7/script.sh: line 1: 89517 Abort trap: 6           ./clang-debug/hyriseTest
> ```
> 
> Haven't seen this one before; looks new.

I think this is probably due to the parsing now being static; I'm gonna try to reproduce this on a Mac and fix it on Wednesday.

dey4ss:
> ```
> [ RUN      ] CsvParserTest.WithScheduler
> libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
> /Users/hyrise-ci/.jenkins/workspace/hyrise_hyrise_PR-1941@tmp/durable-60fa07a7/script.sh: line 1: 89517 Abort trap: 6           ./clang-debug/hyriseTest
> ```
> 
> 
> Haven't seen this one before; looks new.

Okay, the problem seems to be that MacOS is very strict regarding locks and process boundaries. It gets fixed when the lock is a member variable of the parser. All functions need to be non-static then. As I guess removing the scheduled approach would have performance impacts especially for larger CSVs, the ::parse call would be non-static then. Is it better to align the interface with BinaryParser and make its ::parse method also non-static (though it could be static)?

mrks:
If you are at the office, could you stop by for a second?

PR #1944, opened 2019-12-03
Teach Aggregate to use AggregateExpressions

mrks:
Currently, we use AggregateColumnDefinitions, comprised of ColumnIDs and AggregateFunctions to define the aggregate columns. By switching to AggregateExpressions, we
* reduce code duplication (e.g., column name generation)
* pave the way for the aggregate operators to evaluate arithmetic operations in-place


mrks:
Regarding our discussion earlier today.  The second reason not to touch group by ids (which i forgot) was that it will come naturally once we change the soft constraints from table- to LQP-based. Saves us some conflicts for now.

Bouncner:
I don't understand how the group-by IDs come "naturally" with the upcoming LQP change?

mrks:
> I don't understand how the group-by IDs come "naturally" with the upcoming LQP change?

Never mind.

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -351,34 +351,24 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_aggregate_node(
...
-    const auto column_expression = dynamic_cast<const LQPColumnExpression*>(&*aggregate_expression->argument());
-    auto argument_column_id = ColumnID{};
-    if (column_expression && column_expression->column_reference.original_column_id() == INVALID_COLUMN_ID) {
-      // Handle COUNT(*)
-      argument_column_id = INVALID_COLUMN_ID;
-    } else {
-      argument_column_id = node->left_input()->get_column_id(*aggregate_expression->argument());
-    }
-    aggregate_column_definitions.emplace_back(argument_column_id, aggregate_expression->aggregate_function);
+    Assert(lqp_expression->type == ExpressionType::Aggregate,
```
Bouncner: Shouldn't a `DebugAssert` suffice here?

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -351,34 +351,24 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_aggregate_node(
...
+               "' used as AggregateExpression is not an AggregateExpression");
+
+    const auto pqp_expression = _translate_expression(lqp_expression, node->left_input());
+    const auto aggregate_expression = std::static_pointer_cast<AggregateExpression>(pqp_expression);
+    pqp_aggregate_expressions.emplace_back(aggregate_expression);
   }
 
-  // Create GroupByColumns from the GroupBy expressions
+  // Create GroupByColumns from the GroupBy expressions. For now, we expect all GroupBy expressions to be already
+  // present, i.e., we do not calculate them on the fly
```
Bouncner: Trailing dot.

src/lib/operators/abstract_aggregate_operator.cpp
```diff
@@ -60,13 +56,19 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
...
+    const auto column_id = pqp_column->column_id;
+    if (column_id == INVALID_COLUMN_ID) {
+      Assert(aggregate->aggregate_function == AggregateFunction::Count, "Aggregate: Asterisk is only valid with COUNT");
     } else {
-      DebugAssert(aggregate.column < input_table->column_count(), "Aggregate column index out of bounds");
-      Assert(input_table->column_data_type(aggregate.column) != DataType::String ||
-                 (aggregate.function != AggregateFunction::Sum && aggregate.function != AggregateFunction::Avg &&
-                  aggregate.function != AggregateFunction::StandardDeviationSample),
+      DebugAssert(column_id < input_table->column_count(), "Aggregate column index out of bounds");
+      DebugAssert(pqp_column->data_type() == input_table->column_data_type(column_id), "Mismatching column_data_type");
```
Bouncner: Can't it happen that we chose a larger numeric type when building sums?

src/lib/logical_query_plan/lqp_translator.cpp
mrks: The LQPTranslator currently only uses Assert. As those are the asserts that come up regularly when I break things and as the translator cost is quite low, I'd keep it that way.

src/lib/operators/abstract_aggregate_operator.cpp
mrks: This is about the input column, not the output column

PR #1947, opened 2019-12-04
Fix end() for segment iterables

j-tr:
the _offset member variable of the LZ4Iterator should not always be initialized with 0 but with the actual offset of the passed data_it. This leads otherwise to wrong results of the difference calculation between iterators. 
It is now in line with the DictionarySegmentIterable and tested as well.


Bouncner:
Can you please add a description? When did this error occur?

Bouncner:
Justin, are you going to be around this week? I think it's easier for me to just talk with you about it.

j-tr:
will be presumably there on friday. i'll approach you then.

mrks:
> This leads otherwise to wrong results of the difference calculation between iterators.

We do have parameterized iterator tests, don't we? This sounds like those do not cover iterator differences. Could you check that?

j-tr:
this is indeed true. the tests in segment_iterators_test.cpp never use the distance_to function.

mrks:
distance_to is just a helper for the boost adaptor, I'd check `b - a`.

j-tr:
when I add a dummy subtraction with `-` in the test, `distance_to` is called. so i assume that it should be called as well by all other subtractions in the test code. and since this does not happen, i conclude that `-` isn't used. if i'm wrong, whats the best method to check for operator calls?

mrks:
Weird. Could you show what you have tried?

j-tr:
i just added `std::cout << "using difference_to" << std::endl;` to distance_to() and grep the test output for "using difference_to". it occurs 0 times. if im adding `std::cout << "using - " << end - begin << std::endl;` to the lambda argument of `test_all_iterators()` as well, both "using -" and "using difference_to" show up together in the LZ4 tests.

mrks:
Ah, I misread your first comment. So everything is working as expected and we're just missing some tests. I'd do something like:

```
auto a = x.begin();
auto b = x.begin();
b++;
b++;
expect_eq(b - a, 2);
```

```
auto a = x.begin();
auto b = a;
b += 3
expect_eq(b - a, 3);
```

j-tr:
your test will pass, since the difference is calculated on the offset member and it is correctly increased using ++ or +. however the problem is the initialization. The difference of two freshly created iterators is always 0 no matter where they are actually pointing to. So in `test_all_iterators()` `ASSERT_FALSE(end == begin);` will pass but `ASSERT_NE(end - begin, 0);` fails, since end and begin are pointing to the right (different) places, but their offset is both times 0. So in my opinion, it's not only about missing tests

mrks:
Got it. In that case, wouldn't replacing some of the begin()s with end()s help?

j-tr:
not really. and this is acutally the problem. since the difference between two new iterators is currently always 0 it doesnt matter wheather they are begin or end. consequently my pull request forces that the offset is explicitly set at initialization and avoids this.

mrks:
`EXPECT_EQ(x.end() - x.begin(), x.size());`?
`EXPECT_EQ(x.end() - x.begin() + 1, x.size() - 1);`?

j-tr:
i dont see how that could be tested since with_iterators just executes the functor with begin and end as arguments but we can't get the actual size of the underlaying structure from it. 
but i think we are overcomplicating it a little bit here. i'll just talk with @Bouncner tomorrow and i'm sure, we will find a solution quickly.

j-tr:
Is it even necessary to store the _chunk_offset that is causing all the trouble as a member variable? Alternatively, we could store the begin of the underlying container and calculate the offset on the fly when needed. This would add additional iterator difference operations but remove offset additions and subtractions.
Comparing the benchmarks for the current master and the alternative solution (see: https://github.com/hyrise/hyrise/compare/master...j-tr:iterator_offset#diff-c1d1d96af2f07116ac3940a99d811def) for the dictionary segment iterator shows that our current solution performs slightly better. nonetheless, the need to maintain two variables with basically the same information content (_attribute_it and _chunk_offset) seems more error-prone to me.

Unbenannt

Bouncner:
Hmm. I am fine with the change in the LZ4 iterator. But the 12% change for TPC-H Q20 is too much for my taste. So rather keep de-/increasing this single value.

Bouncner:
I am currently changing something on the FoR iterable: don't we have similar problems there as well? There is no `distance` as in the LZ4 iterable, but I wouldn't be able to do `iter.end() - 1` to access the last element. Right?

j-tr:
i don't understand whats the issue with `iter.end() - 1`. from my understanding, we usually use the iterators to call a functor with a begin and end iterator and since we have a decrement implemented, it should be no problem to get the last element by decrementing the end iterator. 

Concerning the parameterized tests: I started unifying them but this got unexpectedly complex since they depend on the encoding type (which is actually an enum value), the segment type and the iterable type. since we have mixed value/type parameters here, gtest doesn't like that. But im sure we could still parameterize it by infering the encoding type from the segment type and use purely type parameterized test.

Bouncner:
When you request end(), it sets `_chunk_offset` to `0u` and a `end()-1` would return a negative chunk_offset in `dereference()` (which is unsigned, thus something very large). So the actual value is correct that is returned when we decrement `end()`, but any functor that uses `chunk_offset` might run into problems. Or do I totally miss something here?

Concerning the test: do you think you can do this as part of this PR?

j-tr:
No you are right. The zero initialization is a problem here.
If there is nothing else urgently, i can can continue with the tests. @Bensk1?

mrks:
> When you request end(), it sets _chunk_offset to 0u

Isn't that the problem in the first place?

mrks:
> Hmm. I am fine with the change in the LZ4 iterator. But the 12% change for TPC-H Q20 is too much for my taste. So rather keep de-/increasing this single value.
>
> @mrks: What's your opinion on that?

This doesn't feel right. Especially the big performance loss should be some alignment issue. I'm running Benchmarks on nemea right now. From a code perspective, I agree with @j-tr.

mrks:
I'm getting differing results:

```diff
 +----------------+----------------+-------+----------------+-------+------------+---------+
 | Benchmark      | prev. iter/s   | runs  | new iter/s     | runs  | change [%] | p-value |
 +----------------+----------------+-------+----------------+-------+------------+---------+
 | TPC-H 01       | 0.591941773891 | 36    | 0.595294058323 | 36    | +1%        |  0.4373 |
 | TPC-H 02       | 68.1781997681  | 4091  | 68.8165359497  | 4129  | +1%        |  0.2225 |
+| TPC-H 03       | 6.9803276062   | 419   | 7.34872531891  | 441   | +5%        |  0.0000 |
 | TPC-H 04       | 6.87540578842  | 413   | 6.86310577393  | 412   | -0%        |  0.7361 |
+| TPC-H 05       | 3.4995496273   | 211   | 3.75077962875  | 226   | +7%        |  0.0000 |
 | TPC-H 06       | 198.962280273  | 11938 | 201.906677246  | 12115 | +1%        |  0.0000 |
 | TPC-H 07       | 7.41563415527  | 445   | 7.38929224014  | 444   | -0%        |  0.6098 |
+| TPC-H 08       | 5.79976272583  | 349   | 6.39093160629  | 384   | +10%       |  0.0000 |
 | TPC-H 09       | 1.49156761169  | 90    | 1.52738904953  | 92    | +2%        |  0.0000 |
 | TPC-H 10       | 2.61136102676  | 157   | 2.63996148109  | 159   | +1%        |  0.2057 |
 | TPC-H 11       | 28.869146347   | 1733  | 28.6622104645  | 1720  | -1%        |  0.0000 |
 | TPC-H 12       | 8.78120422363  | 527   | 8.8173456192   | 530   | +0%        |  0.0000 |
 | TPC-H 13       | 2.16691184044  | 131   | 2.22579503059  | 134   | +3%        |  0.0000 |
 | TPC-H 14       | 45.4273376465  | 2726  | 46.4343147278  | 2787  | +2%        |  0.0000 |
 | TPC-H 15       | 71.3107147217  | 4279  | 71.8279190063  | 4310  | +1%        |  0.0000 |
 | TPC-H 16       | 6.92698097229  | 416   | 6.97177314758  | 419   | +1%        |  0.0578 |
 | TPC-H 17       | 4.65258264542  | 280   | 4.80569219589  | 289   | +3%        |  0.0000 |
 | TPC-H 18       | 0.77457690239  | 47    | 0.80079382658  | 49    | +3%        |  0.0022 |
+| TPC-H 19       | 8.66394901276  | 521   | 10.5141048431  | 631   | +21%       |  0.0000 |
 | TPC-H 20       | 25.9511432648  | 1558  | 26.8259372711  | 1610  | +3%        |  0.0000 |
 | TPC-H 21       | 1.16310191154  | 70    | 1.19819819927  | 72    | +3%        |  0.0001 |
 | TPC-H 22       | 14.0375976562  | 843   | 13.9889278412  | 840   | -0%        |  0.3040 |
 | geometric mean |                |       |                |       | +3%        |         |
 +----------------+----------------+-------+----------------+-------+------------+---------+
```

Again, I'm not 100% convinced that those are actual performance benefits. At least, it does not seem to hurt performance significantly. I'd argue that it's cleaner, so I'd bring it in.

j-tr:
whats your setup? how could one explain these big deviations of the different benchmark runs? did you test the dictionary or LZ4 encoding? i run it on sidon, one numa node, release mode, scale factor 1 and 240 seconds per query.

Bouncner:
It's cleaner, that's why we started the experiment in the first place. So if we can rule out performance regressions for TPC-H I am happy with the change, but not if it turns out that the Q20 drop was really caused by the change. :-/

@j-tr, can you please also change the offset initialisation of the FoR iterators? (and maybe have a quick look over other other segments and rename "start_position" in RLE to `chunk_offset` [RLE looks good though]).

mrks:
nemea, Dictionary encoding (that's what we are talking about, isn't it?), single node, release, SF 1, 60 seconds.

I'm a bit surprised myself. To be honest, I have somehow started to live with seemingly random performance jumps when moving code around. A good candidate is code alignment, but there are other reasons why even seemingly identical code may have weird performance behavior. There is a whole [paper](https://users.cs.northwestern.edu/~robby/courses/322-2013-spring/mytkowicz-wrong-data.pdf) on it (thanks, Martin).

Bouncner:
@j-tr: so the error boils down to setting the `offset` (or whatever other name an iterator uses) is wrongly set for certain segment iterables, right? The distance discussion was a bit misleading, since it was only one way to show the error for LZ4 (which uses offset to calculate the distance) but not for other segments.

Could we add a short test that checks the chunk offset of the returned `SegmentPosition<T>` when we decrease `iter` (with `iter = end()`) until `iter == begin()`? That's the point where most iterables make use of the stored offset member.

Bensk1:
> No you are right. The zero initialization is a problem here.
> If there is nothing else urgently, i can can continue with the tests. @Bensk1?

Sure!

Bouncner:
Any news here?

j-tr:
> Could we add a short test that checks the chunk offset of the returned SegmentPosition<T> when we decrease iter (with iter = end()) until iter == begin()? That's the point where most iterables make use of the stored offset member.

Some iterators seemed to be implemented as forward only iterators (e.g. SimdBp128Iterator), so subtraction lead to segfaults later on. was this a design decision or a lack of implementation? I changed the end operators to support decrement and added a parameterized test for all possible encodings and compressions now.

Bouncner:
> Some iterators seemed to be implemented as forward only iterators (e.g. SimdBp128Iterator), so subtraction lead to segfaults later on. was this a design decision or a lack of implementation? I changed the end operators to support decrement and added a parameterized test for all possible encodings and compressions now.

I am not aware of such a decision. For the vector compression types, it might be ok to be non-random, I guess. At least I cannot think of non-point accesses.

src/test/storage/iterables_test.cpp
```diff
@@ -247,11 +308,11 @@ TEST_F(IterablesTest, ReferenceSegmentIteratorWithIterators) {
 
   auto iterable = ReferenceSegmentIterable<int, EraseReferencedSegmentType::No>{*reference_segment};
 
-  auto sum = uint32_t{0};
+  auto sum = int{0};
```
Bouncner: Why using the sum as an integer when you are going to compare against an unsigned value in EXPECT_EQ?

src/test/storage/iterables_test.cpp
```diff
@@ -247,11 +308,11 @@ TEST_F(IterablesTest, ReferenceSegmentIteratorWithIterators) {
...
   auto accessed_offsets = std::vector<ChunkOffset>{};
-  iterable.with_iterators(SumUpWithIterator{sum, accessed_offsets});
+  iterable.with_iterators(SumUpWithIterator<int>{sum, accessed_offsets});
 
-  EXPECT_EQ(sum, 24'825u);
```
Bouncner: Personally, I like large numbers to be separated.

mrks:
To add to yesterday's discussion: The SimdBp128Decompressor seems to have an expensive but O(1) point access method. Maybe that helps a bit?

j-tr:
@Bouncner i think the simdbp128 problem is solved now and all iterators seem to work also backwards.

src/lib/storage/vector_compression/simd_bp128/simd_bp128_iterator.cpp
```diff
@@ -5,11 +5,11 @@ namespace opossum {
 SimdBp128Iterator::SimdBp128Iterator(const pmr_vector<uint128_t>* data, size_t size, size_t absolute_index)
```
Bouncner: `const`?

src/test/storage/iterables_test.cpp
```diff
@@ -1,3 +1,4 @@
+
```
Bouncner: ?

src/test/storage/iterables_test.cpp
```diff
@@ -9,22 +10,26 @@
 #include "gtest/gtest.h"
```
Bouncner: Merge the master please.

src/test/storage/iterables_test.cpp
```diff
@@ -103,139 +108,197 @@ class IterablesTest : public BaseTest {
...
-  EXPECT_EQ(accessed_offsets, (std::vector<ChunkOffset>{ChunkOffset{0}, ChunkOffset{1}, ChunkOffset{2}}));
-}
+class EncodedSegmentIterablesTest : public IterablesTest,
+                                    public ::testing::WithParamInterface<std::tuple<SegmentEncodingSpec, bool, bool>> {
+ public:
+  uint32_t expected_sum(bool nullable, bool with_position_filter) {
+    if (nullable && with_position_filter) {
+      return 12480u;
+    } else if (nullable && !with_position_filter) {
+      return 24825u;
```
Bouncner: 24`825u?

src/test/storage/iterables_test.cpp
```diff
@@ -103,139 +108,197 @@ class IterablesTest : public BaseTest {
...
-  EXPECT_EQ(accessed_offsets,
-            (std::vector<ChunkOffset>{ChunkOffset{0}, ChunkOffset{1}, ChunkOffset{2}, ChunkOffset{3}}));
+INSTANTIATE_TEST_SUITE_P(SegmentEncoding, EncodedSegmentIterablesTest,
+                         ::testing::Combine(::testing::ValuesIn(all_segment_encoding_specs),
+                                            ::testing::Bool(),   // nullable
+                                            ::testing::Bool()),  // position filter
+                         formatter);
+
+TEST_P(EncodedSegmentIterablesTest, IteratorWithIterators) {
+  auto encoding_spec = std::get<0>(GetParam());
```
Bouncner: const's

src/test/storage/iterables_test.cpp
```diff
@@ -103,139 +108,197 @@ class IterablesTest : public BaseTest {
...
+  auto nullable = std::get<1>(GetParam());
+  auto with_position_filter = std::get<2>(GetParam());
+  std::shared_ptr<Table> test_table = (nullable ? table : table_with_null);
+
+  auto chunk_encoding_spec = ChunkEncodingSpec{test_table->column_count(), EncodingType::Unencoded};
+  for (auto column_id = ColumnID{0}; column_id < test_table->column_count(); ++column_id) {
+    if (encoding_supports_data_type(encoding_spec.encoding_type, test_table->column_data_type(column_id))) {
```
Bouncner: Could we skip if the encoding is not supported? We'll run a couple of duplicated tests with Unencoded here, right?

src/test/storage/iterables_test.cpp
```diff
@@ -103,139 +108,197 @@ class IterablesTest : public BaseTest {
...
-  auto segment = chunk->get_segment(ColumnID{0u});
-  auto int_segment = std::dynamic_pointer_cast<const ValueSegment<int>>(segment);
-
-  auto iterable = ValueSegmentIterable<int>{*int_segment};
-
-  auto sum = uint32_t{0};
-  auto accessed_offsets = std::vector<ChunkOffset>{};
-  iterable.with_iterators(position_filter, SumUpWithIterator{sum, accessed_offsets});
+class EncodedStringSegmentIterablesTest : public IterablesTest,
+                                          public ::testing::WithParamInterface<std::tuple<SegmentEncodingSpec, bool>> {
```
Bouncner: Better use `BaseTestWithParams`.

src/test/storage/iterables_test.cpp
```diff
@@ -103,139 +108,197 @@ class IterablesTest : public BaseTest {
...
+ public:
+  pmr_string expected_concatenation(bool with_position_filter) {
```
Bouncner: const

src/test/storage/iterables_test.cpp
```diff
@@ -9,22 +10,26 @@
...
 namespace opossum {
 
+template <typename DataType>
 struct SumUpWithIterator {
   template <typename Iterator>
   void operator()(Iterator begin, Iterator end) const {
     auto distance = end - begin;
 
-    _sum = 0u;
+    _sum = 0;
```
Bouncner: Feel free to get rid of some whitespaces here if you like.. Too much for my taste.

Bouncner: Also, in sumUp _sum is not initialized explicitly, here it is. I'd vote to initialize it in the struct directly `int sum{0}`.

src/test/storage/iterables_test.cpp
j-tr: i found that `GTEST_SKIP()` can do this. however this produces "skipped tests" in the test resume. is this what we want? the alternative would be to specify the parameter combinations manually and exclude string-specific encodings.

src/test/storage/iterables_test.cpp
j-tr: how can I do so? I think, we need the interface here, since IterablesTest is also derived from BaseTest and using BaseTestWithParams as second base class causes ambiguities.

src/test/storage/iterables_test.cpp
j-tr: i guess the initial idea why sum was initialized outside of the struct was to have a SumUp method that can be called on multiple segments subsequently using the same sum variable. we would lose this functionality if we would initioalize it in the struct. i got rid of the multiple `= 0`s by initializing it outside of the struct. open for discussions.

src/test/storage/iterables_test.cpp
Bouncner: Yep, better change that to signed. And I'd also vote for `int32_t` which is the exact type behind our "int" (see all_type_variant.hpp)

src/test/storage/iterables_test.cpp
Bouncner: I have no opinion here. If it is reused, keep it that way. But I would still unify the structs (even if one of them is not reused).

src/test/storage/iterables_test.cpp
Bouncner: If it's ot straightforward, just ignore my comment. ;)

j-tr:
indeed. in the end, the problem was just that some member variables weren't set properly for the end iterator.

Bouncner:
Do we know for sure that the SimdBp vectors are never accessed randomly via the iterator?

j-tr:
i guess we can't guarantee that. are you aiming at the advance by multiplpe increment/decrements?

Bouncner:
In case you have a debug build laying around: could you break at `dereference()` and run TPC-H with with `-e LZ4`? LZ4 uses SimdBP128 and I hope TPC-H never hits the breakpoint.

Bouncner:
> i guess we can't guarantee that. are you aiming at the advance by multiplpe increment/decrements?

The advance is (unfortunately) a completely different topic. It's probably never going to be called, no matter what you do due to the current problem with input iterator traits.

j-tr:
confirmed. no dereference during TPC-H.

src/test/storage/iterables_test.cpp
Bouncner: @mrks & @Bensk1  : your take unnecessary tests and how to skip them best?

src/test/storage/iterables_test.cpp
```diff
@@ -102,153 +94,243 @@ class IterablesTest : public BaseTest {
...
+  const auto encoding_spec = std::get<0>(GetParam());
+  const auto with_position_filter = std::get<1>(GetParam());
+
+  auto chunk_encoding_spec = ChunkEncodingSpec{test_table->column_count(), EncodingType::Unencoded};
+  for (auto column_id = ColumnID{0}; column_id < test_table->column_count(); ++column_id) {
+    if (encoding_supports_data_type(encoding_spec.encoding_type, test_table->column_data_type(column_id))) {
+      chunk_encoding_spec[column_id] = encoding_spec;
+    } else {
+      if (column_id  == ColumnID{0}) {
+        GTEST_SKIP() << "encoding does't support data type";
```
Bouncner: doesN't

This yields a lot of skipped tests, doesn't it?

src/test/storage/iterables_test.cpp
mrks: > I don't like the use of GTEST_SKIP() here either. I can live with a simple return;

Agreed

src/test/storage/iterables_test.cpp
```diff
@@ -152,9 +152,7 @@ TEST_P(EncodedSegmentIterablesTest, IteratorWithIterators) {
     if (encoding_supports_data_type(encoding_spec.encoding_type, test_table->column_data_type(column_id))) {
       chunk_encoding_spec[column_id] = encoding_spec;
     } else {
-      if (column_id  == ColumnID{0}) {
-        GTEST_SKIP() << "encoding does't support data type";
-      }
+      if (column_id  == ColumnID{0}) return;  // skip test
```
Bouncner: Why do you skip this column id? Please extend the comment.

Bouncner:
Apart from 2-3 comments and small issues (using `int32_t`), this PR looks ready to be merged. When will you be able to incorporate the last changes, @j-tr ?

Besides, do you think it would make sense to add a `PerformanceWarning` in all `advance()` iterator methods that do single in/decrements? Should help us to later pin down problems when the iterators are fixed.

j-tr:
all feedback applied but still failing because of ccache. @mrks can you solve this?

mrks:
Gna. Please try restarting the build. Will look at that once the paper is done.

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
```diff
@@ -101,6 +101,7 @@ class FrameOfReferenceSegmentIterable : public PointAccessibleSegmentIterable<Fr
 
     void advance(std::ptrdiff_t n) {
       // For now, the lazy approach
+      PerformanceWarning("using repeated increment/decrement for random access");
```
mrks: Wouldn't make you re-run CI for that, but since we have to do that anyway: Capitalize 'Using'.

Issue #1643, opened 2019-04-26
SortMergeJoin is not thread-safe


mrks:
`[ FAILED ] SQLiteTestRunnerUnencoded/SQLiteTestRunner.CompareToSQLite/106, where GetParam() = ("SELECT * FROM id_int_int_int_100 AS t1 LEFT JOIN id_int_int_int_100 AS t2 ON t1.a >= t2.a AND t1.b >= t2.b AND t1.c >= t2.c;", false, Unencoded) (106 ms)`

Running `./hyriseTest --gtest_filter=SQLiteTestRunnerUnencoded/SQLiteTestRunner.CompareToSQLite/106` either crashes, passes, or fails on clang-debug with fed77246c3751d2d50566fb27f7898b0c6b3db15 (download zip from Github).

Also, the JoinTestRunner should have a multi-threaded configuration.

mrks:
These rows show up in our result but not in sqlite (which has no NULLs):

```
          id        a        b        c       id        a        b        c
         int      int      int      int      int      int      int      int
 476      16       41       10       30     NULL     NULL     NULL     NULL
 477      84       72       19        2     NULL     NULL     NULL     NULL
```

@mrzzzrm Does this look familiar in any way by chance?

mrks:
This never worked:

```
git submodule update
echo 'SELECT * FROM id_int_int_int_100 AS t1 LEFT JOIN id_int_int_int_100 AS t2 ON t1.a >= t2.a AND t1.b >= t2.b AND t1.c >= t2.c;' > resources/test_data/sqlite_testrunner_queries.sql
cd build
ninja hyriseTest
timeout -s SIGKILL --preserve-status 120 ./hyriseTest --gtest_filter=SQLiteTestRunnerUnencoded/SQLiteTestRunner.CompareToSQLite/0 --gtest_repeat=20
ret=$?
cd ..
git checkout resources/test_data/sqlite_testrunner_queries.sql
if [ "$ret" -eq "0" ]; then return 0; fi; return 1
```

```
95561eca450cac55abc696e0c74be01bd68998ac is the first bad commit
commit 95561eca450cac55abc696e0c74be01bd68998ac
Author: Kolpa <kolyaopahle@gmail.com>
Date:   Tue Mar 26 11:51:15 2019 +0100

    [DYOD18/19] Multi Predicate Join Support (#1482)
```

mrks:
Here's a different Hyrise result:

```
[...]

PR #1948, opened 2019-12-04
Improve get_chunk == nullptr documentation

mrks:
My seminar group found the reference to #1686 confusing and somehow ended up at #1643 (?!) without understanding why we checked for nullptr. Hopefully this is a bit clearer.

### DIFF
--- a/src/lib/storage/table.hpp
+++ b/src/lib/storage/table.hpp
@@ -81,14 +81,14 @@ class Table : private Noncopyable {
    * @defgroup Accessing and adding Chunks
    * @{
    */
-  // Returns the number of chunks, or, more correctly, the ID of the last chunk plus one (see get_chunk)
+  // Returns the number of chunks, or, more correctly, the ID of the last chunk plus one (see get_chunk / #1686).
   // This cannot exceed ChunkID (uint32_t).
   ChunkID chunk_count() const;
 
   // Returns the chunk with the given id. If a previously existing chunk has been physically deleted by the
   // MvccDeletePlugin, this returns nullptr. In the execution engine, it is the GetTable operator's job to
-  // filter these nullptrs and return only existing chunks to the following operator. Thus, all operators
-  // should assert the absence of nullptrs.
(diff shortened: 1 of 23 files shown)


src/lib/storage/table.hpp
```diff
@@ -81,10 +81,14 @@ class Table : private Noncopyable {
    * @defgroup Accessing and adding Chunks
    * @{
    */
-  // returns the number of chunks (cannot exceed ChunkID (uint32_t))
+  // Returns the number of chunks, or, more correctly, the ID of the last chunk plus one (see get_chunk)
```
Bouncner: Trailing dot.

src/lib/storage/table.hpp
```diff
@@ -81,10 +81,14 @@ class Table : private Noncopyable {
...
+  // This cannot exceed ChunkID (uint32_t).
   ChunkID chunk_count() const;
 
-  // returns the chunk with the given id
+  // Returns the chunk with the given id. If a previously existing chunk has been physically deleted by the
+  // MvccDeletePlugin, this returns nullptr. In the execution engine, it is the GetTable operator's job to
+  // filter these nullptrs and return only existing chunks to the following operator. Thus, all operators
+  // should assert the absence of nullptrs.
```
Bouncner: "should assert that this function always returns a chunk."?

mrks: Thus, all other operators should not accept nullptrs and instead assert that this function returned a chunk

?

mrks:
> Actually, I found the note // see #1686 really helpful.

Why not both: `Physically deleted chunk should not reach this point, see get_chunk / #1686.`

src/lib/storage/table.hpp
Bouncner: Accept > expect?

mrks: Not really what I wanted to say. If you wanted to use expect, I would expect "should not have to expect".

PR #1949, opened 2019-12-05
Fix plan reuse in SubqueryToJoinRule

### LINKED ISSUE #1940: Subquery to Join Rule should deep copy LQPs
If a subquery LQP is used twice and the SubqueryToJoinRule can only be applied once, some nodes are part of both the main and the subquery LQP:

`SELECT a FROM id_int_int_int_100 WHERE a IN (SELECT 11) AND b > (SELECT 11);`

## Input
<pre>
[0] [LogicalPlanRootNode] @ 0x1183dc238
 \_[1] [Projection] a @ 0x1183dc198
    \_[2] [Predicate] b > SUBQUERY (LQP, 0x1183dbfb8) @ 0x11902b018
       \_[3] [Predicate] (a) IN SUBQUERY (LQP, 0x1183dbfb8) @ 0x118d0cdd8
          \_[4] [Validate] @ 0x1183dc058
             \_[5] [StoredTable] Name: 'id_int_int_int_100' pruned: 0/10 chunk(s), 2/4 column(s) @ <b>0x118cd7358</b>
-------- Subqueries ---------
0x1183dbfb8:
[0] [Projection] 11 @ 0x1183dbfb8
 \_[1] [DummyTable] @ <b>0x118d0cd18</b>
</pre>

## After SubqueryToJoinRule
<pre>
[0] [LogicalPlanRootNode] @ 0x1183dc238
 \_[1] [Projection] a @ 0x1183dc198
    \_[2] [Predicate] b > SUBQUERY (LQP, 0x1183dbfb8) @ 0x11902b018
       \_[3] [Join] Mode: Semi [a = 11] @ 0x118d0cf58
          \_[4] [Validate] @ 0x1183dc058
          |  \_[5] [StoredTable] Name: 'id_int_int_int_100' pruned: 0/10 chunk(s), 2/4 column(s) @ 0x118cd7358
          \_[6] [Projection] 11 @ 0x1183dc0f8
             \_[7] [DummyTable] @ <b>0x118d0cd18</b>
-------- Subqueries ---------
0x1183dbfb8:
[0] [Projection] 11 @ 0x1183dbfb8
 \_[1] [DummyTable] @ <b>0x118d0cd18</b>
</pre>

Somewhere in the rule, we are forgetting to call deep_copy. This might mean that more complex subqueries might even be cannibalized for the rewrite in a way where it breaks the remaining use of the subquery. This should be tested.

mrks:
Fixes #1940, i.e., the SubqueryToJoinRule modifying the original subquery LQP without any regard for other consumers of that subquery LQP.

### DIFF
--- a/src/lib/optimizer/optimizer.cpp
+++ b/src/lib/optimizer/optimizer.cpp
@@ -206,6 +206,12 @@ void Optimizer::validate_lqp(const std::shared_ptr<AbstractLQPNode>& root_node) 
         // enforced when rules are executed through the optimizer, not when tests call the rule directly.
         Assert(nodes_by_lqp[lqp].contains(output), std::string{"Output `"} + output->description() + "` of node `" +
                                                        node->description() + "` not found in LQP");
+
+        for (const auto& [other_lqp, nodes] : nodes_by_lqp) {
+          if (other_lqp == lqp) continue;
+          Assert(!nodes.contains(node), std::string{"Output `"} + output->description() + "` of node `" +
+                                                       node->description() + "` found in different LQP");
+        }
       }
 
       // Check that all LQPColumnExpressions in the node can be resolved. This does, however, not guarantee that it is
--- a/src/lib/optimizer/optimizer.cpp
+++ b/src/lib/optimizer/optimizer.cpp
@@ -182,11 +182,11 @@ void Optimizer::validate_lqp(const std::shared_ptr<AbstractLQPNode>& root_node) 
     lqps.emplace_back(lqp);
   }
 
-  // Next, collect all nodes found in the entire plan (not only the main LQP)
-  std::unordered_set<std::shared_ptr<const AbstractLQPNode>> all_nodes;
+  std::unordered_map<std::shared_ptr<const AbstractLQPNode>, std::unordered_set<std::shared_ptr<const AbstractLQPNode>>>
+      nodes_by_lqp;
   for (const auto& lqp : lqps) {
-    visit_lqp(lqp, [&all_nodes](const auto& node) {
-      all_nodes.emplace(node);
+    visit_lqp(lqp, [&](const auto& node) {
+      nodes_by_lqp[lqp].emplace(node);
       return LQPVisitation::VisitInputs;
     });
   }
@@ -195,7 +195,7 @@ void Optimizer::validate_lqp(const std::shared_ptr<AbstractLQPNode>& root_node) 
   // (2) Make sure each node has the number of inputs expected for that node type
   // (3) Make sure that for all LQPColumnExpressions, the original_node is part of the LQP
   for (const auto& lqp : lqps) {
-    visit_lqp(lqp, [&all_nodes](const auto& node) {
+    visit_lqp(lqp, [&](const auto& node) {
       // Check that all outputs are part of the LQP
       const auto outputs = node->outputs();
       for (const auto& output : outputs) {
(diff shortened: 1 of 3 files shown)


Bouncner:
I see the changes, they appear to be good. But honestly, I need to talk with you about it because I don't understand the actual problem being fixed.

Further, it would help if you could add a short comment, why you're using a map for the LQPs now (maybe not in code, but in the PR description).

mrks:
> I see the changes, they appear to be good. But honestly, I need to talk with you about it because I don't understand the actual problem being fixed.

Admittedly, this is a bit hard to spot in the PR as it was already part of the documentation before: "LQP nodes should only be part of a single LQP and all of their outputs should be part of the same LQP."

> Further, it would help if you could add a short comment, why you're using a map for the LQPs now (maybe not in code, but in the PR description).

Before, we only checked that all nodes occurred *somewhere* in the LQP (including subqueries), now we check that they appear in the current LQP, but not anywhere else.

src/lib/optimizer/optimizer.cpp
```diff
@@ -204,18 +204,24 @@ void Optimizer::validate_lqp(const std::shared_ptr<AbstractLQPNode>& root_node)
         // outside of the LQP that is currently optimized. For example, if you create a bunch of LQP nodes in the test's
         // SetUp method but these are not used in the current LQP, this can cause this assertion to fail. This is only
         // enforced when rules are executed through the optimizer, not when tests call the rule directly.
-        Assert(all_nodes.contains(output), std::string{"Output `"} + output->description() + "` of node `" +
-                                               node->description() + "` not found in LQP");
+        Assert(nodes_by_lqp[lqp].contains(output), std::string{"Output `"} + output->description() + "` of node `" +
+                                                       node->description() + "` not found in LQP");
+
+        for (const auto& [other_lqp, nodes] : nodes_by_lqp) {
```
Bouncner: Add a (4) to the list of tasks done that states that we are checking if nodes are only part of a single LQP only?

src/lib/optimizer/optimizer.cpp
```diff
@@ -204,18 +204,24 @@ void Optimizer::validate_lqp(const std::shared_ptr<AbstractLQPNode>& root_node)
...
+          if (other_lqp == lqp) continue;
+          Assert(!nodes.contains(node), std::string{"Output `"} + output->description() + "` of node `" +
+                                                       node->description() + "` found in different LQP");
+        }
       }
 
       // Check that all LQPColumnExpressions in the node can be resolved. This does, however, not guarantee that it is
       // correctly used as one of the (transitive) inputs of `node`. Checking that would be more expensive.
```
Bouncner: Should we care about that price ("be more expensive") when validate is only used in debug anyways?

src/test/optimizer/strategy/subquery_to_join_rule_test.cpp
```diff
@@ -689,9 +696,12 @@ TEST_F(SubqueryToJoinRuleTest, IsPredicateNodeJoinCandidateHandlesComparison) {
 
   const auto input_info = SubqueryToJoinRule::is_predicate_node_join_candidate(*lqp);
```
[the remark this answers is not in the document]
mrks: Yes. I found a number of readability / architecture issues in the rule, of which this is one of the smaller ones. However, I don't consider it necessary to refactor the rule (which is working and contained) at the moment.

src/lib/optimizer/optimizer.cpp
mrks: I'd rather rewrite (1):

> Make sure that all outputs found anywhere in the entire plan are also part of the plan

to

> Make sure that all outputs found in an LQP are also part of the same LQP (excluding subqueries)

src/lib/optimizer/optimizer.cpp
mrks: Now:

> Check that all LQPColumnExpressions in the node can be resolved. This mostly guards against expired columns leaving an optimizer rule. It does not guarantee that it the LQPColumnExpressions are correctly returned from one of the (transitive) inputs of `node`. If that is not the case, that will be caught by the LQPTranslator, at the latest. However, feel free to add that check here.

mrks: Read: We probably could afford to do it here, but I don't feel like adding it right now ;)

src/lib/optimizer/optimizer.cpp
Bouncner: Yep.

PR #1951, opened 2019-12-08
Fix broken _scan_sorted_segment

### LINKED ISSUE #1950: TableScan's _scan_sorted_segment broken
In cases where all values of a segment are equal or NULL and the search value is larger than the segment's value (in case of NULL the search value does not matter), `scan_sorted_segment` fails because it receives a BeginIterator that is equals to the EndIterator and then tries to dereference it.

Bensk1:
This fixes #1950 (see description there) by checking if the received begin iterator equals the end iterator.


resources/test_data/tbl/int_sorted_only_null.tbl
```diff
@@ -0,0 +1,7 @@
+a
```
mrks: What is "sorted" about this file?

src/lib/operators/table_scan/column_vs_value_table_scan_impl.cpp
```diff
@@ -153,6 +153,7 @@ void ColumnVsValueTableScanImpl::_scan_sorted_segment(const BaseSegment& segment
                                                          boost::get<ColumnDataType>(value));
 
         sorted_segment_search.scan_sorted_segment([&](auto begin, auto end) {
+          if (begin == end) return;
```
mrks: (Not related to this fix)

It took me a while to understand how `scan_sorted_segment` handles NotEquals. While that behavior is documented in `sorted_segment_search.hpp`, I think a comment "In case of NotEquals, this is called twice" would have helped a lot.

src/test/operators/table_scan_test.cpp
```diff
@@ -301,6 +306,24 @@ TEST_P(OperatorsTableScanTest, SingleScanWithSortedSegmentEquals) {
   EXPECT_TABLE_EQ_UNORDERED(scan->get_output(), expected_result);
 }
 
+TEST_P(OperatorsTableScanTest, SingleScanWithSortedSegmentEqualsAllElementsEqualNull) {
```
mrks: I believe the whole sorted scan is undertested. Even now, we do not test, e.g., LessThan. We probably should adapt the existing tests to cover sorted segments. ScanOnCompressedSegmentsValueGreaterThanMaxDictionaryValue looks like it would have caught this if we had a version with sorted segments.

Your decision if you want to add this in the PR or open an issue.

resources/test_data/tbl/int_sorted_only_null.tbl
Bensk1: Nothing and everything. I think this philosophical question has to be discussed offline during a Gin soiree. Otherwise this PR cannot be merged.

mrks: This sounds like you already started drinking?

src/lib/operators/table_scan/column_vs_value_table_scan_impl.cpp
Bensk1: I will add that

src/test/operators/table_scan_test.cpp
Bensk1: I totally agree. As @Bouncner said, since his group (including @d-justen) is working on the SortedSegmentScan, they could take care of that.

src/lib/operators/table_scan/column_vs_value_table_scan_impl.cpp
Bensk1: Hm, I don't see how it is called twice. The method `_handle_not_equals` determines two ranges and combines them. Is that what you mean?

mrks: Yes - for me, that was quite unexpected.

src/test/operators/table_scan_test.cpp
mrks: Is this tracked somewhere?

src/lib/operators/table_scan/column_vs_value_table_scan_impl.cpp
```diff
@@ -153,6 +153,11 @@ void ColumnVsValueTableScanImpl::_scan_sorted_segment(const BaseSegment& segment
...
+
+          // General note: If the predicate is NotEquals, there might be two matching ranges. scan_sorted_segment
+          // combines these two ranges into a single one. See sorted_segment_search.hpp for further details.
```
mrks: Isn't this misleading? `scan_sorted_segment combines these two ranges into a single one` - the way I read this, the opposite happens. It does NOT combine them but calls the lambda twice.

Bensk1: I think I misunderstood you in the first place. Let's clarify that offline.

src/test/operators/table_scan_test.cpp
Bensk1: No, I will open an issue if we decide to do it like that.

Bensk1:
Dafuq? https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-1951/8/pipeline

mrks:
No idea - open an issue?

Bouncner:
Had the same issue today for Q10 but assumed is was my fault (aggegrate rule).
Haven’t seen this before.

Bensk1:
Me neither, any idea what might have introduced that? I uploaded the log to #1954

Bensk1:
I will add an issue regarding the need for improved scan tests.

PR #1956, opened 2019-12-11
Projection: Better forward of ReferenceSegments

mrks:
For any table, we require all segments to be either ReferenceSegments or not. Thus, when the Projection creates a single ValueSegment, all other ReferenceSegments have to be converted into ValueSegments. Currently, this happens via the ExpressionEvaluator, which has a couple of unnecessary abstractions for this. By doing this step in the projection itself, we can save some time:

```diff
 +----------------+----------------+-------+----------------+-------+------------+---------+
 | Benchmark      | prev. iter/s   | runs  | new iter/s     | runs  | change [%] | p-value |
 +----------------+----------------+-------+----------------+-------+------------+---------+
 | TPC-H 01       | 0.591941773891 | 36    | 0.614538609982 | 37    | +4%        |  0.0000 |
 | TPC-H 02       | 68.1781997681  | 4091  | 66.7094421387  | 4003  | -2%        |  0.0044 |
 | TPC-H 03       | 6.9803276062   | 419   | 7.26622486115  | 436   | +4%        |  0.0000 |
 | TPC-H 04       | 6.87540578842  | 413   | 6.85076665878  | 412   | -0%        |  0.4591 |
+| TPC-H 05       | 3.4995496273   | 211   | 3.85983896255  | 232   | +10%       |  0.0000 |
 | TPC-H 06       | 198.962280273  | 11938 | 198.411514282  | 11905 | -0%        |  0.0458 |
 | TPC-H 07       | 7.41563415527  | 445   | 7.32884645462  | 440   | -1%        |  0.1001 |
+| TPC-H 08       | 5.79976272583  | 349   | 6.23874807358  | 375   | +8%        |  0.0000 |
 | TPC-H 09       | 1.49156761169  | 90    | 1.53204488754  | 92    | +3%        |  0.0000 |
+| TPC-H 10       | 2.61136102676  | 157   | 2.75670576096  | 166   | +6%        |  0.0000 |
 | TPC-H 11       | 28.869146347   | 1733  | 28.2791900635  | 1697  | -2%        |  0.0000 |
 | TPC-H 12       | 8.78120422363  | 527   | 8.70268154144  | 523   | -1%        |  0.0000 |
 | TPC-H 13       | 2.16691184044  | 131   | 2.21104550362  | 133   | +2%        |  0.0000 |
 | TPC-H 14       | 45.4273376465  | 2726  | 46.188533783   | 2772  | +2%        |  0.0000 |
 | TPC-H 15       | 71.3107147217  | 4279  | 68.9120635986  | 4135  | -3%        |  0.0000 |
 | TPC-H 16       | 6.92698097229  | 416   | 6.9563832283   | 418   | +0%        |  0.2133 |
 | TPC-H 17       | 4.65258264542  | 280   | 4.54050970078  | 273   | -2%        |  0.0006 |
+| TPC-H 18       | 0.77457690239  | 47    | 0.826413154602 | 50    | +7%        |  0.0000 |
 | TPC-H 19       | 8.66394901276  | 521   | 8.58924388885  | 516   | -1%        |  0.3671 |
 | TPC-H 20       | 25.9511432648  | 1558  | 25.4696979523  | 1529  | -2%        |  0.0000 |
 | TPC-H 21       | 1.16310191154  | 70    | 1.14416277409  | 69    | -2%        |  0.0992 |
 | TPC-H 22       | 14.0375976562  | 843   | 13.876742363   | 833   | -1%        |  0.0014 |
 | geometric mean |                |       |                |       | +1%        |         |
 +----------------+----------------+-------+----------------+-------+------------+---------+
```

No new tests, as this is a shortcut for something that has to happen anyway.

### DIFF
--- a/src/lib/operators/projection.cpp
+++ b/src/lib/operators/projection.cpp
@@ -12,6 +12,7 @@
 #include "expression/expression_utils.hpp"
 #include "expression/pqp_column_expression.hpp"
 #include "expression/value_expression.hpp"
+#include "storage/segment_iterate.hpp"
 #include "utils/assert.hpp"
 
 namespace opossum {
@@ -81,6 +82,40 @@ std::shared_ptr<const Table> Projection::_on_execute() {
         output_segments[column_id] = input_chunk->get_segment(pqp_column_expression->column_id);
         column_is_nullable[column_id] =
             column_is_nullable[column_id] || input_table.column_is_nullable(pqp_column_expression->column_id);
+      } else if (expression->type == ExpressionType::PQPColumn && !forward_columns) {
+        // Cannot immediately forward the column because it is a ReferenceSegment and we need to return a ValueSegment.
+        // However, instead of going through the ExpressionEvaluator just to materialize the segment, we can do it
+        // here:
+        const auto pqp_column_expression = std::static_pointer_cast<PQPColumnExpression>(expression);
+        const auto segment = input_chunk->get_segment(pqp_column_expression->column_id);
+
+        resolve_data_type(expression->data_type(), [&](const auto data_type) {
+          using DataType = typename decltype(data_type)::type;
+          bool has_null = false;
+          auto values = pmr_concurrent_vector<DataType>(segment->size());
+          auto null_values = pmr_concurrent_vector<bool>(segment->size());
+
+          auto chunk_offset = ChunkOffset{0};
+          segment_iterate<DataType>(*segment, [&](const auto& position) {
+            if (position.is_null()) {
+              has_null = true;
+              null_values[chunk_offset] = true;
+            } else {
+              values[chunk_offset] = position.value();
+            }
+            ++chunk_offset;
+          });
+
+          auto value_segment = std::shared_ptr<ValueSegment<DataType>>{};
+          if (has_null) {
+            value_segment = std::make_shared<ValueSegment<DataType>>(std::move(values), std::move(null_values));
+          } else {
+            value_segment = std::make_shared<ValueSegment<DataType>>(std::move(values));
+          }
+
+          output_segments[column_id] = std::move(value_segment);
+          column_is_nullable[column_id] = has_null;
+        });
       } else {
         auto output_segment = evaluator.evaluate_expression_to_segment(*expression);
         column_is_nullable[column_id] = column_is_nullable[column_id] || output_segment->is_nullable();
--- a/src/lib/operators/projection.cpp
+++ b/src/lib/operators/projection.cpp
@@ -77,15 +77,15 @@ std::shared_ptr<const Table> Projection::_on_execute() {
       const auto& expression = expressions[column_id];
 
       // Forward input column if possible
-      if (expression->type == ExpressionType::PQPColumn && forward_columns) {
+      if (forward_columns) {
         const auto pqp_column_expression = std::static_pointer_cast<PQPColumnExpression>(expression);
         output_segments[column_id] = input_chunk->get_segment(pqp_column_expression->column_id);
         column_is_nullable[column_id] =
             column_is_nullable[column_id] || input_table.column_is_nullable(pqp_column_expression->column_id);
       } else if (expression->type == ExpressionType::PQPColumn && !forward_columns) {
-        // Cannot immediately forward the column because it is a ReferenceSegment and we need to return a ValueSegment.
-        // However, instead of going through the ExpressionEvaluator just to materialize the segment, we can do it
-        // here:
+        // The current column will be returned without any logical modifications. As other columns do get modified (and
+        // returned as a ValueSegment), all segments (including this one) need to become ValueSegments. This segment is
+        // not yet a ValueSegment (otherwise forward_columns would be true); thus we need to materialize it.
         const auto pqp_column_expression = std::static_pointer_cast<PQPColumnExpression>(expression);
         const auto segment = input_chunk->get_segment(pqp_column_expression->column_id);
 
--- a/src/lib/operators/projection.cpp
+++ b/src/lib/operators/projection.cpp
@@ -90,13 +90,13 @@ std::shared_ptr<const Table> Projection::_on_execute() {
         const auto segment = input_chunk->get_segment(pqp_column_expression->column_id);
 
         resolve_data_type(expression->data_type(), [&](const auto data_type) {
-          using DataType = typename decltype(data_type)::type;
+          using ColumnDataType = typename decltype(data_type)::type;
           bool has_null = false;
-          auto values = pmr_concurrent_vector<DataType>(segment->size());
+          auto values = pmr_concurrent_vector<ColumnDataType>(segment->size());
           auto null_values = pmr_concurrent_vector<bool>(segment->size());
 
           auto chunk_offset = ChunkOffset{0};
-          segment_iterate<DataType>(*segment, [&](const auto& position) {
+          segment_iterate<ColumnDataType>(*segment, [&](const auto& position) {
             if (position.is_null()) {
               has_null = true;
               null_values[chunk_offset] = true;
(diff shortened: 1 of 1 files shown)


src/lib/operators/projection.cpp
```diff
@@ -81,6 +82,40 @@ std::shared_ptr<const Table> Projection::_on_execute() {
...
+
+        resolve_data_type(expression->data_type(), [&](const auto data_type) {
+          using ColumnDataType = typename decltype(data_type)::type;
+          bool has_null = false;
+          auto values = pmr_concurrent_vector<ColumnDataType>(segment->size());
+          auto null_values = pmr_concurrent_vector<bool>(segment->size());
+
+          auto chunk_offset = ChunkOffset{0};
+          segment_iterate<ColumnDataType>(*segment, [&](const auto& position) {
+            if (position.is_null()) {
```
Bouncner: Could we get rid of the branch by something like `is_null = pos.is_null();  has_null |= is_null;` and always writing to `values` and `null_values`?

mrks: Yes, but then we would always be writing to values and null_values. Hard to benchmark as we don't really do too much with NULLs yet.

mrks: Not really a good argument, but this is consistent with how we do it in other places.

Bouncner: Well, we don't do much with NULLs but the branch is there anyway. Worth to benchmark it?

mrks: on it

mrks:
```diff
 +----------------+----------------+-------+----------------+-------+------------+---------+
 | Benchmark      | prev. iter/s   | runs  | new iter/s     | runs  | change [%] | p-value |
 +----------------+----------------+-------+----------------+-------+------------+---------+
 | TPC-H 01       | 0.614538609982 | 37    | 0.603783071041 | 37    | -2%        |  0.0076 |
 | TPC-H 02       | 66.7094421387  | 4003  | 67.6907730103  | 4062  | +1%        |  0.0571 |
 | TPC-H 03       | 7.26622486115  | 436   | 7.09939622879  | 426   | -2%        |  0.0005 |
 | TPC-H 04       | 6.85076665878  | 412   | 6.92632436752  | 416   | +1%        |  0.0186 |
-| TPC-H 05       | 3.85983896255  | 232   | 3.6632399559   | 220   | -5%        |  0.0000 |
 | TPC-H 06       | 198.411514282  | 11905 | 198.467041016  | 11909 | +0%        |  0.8153 |
 | TPC-H 07       | 7.32884645462  | 440   | 7.52846670151  | 452   | +3%        |  0.0002 |
 | TPC-H 08       | 6.23874807358  | 375   | 6.14935827255  | 369   | -1%        |  0.0229 |
 | TPC-H 09       | 1.53204488754  | 92    | 1.52867507935  | 92    | -0%        |  0.6390 |
 | TPC-H 10       | 2.75670576096  | 166   | 2.69066119194  | 162   | -2%        |  0.0049 |
 | TPC-H 11       | 28.2791900635  | 1697  | 28.844127655   | 1731  | +2%        |  0.0000 |
 | TPC-H 12       | 8.70268154144  | 523   | 8.7239151001   | 524   | +0%        |  0.0020 |
 | TPC-H 13       | 2.21104550362  | 133   | 2.1796412468   | 131   | -1%        |  0.0000 |
 | TPC-H 14       | 46.188533783   | 2772  | 44.4472465515  | 2667  | -4%        |  0.0000 |
 | TPC-H 15       | 68.9120635986  | 4135  | 67.1021194458  | 4027  | -3%        |  0.0000 |
 | TPC-H 16       | 6.9563832283   | 418   | 6.91787004471  | 416   | -1%        |  0.0978 |
 | TPC-H 17       | 4.54050970078  | 273   | 4.64950180054  | 279   | +2%        |  0.0013 |
-| TPC-H 18       | 0.826413154602 | 50    | 0.77485537529  | 47    | -6%        |  0.0000 |
 | TPC-H 19       | 8.58924388885  | 516   | 8.95484256744  | 538   | +4%        |  0.0000 |
 | TPC-H 20       | 25.4696979523  | 1529  | 26.2662220001  | 1576  | +3%        |  0.0000 |
 | TPC-H 21       | 1.14416277409  | 69    | 1.129114151    | 68    | -1%        |  0.2558 |
 | TPC-H 22       | 13.876742363   | 833   | 13.9337511063  | 837   | +0%        |  0.2627 |
 | geometric mean |                |       |                |       | -1%        |         |
 +----------------+----------------+-------+----------------+-------+------------+---------+
```

As the notable regressions are exactly in those queries that were previously improved, this seems to hurt performance.

Bouncner: Ok, interesting. So the only situation where we __might__ see a benefit will be 50% of NULLs? Probably never going to happen. I try to remember that before I annoy you with that the next time. 🙂

src/lib/operators/projection.cpp
```diff
@@ -81,6 +82,40 @@ std::shared_ptr<const Table> Projection::_on_execute() {
         output_segments[column_id] = input_chunk->get_segment(pqp_column_expression->column_id);
         column_is_nullable[column_id] =
             column_is_nullable[column_id] || input_table.column_is_nullable(pqp_column_expression->column_id);
+      } else if (expression->type == ExpressionType::PQPColumn && !forward_columns) {
+        // Cannot immediately forward the column because it is a ReferenceSegment and we need to return a ValueSegment.
```
julianmenzler: > Cannot immediately forward the column because it is a ReferenceSegment and we need to return a ValueSegment.

I did not get this straight away. Add a comment about how you came to the above conclusion. E.g.:

```cpp
// In this branch, one column gets projected without any modifications. For other columns, however, 
// the projection does more than that. Therefore, we have to return a ValueSegment. Since forward_columns 
// is set to false, the input segment must be of type ReferenceSegment and needs to be materialized. (...)
```

src/lib/operators/projection.cpp
```diff
@@ -81,6 +82,40 @@ std::shared_ptr<const Table> Projection::_on_execute() {
         output_segments[column_id] = input_chunk->get_segment(pqp_column_expression->column_id);
         column_is_nullable[column_id] =
             column_is_nullable[column_id] || input_table.column_is_nullable(pqp_column_expression->column_id);
+      } else if (expression->type == ExpressionType::PQPColumn && !forward_columns) {
```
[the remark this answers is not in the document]
mrks: yepp

src/lib/operators/projection.cpp
mrks: Mmh, I can't point to it, but I found that one hard to follow. I tried to merge the two:

> The current column will be returned without any logical modifications. As other columns do get modified (and returned as a ValueSegment), all segments (including this one) need to become ValueSegments. This segment is not yet a ValueSegment (otherwise forward_columns would be true); thus we need to materialize it.

Better?

src/lib/operators/projection.cpp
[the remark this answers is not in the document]
mrks: Negative, there are some cases where this is not true, e.g., for modified columns on a data table.

src/lib/operators/projection.cpp
julianmenzler: Very good! :)

src/lib/operators/projection.cpp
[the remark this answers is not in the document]
julianmenzler: Wow, good find! Did not think of this szenario. Maybe even worth a comment. ;)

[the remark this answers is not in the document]
mrks: Well... "find" only in the sense that a bunch of tests flared up bright red. :D 
That's also the reason why I think that a comment is not really necessary, as it would not help in understanding the code.

PR #1957, opened 2019-12-11
Aggregate: Don't generate key for int32_t

mrks:
In the hash-based aggregate, we build the AggregateKey, which might have a single entry (group by single column) or multiple. To avoid AllTypeVariants, we need to map incoming values to unique integers. However, mapping ints to ints by effectively dictionary-compressing them makes little sense:

```diff
 +----------------+----------------+-------+---------------+-------+------------+---------+
 | Benchmark      | prev. iter/s   | runs  | new iter/s    | runs  | change [%] | p-value |
 +----------------+----------------+-------+---------------+-------+------------+---------+
 | TPC-H 01       | 0.591941773891 | 36    | 0.60118240118 | 37    | +2%        |  0.0362 |
 | TPC-H 02       | 68.1781997681  | 4091  | 68.4920883179 | 4110  | +0%        |  0.5468 |
+| TPC-H 03       | 6.9803276062   | 419   | 7.33792161942 | 441   | +5%        |  0.0000 |
 | TPC-H 04       | 6.87540578842  | 413   | 6.91902589798 | 416   | +1%        |  0.2511 |
 | TPC-H 05       | 3.4995496273   | 211   | 3.61497426033 | 217   | +3%        |  0.0043 |
 | TPC-H 06       | 198.962280273  | 11938 | 201.134384155 | 12069 | +1%        |  0.0000 |
 | TPC-H 07       | 7.41563415527  | 445   | 7.51439905167 | 451   | +1%        |  0.0566 |
+| TPC-H 08       | 5.79976272583  | 349   | 6.35709524155 | 382   | +10%       |  0.0000 |
 | TPC-H 09       | 1.49156761169  | 90    | 1.52550828457 | 92    | +2%        |  0.0000 |
 | TPC-H 10       | 2.61136102676  | 157   | 2.65954232216 | 160   | +2%        |  0.0347 |
 | TPC-H 11       | 28.869146347   | 1733  | 30.1973304749 | 1812  | +5%        |  0.0000 |
 | TPC-H 12       | 8.78120422363  | 527   | 8.76660633087 | 526   | -0%        |  0.0349 |
 | TPC-H 13       | 2.16691184044  | 131   | 2.27395629883 | 137   | +5%        |  0.0000 |
 | TPC-H 14       | 45.4273376465  | 2726  | 45.9030075073 | 2755  | +1%        |  0.0000 |
+| TPC-H 15       | 71.3107147217  | 4279  | 84.651802063  | 5080  | +19%       |  0.0000 |
 | TPC-H 16       | 6.92698097229  | 416   | 7.06166124344 | 424   | +2%        |  0.0000 |
 | TPC-H 17       | 4.65258264542  | 280   | 4.69012594223 | 282   | +1%        |  0.2701 |
+| TPC-H 18       | 0.77457690239  | 47    | 1.18020284176 | 71    | +52%       |  0.0000 |
 | TPC-H 19       | 8.66394901276  | 521   | 8.94591522217 | 537   | +3%        |  0.0011 |
 | TPC-H 20       | 25.9511432648  | 1558  | 27.0636177063 | 1624  | +4%        |  0.0000 |
 | TPC-H 21       | 1.16310191154  | 70    | 1.18508327007 | 72    | +2%        |  0.0420 |
 | TPC-H 22       | 14.0375976562  | 843   | 14.209446907  | 853   | +1%        |  0.0003 |
+| geometric mean |                |       |               |       | +5%        |         |
 +----------------+----------------+-------+---------------+-------+------------+---------+
```

No tests here as we don't do anything new. The existing tests that cover GROUP BY on int32_t should be sufficient.

@Bouncner - this should increase the impact of #1929.


mrks:
(If you are wondering about the performance improvements I talked about out-of-band: Everything gets better with *Performance first, Correctness second*.)

src/lib/operators/aggregate_hash.cpp
```diff
@@ -215,54 +215,80 @@ void AggregateHash::_aggregate() {
...
-                keys_per_chunk[chunk_id][chunk_offset] = 0u;
+                if (position.is_null()) {
+                  keys_per_chunk[chunk_id][chunk_offset] = 0;
+                } else {
+                  keys_per_chunk[chunk_id][chunk_offset] = position.value() + 1;
+                }
               } else {
-                keys_per_chunk[chunk_id][chunk_offset][group_column_index] = 0u;
+                if (position.is_null()) {
+                  keys_per_chunk[chunk_id][chunk_offset][group_column_index] = 0;
```
Bouncner: `!position.is_null() * (position.value() + 1)`?

src/lib/operators/aggregate_hash.cpp
```diff
@@ -215,54 +215,80 @@ void AggregateHash::_aggregate() {
...
-          const auto chunk_in = input_table->get_chunk(chunk_id);
-          if (!chunk_in) continue;
-
-          const auto base_segment = chunk_in->get_segment(column_id);
-
-          ChunkOffset chunk_offset{0};
-          segment_iterate<ColumnDataType>(*base_segment, [&](const auto& position) {
-            if (position.is_null()) {
+        if constexpr (std::is_same_v<ColumnDataType, int32_t>) {
+          // For values smaller than AggregateKeyEntry, we can use the value itself as an AggregateKeyEntry. We cannot
```
Bouncner: AggregateKeyEntry is the typ of the aggregate key, what does "smaller than" mean here?

Bouncner: And where do we ensure that is still fits even with the special NULL handling?

mrks: > AggregateKeyEntry is the typ of the aggregate key

The type of the aggregate key is AggregateKey, the template argument to _aggregate:

```
  switch (_groupby_column_ids.size()) {
    case 0:
    case 1:
      // No need for a complex data structure if we only have one entry
      _aggregate<AggregateKeyEntry>();
      break;
    case 2:
      // We need to explicitly list all array sizes that we want to support
      _aggregate<std::array<AggregateKeyEntry, 2>>();
      break;
    default:
      PerformanceWarning("No std::array implementation initialized - falling back to vector");
      _aggregate<std::vector<AggregateKeyEntry>>();
      break;
  }
```

> what does "smaller than" mean here?

Would "with a type smaller than" help?

> And where do we ensure that is still fits even with the special NULL handling?

AggregateKeyEntry is 64-bit and we only handle 32-bit here.

src/lib/operators/aggregate_hash.cpp
mrks: Benchmarking

mrks:
```diff
 +----------------+---------------+-------+----------------+-------+------------+---------+
 | Benchmark      | prev. iter/s  | runs  | new iter/s     | runs  | change [%] | p-value |
 +----------------+---------------+-------+----------------+-------+------------+---------+
 | TPC-H 01       | 0.60118240118 | 37    | 0.598081767559 | 36    | -1%        |  0.4720 |
 | TPC-H 02       | 68.4920883179 | 4110  | 68.7720794678  | 4127  | +0%        |  0.5938 |
 | TPC-H 03       | 7.33792161942 | 441   | 7.20852565765  | 433   | -2%        |  0.0018 |
 | TPC-H 04       | 6.91902589798 | 416   | 6.93402576447  | 417   | +0%        |  0.6708 |
 | TPC-H 05       | 3.61497426033 | 217   | 3.72428154945  | 224   | +3%        |  0.0104 |
 | TPC-H 06       | 201.134384155 | 12069 | 201.077758789  | 12065 | -0%        |  0.8469 |
 | TPC-H 07       | 7.51439905167 | 451   | 7.47043275833  | 449   | -1%        |  0.3981 |
 | TPC-H 08       | 6.35709524155 | 382   | 6.37827968597  | 383   | +0%        |  0.5325 |
 | TPC-H 09       | 1.52550828457 | 92    | 1.5229023695   | 92    | -0%        |  0.7316 |
 | TPC-H 10       | 2.65954232216 | 160   | 2.66595959663  | 160   | +0%        |  0.7728 |
 | TPC-H 11       | 30.1973304749 | 1812  | 30.213684082   | 1813  | +0%        |  0.6766 |
 | TPC-H 12       | 8.76660633087 | 526   | 8.73143959045  | 524   | -0%        |  0.0000 |
 | TPC-H 13       | 2.27395629883 | 137   | 2.30064845085  | 139   | +1%        |  0.0000 |
 | TPC-H 14       | 45.9030075073 | 2755  | 46.2060508728  | 2773  | +1%        |  0.0000 |
 | TPC-H 15       | 84.651802063  | 5080  | 85.2034759521  | 5113  | +1%        |  0.0000 |
 | TPC-H 16       | 7.06166124344 | 424   | 7.06917524338  | 425   | +0%        |  0.7488 |
 | TPC-H 17       | 4.69012594223 | 282   | 4.70160293579  | 283   | +0%        |  0.7404 |
 | TPC-H 18       | 1.18020284176 | 71    | 1.20913636684  | 73    | +2%        |  0.0763 |
-| TPC-H 19       | 8.94591522217 | 537   | 8.18227481842  | 492   | -9%        |  0.0000 |
 | TPC-H 20       | 27.0636177063 | 1624  | 26.9943580627  | 1620  | -0%        |  0.3391 |
 | TPC-H 21       | 1.18508327007 | 72    | 1.14148116112  | 69    | -4%        |  0.0023 |
 | TPC-H 22       | 14.209446907  | 853   | 14.2542133331  | 856   | +0%        |  0.3561 |
 | geometric mean |               |       |                |       | -0%        |         |
 +----------------+---------------+-------+----------------+-------+------------+---------+
```

I'd shrug off Q19, but the rest does not seem to make a difference big enough to justify an approach that is arguably less intuitive.

Bouncner: Ack.

src/lib/operators/aggregate_hash.cpp
```diff
@@ -215,54 +215,80 @@ void AggregateHash::_aggregate() {
...
               }
-            } else {
-              auto inserted = id_map.try_emplace(position.value(), id_counter);
-              // store either the current id_counter or the existing ID of the value
-              if constexpr (std::is_same_v<AggregateKey, AggregateKeyEntry>) {
-                keys_per_chunk[chunk_id][chunk_offset] = inserted.first->second;
+              ++chunk_offset;
+            });
+          }
+        } else {
```
mrks: Code below here is unchanged, just differently indented.

src/lib/operators/aggregate_hash.cpp
Bouncner: > Would "with a type smaller than" help?.

Yes. And the same for "values with the same size" ... "types with the same size (i.e., 64bit)" ???

mrks: - s/values with the same size/types with the same size/
- this way, we avoid building the id map

src/lib/operators/aggregate_hash.cpp
```diff
@@ -215,54 +215,80 @@ void AggregateHash::_aggregate() {
...
+            const auto chunk_in = input_table->get_chunk(chunk_id);
+            const auto base_segment = chunk_in->get_segment(column_id);
+            ChunkOffset chunk_offset{0};
+            segment_iterate<ColumnDataType>(*base_segment, [&](const auto& position) {
               if constexpr (std::is_same_v<AggregateKey, AggregateKeyEntry>) {
-                keys_per_chunk[chunk_id][chunk_offset] = 0u;
+                if (position.is_null()) {
+                  keys_per_chunk[chunk_id][chunk_offset] = 0;
+                } else {
+                  keys_per_chunk[chunk_id][chunk_offset] = position.value() + 1;
```
mrks: This is double ungood for negative values. Fix it, test it.

src/lib/operators/aggregate_hash.cpp
Bouncner: Ack.

Bouncner:
Is this PR just waiting for a final review?

mrks:
Yes.

src/lib/operators/aggregate_hash.cpp
```diff
@@ -215,54 +215,89 @@ void AggregateHash::_aggregate() {
...
-                                         decltype(allocator)>(allocator);
-        AggregateKeyEntry id_counter = 1u;
-
-        const auto chunk_count = input_table->chunk_count();
-        for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
-          const auto chunk_in = input_table->get_chunk(chunk_id);
-          if (!chunk_in) continue;
+        if constexpr (std::is_same_v<ColumnDataType, int32_t>) {
+          // For values with a type than AggregateKeyEntry, we can use the value itself as an AggregateKeyEntry. We
+          // cannot do this for types with the same size as AggregateKeyEntry as we need tohave a special NULL value.
```
Bouncner: tohave > to have

src/lib/operators/aggregate_hash.cpp
```diff
@@ -215,54 +215,89 @@ void AggregateHash::_aggregate() {
...
+          // By using the value itself, we can save us the effort of building the id_map.
+          for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
+            const auto chunk_in = input_table->get_chunk(chunk_id);
+            const auto base_segment = chunk_in->get_segment(column_id);
+            ChunkOffset chunk_offset{0};
+            segment_iterate<ColumnDataType>(*base_segment, [&](const auto& position) {
+              const auto normalize = [](const int32_t value) {
+                // We need to convert a potentially negative int32_t value into the uint64_t space. We do not care
+                // about preserving the value, just its uniqueness.
+                const auto shifted_value = static_cast<int64_t>(value) - std::numeric_limits<int32_t>::lowest();
```
Bouncner: Not sure if "normalize" is the best term here. And is that an implicit "underflow" you are using here?

src/lib/operators/aggregate_hash.cpp
```diff
@@ -215,54 +215,89 @@ void AggregateHash::_aggregate() {
...
+              if (position.is_null()) {
+                if constexpr (std::is_same_v<AggregateKey, AggregateKeyEntry>) {
+                  keys_per_chunk[chunk_id][chunk_offset] = 0u;
+                } else {
+                  keys_per_chunk[chunk_id][chunk_offset][group_column_index] = 0u;
+                }
               } else {
-                keys_per_chunk[chunk_id][chunk_offset][group_column_index] = inserted.first->second;
+                auto inserted = id_map.try_emplace(position.value(), id_counter);
+                // store either the current id_counter or the existing ID of the value
```
Bouncner: I don't really understand how this comment is connected to the lines below.

mrks: Doesn't really help, deleted the comment.

src/lib/operators/aggregate_hash.cpp
```diff
@@ -215,54 +215,89 @@ void AggregateHash::_aggregate() {
...
-        auto id_map = std::unordered_map<ColumnDataType, AggregateKeyEntry, std::hash<ColumnDataType>, std::equal_to<>,
-                                         decltype(allocator)>(allocator);
-        AggregateKeyEntry id_counter = 1u;
-
-        const auto chunk_count = input_table->chunk_count();
-        for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
-          const auto chunk_in = input_table->get_chunk(chunk_id);
-          if (!chunk_in) continue;
+        if constexpr (std::is_same_v<ColumnDataType, int32_t>) {
+          // For values with a type than AggregateKeyEntry, we can use the value itself as an AggregateKeyEntry. We
```
Bouncner: with a type than

PR #1959, opened 2019-12-11
ReferenceSegment: Use iterable instead of accessor where possible

mrks:
When we are looking at a ReferenceSegment that points at a single segment, it makes more sense to use that segments iterator instead of using accessors:

```diff
 +----------------+----------------+-------+----------------+-------+------------+---------+
 | Benchmark      | prev. iter/s   | runs  | new iter/s     | runs  | change [%] | p-value |
 +----------------+----------------+-------+----------------+-------+------------+---------+
+| TPC-H 01       | 0.591941773891 | 36    | 0.684819757938 | 42    | +16%       |  0.0000 |
 | TPC-H 02       | 68.1781997681  | 4091  | 66.9826126099  | 4019  | -2%        |  0.0202 |
+| TPC-H 03       | 6.9803276062   | 419   | 7.40183210373  | 445   | +6%        |  0.0000 |
 | TPC-H 04       | 6.87540578842  | 413   | 6.71953058243  | 404   | -2%        |  0.0001 |
+| TPC-H 05       | 3.4995496273   | 211   | 3.78297424316  | 227   | +8%        |  0.0000 |
+| TPC-H 06       | 198.962280273  | 11938 | 236.195739746  | 14172 | +19%       |  0.0000 |
 | TPC-H 07       | 7.41563415527  | 445   | 7.39740276337  | 444   | -0%        |  0.7218 |
+| TPC-H 08       | 5.79976272583  | 349   | 6.13998270035  | 369   | +6%        |  0.0000 |
+| TPC-H 09       | 1.49156761169  | 90    | 1.57876753807  | 95    | +6%        |  0.0000 |
 | TPC-H 10       | 2.61136102676  | 157   | 2.67503213882  | 161   | +2%        |  0.0043 |
 | TPC-H 11       | 28.869146347   | 1733  | 28.764081955   | 1726  | -0%        |  0.0064 |
-| TPC-H 12       | 8.78120422363  | 527   | 7.99289464951  | 480   | -9%        |  0.0000 |
+| TPC-H 13       | 2.16691184044  | 131   | 2.31267619133  | 139   | +7%        |  0.0000 |
+| TPC-H 14       | 45.4273376465  | 2726  | 48.926322937   | 2936  | +8%        |  0.0000 |
+| TPC-H 15       | 71.3107147217  | 4279  | 88.6862106323  | 5322  | +24%       |  0.0000 |
+| TPC-H 16       | 6.92698097229  | 416   | 7.28291988373  | 437   | +5%        |  0.0000 |
 | TPC-H 17       | 4.65258264542  | 280   | 4.54031658173  | 273   | -2%        |  0.0002 |
+| TPC-H 18       | 0.77457690239  | 47    | 0.824911534786 | 50    | +6%        |  0.0000 |
 | TPC-H 19       | 8.66394901276  | 521   | 8.84604549408  | 531   | +2%        |  0.0257 |
 | TPC-H 20       | 25.9511432648  | 1558  | 26.3029403687  | 1579  | +1%        |  0.0000 |
 | TPC-H 21       | 1.16310191154  | 70    | 1.18264293671  | 71    | +2%        |  0.0056 |
+| TPC-H 22       | 14.0375976562  | 843   | 16.4996814728  | 990   | +18%       |  0.0000 |
+| geometric mean |                |       |                |       | +5%        |         |
 +----------------+----------------+-------+----------------+-------+------------+---------+
```

@mflueggen - Since you had some exposure to iterators and accessors, do you think you could have a look?

### DIFF
--- a/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
@@ -67,8 +67,8 @@ std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::scan_chunk(ChunkID chunk_i
           auto right_iterable =
               ReferenceSegmentIterable<ColumnDataType, EraseReferencedSegmentType::No>{*right_typed_segment};
 
-          left_iterable.with_iterators([&](auto left_it, const auto left_end) {
-            right_iterable.with_iterators([&](auto right_it, const auto right_end) {
+          left_iterable.with_iterators([&](auto left_it, const auto [[maybe_unused]] left_end) {
+            right_iterable.with_iterators([&](auto right_it, const auto [[maybe_unused]] right_end) {
               if constexpr (std::is_same_v<std::decay_t<decltype(left_it)>,
                                            std::decay_t<decltype(right_it)>>) {  // NOLINT
                 // Either both reference segments use the MultipleChunkIterator (which uses erased accessors anyway)
--- a/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
@@ -67,8 +67,8 @@ std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::scan_chunk(ChunkID chunk_i
           auto right_iterable =
               ReferenceSegmentIterable<ColumnDataType, EraseReferencedSegmentType::No>{*right_typed_segment};
 
-          left_iterable.with_iterators([&](auto left_it, const auto [[maybe_unused]] left_end) {
-            right_iterable.with_iterators([&](auto right_it, const auto [[maybe_unused]] right_end) {
+          left_iterable.with_iterators([&](auto left_it, [[maybe_unused]] const auto left_end) {
+            right_iterable.with_iterators([&](auto right_it, [[maybe_unused]] const auto right_end) {
               if constexpr (std::is_same_v<std::decay_t<decltype(left_it)>,
                                            std::decay_t<decltype(right_it)>>) {  // NOLINT
                 // Either both reference segments use the MultipleChunkIterator (which uses erased accessors anyway)
--- a/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
@@ -7,7 +7,6 @@
 #include "resolve_type.hpp"
 #include "storage/chunk.hpp"
 #include "storage/create_iterable_from_segment.hpp"
-#include "storage/reference_segment/create_iterable_from_segment.hpp"
 #include "storage/reference_segment/reference_segment_iterable.hpp"
 #include "storage/segment_iterables/any_segment_iterable.hpp"
 #include "storage/segment_iterate.hpp"
--- a/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
@@ -7,6 +7,7 @@
 #include "resolve_type.hpp"
 #include "storage/chunk.hpp"
 #include "storage/create_iterable_from_segment.hpp"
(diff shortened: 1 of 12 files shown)


mrks:
Please hold off the review for now. I did not build this in debug mode, where it has some circular include issues.

src/lib/storage/reference_segment/reference_segment_iterable.hpp
```diff
@@ -71,12 +70,9 @@ class ReferenceSegmentIterable : public SegmentIterable<ReferenceSegmentIterable
           if constexpr (std::is_same_v<SegmentType, LZ4Segment<T>>) return;
 
           if constexpr (!std::is_same_v<SegmentType, ReferenceSegment>) {
-            auto accessor = std::make_shared<SegmentAccessor<T, SegmentType>>(typed_segment);
-
-            auto begin = SingleChunkIterator<SegmentAccessor<T, SegmentType>>{accessor, begin_it, begin_it};
-            auto end = SingleChunkIterator<SegmentAccessor<T, SegmentType>>{accessor, begin_it, end_it};
+            const auto segment_iterable = create_iterable_from_segment<T>(typed_segment);
+            segment_iterable.with_iterators(pos_list, functor);
```
mrks: This is the actual change. The rest is just moved code to deal with header include orders :/

src/lib/storage/create_iterable_from_segment.hpp
```diff
@@ -25,85 +38,36 @@ class ReferenceSegmentIterable;
...
-    return FrameOfReferenceSegmentIterable<T>{segment};
-  }
-#endif
+template <typename T, typename Enabled, bool EraseSegmentType = HYRISE_DEBUG>
+auto create_iterable_from_segment(const FrameOfReferenceSegment<T, Enabled>& segment);
+
+// Fix template deduction so that we can call `create_iterable_from_segment<T, false>` on FrameOfReferenceSegments
+template <typename T, bool EraseSegmentType, typename Enabled>
+auto create_iterable_from_segment(const FrameOfReferenceSegment<T, Enabled>& segment) {
+  return create_iterable_from_segment<T, Enabled, EraseSegmentType>(segment);
```
mflueggen: Couldn't this line be moved into the .ipp file?

src/lib/storage/reference_segment/reference_segment_iterable.hpp
mflueggen: This is a great idea.
But I have to ask you on Wednesday why this only works for `EraseReferencedSegmentType::No`.

src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp
```diff
@@ -67,8 +67,8 @@ std::shared_ptr<PosList> ColumnVsColumnTableScanImpl::scan_chunk(ChunkID chunk_i
           auto right_iterable =
               ReferenceSegmentIterable<ColumnDataType, EraseReferencedSegmentType::No>{*right_typed_segment};
 
-          left_iterable.with_iterators([&](auto left_it, const auto left_end) {
-            right_iterable.with_iterators([&](auto right_it, const auto right_end) {
+          left_iterable.with_iterators([&](auto left_it, [[maybe_unused]] const auto left_end) {
```
mflueggen: Cool.

src/lib/storage/create_iterable_from_segment.hpp
mrks: I guess so, but since the ipp file is not its own compilation unit, it doesn't change anything for the compiler. Thus, for the sake of readability, I preferred not to have another forward declaration.

src/lib/storage/reference_segment/reference_segment_iterable.hpp
mrks: To be honest, I didn't look that far. With erased segment types, we basically declare that we stopped caring about performance anyway. I don't think that there is a general reason why this shouldn't work.

Let's see what CI says about it.

mrks:
Here we go. Let me know if you don't have approval rights (or have additional concerns).

mrks:
Ah, Github changed your review to be binding automatically when i changed the rights.  Nice.

mflueggen:
Looks good.

PR #1961, opened 2019-12-12
[DYOD] Sorted between scan

j-tr:
DYOD Seminar Project: Sort Based Query Optimization

adapted between scan to handle ordererd segments faster


mrks:
Did you already run some benchmarks?

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -132,4 +140,51 @@ void ColumnBetweenTableScanImpl::_scan_dictionary_segment(const BaseDictionarySe
...
+    } else {
+      auto segment_iterable = create_iterable_from_segment(typed_segment);
+      segment_iterable.with_iterators(position_filter, [&](auto segment_begin, auto segment_end) {
+        auto sorted_segment_search =
+            SortedSegmentBetweenSearch(segment_begin, segment_end, order_by_mode, predicate_condition,
+                                       boost::get<ColumnDataType>(left_value), boost::get<ColumnDataType>(right_value));
+
+        sorted_segment_search.scan_sorted_segment([&](auto begin, auto end) {
+          size_t output_idx = matches.size();
+
```
Bensk1: Can [this](https://github.com/hyrise/hyrise/pull/1951) happen here?

j-tr: https://github.com/hyrise/hyrise/pull/1961/files#diff-10c2e575a4fe7a7d31742cce4bea1710R176 prevents this

Bensk1: Ah nice, I see.

Another thing: It won’t make much of a difference but we could store the result of the distance calculation in a variable to determine it only once, right? Then, we could check it before we have the resize call even though it will do nothing in that case anyway. However, I doubt that there is any measurable impact :)

j-tr:
the benchmark shows that there is no real performance change compared to the master. this is probably due to the fact that only Query 4, 10 and 14 can profit from our contribution (because they are the only ones that use a between scan on an ordered column as first operation) and the between scan in these queries just contributes minimally to the overall runtime. 
is it legit to change the sorted columns so that e.g. query 6 could benefit from our improvement, since the between scan in this query contributes significantly to the overall runtime but the scan is performed on a currently not sorted column?

mrks:
> only Query 4, 10 and 14 can profit from our contribution (because they are the only ones that use a between scan on an ordered column as first operation)

Why is it necessary for the between scan to be the first operator? Shouldn't `AbstractDereferencedColumnTableScanImpl::_scan_reference_segment` handle that in a way where this also works for following operator? I would expect for Q6, where the BETWEEN on l_shipdate takes ~10%, to also profit

>  is it legit to change the sorted columns so that e.g. query 6 could benefit from our improvement, since the between scan in this query contributes significantly to the overall runtime but the scan is performed on a currently not sorted column?

From a TPC-H perspective, I believe not. The TPC-H defines which columns we may sort on. Also, this would break the l_shipdate-based chunk pruning. What would be an (potentially illegal) option would be to sort the table by one column (to enable pruning) and then sort the each encoded chunk by a second column (to enable the between scan).

Does this make sense?

mrks:
In any case, you could either write a microbenchmark or modify one of the existing queries to show the benefits.

Bensk1:
> but the scan is performed on a currently not sorted column?

Isn't the between scan performed on `l_shipdate` which is actually sorted?

> From a TPC-H perspective, I believe not. The TPC-H defines which columns we may sort on. Also, this would break the l_shipdate-based chunk pruning. What would be an (potentially illegal) option would be to sort the table by one column (to enable pruning) and then sort the each encoded chunk by a second column (to enable the between scan).

Which columns are allowed to be sorted?

mrks:
Mmmh, after another look into the [specification](http://www.tpc.org/tpc_documents_current_versions/pdf/tpc-h_v2.17.1.pdf), more specifically, 1.5.2, maybe all orders are allowed. Within a compressed chunk, we might be allowed to do whatever we want. For an entire table, I would consider that the tables are supposed to be updated. Assuming that new entries have a higher l_orderkey is reasonable (even if not fully sequential), but I would feel uncomfortable sorting by, e.g., `l_discount`.

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
MasterCarl: Updated in 34e0d18e5a8e9f39ea1e0ff0bb33dc4e9397198b

Boffmann:
> Did you already run some benchmarks?

Yes, we developed on an entire new branch and added it to this PR. This includes benchmarks showing that our solution is indeed faster compared to the intuitive one.

src/lib/operators/table_scan/sorted_segment_between_search.hpp
```diff
@@ -0,0 +1,193 @@
...
+    difference_type size = std::distance(it_first, it_last), bound = 1;
+
+    if (_is_nulls_first) {
+      while (bound < size && (it_first + bound)->is_null()) {
+        bound *= 2;
+      }
+
+      auto end = it_first + std::min(bound, size);
+      _begin = std::lower_bound(it_first + (bound / 2), end, false,
+                                [](const auto& segment_position, const auto& _) { return segment_position.is_null(); });
```
bakoe: (I believe here we should use `std::upper_bound - 1` as well, see comment to line 60..61 in this same file)

src/lib/operators/table_scan/sorted_segment_between_search.hpp
```diff
@@ -0,0 +1,193 @@
...
+    } else {
+      while (bound < size && (it_first + (size - bound))->is_null()) {
+        bound *= 2;
+      }
+
+      auto start = it_first + (size - std::min(bound, size));
+      _end = std::lower_bound(start, it_first + (size - bound / 2), true,
+                              [](const auto& segment_position, const auto& _) { return !segment_position.is_null(); });
```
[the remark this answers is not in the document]
bakoe: Hmm … I think that my description of the `std::lower_bound` call is wrong and should instead be: 

``` diff
- … for which the following condition holds true: The value is not NULL.
+ … for which the following condition does not hold true: The value is not NULL. 
+ (i. e. the first element which is NULL)
```

Thus, it seems like your implementation is perfectly fine. However, I'll leave the review comment here so that you can still have a look at it and double-check. I hope that's okay for you and doesn't lead to more confusion on your side as well. 😉

src/lib/operators/table_scan/sorted_segment_between_search.hpp
bakoe: (Update from https://github.com/hyrise/hyrise/pull/1961#discussion_r367366937 applies here as well)

mrks:
Good comments for the review!

Approving a PR means "everything is a-ok for me". In some cases, you might approve if you have trivial comments like "here's a typo" or "it would be minimally nicer, but equally correct, if you did xyz". Two of your comments sound like things that should be finally resolved before the PR is approved.

I haven't looked at the entire code yet, but have you considered `std::equal_range`? That might save you up to half of the binary searches.

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -132,4 +140,51 @@ void ColumnBetweenTableScanImpl::_scan_dictionary_segment(const BaseDictionarySe
...
 
+void ColumnBetweenTableScanImpl::_scan_sorted_segment(const BaseSegment& segment, const ChunkID chunk_id,
+                                                      PosList& matches,
+                                                      const std::shared_ptr<const PosList>& position_filter,
+                                                      const OrderByMode order_by_mode) const {
+  resolve_data_and_segment_type(segment, [&](const auto type, const auto& typed_segment) {
+    using ColumnDataType = typename decltype(type)::type;
+
+    if constexpr (std::is_same_v<std::decay_t<decltype(typed_segment)>, ReferenceSegment>) {
+      Fail("Expected ReferenceSegments to be handled before calling this method");
```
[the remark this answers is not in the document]
birneamstiel: @bakoe since this function is only called from _scan_non_reference_segment within this file (whose caller resolves the references already, see here and here) this here is fine to assume, I think...

README.md
```diff
@@ -156,4 +156,7 @@ Contact: firstname.lastname@hpi.de
 -   Lukas     Wenzel
 -   Fabian    Wiebe
 -   Tim       Zimmermann
+-   Justin    Trautmann
```
Bouncner: Please sort.

src/benchmark/operators/table_scan_sorted_benchmark.cpp
```diff
@@ -105,35 +118,55 @@ std::shared_ptr<TableWrapper> create_table(const DataType data_type, const int t
 
 void BM_TableScanSorted(
     benchmark::State& state, const int table_size, const double selectivity, const EncodingType encoding_type,
-    const std::string& mode,
+    const std::string& mode, const bool is_between_scan,
```
Bouncner: const

src/lib/operators/table_scan/sorted_segment_between_search.hpp
```diff
@@ -0,0 +1,193 @@
...
+    // If no null values are present
+    if (_is_nulls_first && !it_first->is_null()) {
+      return;
+    }
+    if (!_is_nulls_first && !(it_last - 1)->is_null()) {
+      return;
+    }
+
+    using difference_type = typename std::iterator_traits<IteratorType>::difference_type;
+    difference_type size = std::distance(it_first, it_last), bound = 1;
```
Bouncner: Please don't write on single line. Also, `size` and `bound` don't really tell what they do and `size` can be const.
Could you also add a comment explaining why you use exponential search here?

src/lib/operators/table_scan/sorted_segment_between_search.hpp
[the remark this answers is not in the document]
Boffmann: Thank you very much for your deep thoughts about lower_bound and upper_bound and our algorithm. I checked it three times now (Ascending, Descending, AscendingNullsLast) and in my opinion, there should be no faults. I will discuss this with my teammates again, though, because you never know :)

Boffmann:
> Good comments for the review!
> 
> Approving a PR means "everything is a-ok for me". In some cases, you might approve if you have trivial comments like "here's a typo" or "it would be minimally nicer, but equally correct, if you did xyz". Two of your comments sound like things that should be finally resolved before the PR is approved.
> 
> I haven't looked at the entire code yet, but have you considered `std::equal_range`? That might save you up to half of the binary searches.

Regarding [cppreference.com](https://en.cppreference.com/w/cpp/algorithm/equal_range), std::equal_range returns all elements equivalent to `value` in range `[first, last)`. It therefore includes `first` but excludes `last` which would be, in my opinion, equivalent to:

```
first = std::lower_bound(first, last, value);
last = std::upper_bound(first, last, value);
```

and is therefore only suitable for `PredicateCondition::BetweenInclusive` since e.g. `PredicateCondition::BetweenUpperExclusive` requires

```
first = std::lower_bound(first, last, value);
last = std::lower_bound(first, last, value);
```

which can not be represented by std::equal_range, can it?

mrks:
Ah, yes, not sure what I was thinking there. However, what should be possible is to first calculate the lower bound and then start the search for the upper bound at that iterator, not at begin.

src/lib/operators/table_scan/sorted_segment_between_search.hpp
```diff
@@ -0,0 +1,203 @@
...
+#include "constant_mappings.hpp"
+#include "types.hpp"
+
+namespace opossum {
+
+// Generic class which handles the actual scanning of a sorted segment
+template <typename IteratorType, typename SearchValueType>
+class SortedSegmentBetweenSearch {
+ public:
+  SortedSegmentBetweenSearch(IteratorType begin, IteratorType end, const OrderByMode& order_by, const bool nullable,
```
Bouncner: Let's take a look on this file tomorrow. At first sight, it looks like a copy with few modifications of `sorted_segment_search.hpp`.

MasterCarl: Yes, that was indeed our initial approach.

Bouncner: Makes sense for development to start off quick, but we should make sure that we don't create unnecessary copies.
There is also another reason, we can talk about tomorrow.

MasterCarl:
> Ah, yes, not sure what I was thinking there. However, what should be possible is to first calculate the lower bound and then start the search for the upper bound at that iterator, not at begin.

Thanks for the suggestion @mrks, starting from the lower bound should indeed help us. We did have a function that performed a binary search akin to `std::equal_range` for a while, but the performance seemed to be worse than the combined upper and lower bound searches.

I'm not sure if using `std::equal_range` would have been beneficial anyway since the [possible implementation](https://en.cppreference.com/w/cpp/algorithm/equal_range) consists of `std::lower_bound` and `std::upper_bound` calls anyway.

src/benchmark/operators/table_scan_sorted_benchmark.cpp
MasterCarl: What do you mean by that?

src/lib/operators/table_scan/sorted_segment_between_search.hpp
```diff
@@ -31,6 +31,15 @@ class SortedSegmentBetweenSearch {
 
  private:
   void _exponential_search_for_nulls(IteratorType it_first, IteratorType it_last) {
+    // Uses exponential search to find lower_bound of null values to exclude them from further scanning
```
Bouncner: Use trailing dots and probabaly better a multiline comment `/* */` here.

src/benchmark/operators/table_scan_sorted_benchmark.cpp
Bouncner: I think I wanted to say that you missed using a `const` somewhere and missed the exact line.

Boffmann:
@birneamstiel 
@bakoe
@jonasnoki 

The following changed since you last reviewed our PR:

* We merged sorted_segment_search.hpp and sorted_segment_between_search.hpp in order to reduce code duplication
* We used better names for variables in exponential_search_for_nulls
* We added and adapted some comments

Could you please review these changes as well? Thank you!

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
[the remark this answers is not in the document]
bakoe: I see – can be considered resolved, to my mind. 🙂

bakoe:
> @birneamstiel
> @bakoe
> @jonasnoki
> 
> The following changed since you last reviewed our PR:
> 
> * We merged sorted_segment_search.hpp and sorted_segment_between_search.hpp in order to reduce code duplication
> * We used better names for variables in exponential_search_for_nulls
> * We added and adapted some comments
> 
> Could you please review these changes as well? Thank you!

@Boffmann Sorry for the late reply – I did have another look and the changes you did look good to me. Especially the explanatory comment added to `SortedSegmentSearch` informs about the considerations you did regarding `std::equal_range` quite well. 

Furthermore, my questions/concerns from the last review have been resolved. :+1:

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -34,59 +96,138 @@ class SortedSegmentSearch {
...
+      if ((_end - 1)->value() > *_second_search_value || _begin->value() < _first_search_value) {
+        _begin = _end;
+        return;
+      }
+    }
+
+    // This implementation uses behaviour which resembles std::equal_range's
+    // behaviour. However, std::equal_range returns all elements in range
+    // [first, last), which describes only the PredicateCondition::BetweenInclusive case.
+    // For the other PredicateConditions, different borders are required.
```
Bouncner: borders > bounds?

mrks:
My guess is that the performance regression for Q6 comes from not exploiting the dictionary any more. Instead, the binary search seems to decompress all values it looks at, which will be slower the fewer distinct values we have.

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -35,11 +37,16 @@ std::string ColumnBetweenTableScanImpl::description() const { return "ColumnBetw
...
-    _scan_dictionary_segment(*dictionary_segment, chunk_id, matches, position_filter);
+  const auto ordered_by = _in_table->get_chunk(chunk_id)->ordered_by();
+  if (ordered_by && ordered_by->first == _column_id) {
+    _scan_sorted_segment(segment, chunk_id, matches, position_filter, ordered_by->second);
   } else {
-    _scan_generic_segment(segment, chunk_id, matches, position_filter);
+    // Select optimized or generic scanning implementation based on segment type
+    if (const auto* dictionary_segment = dynamic_cast<const BaseDictionarySegment*>(&segment)) {
+      _scan_dictionary_segment(*dictionary_segment, chunk_id, matches, position_filter);
+    } else {
```
mrks: My guess is that the performance regression for Q6 comes from not exploiting the dictionary any more. Instead, the binary search seems to decompress all values it looks at, which will be slower the fewer distinct values we have.

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -15,16 +17,76 @@ namespace opossum {
...
         _end{end},
         _predicate_condition{predicate_condition},
-        _search_value{search_value},
+        _first_search_value{search_value},
+        _second_search_value{std::nullopt},
+        _nullable{nullable},
         _is_ascending{order_by == OrderByMode::Ascending || order_by == OrderByMode::AscendingNullsLast},
-        _is_nulls_first{order_by == OrderByMode::Ascending || order_by == OrderByMode::Descending} {}
+        _is_nulls_first{order_by == OrderByMode::Ascending || order_by == OrderByMode::Descending},
+        _is_between_scan{false} {}
```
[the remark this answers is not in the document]
Boffmann: Yes you're right. Fixed

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -52,11 +52,12 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
    public:
     explicit Iterator(const ValueIterator& value_it, const NullValueIterator& null_value_it,
-                      const EndPositionIterator& end_position_it, const ChunkOffset start_position)
+                      const EndPositionIterator& end_position_it, const EndPositionIterator& end_position_begin_it,
+                      const ChunkOffset start_position)
         : _value_it{value_it},
           _null_value_it{null_value_it},
           _end_position_it{end_position_it},
-          _end_position_it_begin{end_position_it},
+          _end_position_it_begin{end_position_begin_it},
```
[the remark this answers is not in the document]
j-tr: it's the begin iterator of the end positions. imo member variable and input variable shouldn't be named differently but end_position_it_begin describes pretty much what it is.

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -151,9 +151,10 @@ void ColumnBetweenTableScanImpl::_scan_sorted_segment(const BaseSegment& segment
     } else {
       auto segment_iterable = create_iterable_from_segment(typed_segment);
       segment_iterable.with_iterators(position_filter, [&](auto segment_begin, auto segment_end) {
-        auto sorted_segment_search =
-            SortedSegmentSearch(segment_begin, segment_end, order_by_mode, _column_is_nullable, predicate_condition,
-                                boost::get<ColumnDataType>(left_value), boost::get<ColumnDataType>(right_value));
+        auto typed_left_value = boost::get<ColumnDataType>(left_value);
+        auto typed_right_value = boost::get<ColumnDataType>(right_value);
```
Bouncner: const?

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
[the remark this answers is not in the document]
mrks: I forgot that we are looking at RLE, which has a thing called "end positions". That should only happen when looking at a diff and won't be an issue when actually reading the code.

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -132,4 +139,51 @@ void ColumnBetweenTableScanImpl::_scan_dictionary_segment(const BaseDictionarySe
...
+    if constexpr (std::is_same_v<std::decay_t<decltype(typed_segment)>, ReferenceSegment>) {
+      Fail("Expected ReferenceSegments to be handled before calling this method");
+    } else {
+      auto segment_iterable = create_iterable_from_segment(typed_segment);
+      segment_iterable.with_iterators(position_filter, [&](auto segment_begin, auto segment_end) {
+        auto sorted_segment_search =
+            SortedSegmentSearch(segment_begin, segment_end, order_by_mode, _column_is_nullable, predicate_condition,
+                                boost::get<ColumnDataType>(left_value), boost::get<ColumnDataType>(right_value));
+
+        sorted_segment_search.scan_sorted_segment([&](auto begin, auto end) {
```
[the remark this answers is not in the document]
Boffmann: Yes we can and I did it in commit `906a7625`.

Just for your information: `column_between_table_scan_impl.cpp` and `column_vs_value_table_scan_impl.cpp` could be merged even further (see `_scan_non_reference_segment` for example. I would not do this in this PR though because its not the context of this PR. I would rather focus on seminar related topics)

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -15,16 +17,76 @@ namespace opossum {
...
+     * Uses exponential search to find lower_bound of null values to exclude them from further scanning.
+     * This version of exponential search first reduces the range where the search for NULL values will be performed
+     * and performs a binary search for the exact bound within this range afterwards.
+
+     * Since the null values are either in the beginning or the end of each segment (depending on
+     * how they are ordered) and there are typically very few null values, the amount of steps
+     * used in exponential_search are typically less than the amount of steps taken using only
+     * binary search.
+     */
+  void _exponential_search_for_nulls(IteratorType it_first, IteratorType it_last) {
```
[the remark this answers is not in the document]
Boffmann: I will do some benchmarks as soon as I find the time

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
Bouncner: That might be true for now but that should be no issue once we fix the iterators. Do you see this as an actual issue for this PR?

mrks: Mmh, I originally thought that we should do the binary search on the dictionary, not on the values. However, I just realized that this does not necessarily reduce the number of string comparisons for input segments that are already filtered.

Could you have a look if this is affected by the problem and if the iterator trait hack helps here?

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -34,59 +94,130 @@ class SortedSegmentSearch {
...
     }
 
     // clang-format off
     if (_is_ascending) {
       switch (_predicate_condition) {
-        case PredicateCondition::GreaterThanEquals: _begin = _get_first_bound(); return;
-        case PredicateCondition::GreaterThan: _begin = _get_last_bound(); return;
-        case PredicateCondition::LessThanEquals: _end = _get_last_bound(); return;
-        case PredicateCondition::LessThan: _end = _get_first_bound(); return;
+        case PredicateCondition::GreaterThanEquals: _begin = _get_first_bound(_first_search_value, _begin, _end); return; // NOLINT
```
mrks: Doing only one thing per line prevents people from missing the return:

```
case PredicateCondition::GreaterThanEquals:
  _begin = _get_first_bound(_first_search_value, _begin, _end);
  return;
```

Jenkinsfile
```diff
@@ -47,19 +47,14 @@ try {
   node('linux') {
     def oppossumCI = docker.image('hyrise/opossum-ci:19.10');
     oppossumCI.pull()
-    // create ccache volume on host using:
-    // mkdir /mnt/ccache; mount -t tmpfs -o size=200G none /mnt/ccache
-    // or add it to /etc/fstab:
-    // tmpfs  /mnt/ccache tmpfs defaults,size=200G  0 0
 
-    oppossumCI.inside("-u 0:0 -v /mnt/ccache:/ccache -e \"CCACHE_DIR=/ccache\" -e \"CCACHE_MAXSIZE=200GB\" -e \"CCACHE_SLOPPINESS=file_macro,pch_defines,time_macros\" -e\"CCACHE_DEPEND=1\" -e\"CCACHE_NOHASHDIR=1\" -e\"CCACHE_DEBUG=1\" --privileged=true") {
+    oppossumCI.inside("-u 0:0 --privileged=true") {
```
mrks: You can undo these changes once you merged with the current master.

src/lib/operators/table_scan/abstract_dereferenced_column_table_scan_impl.hpp
```diff
@@ -43,6 +43,40 @@ class AbstractDereferencedColumnTableScanImpl : public AbstractTableScanImpl {
 
   const std::shared_ptr<const Table> _in_table;
   const ColumnID _column_id;
+
+  template <typename IteratorType>
+  void _handle_search_results(IteratorType begin, IteratorType end, const ChunkID chunk_id, PosList& matches,
```
mrks: I don't like the verb "handle". It doesn't really convey any information. Maybe add_rows_to_matches?

src/lib/operators/table_scan/abstract_dereferenced_column_table_scan_impl.hpp
```diff
@@ -43,6 +43,40 @@ class AbstractDereferencedColumnTableScanImpl : public AbstractTableScanImpl {
...
+                              const std::shared_ptr<const PosList>& position_filter) const {
+    if (begin == end) return;
+
+    // General note: If the predicate is NotEquals, there might be two matching ranges. scan_sorted_segment
+    // combines these two ranges into a single one via boost::join(range_1, range_2).
+    // See sorted_segment_search.hpp for further details.
```
mrks: This method is only used by sorted segment searches, correct? Move to `sorted_segment_search.hpp` (and remove this line)?

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -34,59 +94,130 @@ class SortedSegmentSearch {
...
+      _begin = _end;
+      return;
+    }
+
+    // This implementation uses behaviour which resembles std::equal_range's
+    // behaviour since it also calculates a range containing all elements between
+    // the first and second search value. However, std::equal_range returns all
+    // elements in range [first, last), which describes only the
+    // PredicateCondition::BetweenInclusive case. For the other PredicateConditions,
+    // different borders are required.
```
mrks: Is this really correct? If I am not mistaken, even BetweenInclusive cannot be handled by equal_range.

How about this: "However, std::equal_range compares the values using the equals operator and a single comparison value. We need a range on the comparison side, too."

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -34,59 +94,130 @@ class SortedSegmentSearch {
...
   }
 
+  // This function sets the offset(s) that delimit the result set based on the predicate condition and the sort order
+  void _set_begin_and_end_positions_for_between_scan() {
+    DebugAssert(_second_search_value, "Second Search Value must be set for between scan");
+    if (_begin == _end) return;
+
+    auto first_value = _begin->value();
+    auto last_value = (_end - 1)->value();
+    if (!_is_ascending) std::swap(first_value, last_value);
```
mrks: Please explain what you are doing here. Also, it's probably easier if the whole swapping part is limited to a single block of code (see "exchange predicate conditions and search values if descending")

src/lib/operators/table_scan/sorted_segment_search.hpp
MasterCarl: Good catch! I added a note that we will be comparing to two different search values, if two are given.

MasterCarl:
@mrks Cleanup is done, but the benchmarks are still pending.

Jenkinsfile
mrks: According to Github, this still modifies the Jenkinsfile.

j-tr:
We performed Microbenchmarks for the Between search on this branch, this branch without the exponential search, this branch with jan's iterator traits and this branch with iterator traits and without exponential search. 
Looks like the exponential search leads to slightly better results but i'm still not sure if it's worth the additional complexity. 

Unbenannt
Unbenannt4
Unbenannt2
Unbenannt3

MasterCarl:
Looking at the graphs again, it seems to me that the last variety (PR + bensk1/make_iterators_random_access without the exponential search) yields the best results overall.

Bouncner:
> Looking at the graphs again, it seems to me that the last variety (PR + bensk1/make_iterators_random_access without the exponential search) yields the best results overall.

But where did you actually measure the exponential search here?

Boffmann:
> > Looking at the graphs again, it seems to me that the last variety (PR + bensk1/make_iterators_random_access without the exponential search) yields the best results overall.
> 
> But where did you actually measure the exponential search here?

We added null values for our MicroBenchmarks. This is the result:

`With exponential search`
with_exp

`Without exponential search`
without_exp

Bouncner:
Am I wrong when I assume that the shuffled tests should yield the same results and none of your changes is being executed?

If so: what’s going on here?

Bouncner:
I ran the between benchmarks on a physical machine with clang (GCC triggered the page layout flakiness for the benchmark with your branches). To have a fair comparison of both null-searches, I used @Bensk1 iterator fix (the exponential search uses iterator's `advance()`, which works fine with the master, and the alternative uses `std::lower_bound`, which requires the fix).
I only evaluated dictionary encoding with integers.

The two graphs below that even though clang's binaries are less flaky than GCC in this case, they are still flaky as can be been in the Shuffled case. Both executions should trigger the exact same code. However, when looking at both sorted cases, `std::lower_bound` wins.

Hence, I recommend to remove the exponential search.

[actual.pdf](https://github.com/hyrise/hyrise/files/4235949/actual.pdf)

[normalized.pdf](https://github.com/hyrise/hyrise/files/4235964/normalized.pdf)

Bouncner:
So, here are some benchmarks that we have run. Sorry for the delay, this is on me and not the seminar group.

So first, we ran TPC-H with the iterator fix. GCC and Clang on nemea:

Clang:
```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.5978752374649048 | 96    | 1.6469690799713135 | 99    | +3%        |  0.0000 |
 | TPC-H 02       | 70.1949462890625   | 4212  | 71.59683990478516  | 4296  | +2%        |  0.0031 |
 | TPC-H 03       | 7.177587985992432  | 431   | 7.49026346206665   | 450   | +4%        |  0.0000 |
 | TPC-H 04       | 6.276573181152344  | 377   | 6.461714744567871  | 388   | +3%        |  0.0000 |
+| TPC-H 05       | 3.395240306854248  | 204   | 3.5651957988739014 | 215   | +5%        |  0.0000 |
-| TPC-H 06       | 283.6569519042969  | 17020 | 261.6971130371094  | 15702 | -8%        |  0.0000 |
 | TPC-H 07       | 7.624208927154541  | 458   | 7.4563493728637695 | 448   | -2%        |  0.0004 |
 | TPC-H 08       | 5.5869975090026855 | 336   | 5.356699466705322  | 322   | -4%        |  0.0000 |
 | TPC-H 09       | 1.6637862920761108 | 100   | 1.648346185684204  | 99    | -1%        |  0.0470 |
 | TPC-H 10       | 3.3534786701202393 | 202   | 3.337794780731201  | 201   | -0%        |  0.4752 |
 | TPC-H 11       | 28.772266387939453 | 1727  | 28.785558700561523 | 1728  | +0%        |  0.7105 |
 | TPC-H 12       | 14.020085334777832 | 842   | 13.857807159423828 | 832   | -1%        |  0.0000 |
-| TPC-H 13       | 3.0886049270629883 | 186   | 2.5456087589263916 | 153   | -18%       |  0.0000 |
 | TPC-H 14       | 48.606300354003906 | 2917  | 47.6070556640625   | 2857  | -2%        |  0.0000 |
 | TPC-H 15       | 114.03759765625    | 6843  | 112.51571655273438 | 6751  | -1%        |  0.0000 |
 | TPC-H 16       | 8.712960243225098  | 523   | 8.619359016418457  | 518   | -1%        |  0.0000 |
+| TPC-H 17       | 4.246062278747559  | 255   | 4.948129653930664  | 297   | +17%       |  0.0000 |
+| TPC-H 18       | 1.1227768659591675 | 68    | 1.3143688440322876 | 79    | +17%       |  0.0000 |
 | TPC-H 19       | 9.297355651855469  | 558   | 9.370254516601562  | 563   | +1%        |  0.1212 |
 | TPC-H 20       | 25.993438720703125 | 1560  | 26.199663162231445 | 1572  | +1%        |  0.0002 |
 | TPC-H 21       | 1.1182652711868286 | 68    | 1.1019651889801025 | 67    | -1%        |  0.0732 |
 | TPC-H 22       | 15.355514526367188 | 922   | 15.26132869720459  | 916   | -1%        |  0.0010 |
 | geometric mean |                    |       |                    |       | +0%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

GCC:
```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.3420400619506836 | 81    | 1.3390287160873413 | 81    | -0%        |  0.4199 |
 | TPC-H 02       | 80.52922058105469  | 4832  | 80.59638214111328  | 4836  | +0%        |  0.8988 |
 | TPC-H 03       | 8.03616714477539   | 483   | 8.240156173706055  | 495   | +3%        |  0.0000 |
 | TPC-H 04       | 7.071835994720459  | 425   | 7.11731481552124   | 428   | +1%        |  0.2092 |
 | TPC-H 05       | 3.9402852058410645 | 237   | 3.9002909660339355 | 235   | -1%        |  0.2505 |
 | TPC-H 06       | 267.6904602050781  | 16062 | 271.2540283203125  | 16276 | +1%        |  0.0000 |
 | TPC-H 07       | 7.968039035797119  | 479   | 7.985957145690918  | 480   | +0%        |  0.7279 |
+| TPC-H 08       | 5.915480613708496  | 355   | 6.212149143218994  | 373   | +5%        |  0.0000 |
 | TPC-H 09       | 1.6373940706253052 | 99    | 1.6441235542297363 | 99    | +0%        |  0.3458 |
 | TPC-H 10       | 3.4136078357696533 | 205   | 3.432582378387451  | 206   | +1%        |  0.3607 |
 | TPC-H 11       | 31.2896785736084   | 1878  | 30.890840530395508 | 1854  | -1%        |  0.0000 |
 | TPC-H 12       | 14.641908645629883 | 879   | 14.554943084716797 | 874   | -1%        |  0.0001 |
 | TPC-H 13       | 2.3987531661987305 | 144   | 2.4142870903015137 | 145   | +1%        |  0.0070 |
 | TPC-H 14       | 49.7617073059082   | 2986  | 50.26568603515625  | 3016  | +1%        |  0.0000 |
 | TPC-H 15       | 114.22981262207031 | 6854  | 115.52469635009766 | 6932  | +1%        |  0.0000 |
 | TPC-H 16       | 8.13100528717041   | 488   | 8.377473831176758  | 503   | +3%        |  0.0000 |
+| TPC-H 17       | 5.131563663482666  | 308   | 5.395016670227051  | 324   | +5%        |  0.0000 |
 | TPC-H 18       | 1.19792902469635   | 72    | 1.2220771312713623 | 74    | +2%        |  0.0021 |
 | TPC-H 19       | 9.993735313415527  | 600   | 10.28275203704834  | 617   | +3%        |  0.0000 |
 | TPC-H 20       | 29.020416259765625 | 1742  | 28.863025665283203 | 1732  | -1%        |  0.0049 |
 | TPC-H 21       | 1.196044683456421  | 72    | 1.2394263744354248 | 75    | +4%        |  0.0000 |
 | TPC-H 22       | 17.132539749145508 | 1029  | 17.379199981689453 | 1043  | +1%        |  0.0000 |
 | geometric mean |                    |       |                    |       | +1%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

We ran both compilers as a sanity check. Actually, only queries 4, 5, and 10 are changing at all due to the between scan. All other changes are noise.
For Q4, Q5, Q10, the between searches are on the orders table and two of three scans improve by a factor of 2x (just compared the PQP visualizations, no thorough analysis). But all these scans are in the range of microseconds, so they play a role for the runtime.

As benchmarks show that a sorted scan is also faster with defunct random iterators, we are fine for now with TPC-H.

*TPC-H Q6:*
Since we do not pass the sorted flag right now, the between sorted search is not used (it's the third filter in the query). This will be handled in #2000 (@j-tr , @MasterCarl , @Boffmann : please keep an eye on the changes to Q6).

To get an idea of the potential, we ran a modified Q6) where the lineitem table is sorted by `l_discount` (GCC, nemea):
```diff
 +----------------+-------------------+-------+-------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s      | runs  | new iter/s        | runs  | change [%] | p-value |
 +----------------+-------------------+-------+-------------------+-------+------------+---------+
+| TPC-H 06       | 290.9791259765625 | 17460 | 386.4721374511719 | 23189 | +33%       |  0.0000 |
+| geometric mean |                   |       |                   |       | +33%       |         |
 +----------------+-------------------+-------+-------------------+-------+------------+---------+
```

The visualizations for the modified Q6:
Screen Shot 2020-02-26 at 13 49 50
Screen Shot 2020-02-26 at 13 50 07

Bouncner:
Ready to go from my side. @mrks , @bakoe : anything to add?

mrks:
> +33%

That. I want it.

PR #1965, opened 2019-12-15
CI: RelWithDebInfo for TSAN

mrks:
* Hopefully fixes the MVCC Delete Plugin Freezes

* Should help us in tracking down the spurious CI fails. Adds 5 minutes to those steps, but since we are usually waiting for clang-tidy anyway, it should not be noticeable.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -67,8 +67,8 @@ try {
           mkdir clang-release-addr-ub-sanitizers && cd clang-release-addr-ub-sanitizers && cmake -DCI_BUILD=ON -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DENABLE_ADDR_UB_SANITIZATION=ON .. &\
           mkdir clang-release && cd clang-release && cmake -DCI_BUILD=ON -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ .. &\
           mkdir clang-release-addr-ub-sanitizers-no-numa && cd clang-release-addr-ub-sanitizers-no-numa && cmake -DCI_BUILD=ON -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DENABLE_ADDR_UB_SANITIZATION=ON -DENABLE_NUMA_SUPPORT=OFF .. &\
-          mkdir clang-release-thread-sanitizer && cd clang-release-thread-sanitizer && cmake -DCI_BUILD=ON -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DENABLE_THREAD_SANITIZATION=ON .. &\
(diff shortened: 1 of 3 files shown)


src/test/plugins/mvcc_delete_plugin_system_test.cpp
```diff
@@ -245,7 +245,7 @@ TEST_F(MvccDeletePluginSystemTest, CheckPlugin) {
       // To increase the global _last_commit_id, we need to execute a transaction with read-write operators
       // We perform some dummy updates so that the table is unmodified and the validation routine does not complain
       auto pipeline =
-          SQLPipelineBuilder{std::string{"UPDATE " + _t_name_test + " SET number = number WHERE number = 0"}}
+          SQLPipelineBuilder{std::string{"UPDATE " + _t_name_test + " SET number = number WHERE number = -1"}}
```
mrks: Previously, this could have a row-level conflict with the logical delete, leading to 

```
../../../src/test/plugins/mvcc_delete_plugin_system_test.cpp:270: Failure
Expected: (attempts_remaining) > (-1), actual: -1 vs -1
```

Jenkinsfile
```diff
@@ -193,25 +193,25 @@ try {
...
-          stage("clang-release:thread-sanitizer") {
+        }, clangRelWithDebInfoThreadSanitizer: {
+          stage("clang-relwithdebinfo:thread-sanitizer") {
             if (env.BRANCH_NAME == 'master' || full_ci) {
-              sh "export CCACHE_BASEDIR=`pwd`; cd clang-release-thread-sanitizer && make hyriseTest hyriseSystemTest hyriseBenchmarkTPCH -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 6))"
-              sh "TSAN_OPTIONS=\"history_size=7 suppressions=resources/.tsan-ignore.txt\" ./clang-release-thread-sanitizer/hyriseTest clang-release-thread-sanitizer"
-              sh "TSAN_OPTIONS=\"history_size=7 suppressions=resources/.tsan-ignore.txt\" ./clang-release-thread-sanitizer/hyriseSystemTest ${exclude_in_sanitizer_builds} clang-release-thread-sanitizer"
-              sh "TSAN_OPTIONS=\"history_size=7 suppressions=resources/.tsan-ignore.txt\" ./clang-release-thread-sanitizer/hyriseBenchmarkTPCH -s .01 --verify -r 100 --scheduler --clients 10"
+              sh "export CCACHE_BASEDIR=`pwd`; cd clang-relwithdebinfo-thread-sanitizer && make hyriseTest hyriseSystemTest hyriseBenchmarkTPCH -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 6))"
```
Bensk1: CamelCase?

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -45,6 +45,7 @@ void MvccDeletePlugin::_logical_delete_loop() {
         // Calculate metric 1 – Chunk invalidation level
         const double invalidated_rows_ratio = static_cast<double>(chunk->invalid_row_count()) / chunk->size();
         const bool criterion1 = (DELETE_THRESHOLD_PERCENTAGE_INVALIDATED_ROWS <= invalidated_rows_ratio);
+        std::cout << "Chunk " << chunk_id << " - invalidated_rows_ratio: " << invalidated_rows_ratio << std::endl;
```
Bensk1: Could we guard these prints with `#ifdef` and let the CI set a compiler flag?

src/test/plugins/mvcc_delete_plugin_system_test.cpp
```diff
@@ -239,13 +239,23 @@ TEST_F(MvccDeletePluginSystemTest, CheckPlugin) {
 
   // (11) Prepare clean-up of chunk 3
   {
+    // Wait for the previous updates to finish
+    {
+      auto attempts_remaining = max_attempts;
+      while (_counter < INITIAL_CHUNK_COUNT * CHUNK_SIZE - 2) {
```
Bensk1: Add an explanation for the `-2`?

src/test/plugins/mvcc_delete_plugin_system_test.cpp
```diff
@@ -239,13 +239,23 @@ TEST_F(MvccDeletePluginSystemTest, CheckPlugin) {
...
+        std::this_thread::sleep_for(100);
+      }
+      // Check that we have not given up
+      ASSERT_GT(attempts_remaining, -1);
+    }
+
```
Bensk1: Is this part going to stay after we fixed the TSAN issue?

mrks:
> mrks removed the request for review from Bensk1 5 hours ago

Still playing around here, sorry.

Jenkinsfile
mrks: Not a fan of capital case in folder names (differing auto-completion depending on shell settings, glob handling, ...)

src/test/plugins/mvcc_delete_plugin_system_test.cpp
mrks: Yes. To check whether an async process has done its job, we either need to add hooks (which I don't like in the lib code if their sole purpose is to facilitate testing) or wait until we have externally visible effects.

PR #1881, opened 2019-10-14
Make build process faster

mrks:
TODOs:
- [x] add link to blog post in README.md
- [x] check if ccache still makes sense, [find fix for pch issue](https://stackoverflow.com/questions/59095795/gcc-precompiled-header-not-deterministic)
- [x] adapt new_ naming according to whatever we decide in #1877 
- [x] solve JIT issues
- [x] get cmake 3.16 final into the CI image
- [x] verify that asan/ubsan/tsan still work
- [x] decide on which CI steps should use unity builds
- [x] check performance 
- [x] Once unmodified branch runs on Catalina, fix Mac CI build
- [x] Update Docker image with gcc 9, clang, cmake
- [x] Add notice for non-3.16 / non unity build users to CMakeLists.txt
- [x] Have a look at boost warning in cmake
- [x] check if hyrise_impl.a as added by #1474 changes the compile time
- [x] Update cmake on CI Mac (https://github.com/Homebrew/homebrew-core/pull/47226)
- [x] increase ccache folder size
- [x] Add documentation for developers (sloppiness, unity)
- [x] self-review
- [x] Feedback to cmake on unity+clang-tidy
- [x] better describe ccache options

* Keeps the build process from waiting on jemalloc
* Uses -O0 for debug linking
* Enables the use of unity builds (requires cmake 3.16)
* Enables precompiled-headers (requires cmake 3.16)
* Enables dynamic linking for Hyrise
* Uses LLVM lld instead of ld under Linux

Additionally
* Fixes some things that will become warnings with clang 9
* Fixes some ODR violations

compiletime

For details see this blog post: https://medium.com/p/8523135aed72

Performance:
```diff
 +----------------+----------------+-------+---------------+-------+------------+---------+
 | Benchmark      | prev. iter/s   | runs  | new iter/s    | runs  | change [%] | p-value |
 +----------------+----------------+-------+---------------+-------+------------+---------+
 | TPC-H 01       | 0.599761068821 | 36    | 0.60238713026 | 37    | +0%        |  0.5546 |
 | TPC-H 02       | 11.8571109772  | 712   | 12.149933815  | 729   | +2%        |  0.0000 |
 | TPC-H 03       | 6.96436452866  | 418   | 6.91861820221 | 416   | -1%        |  0.1704 |
 | TPC-H 04       | 6.85212564468  | 412   | 6.93864488602 | 417   | +1%        |  0.0036 |
[...]

PR #1968, opened 2019-12-20
Improve size estimation of string segments

Bouncner:
This PR does mainly two things:
 - change the size estimation to sample whenever strings are stored in a vector (e.g., value segments and dictionary segments) to better estimate the size (a newly added mode allows to select if a 1% sample should be used or if all strings should be considered (much slow))
 - add a table "meta_segments_accurate" to the meta tables which is rather expensive, but is accurate for string values and also determines actual distinct counts per segment

This PR will certainly increase the runtime of getting a size estimation, but is only string columns and certain encodings. Further, we misestimated some TPC-H columns (mostly anything that ends with `_comment`) by a factor of two and more. As these columns are responsible for a good chunk of memory in TPC-H, sampling seems to be a much better route here (imo).

In the long run, I am happy to discuss ways to store distinct counts and sufficient string information per segment during the initial encoding.


Bouncner:
For upcoming projects (e.g., @krichly's current project and @Bensk1 potential master's thesis topic), I'd like to include the sort order in the chunk table. But -- while not being used -- chunks can be sorted by multiple columns. I could create a `meta_chunk_sort_orders` or simply put the first sort column (will there ever be more?!?) into the chunk meta table.

Any opinions here?

mrks:
We'll have a similar problem with constraints. Either we need a different table or we simply ignore 1NF and concat. Depends mostly on how this is consumed i guess.

Bouncner:
Since it can easily be joined and I doubt joining is a problem for anybody ever using these tables, I'd vote for the "cleaner" additional table.

src/lib/storage/lz4_segment.cpp
```diff
@@ -442,13 +442,17 @@ std::shared_ptr<BaseSegment> LZ4Segment<T>::copy_using_allocator(const Polymorph
 }
 
 template <typename T>
-size_t LZ4Segment<T>::estimate_memory_usage() const {
+size_t LZ4Segment<T>::memory_usage(const MemoryUsageCalculationMode) const {
+  // MemoryUsageCalculationMode can be ignored since all relevant information can be either obtained directly (e.g.,
+  // size of NULL values vector) or the actual size is already stored (e.g., data_size()).
+
   // The null value vector is only stored if there is at least 1 null value in the segment.
   auto bool_size = size_t{0u};
```
Bensk1: Is `bool_size` the best name here?

src/lib/utils/meta_table_manager.cpp
```diff
@@ -2,18 +2,83 @@
...
 #include "storage/base_encoded_segment.hpp"
+#include "storage/segment_iterables/any_segment_iterable.hpp"
 #include "storage/table.hpp"
 #include "storage/table_column_definition.hpp"
 
+namespace {
+
+using namespace opossum;  // NOLINT
+
+int32_t get_distinct_value_count(const std::shared_ptr<BaseSegment> segment) {
```
Bensk1: Why does this return a signed integer, can it ever become positive?

src/lib/utils/meta_table_manager.cpp
```diff
@@ -2,18 +2,84 @@
...
+  auto distinct_value_count = int32_t{0};
+  resolve_data_type(segment->data_type(), [&](auto type) {
```
Bensk1: Should we have a specialized case for dictionary segments?

src/lib/utils/meta_table_manager.cpp
```diff
@@ -2,18 +2,84 @@
...
+    using ColumnDataType = typename decltype(type)::type;
+    std::unordered_set<ColumnDataType> distinct_values;
+
+    auto iterable = create_any_segment_iterable<ColumnDataType>(*segment);
+    iterable.with_iterators([&](auto it, auto end) {
+      for (auto current_position = size_t{0}; it != end; ++it, ++current_position) {
```
Bensk1: Why do we need `current_position`?

src/lib/utils/meta_table_manager.cpp
```diff
@@ -87,37 +153,79 @@ std::shared_ptr<Table> MetaTableManager::generate_chunks_table() {
...
-        const auto data_type = pmr_string{data_type_to_string.left.at(table->column_data_type(column_id))};
-        AllTypeVariant encoding = NULL_VALUE;
-        if (const auto& encoded_segment = std::dynamic_pointer_cast<BaseEncodedSegment>(segment)) {
-          encoding = pmr_string{encoding_type_to_string.left.at(encoded_segment->encoding_type())};
-        }
+  return output_table;
+}
+                                              {"sorted_by_column_name", DataType::String, false},
+                                              {"sort_order_mode", DataType::String, false}};
+  auto output_table = std::make_shared<Table>(columns, TableType::Data, std::nullopt, UseMvcc::Yes);
```
Bensk1: This format seems completely broken.

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,62 @@
+#pragma once
+
+#include <random>
+
+#include "types.hpp"
+
+namespace opossum {
+
+// In case the initially preallocated capacity is too small, the string is allocated on the heap.
```
Bensk1: This comment does not really sound like method's general comment

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,62 @@
...
+template <typename T>
+size_t memory_usage_of_string_objects(const T& string_vector) {
+  // https://stackoverflow.com/questions/2037209/what-is-a-null-terminated-string
+  const auto sso_string_capacity = std::string("").capacity();
+
+  // Add up sizes for vector object and the pre-initialized strings.
+  auto bytes = string_vector.capacity() * sizeof(pmr_string);
+
+  for (const auto& v : string_vector) {
```
Bensk1: Could we find a better name for `v`?

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include <random>
+
+#include "types.hpp"
+
+namespace opossum {
+
+// In case the initially preallocated capacity is too small, the string is allocated on the heap.
+template <typename T>
+size_t memory_usage_of_string_objects(const T& string_vector) {
```
Bensk1: If this always takes a `string_vector`, why is this templated?

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,62 @@
...
+  if (string_vector.empty()) return base_size;
+
+  if (mode == MemoryUsageCalculationMode::Full) {
+    // Simple (but expensive case): the whole vector's string sizes are aggregated.
+    // std::cout << "estimating fully" << std::endl;
+    return base_size + memory_usage_of_string_objects(string_vector);
+  }
+
+  // For the sampled estimation case, we first create a vector of sample string values and estimate that vector fully.
+  constexpr auto sampling_factor = 0.01f;
```
Bensk1: Just out of curiosity, did you experiment with other sampling factors for our benchmarks?

src/lib/storage/lz4_segment.cpp
Bouncner: Changed to `null_value_vector_size`.

src/lib/utils/meta_table_manager.cpp
Bouncner: Nope, but since we only use this method within the meta table manager to fill an integer columns (and we don't have unsigned column types), I don't think the additional cast would benefit anyone.

src/lib/utils/meta_table_manager.cpp
Bouncner: copy/paste fail. Fixed.

src/lib/utils/size_estimation_utils.hpp
Bouncner: Depending on the caller, it might be a `pmr_vector` or `std::vector` with a `pmr_string` or `std::string`.

src/lib/utils/size_estimation_utils.hpp
Bouncner: Would do you mean with benchmarks?

Bensk1: TPC-H, -DS, JOB. I mean: do you know how accurate the sampling is for different sampling factors.

src/lib/storage/lz4_segment.cpp
Bouncner: Also in ValueSegment.

src/lib/utils/meta_table_manager.cpp
Bouncner: Since it comes for free, I guess it makes sense. Will change.

Right now, I assume that a NULL is not a value. Is everybody onboard with this view? Hence, I would return the size of the dictionary.

src/lib/utils/meta_table_manager.cpp
Bouncner: I accidentally committed half of a WIP function.

src/lib/utils/meta_table_manager.cpp
Bensk1: Hm, I don't know. I could easily imagine this method to be used elsewhere in the future and I had the feeling that we are quite strict with our return types so far.

src/lib/utils/size_estimation_utils.hpp
Bouncner: Changed to `single_string`.

src/lib/utils/size_estimation_utils.hpp
Bouncner:
```
/**
  * This function iterates over a given vector of strings and sums up the memory usage of all elements. Due to the
  * small string optimization (SSO) in most current C++ libraries, each string has an initially allocated buffer (e.g.,
  * 15 chars in GCC's libstdc++). If a string is larger, the string is allocated on the heap and the initial string
  * object stores a pointer to the actual string on the heap.
  */
```

Better?

src/lib/utils/meta_table_manager.cpp
Bouncner: The dictionary handling required another cast, so now returning size_t has the overall less casts. Changed. :)

mrks:
@mflueggen - This should be of interest to you

mrks:
(In the sense of "be aware that it changes")

mrks:
@Bouncner still wip?

Bouncner:
Yep, I think we need to discuss index handling first (Monday meeting).

mflueggen:
Would updating the string size on the fly (on every insert/update/delete) be too expensive or even feasible?

src/lib/utils/meta_table_manager.cpp
Bouncner: Should be fixed.

src/lib/utils/size_estimation_utils.hpp
Bouncner: Unfortunately no. I don't have the numbers.

Bouncner:
> Would updating the string size on the fly (on every insert/update/delete) be too expensive or even feasible?

Currently, we do not store such information in the chunks. It would definitely make sense to store them "somewhere" since there are multiple occasions where it would be almost free (e.g., chunk encoding). But right now, we don't have such statistics and hence we cannot update them during insert/updates.
Conerning the performance question: due to insert-only, the costs might be neglectable as we never have to lock.

Bouncner:
No longer WIP. @Bensk1 : you can take another look at it now.

src/test/utils/size_estimation_utils_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+  // only run a sanity check, assuming all values fit within the SSO capacity.
+  const auto expected_size_sample = sizeof(pmr_vector<pmr_string>) + vector_length * sizeof(pmr_string);
+  EXPECT_LE(expected_size_sample,
+            estimate_string_vector_memory_usage(string_vector, MemoryUsageCalculationMode::Sampled));
+
+  // For the full estimation, we should expect an accurate measurement. Four strings should reside in separate heap
+  // allocations (+ 1 for the added \0 termination).
+  const auto expected_size_full = expected_size_sample + 4 * (large_string_length + 1);
+  EXPECT_EQ(expected_size_full, estimate_string_vector_memory_usage(string_vector, MemoryUsageCalculationMode::Full));
+  EXPECT_GT(expected_size_full * 1.05,
```
Bensk1: What's the purpose of that EXPECT?

Bouncner: Removed.

src/test/utils/size_estimation_utils_test.cpp
```diff
@@ -0,0 +1,52 @@
...
+// Check early out of estimation
+TEST(SizeEstimationUtilsTest, EmptyVector) {
+  const auto empty_vector = pmr_vector<pmr_string>{};
+  const auto expected_size = sizeof(pmr_vector<pmr_string>);
+
+  EXPECT_EQ(expected_size, estimate_string_vector_memory_usage(empty_vector, MemoryUsageCalculationMode::Sampled));
+  EXPECT_EQ(expected_size, estimate_string_vector_memory_usage(empty_vector, MemoryUsageCalculationMode::Full));
+}
+
+// Check correct sampling for shorter vectors than samples being taken
```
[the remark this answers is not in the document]
Bouncner: Changed to ``// Check that sampling works as expected when the input vector is shorter than the minimal sample size.``

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,63 @@
...
+  * small string optimization (SSO) in most current C++ libraries, each string has an initially allocated buffer (e.g.,
+  * 15 chars in GCC's libstdc++). If a string is larger, the string is allocated on the heap and the initial string
+  * object stores a pointer to the actual string on the heap.
+  */
+template <typename T>
+size_t memory_usage_of_string_objects(const T& string_vector) {
+  // https://stackoverflow.com/questions/2037209/what-is-a-null-terminated-string
+  const auto sso_string_capacity = std::string("").capacity();
+
+  // Add up sizes for vector object and the pre-initialized strings.
```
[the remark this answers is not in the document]
Bouncner: `// Add size for pre-initialized strings.``

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,63 @@
...
+  auto bytes = string_vector.capacity() * sizeof(pmr_string);
```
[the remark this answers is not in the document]
Bouncner: Tested on two platforms and various compilers on wandbox.org.
cpp.sh states "The system uses GCC 4.9.2, with Boost 1.55 available.". This GCC does not implement SSO.
As far I have read it, all compilers store the initial char buffer within the string object. Thus is included in sizeof(), contrary to the size behind the potential (larger-than-SSO-sized) string on the heap.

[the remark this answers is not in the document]
Bouncner: Will adapt the comment.

src/lib/utils/meta_table_manager.cpp
```diff
@@ -87,37 +165,60 @@ std::shared_ptr<Table> MetaTableManager::generate_chunks_table() {
...
+
+  auto output_table = std::make_shared<Table>(columns, TableType::Data, std::nullopt, UseMvcc::Yes);
+  gather_segment_meta_data(output_table, MemoryUsageCalculationMode::Sampled);
+
+  return output_table;
+}
+
+std::shared_ptr<Table> MetaTableManager::generate_accurate_segments_table() {
+  PerformanceWarning("Accurate segment information are expensive to gather. Use with caution.");
+  const auto columns = TableColumnDefinitions{
```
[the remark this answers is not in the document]
Bouncner: I have no idea. I changed it back to the way other functions do it, but format changed it back. I don't mind this formatting, but if you want I will disable clang-format and change it.

src/lib/utils/meta_table_manager.cpp
```diff
@@ -2,18 +2,96 @@
...
+            std::stringstream ss;
+            ss << *encoded_segment->compressed_vector_type();
+            vector_compression = pmr_string{ss.str()};
+          }
+        }
+
+        if (mode == MemoryUsageCalculationMode::Full) {
+          const auto distinct_value_count = static_cast<int32_t>(get_distinct_value_count(segment));
+          meta_table->append({pmr_string{table_name}, static_cast<int32_t>(chunk_id), static_cast<int32_t>(column_id),
+                              pmr_string{table->column_name(column_id)}, data_type, distinct_value_count, encoding,
```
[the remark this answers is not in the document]
Bouncner: That means NULL values for the distinct value column? Personally, I don't like it. But I don't have a problem changing that.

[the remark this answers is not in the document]
Bensk1: Right, it would have to be NULL values. I am not sure if we want that.

src/lib/utils/meta_table_manager.cpp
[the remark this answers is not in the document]
Bensk1: No, just keep the clang-format output

src/lib/utils/size_estimation_utils.hpp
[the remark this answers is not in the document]
Bensk1: Hmpf, I just saw that they support C++14 and assumed a newer gcc for that. Alright then.

Bouncner:
Not sure what you don't like about the table header. Would it help to put the distinct count at the end? In that way, it might be more obvious that the "accurate" table just appended that field.

Bensk1:
Seems reasonable

src/lib/storage/fixed_string_dictionary_segment.hpp
```diff
@@ -44,7 +44,7 @@ class FixedStringDictionarySegment : public BaseDictionarySegment {
 
   std::shared_ptr<BaseSegment> copy_using_allocator(const PolymorphicAllocator<size_t>& alloc) const final;
 
-  size_t estimate_memory_usage() const final;
+  size_t memory_usage(const MemoryUsageCalculationMode = MemoryUsageCalculationMode::Full) const final;
```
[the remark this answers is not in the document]
Bouncner: What would you recommend?

src/lib/storage/value_segment.cpp
```diff
@@ -158,15 +159,31 @@ std::shared_ptr<BaseSegment> ValueSegment<T>::copy_using_allocator(const Polymor
...
+  auto null_value_vector_size = size_t{0};
+  if (_null_values) {
+    null_value_vector_size = _null_values->size() * sizeof(bool);
+    // Integer ceiling, since sizeof(bool) equals 1, but boolean vectors are optimized.
+    null_value_vector_size =
+        sizeof(_null_values) +
+        (_null_values->size() % CHAR_BIT ? null_value_vector_size / CHAR_BIT + 1 : null_value_vector_size / CHAR_BIT);
+  }
+
+  return sizeof(*this) + estimate_string_vector_memory_usage(_values, mode) + null_value_vector_size;
```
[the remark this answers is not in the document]
Bouncner: Sounds better. Will change.

src/lib/utils/meta_table_manager.cpp
```diff
@@ -2,18 +2,96 @@
...
+#include "storage/fixed_string_dictionary_segment.hpp"
+#include "storage/segment_iterables/any_segment_iterable.hpp"
 #include "storage/table.hpp"
 #include "storage/table_column_definition.hpp"
 
+namespace {
+
+using namespace opossum;  // NOLINT
+
+size_t get_distinct_value_count(const std::shared_ptr<BaseSegment>& segment) {
```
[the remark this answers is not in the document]
Bouncner: Shall I add a todo?

src/lib/utils/meta_table_manager.cpp
```diff
@@ -2,18 +2,96 @@
...
+
+          if (encoded_segment->compressed_vector_type()) {
+            std::stringstream ss;
+            ss << *encoded_segment->compressed_vector_type();
+            vector_compression = pmr_string{ss.str()};
+          }
+        }
+
+        if (mode == MemoryUsageCalculationMode::Full) {
+          const auto distinct_value_count = static_cast<int32_t>(get_distinct_value_count(segment));
```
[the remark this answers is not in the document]
Bouncner: It's the data type of the table column.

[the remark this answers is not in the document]
Bouncner: Ah got your point. Will change column type to long too prevent problems here.

src/lib/storage/fixed_string_dictionary_segment.hpp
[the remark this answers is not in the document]
mrks: Just require the argument to be passed?

src/lib/utils/meta_table_manager.cpp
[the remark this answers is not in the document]
mrks: Sounds good

src/lib/utils/meta_table_manager.cpp
```diff
@@ -87,37 +165,60 @@ std::shared_ptr<Table> MetaTableManager::generate_chunks_table() {
   return output_table;
 }
 
-std::shared_ptr<Table> MetaTableManager::generate_segments_table() {
+std::shared_ptr<Table> MetaTableManager::generate_chunk_sort_orders_table() {
```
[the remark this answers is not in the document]
Bouncner: I guess you mean part of the chunk table. To be honest (as discussed a couple of weeks ago) I though we support multiple orders since the used sort operator supports it either.

However, we are currently implementing/extending sort flags for the seminar and I wanted to make sure that we support multiple orders whenever they are needed. Maybe we should discuss this in person.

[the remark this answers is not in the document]
mrks: If there is actual work being done in that direction, it's fine. For now, I'd add a comment explaining that. Also, once that is done, we will need another column that reports the "order" of the order, i.e., which column is the primary sort column etc.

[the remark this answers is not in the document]
mrks: Also, somewhat unrelated: Should `Chunk::set_ordered_by` require `!is_mutable()`?

resources/test_data/tbl/meta_tables/meta_segments.tbl
```diff
@@ -1,9 +1,9 @@
-table_name|chunk_id|column_id|column_name|column_data_type|encoding_name
-string|int|int|string|string|string_null
-int_int|0|0|a|int|null
-int_int|0|1|b|int|null
-int_int|1|0|a|int|null
-int_int|1|1|b|int|null
-int_int_int_null|0|0|a|int|RunLength
-int_int_int_null|0|1|b|int|Dictionary
-int_int_int_null|0|2|c|int|null
+table_name|chunk_id|column_id|column_name|column_data_type|encoding_type|vector_compression_type|estimated_size_in_bytes
```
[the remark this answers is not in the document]
Bouncner: But keep `column_data_type`? I thought it is more uniform this way and reflects the actual type `EncodingType` in Hyrise.

src/lib/storage/dictionary_segment.cpp
```diff
@@ -62,11 +63,16 @@ std::shared_ptr<BaseSegment> DictionarySegment<T>::copy_using_allocator(
...
 template <typename T>
-size_t DictionarySegment<T>::estimate_memory_usage() const {
+size_t DictionarySegment<T>::memory_usage(const MemoryUsageCalculationMode mode) const {
   return sizeof(*this) + _dictionary->size() * sizeof(typename decltype(_dictionary)::element_type::value_type) +
          _attribute_vector->data_size();
 }
 
+template <>
+size_t DictionarySegment<pmr_string>::memory_usage(const MemoryUsageCalculationMode mode) const {
+  return sizeof(*this) + estimate_string_vector_memory_usage(*_dictionary, mode) + _attribute_vector->data_size();
```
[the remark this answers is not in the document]
Bouncner: Changed to `string_vector_memory_usage`.

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
```diff
@@ -102,6 +102,7 @@ void FixedStringVector::reserve(const size_t n) { _chars.reserve(n * _string_len
 size_t FixedStringVector::data_size() const { return sizeof(*this) + _chars.size(); }
 
 std::shared_ptr<const pmr_vector<pmr_string>> FixedStringVector::dictionary() const {
+  PerformanceWarning("dictionary() copies the segement's whole dictionary.");
```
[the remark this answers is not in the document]
Bouncner: Oh, thought I was in the fixed string segment. Will removed it. I don't see a reason for that.

resources/test_data/tbl/meta_tables/meta_segments.tbl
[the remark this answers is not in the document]
mrks: `¯\_(ツ)_/¯`

Bensk1:
I wanted to have a really quick test on how the sample rate affects the accuracy to find out if we can decrease it further. I used the name table of the IMDB dataset for this test. A really quick test was not possible, the table has only 6.3M entries but sampling takes ages:

```
SELECT * FROM meta_segments WHERE column_name = 'name'

Execution info: [PARSE: 8 µs 161 ns, SQL TRANSLATE: 2 min 30 s, OPTIMIZE: 3 min 19 s, LQP TRANSLATE: 39 s 967 ms, EXECUTE: 13 s 17 ms (wall time) | QUERY PLAN CACHE HITS: 0/1 statement(s)]
```

Apart from the tremendous times, it is weird that both, the SQL translation and the optimization phase, take so much time. Precise size calculations are quite a bit faster:

```
SELECT * FROM meta_segments_accurate WHERE column_name = 'name'
Execution info: [PARSE: 8 µs 242 ns, SQL TRANSLATE: 309 ms 50 µs, OPTIMIZE: 405 ms 882 µs, LQP TRANSLATE: 75 ms 91 µs, EXECUTE: 27 ms 288 µs (wall time) | QUERY PLAN CACHE HITS: 0/1 statement(s)]
```

Apparently, std::sample is super slow. It does have to copy the strings but I still do not understand how it can be **that** slow. Generating random integers and using these for the samples is much faster. I did a quick (and hopefully fair) comparison [here](http://quick-bench.com/UqiEti_XRJJsL0iIhO2_mj4X_so). Did I get anything wrong?

src/lib/utils/meta_table_manager.cpp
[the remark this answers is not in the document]
Bouncner: Does this comment help?
```
/**
 * At the moment, each chunk can be sorted by exactly one column or none. Hence, having a column within the chunk table
 * would be sufficient. However, this will change in the near future (e.g., when a sort-merge join evicts a chunk that
 * is sorted on two columns). To prepare for this change, this additional table stores the sort orders and allows a
 * chunk to have multiple sort orders. Cascading sort orders for chunks are currently not planned.
 */
```

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,63 @@
...
+  * small string optimization (SSO) in most current C++ libraries, each string has an initially allocated buffer (e.g.,
+  * 15 chars in GCC's libstdc++). If a string is larger, the string is allocated on the heap and the initial string
+  * object stores a pointer to the actual string on the heap.
+  */
+template <typename T>
+size_t memory_usage_of_string_objects(const T& string_vector) {
+  // Get the default capacity of SSO strings.
+  const auto sso_string_capacity = std::string("").capacity();
+
+  // Add size for pre-initialized strings. For compilers using SSO, sizeof(pmr_string) also includes the char buffer.
```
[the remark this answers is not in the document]
Bouncner: Changed to "For stdlib's using SSO, sizeof(pmr_string) also includes the char buffer.".

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,63 @@
...
+  auto bytes = string_vector.capacity() * sizeof(pmr_string);
+
+  for (const auto& single_string : string_vector) {
+    if (single_string.length() > sso_string_capacity) {
+      // Length is guaranteed to return the number of bytes, independent from any potential encoding.
+      // For each string, \0 is appended to denote the end of the string.
+      bytes += single_string.length() + 1;
+    }
```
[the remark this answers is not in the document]
Bouncner: No particular reason. Should be the same in this case.

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+  auto samples_to_draw = std::max(min_rows, static_cast<size_t>(std::ceil(sampling_factor * string_vector.size())));
+  samples_to_draw = std::min(samples_to_draw, string_vector.size());
+  const auto actual_sampling_factor = static_cast<float>(samples_to_draw) / string_vector.size();
+
+  pmr_vector<pmr_string> samples;
+  samples.reserve(samples_to_draw);
+
+  std::sample(string_vector.cbegin(), string_vector.cend(), std::back_inserter(samples), samples_to_draw,
+              std::random_device{});
```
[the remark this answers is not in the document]
Bouncner: I don't know what you mean. :(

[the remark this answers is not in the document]
mrks: back_inserter has to check the capacity of the vector to be inserted to. Same as `vector[]=` vs `emplace_back`. Won't make too much of a difference I guess, more of a theoretical question.

src/test/storage/value_segment_test.cpp
```diff
@@ -103,9 +103,11 @@ TEST_F(StorageValueSegmentTest, MemoryUsageEstimation) {
...
   vs_double.append(42.1337);
 
-  EXPECT_EQ(empty_usage_int + sizeof(int) * 2, vs_int.estimate_memory_usage());
-  EXPECT_EQ(empty_usage_double + sizeof(double), vs_double.estimate_memory_usage());
-  EXPECT_GE(vs_str.estimate_memory_usage(), empty_usage_str + 2 * sizeof(pmr_string));
+  EXPECT_EQ(empty_usage_int + sizeof(int) * 2, vs_int.memory_usage());
+  EXPECT_EQ(empty_usage_double + sizeof(double), vs_double.memory_usage());
+
+  EXPECT_GE(vs_str.memory_usage(), empty_usage_str + 2 * sizeof(pmr_string));
+  EXPECT_GE(vs_str.memory_usage(MemoryUsageCalculationMode::Full), empty_usage_str + 2 * sizeof(pmr_string));
```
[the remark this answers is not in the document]
Bouncner: It's meant to be a sanity check. I could change to a range-like test (size*0.5 < estimate > size*1.05).
For this example, both methods will look at all strings and get an accurate estimation. I wanted to avoid reimplementing have the estimation helpers to get a more accurate estimation.

src/lib/storage/fixed_string_dictionary_segment.hpp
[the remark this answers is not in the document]
Bouncner: Ok.

src/lib/utils/meta_table_manager.cpp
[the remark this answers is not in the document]
Bouncner:
```
// TODO(anyone): #1968 introduced this namespace. With the expected growth of the meta table manager of time, there
//               might be a large number of helper function that are only loosely related to the core functionality
//               of the MetaTableManager. If this becomes the case, restructure and move the functions to other files.
```

Ok?

src/lib/utils/meta_table_manager.cpp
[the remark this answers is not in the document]
Bouncner: > Also, somewhat unrelated: Should `Chunk::set_ordered_by` require `!is_mutable()`?

The cases we are looking at right now create new tables and new chunks. I cannot think of a case where the sorting is set afterwards. Is that what you're talking about?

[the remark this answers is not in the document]
mrks: > "sort-merge join evicts a chunk"

evicts?

> The cases we are looking at right now create new tables and new chunks. I cannot think of a case where the sorting is set afterwards. Is that what you're talking about?

As discussed: _is_mutable should be set automatically for referencing chunks, and in Projection/Aggregate/...

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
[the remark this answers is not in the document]
mrks: How was this resovled?

src/lib/utils/meta_table_manager.cpp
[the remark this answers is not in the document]
mrks: Sounds good

src/lib/utils/size_estimation_utils.hpp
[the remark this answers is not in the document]
mrks: `capacity` is closer to the actually used size in memory. Thinking about it, you might even want to consider `malloc_usable_size`, depending on how you plan to use the data:

```
#include <iostream>
#include <cstdlib>
#include <string>
#include <malloc.h>

void malloc(

int main()
{
    auto s = std::string{"hallomartindieserstringistvolllang"};
    std::cout << s << "\n\tlength: " << s.length() << "\n\tcapacity: " << s.capacity() << "\n\tmalloc_usable_size: " << malloc_usable_size((void*)s.c_str()) << std::endl;
    s = "hallomartindieserstringistkuerzer";
    std::cout << s << "\n\tlength: " << s.length() << "\n\tcapacity: " << s.capacity() << "\n\tmalloc_usable_size: " << malloc_usable_size((void*)s.c_str()) << std::endl;
}
```

```
hallomartindieserstringistvolllang
	length: 34
	capacity: 34
	malloc_usable_size: 40
hallomartindieserstringistkuerzer
	length: 33
	capacity: 34
	malloc_usable_size: 40
```

src/test/storage/value_segment_test.cpp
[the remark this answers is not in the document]
mrks: Which helper? Why can't you use EXPECT_EQ here?

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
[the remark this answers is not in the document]
Bouncner: I did not mean to resolve it. There will be another PR.

src/test/storage/value_segment_test.cpp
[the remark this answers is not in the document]
Bouncner: Rewrote a large part. Will rewrite the test as well. Main motivation was that I did not want to copy all the code of the size estimation helpers to calculate the size. And since the number differ for different stdlib's, we cannot check for a fixed value.
If we decide for your `malloc_usable_size`, the test code will become even longer.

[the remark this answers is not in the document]
mrks: > And since the number differ for different stdlib's, we cannot check for a fixed value.

We already have tests that differ between std lib implementations. Any reason not to do it here?

> If we decide for your malloc_usable_size, the test code will become even longer.

I'm not even sure if that is a good thing. Mostly wanted to point out that even capacity might not be 100% accurate.

src/lib/utils/meta_table_manager.cpp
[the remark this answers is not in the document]
Bouncner: I'll added this tasks to the tasks of the "sort exploitation" group.

src/lib/storage/value_segment.cpp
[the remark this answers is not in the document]
Bouncner: Done.

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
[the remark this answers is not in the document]
Bouncner: Closing, another branch is currently being benchmarked. Will open the PR soon.

src/lib/storage/fixed_string_dictionary_segment.hpp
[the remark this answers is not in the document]
Bouncner: Done.

src/lib/utils/size_estimation_utils.hpp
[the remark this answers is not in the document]
Bouncner: I am not sure this might be going "too far". Besides, I am not sure if the actual number if correct because `c_str()` (might?) creates a copy as it needs to add a null terminator for C-like strings to the end of the string. So the actual string object might be differently aligned than that new returned c_str.

src/lib/utils/size_estimation_utils.hpp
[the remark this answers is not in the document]
Bouncner: Thanks to @Bensk1's benchmark, I revised the code. I am benchmarking it right now. Nonetheless, there is still `std::sample` and that does no longer use `back_inserter`.

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,63 @@
...
+  const auto sample_vector_memory_usage = memory_usage_of_string_objects(samples);
+
+  return base_size + static_cast<size_t>(std::ceil(sample_vector_memory_usage / actual_sampling_factor));
```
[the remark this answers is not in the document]
Bouncner: Changed the calculation to cast both numbers to float first. Besides that I think the use of ceil or round or floor is a matter of taste. But I don't like to return floating byte sizes and the rounded up version of that number makes more sense to me than the rounded down version.

src/test/storage/value_segment_test.cpp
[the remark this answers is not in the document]
Bouncner: Code rewritten. Since we cannot really "observe" the actual implementation, I don't think we need to check for the library being used. Even this "accurate" calculation is just a much better estimation, but not necessarily really 100% correct (added a comment for that).

src/test/utils/meta_table_manager_test.cpp
```diff
@@ -49,11 +49,29 @@ TEST_F(MetaTableManagerTest, TableBasedMetaData) {
     }
   }
 
+  // Test meta table for chunk sort orders
```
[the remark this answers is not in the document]
Bouncner: It is.

src/lib/utils/size_estimation_utils.hpp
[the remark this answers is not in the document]
mrks: https://en.cppreference.com/w/cpp/string/basic_string/c_str

>The pointer obtained from c_str() may be invalidated by:
    Passing a non-const reference to the string to any standard library function, or
    Calling non-const member functions on the string, excluding operator[], at(), front(), back(), begin(), rbegin(), end() and rend(). 

Thus, you are working on the actual data.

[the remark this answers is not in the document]
Bouncner: Hmm? I would read it the other way.

Bouncner:
New performance numbers:
```
SELECT * FROM meta_segments WHERE column_name = 'name'

Execution info: [PARSE: 4 µs 630 ns, SQL TRANSLATE: 5 s 186 ms, OPTIMIZE: 7 s 304 ms, LQP TRANSLATE: 1 s 457 ms, EXECUTE: 472 ms 794 µs (wall time) | QUERY PLAN CACHE HITS: 0/1 statement(s)]

SELECT * FROM meta_segments_accurate WHERE column_name = 'name'

Execution info: [PARSE: 4 µs 305 ns, SQL TRANSLATE: 304 ms 66 µs, OPTIMIZE: 369 ms 440 µs, LQP TRANSLATE: 83 ms 182 µs, EXECUTE: 26 ms 345 µs (wall time) | QUERY PLAN CACHE HITS: 0/1 statement(s)]
```

I guess there several issues hidden here. The table generation methods are called 30 times when executing the query the first time. Once the plan exists, it is only called once as expected. So there might be something fishy in the parsing, translation, and optimization.

I don't think this should be solved in this PR, but I'll like to tackle the issue that the accurate table appears to be generated faster than the inaccurate one. The sampling part does not appear to be reason (as far as I can tell).

Bouncner:
There are too many issues with the performance right now. I'll need a bit more time. Will reopen later.

Bensk1:
> New performance numbers:
> 
> ```
> SELECT * FROM meta_segments WHERE column_name = 'name'
> 
> Execution info: [PARSE: 4 µs 630 ns, SQL TRANSLATE: 5 s 186 ms, OPTIMIZE: 7 s 304 ms, LQP TRANSLATE: 1 s 457 ms, EXECUTE: 472 ms 794 µs (wall time) | QUERY PLAN CACHE HITS: 0/1 statement(s)]
> 
> SELECT * FROM meta_segments_accurate WHERE column_name = 'name'
> 
> Execution info: [PARSE: 4 µs 305 ns, SQL TRANSLATE: 304 ms 66 µs, OPTIMIZE: 369 ms 440 µs, LQP TRANSLATE: 83 ms 182 µs, EXECUTE: 26 ms 345 µs (wall time) | QUERY PLAN CACHE HITS: 0/1 statement(s)]
> ```
> 
> I guess there several issues hidden here. The table generation methods are called 30 times when executing the query the first time. Once the plan exists, it is only called once as expected. So there might be something fishy in the parsing, translation, and optimization.
> 
> I don't think this should be solved in this PR, but I'll like to tackle the issue that the accurate table appears to be generated faster than the inaccurate one. I am looking at the sampling part again now.

Are the table generation methods called equally often (more than 30 times) for both tables?

Bouncner:
> > New performance numbers:
> > ```
> > SELECT * FROM meta_segments WHERE column_name = 'name'
> > 
> > Execution info: [PARSE: 4 µs 630 ns, SQL TRANSLATE: 5 s 186 ms, OPTIMIZE: 7 s 304 ms, LQP TRANSLATE: 1 s 457 ms, EXECUTE: 472 ms 794 µs (wall time) | QUERY PLAN CACHE HITS: 0/1 statement(s)]
> > 
> > SELECT * FROM meta_segments_accurate WHERE column_name = 'name'
> > 
> > Execution info: [PARSE: 4 µs 305 ns, SQL TRANSLATE: 304 ms 66 µs, OPTIMIZE: 369 ms 440 µs, LQP TRANSLATE: 83 ms 182 µs, EXECUTE: 26 ms 345 µs (wall time) | QUERY PLAN CACHE HITS: 0/1 statement(s)]
> > ```
> > 
> > 
> > I guess there several issues hidden here. The table generation methods are called 30 times when executing the query the first time. Once the plan exists, it is only called once as expected. So there might be something fishy in the parsing, translation, and optimization.
> > I don't think this should be solved in this PR, but I'll like to tackle the issue that the accurate table appears to be generated faster than the inaccurate one. I am looking at the sampling part again now.
> 
> Are the table generation methods called equally often (more than 30 times) for both tables?

I'll check that later. I think yes, but I am not 100% sure.

Bouncner:
Will probably reopen tomorrow. Just for your interest: http://quick-bench.com/4fVTdnyv4owwkHnQwNY2B-J2wbE

Even when using integers, `std::sample` is very slow. The smaller the sample, the better maps are (with a slight advantage for `std::map` over `std::unordered_map`, which decreases with larger samples).

mrks:
>  Complexity
Linear in std::distance(first,last) 

I can confirm that the difference increases by 10x when increasing SIZE in your example.

Weird. I wonder what reasons they had to choose an implementation that is linear to the size of the sampled vector, not the number of samples.

Bensk1:
The current implementation is stable. Maybe that is the reason?

Bensk1:
> Will probably reopen tomorrow. Just for your interest: http://quick-bench.com/4fVTdnyv4owwkHnQwNY2B-J2wbE
> 
> Even when using integers, `std::sample` is very slow. The smaller the sample, the better maps are (with a slight advantage for `std::map` over `std::unordered_map`, which decreases with larger samples).

I was wondering for what we need the maps. That's for sampling without replacement (zurücklegen), right? I would also be fine with simple sampling, just filling up a vector with ints.

mrks:
> The current implementation is stable. Maybe that is the reason?

Would also be possible with the `std::set` approach, but maybe the libstdc++ didn't consider / care about the `input >>> sample` case

Bensk1:
The `std::set` approach can get really slow if the sample size is close to the input size.

Bouncner:
> I was wondering for what we need the maps. That's for sampling without replacement (zurücklegen), right? I would also be fine with simple sampling, just filling up a vector with ints.

Mainly to avoid duplicates which might yield much smaller sample lists than desired.

> The std::set approach can get really slow if the sample size is close to the input size.

Yes, but the whole idea of sampling is to have a significantly smaller sample here. Otherwise, a non-sampling consecutive loop is by far faster anyways.

Concerning `std::sample`. Thinking about it, it's not trivial. You don't know the values ahead (which we know here and what makes life much simpler for us) and ensuring that you yield the desired number of elements in stable order and with randomized jumps probably isn't that easy.

Bensk1:
> > The std::set approach can get really slow if the sample size is close to the input size.
> 
> Yes, but the whole idea of sampling is to have a significantly smaller sample here. Otherwise, a non-sampling consecutive loop is by far faster anyways.

This was only a thought on why the stdlib developers did not use something like the set approach. In our case it's true that the sample is significantly smaller. In general, this might not be the case and they have to provide a sample function that works in all cases.

Bensk1:
I am fine with your set or map implementation.

Bouncner:
> > Will probably reopen tomorrow. Just for your interest: http://quick-bench.com/4fVTdnyv4owwkHnQwNY2B-J2wbE
> > Even when using integers, `std::sample` is very slow. The smaller the sample, the better maps are (with a slight advantage for `std::map` over `std::unordered_map`, which decreases with larger samples).
> 
> I was wondering for what we need the maps. That's for sampling without replacement (zurücklegen), right? I would also be fine with simple sampling, just filling up a vector with ints.

Sorry, I first wrote `map` while meaning `set` and then also misread your question. :D
Narf ... numbers of coming up in a few minutes.

Bouncner:
@Bensk1, to answer to your question concerning the execution counts, I added a little print to the functions in the `meta_table_manager.cpp`.
For TPC-H (SF 0.01 and 0.1) and IMDB, both tables are created 12 and 12 times. I have seen 30 times before and I don't know what changed. But whatever the number is, we probably need to change that behaviour.
The gists:
TPCH: https://gist.github.com/Bouncner/d4df600591937c88e511b68f3b77e7b3
IMDB: https://gist.github.com/Bouncner/0f84990cc51ed9e95417fc11bd93bb17

For the IMDB `name` table, the accurate information takes 4s (which I would consider ok) for an Unencoded table and 100ms for the inaccurate table. So sampling and not collecting distinct counts has a large impact here. For the dictionary-encoded version, there is almost no difference. The reason is here mainly that the distinct counts take quite long to calculate which comes for free with dictionary segments. The sample is faster, but it's barely noticable for dictionary segments.

If you look at the execution time and the running times of the other query phases, it looks like we are creating the table 10 times in the SQL translation phase and once in the optimization phase.

src/lib/utils/size_estimation_utils.hpp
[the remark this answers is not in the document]
Bouncner: As discussed below, we no longer use `std::sample`.

src/lib/utils/size_estimation_utils.hpp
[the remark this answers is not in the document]
Bouncner: It's not outdated. I dislike Github very much. Did I answer your question?

mrks:
As `SQLTranslator::_translate_stored_table` only needs the `table->column_definitions`, we could expose those in the `MetaTableManager`. The translator could then check for `MetaTableManager::is_meta_table_name` and retrieve the stored definitions instead of requiring the actual tables to be built.

In the optimizer, the two `get_table` calls that are used to retrieve the soft unique constraints should be addressed by @julianmenzler's work, so what remains is the ChunkPruningRule. Skip chunk pruning for meta tables?

src/lib/utils/size_estimation_utils.hpp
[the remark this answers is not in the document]
mrks: I must have misread something.

Bouncner:
> As `SQLTranslator::_translate_stored_table` only needs the `table->column_definitions`, we could expose those in the `MetaTableManager`. The translator could then check for `MetaTableManager::is_meta_table_name` and retrieve the stored definitions instead of requiring the actual tables to be built.

Would you consider this to be part of this PR? I am not quite sure that I understand what needs to be done. Never looked into the SQL translation.

 
> In the optimizer, the two `get_table` calls that are used to retrieve the soft unique constraints should be addressed by @julianmenzler's work, so what remains is the ChunkPruningRule. Skip chunk pruning for meta tables?

Sounds somehow hacky but also ok to me. :)
Part of this PR?

mrks:
Doesn't have to be part of this PR, which does what is advertised and does not break anything. Meta tables being generated more often than necessary is a different thing.

mrks:
Ah, wait a second. I just grepped for `get_table` in the optimizer folder, but the CardinalityEstimator might touch that, too. Forget about the optimizer part. The SQLTranslator part should still make sense though.

Bouncner:
Ready to be dismantled again.

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
mrks: Mmmh, that new keyboard doesn't seem to work well, either ;)

"segement"

src/test/utils/meta_table_manager_test.cpp
mrks: Why isn't this part of TableBasedMetaData?

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,77 @@
...
+/**
+  * This function iterates over the given string vector @param string_vector strings and sums up the memory usage. Due
+  * to the small string optimization (SSO) in most current C++ libraries, each string has an initially allocated buffer
+  * (e.g., 15 chars in GCC's libstdc++). If a string is larger, the string is allocated on the heap and the initial
+  * string object stores a pointer to the actual string on the heap.
+  *
+  * Depending on the @param mode, either all strings all considered or a sample is taken and evaluated.
+  *
+  * Please note, that there are still differences between the stdlib's. Thus, even the full size accumulation is not
+  * guaranteed to be 100% accurate for all libraries.
```
mrks: Not sure how these two things belong together. What has the difference in SSO to do with the accuracy of this?

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,77 @@
...
+size_t string_vector_memory_usage(const V& string_vector, const MemoryUsageCalculationMode mode) {
+  using StringType = typename V::value_type;
+  const auto base_size = sizeof(V);
+
+  // Early out
+  if (string_vector.empty()) return base_size;
+
+  // Get the default pre-allocated capacity of SSO strings.
+  const auto sso_string_capacity = std::string("").capacity();
+  auto sso_exceeding_bytes = [&](const auto& single_string) -> size_t {
```
mrks: string_heap_size?

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,77 @@
...
+    // Run the (expensive) calculation of aggregating the whole vector's string sizes when full estimation is desired
+    // or the given input vector is small.
+    auto elements_size = string_vector.capacity() * sizeof(StringType);
+    for (const auto& single_string : string_vector) {
+      elements_size += sso_exceeding_bytes(single_string);
+    }
+    return base_size + elements_size;
+  }
+
+  std::default_random_engine generator{std::random_device{}()};
```
mrks: Add a comment on why we don't like std::sample?

src/test/storage/value_segment_test.cpp
```diff
@@ -103,9 +103,11 @@ TEST_F(StorageValueSegmentTest, MemoryUsageEstimation) {
...
   vs_double.append(42.1337);
 
-  EXPECT_EQ(empty_usage_int + sizeof(int) * 2, vs_int.estimate_memory_usage());
-  EXPECT_EQ(empty_usage_double + sizeof(double), vs_double.estimate_memory_usage());
-  EXPECT_GE(vs_str.estimate_memory_usage(), empty_usage_str + 2 * sizeof(pmr_string));
+  EXPECT_EQ(empty_usage_int + sizeof(int) * 2, vs_int.memory_usage(MemoryUsageCalculationMode::Sampled));
+  EXPECT_EQ(empty_usage_double + sizeof(double), vs_double.memory_usage(MemoryUsageCalculationMode::Sampled));
+
+  EXPECT_GE(vs_str.memory_usage(MemoryUsageCalculationMode::Sampled), empty_usage_str + 2 * sizeof(pmr_string));
+  EXPECT_GE(vs_str.memory_usage(MemoryUsageCalculationMode::Full), empty_usage_str + 2 * sizeof(pmr_string));
```
mrks: I'd still prefer accurate numbers :(

See `TEST_F(BTreeIndexTest, MemoryConsumptionVeryShortStringNoNulls)` for an example

Bensk1: I totally agree!

mrks: Looks like `BTreeIndexImpl<pmr_string>::_add_to_heap_memory_usage` should be replaced by a call to the new implementation, too.

src/lib/utils/size_estimation_utils.hpp
Bouncner: Clang and GCC (or better: their libs) use different ways to implement SSO. The current calculations should be somewhat accurate but I cannot guarantee it (https://shaharmike.com/cpp/std-string/), especially not for upcoming versions.

src/lib/utils/size_estimation_utils.hpp
Bouncner:
```
// We manually create sample positions as this turned out to be much faster than using std::iota and std::sample.
  // Since we want an ordered position list (this potentially increases the performance when accessing the segment), we
  // can directly use std::set to generate distinct sample positions. std::set is slightly faster than
  // std::unordered_set + sorting for small sample sizes.
```

Ok?

src/test/storage/value_segment_test.cpp
Bouncner: I have overseen this test. GE should no longer be required.

src/lib/utils/size_estimation_utils.hpp
mrks: Perfect

src/lib/utils/size_estimation_utils.hpp
mrks: For me, those are two different things. First, the standard does not prescribe an exact allocation strategy, so our approach may silently become invalid in the future. Second, the different implementations do different things and users should be aware of that.

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,77 @@
...
+template <typename V>
+size_t string_vector_memory_usage(const V& string_vector, const MemoryUsageCalculationMode mode) {
+  using StringType = typename V::value_type;
+  const auto base_size = sizeof(V);
+
+  // Early out
+  if (string_vector.empty()) return base_size;
+
+  // Get the default pre-allocated capacity of SSO strings.
+  const auto sso_string_capacity = std::string("").capacity();
```
mrks: Should probably use StringType

src/test/storage/value_segment_test.cpp
Bouncner: I don't really like tests such as in the BTree because it's looks like it is accurate by even specializing on the library, but it's actually just testing our own interpretation of the libraries (which is *wrong* here for llvm).

Just as I side note. Will adapt it.

src/lib/utils/size_estimation_utils.hpp
Bouncner: So what's your recommendation?

src/test/utils/meta_table_manager_test.cpp
Bouncner: Hmm? It is part of `TableBasedMetaData`. Or do you mean the first loop?

src/lib/utils/size_estimation_utils.hpp
mrks: `s/Thus, even/Also/` would be enough for me. I'm just bugged by the text suggesting a logical connection where I don't see one.

src/test/storage/value_segment_test.cpp
Bouncner: Changed the test of the full estimation to EQ.

src/lib/utils/size_estimation_utils.hpp
Bouncner: Ah ok. Now I get you. Changed.

src/test/utils/meta_table_manager_test.cpp
mrks: Yes, I mean the magic thingy that does all sorts of tests where you added `segments_accurate`.

Bouncner: The executed encoding and inserts of tuples would not have an effect. I could add the table there but the TBL would not change for the updated version.
Your call.

mrks: Ah, I didn't realize that `chunk_sort_orders` only contains rows for chunks that are actually ordered. Never mind.

Bouncner:
@mrks , @Bensk1 : anything more to do here? I don't want to merge (assuming the CI ever catches up) as long as there are request for changes.

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return base_size + elements_size;
+  }
+
+  // We manually create sample positions as this turned out to be much faster than using std::iota and std::sample.
+  // Since we want an ordered position list (this potentially increases the performance when accessing the segment), we
+  // can directly use std::set to generate distinct sample positions. std::set is slightly faster than
+  // std::unordered_set + sorting for small sample sizes.
+  std::default_random_engine generator{std::random_device{}()};
+  std::uniform_int_distribution<int> distribution(0, static_cast<int>(samples_to_draw));
```
Bensk1: Why do we have `int` here?

Bouncner: That was required by the usage of the `int` template parameter. And that did not make sense.

Thanks for pointing out. Changed.

mrks:
Nope, I think my being-a-pain-in-the-ass-meter was well filled with this one...

mrks:
Couple of things that I found:

### --track-origins=yes gives slightly more information:
```
==57214==  Uninitialised value was created by a heap allocation
==57214==    at 0x483A7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==57214==    by 0x189378B: allocate (memory_resource.hpp:44)
==57214==    by 0x189378B: allocate (polymorphic_allocator.hpp:82)
==57214==    by 0x189378B: allocate (alloc_traits.h:306)
==57214==    by 0x189378B: _M_allocate (stl_vector.h:343)
==57214==    by 0x189378B: _M_create_storage (stl_vector.h:358)
==57214==    by 0x189378B: _Vector_base (stl_vector.h:302)
==57214==    by 0x189378B: vector (stl_vector.h:521)
==57214==    by 0x189378B: opossum::SizeEstimationUtilsTest_StringVectorExceedingSSOLengths_Test::TestBody() (src/test/utils/size_estimation_utils_test.cpp:34)
```

Not sure why valgrind thinks that memory is uninitialized.

### Single string test
This also fails, both for std::string and pmr_string. It probably should be added to the tests:

```
TEST(SizeEstimationUtilsTest, SingleString) {
  auto string1 = std::string{"a"};
  auto string2 = pmr_string{"b"};

  EXPECT_EQ(string_heap_size(string1), 0);
  EXPECT_EQ(string_heap_size(string2), 0);
}
```

Thus, this doesn't have anything to do with the vector.

```
==643==  Uninitialised value was created by a stack allocation
==643==    at 0x1893E50: _M_local_data (src/test/utils/size_estimation_utils_test.cpp:8)
==643==    by 0x1893E50: basic_string<std::allocator<char> > (basic_string.h:526)
==643==    by 0x1893E50: opossum::SizeEstimationUtilsTest_SingleString_Test::TestBody() (src/test/utils/size_estimation_utils_test.cpp:10)
```

### Compiler
This only seems to happen with clang.

mrks:
Valgrind is drunk:

```diff
diff --git a/src/lib/utils/size_estimation_utils.hpp b/src/lib/utils/size_estimation_utils.hpp
index 66ae01e0..b9d94bc0 100644
--- a/src/lib/utils/size_estimation_utils.hpp
+++ b/src/lib/utils/size_estimation_utils.hpp
@@ -20,7 +21,7 @@ size_t string_heap_size(const T& string) {
     // (cf. https://shaharmike.com/cpp/std-string/).
     return string.capacity() + 1;
   }
-  return 0;
+  return string.capacity() - sso_string_capacity;
 }

 /**
```

fixes it.

Bouncner:
That's quite unintuitive, isn't it? At least add a comment?

mrks:
Have you looked at the actual commit?

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -12,15 +12,18 @@ namespace opossum {
...
   if (string.capacity() > sso_string_capacity) {
     // For heap-allocated strings, \0 is appended to denote the end of the string. capacity() is used over length()
     // since some libraries (e.g. llvm's libc++) also over-allocate the heap strings
     // (cf. https://shaharmike.com/cpp/std-string/).
     return string.capacity() + 1;
   }
-  return 0;
+
+  // valgrind is drunk. This could be `return 0`, but valgrind doesn't like it.
+  DebugAssert(string.capacity() - sso_string_capacity == 0, "SSO does not meet expectations");
```
Bouncner: `string.capacity() == sso_string_capacity` ?

mrks: I wanted to convey that this is the exact thing as `0`. No hard feelings about it though.

Bouncner:
Just did and deleted my comment. ;)
You're too fast.

Added a short comment. Do you know where the conflict comes from?

mrks:
#1881. I have a local, merged version. If you don't mind, I'll push that once it built successfully.

src/lib/utils/size_estimation_utils.hpp
mrks: In the sense of "I want to return 0, but can't. Here, it is the same thing".

Bouncner:
Sure, go ahead.

src/lib/utils/size_estimation_utils.hpp
Bouncner: Ack.

Bouncner:
@Bensk1 , @mrks : is the knaller zündable?

@mrks : thanks for the valgrind help!

mrks:
From my side yes. @Bensk1's point seem to be addressed as well, so if you want this in for some reason, I'd guess it's ok to discard his review.

Bensk1:
Please zünd the knaller! And start enjoying your weekend finally!

mrks:
Please remember to clean up the commit message before zünding the knaller

Bouncner:
Oh, I never do. :-(

Will do it next time.

PR #1973, opened 2020-01-08
Remove dictionary() from FixedStringDictionarySegment

Bouncner:
In the discussion of #1968 we found a rather strange method named `dictionary()` within the fixed string vector. While this method has a misleading name, it also copied the whole vector into a new vector. This has (probably) been done since the other dictionary segment implementation did the same.

This PR removes this method altogether. The like matcher needed to be adapted to handle both `pmr_string` as well as `std::string_view`.

This PR has been done on sidon with clang as a sanity check. It compares the current master against this PR with an encoding config that encodes all string columns as fixed string dictionaries.
This is a "sanity check" as I did the same measurement with GCC on nemea and Q19 degraded by 6%. I assume this is the current "flakiness" of our benchmarks, but I cannot rule it out. At least for clang, the performance improves.
Q13 does improve >10% in all my measured.

```diff
 +----------------+--------------------+------+---------------------+------+------------+---------+
 | Benchmark      | prev. iter/s       | runs | new iter/s          | runs | change [%] | p-value |
 +----------------+--------------------+------+---------------------+------+------------+---------+
 | TPC-H 01       | 0.37529256939888   | 23   | 0.37538352608680725 | 23   | +0%        |  0.9777 |
 | TPC-H 02       | 41.59766387939453  | 2496 | 42.229896545410156  | 2534 | +2%        |  0.0844 |
 | TPC-H 03       | 4.429035663604736  | 266  | 4.392274856567383   | 264  | -1%        |  0.0342 |
 | TPC-H 04       | 3.7825026512145996 | 227  | 3.753763198852539   | 226  | -1%        |  0.1750 |
 | TPC-H 05       | 2.3961899280548096 | 144  | 2.399163246154785   | 144  | +0%        |  0.9045 |
 | TPC-H 06       | 116.68707275390625 | 7002 | 119.15573120117188  | 7150 | +2%        |  0.0000 |
 | TPC-H 07       | 5.18301248550415   | 311  | 5.272796630859375   | 317  | +2%        |  0.0122 |
 | TPC-H 08       | 3.8212928771972656 | 230  | 3.7169783115386963  | 224  | -3%        |  0.0000 |
 | TPC-H 09       | 1.1397175788879395 | 69   | 1.1574090719223022  | 70   | +2%        |  0.0008 |
 | TPC-H 10       | 1.9381413459777832 | 117  | 1.9279181957244873  | 116  | -1%        |  0.4811 |
 | TPC-H 11       | 19.53865623474121  | 1173 | 19.824811935424805  | 1190 | +1%        |  0.0000 |
 | TPC-H 12       | 9.314201354980469  | 559  | 9.548490524291992   | 573  | +3%        |  0.0000 |
+| TPC-H 13       | 1.649482250213623  | 99   | 1.9548671245574951  | 118  | +19%       |  0.0000 |
 | TPC-H 14       | 32.581275939941406 | 1955 | 32.32889938354492   | 1940 | -1%        |  0.0000 |
 | TPC-H 15       | 47.39055252075195  | 2844 | 47.04608917236328   | 2823 | -1%        |  0.0000 |
 | TPC-H 16       | 5.05262565612793   | 304  | 5.022717475891113   | 302  | -1%        |  0.0001 |
 | TPC-H 17       | 2.949676036834717  | 177  | 2.9436872005462646  | 177  | -0%        |  0.4594 |
 | TPC-H 18       | 0.8713483214378357 | 53   | 0.8831719160079956  | 54   | +1%        |  0.2997 |
 | TPC-H 19       | 5.604300022125244  | 337  | 5.420045375823975   | 326  | -3%        |  0.0000 |
+| TPC-H 20       | 16.05953025817871  | 964  | 16.864906311035156  | 1012 | +5%        |  0.0000 |
 | TPC-H 21       | 0.7333243489265442 | 44   | 0.7469368577003479  | 45   | +2%        |  0.0766 |
 | TPC-H 22       | 9.163599014282227  | 550  | 9.034906387329102   | 543  | -1%        |  0.0000 |
 | geometric mean |                    |      |                     |      | +1%        |         |
 +----------------+--------------------+------+---------------------+------+------------+---------+
```

Furhermore, the projection has been improved by avoiding the compression of attribute vectors when passing on dictionary-encoded segments. The creation of the attribute vector improve by 10x, but all relevant queries are dominated by aggregations. Thus, no TPC-H improvements for the projection.


mrks:
pr name?

Bouncner:
Whoopsie? Didn’t want click on publish. Will change as soon as I am on the PC.

But it’s ready to be reviewed.

src/lib/expression/evaluation/like_matcher.hpp
```diff
@@ -82,19 +82,23 @@ class LikeMatcher {
    *    LikeMatcher{"%hello%"}.resolve(false, [](const auto& matcher) {
    *        std::cout << matcher("He said hello!") << std::endl;
    *    }
+   *
+   * The first template type T defaults to pmr_string, but can also be std::string_view (e.g., for fixed string
+   * dictionary segments). The template type for the functor is the second template argument, since it usually
+   * deduced while T is not decuded automatically.
```
mrks: typo

src/lib/expression/evaluation/like_matcher.hpp
```diff
@@ -82,19 +82,23 @@ class LikeMatcher {
...
    */
-  template <typename Functor>
+  template <typename T = pmr_string, typename Functor>
   void resolve(const bool invert_results, const Functor& functor) const {
     if (std::holds_alternative<StartsWithPattern>(_pattern_variant)) {
       const auto& prefix = std::get<StartsWithPattern>(_pattern_variant).string;
-      functor([&](const pmr_string& string) -> bool {
+      functor([&](const T& string) -> bool {
```
mrks: Can't you accept auto here and get rid of `T`?

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_iterator.hpp
```diff
@@ -61,8 +61,9 @@ class FixedStringIterator : public boost::iterator_facade<FixedStringIterator<On
   }
 
   template <bool OnConstStorageLocal = OnConstStorage>
-  std::enable_if_t<OnConstStorageLocal, const std::string_view> dereference() const {  // NOLINT
-    return std::string_view{&_chars[_pos], strnlen(&_chars[_pos], _string_length)};
+  std::enable_if_t<OnConstStorageLocal, const std::string_view&> dereference() const {  // NOLINT
+    _temporary_return_value = std::string_view{&_chars[_pos], strnlen(&_chars[_pos], _string_length)};
+    return _temporary_return_value;
```
mrks: Why is this necessary?

src/lib/expression/evaluation/like_matcher.hpp
```diff
@@ -82,19 +82,23 @@ class LikeMatcher {
    *    LikeMatcher{"%hello%"}.resolve(false, [](const auto& matcher) {
    *        std::cout << matcher("He said hello!") << std::endl;
    *    }
+   *
```
mrks: Wasn't the main idea to get rid of `FixedStringVector::dictionary`?

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_iterator.hpp
Bouncner: To be able to return a reference.

src/lib/expression/evaluation/like_matcher.hpp
Bouncner: Yes, did I overlook something? The function is gone.

Bouncner: Yes, did I overlook something? The function is gone.

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_iterator.hpp
Bouncner: The previous iterator returned the string_view by value and that disabled range-based loops.

src/lib/expression/evaluation/like_matcher.hpp
Bouncner: Wow. Did not know that's possible. 

Is that guaranteed to work as expected? I wonder if `resolve`'s inner functions are instantiated for several every type now instead of `resolve` being instantiated twice for `pmr_string` and `std::string_view`?

Bouncner:
Will run some measurements to sure nothing broke. After that, I am going to merge the master in hope to get better Q19 results (#1971).

src/lib/expression/evaluation/like_matcher.hpp
Bouncner: Comment might be removed when `auto` works as expected. But do you mean the missing "is"? Cannot find a typo.

Bouncner:
I have no idea what just happened. But the performance when using `auto` increased quite nicely over the previous commit.

```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 0.6648398041725159 | 40    | 0.6692331433296204 | 41    | +1%        |  0.3184 |
 | TPC-H 02       | 68.52555084228516  | 4112  | 69.50872802734375  | 4171  | +1%        |  0.0382 |
 | TPC-H 03       | 7.341727256774902  | 441   | 7.360022068023682  | 442   | +0%        |  0.6959 |
 | TPC-H 04       | 6.742325782775879  | 405   | 6.699501991271973  | 402   | -1%        |  0.1404 |
+| TPC-H 05       | 3.5724611282348633 | 215   | 3.827176570892334  | 230   | +7%        |  0.0000 |
 | TPC-H 06       | 216.87339782714844 | 13013 | 226.55169677734375 | 13594 | +4%        |  0.0000 |
 | TPC-H 07       | 7.470024585723877  | 449   | 7.523620128631592  | 452   | +1%        |  0.2898 |
 | TPC-H 08       | 5.7521071434021    | 346   | 5.731477737426758  | 344   | -0%        |  0.6785 |
 | TPC-H 09       | 1.5733438730239868 | 95    | 1.5810754299163818 | 95    | +0%        |  0.3466 |
 | TPC-H 10       | 3.0102829933166504 | 181   | 3.0070436000823975 | 181   | -0%        |  0.8575 |
 | TPC-H 11       | 29.54155731201172  | 1773  | 29.839937210083008 | 1791  | +1%        |  0.0000 |
 | TPC-H 12       | 12.211451530456543 | 733   | 12.437628746032715 | 747   | +2%        |  0.0000 |
 | TPC-H 13       | 2.633223056793213  | 158   | 2.6608848571777344 | 160   | +1%        |  0.0000 |
 | TPC-H 14       | 45.3634033203125   | 2722  | 45.84188461303711  | 2751  | +1%        |  0.0000 |
 | TPC-H 15       | 94.91619110107422  | 5695  | 96.45069885253906  | 5788  | +2%        |  0.0000 |
 | TPC-H 16       | 7.1049041748046875 | 427   | 7.1522297859191895 | 430   | +1%        |  0.0140 |
 | TPC-H 17       | 4.5060553550720215 | 271   | 4.4468913078308105 | 267   | -1%        |  0.0896 |
 | TPC-H 18       | 1.2890011072158813 | 78    | 1.262071967124939  | 76    | -2%        |  0.1463 |
+| TPC-H 19       | 7.773528575897217  | 467   | 9.15342903137207   | 550   | +18%       |  0.0000 |
 | TPC-H 20       | 27.124622344970703 | 1628  | 27.16774559020996  | 1631  | +0%        |  0.4835 |
+| TPC-H 21       | 1.0977180004119873 | 66    | 1.192076325416565  | 72    | +9%        |  0.0000 |
+| TPC-H 22       | 15.994616508483887 | 960   | 16.847970962524414 | 1011  | +5%        |  0.0000 |
 | geometric mean |                    |       |                    |       | +2%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

I'll merge the master and see what changes. But if that performance holds, FS string segments would improve over the default dictionary segments. Fingers crossed!

src/lib/expression/evaluation/like_matcher.hpp
Bouncner: See below. Auto is now used.

Bouncner:
@mrks: any idea what the current TPCC issue is about? (e9dda56)

src/lib/expression/evaluation/like_matcher.hpp
mrks: decuded

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_iterator.hpp
mrks: Do you have performance numbers for this?

I have tried this approach (storing temporaries and returning references) a couple of times and always ran into problems.

src/lib/expression/evaluation/like_matcher.hpp
mrks: Unless Github is lying to me, fixed_string_vector.[ch]pp are not even touched.

mrks:
> @mrks: any idea what the current TPCC issue is about? (e9dda56)

Interesting - I have been getting that in #1351 as well and was sure that it was related to those changes. Will have a look.

mrks:
Are those benchmarks using the dictionary or the fixed-string dictionary?

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_iterator.hpp
Bouncner: I don't have numbers only on this. There would need to be quite some rewrite I fear to have a good comparison.

mrks: In that case and unless it is necessary for this PR, I'd leave it out. My experience is that holding temporaries in the iterators and returning references to those is a Pandora's box.

Bouncner:
The previous comparison was master vs. this PR and both used FSDict for all string columns.

Here the numbers with current master.

Master vs. this PR with both using Fixed String Dictionaries for string columns:
```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 0.6597551107406616 | 40    | 0.6443614363670349 | 39    | -2%        |  0.0007 |
 | TPC-H 02       | 70.57283020019531  | 4235  | 72.77520751953125  | 4367  | +3%        |  0.0000 |
 | TPC-H 03       | 7.241426944732666  | 435   | 7.245577335357666  | 435   | +0%        |  0.9096 |
 | TPC-H 04       | 6.788454532623291  | 408   | 6.830118656158447  | 410   | +1%        |  0.2238 |
 | TPC-H 05       | 3.7318551540374756 | 224   | 3.8460309505462646 | 231   | +3%        |  0.0049 |
 | TPC-H 06       | 221.99072265625    | 13320 | 226.85472106933594 | 13612 | +2%        |  0.0000 |
 | TPC-H 07       | 7.618153095245361  | 458   | 7.674581527709961  | 461   | +1%        |  0.2371 |
 | TPC-H 08       | 6.435942649841309  | 387   | 6.617753505706787  | 398   | +3%        |  0.0000 |
 | TPC-H 09       | 1.564125418663025  | 94    | 1.5927869081497192 | 96    | +2%        |  0.0003 |
 | TPC-H 10       | 2.9977939128875732 | 180   | 3.0369393825531006 | 183   | +1%        |  0.0404 |
 | TPC-H 11       | 30.618030548095703 | 1838  | 30.272178649902344 | 1817  | -1%        |  0.0000 |
 | TPC-H 12       | 12.301977157592773 | 739   | 12.562972068786621 | 754   | +2%        |  0.0000 |
+| TPC-H 13       | 2.245063066482544  | 135   | 2.7039954662323    | 163   | +20%       |  0.0000 |
 | TPC-H 14       | 45.44719314575195  | 2727  | 46.71722412109375  | 2804  | +3%        |  0.0000 |
 | TPC-H 15       | 95.4798583984375   | 5729  | 97.61774444580078  | 5858  | +2%        |  0.0000 |
 | TPC-H 16       | 7.063474655151367  | 424   | 7.363173007965088  | 442   | +4%        |  0.0000 |
+| TPC-H 17       | 4.605818748474121  | 277   | 4.964298248291016  | 299   | +8%        |  0.0000 |
 | TPC-H 18       | 1.2858281135559082 | 78    | 1.2641243934631348 | 76    | -2%        |  0.1709 |
 | TPC-H 19       | 9.43991756439209   | 567   | 9.576200485229492  | 575   | +1%        |  0.1275 |
 | TPC-H 20       | 27.861751556396484 | 1672  | 28.287826538085938 | 1698  | +2%        |  0.0000 |
 | TPC-H 21       | 1.1929067373275757 | 72    | 1.1576995849609375 | 70    | -3%        |  0.0002 |
 | TPC-H 22       | 16.028148651123047 | 962   | 16.134994506835938 | 969   | +1%        |  0.0557 |
 | geometric mean |                    |       |                    |       | +2%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

Current master with dictionary encoding vs. this PR with Fixed String Dictionaries:
```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
-| TPC-H 01       | 0.6797015070915222 | 41    | 0.6443614363670349 | 39    | -5%        |  0.0000 |
 | TPC-H 02       | 71.73686218261719  | 4305  | 72.77520751953125  | 4367  | +1%        |  0.0273 |
-| TPC-H 03       | 7.669553279876709  | 461   | 7.245577335357666  | 435   | -6%        |  0.0000 |
 | TPC-H 04       | 7.0554070472717285 | 424   | 6.830118656158447  | 410   | -3%        |  0.0000 |
+| TPC-H 05       | 3.533586263656616  | 213   | 3.8460309505462646 | 231   | +9%        |  0.0000 |
 | TPC-H 06       | 226.8496551513672  | 13611 | 226.85472106933594 | 13612 | +0%        |  0.9826 |
 | TPC-H 07       | 7.648836135864258  | 459   | 7.674581527709961  | 461   | +0%        |  0.5930 |
 | TPC-H 08       | 6.486102104187012  | 390   | 6.617753505706787  | 398   | +2%        |  0.0001 |
 | TPC-H 09       | 1.6011005640029907 | 97    | 1.5927869081497192 | 96    | -1%        |  0.3397 |
 | TPC-H 10       | 2.9560153484344482 | 178   | 3.0369393825531006 | 183   | +3%        |  0.0000 |
 | TPC-H 11       | 30.782114028930664 | 1847  | 30.272178649902344 | 1817  | -2%        |  0.0000 |
-| TPC-H 12       | 14.558093070983887 | 874   | 12.562972068786621 | 754   | -14%       |  0.0000 |
+| TPC-H 13       | 2.3789782524108887 | 143   | 2.7039954662323    | 163   | +14%       |  0.0000 |
 | TPC-H 14       | 48.788917541503906 | 2928  | 46.71722412109375  | 2804  | -4%        |  0.0000 |
 | TPC-H 15       | 98.39212799072266  | 5904  | 97.61774444580078  | 5858  | -1%        |  0.0000 |
 | TPC-H 16       | 7.281911849975586  | 437   | 7.363173007965088  | 442   | +1%        |  0.0000 |
+| TPC-H 17       | 4.633598804473877  | 279   | 4.964298248291016  | 299   | +7%        |  0.0000 |
 | TPC-H 18       | 1.2456644773483276 | 75    | 1.2641243934631348 | 76    | +1%        |  0.3110 |
+| TPC-H 19       | 8.860579490661621  | 532   | 9.576200485229492  | 575   | +8%        |  0.0000 |
 | TPC-H 20       | 27.87569808959961  | 1673  | 28.287826538085938 | 1698  | +1%        |  0.0000 |
 | TPC-H 21       | 1.2029165029525757 | 73    | 1.1576995849609375 | 70    | -4%        |  0.0000 |
 | TPC-H 22       | 15.581209182739258 | 935   | 16.134994506835938 | 969   | +4%        |  0.0000 |
 | geometric mean |                    |       |                    |       | +0%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

So there is no (end-to-end) gain over dictionary encoding when taking all TPC-H queries. I will benchmark TPC-DS.

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_iterator.hpp
Bouncner: It is necessary for this PR since operators use range-based for-loops with references.

mrks:
> @mrks: any idea what the current TPCC issue is about? (e9dda56)

Could you check if reverting e93f48c02d3f683a625478eb4d1101b7783968b2 "fixes" it here as well?

src/lib/expression/evaluation/like_matcher.hpp
Bouncner: Whooopsie. Now it's really gone.

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_iterator.hpp
mrks: Mind to give me a line reference?

As far as I understand it, this violates [[iterator.concept.forward]](https://eel.is/c++draft/iterator.concept.forward):
> Pointers and references obtained from a forward iterator into a range [i, s) shall remain valid while [i, s) continues to denote a range.

Bouncner: It's here:
```
../src/lib/operators/table_scan/column_like_table_scan_impl.cpp:124:10: note: use non-reference type 'std::__1::basic_string_view<char, std::__1::char_traits<char> >'
    for (const auto& value : dictionary) {
```

We could iteratively build up a vector of `std::string_view` within the iterator but I have no idea whether this would help here.

I tried to read something about such proxy iterators but could not really find a proper source. I found a paper that discusses such proxy objects (https://sean-parent.stlab.cc/papers/2008-03-sac/p171-pirkelbauer.pdf), but I don't understand a word. :D

Bouncner: Another link that might help here: https://stackoverflow.com/questions/55304218/why-does-the-parallel-for-each-require-forward-iterators.

mrks: Isn't the error just a clang warning that we could suppress locally?

Bouncner: Trying it out. Did not expect that it's ok to "overrule" the request for a reference.

mrks: `Wrange-loop-analysis` is just a performance warning, which does not seem to understand what we are doing here.

Bouncner: But doesn't the warning `loop variable A is always a copy because the range of type B does not return a reference` indicate that we are now creating copies? I don't understand why there is no copy elision.

mrks: Could you give me the entire error message please?

mrks: Wait a sec

mrks: Clang sees that you are using `const auto&` for the loop variable (which is correct for regular dictionaries):

```
// Warn when the loop variable is a const reference that creates a copy.
// Suggest using the non-reference type for copies.  If a copy can be prevented
// suggest the const reference type that would do so.
```

As `operator*` does not return a reference (SemaStmt.cpp:2417), clang issues a warning:

```
    // The range always returns a copy, so a temporary is always created.
    // Suggest removing the reference from the loop variable.
    SemaRef.Diag(VD->getLocation(), diag::warn_for_range_variable_always_copy)
        << VD << RangeInitType;
```

Normally, when the iterator returns a proxy object, `for (auto x : y)` would be correct and should not raise a warning for clang. In this case, we want `const auto&` for DictionarySegment and `auto` for FixedStringDictionarySegment. `const auto&` avoids copies for DictionarySegments and works for `FixedStringDictionarySegment` because of reference lifetime extension.

I believe that clang's error message is a bit misleading. This is not really a copy being made and lifetime-extended, but a proxy.

src/lib/operators/table_scan/column_like_table_scan_impl.cpp
```diff
@@ -121,11 +121,18 @@ std::pair<size_t, std::vector<bool>> ColumnLikeTableScanImpl::_find_matches_in_d
   dictionary_matches.reserve(dictionary.size());
 
   _matcher.resolve(_invert_results, [&](const auto& matcher) {
+    // As Fixed StringVector iterators return an std::string_view value, we disable clang's -Wrange-loop-analysis error
+    // about a potential copy for the loop value.
```
mrks: I'd add a comment saying that we do want `const auto&` for regular dictionaries (see my last comment)

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_iterator.hpp
Bouncner: This is ref vs no-ref in the iterator:

```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 0.6443614363670349 | 39    | 0.6555484533309937 | 40    | +2%        |  0.0119 |
 | TPC-H 02       | 72.77520751953125  | 4367  | 69.72757720947266  | 4184  | -4%        |  0.0000 |
 | TPC-H 03       | 7.245577335357666  | 435   | 7.494514465332031  | 450   | +3%        |  0.0000 |
 | TPC-H 04       | 6.830118656158447  | 410   | 6.830428600311279  | 410   | +0%        |  0.9915 |
-| TPC-H 05       | 3.8460309505462646 | 231   | 3.63138484954834   | 218   | -6%        |  0.0000 |
-| TPC-H 06       | 226.85472106933594 | 13612 | 206.52418518066406 | 12392 | -9%        |  0.0000 |
 | TPC-H 07       | 7.674581527709961  | 461   | 7.545098304748535  | 453   | -2%        |  0.0097 |
 | TPC-H 08       | 6.617753505706787  | 398   | 6.371017932891846  | 383   | -4%        |  0.0000 |
 | TPC-H 09       | 1.5927869081497192 | 96    | 1.577647089958191  | 95    | -1%        |  0.0909 |
 | TPC-H 10       | 3.0369393825531006 | 183   | 2.998673677444458  | 180   | -1%        |  0.0466 |
 | TPC-H 11       | 30.272178649902344 | 1817  | 30.27552604675293  | 1817  | +0%        |  0.9270 |
 | TPC-H 12       | 12.562972068786621 | 754   | 12.499527931213379 | 750   | -1%        |  0.0000 |
 | TPC-H 13       | 2.7039954662323    | 163   | 2.703594207763672  | 163   | -0%        |  0.9584 |
 | TPC-H 14       | 46.71722412109375  | 2804  | 45.88617706298828  | 2754  | -2%        |  0.0000 |
-| TPC-H 15       | 97.61774444580078  | 5858  | 92.44818115234375  | 5547  | -5%        |  0.0000 |
 | TPC-H 16       | 7.363173007965088  | 442   | 7.072004795074463  | 425   | -4%        |  0.0000 |
-| TPC-H 17       | 4.964298248291016  | 299   | 4.561469554901123  | 274   | -8%        |  0.0000 |
 | TPC-H 18       | 1.2641243934631348 | 76    | 1.2041053771972656 | 73    | -5%        |  0.0010 |
-| TPC-H 19       | 9.576200485229492  | 575   | 8.67216968536377   | 521   | -9%        |  0.0000 |
 | TPC-H 20       | 28.287826538085938 | 1698  | 27.54608917236328  | 1653  | -3%        |  0.0000 |
+| TPC-H 21       | 1.1576995849609375 | 70    | 1.2195662260055542 | 74    | +5%        |  0.0000 |
-| TPC-H 22       | 16.134994506835938 | 969   | 14.36899471282959  | 863   | -11%       |  0.0000 |
 | geometric mean |                    |       |                    |       | -3%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

mrks: And that's for FixedStringDictionaries, right?

Bouncner: Yes, all strings are fixed string dictionary segments. "prev" is the version with references, "new" is the version without.

Any ideas what is happening here? Could it be that there are actually copies being created?

mrks: Mmmh, I don't think that it is in the like_matcher, as that is only used for a couple of queries. My guess is on the scan of dictionary-compressed segments, and more specifically somewhere around `ColumnVsValueTableScanImpl::_get_search_value_id`.

Just to show that the references are broken:

```
TEST_F(FixedStringVectorTest, Foo) {
  std::vector<pmr_string> v = {"abc", "x", "ghi"};
  auto fs_vector = FixedStringVector{v.cbegin(), v.cend(), 3};


  const auto& fs_lower_bound = *std::lower_bound(fs_vector.cbegin(), fs_vector.cend(), "def");
  const auto& v_lower_bound = *std::lower_bound(v.cbegin(), v.cend(), "def");
  EXPECT_EQ(fs_lower_bound, v_lower_bound);
}
```

```
[ RUN      ] FixedStringVectorTest.Foo
../src/test/storage/fixed_string_vector_test.cpp:76: Failure
Expected equality of these values:
  fs_lower_bound
    Which is: { ' ' (32, 0x20), '\xE9' (233), '\xE3' (227), '\xEF' (239), '\xFC' (252), '\x7F' (127), '\0', '\0', '\xF1' (241), '\x82' (130), 'N' (78, 0x4E), 'p' (112, 0x70), '\x13' (19), 'V' (86, 0x56), '\0', '\0', '\0', 'B' (66, 0x42) }
  v_lower_bound
    Which is: { 'x' (120, 0x78) }
[  FAILED  ] FixedStringVectorTest.Foo (0 ms)
```

mrks:
You can revert 9e69b30. That was not the issue. :/

src/lib/operators/table_scan/column_like_table_scan_impl.cpp
Bouncner: Changed to
```
// For the loop through the dictionary, we want to use const auto& for DictionaySegments. However,
    // FixedStringVector iterators return an std::string_view value. Thus, we disable clang's -Wrange-loop-analysis
    // error about a potential copy for the loop value.
```

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_iterator.hpp
```diff
@@ -73,6 +73,7 @@ class FixedStringIterator : public boost::iterator_facade<FixedStringIterator<On
   const size_t _string_length;
   Storage& _chars;
   size_t _pos;
+  mutable std::string_view _temporary_return_value;
```
mrks: Now unused, correct?

Bouncner:
Looked at TPC-H again. Apart from the LikeMatcher, there shouldn't be any changes. I am surprised that so many queries degraded.

Any ideas what would be a proper benchmark? Shall I run the DS benchmark?

mrks:
Maybe have a look at a simple query and try to reduce it to a single operator?

Bouncner:
> Maybe have a look at a simple query and try to reduce it to a single operator?

But what would I test? For all the queries that use the LikeMatcher (Q2, Q9, twice Q13, and Q16) performance does not change or improves. For the other queries I'd like to rule out that this PR's changes affect them, but I am not certain how to do that.

Will run DS and TPC-H now.

mrks:
First, you could simply compare the query plans and see if the performance reduction is evenly distributed across all operators or if some stick out. If the performance decreases for non-LIKE queries, you could start by reducing those queries and re-running the comparison (e.g., remove a GROUP BY column, remove an aggregate altogether, reduce the number of joined tables...).

Bouncner:
The issue is probably related to the general flaky performance.

Here a comparison for TPC-H and TPC-DS (-t 180, SF 1, ST, sidon) with default configuration (i.e., everything is stored in DictionarySegments).

TPC-H:
```diff
 +----------------+--------------------+-------+---------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s          | runs  | change [%] | p-value |
 +----------------+--------------------+-------+---------------------+-------+------------+---------+
 | TPC-H 01       | 0.4693295359611511 | 85    | 0.46534356474876404 | 84    | -1%        |  0.0010 |
 | TPC-H 02       | 59.037994384765625 | 10627 | 57.791236877441406  | 10403 | -2%        |  0.0000 |
-| TPC-H 03       | 5.3013691902160645 | 955   | 4.994623184204102   | 900   | -6%        |  0.0000 |
 | TPC-H 04       | 5.033794403076172  | 907   | 4.912196159362793   | 885   | -2%        |  0.0000 |
-| TPC-H 05       | 2.8756191730499268 | 518   | 2.690255880355835   | 485   | -6%        |  0.0000 |
 | TPC-H 06       | 125.64569854736328 | 22617 | 123.18753051757812  | 22174 | -2%        |  0.0000 |
 | TPC-H 07       | 6.045814037322998  | 1089  | 5.980471611022949   | 1077  | -1%        |  0.0215 |
 | TPC-H 08       | 4.172440528869629  | 752   | 4.177720069885254   | 753   | +0%        |  0.6061 |
 | TPC-H 09       | 1.1797435283660889 | 213   | 1.1617437601089478  | 210   | -2%        |  0.0000 |
 | TPC-H 10       | 2.158273696899414  | 389   | 2.1321537494659424  | 384   | -1%        |  0.0019 |
 | TPC-H 11       | 23.872095108032227 | 4297  | 23.662601470947266  | 4260  | -1%        |  0.0000 |
 | TPC-H 12       | 10.873621940612793 | 1958  | 10.744891166687012  | 1935  | -1%        |  0.0000 |
 | TPC-H 13       | 1.7562596797943115 | 317   | 1.6778802871704102  | 303   | -4%        |  0.0000 |
 | TPC-H 14       | 36.467926025390625 | 6565  | 35.616844177246094  | 6412  | -2%        |  0.0000 |
 | TPC-H 15       | 67.65521240234375  | 12178 | 65.89962768554688   | 11862 | -3%        |  0.0000 |
 | TPC-H 16       | 5.36857795715332   | 967   | 5.270361423492432   | 949   | -2%        |  0.0000 |
 | TPC-H 17       | 3.155538320541382  | 568   | 3.110901117324829   | 560   | -1%        |  0.0000 |
 | TPC-H 18       | 0.9457144141197205 | 171   | 0.9190170764923096  | 166   | -3%        |  0.0021 |
 | TPC-H 19       | 5.863828659057617  | 1056  | 5.839353084564209   | 1052  | -0%        |  0.1970 |
 | TPC-H 20       | 20.17174530029297  | 3631  | 19.86154556274414   | 3576  | -2%        |  0.0000 |
-| TPC-H 21       | 0.9353960156440735 | 169   | 0.8823478817939758  | 159   | -6%        |  0.0000 |
 | TPC-H 22       | 9.570918083190918  | 1723  | 9.558195114135742   | 1721  | -0%        |  0.4160 |
 | geometric mean |                    |       |                     |       | -2%        |         |
 +----------------+--------------------+-------+---------------------+-------+------------+---------+
```

TPC-DS:
```diff
 +----------------+--------------------+------+--------------------+------+------------+---------+
 | Benchmark      | prev. iter/s       | runs | new iter/s         | runs | change [%] | p-value |
 +----------------+--------------------+------+--------------------+------+------------+---------+
 | 01             | 14.843240737915039 | 2672 | 14.885199546813965 | 2680 | +0%        |  0.0000 |
+| 03             | 16.70875358581543  | 3008 | 19.038572311401367 | 3427 | +14%       |  0.0000 |
-| 06             | 18.952564239501953 | 3412 | 16.738052368164062 | 3013 | -12%       |  0.0000 |
 | 07             | 6.250863552093506  | 1126 | 6.106761932373047  | 1100 | -2%        |  0.0000 |
 | 09             | 4.156042575836182  | 749  | 4.153772830963135  | 748  | -0%        |  0.0162 |
 | 10             | 10.524148941040039 | 1895 | 10.370977401733398 | 1867 | -1%        |  0.0000 |
 | 13             | 5.778879642486572  | 1041 | 5.627763271331787  | 1013 | -3%        |  0.0000 |
-| 15             | 33.0264778137207   | 5945 | 30.22570037841797  | 5441 | -8%        |  0.0000 |
+| 17             | 10.503890037536621 | 1891 | 11.30838394165039  | 2036 | +8%        |  0.0000 |
 | 19             | 25.819658279418945 | 4648 | 25.47317123413086  | 4586 | -1%        |  0.0000 |
 | 25             | 15.89291000366211  | 2861 | 15.909163475036621 | 2864 | +0%        |  0.7637 |
 | 26             | 12.601251602172852 | 2269 | 11.982741355895996 | 2157 | -5%        |  0.0000 |
 | 28             | 8.721163749694824  | 1570 | 8.574016571044922  | 1544 | -2%        |  0.0000 |
-| 29             | 10.093608856201172 | 1817 | 9.144375801086426  | 1647 | -9%        |  0.0000 |
 | 31             | 3.1150012016296387 | 561  | 3.0509660243988037 | 550  | -2%        |  0.0000 |
-| 34             | 17.147981643676758 | 3087 | 15.816271781921387 | 2848 | -8%        |  0.0000 |
 | 35             | 3.3920466899871826 | 611  | 3.3604063987731934 | 605  | -1%        |  0.0000 |
 | 39a            | 1.603397250175476  | 289  | 1.590185523033142  | 287  | -1%        |  0.1689 |
-| 39b            | 1.9373661279678345 | 349  | 1.556889533996582  | 281  | -20%       |  0.0000 |
 | 41             | 23.383499145507812 | 4210 | 23.417627334594727 | 4216 | +0%        |  0.0000 |
 | 42             | 27.815750122070312 | 5007 | 26.84593391418457  | 4833 | -3%        |  0.0000 |
 | 43             | 2.046342611312866  | 369  | 2.044631242752075  | 369  | -0%        |  0.6669 |
 | 45             | 38.4044303894043   | 6913 | 37.90494155883789  | 6823 | -1%        |  0.0000 |
 | 48             | 4.613901615142822  | 831  | 4.454866886138916  | 802  | -3%        |  0.0000 |
 | 50             | 19.538026809692383 | 3517 | 19.38546371459961  | 3490 | -1%        |  0.0001 |
 | 52             | 27.725143432617188 | 4991 | 28.10042381286621  | 5059 | +1%        |  0.0000 |
 | 55             | 28.51858901977539  | 5134 | 27.875102996826172 | 5018 | -2%        |  0.0000 |
 | 62             | 6.580082893371582  | 1185 | 6.569107532501221  | 1183 | -0%        |  0.0002 |
-| 65             | 5.255367279052734  | 946  | 4.991279602050781  | 899  | -5%        |  0.0000 |
+| 69             | 7.335076808929443  | 1321 | 8.507254600524902  | 1532 | +16%       |  0.0000 |
 | 73             | 17.45408058166504  | 3142 | 18.06098747253418  | 3251 | +3%        |  0.0000 |
 | 79             | 7.821459770202637  | 1408 | 7.655663013458252  | 1379 | -2%        |  0.0000 |
 | 81             | 21.593002319335938 | 3887 | 21.352954864501953 | 3844 | -1%        |  0.0000 |
 | 83             | 40.1932487487793   | 7235 | 39.920692443847656 | 7186 | -1%        |  0.0000 |
 | 85             | 7.403775215148926  | 1333 | 7.070439338684082  | 1273 | -5%        |  0.0000 |
 | 88             | 3.8664844036102295 | 696  | 3.6788251399993896 | 663  | -5%        |  0.0000 |
 | 91             | 28.522581100463867 | 5135 | 27.73870086669922  | 4993 | -3%        |  0.0000 |
 | 93             | 2.4488167762756348 | 441  | 2.3940093517303467 | 431  | -2%        |  0.0000 |
 | 96             | 35.928062438964844 | 6468 | 35.80656433105469  | 6446 | -0%        |  0.0000 |
 | 97             | 1.3721529245376587 | 247  | 1.3660931587219238 | 246  | -0%        |  0.2769 |
 | 99             | 3.272538661956787  | 590  | 3.2642812728881836 | 588  | -0%        |  0.0015 |
 | geometric mean |                    |      |                    |      | -2%        |         |
 +----------------+--------------------+------+--------------------+------+------------+---------+
```

Bouncner:
Here for a configuration that uses FixedStringDictionary for all string columns, comparing master against this PR.

TPC-H:
```diff
 +----------------+---------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s        | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+---------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 0.44814351201057434 | 81    | 0.446403831243515  | 81    | -0%        |  0.1333 |
 | TPC-H 02       | 58.76017379760742   | 10577 | 57.977176666259766 | 10436 | -1%        |  0.0003 |
 | TPC-H 03       | 5.2487030029296875  | 945   | 5.220543384552002  | 940   | -1%        |  0.0374 |
 | TPC-H 04       | 4.922954559326172   | 887   | 4.875730037689209  | 878   | -1%        |  0.0000 |
 | TPC-H 05       | 2.643258810043335   | 476   | 2.649756669998169  | 477   | +0%        |  0.6133 |
 | TPC-H 06       | 125.32432556152344  | 22559 | 125.76953887939453 | 22639 | +0%        |  0.0126 |
 | TPC-H 07       | 6.0103020668029785  | 1082  | 6.004411220550537  | 1081  | -0%        |  0.8412 |
 | TPC-H 08       | 4.2215895652771     | 760   | 4.204634666442871  | 757   | -0%        |  0.1309 |
 | TPC-H 09       | 1.1572321653366089  | 209   | 1.1752519607543945 | 212   | +2%        |  0.0000 |
 | TPC-H 10       | 2.188042640686035   | 394   | 2.1766536235809326 | 392   | -1%        |  0.1340 |
 | TPC-H 11       | 23.90055274963379   | 4303  | 23.646697998046875 | 4257  | -1%        |  0.0000 |
 | TPC-H 12       | 10.4364652633667    | 1879  | 10.358449935913086 | 1865  | -1%        |  0.0000 |
+| TPC-H 13       | 1.5701308250427246  | 283   | 1.9659138917922974 | 354   | +25%       |  0.0000 |
 | TPC-H 14       | 35.2235107421875    | 6341  | 35.1142692565918   | 6321  | -0%        |  0.0000 |
 | TPC-H 15       | 67.83992767333984   | 12212 | 67.70269775390625  | 12187 | -0%        |  0.0000 |
 | TPC-H 16       | 5.223020553588867   | 941   | 5.216329574584961  | 939   | -0%        |  0.4172 |
 | TPC-H 17       | 3.1712310314178467  | 571   | 3.1417458057403564 | 566   | -1%        |  0.0000 |
 | TPC-H 18       | 0.925918459892273   | 167   | 0.9464847445487976 | 171   | +2%        |  0.0165 |
 | TPC-H 19       | 5.864799976348877   | 1056  | 5.8232550621032715 | 1049  | -1%        |  0.0280 |
 | TPC-H 20       | 20.296483993530273  | 3654  | 20.27878189086914  | 3651  | -0%        |  0.6755 |
 | TPC-H 21       | 0.8944036364555359  | 161   | 0.9182919859886169 | 166   | +3%        |  0.0000 |
 | TPC-H 22       | 9.863717079162598   | 1776  | 10.100380897521973 | 1819  | +2%        |  0.0000 |
 | geometric mean |                     |       |                    |       | +1%        |         |
 +----------------+---------------------+-------+--------------------+-------+------------+---------+
```

TPC-DS (Queries 39 with the rather large bump are not using LIKEs):
```diff
 +----------------+--------------------+------+--------------------+------+------------+---------+
 | Benchmark      | prev. iter/s       | runs | new iter/s         | runs | change [%] | p-value |
 +----------------+--------------------+------+--------------------+------+------------+---------+
 | 01             | 15.085593223571777 | 2716 | 14.974433898925781 | 2696 | -1%        |  0.0000 |
+| 03             | 16.636707305908203 | 2995 | 19.180208206176758 | 3453 | +15%       |  0.0000 |
+| 06             | 16.11261558532715  | 2901 | 17.442350387573242 | 3140 | +8%        |  0.0000 |
 | 07             | 6.281346321105957  | 1131 | 6.086889266967773  | 1096 | -3%        |  0.0000 |
 | 09             | 4.181926250457764  | 753  | 4.135385513305664  | 745  | -1%        |  0.0000 |
+| 10             | 9.128987312316895  | 1644 | 10.497650146484375 | 1890 | +15%       |  0.0000 |
 | 13             | 5.795299530029297  | 1044 | 5.769854545593262  | 1039 | -0%        |  0.2484 |
+| 15             | 30.398069381713867 | 5472 | 32.4474983215332   | 5841 | +7%        |  0.0000 |
+| 17             | 10.014714241027832 | 1803 | 11.389898300170898 | 2051 | +14%       |  0.0000 |
 | 19             | 23.308712005615234 | 4196 | 24.377225875854492 | 4388 | +5%        |  0.0000 |
 | 25             | 15.978787422180176 | 2877 | 15.4947509765625   | 2790 | -3%        |  0.0000 |
-| 26             | 12.904751777648926 | 2323 | 12.228285789489746 | 2202 | -5%        |  0.0000 |
 | 28             | 8.737271308898926  | 1573 | 8.690947532653809  | 1565 | -1%        |  0.0000 |
-| 29             | 9.955366134643555  | 1792 | 9.29418659210205   | 1674 | -7%        |  0.0000 |
 | 31             | 2.980335235595703  | 537  | 2.984470844268799  | 538  | +0%        |  0.5179 |
-| 34             | 17.283695220947266 | 3112 | 16.409910202026367 | 2954 | -5%        |  0.0000 |
 | 35             | 3.3987207412719727 | 612  | 3.3517837524414062 | 604  | -1%        |  0.0000 |
-| 39a            | 2.077641725540161  | 374  | 1.7380636930465698 | 313  | -16%       |  0.0000 |
-| 39b            | 2.105422019958496  | 379  | 1.552520513534546  | 280  | -26%       |  0.0000 |
 | 41             | 23.4580020904541   | 4223 | 23.471529006958008 | 4225 | +0%        |  0.0007 |
+| 42             | 27.636655807495117 | 4975 | 30.400493621826172 | 5473 | +10%       |  0.0000 |
 | 43             | 2.0144779682159424 | 363  | 1.9931029081344604 | 359  | -1%        |  0.0000 |
 | 45             | 38.481571197509766 | 6927 | 37.89495849609375  | 6822 | -2%        |  0.0000 |
 | 48             | 4.461430072784424  | 804  | 4.5855631828308105 | 826  | +3%        |  0.0000 |
 | 50             | 19.570606231689453 | 3523 | 19.721097946166992 | 3550 | +1%        |  0.0000 |
 | 52             | 29.581783294677734 | 5325 | 28.640363693237305 | 5156 | -3%        |  0.0000 |
 | 55             | 30.020998001098633 | 5404 | 29.43992805480957  | 5300 | -2%        |  0.0000 |
 | 62             | 6.471374988555908  | 1165 | 6.407708168029785  | 1154 | -1%        |  0.0000 |
 | 65             | 4.978475570678711  | 897  | 4.961326599121094  | 894  | -0%        |  0.2213 |
 | 69             | 8.1957368850708    | 1476 | 8.178154945373535  | 1473 | -0%        |  0.5248 |
-| 73             | 18.29758644104004  | 3294 | 17.237646102905273 | 3103 | -6%        |  0.0000 |
 | 79             | 7.703390598297119  | 1387 | 7.515801429748535  | 1353 | -2%        |  0.0000 |
 | 81             | 21.172481536865234 | 3812 | 20.930471420288086 | 3768 | -1%        |  0.0000 |
 | 83             | 39.73696517944336  | 7153 | 39.39996337890625  | 7092 | -1%        |  0.0000 |
 | 85             | 7.196558475494385  | 1296 | 7.327095985412598  | 1319 | +2%        |  0.0404 |
 | 88             | 3.60707426071167   | 650  | 3.6809773445129395 | 663  | +2%        |  0.0000 |
 | 91             | 28.64310646057129  | 5156 | 28.346866607666016 | 5103 | -1%        |  0.0000 |
 | 93             | 2.4465627670288086 | 441  | 2.4388344287872314 | 439  | -0%        |  0.0251 |
 | 96             | 35.730464935302734 | 6432 | 35.90809631347656  | 6464 | +0%        |  0.0000 |
 | 97             | 1.379141926765442  | 249  | 1.3372018337249756 | 241  | -3%        |  0.0000 |
 | 99             | 3.2076704502105713 | 578  | 3.1841673851013184 | 574  | -1%        |  0.0000 |
 | geometric mean |                    |      |                    |      | -1%        |         |
 +----------------+--------------------+------+--------------------+------+------------+---------+
```

src/lib/operators/projection.cpp
```diff
@@ -120,27 +120,39 @@ std::shared_ptr<const Table> Projection::_on_execute() {
...
                         referenced_dictionary_segment->null_value_id());
+                  } else if constexpr (std::is_same_v<DictionarySegmentType,  // NOLINT
+                                                      FixedStringDictionarySegment<ColumnDataType>>) {
+                    const auto compressed_attribute_vector =
+                        create_filtered_and_compressed_pos_list(typed_segment, pos_list);
+                    const auto& dictionary = typed_segment.fixed_string_dictionary();
+
+                    output_segments[column_id] = std::make_shared<FixedStringDictionarySegment<ColumnDataType>>(
+                        dictionary, std::move(compressed_attribute_vector),
+                        referenced_dictionary_segment->null_value_id());
```
mrks: Hacked together a quick test for this (sans FixedString). Could you add it?

```
TEST_F(OperatorsProjectionTest, ReusesDictionaryWhenForwarding) {
  // Checks that instead of materializing all values in an imperfectly forwarded DictionarySegment, the dictionary
  // is re-used and only the attribute vector is materialized.

  auto dict_table = load_table("resources/test_data/tbl/int_float4.tbl", 3);
  ChunkEncoder::encode_all_chunks(dict_table, EncodingType::Dictionary);
  auto table_wrapper = std::make_shared<TableWrapper>(dict_table);
  table_wrapper->execute();

  const auto table_scan = create_table_scan(table_wrapper, ColumnID{0}, PredicateCondition::Equals, 123456);
  table_scan->execute();
  const auto projection =
      std::make_shared<opossum::Projection>(table_scan, expression_vector(a_b, a_a, add_(a_b, a_a)));
  projection->execute();

  const auto input_segment = dict_table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
  const auto input_dictionary_segment = std::dynamic_pointer_cast<DictionarySegment<int>>(input_segment);
  ASSERT_TRUE(input_dictionary_segment);

  const auto& output_table = projection->get_output();
  ASSERT_EQ(output_table->chunk_count(), 3);

  // Note that column a now has column id b
  const auto output_segment = output_table->get_chunk(ChunkID{0})->get_segment(ColumnID{1});
  const auto output_dictionary_segment = std::dynamic_pointer_cast<DictionarySegment<int>>(output_segment);
  ASSERT_TRUE(output_dictionary_segment);

  EXPECT_TRUE(output_dictionary_segment->dictionary() == input_dictionary_segment->dictionary());

  const auto& output_attribute_vector = output_dictionary_segment->attribute_vector();
  EXPECT_EQ(output_attribute_vector->size(), 1);

  const auto output_attribute_vector_decompressor = output_attribute_vector->create_base_decompressor();
  EXPECT_EQ(output_attribute_vector_decompressor->get(0), ValueID{1});
}
```

src/lib/operators/projection.cpp
```diff
@@ -120,27 +120,39 @@ std::shared_ptr<const Table> Projection::_on_execute() {
...
+                        create_filtered_and_compressed_pos_list(typed_segment, pos_list);
+                    const auto& dictionary = typed_segment.dictionary();
 
-                    auto compressed_attribute_vector =
-                        compress_vector(filtered_attribute_vector, VectorCompressionType::FixedSizeByteAligned, {});
                     output_segments[column_id] = std::make_shared<DictionarySegment<ColumnDataType>>(
                         dictionary, std::move(compressed_attribute_vector),
                         referenced_dictionary_segment->null_value_id());
+                  } else if constexpr (std::is_same_v<DictionarySegmentType,  // NOLINT
+                                                      FixedStringDictionarySegment<ColumnDataType>>) {
```
mrks: What happened to clang-tidy here? Is it drunk?

Bouncner: `src/lib/operators/projection.cpp:143:  If an else has a brace on one side, it should have it on both  [readability/braces] [5]`

src/lib/operators/projection.cpp
Bouncner: `// Note that column a now has column id b`: do you mean `column id 1`?

src/lib/operators/projection.cpp
```diff
@@ -120,27 +120,39 @@ std::shared_ptr<const Table> Projection::_on_execute() {
...
                     auto chunk_offset = ChunkOffset{0};
-                    iterable.with_iterators(pos_list, [&](auto it, auto end) {
+                    iterable.with_iterators(input_pos_list, [&](auto it, auto end) {
                       while (it != end) {
                         filtered_attribute_vector[chunk_offset] = it->value();
                         ++it;
                         ++chunk_offset;
                       }
                     });
+                    return std::make_shared<FixedSizeByteAlignedVector<uint32_t>>(std::move(filtered_attribute_vector));
```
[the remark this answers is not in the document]
Bouncner: Should have addressed all points concerning the lambda.

src/lib/operators/projection.cpp
```diff
@@ -133,21 +133,23 @@ std::shared_ptr<const Table> Projection::_on_execute() {
                         ++chunk_offset;
                       }
                     });
+                    // As dictionary segments expect a BaseCompressedVector, we use a max-width (i.e., uint32_t)
+                    // FixedByteAlignedVector as we can directly forward the new attribute vector.
```
mrks: DictionarySegments want a BaseCompressedVector, not an `std::vector<ValueId>` for the attribute vector. Luckily the latter can be wrapped into a FixedSizeByteAligned<uint32_t> without copying.

Bouncner:
```
// DictionarySegments take BaseCompressedVectors, not an std::vector<ValueId> for the attribute
// vector. But the latter can be wrapped into a FixedSizeByteAligned<uint32_t> without copying.
```

?

PR #1974, opened 2020-01-09
Enable encoding after binary import

Bouncner:
While working on another PR, I found that tables are not encoded when being imported (e.g, using `load tpch_cached_tables/sf-0.100000/lineitem.bin lineitem LZ4`)in the console (not only here). 

The reason was that the binary importer does not finalize chunk. This PR finalizes chunk in the binary parser and adds tests.

Maybe an open discussion: do we want to finalize the last chunk? I'd say no, since this is unexpected behaviour in my opinion. It also could lead to cases where we finalize a single chunk with a few rows.

### DIFF
--- a/src/lib/import_export/binary/binary_parser.cpp
+++ b/src/lib/import_export/binary/binary_parser.cpp
@@ -28,7 +28,7 @@ std::shared_ptr<Table> BinaryParser::parse(const std::string& filename) {
 
   auto [table, chunk_count] = _read_header(file);
   for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
-    _import_chunk(file, table);
+    _import_chunk(file, table, chunk_id < (chunk_count - 1));
   }
 
   return table;
@@ -98,7 +98,7 @@ std::pair<std::shared_ptr<Table>, ChunkID> BinaryParser::_read_header(std::ifstr
   return std::make_pair(table, chunk_count);
 }
 
-void BinaryParser::_import_chunk(std::ifstream& file, std::shared_ptr<Table>& table) {
+void BinaryParser::_import_chunk(std::ifstream& file, std::shared_ptr<Table>& table, const bool finalize_chunk) {
   const auto row_count = _read_value<ChunkOffset>(file);
 
   Segments output_segments;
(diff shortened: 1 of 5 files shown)


src/lib/import_export/binary/binary_parser.cpp
```diff
@@ -28,7 +28,7 @@ std::shared_ptr<Table> BinaryParser::parse(const std::string& filename) {
 
   auto [table, chunk_count] = _read_header(file);
   for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
-    _import_chunk(file, table);
+    _import_chunk(file, table, chunk_id < (chunk_count - 1));
```
mrks: I'd prefer for `chunk_id < (chunk_count - 1)` to be in its own variable first. When reading this, it is hard to guess what it means

Bouncner:
> > Maybe an open discussion: do we want to finalize the last chunk? I'd say no, since this is unexpected behaviour in my opinion. It also could lead to cases where we finalize a single chunk with a few rows.
> 
> Maybe this should be consistent with what the benchmarks are doing. I believe those are finalizing everything.

From a benchmarking point of view that makes sense. From a user/console point of view it's somewhat surprising. Your call.

src/lib/import_export/binary/binary_parser.cpp
Bouncner: Will change once we decide for or against finalization.

mrks:
@Bensk1 - any opinion on this?

src/lib/import_export/binary/binary_parser.cpp
```diff
@@ -109,6 +109,10 @@ void BinaryParser::_import_chunk(std::ifstream& file, std::shared_ptr<Table>& ta
 
   const auto mvcc_data = std::make_shared<MvccData>(row_count, CommitID{0});
   table->append_chunk(output_segments, mvcc_data);
+
+  if (finalize_chunk) {
+    table->get_chunk(ChunkID{table->chunk_count() - 1})->finalize();
```
Bensk1: Can we use the table's `last_chunk()` method here?

Bouncner: Ah ... wasn't aware of it. Will change.

Bouncner:
Just gave it a second choice and my opinion on console loading won't change, but we mostly use binaries for exporting/importing during benchmark runs and here, we could end up with a non-finalized dictionary-encoded last chunk.
I could check in the binary import what the encoding was, but that doesn't sound clean to me.

So I comply and we say all  loaded chunks are finalized?

Bensk1:
Hard to decide. We might finalize a chunk with a few values but we also might not finalize a chunk that is almost full and that is never being filled up. 

From a perspective what most people do with Hyrise I would probably finalize all chunks because it will likely be benchmarking.

mrks:
Sounds good.

Bouncner:
Changed to finalizing all binary loaded chunks.

PR #1978, opened 2020-01-10
Test SQL COPY import end-to-end

dey4ss:
refs #1933


mrks:
Mmmh, as this touches the server... :D

PR #1980, opened 2020-01-11
Fix Multi-threading issues

mrks:
The recent TPC-C fails were caused by improved performance, which caused more MT issues to surface

* Cache::try_get did not acquire the mutex, so the GDFSCache got corrupted
* TPCCTableGenerator called a method on an expired object
* JoinHash could not deal with the table size / chunk count increasing while the table was being processed

Once this is merged, I can bring Jenkins back to 4 workers.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -275,9 +275,9 @@ try {
             if (env.BRANCH_NAME == 'master' || full_ci) {
               sh "mkdir ./clang-release-memcheck-test"
               // If this shows a leak, try --leak-check=full, which is slower but more precise
-              sh "valgrind --tool=memcheck --track-origins=yes --error-exitcode=1 --gen-suppressions=all --num-callers=25 --suppressions=resources/.valgrind-ignore.txt ./clang-release/hyriseTest clang-release-memcheck-test --gtest_filter=-NUMAMemoryResourceTest.BasicAllocate"
-              sh "valgrind --tool=memcheck --track-origins=yes --error-exitcode=1 --gen-suppressions=all --num-callers=25 --suppressions=resources/.valgrind-ignore.txt ./clang-release/hyriseBenchmarkTPCH -s .01 -r 1 --scheduler --cores 10"
-              sh "valgrind --tool=memcheck --track-origins=yes --error-exitcode=1 --gen-suppressions=all --num-callers=25 --suppressions=resources/.valgrind-ignore.txt ./clang-release/hyriseBenchmarkTPCC -s 1 --scheduler --cores 10"
+              sh "valgrind --tool=memcheck --error-exitcode=1 --gen-suppressions=all --num-callers=25 --suppressions=resources/.valgrind-ignore.txt ./clang-release/hyriseTest clang-release-memcheck-test --gtest_filter=-NUMAMemoryResourceTest.BasicAllocate"
+              sh "valgrind --tool=memcheck --error-exitcode=1 --gen-suppressions=all --num-callers=25 --suppressions=resources/.valgrind-ignore.txt ./clang-release/hyriseBenchmarkTPCH -s .01 -r 1 --scheduler --cores 10"
+              sh "valgrind --tool=memcheck --error-exitcode=1 --gen-suppressions=all --num-callers=25 --suppressions=resources/.valgrind-ignore.txt ./clang-release/hyriseBenchmarkTPCC -s 1 --scheduler --cores 10"
             } else {
               Utils.markStageSkippedForConditional("memcheckReleaseTest")
             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -179,6 +179,7 @@ try {
               sh "LSAN_OPTIONS=suppressions=resources/.lsan-ignore.txt ASAN_OPTIONS=suppressions=resources/.asan-ignore.txt ./clang-debug-addr-ub-sanitizers/hyriseTest clang-debug-addr-ub-sanitizers"
               sh "LSAN_OPTIONS=suppressions=resources/.lsan-ignore.txt ASAN_OPTIONS=suppressions=resources/.asan-ignore.txt ./clang-debug-addr-ub-sanitizers/hyriseSystemTest ${tests_excluded_in_sanitizer_builds} clang-debug-addr-ub-sanitizers"
               sh "LSAN_OPTIONS=suppressions=resources/.lsan-ignore.txt ASAN_OPTIONS=suppressions=resources/.asan-ignore.txt ./clang-debug-addr-ub-sanitizers/hyriseBenchmarkTPCH -s .01 --verify -r 1"
+              sh "LSAN_OPTIONS=suppressions=resources/.lsan-ignore.txt ASAN_OPTIONS=suppressions=resources/.asan-ignore.txt ./scripts/test/hyriseBenchmarkTPCC_test.py clang-debug-addr-ub-sanitizers"
             } else {
               Utils.markStageSkippedForConditional("clangDebugAddrUBSanitizers")
             }
@@ -205,6 +206,7 @@ try {
               sh "LSAN_OPTIONS=suppressions=resources/.lsan-ignore.txt ASAN_OPTIONS=suppressions=resources/.asan-ignore.txt ./clang-release-addr-ub-sanitizers/hyriseTest clang-release-addr-ub-sanitizers"
               sh "LSAN_OPTIONS=suppressions=resources/.lsan-ignore.txt ASAN_OPTIONS=suppressions=resources/.asan-ignore.txt ./clang-release-addr-ub-sanitizers/hyriseSystemTest ${tests_excluded_in_sanitizer_builds} clang-release-addr-ub-sanitizers"
               sh "LSAN_OPTIONS=suppressions=resources/.lsan-ignore.txt ASAN_OPTIONS=suppressions=resources/.asan-ignore.txt ./clang-release-addr-ub-sanitizers/hyriseBenchmarkTPCH -s .01 --verify -r 100 --scheduler --clients 10"
+              sh "LSAN_OPTIONS=suppressions=resources/.lsan-ignore.txt ASAN_OPTIONS=suppressions=resources/.asan-ignore.txt ./scripts/test/hyriseBenchmarkTPCC_test.py clang-release-addr-ub-sanitizers"
             } else {
               Utils.markStageSkippedForConditional("clangReleaseAddrUBSanitizers")
             }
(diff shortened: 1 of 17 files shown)


src/benchmarklib/tpcc/tpcc_table_generator.cpp
```diff
@@ -513,28 +513,27 @@ std::unordered_map<std::string, BenchmarkTableInfo> TPCCTableGenerator::generate
...
+
+  return std::unordered_map<std::string, BenchmarkTableInfo>({{"ITEM", BenchmarkTableInfo{item_table}},
+                                                              {"WAREHOUSE", BenchmarkTableInfo{warehouse_table}},
+                                                              {"STOCK", BenchmarkTableInfo{stock_table}},
+                                                              {"DISTRICT", BenchmarkTableInfo{district_table}},
+                                                              {"CUSTOMER", BenchmarkTableInfo{customer_table}},
+                                                              {"HISTORY", BenchmarkTableInfo{history_table}},
+                                                              {"ORDER", BenchmarkTableInfo{order_table}},
+                                                              {"ORDER_LINE", BenchmarkTableInfo{order_line_table}},
+                                                              {"NEW_ORDER", BenchmarkTableInfo{new_order_table}}});
```
mrks: TSAN was unhappy about this. Since we don't yet use TPC-C frequently, I removed it for now.

src/benchmarklib/tpcc/tpcc_table_generator.hpp
```diff
@@ -164,7 +164,7 @@ class TPCCTableGenerator : public AbstractTableGenerator {
           }
 
           // reset data
-          data.clear();
+          data = decltype(data){};
```
mrks: `data` was moved before - we cannot call `clear()` on it

src/lib/cache/cache.hpp
```diff
@@ -24,69 +24,90 @@ class Cache {
...
+    std::shared_lock<std::shared_mutex> lock(_mutex);
+    return _impl->size();
+  }
 
   // Returns a reference to the underlying cache.
-  AbstractCacheImpl<Key, Value>& cache() { return *_impl; }
-  const AbstractCacheImpl<Key, Value>& cache() const { return *_impl; }
+  // TODO: Is this actually needed?
+  AbstractCacheImpl<Key, Value>& unsafe_cache() { return *_impl; }
+  const AbstractCacheImpl<Key, Value>& unsafe_cache() const { return *_impl; }
```
mrks: @Bouncner @Bensk1 ?

src/lib/operators/insert.cpp
```diff
@@ -253,7 +253,7 @@ void Insert::_on_rollback_records() {
 
     for (auto chunk_offset = target_chunk_range.begin_chunk_offset; chunk_offset < target_chunk_range.end_chunk_offset;
          ++chunk_offset) {
-      mvcc_data->begin_cids[chunk_offset] = 0u;
+      mvcc_data->set_begin_cid(chunk_offset, 0u);
```
mrks: So that the tsan suppression applies

resources/.tsan-ignore.txt
```diff
@@ -1,7 +1,7 @@
 # There are multiple data races identified in the uninstrumented libtbb (probably false positives)
 # However, be aware of concurrent_vector entries being potentially uninitialized:
 #   https://software.intel.com/en-us/blogs/2009/04/09/delusion-of-tbbconcurrent_vectors-size-or-3-ways-to-traverse-in-parallel-correctly
-race:^tbb::*
+race:tbb::*
```
Bouncner: What's that?

Jenkinsfile
```diff
@@ -226,6 +230,7 @@ try {
               sh "TSAN_OPTIONS=\"history_size=7 suppressions=resources/.tsan-ignore.txt\" ./clang-relwithdebinfo-thread-sanitizer/hyriseTest clang-relwithdebinfo-thread-sanitizer"
               sh "TSAN_OPTIONS=\"history_size=7 suppressions=resources/.tsan-ignore.txt\" ./clang-relwithdebinfo-thread-sanitizer/hyriseSystemTest ${tests_excluded_in_sanitizer_builds} clang-relwithdebinfo-thread-sanitizer"
               sh "TSAN_OPTIONS=\"history_size=7 suppressions=resources/.tsan-ignore.txt\" ./clang-relwithdebinfo-thread-sanitizer/hyriseBenchmarkTPCH -s .01 --verify -r 100 --scheduler --clients 10"
+              sh "TSAN_OPTIONS=\"history_size=7 suppressions=resources/.tsan-ignore.txt\" ./scripts/test/hyriseBenchmarkTPCC_test.py clang-relwithdebinfo-thread-sanitizer"
```
Bouncner: I am for the change. But have you an idea about the run time?

scripts/test/hyriseBenchmarkTPCC_test.py
```diff
@@ -30,15 +29,14 @@ def main():
   arguments = {}
   arguments["--scale"] = "1"
   arguments["--time"] = "60"
-  arguments["--runs"] = "100"
```
Bouncner: No idea about TPC-C. Is that an intended change? Is one run sufficient?

src/benchmarklib/tpcc/tpcc_table_generator.cpp
```diff
@@ -513,28 +513,27 @@ std::unordered_map<std::string, BenchmarkTableInfo> TPCCTableGenerator::generate
...
-      {{"ITEM", BenchmarkTableInfo{item_table.get()}},
-       {"WAREHOUSE", BenchmarkTableInfo{warehouse_table.get()}},
-       {"STOCK", BenchmarkTableInfo{stock_table.get()}},
-       {"DISTRICT", BenchmarkTableInfo{district_table.get()}},
-       {"CUSTOMER", BenchmarkTableInfo{customer_table.get()}},
-       {"HISTORY", BenchmarkTableInfo{history_table.get()}},
-       {"ORDER", BenchmarkTableInfo{order_table.get()}},
-       {"ORDER_LINE", BenchmarkTableInfo{order_line_table.get()}},
-       {"NEW_ORDER", BenchmarkTableInfo{new_order_table.get()}}});
+  auto item_table = generate_item_table();
```
Bouncner: Curious: has this been an issue with asynchronous creation?

src/lib/cache/cache.hpp
```diff
@@ -24,69 +24,90 @@ class Cache {
...
-  size_t size() const { return _impl->size(); }
+  size_t size() const {
+    std::shared_lock<std::shared_mutex> lock(_mutex);
+    return _impl->size();
+  }
 
   // Returns a reference to the underlying cache.
-  AbstractCacheImpl<Key, Value>& cache() { return *_impl; }
-  const AbstractCacheImpl<Key, Value>& cache() const { return *_impl; }
+  // TODO: Is this actually needed?
```
Bouncner: I did not follow the whole "caches no longer global" discussion. Is this a place where we need this access?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -162,17 +162,17 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
     // Executes the given action for every row id of the table in this range.
     template <typename F>
     void for_every_row_id(std::unique_ptr<MaterializedSegmentList<T>>& table, F action) {
-      for (size_t cluster = start.cluster; cluster <= end.cluster; ++cluster) {
-        size_t start_index = (cluster == start.cluster) ? start.index : 0;
-        size_t end_index = (cluster == end.cluster) ? end.index : (*table)[cluster]->size();
 // False positive with gcc and tsan (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92194)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
```
Bouncner: Could we use `[[maybe unused]]` here? Or does TSAN ignore it?

src/test/cache/cache_test.cpp
```diff
@@ -246,7 +246,8 @@ TEST(CachePolicyTest, Iterators) {
   auto element_count = size_t{0};
   auto value_sum = size_t{0};
 
-  for (const auto& [key, value] : cache) {
+  for (auto it = cache.unsafe_begin(); it != cache.unsafe_end(); ++it) {
```
Bouncner: So the rename of the method leads to not being able to use the range-loops anymore? It might be just me, but I would have thought that it's clear that iterating a cache should be used with great caution.

resources/.tsan-ignore.txt
mrks: Not sure about this one, but I feel like less tbb false-positives popped up with this change.

Jenkinsfile
mrks: Nope, will see once the CI run completes.

scripts/test/hyriseBenchmarkTPCC_test.py
mrks: `runs` defaults to infinity. 100 was way too low.

src/benchmarklib/tpcc/tpcc_table_generator.cpp
mrks: Did Github hide my comment again? :(
TSan didn't like it. I can't find anything wrong with it, but as long as we don't use TPC-C frequently and this only saves a couple of seconds, I rather be on the save side.

src/lib/cache/cache.hpp
mrks: You tell me. The code base doesn't use it. If you and @Bensk1 don't use it in your code, I'd throw it out

src/lib/operators/join_sort_merge.cpp
mrks: uninitialized, not unused.

src/test/cache/cache_test.cpp
mrks: > So the rename of the method leads to not being able to use the range-loops anymore?

Yes.

> I would have thought that it's clear that iterating a cache should be used with great caution.

I believe that someone looking at the class and seeing a mutex being used for get/set is right to believe that the class transparently synchronizes accesses. Mixing thread-safe and -unsafe methods in one class is almost a guarantee for disaster. In my opinion, using range loops is by far the lesser evil.

src/lib/cache/cache.hpp
Bouncner: In my plugin, I use `for (const auto& [query_string, physical_query_plan] : *SQLPipelineBuilder::default_pqp_cache)`. Can't find a use of `cache()`.

@Bensk1 ?

src/benchmarklib/tpcc/tpcc_table_generator.cpp
Bouncner: I was reviewing when you added the comment. Saw it afterwards and makes sense.

resources/.tsan-ignore.txt
Bouncner: Can we at least have a comment explaining the actual line? I am wondering what the change actually does.

src/test/cache/cache_test.cpp
Bouncner: Sounds reasonable to me.

mrks:
CI runtime is not increased as the third phase is dominated by other steps. Can you confirm the zünderability of the knaller?

Jenkinsfile
```diff
@@ -137,7 +137,6 @@ try {
               sh "./scripts/test/hyriseConsole_test.py gcc-debug"
               sh "./scripts/test/hyriseBenchmarkJoinOrder_test.py gcc-debug"
               sh "./scripts/test/hyriseBenchmarkFileBased_test.py gcc-debug"
-              sh "./scripts/test/hyriseBenchmarkTPCC_test.py gcc-debug"
```
Bouncner: Reason?

Jenkinsfile
```diff
@@ -272,7 +276,8 @@ try {
               sh "mkdir ./clang-release-memcheck-test"
               // If this shows a leak, try --leak-check=full, which is slower but more precise
               sh "valgrind --tool=memcheck --error-exitcode=1 --gen-suppressions=all --num-callers=25 --suppressions=resources/.valgrind-ignore.txt ./clang-release/hyriseTest clang-release-memcheck-test --gtest_filter=-NUMAMemoryResourceTest.BasicAllocate"
-              sh "valgrind --tool=memcheck --error-exitcode=1 --gen-suppressions=all --num-callers=25 --suppressions=resources/.valgrind-ignore.txt ./clang-release/hyriseBenchmarkTPCH -s .01 --verify -r 1"
+              sh "valgrind --tool=memcheck --error-exitcode=1 --gen-suppressions=all --num-callers=25 --suppressions=resources/.valgrind-ignore.txt ./clang-release/hyriseBenchmarkTPCH -s .01 -r 1 --scheduler --cores 10"
```
Bouncner: You removed the `track_origins` flag again. Reason?

mrks: Too slow

Jenkinsfile
mrks: This whole step is the slowest one, so we rather execute TPC-C in a different step

src/lib/cache/cache.hpp
```diff
@@ -24,69 +24,84 @@ class Cache {
...
 
-  // Returns the access frequency of a cached item (=1 for set(), +1 for each get()).
-  // Returns 0 for keys not being cache-resident.
-  size_t size() const { return _impl->size(); }
+  void resize(size_t capacity) {
+    std::unique_lock<std::shared_mutex> lock(_mutex);
+    _impl->resize(capacity);
+  }
 
-  // Returns a reference to the underlying cache.
```
Bouncner: Overlooked that. Reason to remove it?

mrks:
It was unused in the main code base. If you need it for the things you do with the LQP/PQP, feel free to bring it back as `unsafe_cache`.


________________________________

PR #1982, opened 2020-01-11
Fix crashing randomized test: missing fixtures keeps storage manager alive

Bouncner:
#1968 changed a test to be non-fixtured (changed from `TEST_F` to `TEST`) which uses the storage manager. This was dump because the storage manager is not reset after the test.
For the randomized tests of the CI, by chance there could be tests following that also create a table `int_int` (e.g., the union tests). This will crash since the table already exists.

I grepped for tests that contain the `storage_manager` and do not use fixtures. The found tests look good. I also looked for cases that do not inherit from BaseTest, but did not find one that uses the storage manager.

The greps:
```
find . -type f -exec grep -q storage_manager {} \; -exec grep -l '^TEST(' {} \;
find . -type f -exec grep -q storage_manager {} \; -exec grep -l 'Test : public ::testing::Test' {} \;
```

We should not allow noobs to write tests.

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -21,15 +21,15 @@ if [ ! -z "$output" ]; then
 	exitcode=1
 fi
 
-# Gtest's TEST() should not be used. Use TEST_F() instead.
+# Gtest's TEST() should not be used. Use TEST_F() instead. This might require additional test classes but ensures that state is cleaned up properly.
 output=$(grep -rn '^TEST(' src/test | sed 's/^\([a-zA-Z/._]*:[0-9]*\).*/\1  TEST() should not be used as it does not clean up global state (e.g., the Hyrise singleton)./')
 if [ ! -z "$output" ]; then
 	echo "$output"
 	exitcode=1
 fi
 
-# Gtest's TEST() should not be used. Use TEST_F() instead.
-output=$(grep -rn ':\w*(?:public|protected|private)?\w+::testing::Test' src/test | sed 's/^\([a-zA-Z/._]*:[0-9]*\).*/\1  TEST() should not be used as it does not clean up global state (e.g., the Hyrise singleton)./')
+# Tests should inherit from BaseTest or BaseTestWithParams of base_test.hpp to ensure proper destruction.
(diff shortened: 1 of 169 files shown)


mrks:
Could you remind me why we use non-fixture tests at all?

Bouncner:
I’d say performance (no `SetUp()` being called what make a real difference) and stating that this is an isolated test (which is good if no idiot would use globals in there).

mrks:
How much performance would we lose if we switched all `TEST(` with `TEST_F(`?

> stating that this is an isolated test

Mmmh, not sure what this information is good for, especially considering how it can silently break things if someone down the line uses, e.g., the StorageManager.

Bouncner:
Personally, I like the information that a test is a “pure unit” test that does not have a full setup created upfront (and these setups are sometimes quite large).

Will test the performance change.

Bouncner:
I changed most of the tests from TEST() to TEST_F(), excluding a few where only single functions are tests (e.g. test of format utils).

I could not measure a difference in runtime. It's just a tiny number of tests that have changed.

Bouncner:
We can get rid of all TEST() uses and might be able to flag them in the future, but we would need small class instances (like in the meta table tests: `class MetaTableManagerTest : public BaseTest {};`) even for the smallest tests.

I have no opinions on that. I actually like TEST(), but I agree that it can be problematic (would I just proved with the last commit to master).

mrks:
> but we would need small class instances (like in the meta table tests: class MetaTableManagerTest : public BaseTest {};) even for the smallest tests.

Works for me. Alternatively, I believe we could also use `TEST_F(BaseTest, foo)`, but that doesn't look nice in the console output.

> might be able to flag them in the future

Given that test fails caused by a previous test not properly cleaning up are hard to debug, making lint.sh complain sounds like a good thing.

Bouncner:
> Works for me. Alternatively, I believe we could also use `TEST_F(BaseTest, foo)`, but that doesn't look nice in the console output.

Naaaa, veto. I can live with the additional line but not with having unit tests grouped in some "random" test suite. Igittifui. :)

> Given that test fails caused by a previous test not properly cleaning up are hard to debug, making lint.sh complain sounds like a good thing.

Maybe have a coffee together? Never worked with the linter tools.

Bouncner:
Ah, it's easier than expected. I am on it. :)

mrks:
What's WIP about this?

Bouncner:
Only the upcoming removal of gtest includes. Wanted to avoid the merge.

Feel free to review.

Bouncner:
No longer WIP.

PR #1984, opened 2020-01-13
Fix two small issues in the micro benchmarks

Bouncner:
The effect of not running the micro benchmarks has shown. One issue in the sort table scan benchmark was that the `ChunkEncoder` no longer supports encoding of mutable chunks (it's now the writers task to finalize chunks). The other issue was that a scale factor of 0.001f for the TPC-H data generator is no longer possible. Changed to 0.01f.


mrks:
Did this already include #1980?

Bouncner:
Nope, no it does.

PR #1986, opened 2020-01-14
Support SQL COPY for Export

dey4ss:
refs #1933


src/lib/operators/export.cpp
```diff
@@ -10,7 +10,10 @@
 namespace opossum {
 
 Export::Export(const std::shared_ptr<const AbstractOperator>& in, const std::string& filename, const FileType& type)
-    : AbstractReadOnlyOperator(OperatorType::Export, in), _filename(filename), _type(type) {}
+    : AbstractReadOnlyOperator(OperatorType::Export, in), _filename(filename), _type(type) {
```
dey4ss: I do not know whether it is better to use an input operator here or, to be kind of align with the ImportOperator, only provide a table name, which is then accessed from the storage manager in the ::on_execute(). This, for example, has an influence whether we need to check for the table existence in `sql_pipeline_statement.cpp` (when we provide a table name) or in `lqp_translator.cpp` or even in `sql_translator.cpp` (when we use an input operator).

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -309,6 +310,12 @@ void SQLPipelineStatement::_precheck_ddl_operators(const std::shared_ptr<Abstrac
       AssertInput(file.good(), "There is no file '" + import->filename + "'.");
       break;
     }
+    /*case OperatorType::Export: {
+      const auto export_table = std::dynamic_pointer_cast<Export>(pqp);
+      AssertInput(storage_manager.has_table(export_table->tablename),
+                  "There is no table '" + export_table->tablename + "'.");
+      break;
+    }*/
```
Bensk1: Is this till WIP?

dey4ss: Exactly, this is related to my question above. Giving the operator another operator as input vs. a table name decides on where we check if this table exists. With a table name we can check here, with an operator we need either to check in LQPTranslator or even SQLTranslator.

src/lib/operators/export.cpp
mrks: Is there any reason not to use an input operator?

dey4ss: Using a table name aligns with DropTable, DropView etc. that also access data maintained by the StorageManager. Checking if this data (i.e. table) exists can then also be done in SQLPipelineStatement, where it is done for the other operators.

mrks: > that also access data maintained by the StorageManager

So does pretty much everyone else.

The big benefit of making this an operator (which I just thought of now) is that it respects the MVCC visiblity of rows.

mrks:
What's the status here?

Bensk1:
I will have a look sometime today.

src/lib/logical_query_plan/export_node.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include "base_non_query_node.hpp"
+#include "enable_make_for_lqp_node.hpp"
+#include "import_export/file_type.hpp"
+#include "storage/table_column_definition.hpp"
+
+namespace opossum {
+
+/**
+ * This node type represents the COPY TO management command.
+ */
```
Bensk1: The ImportNode says, `IMPORT / COPY FROM management command.` Should we say `EXPORT /...` here?

src/lib/sql/sql_translator.cpp
```diff
@@ -1214,6 +1217,16 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_import(const hsql::Im
...
 }
 
+// NOLINTNEXTLINE - while this particular method could be made static, others cannot.
+std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_export(const hsql::ExportStatement& export_statement) {
+  // Get validated table as input
+  auto sql_identifier_resolver = std::make_shared<SQLIdentifierResolver>();
+  auto lqp = _translate_stored_table(export_statement.tableName, sql_identifier_resolver);
+
+  return ExportNode::make(export_statement.tableName, export_statement.filePath,
+                          import_type_to_file_type(export_statement.type), lqp);
```
Bensk1: import_type sounds strange in the context of an export.

src/test/operators/export_test.cpp
```diff
@@ -123,37 +123,34 @@ TEST_F(OperatorsExportTest, NonsensePath) {
...
   auto table_wrapper = std::make_shared<TableWrapper>(std::move(table));
   table_wrapper->execute();
-  auto exporter = std::make_shared<opossum::Export>(table_wrapper, test_filename + ".bin");
-  exporter->execute();
-
-  // need to load table again since it moved
-  table = load_table("resources/test_data/tbl/float.tbl");
 
-  EXPECT_TABLE_EQ_ORDERED(exporter->get_output(), table);
+  EXPECT_THROW(std::make_shared<opossum::Export>(table_wrapper, "not_existing_file.mp3"), std::exception);
```
Bensk1: Not specific to this line. I just chose a random place to put this comment. Should we test that COPY FROM/IMPORT and COPY TO/EXPORT are correctly translated with SQLTranslatorTests?

src/lib/logical_query_plan/export_node.hpp
dey4ss: The SQLParser supports both the old IMPORT FROM ... INTO and COPY ... FROM ... and generates an ImportStatement from both, while it only makes an ExportStatement from COPY ... TO ...
That's why we have a difference here.

src/lib/sql/sql_translator.cpp
dey4ss: Actually, the parser uses an ImportType to indicate the file type both for ImportStatement and ExportStatement due to we wanted to be downwards compatible there

src/test/operators/export_test.cpp
dey4ss: I will add this

src/lib/logical_query_plan/export_node.hpp
Bensk1: Alright

src/lib/sql/sql_translator.cpp
Bensk1: So it is an ImportExportType? But it's probably not worth changing it.

dey4ss: We use it for both, exactly, but its name in the parser stayed ImportType so others would not need to adapt their code if they use an ImportStatement anywhere.

src/test/operators/export_test.cpp
Bensk1: This could also be an operator test, couldn't it?

dey4ss: I'm not sure since we want to check if the input operator of the ExportOperator is a ValidateOperator in the LQP.  So this is kind of out of scope of the operator test. We check if a ValidateOperator is part of the LQP when the SQL query is translated with UseMvcc::True, but I thought more of an integration test (so if everything works when running the server).

Bensk1: Ah, I see, I misread your previous post. Checking the correct input for the ExportOperator could be part of the SQLTranslatorTest?

Bensk1: Also, we could add a note at the end of sqlite_testrunnter_queries.sql that we cannot test copy there.

dey4ss: This is what I added recently, we ensure to have a ValidateNode as input of the ExportNode when UseMvcc::Yes.

Bensk1: Discussed offline.

src/lib/logical_query_plan/export_node.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include "constant_mappings.hpp"
+
+namespace opossum {
+
+ExportNode::ExportNode(const std::string& init_tablename, const std::string& init_filename,
+                       const FileType init_file_type)
+    : BaseNonQueryNode(LQPNodeType::Export),
+      tablename(init_tablename),
+      filename(init_filename),
```
mrks: table_name / file_name? I think that's more consistent with the rest of the code.

src/lib/sql/sql_translator.cpp
```diff
@@ -1214,6 +1217,16 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_import(const hsql::Im
                           import_type_to_file_type(import_statement.type));
 }
 
+// NOLINTNEXTLINE - while this particular method could be made static, others cannot.
+std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_export(const hsql::ExportStatement& export_statement) {
+  // Get validated table as input
```
mrks: ... if MVCC is enabled

src/test/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -964,4 +969,24 @@ TEST_F(LQPTranslatorTest, CreatePreparedPlan) {
   EXPECT_EQ(prepare->prepared_plan(), prepared_plan);
 }
 
+TEST_F(LQPTranslatorTest, Export) {
+  const auto lqp = ExportNode::make("a_table", "a_file.tbl", FileType::Auto, ValidateNode::make(int_float_node));
```
mrks: Please have a look at other LQP tests regarding the formatting. ValidateNode should be on its own line and indented.

src/test/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -964,4 +969,24 @@ TEST_F(LQPTranslatorTest, CreatePreparedPlan) {
...
+
+  const auto pqp = LQPTranslator{}.translate_node(lqp);
+  const auto exporter = std::dynamic_pointer_cast<Export>(pqp);
+
+  EXPECT_EQ(exporter->type(), OperatorType::Export);
+  EXPECT_EQ(exporter->input_left()->type(), OperatorType::Validate);
+}
+
+TEST_F(LQPTranslatorTest, Import) {
```
mrks: Not an LQPTranslator thing, but do we have a test that makes sure that after an import, MVCC information is correctly set?

src/test/server/server_test_runner.cpp
```diff
@@ -40,6 +41,8 @@ class ServerTestRunner : public BaseTest {
     // Give the server time to shut down gracefully before force-closing the socket it's working on
     std::this_thread::sleep_for(std::chrono::milliseconds(10));
     _server_thread->join();
+
+    std::remove(_export_filename.c_str());
```
mrks: What happens to following tests if the process dies before this is reached?

src/test/server/server_test_runner.cpp
```diff
@@ -102,6 +106,78 @@ TEST_F(ServerTestRunner, TestCopyImport) {
...
+  pqxx::connection connection{_connection_string};
+
+  // We use nontransactions because the regular transactions use "begin" and "commit" keywords that we do not support.
+  // Nontransactions auto commit.
+  pqxx::nontransaction transaction{connection};
+
+  const auto result = transaction.exec("COPY table_a TO '" + _export_filename + "';");
+
+  std::ifstream file{_export_filename};
+  EXPECT_TRUE(file.good());
```
mrks: Check that the file is equal to something expected?

src/test/server/server_test_runner.cpp
```diff
@@ -102,6 +106,78 @@ TEST_F(ServerTestRunner, TestCopyImport) {
...
+  // Check whether server is still running and connection established
+  const auto result = transaction.exec("SELECT * FROM table_a;");
+  EXPECT_EQ(result.size(), _table_a->row_count());
+}
+
+TEST_F(ServerTestRunner, TestCopyIntegration) {
+  pqxx::connection connection{_connection_string};
+
+  // We use nontransactions because the regular transactions use "begin" and "commit" keywords that we do not support.
+  // Nontransactions auto commit.
```
mrks: @klauck We should clean up all of these comments once we actually do support it.

src/test/sql/sql_translator_test.cpp
```diff
@@ -2604,4 +2608,63 @@ TEST_F(SQLTranslatorTest, WithClauseTableMasking) {
   EXPECT_LQP_EQ(actual_lqp, expected_lqp);
 }
 
+TEST_F(SQLTranslatorTest, CopyStatementImport) {
+  const auto actual_lqp_a = compile_query("COPY a_table FROM 'a_file.tbl';");
```
mrks: This makes it easier to follow a single test:

```
{
  const auto actual_lqp = compile_query("COPY a_table FROM 'a_file.tbl';");
  const auto expected_lqp = ImportNode::make("a_table", "a_file.tbl", FileType::Auto);
}
```

src/test/server/server_test_runner.cpp
dey4ss: A following test here (either `TestCopyExport` or `TestCopyIntegration`) could accidentally be green as we do not compare the files. I could copy the file comparison from `BinaryWriterTest` to at least make sure the other test would fail in advance. For re-running, if the file still exists, we could still have a false positive, which is also the case for any other export test. Therefore, I think removing the file in `SetUp()` rather than in `TearDown()` would be a better  solution.

src/test/server/server_test_runner.cpp
dey4ss: I could copy that from `BinaryWriterTest`.

src/test/logical_query_plan/lqp_translator_test.cpp
dey4ss: I doubt it, so I'm gonna add a case in `ImportTest`.

dey4ss:
I just wonder if the input node should also be copied at `ExportNode::on_deep_copy()` (I guess yes)?

mrks:
No. Inputs are handled by the abstract class. That's the same for hash, copy, and equals. Could you add that to abstract_lqp_node.hpp, please?

src/test/server/server_test_runner.cpp
```diff
@@ -45,6 +46,30 @@ class ServerTestRunner : public BaseTest {
...
+    std::istreambuf_iterator<char> iterator_created(created);
+    std::istreambuf_iterator<char> end;
+
+    while (iterator_original != end && iterator_created != end) {
+      if (*iterator_original != *iterator_created) return false;
+      ++iterator_original;
+      ++iterator_created;
+    }
+    return ((iterator_original == end) && (iterator_created == end));
+  }
```
mrks: Deduplicate and move to base_test? Maybe we should also move this and the existing implementations from base_test.hpp to base_test.cpp.

dey4ss: I just moved all method implementations from base_test.hpp to base_test.cpp. The code compiles, but when running the test binary, I get errors like

```
dyld: lazy symbol binding failed: Symbol not found: __ZN7opossum17BaseTestWithParamIvE38get_supporting_segment_encodings_specsENS_6detail8DataTypeEb
  Referenced from: /Users/epic/h2/cmake-build-debug/./hyriseTest
  Expected in: flat namespace

dyld: Symbol not found: __ZN7opossum17BaseTestWithParamIvE38get_supporting_segment_encodings_specsENS_6detail8DataTypeEb
  Referenced from: /Users/epic/h2/cmake-build-debug/./hyriseTest
  Expected in: flat namespace
```

This occurs for different methods if I run different tests.

mrks: When implementing templates in the cpp file you need to instantiate them. Can you get around using templates? For example, I don't see why `BaseTestWithParam<ParamType>::get_supporting_segment_encodings_specs` can't be free-floating.

dey4ss: Okay, I thought there was a good reason for the methods being members of BaseTestWithParam. As none of them uses the ParamType, it would mean to exclude all of them (besides the destructor) from the class, right? If this is the case, should they rather be moved to a separate pair of cpp and hpp file that is included by base_test.hpp as they would not really have something in common anymore?

mrks: Meh, I don't care too much about clean headers in the test. If it's a separate file, base_test should include it for convenience, so... whatever.

dey4ss: Alright, I will not move the code to a separate header then.

dey4ss:
> have a look at `EXPLICITLY_INSTANTIATE_DATA_TYPES`

As far as I understand it, that Macro instantiates all templated classes for the data types. So making `create_dict_segment_by_type()` a member of BaseTestWithParam again and using the Macro does not help, as we want the instantiations not for the class template but for the additional function template, right? So should I add something like `EXPLICITLY_INSTANTIATE_DATA_TYPES_FOR_METHOD` that takes a method and its return type?

mrks:
If that doesn't take too much time, that would be great. We'd like to limit the number of places where the list of supported data types appear.

dey4ss:
> If that doesn't take too much time, that would be great. We'd like to limit the number of places where the list of supported data types appear.

I'll have a look a it

mrks:
Would it be easier if you didn't have to write a general macro used by everyone but would just use `BOOST_PP_SEQ_FOR_EACH(..., ..., DATA_TYPES)` immediately?

dey4ss:
> Would it be easier if you didn't have to write a general macro used by everyone but would just use `BOOST_PP_SEQ_FOR_EACH(..., ..., DATA_TYPES)` immediately?

I basically did this and defined a Macro to instantiate the specific function, which is called by `BOOST_PP_SEQ_FOR_EACH` ignoring the data.

PR #1989, opened 2020-01-15
Fix generate_tpcc in Console (#1988)

### LINKED ISSUE #1988: Console crashes on generate_tpcc

Bensk1:
Fixes #1988. While I was looking at it, I made the data type for `num_warehouses` consistent because the TPCCTableGenerator was taking an `int` in the constructor and then storing that into a `size_t`.


Bensk1:
@mrks have you seen that before? https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-1989/4/pipeline/

PR #1990, opened 2020-01-20
Enable Cache Implementation Access

Bouncner:
This PR brings back the possibility to access the cache implementation that is used. This is necessary to access certain methods that are not implemented for all cache implementations (e.g., the access counter via `frequency()` in GDFS). This method is called `unsafe_cache()` in accordance with the recent changes to the cache (#1980) and to denote that this access is inherently unsafe as concurrent modifications are likely to happen.

Further:
 * added a test that documents the behaviour of `default_pqp_cache` which was not obvious to me, thus it might help others as well
 * removed `frequency()` from cache.hpp as it is not part of the abstract cache interface anyways (see discussion in #1932)

### DIFF
--- a/src/test/sql/query_plan_cache_test.cpp
+++ b/src/test/sql/query_plan_cache_test.cpp
@@ -148,18 +148,18 @@ TEST_F(QueryPlanCacheTest, AutomaticQueryOperatorCacheLRUK2) {
 TEST_F(QueryPlanCacheTest, CachedPQPFrequencyCount) {
   // Create pipeline and pass pqp cache. Verify that this does not change default_pqp_cache.
   auto sql_pipeline = SQLPipelineBuilder{Q1}.with_pqp_cache(cache).create_pipeline_statement();
-  EXPECT_EQ(SQLPipelineBuilder::default_pqp_cache, nullptr);
+  EXPECT_EQ(Hyrise::get().default_pqp_cache, nullptr);
 
   // Setting default_pqp_cache and verify it's set.
-  SQLPipelineBuilder::default_pqp_cache = cache;
-  EXPECT_NE(SQLPipelineBuilder::default_pqp_cache, nullptr);
+  Hyrise::get().default_pqp_cache = cache;
+  EXPECT_NE(Hyrise::get().default_pqp_cache, nullptr);
 
   // Create new pipeline, without setting a cache (default cache set previously). Execute pipeline and check if
   // frequency of query is as expected.
   auto new_sql_pipeline = SQLPipelineBuilder{Q1}.create_pipeline_statement();
   new_sql_pipeline.get_result_table();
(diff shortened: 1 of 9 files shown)


src/test/sql/query_plan_cache_test.cpp
```diff
@@ -143,4 +143,24 @@ TEST_F(QueryPlanCacheTest, AutomaticQueryOperatorCacheLRUK2) {
...
 
+// Check access to PQP cache. When set, check the underlying cache implementation, and verify that it is a GDFS cache
+// that supports retrieving the cache frequency count.
+TEST_F(QueryPlanCacheTest, CachedPQPFrequencyCount) {
+  // Create pipeline and pass pqp cache. Verify that this does not change default_pqp_cache.
+  auto sql_pipeline = SQLPipelineBuilder{Q1}.with_pqp_cache(cache).create_pipeline_statement();
+  EXPECT_EQ(SQLPipelineBuilder::default_pqp_cache, nullptr);
+
+  // Setting default_pqp_cache and verify it's set.
+  SQLPipelineBuilder::default_pqp_cache = cache;
```
mrks: This looks like it might break test isolation? Also, should Hyrise::reset reset the default caches?

Bouncner: The clean up executes:
```
~BaseTestWithParam() {
    Hyrise::reset();
    SQLPipelineBuilder::default_pqp_cache = nullptr;
    SQLPipelineBuilder::default_lqp_cache = nullptr;
  }
```

Thus, I think it's safe here. Now that you ask, I'd vote for `Hyrise::reset()` resetting the default caches. But I am not sure what the motivation was not to do it.

Bouncner: I'll address it with the latest issue on BaseTest?

Bouncner: If you agree, we can resolve and proceed.

mrks: Looks like the other thing will be part of #1986, so maybe do this one here?

mrks:
I was just about to add the missing header, but then started thinking about why those default caches are still static. Shouldn't those be part of the Hyrise class?

mrks:
(also see https://github.com/hyrise/hyrise/pull/1615)

Bouncner:
To be honest: I have no idea what the initial intention to build it that way was. So I am not the one to answer your question.

Whatever makes sense: I'll do it or you can go ahead and push it here. No opinion as long as I am able to access the caches in some way.

mrks:
We just didn't have the Hyrise class at the time.

Bouncner:
Hmm. I am not sure how to implement this change. :-/

mrks:
> Should the hyrise object know about caches? Not sure.

Certainly better than having a static object with all associated problems (easy to forget when resetting, static initialization order fiasco, ...).

Please have a look and zünd the knaller if you agree.

PR #1993, opened 2020-01-20
[DYOD] Support for Transaction Statements

youri-k:
Open TODOs:
- [x] error handling
- [x] correct output for transaction statements
- [x] transaction specific tests


youri-k:
@Dencrash even though there are a few open todos, we think your group can start reviewing this :slightly_smiling_face:

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -191,30 +201,67 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
...
+      return {std::make_shared<JobTask>([this] {
+        if (!_transaction_context || !_transaction_context->is_auto_commit())
+          this->set_warning_message(std::string("WARNING: Cannot begin transaction inside an active transaction."));
+        else
+          _transaction_context = Hyrise::get().transaction_manager.new_transaction_context(false);
+      })};
+    case hsql::kCommitTransaction:
+      return {std::make_shared<JobTask>([this] {
+        if (!_transaction_context || _transaction_context->is_auto_commit())
+          this->set_warning_message(std::string("WARNING: Cannot commit since there is no active transaction."));
```
mrks: Why is this a warning, not an (input) assertion? Also, information should not be passed around in strings. If a string is necessary, it should be generated at the latest moment possible.

src/test/sql/sql_pipeline_test.cpp
```diff
@@ -445,11 +445,11 @@ TEST_F(SQLPipelineTest, UpdateWithTransactionFailure) {
   const auto sql =
       "UPDATE table_a SET a = 1 WHERE a = 12345; UPDATE table_a SET a = 1 WHERE a = 123; "
       "UPDATE table_a SET a = 1 WHERE a = 1234";
-  auto transaction_context = Hyrise::get().transaction_manager.new_transaction_context();
+  auto transaction_context = Hyrise::get().transaction_manager.new_transaction_context(false);
```
mrks: The `false` argument is not understandable without looking into the method. Consider using an `enum class` (compare to UseMvcc::Yes/No):

https://akrzemi1.wordpress.com/2017/02/16/toggles-in-functions/

src/lib/concurrency/transaction_context.cpp
```diff
@@ -10,11 +10,13 @@
 
 namespace opossum {
 
-TransactionContext::TransactionContext(const TransactionID transaction_id, const CommitID snapshot_commit_id)
+TransactionContext::TransactionContext(const TransactionID transaction_id, const CommitID snapshot_commit_id,
+                                       bool is_auto_commit)
```
mrks: Same thing as in the previous comment (hoping Github keeps the order): bool parameters make it hard to understand the function call. Same for `rollback` and `_mark_as_rolled_back`

src/lib/concurrency/transaction_context.hpp
```diff
@@ -80,14 +81,14 @@ class TransactionContext : public std::enable_shared_from_this<TransactionContex
...
-   * Returns true if rollback has been called.
+   * Returns true if transaction has been aborted.
    */
   bool aborted() const;
 
   /**
    * Aborts and rolls back the transaction.
    */
-  void rollback();
+  void rollback(bool is_explicit = false);
```
mrks: Please doc the parameter and why this information is necessary.

src/lib/concurrency/transaction_manager.hpp
```diff
@@ -56,7 +56,7 @@ class TransactionManager : public Noncopyable {
   /**
    * Creates a new transaction context
    */
-  std::shared_ptr<TransactionContext> new_transaction_context();
+  std::shared_ptr<TransactionContext> new_transaction_context(bool is_auto_commit = true);
```
mrks: Does it make sense for this to be the default?

mrks: Please doc the parameter and what impact it has.

src/lib/server/query_handler.cpp
```diff
@@ -83,4 +89,32 @@ std::shared_ptr<const Table> QueryHandler::execute_prepared_plan(
...
 }
 
+void QueryHandler::handle_transaction_statement_message(ExecutionInformation& execution_info,
+                                                        SQLPipeline& sql_pipeline) {
+  // handle custom user feedback (command complete messages) for transaction statements
+  auto sql_statement = sql_pipeline.get_parsed_sql_statements().back();
+  auto& statements = sql_statement->getStatements();
+
+  if (statements.front()->isType(hsql::StatementType::kStmtTransaction)) {
+    auto* transaction_statement = reinterpret_cast<hsql::TransactionStatement*>(statements.front());
```
mrks: Why do you need a reinterpret_cast here?

src/lib/sql/sql_pipeline.cpp
```diff
@@ -219,22 +227,37 @@ std::pair<SQLPipelineStatus, const std::vector<std::shared_ptr<const Table>>&> S
...
 
+    if (pipeline_statement->warning_message())
+      _warning_messages.emplace_back(pipeline_statement->warning_message().value());
+
     DebugAssert(statement_status == SQLPipelineStatus::Success, "Unexpected pipeline status");
 
     _result_tables.emplace_back(table);
+
+    switch (pipeline_statement->transaction_context()->phase()) {
+      // Auto-commit statements should always be committed or rolled back at this point
```
mrks: Is this checked somewhere? Assertions are better than comments.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -145,7 +155,7 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
     return _physical_plan;
   }
 
-  // If we need a transaction context but haven't passed one in, this is the latest point where we can create it
+  // If we need a transaction context but haven't passed one in, this is the last point where we can create it
```
mrks: :)

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -191,30 +201,67 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
...
+  } else {
+    auto operator_tasks = OperatorTask::make_tasks_from_operator(get_physical_plan(), _cleanup_temporaries);
+    _tasks = std::vector<std::shared_ptr<AbstractTask>>(operator_tasks.cbegin(), operator_tasks.cend());
+  }
+
   return _tasks;
 }
 
+std::vector<std::shared_ptr<AbstractTask>> SQLPipelineStatement::_get_transaction_tasks() {
+  auto sql_statement = get_parsed_sql_statement();
```
mrks: const auto&?

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -191,30 +201,67 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
...
+  const std::vector<hsql::SQLStatement*>& statements = sql_statement->getStatements();
+  auto* transaction_statement = reinterpret_cast<hsql::TransactionStatement*>(statements.front());
```
mrks: Why do you need a reinterpret_cast?

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -191,30 +201,67 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
...
+        if (!_transaction_context || !_transaction_context->is_auto_commit())
+          this->set_warning_message(std::string("WARNING: Cannot begin transaction inside an active transaction."));
+        else
+          _transaction_context = Hyrise::get().transaction_manager.new_transaction_context(false);
+      })};
+    case hsql::kCommitTransaction:
+      return {std::make_shared<JobTask>([this] {
+        if (!_transaction_context || _transaction_context->is_auto_commit())
+          this->set_warning_message(std::string("WARNING: Cannot commit since there is no active transaction."));
+        else
```
mrks: Multi-line blocks should have curly braces (more than once in this method).

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -191,30 +201,67 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
...
 
+std::vector<std::shared_ptr<AbstractTask>> SQLPipelineStatement::_get_transaction_tasks() {
+  auto sql_statement = get_parsed_sql_statement();
+  const std::vector<hsql::SQLStatement*>& statements = sql_statement->getStatements();
+  auto* transaction_statement = reinterpret_cast<hsql::TransactionStatement*>(statements.front());
+  switch (transaction_statement->command) {
+    case hsql::kBeginTransaction:
+      return {std::make_shared<JobTask>([this] {
+        if (!_transaction_context || !_transaction_context->is_auto_commit())
+          this->set_warning_message(std::string("WARNING: Cannot begin transaction inside an active transaction."));
```
mrks: This error message does not seem to match the check. Are these cases tested?

src/test/server/transaction_handling_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  const auto transaction_ctx = Hyrise::get().transaction_manager.new_transaction_context();
+
+  auto [execution_information, transaction_context] =
+      QueryHandler::execute_pipeline(query, SendExecutionInfo::Yes, transaction_ctx);
+
+  // rollback transaction statement is executed successfully
+  // in this case the second insert into the table gets rolled back
+  EXPECT_TRUE(execution_information.error_message.empty());
+  EXPECT_EQ(execution_information.result_table->column_count(), 1);
```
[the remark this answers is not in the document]
mrks: You mean a check whether the row in the user table with the id 2 is not visible to other transactions, right?

This would be more interesting for a DELETE, where we should make sure that the row is "unlocked".

mrks: Why are you checking the column_count?

src/test/server/transaction_handling_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+TEST_F(TransactionHandlingTest, TransactionContextTest) {
```
mrks: What are you testing here? "TransactionContextTest" does not help too much.

src/test/server/transaction_handling_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+      QueryHandler::execute_pipeline(query, SendExecutionInfo::Yes, transaction_ctx);
+  execution_information = execution_info_transaction_context_pair.first;
+
+  // now that the user rolled back,
+  // the transaction context is in the successful state of having been rolled back on purpose
+  EXPECT_EQ(transaction_ctx->phase(), TransactionPhase::ExplicitlyRolledBack);
+  EXPECT_TRUE(execution_information.error_message.empty());
+
+  // internally the transaction context returned by the pipeline is has been set to nullptr,
+  // in order to force creating a new one, again in "auto-commit" mode
```
mrks: This sentence got stuck in my parser. Could you rephrase it?

src/lib/concurrency/transaction_context.cpp
```diff
@@ -10,11 +10,13 @@
 
 namespace opossum {
 
-TransactionContext::TransactionContext(const TransactionID transaction_id, const CommitID snapshot_commit_id)
+TransactionContext::TransactionContext(const TransactionID transaction_id, const CommitID snapshot_commit_id,
```
mrks: (Random line)
Have you tried adding a couple of SQL strings with begin/commit/rollback to sqlite_testrunner_queries.sql? While no replacement for unit tests, that should make it easy to further improve the test coverage.

src/lib/concurrency/transaction_context.cpp
```diff
@@ -28,16 +30,17 @@ TransactionContext::~TransactionContext() {
                   }
                 }
 
-                const auto is_rolled_back = _phase == TransactionPhase::RolledBack;
-                return (!an_operator_failed || is_rolled_back);
+                const auto is_rolled_back = _phase == TransactionPhase::ErrorRolledBack;
```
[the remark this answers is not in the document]
Till-Later: If one of the operator has failed, we expect the TransactionContext to Transition to the ErrorRolledBack-state. ExplicitlyRolledBack-state is only for valid transactions. However, I changed the variable name to `is_error_rolled_back` 👍

src/lib/concurrency/transaction_context.hpp
```diff
@@ -53,7 +54,7 @@ class TransactionContext : public std::enable_shared_from_this<TransactionContex
   friend class TransactionManager;
 
  public:
-  TransactionContext(TransactionID transaction_id, CommitID snapshot_commit_id);
+  TransactionContext(TransactionID transaction_id, CommitID snapshot_commit_id, bool is_auto_commit = true);
```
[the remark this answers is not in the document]
Till-Later: I don't think so. One argument could be: Auto-commit Transactions are created without a specific user command. Now these transactions don't explicitly need to state their implicity in the constructor, since auto-commit is the standard behavior.

src/lib/operators/abstract_operator.hpp
```diff
@@ -18,6 +18,11 @@ class TransactionContext;
 enum class OperatorType {
   Aggregate,
   Alias,
+  CreateTable,
+  CreatePreparedPlan,
+  CreateView,
+  DropTable,
+  DropView,
```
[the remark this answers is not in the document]
Till-Later: I guess formatter doing formatting things...?

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -124,7 +124,7 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_by_node_type(
     case LQPNodeType::Validate:           return _translate_validate_node(node);
     case LQPNodeType::Union:              return _translate_union_node(node);
 
-      // Maintenance operators
+    // Maintenance operators
```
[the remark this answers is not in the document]
Till-Later: See below

src/lib/sql/sql_pipeline_statement.hpp
```diff
@@ -23,11 +27,11 @@ struct SQLPipelineStatementMetrics {
 };
 
 enum class SQLPipelineStatus {
-  NotExecuted,  // The pipeline or the pipeline statement has been not been executed yet.
+  NotExecuted,  // The pipeline or the pipeline statement has not been executed yet.
   Success,      // The pipeline or the pipeline statement has been executed successfully. If use_mvcc is set but no
                 //     transaction_context was supplied, the statement has been auto-committed. If a context was
                 //     supplied, that context continues to be active (i.e., is not yet committed).
-  RolledBack    // The pipeline or the pipeline statement caused a transaction conflict and has been rolled back.
+  Failure       // The pipeline or the pipeline statement caused a transaction conflict and has been rolled back.
```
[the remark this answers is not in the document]
Till-Later: Explicitly rolling back the transaction can happen thorugh the ROLLBACK; command by the client and is considered as correct behavior. It can be done, even if there is no transaction conflict present. Otherwise we would return the following error message to the client ` "Transaction conflict, transaction was rolled back. Failed statement: " + failed_statement}`.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -191,30 +201,67 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
...
+    _tasks = _get_transaction_tasks();
+  } else {
+    auto operator_tasks = OperatorTask::make_tasks_from_operator(get_physical_plan(), _cleanup_temporaries);
+    _tasks = std::vector<std::shared_ptr<AbstractTask>>(operator_tasks.cbegin(), operator_tasks.cend());
+  }
+
   return _tasks;
 }
 
+std::vector<std::shared_ptr<AbstractTask>> SQLPipelineStatement::_get_transaction_tasks() {
```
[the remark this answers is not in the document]
Till-Later: I just tried to change it. The return value is stored into `std::vector<std::shared_ptr<AbstractTask>>` by the calling function `SQLPipelineStatement::get_tasks()`. This means, the return value would have to get implicitly converted from `std::vector<std::shared_ptr<JobTask>>` to `std::vector<std::shared_ptr<AbstractTask>>`. However, it turns out, that vectors cannot do this kind of conversion implicitly. I guess it would be fine, if the returned Task weren't encapsulated inside a vector.

src/lib/server/result_serializer.hpp
```diff
@@ -7,6 +7,8 @@
 
 namespace opossum {
 
+struct ExecutionInformation;
```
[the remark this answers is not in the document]
Till-Later: Yes, I think an include would to the job as well. We could change it, but I will leave it, since (at least I think so) Forward Declarations instead of includes in Header Files can reduce compile time.

src/lib/server/postgres_protocol_handler.hpp
```diff
@@ -64,6 +65,9 @@ class PostgresProtocolHandler {
   // Send error message to client if there is an error during parsing or execution
   void send_error_message(const ErrorMessage& error_message);
 
+  // Send warning message to client if there is an error during parsing or execution
+  void send_warning_message(const WarningMessage& warning_message);
```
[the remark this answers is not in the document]
Till-Later: It was supposed to be consistent the the ErrorMessage tyoe. However I changed it to std::string, since this has been a discussion among our group before.

src/lib/sql/sql_pipeline_statement.cpp
Till-Later: Throwing an exception here in the regular way would cause the database to crash. This lambda is executed by the Hyrise scheduler, which has no mechanism to catch exceptions. Instead we behave in accordance with the PostgreSQL-protocol and send a special warning message packet.

I don't see a way to generate this message at a later point without introducing some additional enum class or some other kind of mapping. However this would not be consistent with regular error messags, where strings are generated at the code position where the error is thrown.

mrks: > Throwing an exception here in the regular way would cause the database to crash. This lambda is executed by the Hyrise scheduler, which has no mechanism to catch exceptions.

Hmm, I see what you mean. If we had an AssertInput outside of the lambda, this should work, right? Can't we track `explicit_transaction_active` (naming tbd) outside of the lambda? We just need to know if there is an active transaction when we enter `get_tasks` and modify the flag accordingly when we see kCommit... whatever.

> However this would not be consistent with regular error messags, where strings are generated at the code position where the error is thrown.

Could you give me a pointer to the existing code? This doesn't sound like something we should do.

Also, one more thing: The hsql namespace should not make it outside of the SQLTranslator. Could you please translate kCommitTransaction etc. into a Hyrise-internal symbol? That improves the encapsulation.

src/lib/server/result_serializer.hpp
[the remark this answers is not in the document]
mrks: Yepp. Whenever* possible, forward declarations are better than including the related header.

*) I feel like I need to add some relativization. There are probably cases where using the include helps with optimizations or something else. As a rule of thumb, forward declarations are better.

src/test/server/transaction_handling_test.cpp
odanivan: I changed the name of the test and added some comments.

src/test/server/transaction_handling_test.cpp
odanivan: Ups, should have been row_count() of course.

src/test/server/transaction_handling_test.cpp
odanivan: Yes. I hope the change I made fixes it.

src/lib/concurrency/transaction_context.cpp
odanivan: I added some transaction statements to the sqlite test runner now and the tests pass. Let us know if you want us to add some more queries.

src/lib/concurrency/transaction_manager.hpp
Till-Later: One argument could be: Auto-commit Transactions are implicitly created without a specific user command. Now these transactions don't explicitly need to state their implicity during creation, since auto-commit is the standard behavior. 

Additionally Auto-Commit is far more common in the codebase.

src/lib/sql/sql_pipeline_statement.cpp
youri-k: I tried to change the warning message into a `FailInput`, but there were a couple of things that didn't work. We couldn't send WarningMessages to the `psql` Client anymore, but that was easily fixable by adding a new `WarnInput` method. However, throwing in the sql pipeline statement breaks the pipeline and this isn't expected for a warning. Additionally, I tried to pull the check out of the lambda, but than we have nothing to return in the method and I thought adding another bool to the lambda wouldn't make it any nicer than the current solution.

If you got any other ideas, I'll gladly try to implement them.

src/lib/concurrency/transaction_context.cpp
odanivan: Yes, we have now added some SQL queries containing transaction statements to the sqlite test runner and the tests pass.

src/test/server/transaction_handling_test.cpp
odanivan: Ups. Yes, that should have been row_count.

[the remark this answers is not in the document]
odanivan: We now added another unit test that checks for visibility / uncommitted transaction side effects. Thanks for the input!

README.md
```diff
@@ -157,4 +157,6 @@ Contact: firstname.lastname@hpi.de
 -   Lukas     Wenzel
 -   Fabian    Wiebe
 -   Tim       Zimmermann
-
+-   Ivan      Illic
```
mrks: Please make sure this is sorted.

src/test/server/transaction_handling_test.cpp
```diff
@@ -78,13 +82,72 @@ TEST_F(TransactionHandlingTest, TransactionContextTest) {
...
+  // that's why the change is not visible yet
+  EXPECT_TRUE(execution_information_2.error_message.empty());
+  EXPECT_EQ(execution_information_2.result_table->row_count(), 1);
+
+  // user A finally commits the transaction
+  query = "COMMIT;";
+
+  execution_info_transaction_context_pair =
+      QueryHandler::execute_pipeline(query, SendExecutionInfo::Yes, transaction_ctx);
+  execution_information = execution_info_transaction_context_pair.first;
```
mrks: Could also use `const auto [execution_information, _] = ...` (structured binding)

src/lib/sql/sql_translator.cpp
```diff
@@ -194,6 +194,8 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_statement(const hsql:
     case hsql::kStmtExport:
       return _translate_export(static_cast<const hsql::ExportStatement&>(statement));
     case hsql::kStmtTransaction:
+      // The transaction statements are handled directly in the SQLPipelineStatement,
+      //  but the translation is still called, so we need some kind of handling here.
```
mrks: ... so we return a dummy node here?

src/lib/sql/sql_pipeline_statement.hpp
```diff
@@ -134,8 +133,10 @@ class SQLPipelineStatement : public Noncopyable {
   // Delete temporary tables
   const CleanupTemporaries _cleanup_temporaries;
 
-  std::optional<std::string> _warning_message;
-  bool _is_transaction_statement;
+  std::optional<std::string> _warning_message = std::nullopt;
+
+  // Might be the Statement's own transaction context, or the one shared by all Statements in a Pipeline
```
mrks: Own transaction context **(if in auto-commit mode)**?

src/lib/server/query_handler.cpp
```diff
@@ -96,7 +96,7 @@ void QueryHandler::handle_transaction_statement_message(ExecutionInformation& ex
   auto& statements = sql_statement->getStatements();
 
   if (statements.front()->isType(hsql::StatementType::kStmtTransaction)) {
-    auto* transaction_statement = reinterpret_cast<hsql::TransactionStatement*>(statements.front());
+    auto* transaction_statement = dynamic_cast<hsql::TransactionStatement*>(statements.front());
```
mrks: Since you are not testing this for `nullptr`, `(const) auto& ... = dynamic_cast<...&>(*statements.front())` would be the canonical way.

src/lib/concurrency/transaction_context.hpp
```diff
@@ -81,14 +90,17 @@ class TransactionContext : public std::enable_shared_from_this<TransactionContex
   TransactionPhase phase() const;
 
   /**
-   * Returns true if transaction has been aborted.
+   * Returns true if transaction has been aborted due to an error.
```
mrks: Same as before: We don't really have errors. If this is only for RollBackAfterConflict, maybe rename the method? Above you have used "failed", maybe that helps?

src/lib/concurrency/transaction_context.hpp
```diff
@@ -15,46 +15,55 @@ class CommitContext;
...
+  RolledBackByUser,             // Transaction has been rolled back due to ROLLBACK;-statement. (Considered a success)
+  Committing,                   // Commit ID has been assigned. Operators may commit records.
+  Committed,                    // Transaction has been committed.
 };
 
 std::ostream& operator<<(std::ostream& stream, const TransactionPhase& phase);
 
 /**
- * @brief Representation of a transaction
+ * @brief Representation of a transactionsql
```
mrks: ?

src/lib/concurrency/transaction_context.hpp
```diff
@@ -15,46 +15,55 @@ class CommitContext;
...
  */
 enum class TransactionPhase {
-  Active,                // Transaction has just been created. Operators may be executed.
-  Aborted,               // One of the operators failed. Transaction needs to be rolled back.
-  ErrorRolledBack,       // Transaction has been rolled back due to error.
-  ExplicitlyRolledBack,  // Transaction has been rolled back due to rollback transaction statement.
-  Committing,            // Commit ID has been assigned. Operators may commit records.
-  Committed,             // Transaction has been committed.
+  Active,                       // Transaction has just been created. Operators may be executed.
+  Aborted,                      // One of the operators failed. Transaction needs to be rolled back.
```
mrks: Old text, but we should change the terminology away from "failure" here, too. Maybe "ran into a transaction conflict"? Next lines, too.

src/lib/server/query_handler.cpp
mrks: Making this its own comment so that it does not get lost in the other discussion:

> Also, one more thing: The hsql namespace should not make it outside of the SQLTranslator. Could you please translate kCommitTransaction etc. into a Hyrise-internal symbol? That improves the encapsulation.

src/lib/sql/sql_pipeline_statement.cpp
mrks: Maybe I have a wrong understanding of what a "warning" is in psql. My idea was to reject the execution of the current SQL string. Is the Postgres standard to just print an info message saying "Whoopsie, this didn't make sense, I just ignored your transaction statement and did potentially non-atomic auto-commit modifications to the database"? If yes, I understand what you mean. Would it be possible to deviate from that scheme and make it a hard error? I can't see any case where we would want a soft warning (that will be ignored) instead of a hard fail.

mrks:
I think we are basically done once the warning/error thing is resolved. Let's hope that I understood you right on that.

README.md
klauck: Done.

src/lib/sql/sql_pipeline_statement.hpp
```diff
@@ -123,6 +133,9 @@ class SQLPipelineStatement : public Noncopyable {
 
   // Delete temporary tables
   const CleanupTemporaries _cleanup_temporaries;
+
+  std::optional<std::string> _warning_message;
+  bool _is_transaction_statement;
```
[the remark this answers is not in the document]
klauck: Was moved to a method.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -38,12 +39,13 @@ SQLPipelineStatement::SQLPipelineStatement(const std::string& sql, std::shared_p
...
       _use_mvcc(use_mvcc),
-      _auto_commit(_use_mvcc == UseMvcc::Yes && !transaction_context),
       _transaction_context(transaction_context),
       _optimizer(optimizer),
       _parsed_sql_statement(std::move(parsed_sql)),
       _metrics(std::make_shared<SQLPipelineStatementMetrics>()),
-      _cleanup_temporaries(cleanup_temporaries) {
+      _cleanup_temporaries(cleanup_temporaries),
+      _warning_message(std::nullopt),
+      _is_transaction_statement(false) {
```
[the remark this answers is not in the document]
klauck: The member _is_transaction_statement was deletes.

src/lib/sql/sql_pipeline_statement.cpp
klauck: Link to PostgreSQL error messages https://www.postgresql.org/docs/12/plpgsql-errors-and-messages.html#PLPGSQL-STATEMENTS-RAISE
What do you mean by "hard error"?
Option 1: Server crash: I think we should not do this, but be at least be a bit robust. Note, invalid sql does neither crash the server, but reports an error to the user.
Option 2: Error message to user: The error level does not change the fact that we must store the statement status.

src/lib/concurrency/transaction_manager.hpp
klauck: 1) Does the argument convince you?
2) Or should we remove the default for readability?

I prefer the second.

mrks: Agreed. I don't know of a general "default argument rule". My personal approach is to use defaults only when the default is painfully obvious or when the usual caller would simply not care about it. Here, neither applies.

src/lib/sql/sql_pipeline_statement.cpp
mrks: I'm not sure if I completely understand the issue here. If I submit an invalid SQL query through psql, I get this:

```
=> SELECT foobar;
Invalid input error: Couldn't resolve identifier 'foobar' or it is ambiguous
```

This is an AssertInput and is caught by `Session::run`. Why doesn't this work here?

mrks: > My idea was to reject the execution of the current SQL string. Is the Postgres standard to just print an info message saying "Whoopsie, this didn't make sense, I just ignored your transaction statement and did potentially non-atomic auto-commit modifications to the database"?

So the latter is true? If you ask me, that doesn't make any sense and I can't see why a user would want that. Especially if queries are not sent manually but as part of a benchmark or an application, this seems like a recipe for desaster. In conformance with our "fail often, fail loud" principle, I would expect the database to reject such a broken query.

klauck: > My idea was to reject the execution of the current SQL string. Is the Postgres standard to just print an info message saying "Whoopsie, this didn't make sense, I just ignored your transaction statement and did potentially non-atomic auto-commit modifications to the database"?
The current SQL statement is rejected for errors AND warnings. (This is what Postgres and we do currently.)
However, in Posgres, errors abort the current transaction. Warnings preserve the Transaction status. In both cases, the next statement is executed normally.
From the Postgres documentation "the frontend must then take responsibility for figuring out what happens if the earlier command fails and already-issued later commands succeed."

We currently abort the complete pipeline as we catch the error only in Session::run. I think it is ok for a research database. However, users often cannot directly control which statements end up in one pipeline.
The longer I think about it, I prefer using AssertInput also for warnings, although it is not the default behavior, which we also not use/guanrantee for errors.
Advantages:
- Simpler code
- The user probably did something unintentional anyway.

If you agree, I switch from warnings to errors.

klauck: > Also, one more thing: The hsql namespace should not make it outside of the SQLTranslator. Could you please translate kCommitTransaction etc. into a Hyrise-internal symbol? That improves the encapsulation.

Note, in the current master: SQLPipeline uses hsql-things, too:
`switch (statement->type()) {
      // Check if statement alters the structure of the database in a way that following statements might depend upon.
      case hsql::StatementType::kStmtImport:
      case hsql::StatementType::kStmtCreate:
      case hsql::StatementType::kStmtDrop:
      case hsql::StatementType::kStmtAlter:
      case hsql::StatementType::kStmtRename:`
If I should change this, I would be happy about a proposal in which way to adaptd the code. Currently we use a shortcut before the statement translation happens.

mrks: > If you agree, I switch from warnings to errors.

Agreed. To add to the advantages: This seems like the easier approach and, if we ever come to the point where it is not sufficient any more, we can still adapt it. The technical debt seems smaller than adding error handling for a case that should not exist.

> Note, in the current master: SQLPipeline uses hsql-things

Hmpf, ok. Moving this into the SQLTranslator just for the sake of it being somewhere else doesn't help anyone. Just ignore that point.

src/lib/sql/sql_pipeline_statement.hpp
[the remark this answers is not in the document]
klauck: Done.

src/lib/sql/sql_pipeline_statement.cpp
klauck: The code is correct but complicated, because SQLPipelineStatement::get_physical_plan() is called before get_transaction_tasks() (to _precheck_ddl_operators) und requires a TransactionContext.
As a result, when get_transaction_tasks() is called, there may exist an automatically created auto-commit-transaction-context.

There are two possibilities to make the code easier:
1) use an extra switch in get_result_table to check for transaction statements before _precheck_ddl_operators is called
2) move _precheck_ddl_operators inside get_tasks(). I currently do not understand why it is outside.

src/lib/sql/sql_translator.cpp
klauck: Is this a proposal to improve the comment?

src/lib/sql/sql_pipeline_statement.hpp
klauck: Done.

src/lib/server/query_handler.cpp
klauck: Obsolete (as hsql is already used in the pipeline).

src/lib/sql/sql_pipeline_statement.cpp
klauck: Done.

src/lib/sql/sql_pipeline_statement.cpp
klauck: Done (2.).

src/lib/sql/sql_translator.cpp
klauck: Done.

src/lib/server/query_handler.cpp
klauck: Done.

src/lib/concurrency/transaction_context.hpp
klauck: @mrks "Failed" is one operator state `enum class ReadWriteOperatorState`.
Should we keep "failed operators" but use "conflicting transactions"?

src/lib/concurrency/transaction_context.hpp
klauck: Undone.

mrks:
@klauck Could you merge the master please?

klauck:
> @klauck Could you merge the master please?

Yes (, tomorrow).

klauck:
@mrks master is merged now
Two open questions from my side:
(1) Should we change `enum class RollBackReason : bool { RollBackByUser, RollBackAfterConflict };` to `enum class RollBackReason : bool { ByUser, AfterConflict };` ?
(2) A new commit on the hyrise/master added a default parameter for creating transaction contexts. We discussed, whether we should remove it. Any final opinion?
One observation on (2): For many tests, the parameter can be arbitrarily set, because the auto commit logic (implemented in the SQLPipeline) is bypassed and transactions have to be committed explicitly anyway.

mrks:
>  (1) Should we change enum class RollBackReason : bool { RollBackByUser, RollBackAfterConflict }; to enum class RollBackReason : bool { ByUser, AfterConflict }; ?

Or even `RollbackReason::User` and `RollbackReason::Conflict`? (Note: rollback, not roll back)

> (2) A new commit on the hyrise/master added a default parameter for creating transaction contexts. We discussed, whether we should remove it. Any final opinion?

Yes. I didn't think too much about it back then, but the discussion that we had in here makes more sense. Same for `_mark_as_rolled_back(RollBackReason rollback_reason = RollBackReason::RollBackAfterConflict)` and it's colleagues. The conflict rollback is not really more "default" than the user one.

klauck:
hyrise/master kann ich morgen noch mergen

src/lib/server/session.cpp
```diff
@@ -10,7 +10,8 @@ namespace opossum {
 Session::Session(boost::asio::io_service& io_service, const SendExecutionInfo send_execution_info)
     : _socket(std::make_shared<Socket>(io_service)),
       _postgres_protocol_handler(std::make_shared<PostgresProtocolHandler<Socket>>(_socket)),
-      _send_execution_info(send_execution_info) {}
+      _send_execution_info(send_execution_info),
+      _transaction_context(nullptr) {}
```
mrks, line 14: remove

src/lib/server/session.cpp
```diff
@@ -105,7 +106,8 @@ void Session::_handle_simple_query() {
   // A simple query command invalidates unnamed portals
   _portals.erase("");
 
-  const auto execution_information = QueryHandler::execute_pipeline(query, _send_execution_info);
+  const auto [execution_information, transaction_context] =
+      QueryHandler::execute_pipeline(query, _send_execution_info, _transaction_context);
```
mrks, line 110: Use std::tie to avoid having both `transaction_context` and `_transaction_context`

src/lib/server/session.cpp
```diff
@@ -187,8 +192,9 @@ void Session::_handle_execute() {
 
   if (portal_name.empty()) _portals.erase(portal_it);
 
-  if (!_transaction) _transaction = Hyrise::get().transaction_manager.new_transaction_context();
-  physical_plan->set_transaction_context_recursively(_transaction);
+  if (!_transaction_context)
+    _transaction_context = Hyrise::get().transaction_manager.new_transaction_context(AutoCommit::No);
```
mrks, line 196: braces

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -190,65 +188,99 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
...
+    auto operator_tasks = OperatorTask::make_tasks_from_operator(get_physical_plan());
+    _tasks = std::vector<std::shared_ptr<AbstractTask>>(operator_tasks.cbegin(), operator_tasks.cend());
+  }
   return _tasks;
 }
 
+std::vector<std::shared_ptr<AbstractTask>> SQLPipelineStatement::_get_transaction_tasks() {
+  const auto& sql_statement = get_parsed_sql_statement();
+  const std::vector<hsql::SQLStatement*>& statements = sql_statement->getStatements();
+  auto* transaction_statement = dynamic_cast<hsql::TransactionStatement*>(statements.front());
```
mrks, line 209: No need for raw ptr

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -190,65 +188,99 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
...
-  if (was_rolled_back()) {
-    return {SQLPipelineStatus::RolledBack, _result_table};
+  if (has_failed()) {
+    return {SQLPipelineStatus::Failure, _result_table};
   }
 
-  if (_auto_commit) {
+  if (_use_mvcc == UseMvcc::Yes && _transaction_context->is_auto_commit()) {
     _transaction_context->commit();
   }
```
mrks, line 267: Check if responsibilities are clearly documented.

src/lib/sql/sql_pipeline.hpp
```diff
@@ -99,7 +98,7 @@ class SQLPipeline : public Noncopyable {
 
   std::vector<std::shared_ptr<SQLPipelineStatement>> _sql_pipeline_statements;
 
-  const std::shared_ptr<TransactionContext> _transaction_context;
+  std::shared_ptr<TransactionContext> _transaction_context;
```
mrks, line 101: doc

src/lib/sql/sql_pipeline.cpp
```diff
@@ -217,22 +225,38 @@ std::pair<SQLPipelineStatus, const std::vector<std::shared_ptr<const Table>>&> S
...
 
-      return {SQLPipelineStatus::RolledBack, _result_tables};
+      return {SQLPipelineStatus::Failure, _result_tables};
     }
 
     DebugAssert(statement_status == SQLPipelineStatus::Success, "Unexpected pipeline status");
 
     _result_tables.emplace_back(table);
+
+    const auto& previous_transaction_context = pipeline_statement->transaction_context();
```
mrks, line 247: new_transaction_context?

src/lib/sql/sql_pipeline.cpp
```diff
@@ -217,22 +225,38 @@ std::pair<SQLPipelineStatus, const std::vector<std::shared_ptr<const Table>>&> S
...
-    if (statement_status == SQLPipelineStatus::RolledBack) {
+    if (statement_status == SQLPipelineStatus::Failure) {
       _failed_pipeline_statement = pipeline_statement;
 
-      if (_transaction_context) {
+      if (_transaction_context && !_transaction_context->is_auto_commit()) {
         // The pipeline was executed using a transaction context (i.e., no auto-commit after each statement).
         // Previously returned results are invalid.
         _result_tables.clear();
+        _transaction_context = nullptr;
```
mrks, line 237: Why is this change necessary?

src/lib/sql/sql_pipeline.cpp
```diff
@@ -217,22 +225,38 @@ std::pair<SQLPipelineStatus, const std::vector<std::shared_ptr<const Table>>&> S
...
     }
 
     DebugAssert(statement_status == SQLPipelineStatus::Success, "Unexpected pipeline status");
 
     _result_tables.emplace_back(table);
+
+    const auto& previous_transaction_context = pipeline_statement->transaction_context();
+
+    if (previous_transaction_context->is_auto_commit()) {
+      DebugAssert(previous_transaction_context->phase() == TransactionPhase::Committed ||
```
mrks, line 250: Assert

src/lib/sql/sql_pipeline.cpp
```diff
@@ -217,22 +225,38 @@ std::pair<SQLPipelineStatus, const std::vector<std::shared_ptr<const Table>>&> S
...
+                      previous_transaction_context->phase() == TransactionPhase::RolledBackByUser,
```
mrks, line 251: Can an auto commit transaction be in TransactionPhase::RolledBackByUser? If not, should _transition guard against this?

src/lib/sql/sql_pipeline.cpp
```diff
@@ -217,22 +225,38 @@ std::pair<SQLPipelineStatus, const std::vector<std::shared_ptr<const Table>>&> S
...
+                  "Auto-commit statements should always be committed or rolled back at this point");
+      _transaction_context = nullptr;
+    } else if (previous_transaction_context->phase() == TransactionPhase::Committed ||
+               previous_transaction_context->phase() == TransactionPhase::RolledBackByUser) {
+      _transaction_context = nullptr;
+    } else {
+      _transaction_context = previous_transaction_context;
```
mrks, line 258: doc

src/test/CMakeLists.txt
```diff
@@ -143,6 +143,7 @@ set(
     scheduler/scheduler_test.cpp
     server/mock_socket.hpp
     server/postgres_protocol_handler_test.cpp
+    server/transaction_handling_test.cpp
```
mrks, line 146: sort

src/test/operators/change_meta_table_test.cpp
```diff
@@ -92,7 +92,7 @@ TEST_F(ChangeMetaTableTest, OnlyAllowsAutoCommit) {
 
   EXPECT_THROW(change_meta_table->execute(), std::exception);
 
-  transaction_context->rollback();
+  transaction_context->rollback(RollbackReason::Conflict);
```
mrks, line 95: Change to User

src/test/server/transaction_handling_test.cpp
```diff
@@ -0,0 +1,151 @@
...
+  std::string query = "CREATE TABLE users (id INT); INSERT INTO users(id) VALUES (1);";
+
+  auto transaction_ctx = Hyrise::get().transaction_manager.new_transaction_context(AutoCommit::Yes);
+
+  auto execution_info_transaction_context_pair =
+      QueryHandler::execute_pipeline(query, SendExecutionInfo::Yes, transaction_ctx);
+
+  auto execution_information = execution_info_transaction_context_pair.first;
+
+  // normally, when user has not begun a transaction yet, the transaction context is in "auto-commit" mode
```
mrks, line 55: misleading

src/test/server/transaction_handling_test.cpp
```diff
@@ -0,0 +1,151 @@
...
+  EXPECT_TRUE(execution_information.error_message.empty());
+  EXPECT_EQ(transaction_ctx->is_auto_commit(), true);
+  EXPECT_EQ(transaction_ctx->phase(), TransactionPhase::Committed);
+
+  transaction_ctx = execution_info_transaction_context_pair.second;
```
mrks, line 60: structured binding (more than once in this file)

src/test/server/transaction_handling_test.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+  query = "BEGIN; INSERT INTO users(id) VALUES (2);";
```
mrks, line 62: scope

src/test/server/transaction_handling_test.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+  auto execution_info_transaction_context_pair =
+      QueryHandler::execute_pipeline(query, SendExecutionInfo::Yes, transaction_ctx);
+
+  auto execution_information = execution_info_transaction_context_pair.first;
+
+  transaction_ctx = execution_info_transaction_context_pair.second;
+
+  // user A wants to delete a row from the table and starts transaction
+  query = "BEGIN; DELETE FROM users WHERE id = 1;";
```
mrks, line 107: scope

src/test/server/transaction_handling_test.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+namespace opossum {
+
+class TransactionHandlingTest : public BaseTest {
+ protected:
+  void SetUp() override {}
+};
+
+TEST_F(TransactionHandlingTest, CreateTableWithinTransaction) {
+  const std::string query = "BEGIN; CREATE TABLE users (id INT); INSERT INTO users(id) VALUES (1); COMMIT;";
```
mrks, line 13: Do we have any failure case (e.g., BEGIN BEGIN) tests?

mrks:
Please review 05168e6e. If I didn't add any new mistakes, that should basically be it.

Open issues:
- [ ] Is there anything that keeps us from changing the console's manual transaction handling to this approach as part of a separate issue?
- [ ] Using the `psql` client, statements are sent one by one. As a result, `BEGIN; INSERT INTO t (a) VALUES (1); BEGIN; INSERT INTO t (a) VALUES (2); COMMIT;` shows an error message, but still inserts 2. Is this part of the Postgres standard? If yes, we should add a warning (e.g., "Depending on your client, following statements might have still been sent and executed"). If the wire format generally allows multiple statements in one packet, this should be tested.
- [ ] Run TPC-H with a low scale factor verification

klauck:
> Please review [05168e6](https://github.com/hyrise/hyrise/commit/05168e6e9fe58d563a1606b93d80f0510823661d). If I didn't add any new mistakes, that should basically be it.
> 
> Open issues:
> 
>     * [ ]  Is there anything that keeps us from changing the console's manual transaction handling to this approach as part of a separate issue?
I do not think so.
I will look into it after this PR is merged.

> 
>     * [ ]  Using the `psql` client, statements are sent one by one. As a result, `BEGIN; INSERT INTO t (a) VALUES (1); BEGIN; INSERT INTO t (a) VALUES (2); COMMIT;` shows an error message, but still inserts 2. Is this part of the Postgres standard? If yes, we should add a warning (e.g., "Depending on your client, following statements might have still been sent and executed"). If the wire format generally allows multiple statements in one packet, this should be tested.
> 
>     * [ ]  Run TPC-H with a low scale factor verification

I look into the rest next week after my lecture on monday

mrks:
No notable performance regressions for SF 0.01:

```diff
 +----------------+--------------------+--------+--------------------+--------+------------+---------+
 | Benchmark      | prev. iter/s       | runs   | new iter/s         | runs   | change [%] | p-value |
 +----------------+--------------------+--------+--------------------+--------+------------+---------+
 | TPC-H 01       | 137.33602905273438 | 8241   | 138.9593048095703  | 8338   | +1%        |  0.0000 |
 | TPC-H 02       | 1773.5830078125    | 106415 | 1791.017578125     | 107462 | +1%        |  0.0357 |
 | TPC-H 03       | 687.5708618164062  | 41255  | 667.2001342773438  | 40033  | -3%        |  0.0000 |
 | TPC-H 04       | 990.59716796875    | 59436  | 968.95654296875    | 58138  | -2%        |  0.0000 |
 | TPC-H 05       | 598.6329345703125  | 35918  | 576.728759765625   | 34604  | -4%        |  0.0000 |
 | TPC-H 06       | 9406.9560546875    | 564418 | 9479.8017578125    | 568789 | +1%        |  0.0000 |
-| TPC-H 07       | 402.83795166015625 | 24171  | 381.6468811035156  | 22899  | -5%        |  0.0002 |
 | TPC-H 08       | 62.33194351196289  | 3740   | 63.546207427978516 | 3813   | +2%        |  0.0103 |
 | TPC-H 09       | 300.8977966308594  | 18054  | 296.0789794921875  | 17765  | -2%        |  0.0000 |
 | TPC-H 10       | 464.8163146972656  | 27889  | 459.7134704589844  | 27583  | -1%        |  0.0000 |
 | TPC-H 11       | 3002.0478515625    | 180123 | 2924.9990234375    | 175500 | -3%        |  0.0000 |
 | TPC-H 12       | 1294.5419921875    | 77673  | 1306.384033203125  | 78384  | +1%        |  0.0000 |
 | TPC-H 13       | 455.8348693847656  | 27351  | 452.7369079589844  | 27165  | -1%        |  0.0000 |
 | TPC-H 14       | 2897.21484375      | 173833 | 2897.6240234375    | 173858 | +0%        |  0.3600 |
 | TPC-H 15       | 787.6502075195312  | 47260  | 804.054931640625   | 48244  | +2%        |  0.0000 |
 | TPC-H 16       | 400.878173828125   | 24053  | 402.59942626953125 | 24156  | +0%        |  0.0000 |
-| TPC-H 17       | 1360.6259765625    | 81638  | 1231.83984375      | 73911  | -9%        |  0.0000 |
 | TPC-H 18       | 365.6825866699219  | 21941  | 374.9067077636719  | 22495  | +3%        |  0.0000 |
 | TPC-H 19       | 98.49745178222656  | 5910   | 99.76142120361328  | 5986   | +1%        |  0.0000 |
 | TPC-H 20       | 320.60443115234375 | 19237  | 322.5276794433594  | 19352  | +1%        |  0.0005 |
 | TPC-H 21       | 243.77330017089844 | 14627  | 238.33042907714844 | 14300  | -2%        |  0.0000 |
 | TPC-H 22       | 593.14013671875    | 35589  | 608.4427490234375  | 36507  | +3%        |  0.0000 |
 | geometric mean |                    |        |                    |        | -1%        |         |
 +----------------+--------------------+--------+--------------------+--------+------------+---------+
```

mrks:
As you've already approved the PR, could you ping me when you're done with your new review?

klauck:
>     * Using the `psql` client, statements are sent one by one. As a result, `BEGIN; INSERT INTO t (a) VALUES (1); BEGIN; INSERT INTO t (a) VALUES (2); COMMIT;` shows an error message, but still inserts 2. Is this part of the Postgres standard? If yes, we should add a warning (e.g., "Depending on your client, following statements might have still been sent and executed"). If the wire format generally allows multiple statements in one packet, this should be tested.
Short: yes, it is Postgres standard

Long:
Postgres does not have the notion of an sql pipeline.
As a result, if an earlier statement fails, (in Postgres) there is no distinction whether the next statement was part of the query string (Hyrise pipeline) or not.
It is executed in any case.
(Postgres Docs: "the frontend must then take responsibility for figuring out what happens if the earlier command fails and already-issued later commands succeed")

klauck:
>     * If the wire format generally allows multiple statements in one packet, this should be tested.
There are tests, which call `QueryHandler::execute_pipeline()` with multiple statements, e.g., in transaction_handling_test.cpp

klauck:
> As you've already approved the PR, could you ping me when you're done with your new review?
I am done.
(The console update can be moved to an issue.)

mrks:
Thanks again everyone. I think this fits in quite well.

klauck:
Thank you, too, for your support and final changes.

PR #1995, opened 2020-01-22
Segment access counters

mflueggen:
This PR provides segments with the possibility to keep track of how many times they were accessed and what access pattern was used. This is done by the caller (iterator, accessor) executing the corresponding function of the segment’s counter.

Currently the categories `other` and `dictionary_access` may only be partially implemented. `other` counts accesses which do not fit into the other (haha) categories (like `ValueSegment::append`). `dictionary_access` increases when  `DictionarySegment::_dictionary` is requested.


resources/test_data/tbl/meta_tables/meta_segments.tbl
```diff
@@ -1,9 +1,9 @@
...
-int_int_int_null|0|1|b|int|Dictionary|SimdBp128|100
-int_int_int_null|0|2|c|int|null|null|265
\ No newline at end of file
+int_int|0|0|a|int|null|null|240
+int_int|0|1|b|int|null|null|240
+int_int|1|0|a|int|null|null|236
+int_int|1|1|b|int|null|null|236
+int_int_int_null|0|0|a|int|RunLength|null|160
+int_int_int_null|0|1|b|int|Dictionary|SimdBp128|164
+int_int_int_null|0|2|c|int|null|null|329
```
Bensk1: Where do these changes come from? Is #1968 not merged yet? Could you merge the current master?

mflueggen: The tests failed after having implemented the access counters. The expected values were 64 bytes off. Exactly the size of the access counter structure.

Bensk1: Alright!

src/lib/import_export/binary/binary_parser.hpp
```diff
@@ -1,104 +1,104 @@
...
-  // Reads row_count many strings from input file. String lengths are encoded in type T.
-  static pmr_vector<pmr_string> _read_string_values(std::ifstream& file, const size_t count);
-
-  // Reads a single value of type T from the input file.
-  template <typename T>
-  static T _read_value(std::ifstream& file);
-};
-
-}  // namespace opossum
+#pragma once
```
mrks: Nothing changed here, right? If this is just CR/LF vs LF, we should solve it outside of this PR.

src/lib/storage/value_segment.cpp
```diff
@@ -150,12 +156,16 @@ template <typename T>
...
+  std::shared_ptr<BaseSegment> copied_segment;
   if (is_nullable()) {
     pmr_concurrent_vector<bool> new_null_values(*_null_values, alloc);  // NOLINT(cppcoreguidelines-slicing) (see above)
-    return std::allocate_shared<ValueSegment<T>>(alloc, std::move(new_values), std::move(new_null_values));
+    copied_segment = std::allocate_shared<ValueSegment<T>>(alloc, std::move(new_values), std::move(new_null_values));
   } else {
-    return std::allocate_shared<ValueSegment<T>>(alloc, std::move(new_values));
+    copied_segment = std::allocate_shared<ValueSegment<T>>(alloc, std::move(new_values));
   }
+  copied_segment->access_counter.set_counter_values(access_counter);
```
mrks: We should discuss what the expected behavior here is. I could find reasons for both the values to be copied or to be reset.

src/lib/storage/segment_accessor.hpp
```diff
@@ -71,7 +79,7 @@ class MultipleChunkReferenceSegmentAccessor final : public AbstractSegmentAccess
 
     const auto chunk_id = row_id.chunk_id;
 
-    // Grow the _accessors vector faster than linearly if the chunk_id is out of its current bounds
+    // Grow the _accessor's vector faster than linear if the chunk_id is out of its current bounds
```
mrks: Both look correct to me in the original version

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,168 @@
...
+    case SegmentAccessCounter::AccessPattern::Rnd:
+      _counter.iterator_random_access += positions->size();
+  }
+}
+
+// count is currently not in use.
+void SegmentAccessCounter::on_accessor_create(uint64_t count) { ++_counter.accessor_create; }
+
+// chunk_offset is currently not in use. It may be used in the future to derive the access pattern.
+void SegmentAccessCounter::on_accessor_access(uint64_t count, ChunkOffset chunk_offset) {
```
mrks: YAGNI. As we don't build interfaces consumed by anyone else, we should only add it once needed. Also, I believe that I've seen `chunk_offset == 0` being passed, so in that case, adding it later would give us wrong results.

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,168 @@
...
+template <typename T>
+SegmentAccessCounter::Counter<T>::Counter()
+    : other{0},
+      iterator_create{0},
+      iterator_seq_access{0},
+      iterator_increasing_access{0},
+      iterator_random_access{0},
+      accessor_create{0},
+      accessor_access{0},
+      dictionary_access{0} {}
```
mrks: Move to declaration?

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,168 @@
...
+  _counter.iterator_create = iterator_create;
+  _counter.iterator_seq_access = iterator_seq_access;
+  _counter.iterator_increasing_access = iterator_increasing_access;
+  _counter.iterator_random_access = iterator_random_access;
+  _counter.accessor_create = accessor_create;
+  _counter.accessor_access = accessor_access;
+  _counter.dictionary_access = dictionary_access;
+}
+
+void SegmentAccessCounter::set_counter_values(const Counter<uint64_t>& counter) {
```
mrks: How is this different from `operator=`?

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,168 @@
...
+  counter.iterator_seq_access = _counter.iterator_seq_access;
+  counter.other = _counter.other;
+  counter.dictionary_access = _counter.dictionary_access;
+  counter.accessor_access = _counter.accessor_access;
+  counter.accessor_create = _counter.accessor_create;
+  counter.iterator_create = _counter.iterator_create;
+  return counter;
+}
+
+void SegmentAccessCounter::set_counter_values(uint64_t other, uint64_t iterator_create, uint64_t iterator_seq_access,
```
mrks: What is this needed for?

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,168 @@
...
+}
+
+void SegmentAccessCounter::on_dictionary_access(uint64_t count) { _counter.dictionary_access += count; }
+
+void SegmentAccessCounter::on_other_access(uint64_t count) { _counter.other += count; }
+
+void SegmentAccessCounter::reset() { _counter.reset(); }
+
+// returns a copy of the segments counter.
+SegmentAccessCounter::Counter<uint64_t> SegmentAccessCounter::counter() const {
```
mrks: How is this different from the copy constructor?

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,168 @@
...
+// chunk_offset is currently not in use. It may be used in the future to derive the access pattern.
+void SegmentAccessCounter::on_accessor_access(uint64_t count, ChunkOffset chunk_offset) {
+  _counter.accessor_access += count;
+}
+
+void SegmentAccessCounter::on_dictionary_access(uint64_t count) { _counter.dictionary_access += count; }
+
+void SegmentAccessCounter::on_other_access(uint64_t count) { _counter.other += count; }
+
+void SegmentAccessCounter::reset() { _counter.reset(); }
```
mrks: Could this be covered by `s = SegmentAccessCounter{};`?

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include <memory>
+#include <string>
+
+#include "storage/pos_list.hpp"
+#include "types.hpp"
+
+namespace opossum {
+class Table;
+
+class SegmentAccessCounter {
```
mrks: Could you add some documentation in this file?

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+  void on_iterator_create(const std::shared_ptr<const PosList>& positions);
+
+  void on_accessor_create(uint64_t count);
+
+  void on_accessor_access(uint64_t count, ChunkOffset chunk_offset);
+
+  void on_dictionary_access(uint64_t count);
+
+  void on_other_access(uint64_t count);
```
mrks: What is an "other access"? Why don't I see a member field for it?

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include "types.hpp"
+
+namespace opossum {
+class Table;
+
+class SegmentAccessCounter {
+  friend class SegmentAccessCounterTest;
+
+ public:
+  template <typename T>
```
mrks: What different types do we have / need?

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,98 @@
...
+  class Counter {
+   public:
+    T other;
+    T iterator_create;
+    T iterator_seq_access;
+    T iterator_increasing_access;
+    T iterator_random_access;
+    T accessor_create;
+    T accessor_access;
+    T dictionary_access;
```
mrks: I wonder if having an `enum class` with the different types of counters and an array with their values would make things like `operator=` and copying easier.

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+    T sum() const;
+
+    std::string to_string() const;
+
+    inline static const std::string HEADERS =
+        "Other,IteratorCreate,IteratorSeqAccess,IteratorIncreasingAccess,"
+        "IteratorRandomAccess,AccessorCreate,AccessorAccess,"
+        "DictionaryAccess";
+  };
```
mrks: If you use the enum, you could create a string mapping and ensure that the order is correct and complete.

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,168 @@
...
+    case SegmentAccessCounter::AccessPattern::RndInc:
+    case SegmentAccessCounter::AccessPattern::RndDec:
+      _counter.iterator_increasing_access += positions->size();
+      break;
+    case SegmentAccessCounter::AccessPattern::Rnd:
+      _counter.iterator_random_access += positions->size();
+  }
+}
+
+// count is currently not in use.
```
mrks: Then don't add it. ;)

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,98 @@
...
+  // For access pattern analysis: The following enums are used to determine how an iterator iterates over its elements.
+  // This is done by analysing the first elements in a given PosList and a state machine, defined below.
+  // There are six AccessPatterns:
+  // 0 (unknown), equivalent to IteratorSeqAccess
+  // 1 (sequentially increasing), difference between two neighboring elements is 0 or 1.
+  // 2 (increasing randomly)
+  // 3 (sequentially decreasing), difference between two neighboring elements is -1 or 0.
+  // 4 (decreasing randomly)
+  // 5 (random access)
+  enum class AccessPattern { Unknown, SeqInc, RndInc, SeqDec, RndDec, Rnd };
```
mrks: We don't like abbreviations

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,98 @@
...
+  // There are five possible inputs
+  enum class Input { Zero, One, Positive, NegativeOne, Negative };
+
+  constexpr static const std::array<std::array<AccessPattern, 5 /*|Input|*/>, 6 /*|AccessPattern|*/> _transitions{
```
mrks: This is only used in one method. Move it there?

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,168 @@
...
+T SegmentAccessCounter::Counter<T>::sum() const {
+  return other + iterator_create + iterator_seq_access + iterator_increasing_access + iterator_random_access +
+         accessor_create + accessor_access + dictionary_access;
+}
+
+// Explicit instantiation of Counters used
+template class SegmentAccessCounter::Counter<uint64_t>;
+template class SegmentAccessCounter::Counter<std::atomic_uint64_t>;
+
+SegmentAccessCounter::AccessPattern SegmentAccessCounter::_iterator_access_pattern(
```
mrks: doc?

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,168 @@
...
+      iterator_random_access{0},
+      accessor_create{0},
+      accessor_access{0},
+      dictionary_access{0} {}
+
+template <typename T>
+std::string SegmentAccessCounter::Counter<T>::to_string() const {
+  return (boost::format("%d,%d,%d,%d,%d,%d,%d,%d") % other % iterator_create % iterator_seq_access %
+          iterator_increasing_access % iterator_random_access % accessor_create % accessor_access % dictionary_access)
+      .str();
```
mrks: Having that array would also make this part shorter / cleaner

src/test/storage/segment_access_counter_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+  EXPECT_EQ(0, counter.sum());
+  counter.other = 1;
+  counter.iterator_create = 2;
+  counter.iterator_seq_access = 3;
+  counter.iterator_increasing_access = 4;
+  counter.iterator_random_access = 5;
+  counter.accessor_create = 6;
+  counter.accessor_access = 7;
+  counter.dictionary_access = 8;
+  EXPECT_EQ(36, counter.sum());
```
mrks: What is this supposed to test? Also, what is sum? Why does it contain the creation of things?

src/test/storage/segment_access_counter_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+template <typename T>
+class SegmentAccessCounterCounterTest : public BaseTest {
+ public:
+  static void SetUpTestCase() {}
+
+  void SetUp() override {}
+
+  void TearDown() override {}
+};
```
mrks: What are these needed for?

src/lib/storage/value_segment.cpp
mflueggen: That is a tough question.

1. A freshly copied segment hasn’t been accessed yet. One could expect the counters to be zero.
2. When copying an object, one might expect an exact duplicate.

Personally, I think I would expect an exact duplicate. I can always reset the counters manually, if required, and will know, that I am not holding an exact copy anymore.

mrks: 2 sounds reasonable. That should explicitly be documented somewhere.

src/lib/storage/segment_accessor.hpp
mflueggen: I am not sure about "linearly". I would say it refers to to "faster" and not to "grow", but I really have no idea.

mrks: It does. In any case, "faster" is an comparative adjective, "grow" is a verb. Both are described by the adverb form (linearly).

src/lib/storage/dictionary_segment.cpp
```diff
@@ -39,6 +39,7 @@ AllTypeVariant DictionarySegment<T>::operator[](const ChunkOffset chunk_offset)
 
 template <typename T>
 std::shared_ptr<const pmr_vector<T>> DictionarySegment<T>::dictionary() const {
+  access_counter.get(SegmentAccessCounter::AccessType::Dictionary) += _dictionary->size();
```
mflueggen: Maybe even sequential?

src/lib/storage/segment_access_counter.hpp
mrks: What does "how" mean? Maybe you could also add who increments the counter and how it is used (i.e., consumers have to track differences).

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,73 @@
...
+    Sequential /* 0, 1, 1, 2, 3, 4 */,
+    Increasing /* 0, 0, 1, 2, 4, 8, 17 */ /* or decreasing */,
+    Random /* 0, 1, 0, 42 */,
+    Dictionary,
+    Count /* Dummy entry to describe the number of elements in this enum class. */
+  };
+
+  inline static const std::array<std::string, (size_t)AccessType::Count> access_type_string_mapping = {
+    "Point", "Sequential", "Increasing", "Random", "Dictionary"
+  };
```
mrks: What is this used for? If this is just used for debugging, overloading `operator<<` makes the usage easier. If you actually need the string, it might be better to use an unordered_map here. It is slightly more expensive, but you don't run into issues when enum values are changed.

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+ public:
+  using CounterType = std::atomic_uint64_t;
+
+  enum class AccessType {
+    Point /* Single point access */,
+    Sequential /* 0, 1, 1, 2, 3, 4 */,
+    Increasing /* 0, 0, 1, 2, 4, 8, 17 */ /* or decreasing */,
+    Random /* 0, 1, 0, 42 */,
+    Dictionary,
```
mrks: This could use a comment, too. People might be surprised to see a pos list above and dictionary here.

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,73 @@
...
+  SegmentAccessCounter();
+  SegmentAccessCounter(const SegmentAccessCounter& counter);
+  SegmentAccessCounter& operator=(const SegmentAccessCounter& counter);
+
+  CounterType& get(const AccessType type);
+  const CounterType& get(const AccessType type) const;
+
+  static AccessType access_type(const PosList& positions);
+
+  uint64_t sum() const;
```
mrks: Still not sure what this is good for. In tests, you can also compare against `EXPECT_EQ(..., {1, 5, 4});`

src/lib/storage/segment_access_counter.hpp
```diff
@@ -0,0 +1,73 @@
...
+class SegmentAccessCounter {
+  friend class SegmentAccessCounterTest;
+
+ public:
+  using CounterType = std::atomic_uint64_t;
+
+  enum class AccessType {
+    Point /* Single point access */,
+    Sequential /* 0, 1, 1, 2, 3, 4 */,
+    Increasing /* 0, 0, 1, 2, 4, 8, 17 */ /* or decreasing */,
```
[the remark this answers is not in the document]
mrks: Maybe call the type "Monotonic"?

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,131 @@
...
+      }
+    }
+  }
+}
+
+SegmentAccessCounter::AccessPattern SegmentAccessCounter::_access_pattern(const PosList& positions) {
+  // There are five possible inputs
+  enum class Input { Zero, One, Positive, NegativeOne, Negative };
+
+  constexpr std::array<std::array<AccessPattern, 5 /*|Input|*/>, 6 /*|AccessPattern|*/> _transitions{
```
mrks: Could you doc the algorithm in ~two lines?

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,131 @@
...
+  Fail("This code should never be reached.");
+}
+
+void SegmentAccessCounter::reset() {
+  for (auto& counter : _counters) counter = 0ul;
+}
+
+// Sets all counters of all segments in tables to zero.
+void SegmentAccessCounter::reset(const std::map<std::string, std::shared_ptr<Table>>& tables) {
+  for (const auto& [table_name, table_ptr] : tables) {
```
mrks: Just `table` (instead of `table_ptr`) would be more consistent with the rest of the code base; same with chunk.

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,131 @@
...
+    for (auto chunk_id = ChunkID{0}; chunk_id < table_ptr->chunk_count(); ++chunk_id) {
+      const auto chunk_ptr = table_ptr->get_chunk(chunk_id);
+      for (auto column_id = ColumnID{0}, count = static_cast<ColumnID>(chunk_ptr->column_count()); column_id < count;
```
mrks: count of what? Also, I'd move it out of the for clause to improve readability.

src/lib/storage/segment_access_counter.cpp
```diff
@@ -0,0 +1,131 @@
...
+  }
+  Fail("This code should never be reached.");
+}
+
+void SegmentAccessCounter::reset() {
+  for (auto& counter : _counters) counter = 0ul;
+}
+
+// Sets all counters of all segments in tables to zero.
+void SegmentAccessCounter::reset(const std::map<std::string, std::shared_ptr<Table>>& tables) {
```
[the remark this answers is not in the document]
mflueggen: True.

src/lib/storage/segment_access_counter.hpp
mflueggen: The function can be nice if you just want to check if the segment has been accessed at all.

mrks: `!= ArrayType{}`?

src/lib/storage/segment_access_counter.hpp
mflueggen: I use the mapping to export the counters to a csv file. I could just create the headers in my export function. However, I'd perfer having the mapping at one central place to reduce the effort if somebody would ever change the access types. I'll create a map.

src/lib/storage/segment_access_counter.hpp
mflueggen: That is a pretty cool idea.

src/lib/storage/dictionary_segment.cpp
```diff
@@ -39,6 +39,7 @@ AllTypeVariant DictionarySegment<T>::operator[](const ChunkOffset chunk_offset)
 
 template <typename T>
 std::shared_ptr<const pmr_vector<T>> DictionarySegment<T>::dictionary() const {
+  access_counter[SegmentAccessCounter::AccessType::Dictionary] += _dictionary->size();
```
mflueggen: Increasing the counter by dictionary->size() is not correct for all cases but probably okay at most times...

mrks:
Could you please make sure the master is merged and compare TPC-H for the current master with this branch?

src/lib/storage/segment_access_counter.hpp
```diff
@@ -10,9 +10,12 @@
...
 
 namespace opossum {
-class Table;
 
-// The SegmentAccessCounter is a collection of counters to register how often and how a segment is accessed.
+// The SegmentAccessCounter is a collection of counters to count how often a segment is accessed.
+// It contains several counters (see AccessType) to differentiate between different access types, like
+// sequential or random access. The individual counters can be accessed using the [] operator.
+// The counters are currently updated by the iterators, segment accessors or from withing the segment itself.
+// Use function access_type(const PosList& positions) to derive the access pattern from the position list.
```
mrks: Suggestion: Move the last line to the function itself and make it "For a given position list, this determines whether its entries are in sequential, monotonic, or random order. It only looks at the first n values."

mflueggen:
> 
> 
> Could you please make sure the master is merged and compare TPC-H for the current master with this branch?

Sure. With what settings should I be executing the TPC-H benchmark. Should I do two runs: One singlethreaded and one multithreaded?

mrks:
Sounds good. Let's do settings for ST and `--scheduler --clients=[numcpus] -t 180` for MT on a physical server.

mflueggen:
Benchmark results on Pella with `numactl -N 2 -m 2`:
(A positive change is in favour of the PR...)

`./hyriseBenchmarkTPCH --output singlethreaded.json`
grafik

`./hyriseBenchmarkTPCH --scheduler --clients=20 -t 180 --output multithreaded.json`
grafik

mrks:
Ok, so nothing that would be a concern. Nice. Anything missing from your side?

mflueggen:
I think we are good as long as the majority of accesses happens over the iterators and accessors. In the near future I can imagine segment type specific adjustments and additional counters, as we will see what we are missing. But I guess that is secondary, as long as the interface does not change.

mrks:
It's not even that permanent. We can still change everything if we see the need to. Just wanted to make sure you have nothing you wanted to add to this PR.

Issue #1979, opened 2020-01-10
Meta Tables: tables are created multiple times when being queried


Bouncner:
As discussed in #1968, meta tables are generated several times during a query.

Apparently, they are created 10 times during the SQL translation. What might be needed is something that allows users of the storage manager to retrieve the properties of the meta tables without generating the actual rows and then having the GetTable operator finally creating the rows.

mrks:
@j-tr Is this resolved? If yes, please remember to add `fixes #...` to the PRs. If not, what is missing?

PR #1998, opened 2020-01-24
Generate MetaTables only once

j-tr:
Addressing issue: #1979 meta tables are created multiple times when being queried

idea: 
* StorageManager::get_table() produces an empty table stub that can be used by the sql translator
* GetTable::GetTable() triggers the actual generation of the content 

potential problem: 
* there might be multiple GetTable() calls in one query yielding again multiple table generations

### DIFF
--- a/src/lib/utils/meta_table_manager.cpp
+++ b/src/lib/utils/meta_table_manager.cpp
@@ -90,12 +90,18 @@ auto gather_segment_meta_data(const std::shared_ptr<Table>& meta_table, const Me
 namespace opossum {
 
 MetaTableManager::MetaTableManager() {
-  _methods["tables"] = &MetaTableManager::generate_tables_table;
-  _methods["columns"] = &MetaTableManager::generate_columns_table;
-  _methods["chunks"] = &MetaTableManager::generate_chunks_table;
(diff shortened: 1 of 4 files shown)


src/lib/utils/meta_table_manager.cpp
```diff
@@ -90,12 +90,18 @@ auto gather_segment_meta_data(const std::shared_ptr<Table>& meta_table, const Me
...
 
 MetaTableManager::MetaTableManager() {
-  _methods["tables"] = &MetaTableManager::generate_tables_table;
-  _methods["columns"] = &MetaTableManager::generate_columns_table;
-  _methods["chunks"] = &MetaTableManager::generate_chunks_table;
-  _methods["chunk_sort_orders"] = &MetaTableManager::generate_chunk_sort_orders_table;
-  _methods["segments"] = &MetaTableManager::generate_segments_table;
-  _methods["segments_accurate"] = &MetaTableManager::generate_accurate_segments_table;
+  _methods["tables"] = {&MetaTableManager::generate_tables_table,
+                              &MetaTableManager::generate_tables_table_stub};
```
mrks: Any reason not to use a boolean parameter that tells the method to fill the table with data (or not)?

PR #2001, opened 2020-01-31
Generate meta tables only once during translation

j-tr:
Addressing issue: #1979 meta tables are created multiple times when being queried

Fixed by using StaticTableNodes for meta tables and don't cache LQPs/PQPs that involve meta tables.


src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -92,6 +92,8 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_unoptimized_lo
   DebugAssert(lqp_roots.size() == 1, "LQP translation returned no or more than one LQP root for a single statement.");
   _unoptimized_logical_plan = lqp_roots.front();
 
+  _cacheable = sql_translator.cacheable();
```
mrks: Convenient, but I am not sure if this is really a property/responsibility of the SQLTranslator. Another option would be to have this property stored in the StoredTableNode, but then we would have to traverse each result and check for StoredTableNodes, which feels like pointless work. :/

src/lib/sql/sql_translator.cpp
```diff
@@ -527,8 +527,11 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_origin(const hsq
         }
 
       } else if (Hyrise::get().storage_manager.has_table(hsql_table_ref.name)) {
-        lqp = _translate_stored_table(hsql_table_ref.name, sql_identifier_resolver);
-
+        if (MetaTableManager::is_meta_table_name(hsql_table_ref.name)) {
```
mrks: Do we still need the StorageManager to return `has_table` or can we completely switch to the SQLTranslator handling this?

src/lib/sql/sql_translator.cpp
```diff
@@ -654,6 +657,32 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_stored_table(
...
 }
 
+std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_meta_table(
+    const std::string& name, const std::shared_ptr<SQLIdentifierResolver>& sql_identifier_resolver) {
+  AssertInput(Hyrise::get().storage_manager.has_table(name), std::string{"Did not find a table with name "} + name);
+  AssertInput(MetaTableManager::is_meta_table_name(name), std::string{"Did not find a meta table with name "} + name);
+
+  _cacheable = false;
+
+  const auto meta_table = Hyrise::get().storage_manager.get_table(name);
```
mrks: Same thing as above, can't we talk directly to the MetaTableManager?

src/lib/sql/sql_translator.cpp
```diff
@@ -654,6 +657,32 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_stored_table(
...
+
+  const auto static_table_node = StaticTableNode::make(meta_table);
+  const auto validated_static_table_node = _validate_if_active(static_table_node);
+
+  // Publish the columns of the table in the SQLIdentifierResolver
```
mrks: This could be extended a bit

src/lib/sql/sql_translator.hpp
```diff
@@ -51,6 +51,8 @@ class SQLTranslator final {
    */
   static std::shared_ptr<AbstractExpression> translate_hsql_expr(const hsql::Expr& hsql_expr, const UseMvcc use_mvcc);
 
+  bool cacheable();
```
mrks: If it stays: doc?

src/lib/sql/sql_pipeline_statement.cpp
mrks: In any case, we should have some documentation here on why some things are not cacheable.

j-tr: indeed. another option would be to put the flag in the AbstractLQPNode and set it for the root node if the whole LQP contains a metatable access. this way, neither the sql_translator nor the sql_pipelined_statement would need a member variable. what do you think about this?

mrks: By design, LQP nodes shouldn't statically hold any information about their leafs. If they did, e.g., hold the order of expressions, LQP manipulations below the node would become a pain.

In this case, doing so would probably be harmless, but I don't want to open Pandora's box.

j-tr: our discussed solution of passing the parser result to the SQLTranslator constructor isn't that straight forward since other instances of the SQLTranslator are sometimes used within the initial sql translater in order to translate other kinds of SQL statements that don't need any sqlParserResults. see:
https://github.com/j-tr/hyrise/blob/06a5f900999a497c30a96dd5e9bb8f79e60f2627/src/lib/sql/sql_translator.cpp#L288
https://github.com/j-tr/hyrise/blob/06a5f900999a497c30a96dd5e9bb8f79e60f2627/src/lib/sql/sql_translator.cpp#L308
I think the with_translator could be omitted and the method could be called directly instead but the one in translate_hsql_expr needs to have an empty context. Do you have an idea how to solve this?

src/lib/sql/sql_translator.cpp
```diff
@@ -654,6 +657,32 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_stored_table(
...
+std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_meta_table(
+    const std::string& name, const std::shared_ptr<SQLIdentifierResolver>& sql_identifier_resolver) {
+  AssertInput(Hyrise::get().storage_manager.has_table(name), std::string{"Did not find a table with name "} + name);
+  AssertInput(MetaTableManager::is_meta_table_name(name), std::string{"Did not find a meta table with name "} + name);
+
+  _cacheable = false;
+
+  const auto meta_table = Hyrise::get().storage_manager.get_table(name);
+
+  const auto static_table_node = StaticTableNode::make(meta_table);
```
mrks: Should we have a flag in StoredTableNode that prevents deep_copy from being called on meta tables?

src/test/sql/sql_translator_test.cpp
```diff
@@ -1875,6 +1875,15 @@ TEST_F(SQLTranslatorTest, ShowColumns) {
   EXPECT_LQP_EQ(actual_lqp, expected_lqp);
 }
 
+TEST_F(SQLTranslatorTest, SelectMetaTable) {
```
mrks: We should have more complicated tests on the parts that call their own translator (e.g. subqueries?)

src/lib/sql/sql_pipeline_statement.cpp
mrks: Hmpf . :/

in that case, returning a pair with TranslationInfo might be easier. Can we make the ID thing used by the server part of that, too?

j-tr: i now implemented a solution that avoids modifying the return types of all translate_* functions by still using a private member variable _cacheable but without having a public getter for it. This way, from an external view, the SQL translator is stateless

src/lib/sql/sql_translator.cpp
j-tr: do you mean StaticTableNode? what's exactly the problem with deep copy here?

mrks: deep_copy is called when a table is reused. Those table should not be reused, but recreated, so this would be a sanity check.

j-tr: i don't get it yet. isn't a table noncopyable?

mrks: Yes, but deep-copying the StaticTableNode only copies the pointer to the table. As such, while the table is not copied, it is erroneously reused.

j-tr: i see. maybe a similar but different problem that came to my mind: how do we deal with static table nodes in prepared statements? they might get outdated at the time of execution.

mrks: Good that you think of it. I don't think that we lose anything if we simply prohibit the combination of meta tables and prepared statements.

mrks:
Please ping me when this is ready for another look

j-tr:
> 
> 
> Please ping me when this is ready for another look

@mrks ping

src/lib/logical_query_plan/static_table_node.cpp
```diff
@@ -56,6 +56,7 @@ size_t StaticTableNode::_on_shallow_hash() const {
 }
 
 std::shared_ptr<AbstractLQPNode> StaticTableNode::_on_shallow_copy(LQPNodeMapping& node_mapping) const {
+  Assert(_copyable, "StaticTableNode cannot be shallow copied");
```
mrks: *This* StaticTableNode cannot be shallow copied?

src/lib/logical_query_plan/static_table_node.hpp
```diff
@@ -24,6 +24,8 @@ class StaticTableNode : public EnableMakeForLQPNode<StaticTableNode>, public Bas
 
  protected:
   mutable std::optional<std::vector<std::shared_ptr<AbstractExpression>>> _column_expressions;
+  // static table nodes that reference meta tables should not be copied
+  const bool _copyable;
```
mrks: Suggestion: "Some tables should not be copied but recreated. Currently, this applies to meta tables."

Also, I'd move that to the constructor (which has no documentation on that parameter right now)

src/lib/sql/sql_translator.cpp
```diff
@@ -136,19 +136,10 @@ SQLTranslator::SQLTranslator(const UseMvcc use_mvcc)
...
-
-  for (const auto& [value_placeholder_id, parameter_id] : parameter_ids_of_value_placeholders) {
-    parameter_ids[value_placeholder_id] = parameter_id;
-  }
-
-  return parameter_ids;
-}
-
-std::vector<std::shared_ptr<AbstractLQPNode>> SQLTranslator::translate_parser_result(
+std::pair<std::vector<std::shared_ptr<AbstractLQPNode>>, TranslationInfo> SQLTranslator::translate_parser_result(
```
mrks: Change this to `SQLTranslationResult` and make the vector a part of it? That would save us from having to remember what `.first` and `.second` are.

src/lib/sql/sql_translator.cpp
```diff
@@ -654,6 +648,32 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_stored_table(
...
+std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_meta_table(
+    const std::string& name, const std::shared_ptr<SQLIdentifierResolver>& sql_identifier_resolver) {
+  AssertInput(MetaTableManager::is_meta_table_name(name), std::string{"Did not find a meta table with name "} + name);
+
+  // MetaTables are non-cacheable because they might contain information about the general system state
+  // that can change at any time
+  _cacheable = false;
+
+  const auto meta_table_name = name.substr(MetaTableManager::META_PREFIX.size());
+  const auto meta_table = Hyrise::get().meta_table_manager.generate_table(meta_table_name);
```
mrks: Please describe why we are doing this here even though it does not really belong to the translation.

src/lib/sql/sql_translator.hpp
```diff
@@ -171,9 +174,16 @@ class SQLTranslator final {
       const std::vector<SelectListElement>& select_list_elements);
 
  private:
+  /**
+  * @return after translate_*(), contains the ParameterIDs allocated for the placeholders in the query
+  */
+  std::vector<ParameterID> _parameter_ids_of_value_placeholders() const;
```
mrks: What do we still need this for, can't it be part of the info/result?

j-tr:
` SELECT * FROM meta_tables WHERE table_name != (SELECT table_name from meta_tables)`
this example query still suffers from two different problems: the table is generated twice, and the execution fails since the where clause currently needs to copy the table and we just prohibited this.
in my opinion, in this case we would rather want to have the table copied and not regenerate it and we should cache a table for the time of one query translation but prevent the caching of the complete lqp. i thought of introducing a map of meta table names and static table nodes that holds all created nodes containing meta tables and is cleared at the end of the translate_parser_result method. what do you think?

mrks:
If that's easy to do, that sounds like a good idea. I wouldn't spend too much time on it though - I don't see this actually becoming an issue anytime soon.

j-tr:
hm. forgot about that there ist again the problem of the independent instance of the SQLTranslator for subqueries. So we would need to pass the intra-query cache somehow to the new translator instance that is used for each subqueries and back, because potentially multiple subqueries could all use independent SQLTranslators but operate on the same meta table and should all know about what meta tables have already been used by the other subqueries. This seems somehow very hacky but just prohibiting subqueries on meta tables seems to be no good solution neither.

mrks:
Hmm, we could also translate the tables without materializing them and visit the LQP (with sub-LQPs) later on, replacing all StoredTableNodes with StaticTableNodes.

j-tr:
i went for the solution that passes a shared pointer to a cache to all the subqueries.

src/lib/sql/sql_translator.hpp
```diff
@@ -33,23 +43,20 @@ class SQLTranslator final {
...
    * @param use_mvcc  Whether ValidateNodes should be compiled into the plan
    */
-  explicit SQLTranslator(const UseMvcc use_mvcc);
-
-  /**
-   * @return after translate_*(), contains the ParameterIDs allocated for the placeholders in the query
-   */
-  std::vector<ParameterID> parameter_ids_of_value_placeholders() const;
+  explicit SQLTranslator(const UseMvcc use_mvcc,
+                         std::shared_ptr<std::unordered_map<std::string, std::shared_ptr<Table>>> meta_tables =
```
mrks: Doc the parameter?

Also: Is this parameter meant to be used externally? If not, should we have a private constructor that takes this argument?

src/test/sql/sql_translator_test.cpp
```diff
@@ -1865,16 +1868,44 @@ TEST_F(SQLTranslatorTest, ShowColumns) {
...
+  const LQPColumnReference table_name_column{static_table_node, meta_table->column_id_by_name("table_name")};
+  const LQPColumnReference column_count_column{static_table_node, meta_table->column_id_by_name("column_count")};
+
+  const auto expected_subquery_lqp =
+      ProjectionNode::make(expression_vector(table_name_column, column_count_column), static_table_node);
+  const auto expected_lqp = ProjectionNode::make(expression_vector(table_name_column), expected_subquery_lqp);
+
+  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+}
+
```
mrks: Do we have a test where the same meta table is used twice and where we test that both StaticTableNodes point to the same table?

src/lib/sql/sql_translator.hpp
```diff
@@ -20,6 +20,16 @@ namespace opossum {
 class AggregateNode;
 class LQPSubqueryExpression;
 
+struct TranslationInfo {
+  bool cacheable{true};
+  std::vector<ParameterID> parameter_ids_of_value_placeholders{};
+};
+
+struct SQLTranslationResult {
```
mrks: Please doc the two structs

src/lib/sql/sql_translator.cpp
```diff
@@ -654,6 +658,41 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_stored_table(
...
+  std::shared_ptr<Table> meta_table;
+  if (_meta_tables->contains(meta_table_name)) {
+    meta_table = _meta_tables->at(meta_table_name);
+  } else {
+    meta_table = Hyrise::get().meta_table_manager.generate_table(meta_table_name);
+    (*_meta_tables)[meta_table_name] = meta_table;
+  }
+
+  const auto static_table_node = StaticTableNode::make(meta_table, false);
+  const auto validated_static_table_node = _validate_if_active(static_table_node);
```
Bensk1: You might have discussed that already but why do we need validation for meta tables?

src/lib/sql/sql_translator.cpp
```diff
@@ -132,23 +132,14 @@ bool is_trivial_join_predicate(const AbstractExpression& expression, const Abstr
...
-  auto parameter_ids = std::vector<ParameterID>{parameter_ids_of_value_placeholders.size()};
-
-  for (const auto& [value_placeholder_id, parameter_id] : parameter_ids_of_value_placeholders) {
-    parameter_ids[value_placeholder_id] = parameter_id;
-  }
-
-  return parameter_ids;
-}
+SQLTranslationResult SQLTranslator::translate_parser_result(const hsql::SQLParserResult& result) {
+  _cacheable = true;
```
Bensk1: Do we need that since `_cacheable` default to `true`?

src/lib/sql/sql_translator.hpp
j-tr: private constructor is a good idea. done

src/lib/sql/sql_translator.cpp
j-tr: you are right. this seems useless. 
but since the latest changes in the master allow entries of certain meta tables to be updated, inserted or deleted and these operators require the table to be validated, this validation step is necessary. 
An option would be to check for mutability of the meta table and validate accordingly, but i don't know if this really has an impact, since so far, meta tables tend to be very small.

src/lib/sql/sql_translator.cpp
j-tr: imo yes. since one parser can be used to translate multiple queries subesquently, the cacheable attribute needs to be reset at each call. instead we can omit the initialization

Bensk1: Sounds reasonable, it should stay as is.

src/lib/sql/sql_translator.cpp
Bensk1: Correct, we do have inserts, updates, and deletes on meta tables now but the tables are still generated on the fly for each access. For example, updates do not really change a table but some internal data structure that is part of the meta table object. Therefore, they should never contain invisible rows.

Bensk1:
Strange CI error, I haven't seen this before. @Bouncner, have you? 

```
[ RUN      ] StressTest.TestTransactionConflicts
/Users/hyrise-ci/.jenkins/workspace/hyrise_hyrise_PR-2001@tmp/durable-24ffc3b0/script.sh: line 1: 93934 Bus error: 10           ./clang-debug/hyriseSystemTest --gtest_filter=-TPCCTest*:TpcdsTableGeneratorTest.*:TPCHTableGeneratorTest.RowCountsMediumScaleFactor:*.CompareToSQLite/Line1*WithLZ4
script returned exit code 138
```

Bouncner:
Nope, haven’t seen this one before.

Bensk1:
It's working now. This should be ready to merge as soon as we resolve the remaining discussion rergarding validation and Markus' last question.

src/lib/sql/sql_translator.cpp
j-tr: we should never need it. it was just required for all kinds of tables from the tests. removed it for meta tables in the tests

src/test/sql/sql_translator_test.cpp
j-tr: we have a little conflict here with the copyable flag of storedTableNodes. Right now, the only situation, where this could happen is in a where clause but since where clauses copy nodes, they are currently not allowed for meta tables. other use cases such as unions are not yet implemented. 
imo, we should drop the copyable flag, since copying a meta table within one query isn't a problem anyway

Bensk1: What do you mean by where clauses are not allowed for meta tables? `SELECT * FORM meta_segments WHERE table_name = 'lineitem'` does work.

j-tr: yes. but if the where clause contains a subquery, it's not allowed anymore. E.g.
`SELECT * FROM meta_tables WHERE table_name != (SELECT table_name from meta_tables LIMIT 1)`

j-tr: anyway, there is now a test that covers the case where a meta table is used twice but generated only once. somehow, i thought we need a where clause for this to happen but a self join causes this as well

Bensk1: Great.

Bensk1:
So, this should be good to go?

j-tr:
ok from my side

PR #2002, opened 2020-02-03
Jenkins: disable ccache for clang

mrks:
Hopefully this stops CI from ruining everyone's day.

Seems to be a known issue: https://gitlab.kitware.com/cmake/cmake/issues/19923

Also removes the no-numa builds. We don't do too much differently there and the no-libnuma support is tested in the mac build anyway.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -95,7 +95,7 @@ try {
 
         parallel clangDebug: {
           stage("clang-debug") {
-            sh "export CCACHE_BASEDIR=`pwd`; cd clang-debug && make all -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 4)) && ../scripts/analyze_ccache_usage.py"
+            sh "cd clang-debug && make all -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 4))"
             sh "./clang-debug/hyriseTest clang-debug"
           }
         }, gccDebug: {
@@ -114,7 +114,7 @@ try {
         parallel clangRelease: {
           stage("clang-release") {
             if (env.BRANCH_NAME == 'master' || full_ci) {
-              sh "export CCACHE_BASEDIR=`pwd`; cd clang-release && make all -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 6)) && ../scripts/analyze_ccache_usage.py"
+              sh "cd clang-release && make all -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 6))"
               sh "./clang-release/hyriseTest clang-release"
               sh "./clang-release/hyriseSystemTest clang-release"
               sh "./scripts/test/hyriseConsole_test.py clang-release"
@@ -159,7 +159,7 @@ try {
           stage("clang-debug-unity-odr") {
             if (env.BRANCH_NAME == 'master' || full_ci) {
               // Check if unity builds work even if everything is batched into a single compilation unit. This helps prevent ODR (one definition rule) issues.
-              sh "export CCACHE_BASEDIR=`pwd`; cd clang-debug-unity-odr && make all -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 3))"
+              sh "cd clang-debug-unity-odr && make all -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 3))"
             } else {
               Utils.markStageSkippedForConditional("clangDebugUnity")
             }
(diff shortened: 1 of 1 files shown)


Bouncner:
Just curious: why disabling ccache and not PCH?

mrks:
You can't disable PCH from the outside, so we would need another CMake flag (and thus, a different configuration).

Jenkinsfile
```diff
@@ -74,26 +74,26 @@ try {
           disable_aslr = 'setarch x86_64 -R'
 
           // Run cmake once in isolation and build jemalloc to avoid race conditions with autoconf (#1413)
-          sh "mkdir clang-debug && cd clang-debug &&                                                       ${cmake} ${ccache} ${debug}   ${clang}          .. && make -j libjemalloc-build"
```
Bouncner: Why only for debug? I don’t get what we are doing here from the comment.

Jenkinsfile
```diff
@@ -74,26 +74,26 @@ try {
...
+          sh "mkdir clang-debug && cd clang-debug &&                                                                   ${cmake} ${debug}                    ${clang} ${unity} .. && make -j libjemalloc-build"
 
           // Configure the rest in parallel
-          // Most builds don't use unity builds as this would break the ccache use across different PRs. For clang-tidy, ccache does not have an effect, so we use unity there.
+          // clang + precompiled headers + ccache is broken: https://gitlab.kitware.com/cmake/cmake/issues/19923
+          // Some confirgurations don't use unity builds as this would break the ccache use across different PRs.
```
Bouncner: Typo

Jenkinsfile
```diff
@@ -74,26 +74,26 @@ try {
...
+          mkdir clang-release && cd clang-release &&                                                                   ${cmake}           ${release}        ${clang}          .. &\
+          mkdir clang-relwithdebinfo-thread-sanitizer && cd clang-relwithdebinfo-thread-sanitizer &&                   ${cmake}           ${relwithdebinfo} ${clang}          -DENABLE_THREAD_SANITIZATION=ON .. &\
           mkdir gcc-debug && cd gcc-debug &&                                                                           ${cmake} ${ccache} ${debug}          ${gcc}            .. &\
           mkdir gcc-release && cd gcc-release &&                                                                       ${cmake} ${ccache} ${release}        ${gcc}            .. &\
           wait"
         }
 
         parallel clangDebug: {
           stage("clang-debug") {
-            sh "export CCACHE_BASEDIR=`pwd`; cd clang-debug && make all -j \$(( \$(cat /proc/cpuinfo | grep processor | wc -l) / 4)) && ../scripts/analyze_ccache_usage.py"
```
Bouncner: Aren’t you afraid we won’t put the analayze script back in once the cmake bug is fixed?

mrks: Doesn't really matter. That was just for me to check things in the output of Jenkins. If we put ccache back in and forget it, the worst thing that could happen is someone missing and adding it.

Jenkinsfile
mrks: better?

Bouncner: So it’s not about debug. It’s about running autoconf “anyhow” and afterwards everything works?

mrks: correct

Bouncner: Remove the empty line + replace “really” with “using debug here”)?

For my understanding: for cmake you need a new cmake run every time you change from debug to release. For automake, that’s not the case?

mrks: Correct. I don't want to mention debug here because it has nothing to do with debug.

If you don't insist, I#d rather merge this once it's green. This is *the* PR needed to unfuck CI.

Bouncner: The empty line still bothers me though. :D

Bouncner:
Just a note: depending on how long it takes to fix this issue, we might want to fix it outside of Jenkins as well. Got the same problem on nemea.

mrks:
Any chance you are changing files while a build is in progress? I saw the same thing when I started ninja and while waiting for the build to finished, changed the header again (e.g., added a comment).

Bouncner:
Nope. It was a benchmark run over night. After the first run (which built just fine), another branch was checked out and ninja was called. No concurrent modifications.

mrks:
Do you still have the error message? I'd like to see which files were affected.

Bouncner:
I am currently running a benchmark. After that, I can rebuild the tests which failed. Do you want the log or the folder to look into it yourself?

mrks:
Let's start with just one or two lines from the log.

Bouncner:
https://gist.github.com/Bouncner/af53fa445f66fb3401e62c3ede73f2c4

mrks:
Is that a nemea folder that I can access?

Bouncner:
Feel free. I am working in the `rel` directory.

mrks:
Working on the HP thing right now. If this is reproducible, let's look at it together next week.

Bouncner:
Ok.

PR #2004, opened 2020-02-03
Sort merge join: do not concatenate clusters

Bouncner:
This PR is a rather simple change for the SMJ. Right now, after the actual merge is done (which is done on multiple clusters in parallel for large joins), the clusters are concatenated into a single large chunk at the end.

This has various problems: (i) it is unnecessary and slows the join down, (ii) it prohibits exploiting the sorted flag (given per chunk, but no longer possible when concatenating), and (iii) is hinders the following operators from parallelizing over chunks.

This PR no longer concatenates chunks and further parallelizes the output writing as it has shown to be significant in some situations (does not matter for ST execution).


src/lib/operators/join_sort_merge.cpp
```diff
@@ -910,31 +889,58 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
-        output_right->push_back(NULL_ROW_ID);
+    if (include_null_left || include_null_right) {
+      auto null_output_left = std::make_shared<PosList>();
+      auto null_output_right = std::make_shared<PosList>();
+      null_output_left->reserve(_null_rows_left->size());
+      null_output_right->reserve(_null_rows_right->size());
+
+      // Add the outer join rows which had a null value in their join column
+      if (include_null_left) {
+        for (auto row_id_left : *_null_rows_left) {
```
mrks: const?

Bouncner:
Closing for now as there is an issue with debug builds failing which run just fine in release.

Bouncner:
Hash Join performance:
```
----------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations
----------------------------------------------------------------------------
BM_Join_SmallAndSmall<JoinHash>       255442 ns       255447 ns         2051
BM_Join_SmallAndBig<JoinHash>      159634924 ns    159602007 ns            5
BM_Join_MediumAndMedium<JoinHash>   15220566 ns     15220865 ns           44
```

Old SortMergeJoin:
```
---------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations
---------------------------------------------------------------------------------
BM_Join_SmallAndSmall<JoinSortMerge>       267951 ns       267956 ns         2629
BM_Join_SmallAndBig<JoinSortMerge>      620472431 ns    620473053 ns            1
BM_Join_MediumAndMedium<JoinSortMerge>   18120199 ns     18120300 ns           37
```

New SortMergeJoin:
```
---------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations
---------------------------------------------------------------------------------
BM_Join_SmallAndSmall<JoinSortMerge>       269358 ns       269350 ns         2576
BM_Join_SmallAndBig<JoinSortMerge>      593867779 ns    593871558 ns            1
BM_Join_MediumAndMedium<JoinSortMerge>   16502659 ns     16500681 ns           41
```

Benchmarks are noisy. There is a small improvement (especially for large joins), but not much. More important is the ability now to mark chunks as sorted.

Bouncner:
In case somebody wonders: for very large joins the SMJ remains to be faster than the HashJoin (that is without any bloom filters, which can have a vast impact in realistic joins) but it moved to 2Mx2M joins. So nothing we will probably see soon.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -910,31 +891,73 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
+        null_output_left->insert(null_output_left->end(), _null_rows_right->size(), NULL_ROW_ID);
+        null_output_right->reserve(_null_rows_right->size());
+        for (const auto& row_id_right : *_null_rows_right) {
+          null_output_right->push_back(row_id_right);
+        }
+      }
+
+      if (!null_output_left->empty()) {
+        _output_pos_lists_left.push_back(null_output_left);
+        _output_pos_lists_right.push_back(null_output_right);
```
mrks: Probably a stupid question, but why don't we have to do the same check for the right side?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -910,31 +891,73 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
     }
 
-    // Add the segments from both input tables to the output
-    Segments output_segments;
-    _add_output_segments(output_segments, _sort_merge_join.input_table_left(), output_left);
-    _add_output_segments(output_segments, _sort_merge_join.input_table_right(), output_right);
+    // Wait for all chunk creation tasks to finish
+    Hyrise::get().scheduler()->wait_for_tasks(output_jobs);
+
+    // Remove empty chunks which occur due to empty radix clusters or not matching tuples of clusters.
```
mrks: that or ', which'?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -910,31 +891,73 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
+    output_chunks.erase(
+        std::remove_if(output_chunks.begin(), output_chunks.end(),
+                       [](const auto& output_chunk) { return !output_chunk || output_chunk->size() == 0; }),
+        output_chunks.end());
 
-    // Build the output_table with one Chunk
-    return _sort_merge_join._build_output_table({std::make_shared<Chunk>(output_segments)});
+    // TODO(anyone): when setting chunk ordered by flags, check if output is really sorted for NULL-adding joins.
```
mrks: Are you suggesting adding a sanity check here? Is this something that is potentially broken?

Bouncner: It is a hint for anyone (i.e., the exploitation seminar group) to not blindly set the flag for all chunks.
I am not certain, that such chunks are sorted. So “anybody” should either not set flags for NULLable  segments or take a deeper look (maybe with a DebugAssert on is_sorted) into nulling joins.

mrks: Is this really an actionable TODO then or is it more of a warning?

Bouncner: It’s a to do, but I’ll talk to my seminar group. Probably van be removed as they are aware of it anyways.

Bouncner: Pinging @j-tr, @MasterCarl, @Boffmann here. I will remove the comment as it is not obvious what I am talking about.
Still, I want you to be aware of the "NULL chunks" for your own work on setting the sort flags.

src/lib/operators/join_sort_merge.cpp
Bouncner: The lists should have the same length. Added a DebugAssert.

mrks:
CI Fail is likely caused by #2007 not being part of this branch. Please merge the current master.

Bouncner:
@j-tr, @MasterCarl, @Boffmann could you take a look at this PR, please?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -910,31 +891,74 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
+
+    // Intermediate structure for output chunks (to avoid concurrent appending to table)
+    std::vector<std::shared_ptr<Chunk>> output_chunks(_output_pos_lists_left.size());
+
+    // Determine if writing output in parallel is necessary.
+    // As partitions ought to be roughly equally sized, looking at the first should be sufficient.
+    const auto write_output_concurrently = _cluster_count > 1 && _output_pos_lists_left[0]->size() > 10'000;
+
+    std::vector<std::shared_ptr<AbstractTask>> output_jobs;
+    output_jobs.reserve(_output_pos_lists_left.size());
```
j-tr: i don't know how much performance impact it is, but couldn't we avoid reserving extra memory in case we don't write concurrently in the end?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -910,31 +891,74 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
+        write_output_chunk();
       }
     }
 
-    // Add the segments from both input tables to the output
-    Segments output_segments;
-    _add_output_segments(output_segments, _sort_merge_join.input_table_left(), output_left);
-    _add_output_segments(output_segments, _sort_merge_join.input_table_right(), output_right);
+    // Wait for all chunk creation tasks to finish
+    Hyrise::get().scheduler()->wait_for_tasks(output_jobs);
```
j-tr: don't know how simple the logic behind wait_for_tasks is but could we avoid calling it by checking write_output_concurrently before?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -910,31 +891,74 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractJoinOperatorImpl {
...
-      for (auto row_id_right : *_null_rows_right) {
-        output_left->push_back(NULL_ROW_ID);
-        output_right->push_back(row_id_right);
+
+    // Intermediate structure for output chunks (to avoid concurrent appending to table)
+    std::vector<std::shared_ptr<Chunk>> output_chunks(_output_pos_lists_left.size());
+
+    // Determine if writing output in parallel is necessary.
+    // As partitions ought to be roughly equally sized, looking at the first should be sufficient.
+    const auto write_output_concurrently = _cluster_count > 1 && _output_pos_lists_left[0]->size() > 10'000;
```
MasterCarl: Is 10'000 a commonly used constant in Hyrise? It might be better to collect such parameters somewhere.

Bouncner: No, it's not commonly used in Hyrise. This number here has been evaluated with some microbenchmarks but I doubt there is a proper way to set it (it depends on concurrent queries, optimizing for latency or throughput and more) ...

@mrks : what's your take here?

src/lib/operators/join_sort_merge.cpp
Bouncner: @mrks : do you have an idea about the impact of `wait_for_tasks()`. Given that SMJ is the fastest join for very small data sets, it might make sense to do the check here (even though we will not switch to SMJs in the near future).

src/lib/operators/join_sort_merge.cpp
Bouncner: The number of output lists is given by the radix count. For small joins, where the `reserve()` could be noticable, the number of pos lists is very small and it should basically come for free.
In other cases, the joins are to large that it will never be noticable anyway.

It's probably impossible to measure the impact given the general complexity of join operators, but I'd vote for reserving vector whenever we have a decent idea how large they might become.

src/lib/operators/join_sort_merge.cpp
mrks: Mmmh, as it is only used here and it is truly a magic constant, giving it a name doesn't really change things for me. No strong opinion on this.

src/lib/operators/join_sort_merge.cpp
mrks: It just iterates over the input vector. If the vector is empty, it should have a negligible cost. I still like checking for `write_output_concurrently` because it reminds readers that this is not always done.

Bouncner:
Thanks @MasterCarl and @j-tr for your review. Should be addressed in 6a61d3a.

ArneMayer:
Great find :)

ArneMayer:
If you ever want me to review smth regardingn SMJ, hit me up :)

PR #1999, opened 2020-01-29
[DYOD] [WIP] Make sort operator multi-column capable and add SortMicroBenchmarks

jonasnoki:

PR #1987, opened 2020-01-15
[DYOD] Add null value specification to SyntheticTableGenerator

birneamstiel:
This will add a way to specify which part (ratio) of the generated table will be filled with null values. Since this would add another specification on per-column-basis to `generateTable` we decided to factor out all specifications that apply per column into one vector of a newly introduced struct `ColumnSpecification`.

### TODO
- [x] check for other usages
- [x] add FullCI label

PR #2005, opened 2020-02-05
[DYOD] Make Sort operator multi column capable and adjust LQP translator accordingly

bakoe:
Depends on/built upon #1987 

This will …

- make the existing Sort operator (with an otherwise unchanged implementation) capable of sorting multiple columns at once (as opposed to before, where it was only capable of sorting one column) and
- adjust the LQP Translator accordingly to not create n Sort operator instances for n to-be-sorted columns anymore, but only 1.

### Benchmarking results

<details>
  <summary>Former results (click to open)</summary>
    
  
(moved here from #1999's comments)

Updated TPC-H performance comparison of the **master** branch (d2571500dc039dd0e7a483ec2ea6e43b5b381f9d) vs. our current PR state (1abb77e6380416827d7a8fc3f41685d14dac9d6c):

```diff
 +----------------+--------------------+------+--------------------+------+------------+---------+
 | Benchmark      | prev. iter/s       | runs | new iter/s         | runs | change [%] | p-value |
 +----------------+--------------------+------+--------------------+------+------------+---------+
 | TPC-H 01       | 0.6615901589393616 | 40   | 0.6531713604927063 | 40   | -1%        |  0.2005 |
+| TPC-H 02       | 40.73917007446289  | 2445 | 46.932132720947266 | 2816 | +15%       |  0.0000 |
 | TPC-H 03       | 4.546040058135986  | 273  | 4.715278148651123  | 284  | +4%        |  0.0000 |
 | TPC-H 04       | 4.210179328918457  | 253  | 4.208194732666016  | 253  | -0%        |  0.9496 |
+| TPC-H 05       | 2.4139575958251953 | 145  | 2.5666661262512207 | 154  | +6%        |  0.0000 |
 | TPC-H 06       | 115.94351959228516 | 6957 | 118.49485778808594 | 7110 | +2%        |  0.0000 |
 | TPC-H 07       | 5.204839706420898  | 313  | 5.2055439949035645 | 313  | +0%        |  0.9856 |
 | TPC-H 08       | 3.62990665435791   | 218  | 3.4964518547058105 | 210  | -4%        |  0.0000 |
 | TPC-H 09       | 1.1806957721710205 | 71   | 1.1793214082717896 | 71   | -0%        |  0.8850 |
 | TPC-H 10       | 1.8862420320510864 | 114  | 1.847568154335022  | 111  | -2%        |  0.0684 |
 | TPC-H 11       | 18.202131271362305 | 1093 | 18.234140396118164 | 1095 | +0%        |  0.4789 |
 | TPC-H 12       | 9.468599319458008  | 569  | 9.594390869140625  | 576  | +1%        |  0.0000 |
-| TPC-H 13       | 1.5981374979019165 | 96   | 1.4652011394500732 | 88   | -8%        |  0.0000 |
-| TPC-H 14       | 31.862634658813477 | 1912 | 29.74830436706543  | 1785 | -7%        |  0.0000 |
 | TPC-H 15       | 62.11576843261719  | 3727 | 61.88772201538086  | 3714 | -0%        |  0.1134 |
+| TPC-H 16       | 4.856360912322998  | 292  | 5.617599964141846  | 338  | +16%       |  0.0000 |
+| TPC-H 17       | 2.6693525314331055 | 161  | 2.825106143951416  | 170  | +6%        |  0.0000 |
-| TPC-H 18       | 0.7827396392822266 | 47   | 0.7413405179977417 | 45   | -5%        |  0.0205 |
 | TPC-H 19       | 4.921579837799072  | 296  | 4.923618793487549  | 296  | +0%        |  0.9590 |
 | TPC-H 20       | 16.069896697998047 | 965  | 16.119413375854492 | 968  | +0%        |  0.5094 |
 | TPC-H 21       | 0.7259087562561035 | 44   | 0.7365230917930603 | 45   | +1%        |  0.2648 |
 | TPC-H 22       | 8.676556587219238  | 521  | 8.752893447875977  | 526  | +1%        |  0.0503 |
 | geometric mean |                    |      |                    |      | +1%        |         |
 +----------------+--------------------+------+--------------------+------+------------+---------+
```
</details>

TPC-H performance comparison of the master branch (acac68813572a4c929855de7a717e4870eee0242) to this PR's state (f36a90c763ec4733f291da4f09c8958cb7f550fd):

```diff
 +----------------+--------------------+------+--------------------+------+------------+---------+
 | Benchmark      | prev. iter/s       | runs | new iter/s         | runs | change [%] | p-value |
 +----------------+--------------------+------+--------------------+------+------------+---------+
 | TPC-H 01       | 0.671251654624939  | 41   | 0.6837134957313538 | 42   | +2%        |  0.0045 |
+| TPC-H 02       | 51.17995834350586  | 3071 | 53.962589263916016 | 3238 | +5%        |  0.0000 |
 | TPC-H 03       | 5.554073333740234  | 334  | 5.605419635772705  | 337  | +1%        |  0.1193 |
 | TPC-H 04       | 4.674902439117432  | 281  | 4.715536594390869  | 283  | +1%        |  0.1910 |
+| TPC-H 05       | 2.6141672134399414 | 157  | 2.807044506072998  | 169  | +7%        |  0.0000 |
 | TPC-H 06       | 139.84872436523438 | 8391 | 137.7901153564453  | 8268 | -1%        |  0.0000 |
 | TPC-H 07       | 6.158907413482666  | 370  | 6.113049507141113  | 367  | -1%        |  0.2866 |
 | TPC-H 08       | 3.9812259674072266 | 239  | 4.041755199432373  | 243  | +2%        |  0.0189 |
 | TPC-H 09       | 1.3156601190567017 | 79   | 1.3245304822921753 | 80   | +1%        |  0.2410 |
 | TPC-H 10       | 2.510338068008423  | 151  | 2.519153356552124  | 152  | +0%        |  0.6903 |
 | TPC-H 11       | 22.485443115234375 | 1350 | 22.502519607543945 | 1351 | +0%        |  0.7085 |
 | TPC-H 12       | 10.463584899902344 | 628  | 10.217452049255371 | 614  | -2%        |  0.0000 |
 | TPC-H 13       | 1.651613473892212  | 100  | 1.6560580730438232 | 100  | +0%        |  0.6203 |
 | TPC-H 14       | 36.24421691894531  | 2175 | 35.97789764404297  | 2159 | -1%        |  0.0000 |
 | TPC-H 15       | 69.62496948242188  | 4178 | 67.10285949707031  | 4027 | -4%        |  0.0000 |
+| TPC-H 16       | 6.1179094314575195 | 368  | 6.676680564880371  | 401  | +9%        |  0.0000 |
 | TPC-H 17       | 3.0184719562530518 | 182  | 3.0089809894561768 | 181  | -0%        |  0.6901 |
-| TPC-H 18       | 0.8787471652030945 | 53   | 0.7270537614822388 | 44   | -17%       |  0.0000 |
 | TPC-H 19       | 6.065253734588623  | 364  | 6.153009414672852  | 370  | +1%        |  0.0634 |
 | TPC-H 20       | 20.36804962158203  | 1223 | 19.955921173095703 | 1198 | -2%        |  0.0000 |
 | TPC-H 21       | 0.8674397468566895 | 53   | 0.8775533437728882 | 53   | +1%        |  0.3617 |
 | TPC-H 22       | 9.470969200134277  | 569  | 9.851902961730957  | 592  | +4%        |  0.0000 |
 | geometric mean |                    |      |                    |      | +0%        |         |
 +----------------+--------------------+------+--------------------+------+------------+---------+
```


mrks:
What's with #1999? Does this supersede the previous PR?

bakoe:
@mrks Yes, #1999 is superseded/replaced by this PR. Sorry for not yet having closed #1999.

Based on [this comment from @tobodner to #1999](https://github.com/hyrise/hyrise/pull/1999#issuecomment-581820279) and to make our PRs easier to digest, we planned to structure our PRs as follows:

- Adjustments to the SyntheticTableGenerator (PR #1987 already created and reviewed)
- Extension of Sort Operator and LQP Translator to allow multi column sorting in one operator instance (without micro benchmark adjustments) (already extracted from #1999 to this PR – #2005)
- Extension of the Sort operator’s micro benchmarks incl. NULL ratios created by the updated SyntheticTableGenerator and multi column benchmarks utilizing the updated Sort operator interface from #2005 (to-be-extracted from #1999) 
- Enabling of parallelization in the Sort operator (to-be-created from our fork’s corresponding branch) 

This would result in a total of 4 PRs (of which one has already been successfully reviewed) which – although depending on each other in the stated order and thus having to be reviewed in this order – we hope would allow reviewers to focus better on each of the individual changes we did (which are “semantically” quite Independent from each other). 

Does this make sense?

birneamstiel:
@tobodner the **full ci** pipline has been pending for about 12 hours now, is this an expected waiting time or should I try retriggering it? Any suggestions?

mrks:
It's not stuck, it's green. You need to set the FullCI tag if your PR is complete enough to go through the entire CI Pipeline. Just did so and restarted the last run.

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,78 @@ const std::string& Sort::name() const {
...
+    auto sort_definition = _sort_definitions[column_id];
+    _sort_definition_data_types[column_id] = input_table_left()->column_data_type(sort_definition.column);
+  }
+
+  std::shared_ptr<const Table> output = input_table_left();
+  std::shared_ptr<PosList> previously_sorted_pos_list = std::shared_ptr<PosList>(nullptr);
+
+  _validate_sort_definitions();
+
+  for (auto column_id = _sort_definitions.size(); column_id-- != 0;) {
```
j-tr: this might be hard to read, maybe better `... ; column_id >= 0; --column_id) ...`

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,78 @@ const std::string& Sort::name() const {
...
+
+        // Set the ordered_by attribute of the output's chunks to the column the chunk was sorted by last (because this
+        // is the column that the chunk is ordered by in any case; whereas it is not necessarily sorted by the other
+        // columns -- on their own -- as well).
+        const auto chunk_count = output_mutable->chunk_count();
+        for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+          output_mutable->get_chunk(chunk_id)->set_ordered_by(
+              std::make_pair(sort_definition.column, sort_definition.order_by_mode));
+        }
+        output = output_mutable;
```
j-tr: would if be possible to move this part at the very end of the function outside of the for loop? this way, we wouldnt need is_last_sorting_run anymore and sorting and merialization would be logically more seperated

src/lib/operators/sort.cpp
```diff
@@ -200,41 +244,69 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
-    }
-
-    return output;
+    return _row_id_value_vector;
   }
 
+ protected:
   // completely materializes the sort column to create a vector of RowID-Value pairs
-  void _materialize_sort_column() {
+  void _materialize_sort_column(const std::shared_ptr<PosList>& pos_list = nullptr) {
```
j-tr: do we really need the nullptr default value here?

src/lib/operators/sort.cpp
```diff
@@ -200,41 +244,69 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
+        if (accessor) {
+          const auto typed_value = accessor->access(chunk_offset);
+          if (!typed_value) {
+            null_value_rows.emplace_back(row_id, SortColumnType{});
+          } else {
+            row_id_value_vector.emplace_back(row_id, typed_value.value());
+          }
         } else {
-          row_id_value_vector.emplace_back(RowID{chunk_id, position.chunk_offset()}, position.value());
+          const auto value = (*base_segment)[chunk_offset];
```
j-tr: i'm not sure of the real performance impact of the [] operator but maybe it could be avoided by iterating over the segment and check if the offset is in the pos list. I think the performance somehow depends on how much of the segment is covered by the poslist.

src/lib/operators/sort.hpp
```diff
@@ -38,17 +47,19 @@ class Sort : public AbstractReadOnlyOperator {
...
   template <typename SortColumnType>
   class SortImpl;
+
   template <typename SortColumnType>
   class SortImplMaterializeOutput;
 
-  std::unique_ptr<AbstractReadOnlyOperatorImpl> _impl;
-  const ColumnID _column_id;
-  const OrderByMode _order_by_mode;
+  std::shared_ptr<const Table> _get_materialized_output(const std::shared_ptr<PosList>& pos_list = nullptr);
```
j-tr: do we need the default parameter here?

src/lib/operators/sort.hpp
```diff
@@ -15,18 +15,27 @@
 namespace opossum {
 
 /**
- * Operator to sort a table by a single column. This implements a stable sort, i.e., rows that share the same value will
- * maintain their relative order.
- * Multi-column sort is not supported yet. For now, you will have to sort by the secondary criterion, then by the first
+ * Defines in which order a certain column should be sorted.
+ */
+struct SortColumnDefinition final {
```
MasterCarl: This is only a minor issue, but I think there might be a better name than _Definition_, e.g. `ColumnSortParameter`.

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,78 @@ const std::string& Sort::name() const {
...
 std::shared_ptr<const Table> Sort::_on_execute() {
-  _impl = make_unique_by_data_type<AbstractReadOnlyOperatorImpl, SortImpl>(
-      input_table_left()->column_data_type(_column_id), input_table_left(), _column_id, _order_by_mode,
-      _output_chunk_size);
-  return _impl->_on_execute();
+  _sort_definition_data_types.resize(_sort_definitions.size());
+  for (auto column_id = ColumnID{0}; column_id < _sort_definitions.size(); ++column_id) {
+    auto sort_definition = _sort_definitions[column_id];
+    _sort_definition_data_types[column_id] = input_table_left()->column_data_type(sort_definition.column);
+  }
```
MasterCarl: You might use `std::transform` here to make the code more readable.

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,78 @@ const std::string& Sort::name() const {
...
+    DataType data_type = _sort_definition_data_types[column_id];
+
+    resolve_data_type(data_type, [&](auto type) {
+      using ColumnDataType = typename decltype(type)::type;
+
+      auto sort_single_col_impl = SortImpl<ColumnDataType>(input_table_left(), sort_definition.column,
+                                                           sort_definition.order_by_mode, _output_chunk_size);
+      const auto row_id_value_vector = sort_single_col_impl.sort_one_column(previously_sorted_pos_list);
+
+      // 3. Materialization of the result: We take the sorted ValueRowID Vector, create chunks fill them until they are
```
MasterCarl: I can't find points 1 and 2. Also needs a comma after create chunks.

MasterCarl: Please also use consistent naming between ValueRowID and row_id_value_vector.

src/lib/operators/sort.cpp
```diff
@@ -200,41 +244,69 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
     row_id_value_vector.reserve(_table_in->row_count());
 
     auto& null_value_rows = *_null_value_rows;
 
-    const auto chunk_count = _table_in->chunk_count();
-    for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
-      const auto chunk = _table_in->get_chunk(chunk_id);
-      Assert(chunk, "Physically deleted chunk should not reach this point, see get_chunk / #1686.");
+    if (pos_list) {
+      auto segment_ptr_and_accessor_by_chunk_id =
```
MasterCarl: It took me a while to realize this structure is actually written to. Maybe add a comment as to what purpose it serves and how its contents are updated.

src/lib/operators/sort.cpp
```diff
@@ -200,41 +244,69 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
+            row_id_value_vector.emplace_back(row_id, boost::get<SortColumnType>(value));
+          }
         }
-      });
+      }
+    } else {
+      const auto chunk_count = _table_in->chunk_count();
+      for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
+        const auto chunk = _table_in->get_chunk(chunk_id);
+        Assert(chunk, "Did not expect deleted chunk here.");  // see #1686
```
MasterCarl: Maybe add full URL https://github.com/hyrise/hyrise/pull/1686

src/lib/operators/sort.cpp
birneamstiel: Yep you're right, increases the logical structure... It's not possible to completely eliminate the check for `is_last_sorting_run` here since we need the `ColumnDataType` for the materialization.

Moved the setting of `ordered_by` to the end of this method in 8293aeaa193ad73bce0707dc5fd34dab197ef6ec

src/lib/operators/sort.cpp
birneamstiel: We don't! fixed in 95a08c007e269e5beafd61e6f4c707621eac7003

src/lib/operators/sort.hpp
birneamstiel: no, fixed in 95a08c007e269e5beafd61e6f4c707621eac7003

src/lib/operators/sort.hpp
birneamstiel: Hmm actually I think _Definition_ is a better fit here since it is indeed defining how the column should be sorted... other opinions?

src/lib/operators/sort.cpp
birneamstiel: Good idea, thank you! Done in 40bc0bf1d6c66f646268bb8d1f601cf33e21bd44

src/lib/operators/sort.cpp
birneamstiel: fixed in 32efd0affe63408e18085d1837f40091cb078bce

src/lib/operators/sort.cpp
birneamstiel: Done in f6ea5af89bdb12244547fb8a313f682459053c31

src/lib/operators/sort.cpp
```diff
@@ -200,41 +244,69 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
     auto& row_id_value_vector = *_row_id_value_vector;
     row_id_value_vector.reserve(_table_in->row_count());
 
     auto& null_value_rows = *_null_value_rows;
 
-    const auto chunk_count = _table_in->chunk_count();
-    for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
-      const auto chunk = _table_in->get_chunk(chunk_id);
-      Assert(chunk, "Physically deleted chunk should not reach this point, see get_chunk / #1686.");
+    if (pos_list) {
```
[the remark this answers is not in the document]
birneamstiel: Done in f6ea5af89bdb12244547fb8a313f682459053c31

src/lib/operators/sort.cpp
j-tr: wouldn't it be possible to have another call or resolve_data_type outside the loop?

src/lib/operators/sort.hpp
```diff
@@ -38,17 +47,19 @@ class Sort : public AbstractReadOnlyOperator {
...
   template <typename SortColumnType>
   class SortImpl;
+
   template <typename SortColumnType>
   class SortImplMaterializeOutput;
 
-  std::unique_ptr<AbstractReadOnlyOperatorImpl> _impl;
-  const ColumnID _column_id;
-  const OrderByMode _order_by_mode;
+  std::shared_ptr<const Table> _get_materialized_output(const std::shared_ptr<PosList>& pos_list);
```
mrks: "get" sounds like a cheap step. Maybe "materialize_output"?

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,77 @@ const std::string& Sort::name() const {
...
-      _output_chunk_size);
-  return _impl->_on_execute();
+  _sort_definition_data_types.resize(_sort_definitions.size());
+  std::transform(
+      _sort_definitions.begin(), _sort_definitions.end(), _sort_definition_data_types.begin(),
+      [this](const auto sort_definition) { return input_table_left()->column_data_type(sort_definition.column); });
+
+  std::shared_ptr<const Table> output = input_table_left();
+  std::shared_ptr<Table> output_mutable;
+  std::shared_ptr<PosList> previously_sorted_pos_list = std::shared_ptr<PosList>(nullptr);
```
mrks: Just sorted_pos_list?

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,77 @@ const std::string& Sort::name() const {
...
-void Sort::_on_cleanup() { _impl.reset(); }
+void Sort::_on_cleanup() {}
+
+/**
+ * Asserts that all column definitions are valid.
+ */
+void Sort::_validate_sort_definitions() const {
+  const auto input_table = input_table_left();
+  for (const auto& column_sort_definition : _sort_definitions) {
+    Assert(column_sort_definition.column != INVALID_COLUMN_ID, "Sort: Invalid column in sort definition");
```
mrks: Checking that the column id is within input_table->column_count() would be even safer.

src/lib/operators/sort.cpp
```diff
@@ -200,41 +243,72 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
-    }
-
-    return output;
+    return _row_id_value_vector;
   }
 
+ protected:
   // completely materializes the sort column to create a vector of RowID-Value pairs
-  void _materialize_sort_column() {
+  void _materialize_sort_column(const std::shared_ptr<PosList>& pos_list) {
```
mrks: Please make the name describe what the parameter does, not what type it is.

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,77 @@ const std::string& Sort::name() const {
...
+                                                           sort_definition.order_by_mode, _output_chunk_size);
+      const auto row_id_value_vector = sort_single_col_impl.sort_one_column(previously_sorted_pos_list);
+
+      if (is_last_sorting_run) {
+        auto materialization = std::make_shared<SortImplMaterializeOutput<ColumnDataType>>(
+            input_table_left(), row_id_value_vector, _output_chunk_size);
+
+        output_mutable = materialization->execute();
+      } else {
+        previously_sorted_pos_list.reset();
```
mrks: reset is redundant if you overwrite it

src/lib/operators/sort.cpp
```diff
@@ -177,10 +221,9 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
     _null_value_rows = std::make_shared<std::vector<RowIDValuePair>>();
   }
 
- protected:
-  std::shared_ptr<const Table> _on_execute() override {
+  std::shared_ptr<std::vector<RowIDValuePair>> sort_one_column(const std::shared_ptr<PosList>& pos_list = nullptr) {
```
mrks: Please describe what the parameter does, not what type it is

src/lib/operators/sort.cpp
mrks: It might be easier to understand the effects of this method if, instead of modifying the object, it would be static and simply returned the materialized column. While it was this way before, changing this should only take a handful of lines.

src/lib/operators/sort.cpp
```diff
@@ -200,41 +243,72 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
+                                                std::shared_ptr<AbstractSegmentAccessor<SortColumnType>>>>();
+      segment_ptr_and_accessor_by_chunk_id.reserve(_table_in->chunk_count());
+
+      // When there was a preceding sorting run, we materialize according to the passed PosList.
+      for (RowID row_id : *pos_list) {
+        const auto [chunk_id, chunk_offset] = row_id;
+
+        auto& segment_ptr_and_typed_ptr_pair = segment_ptr_and_accessor_by_chunk_id[chunk_id];
+        auto& base_segment = segment_ptr_and_typed_ptr_pair.first;
+        auto& accessor = segment_ptr_and_typed_ptr_pair.second;
```
mrks: structured bindings?

src/lib/operators/sort.cpp
```diff
@@ -200,41 +243,72 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
-          null_value_rows.emplace_back(RowID{chunk_id, position.chunk_offset()}, SortColumnType{});
+        // If the input segment is not a ReferenceSegment, we can take a fast(er) path
+        if (accessor) {
+          const auto typed_value = accessor->access(chunk_offset);
+          if (!typed_value) {
+            null_value_rows.emplace_back(row_id, SortColumnType{});
+          } else {
+            row_id_value_vector.emplace_back(row_id, typed_value.value());
+          }
         } else {
```
mrks: Is the else path taken? I would expect create_segment_accessor to always return something

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,77 @@ const std::string& Sort::name() const {
...
+    output_mutable->get_chunk(chunk_id)->set_ordered_by(
+        std::make_pair(final_sort_definition.column, final_sort_definition.order_by_mode));
+  }
+  output = output_mutable;
+
+  return output;
 }
 
-void Sort::_on_cleanup() { _impl.reset(); }
+void Sort::_on_cleanup() {}
```
mrks: Why implement it if it is empty?

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,77 @@ const std::string& Sort::name() const {
...
 std::shared_ptr<const Table> Sort::_on_execute() {
-  _impl = make_unique_by_data_type<AbstractReadOnlyOperatorImpl, SortImpl>(
-      input_table_left()->column_data_type(_column_id), input_table_left(), _column_id, _order_by_mode,
-      _output_chunk_size);
-  return _impl->_on_execute();
+  _sort_definition_data_types.resize(_sort_definitions.size());
+  std::transform(
+      _sort_definitions.begin(), _sort_definitions.end(), _sort_definition_data_types.begin(),
+      [this](const auto sort_definition) { return input_table_left()->column_data_type(sort_definition.column); });
+
```
mrks: I'm not a big fan of transform. Simply iterating over the column ids and writing into the _sort_definitions vector would probably be more intuitive and might even be shorter.

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,77 @@ const std::string& Sort::name() const {
...
+  std::shared_ptr<const Table> output = input_table_left();
```
mrks: output = input? Why?

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,77 @@ const std::string& Sort::name() const {
...
+  for (auto column_id = _sort_definitions.size(); column_id-- != 0;) {
+    const bool is_last_sorting_run = (column_id == 0);
+
+    auto sort_definition = _sort_definitions[column_id];
+    DataType data_type = _sort_definition_data_types[column_id];
+
+    resolve_data_type(data_type, [&](auto type) {
+      using ColumnDataType = typename decltype(type)::type;
+
+      auto sort_single_col_impl = SortImpl<ColumnDataType>(input_table_left(), sort_definition.column,
```
mrks: s/col/column/

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,77 @@ const std::string& Sort::name() const {
...
+            input_table_left(), row_id_value_vector, _output_chunk_size);
+
+        output_mutable = materialization->execute();
+      } else {
+        previously_sorted_pos_list.reset();
+        previously_sorted_pos_list = std::make_shared<PosList>();
+
+        for (auto& row_id_value : *row_id_value_vector) {
+          previously_sorted_pos_list->emplace_back(row_id_value.first);
+        }
```
mrks: reserve?

src/lib/operators/sort.cpp
```diff
@@ -200,41 +200,76 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
-      Assert(chunk, "Physically deleted chunk should not reach this point, see get_chunk / #1686.");
+    if (pos_list) {
+      auto segment_ptr_and_accessor_by_chunk_id =
+          std::unordered_map<ChunkID, std::pair<std::shared_ptr<const BaseSegment>,
+                                                std::shared_ptr<AbstractSegmentAccessor<SortColumnType>>>>();
+      segment_ptr_and_accessor_by_chunk_id.reserve(_table_in->chunk_count());
 
-      auto base_segment = chunk->get_segment(_column_id);
+      // When there was a preceding sorting run, we materialize according to the passed PosList.
+      for (RowID row_id : *pos_list) {
```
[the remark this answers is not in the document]
birneamstiel: Since we're using row_id (as RowID) further down the method I'd keep it like that...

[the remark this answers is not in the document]
mrks: Yepp, must have missed that.

src/lib/operators/sort.cpp
birneamstiel: Hm not sure what you're aiming at, is ist about the naming? So sort_s_column_impl would be better from your point of view?

src/lib/operators/sort.cpp
birneamstiel: So basically like we had it before? https://github.com/hyrise/hyrise/pull/2005#discussion_r379437147

src/test/operators/sort_test.cpp
```diff
@@ -3,6 +3,7 @@
 #include <utility>
 
 #include "base_test.hpp"
+#include "gtest/gtest.h"
```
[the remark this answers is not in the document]
birneamstiel: removed in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
```diff
@@ -33,136 +20,150 @@ const std::string& Sort::name() const {
...
-  // creates a new table with reference segments
-  SortImplMaterializeOutput(const std::shared_ptr<const Table>& in,
-                            const std::shared_ptr<std::vector<std::pair<RowID, SortColumnType>>>& id_value_map,
-                            const size_t output_chunk_size)
-      : _table_in(in), _output_chunk_size(output_chunk_size), _row_id_value_vector(id_value_map) {}
+  _validate_sort_definitions();
+
+  for (auto column_id = _sort_definitions.size(); column_id-- != 0;) {
+    auto sort_definition = _sort_definitions[column_id];
+    DataType data_type = _sort_definition_data_types[column_id];
```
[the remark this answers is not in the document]
birneamstiel: Done in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
```diff
@@ -200,41 +200,76 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
     auto& row_id_value_vector = *_row_id_value_vector;
     row_id_value_vector.reserve(_table_in->row_count());
 
     auto& null_value_rows = *_null_value_rows;
 
-    const auto chunk_count = _table_in->chunk_count();
-    for (ChunkID chunk_id{0}; chunk_id < chunk_count; ++chunk_id) {
-      const auto chunk = _table_in->get_chunk(chunk_id);
-      Assert(chunk, "Physically deleted chunk should not reach this point, see get_chunk / #1686.");
+    if (pos_list) {
```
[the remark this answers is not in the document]
birneamstiel: I'm not really sure if I understood correctly. Are you saying that every branch should be in a separate method and the decision/condition should happen before they are called?
I refactored this a bit, basically I just factored out the if/then block... see 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
```diff
@@ -200,41 +200,76 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
+      auto segment_ptr_and_accessor_by_chunk_id =
+          std::unordered_map<ChunkID, std::pair<std::shared_ptr<const BaseSegment>,
+                                                std::shared_ptr<AbstractSegmentAccessor<SortColumnType>>>>();
+      segment_ptr_and_accessor_by_chunk_id.reserve(_table_in->chunk_count());
 
-      auto base_segment = chunk->get_segment(_column_id);
+      // When there was a preceding sorting run, we materialize according to the passed PosList.
```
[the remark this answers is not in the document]
birneamstiel: Done in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.hpp
birneamstiel: Fixed in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
birneamstiel: Implemented in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
birneamstiel: Done in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
birneamstiel: Changed in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
birneamstiel: Removed in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
birneamstiel: Done in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
birneamstiel: Done in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
birneamstiel: Removed in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
birneamstiel: Done in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.hpp
```diff
@@ -15,18 +15,27 @@
...
+ * value will maintain their relative order.
+ * By passing multiple sort column definitions it is possible to sort multiple columns with one operator run.
  */
 class Sort : public AbstractReadOnlyOperator {
  public:
-  // The parameter chunk_size sets the chunk size of the output table, which will always be materialized
-  Sort(const std::shared_ptr<const AbstractOperator>& in, const ColumnID column_id,
-       const OrderByMode order_by_mode = OrderByMode::Ascending, const size_t output_chunk_size = Chunk::DEFAULT_SIZE);
+  Sort(const std::shared_ptr<const AbstractOperator>& in, const std::vector<SortColumnDefinition>& sort_definitions,
+       size_t output_chunk_size = Chunk::DEFAULT_SIZE);
```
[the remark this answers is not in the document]
birneamstiel: Done in 2e4a1e739bcc9286a783cba00336ca8e61325125

src/lib/operators/sort.cpp
```diff
@@ -33,136 +20,150 @@ const std::string& Sort::name() const {
...
-// This class fulfills only the materialization task for a sorted row_id_value_vector.
-template <typename SortColumnType>
-class Sort::SortImplMaterializeOutput {
- public:
-  // creates a new table with reference segments
-  SortImplMaterializeOutput(const std::shared_ptr<const Table>& in,
-                            const std::shared_ptr<std::vector<std::pair<RowID, SortColumnType>>>& id_value_map,
-                            const size_t output_chunk_size)
-      : _table_in(in), _output_chunk_size(output_chunk_size), _row_id_value_vector(id_value_map) {}
+  _validate_sort_definitions();
```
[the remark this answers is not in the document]
birneamstiel: Done in c48f79e89e8b8861b6557dfe19f30e784c72d37a

src/lib/operators/sort.cpp
bakoe: IIRC, this is there in order to just output the input table _as is_ when the Sort operator is used with an empty `SortDefinition` struct vector as its input configuration (i. e. when there are not columns which the input should be sorted by). Does that make sense? Probably we should add a corresponding comment to the line then … 🙃

src/test/operators/sort_test.cpp
```diff
@@ -107,7 +114,9 @@ TEST_P(OperatorsSortTest, DoubleSortOfOneColumn) {
 TEST_P(OperatorsSortTest, DescendingSortOfOneColumn) {
   std::shared_ptr<Table> expected_result = load_table("resources/test_data/tbl/int_float_reverse.tbl", 2);
 
-  auto sort = std::make_shared<Sort>(_table_wrapper, ColumnID{0}, OrderByMode::Descending, 2u);
+  auto sort = std::make_shared<Sort>(
+      _table_wrapper, std::vector<SortColumnDefinition>{SortColumnDefinition{ColumnID{0}, OrderByMode::Descending}},
```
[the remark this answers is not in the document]
bakoe: I think it actually is necessary because of the `explicit`-ness of the `SortColumnDefinition` constructor.

Our way to this code was as follows:
- The previously [non-explicit constructor](https://github.com/bakoe/hyrise/blob/80232da1c62a902383ead90dd5678726f2241d52/src/lib/operators/sort.hpp#L21-L22) (and thus omitting of `std::vector<SortColumnDefinition>` in its usage, as proposed by you) [lead to an error in the Lint stage of the CI pipeline](https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-2005/4/pipeline#step-51-log-2) (`runtime/explicit `)
- Thus, [we made the `SortColumnDefinition` constructor `explicit`](https://github.com/hyrise/hyrise/commit/e220caf3e6fbc969559f991036e3a616c63a210f#diff-f0417e0e36c6f2046bc9907546366527L21-R21). This [lead to an error in the gccDebug stage of the CI pipeline](https://github.com/hyrise/hyrise/commit/e220caf3e6fbc969559f991036e3a616c63a210f#diff-f0417e0e36c6f2046bc9907546366527L21-R21) (`converting to 'opossum::SortColumnDefinition' from initializer list would use explicit constructor `)
- Thus, we made the usage of the `SortColumnDefinition` constructor explicit in 9937bc5750afb6a8d601465b7a9321b8ac8a5970.. 4b16c695a4bb701cadead86e35612d71ee28391f, leading to the existence of `std::vector<SortColumnDefinition>` here and in other places. 

Does this make sense?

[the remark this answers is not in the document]
mrks: Hmpf. I would have thought that one `SortColumnDefinition` should be sufficient to satisfy the explicitness, but apparently not. Thanks for the explanation!

src/lib/operators/sort.cpp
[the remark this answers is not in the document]
mrks: Yepp. This is a bit of a balancing act and we do not want to encourage a bunch of microfunctions.

src/lib/operators/sort.cpp
```diff
@@ -200,41 +200,76 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
+        }
 
-      segment_iterate<SortColumnType>(*base_segment, [&](const auto& position) {
-        if (position.is_null()) {
-          null_value_rows.emplace_back(RowID{chunk_id, position.chunk_offset()}, SortColumnType{});
+        // If the input segment is not a ReferenceSegment, we can take a fast(er) path
+        if (accessor) {
+          const auto typed_value = accessor->access(chunk_offset);
+          if (!typed_value) {
+            null_value_rows.emplace_back(row_id, SortColumnType{});
```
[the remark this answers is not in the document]
mrks: On second thought, since the number of null values is unknown, we don't have a good guess. This would have been different in cases where we use a `vector<bool>` to store null values.

src/lib/operators/sort.cpp
mrks: Hmpf. Didn't see that. Another example of code reviews not being an exact science. :/

Personally, I don't see a big advantage on the side of `std::transform` (other than std::execution, which we won't use anytime soon). On the other hand, I find it more natural to read the `for` loop: "Iterate over all columns and write them to this vector". I'm done with reading the for loop by the time it takes me to recall the order of arguments for `std::transform` and I suspect that upcoming student generations, who are not familiar with the STL, will feel the same.

src/lib/operators/sort.cpp
mrks: Can you come up with a case where this would be legal? If someone creates a sort operator but does not declare what the order should be, the constructor (not _validate_sort_definitions, which is called much later) should just go up in flames.

src/lib/operators/sort.cpp
mrks: My fault. Shouldn't use [sed syntax](https://unix.stackexchange.com/questions/159367/using-sed-to-find-and-replace) in comments. "Replace col with column" would have been not that much worse to type. ;)

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,76 @@ const std::string& Sort::name() const {
...
+    });
+  }
+
+  auto final_sort_definition = _sort_definitions[0];
+  // Set the ordered_by attribute of the output's chunks to the column the chunk was sorted by last (because this
+  // is the column that the chunk is ordered by in any case; whereas it is not necessarily sorted by the other
+  // columns -- on their own -- as well).
+  const auto chunk_count = output_mutable->chunk_count();
+  for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+    output_mutable->get_chunk(chunk_id)->set_ordered_by(
```
Boffmann:
```suggestion
    const auto chunk = output->get_chunk(chunk_id);
    if (chunk->is_mutable()) {
        chunk->finalize();
    }
    chunk->set_ordered_by(
```

Hint: We used `chunk->finalize` whenever the `ordered_by` property is set to ensure that the chunk is always ordered. The performance impacts on this are still measured, though.

src/lib/operators/sort.cpp
```diff
@@ -200,41 +243,74 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
+      const auto [chunk_id, chunk_offset] = row_id;
+
+      auto& [base_segment, accessor] = segment_ptr_and_accessor_by_chunk_id[chunk_id];
+
+      if (!base_segment) {
+        base_segment = _table_in->get_chunk(chunk_id)->get_segment(_column_id);
+        accessor = create_segment_accessor<SortColumnType>(base_segment);
+      }
+
+      // If the input segment is not a ReferenceSegment, we can take a fast(er) path
```
Boffmann:
```suggestion
      // If the input segment is not a ReferenceSegment, we can take a fast(er) path.
```

src/lib/operators/sort.cpp
```diff
@@ -200,41 +243,74 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
+      if (accessor) {
+        const auto typed_value = accessor->access(chunk_offset);
+        if (!typed_value) {
+          (*_null_value_rows).emplace_back(row_id, SortColumnType{});
```
Boffmann:
```suggestion
          _null_value_rows->emplace_back(row_id, SortColumnType{});
```

src/lib/operators/sort.cpp
```diff
@@ -200,41 +243,74 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
         } else {
-          row_id_value_vector.emplace_back(RowID{chunk_id, position.chunk_offset()}, position.value());
+          (*_row_id_value_vector).emplace_back(row_id, typed_value.value());
```
Boffmann:
```suggestion
          _row_id_value_vector->emplace_back(row_id, typed_value.value());
```

src/lib/operators/sort.cpp
```diff
@@ -200,41 +243,74 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
         }
-      });
+      } else {
+        const auto value = (*base_segment)[chunk_offset];
+        if (variant_is_null(value)) {
+          (*_null_value_rows).emplace_back(row_id, SortColumnType{});
```
Boffmann:
```suggestion
          _null_value_rows->emplace_back(row_id, SortColumnType{});
```

src/lib/operators/sort.cpp
```diff
@@ -200,41 +243,74 @@ class Sort::SortImpl : public AbstractReadOnlyOperatorImpl {
...
+        } else {
+          (*_row_id_value_vector).emplace_back(row_id, boost::get<SortColumnType>(value));
```
Boffmann:
```suggestion
          _row_id_value_vector->emplace_back(row_id, boost::get<SortColumnType>(value));
```

mrks:
" +5,111 −2,296 " - Not sure what happened here. Superseded by #2064

Issue #2003, opened 2020-02-03
Dynamically Load Plug-Ins and Persist Settings


dey4ss:
As we can load/unload plug-ins with the Console, we would like to also be able to do so via the server. The suggested approach is to do this with a new meta table: SELECT shows the loaded plug-ins, INSERT loads new ones and DELETE unloads them.
Since plugins (and other components) may have settings, we would like to have a central point for showing and changing those settings. This can also be done via a meta table.

Steps needed are:
- [x] add new meta tables `meta_plugins` and `meta_settings` in MetaTableManager that show the plugins fetched from the PluginManager/all settings that are configurable in Hyrise
- [x] for each meta table: indicate if an operation is valid
- [x] forward mutations to meta tables
- [x] add settings and a settings manager
- [x] tests

dey4ss:
@Bensk1 @mrks Starting to impelemt what we discussed earlier, some questions came up to my mind:

1. updating/deleting in meta tables: As I know generating some meta tables takes a lot of time, I thought about making these operations not taking too long. When altering those "tables" as usual, we would need to generate the tables again to find the rows that match in the specific function. That is why I suggest to be the first column of such meta tables a key. Then, we would only need to generate the table in the MetaTableManager, find the rows and do the update directly.
2. inputs for update/delete/insert to meta tables: I thought about checking if the values match the column definitions of the meta tables. Let us assume the settings table we discussed. We cannot alter the description there, so we can only update name/value. Should the MetaSettingsTable only accept updates à la ('name', 'value') which does not fit the resulting table or ('name', 'value', 'description') and ignore the description? This also affects wehere we check if the values have the right format; the first version needs to be checked by the implementation, the second one can be checked by the abstract class looking on the column definitions, which are a member variable.

mrks:
1: Not sure if I can follow. What do you mean by making them a key? Hyrise does not really have the notion of keys yet. In any case, the generation of meta tables being expensive is more a problem of the segments_accurate table (where a lot of work is done) and shouldn't be one for the settings table.
2: If I remember correctly, the update operator gets two input tables: References to the rows being updated and the new values. The second table is always complete (i.e., holds all columns), and values that are not modified are copied from the existing table. Thus, you will always get a value for the description. We could check if the description matches the old one if that does not take more than maybe five lines of code, but I won't lose sleep if modifications to the description are simply ignored.

dey4ss:
[...]

PR #2006, opened 2020-02-06
Refactor Meta Tables and add Settings

dey4ss:
refs #2003


mrks:
What's the idea behind this PR? Are you still working on this or are you looking for early feedback?

dey4ss:
> What's the idea behind this PR? Are you still working on this or are you looking for early feedback?

Indeed I think feedback is always a good thing, that's why I wanted this to be open for discussions.

src/lib/utils/meta_table_manager.cpp
```diff
@@ -10,11 +10,18 @@
...
 #include "storage/table.hpp"
 #include "storage/table_column_definition.hpp"
+#include "utils/plugin_manager.hpp"
 
 namespace {
 
 using namespace opossum;  // NOLINT
 
+#ifdef __APPLE__
+#define DYNAMIC_LIBRARY_SUFFIX ".dylib"
```
mrks: Not a fan of macros. I don't know a way around `ifdef __APPLE__`, but we can use a constexpr for the string.

src/lib/utils/meta_table_manager.cpp
```diff
@@ -96,18 +103,53 @@ MetaTableManager::MetaTableManager() {
...
   _methods["segments_accurate"] = &MetaTableManager::generate_accurate_segments_table;
+  _methods["plugins"] = &MetaTableManager::generate_plugins_table;
+
+  //const auto a = std::make_tuple(&MetaTableManager::_insert_into_plugins, &MetaTableManager::_delete_from_plugins);
+  //auto a =  {&MetaTableManager::_insert_into_plugins, &MetaTableManager::_delete_from_plugins};
+  tuple<std::function<std::shared_ptr<Table>(const std::string&)>,
+        std::function<std::shared_ptr<Table>(const std::string&)>>(&MetaTableManager::_insert_into_plugins,
+                                                                   &MetaTableManager::_delete_from_plugins) a;
+
+  _mutating_methods["plugins"] = a;
```
mrks: Change `_methods["..."]` to hold a class that has methods for get, insert, and delete (which may default to something throwing if nothing is implemented)? That way, we can move those classes into separate files and the MetaTableManager itself doesn't have to know anything about plugins.

src/lib/utils/meta_table_manager.cpp
```diff
@@ -229,6 +271,26 @@ std::shared_ptr<Table> MetaTableManager::generate_accurate_segments_table() {
...
+  return output_table;
+}
+
+void MetaTableManager::_insert_into_plugins(const std::string& value) {
+  Hyrise::get().plugin_manager.load_plugin(value + DYNAMIC_LIBRARY_SUFFIX);
+}
+
+void MetaTableManager::_delete_from_plugins(const std::string& value) {
+  Hyrise::get().plugin_manager.unload_plugin(value);
+}
```
mrks: Not sure how we end up here. Insert and delete operate on table. The mechanism for converting a table modification to a method like this should be generic. Things that come to mind:

* Tables have multiple columns. While we only use a single field here, we should plan for that.
* Tables have multiple rows. We should be able to do `UPDATE meta_memory_budgets SET budget = budget * 2 WHERE ...`.

Maybe we should do some brainstorming next Wednesday on what settings we want to set this way. Not necessarily because we need to implement them right now, but because we should make sure that whatever we come up with here doesn't block us in the future.

src/lib/utils/plugin_manager.cpp
```diff
@@ -19,6 +19,19 @@ bool PluginManager::_is_duplicate(const std::unique_ptr<AbstractPlugin>& plugin)
   return false;
 }
 
+const std::vector<PluginName> PluginManager::plugin_names() const {
```
mrks: `loaded_plugins`?

src/lib/utils/meta_table_manager.cpp
mrks: The SQLTranslator could check if a method (e.g., INSERT) is set and throw an input exception if it is not.

Also, we probably should discuss how to handle update. DELETE+INSERT doesn't feel right as we don't have proper MVCC on the meta tables and shouldn't leave the DB in an inconsistent state

dey4ss: > That way, we can move those classes into separate files and the MetaTableManager itself doesn't have to know anything about plugins.

Okay, do you have any suggestions where those classes should be located at best? `utils/meta_table_manipulation`?

> DELETE+INSERT doesn't feel right as we don't have proper MVCC on the meta tables and shouldn't leave the DB in an inconsistent state

MVCC with meta tables is something that also came to my mind. Is this a general thing that should be thought of? I mean, since tey are generated on the fly and will be wrapped in a static table node, does having MVCC here really makes sense?

src/lib/utils/meta_table_manager.cpp
dey4ss: Since next week is the TaC II block week, do you also have some time later on today?

src/lib/utils/meta_table_manager.cpp
mrks: > Okay, do you have any suggestions where those classes should be located at best? utils/meta_table_manipulation?

Just `utils/meta_tables`? Have a class for each meta table and that class either implements modifications or not.

> MVCC with meta tables is something that also came to my mind. Is this a general thing that should be thought of?

Oh god, no. Quite the opposite - I was trying to say that our UPDATE=DELETE+INSERT concept might be a disadvantage here.

src/lib/utils/meta_table_manager.cpp
mrks: Sure. Maybe 13:00, also pinging @Bouncner and @Bensk1 (and anyone else who might care)

Bensk1: Alright!

mrks:
Due to the length of the PR list, I'd suggest to close this while you're working on it and re-open it once you want people to have another look.

dey4ss:
> Due to the length of the PR list, I'd suggest to close this while you're working on it and re-open it once you want people to have another look.

Alright

src/lib/utils/settings_manager.cpp
```diff
@@ -0,0 +1,36 @@
+#include "settings_manager.hpp"
+
+#include "utils/settings/mock_setting.hpp"
+
+namespace opossum {
+
+bool SettingsManager::has_setting(const std::string& name) const { return _settings.count(name); }
+
+void SettingsManager::add(const std::string& name, std::shared_ptr<AbstractSetting> setting) {
```
dey4ss: Something around here seems not to work, maybe because of using the attribute of the setting as the key?

src/lib/utils/settings/abstract_setting.hpp
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * This is an abstract class for all settings objects.
+ *
+ * Settings register itself at the SettingsManager.
+ * Therefore it is neccessary that they are owned by a shared pointer (instantiate with std::make_shared).
+ * Thus, you need to call unenroll() when the setting is not longer needed (it won't get destructed otherwise,
+ * as SettingsManager would still hold a shared pointer to it).
+ *
+ * Settings provide a getter/setter with a string.
+ * When the set(...) method ia called, a setting can invoke a method of its parent that triggers
```
Bensk1: is

src/lib/utils/settings/abstract_setting.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * the immediate appliance of the changed value.
+ *
+ * Settings have a unique name that consists of the name of its parent and the setting
+ * name (e.g. "TaskScheduler.workers")
```
Bensk1: The concept of a parent should probably be introduced above.

src/lib/utils/settings/abstract_setting.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include "hyrise.hpp"
+#include "utils/assert.hpp"
+
+namespace opossum {
+
+AbstractSetting::AbstractSetting(const std::string& init_name) : name(init_name) {}
+
+void AbstractSetting::enroll() {
+  auto name_copy = name;
```
Bensk1: Why do we need the name_copy here?

src/lib/utils/settings_manager.cpp
Bensk1: What exactly is not working?

dey4ss: I sometimes get std::bad_alloc or segfaults when accessing or adding settings.

src/lib/utils/meta_tables/meta_mock_table.hpp
```diff
@@ -0,0 +1,46 @@
+#pragma once
+
+#include "utils/meta_tables/abstract_meta_table.hpp"
+
+namespace opossum {
+
+/**
+ * This is a class that mocks the behavior of a meta table.
+ * It can be useful in tests as it provides how often methods were called.
```
mrks: Does this have to be in `src/lib`?

src/lib/utils/meta_tables/meta_plugins_table.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include <boost/algorithm/string.hpp>
+
+#include "hyrise.hpp"
+
+#ifdef __APPLE__
+constexpr char DYNAMIC_LIBRARY_SUFFIX[] = ".dylib";
+#elif __linux__
+constexpr char DYNAMIC_LIBRARY_SUFFIX[] = ".so";
+#endif
```
mrks: This doesn't seem to be the responsibility of the meta table.

src/lib/utils/meta_tables/meta_plugins_table.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+std::string MetaPluginsTable::get_full_file_name(const std::string& file_name) const {
+  auto extension = std::string{std::filesystem::path{file_name}.extension()};
+  boost::algorithm::to_lower(extension);
+  if (extension == DYNAMIC_LIBRARY_SUFFIX) {
+    return file_name;
+  }
+
+  return file_name + DYNAMIC_LIBRARY_SUFFIX;
+}
```
mrks: Is this the tables responsibility?

src/lib/utils/meta_tables/meta_settings_table.cpp
```diff
@@ -0,0 +1,41 @@
+#include "meta_settings_table.hpp"
+
+#include "hyrise.hpp"
+#include "utils/assert.hpp"
+
+namespace opossum {
+
+MetaSettingsTable::MetaSettingsTable()
+    : AbstractMetaTable(TableColumnDefinitions{{"setting_name", DataType::String, false},
```
mrks: Just name?

src/lib/utils/meta_tables/meta_settings_table.cpp
```diff
@@ -0,0 +1,41 @@
...
+const std::string& MetaSettingsTable::name() const {
+  static const auto name = std::string{"settings"};
+  return name;
+}
+
+bool MetaSettingsTable::can_update() const { return true; }
+
+std::shared_ptr<Table> MetaSettingsTable::_on_generate() const {
+  auto output_table = std::make_shared<Table>(_column_definitions, TableType::Data, std::nullopt, UseMvcc::Yes);
+  const auto setting_names = Hyrise::get().settings_manager.all_settings();
```
mrks: I would have expected `all_settings` to return a map (or similar). Maybe `setting_names`?

src/lib/utils/meta_tables/segment_meta_data.hpp
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include "storage/table.hpp"
+
+namespace opossum {
+
+size_t get_distinct_value_count(const std::shared_ptr<BaseSegment>& segment);
+
+void gather_segment_meta_data(const std::shared_ptr<Table>& meta_table, const MemoryUsageCalculationMode mode);
```
mrks: doc?

src/lib/utils/plugin_manager.hpp
```diff
@@ -26,6 +26,8 @@ class PluginManager : public Noncopyable {
   void load_plugin(const std::filesystem::path& path);
   void unload_plugin(const PluginName& name);
 
+  const std::vector<PluginName> loaded_plugins() const;
```
mrks: I believe there is hardly any reason to ever return a const value. References yes, but returns by value are by definition the property of the caller.

src/lib/utils/settings/abstract_setting.cpp
```diff
@@ -0,0 +1,20 @@
+#include "abstract_setting.hpp"
+
+#include "hyrise.hpp"
+#include "utils/assert.hpp"
+
+namespace opossum {
+
+AbstractSetting::AbstractSetting(const std::string& init_name) : name(init_name) {}
+
+void AbstractSetting::enroll() {
```
mrks: register/unregister?

src/lib/utils/settings/abstract_setting.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace opossum {
+
+/**
+ * This is an abstract class for all settings objects.
+ *
+ * Settings register itself at the SettingsManager.
+ * Therefore it is neccessary that they are owned by a shared pointer (instantiate with std::make_shared).
+ * Thus, you need to call unenroll() when the setting is not longer needed (it won't get destructed otherwise,
+ * as SettingsManager would still hold a shared pointer to it).
```
mrks: This sounds a lot like a `unique_ptr` to me.

src/lib/utils/settings/abstract_setting.hpp
mrks: I'd welcome some information on who creates setting objects.

src/lib/utils/settings/mock_setting.cpp
```diff
@@ -0,0 +1,27 @@
+#include "mock_setting.hpp"
```
mrks: Again: needed in src/lib?

src/lib/utils/settings_manager.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <functional>
+#include <unordered_map>
+
+#include "types.hpp"
+#include "utils/settings/abstract_setting.hpp"
+
+namespace opossum {
+
+class SettingsManager : public Noncopyable {
```
mrks: doc

src/lib/utils/meta_tables/meta_settings_table.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+  return output_table;
+}
+
+void MetaSettingsTable::_on_update(const std::vector<AllTypeVariant>& selected_values,
+                                   const std::vector<AllTypeVariant>& update_values) {
+  const auto& name = std::string{boost::get<pmr_string>(selected_values.at(0))};
+  Assert(Hyrise::get().settings_manager.has_setting(name), "No setting named " + name + " found.");
+
+  const auto& value = std::string{boost::get<pmr_string>(update_values.at(1))};
```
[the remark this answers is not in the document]
dey4ss: The first value in the row (i.e., vector of AllTypeVariants) is the setting name, the second one is the setting value (that we want to update).

src/lib/utils/settings/abstract_setting.cpp
dey4ss: Just a relict of my tries to make the setting registration work :D

src/lib/operators/mutate_meta_table.cpp
```diff
@@ -0,0 +1,54 @@
...
+  static const auto name = std::string{"Mutate Meta Table"};
+  return name;
+}
+
+std::shared_ptr<const Table> MutateMetaTable::_on_execute(std::shared_ptr<TransactionContext> context) {
+  // Meta tables can only be modified during non-transactions
+  if (context->auto_commit() == AutoCommit::No) {
+    _mark_as_failed();
+    return nullptr;
+  }
```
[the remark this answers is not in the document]
dey4ss: Apparently, the SQLTranslator does not know anything about TransactionContexts. The SQLPipelineStatement sets the context for every operator after generating the physical plan. 
An Assert here would also do the trick with crashing everything.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -147,7 +147,7 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
 
   // If we need a transaction context but haven't passed one in, this is the latest point where we can create it
   if (!_transaction_context && _use_mvcc == UseMvcc::Yes) {
-    _transaction_context = Hyrise::get().transaction_manager.new_transaction_context();
+    _transaction_context = Hyrise::get().transaction_manager.new_transaction_context(AutoCommit::Yes);
```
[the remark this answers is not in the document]
dey4ss: This is basically the only point where a transaction context for an auto commit statement is obtained

src/lib/utils/meta_table_manager.cpp
```diff
@@ -1,237 +1,94 @@
...
 
-std::shared_ptr<Table> MetaTableManager::generate_accurate_segments_table() {
-  PerformanceWarning("Accurate segment information are expensive to gather. Use with caution.");
-  const auto columns = TableColumnDefinitions{
-      {"table_name", DataType::String, false},       {"chunk_id", DataType::Int, false},
-      {"column_id", DataType::Int, false},           {"column_name", DataType::String, false},
-      {"column_data_type", DataType::String, false}, {"distinct_value_count", DataType::Long, false},
-      {"encoding_type", DataType::String, true},     {"vector_compression_type", DataType::String, true},
-      {"size_in_bytes", DataType::Long, false}};
+void MetaTableManager::update(const std::string& table_name, const std::shared_ptr<const Table>& selected_values,
```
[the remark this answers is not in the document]
dey4ss: I guess it won't be bad

src/lib/utils/meta_tables/abstract_meta_table.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * This is an abstract class for all meta table objects.
+ * Meta tables are significant different from normal tables.
+ * Information provided by the table is usually not persisted, but gathered on the fly.
+ *
+ * A meta table provides methods for retrieving, inserting, deleting or updating
+ * and the information if the operation may be done.
+ *
+ * Meta tables should be declared as members in the MetaTableManager constructor.
```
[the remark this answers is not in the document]
dey4ss: In the method body of the constructor is a vector with shared pointers to the meta tables

src/lib/utils/meta_tables/meta_accurate_segments_table.cpp
```diff
@@ -0,0 +1,33 @@
+#include "meta_accurate_segments_table.hpp"
```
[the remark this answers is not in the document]
dey4ss: Exactly

src/lib/utils/meta_tables/meta_mock_table.hpp
dey4ss: I took the MockNode as orientation, this can also be in test/utils

src/lib/operators/mutate_meta_table.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+  switch (_mutation_type) {
+    case MetaTableMutation::Insert:
+      Hyrise::get().meta_table_manager.insert_into(_table_name, input_table_right());
```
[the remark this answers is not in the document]
dey4ss: The compiler won't let me name the function just delete. The same problem with register for the settings, that's why I chose those names. But maybe there are better ones.

src/lib/utils/meta_tables/meta_accurate_segments_table.cpp
```diff
@@ -0,0 +1,33 @@
...
+                                               {"column_id", DataType::Int, false},
+                                               {"column_name", DataType::String, false},
+                                               {"column_data_type", DataType::String, false},
+                                               {"distinct_value_count", DataType::Long, false},
+                                               {"encoding_type", DataType::String, true},
+                                               {"vector_compression_type", DataType::String, true},
+                                               {"size_in_bytes", DataType::Long, false}}) {}
+
+const std::string& MetaAccurateSegmentsTable::name() const {
+  static const auto name = std::string{"segments_accurate"};
```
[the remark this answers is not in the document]
dey4ss: The old name was `segments_accurate`. I agree that the name and the class name should be aligned, so which one is nicer?

src/lib/utils/settings/abstract_setting.hpp
dey4ss: I have "A setting is an adjusting screw of any component with a variable value. That component (or "parent") instantiates and destroys a setting." now.

src/lib/utils/settings/abstract_setting.hpp
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Settings provide a getter/setter with a string.
```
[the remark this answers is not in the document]
dey4ss: "Settings provide a getter/setter that returns/requires a std::string."?

src/lib/sql/sql_translator.cpp
```diff
@@ -409,33 +421,62 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_insert(const hsql::In
...
+
+  bool is_meta_table = false;
+
   const auto sql_identifier_resolver = std::make_shared<SQLIdentifierResolver>();
   auto data_to_delete_node = _translate_stored_table(delete_statement.tableName, sql_identifier_resolver);
 
-  AssertInput(!MetaTableManager::is_meta_table_name(delete_statement.tableName), "Cannot modify meta tables");
+  if (MetaTableManager::is_meta_table_name(table_name)) {
+    const auto meta_table_name = table_name.substr(MetaTableManager::META_PREFIX.size());
+    AssertInput(Hyrise::get().meta_table_manager.has_table(meta_table_name) &&
```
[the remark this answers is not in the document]
dey4ss: I now had a look at it and do not assert the existence of a meta table, as everything is done with the existing methods.

src/lib/utils/settings_manager.cpp
dey4ss: e.g. hyriseTest built with clang
```
[ RUN      ] MetaSettingsTest.IsUpdateable
unknown file: Failure
C++ exception with description "std::bad_alloc" thrown in SetUp().
```
where this is works with gcc (but there the names are not set correctly).

mrks: Sounds like memory corruption to me. I'd try asan/ubsan.

dey4ss: Yep, was absolutely stupid. Stored `std::string& name` as AbstractSetting member.

src/lib/utils/settings/abstract_setting.hpp
Bensk1: Sounds good. Often, the term "knob" is used for database systems.

src/lib/utils/settings/abstract_setting.hpp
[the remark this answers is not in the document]
Bensk1: Yes!

src/lib/utils/meta_tables/meta_accurate_segments_table.cpp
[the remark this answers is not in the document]
Bensk1: I don't really mind. If I had to choose I would go for `segments_accurate`

src/lib/utils/meta_tables/meta_settings_table.cpp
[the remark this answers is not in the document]
Bensk1: Thanks.

Bensk1:
Any large questions/problems left or is it mostly just incorporating the feedback?

dey4ss:
> 
> 
> Any large questions/problems left or is it mostly just incorporating the feedback?

The last one ;)
I hope I get everything done before going on vacation on Monday.

src/lib/utils/settings/abstract_setting.hpp
dey4ss: Can I also get something like `shared_from_this()` on the object to auto-register settings at the settings manager after `std::make_unique<>()`?

src/lib/utils/meta_tables/abstract_meta_table.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+  /*
+   * Manipulate the meta table by calling _on_insert() / _on_remove.
+   * Additionally, checks if the input values match the column definitions are done.
+   */
+  void insert(const std::vector<AllTypeVariant>& values);
+  void remove(const std::vector<AllTypeVariant>& values);
+  void update(const std::vector<AllTypeVariant>& selected_values, const std::vector<AllTypeVariant>& update_values);
+
+  void _assert_data_types(const std::vector<AllTypeVariant>& values) const;
```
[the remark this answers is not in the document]
dey4ss: Also the remove, update, insert methods are protected, as we only want the MetaTableManager to call them.

src/lib/operators/mutate_meta_table.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+class TransactionContext;
+
+/**
+ * Operator that modifies meta tables.
+ * Expects the table name of the table to modify as a string and both
+ * the values to modify and the fully modified values.
+ *
+ * This is not MVCC safe, so we do nothing on commit or rollback.
+ * This is why the operator should only be used in auto-commit queries.
```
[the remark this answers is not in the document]
dey4ss: I changed the behavior of the operator to assert that the transaction is auto-commit, so we throw an exception.

src/lib/utils/settings/abstract_setting.cpp
Bensk1: I like that better, yes.

Bensk1:
Great, just give us a ping for a final review.

src/lib/utils/settings/abstract_setting.cpp
dey4ss: `register` is a reserved keyword, so something like `register_at_settings_manager()`?

src/lib/logical_query_plan/mutate_meta_table_node.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <vector>
+
+#include "base_non_query_node.hpp"
+
+namespace opossum {
+
+class AbstractExpression;
+
+/*
+ * Node type to represent mutations (insert, delete, update) of a meta table.
```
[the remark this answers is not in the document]
mrks: I was actually thinking about the node's inputs, not the constructor arguments, sorry.

src/lib/operators/mutate_meta_table.cpp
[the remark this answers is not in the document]
mrks: We might be able to revisit this with the pending transaction PR

src/lib/operators/mutate_meta_table.cpp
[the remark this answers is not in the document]
mrks: Forgot about `deleted as we luckily never use it.

src/lib/utils/meta_tables/abstract_meta_table.hpp
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * This is an abstract class for all meta table objects.
+ * Meta tables are significantly different from normal tables.
+ * Information provided by the table is usually not persisted, but gathered on the fly.
+ *
+ * A meta table may provide methods for retrieving, inserting, deleting or updating.
+ * Not all of these operations are allowed for all meta tables. Therefore, they also provide methods
+ * that indicate whether the particular operation is allowed, e.g., can_insert().
+ *
+ * Modifications of a meta table's values ususally trigger some actions, like loading plugins.
```
mrks: Typo

src/lib/utils/meta_tables/abstract_meta_table.hpp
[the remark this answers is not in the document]
mrks: Got it. I wouldn't really call it a member, which is a term usually reserved for class members. Just `added` ?

src/lib/utils/meta_tables/abstract_meta_table.hpp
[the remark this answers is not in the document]
mrks: In that case, their naming is wrong, isn't it?

src/lib/logical_query_plan/mutate_meta_table_node.hpp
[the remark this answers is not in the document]
dey4ss: Okay :D

src/lib/utils/meta_tables/abstract_meta_table.hpp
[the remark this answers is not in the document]
dey4ss: Yes, I already changed that

src/lib/utils/settings/abstract_setting.hpp
mrks: Ah, I was on the wrong track. Does the comment need the technical justification? Maybe just "settings need to be registered and deregistered in the SettingsManager. If a plugin is unloaded and does not remove its settings, this might lead to undefined behavior.'?

src/lib/utils/settings/abstract_setting.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+namespace opossum {
+
+/**
+ * This is an abstract class for all settings objects.
+ * A setting is a configuration knob of any component with a variable value.
+ * That component (or "parent") instantiates and destroys a setting.
```
mrks: Parent makes me think of class hierarchies. Maybe "component configured by the setting"?

src/lib/utils/settings_manager.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <unordered_map>
+
+#include "types.hpp"
+#include "utils/settings/abstract_setting.hpp"
+
+namespace opossum {
+
+/**
+ * This component is the central point for accessing all Setting objects.
+ * Settings register here, so any component can get or change their values.
```
mrks: You really like 'component', right? To me, it means only little more than 'thing'. Also, you are using the word twice here, referring to two different things.

"The SettingsManager is ..." and "Settings registered here can be changed from everywhere. Users change settings through the settings meta table."?

src/test/sql/sql_translator_test.cpp
```diff
@@ -1990,6 +2005,23 @@ TEST_F(SQLTranslatorTest, DeleteConditional) {
   EXPECT_LQP_EQ(actual_lqp, expected_lqp);
 }
 
+TEST_F(SQLTranslatorTest, DeleteMetaTable) {
```
mrks: DeleteFrom...?

src/test/utils/meta_tables/meta_table_test.cpp
```diff
@@ -0,0 +1,138 @@
...
+  static MetaTableNames meta_table_names() {
+    MetaTableNames names;
+    for (auto& table : MetaTableTest::meta_tables()) {
+      names.push_back(table->name());
+    }
+
+    return names;
+  }
+
+  const std::shared_ptr<Table> generateTable(const std::shared_ptr<AbstractMetaTable>& table) const {
```
mrks: Naming

src/test/utils/setting_test.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+  EXPECT_FALSE(settings_manager.has_setting("mock_setting"));
+
+  mock_setting->enroll();
+
+  EXPECT_TRUE(settings_manager.has_setting("mock_setting"));
+
+  mock_setting->unenroll();
+
+  EXPECT_FALSE(settings_manager.has_setting("mock_setting"));
```
mrks: No need for those empty lines

src/lib/utils/settings/abstract_setting.cpp
mrks: Argh, second keyword fail for today. Clunky, but works for me.

dey4ss:
@Bensk1 I guess I'm done

Bensk1:
Sounds like ready to merge for me as soon as the CI is done. Thanks for getting it done that quickly.

PR #2007, opened 2020-02-06
Change line endings from CRLF to LF

mflueggen:
I found out by accident, that a few files had CRLF instead of LF as line endings. This PR should correct the line endings in the files.


mrks:
@dey4ss Looks like your system uses CRLF. Any chance you can fix that for the future?

mrks:
Re-running after cleaning ccache. Maybe that has issues with differing line endings. Please ping me if this doesn't fix it.

dey4ss:
> @dey4ss Looks like your system uses CRLF. Any chance you can fix that for the future?

Oh sorry, I'm gonna change this

PR #2008, opened 2020-02-07
Plot operator breakdown

mrks:
This adds a table at the lower right corner of each PQP visualization that shows the aggregate runtime per operator:

Screenshot 2020-02-07 at 08 41 51

Also, this adds a script that accumulates this information over different *PQP.svg files:

operator_breakdown

I don't particularly like the colors but haven't found a nice color scheme that doesn't imply some type of ordering.

### DIFF
--- a/scripts/plot_operator_breakdown.py
+++ b/scripts/plot_operator_breakdown.py
@@ -41,15 +41,9 @@ for file in sorted(glob.glob('*-PQP.svg')):
         for operator_name in row_strings[0].split('|'):
             operator_names.append(operator_name.strip())
 
-        # Quick'n dirty conversion from time string to nanoseconds
+        # Convert time string to nanoseconds and add to operator_durations
         for operator_duration_str in row_strings[1].split('|'):
-            operator_duration_str = operator_duration_str.replace(' min ', ' * 60 * 1e9 + ')
-            operator_duration_str = operator_duration_str.replace(' s ', ' * 1e9 + ')
-            operator_duration_str = operator_duration_str.replace(' ms ', ' * 1e6 + ')
-            operator_duration_str = operator_duration_str.replace(' µs ', ' * 1e3 + ')
-            operator_duration_str = operator_duration_str.replace(' ns ', ' + ')
-            operator_duration_str += '0'
-            operator_duration = float(eval(operator_duration_str))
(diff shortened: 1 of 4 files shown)


Bouncner:
What about ColorBrewer's `Set1`?
https://ggplot2.tidyverse.org/reference/scale_brewer.html
It's also directly available in MPL.

Bouncner:
How hard would it be to sort the table?

In general: great additional. I found myself to be looking at PQP and manually accumulating runtimes by far too often lately.

mrks:
> What about ColorBrewer's Set1?

Looks more like a scale than discrete values

> How hard would it be to sort the table?

Shouldn't be too bad. By operator name or by time spent?

Bouncner:
> Looks more like a scale than discrete values
Are we talking about the same color set? http://colorbrewer2.org/#type=qualitative&scheme=Set1&n=3 (Set1 and Dark2 are quite common in many papers).

> Shouldn't be too bad. By operator name or by time spent?
I'd say time spent descending.

scripts/plot_operator_breakdown.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # The svg table stores data in a columnar orientation, so we first extract the operator names, then their
+        # durations
+        operator_names = []
+        operator_durations = []
+
+        for operator_name in row_strings[0].split('|'):
+            operator_names.append(operator_name.strip())
+
+        # Quick'n dirty conversion from time string to nanoseconds
```
j-tr: maybe less dirty:
```
operator_duration = pd.Timedelta(operator_duration_str.replace('µ', 'u').replace('min', 'm')).total_seconds() * 1e9
```

mrks: You magician!

mrks:
> Are we talking about the same color set? http://colorbrewer2.org/#type=qualitative&scheme=Set1&n=3 (Set1 and Dark2 are quite common in many papers).

Yes. I was thinking about using that in the paper, too, but I don't like that the same hue is used multiple times. For me, that implies some connection between the values with the same hue.

mrks:
> I'd say time spent descending.

Sic fiat.

mrks:
@j-tr Ready to go?

PR #2009, opened 2020-02-08
[DYOD] Cost Model Calibration

boehmchen:
If you have any questions, please do not hesitate to ask us.


src/calibration/benchmark_builder.cpp
```diff
@@ -0,0 +1,94 @@
...
+      // auto item_runner = std::make_unique<TPCHBenchmarkItemRunner>(config, USE_PREPARED_STATEMENTS, SCALE_FACTOR, tpch_query_ids_benchmark);
+      auto item_runner = std::make_unique<TPCHBenchmarkItemRunner>(_config, USE_PREPARED_STATEMENTS, SCALE_FACTOR);
+      auto benchmark_runner = std::make_shared<BenchmarkRunner>(
+              *_config, std::move(item_runner), std::make_unique<TPCHTableGenerator>(SCALE_FACTOR, _config), BenchmarkRunner::create_context(*_config));
+      Hyrise::get().benchmark_runner = benchmark_runner;
+      benchmark_runner->run();
+    }
+
+    void BenchmarkBuilder::_run_tcpds(const float SCALE_FACTOR) const {
+      const std::string query_path = "hyrise/resources/benchmark/tpcds/tpcds-result-reproduction/query_qualification";
```
Bouncner: That should have been only necessary for my plugin, not for the hyrise binary.

src/calibration/benchmark_builder.cpp
```diff
@@ -0,0 +1,94 @@
...
+      auto query_generator = std::make_unique<FileBasedBenchmarkItemRunner>(_config, query_path, std::unordered_set<std::string>{});
+      auto table_generator = std::make_unique<TpcdsTableGenerator>(SCALE_FACTOR, _config);
+      auto benchmark_runner = std::make_shared<BenchmarkRunner>(*_config, std::move(query_generator), std::move(table_generator),
+                                                                opossum::BenchmarkRunner::create_context(*_config));
+      Hyrise::get().benchmark_runner = benchmark_runner;
+      benchmark_runner->run();
+    }
+
+    void BenchmarkBuilder::_run_job(const float SCALE_FACTOR) const { //TODO Remove SCALE_FACTOR HERE
+      const auto table_path = "hyrise/imdb_data";
```
Bouncner: See above.

src/calibration/table_export.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include "storage/table.hpp"
+#include "calibration_table_wrapper.hpp"
+
+namespace opossum {
+
+enum TableExportType {
+    TABLE, COLUMN, SEGMENT
+};
+
+class TableExport{
```
Bouncner: You do not export tables here.

src/calibration/table_generator.cpp
```diff
@@ -0,0 +1,61 @@
+#include "table_generator.hpp"
+#include "storage/table.hpp"
+#include "constant_mappings.hpp"
+
+namespace opossum {
+
+  TableGenerator::TableGenerator(std::shared_ptr<TableGeneratorConfig> config) {
```
Bouncner: I would rather vote for CalibrationTableGenerator (if that is necessary at all) since we have multiple TableGenerators already. Since this does not generate a table but rather multiple tables, the name isn't a good fit anyways.
Maybe CalibrationSetup or something similar?

python/data/TableScan_train.csv
```diff
@@ -0,0 +1,2497 @@
+INPUT_ROWS_LEFT,OUTPUT_ROWS,RUNTIME_NS,COLUMN_TYPE,TABLE_NAME,COLUMN_NAME
```
Bouncner: Please do not commit such files. Further, the name is not well formatted.

python/data/table_scans.csv
```diff
@@ -0,0 +1,735 @@
+QUERY_HASH,SCAN_TYPE,TABLE_NAME,COLUMN_NAME,INPUT_ROWS,OUTPUT_ROWS,RUNTIME_NS,DESCRIPTION
```
Bouncner: Please do not commit such files.

src/calibration/measurement_export.cpp
```diff
@@ -0,0 +1,230 @@
...
+        case OperatorType::CreatePreparedPlan:  _export_generic(op);    break;
+        case OperatorType::CreateView:          _export_generic(op);    break;
+        case OperatorType::DropTable:           _export_generic(op);    break;
+        case OperatorType::DropView:            _export_generic(op);    break;
+        case OperatorType::Mock:                _export_generic(op);    break;
+      }
+    }
+
+    //TODO Construct path before starting export-> save it in map
+    std::string MeasurementExport::_path(OperatorType operator_type) const {
```
Bouncner: Measurement is to broad for my taste. `Calibration`? `CalibrationMeasurement`?

Please consider that this is a rather narrow area of Hyrise that you are working on here. Your names should not confuse other students which use "core Hyrise".

src/calibration/measurement_export.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+                // SCAN_TYPE
+                if (original_node == node->left_input()) {
+                  ss << "COLUMN_SCAN";
+                } else {
+                  ss << "REFERENCE_SCAN";
+                }
+                ss << _separator;
+
+                // TABLE_NAME
```
Bouncner: That comment does not help.

.gitignore
```diff
@@ -42,4 +42,4 @@ cmake-build-*
 default.profdata
 default.profraw
 
-tpcds_cached_tables/
```
Bensk1, line 45: We should not remove that. However, I am not sure why this is part of the master and `tpch_cached_tables` is not.

python/cost_models.py
```diff
@@ -0,0 +1,220 @@
...
+    abline_values = range(int(axis_min), int(axis_max), int((axis_max-axis_min)/100))
+
+    # Plot the best fit line over the actual values
+    plt.plot(abline_values, abline_values, c = 'r', linestyle="-")
+    plt.title('{}_{}_{}; Score: {}'.format(encoding, scan, method, model_scores['R2']))
+    plt.ylim([axis_min, axis_max])
+    plt.xlim([axis_min, axis_max])
+    plt.xlabel("Real Time")
+    plt.ylabel("Predicted Time")
+    output_path = '{}/Plots/{}_{}_{}'.format(out, method, encoding, scan)
```
Bensk1, line 64: For the future, consider using python's f-strings.

src/calibration/calibration.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+using namespace opossum;  // NOLINT
+
+int main() {
+  auto table_config = std::make_shared<TableGeneratorConfig>(TableGeneratorConfig{
+      {DataType::Double, DataType::Float, DataType::Int, DataType::Long, DataType::String, DataType::Null},
+      {EncodingType::Dictionary},
+      {ColumnDataDistribution::make_uniform_config(0.0, 1000.0)},
+      {100000},
+      {1500, 3000, 6000, 10000, 20000, 30000, 60175, 25, 15000, 2000, 8000, 5, 100}});
```
Bensk1, line 21: I would put these (and the ones above) into variables, e.g., `constexpr auto row_counts ...` to make it more descriptive

src/calibration/calibration.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include "table_feature_export.hpp"
+
+using namespace opossum;  // NOLINT
+
+int main() {
+  auto table_config = std::make_shared<TableGeneratorConfig>(TableGeneratorConfig{
+      {DataType::Double, DataType::Float, DataType::Int, DataType::Long, DataType::String, DataType::Null},
+      {EncodingType::Dictionary},
+      {ColumnDataDistribution::make_uniform_config(0.0, 1000.0)},
+      {100000},
```
Bensk1, line 20: This is the chunk_size, right? How about using `Chunk::DEFAULT_SIZE`?

src/calibration/calibration.cpp
```diff
@@ -0,0 +1,57 @@
...
+  auto const path_train = "./data/train";
+  auto const path_test = "./data/test";
+
+  const auto feature_export = OperatorFeatureExport(path_train);
+  auto lqp_generator = CalibrationLQPGenerator();
+  auto table_export = TableFeatureExport(path_train);
+
+  auto benchmark_runner = CalibrationBenchmarkRunner(path_test);
+
+  benchmark_runner.run_benchmark(BenchmarkType::TCPH, 0.01f, 10);
```
Bensk1, line 34: Again, since these numbers look like configuration parameters, I would treat them as such and put them as constants in a place that is properly visible.

src/calibration/calibration.cpp
```diff
@@ -0,0 +1,57 @@
...
+  const auto tables = table_generator.generate();
+
+  auto const path_train = "./data/train";
+  auto const path_test = "./data/test";
+
+  const auto feature_export = OperatorFeatureExport(path_train);
+  auto lqp_generator = CalibrationLQPGenerator();
+  auto table_export = TableFeatureExport(path_train);
+
+  auto benchmark_runner = CalibrationBenchmarkRunner(path_test);
```
Bensk1, line 32: This implies that all benchmark data should be used for testing only, right?

src/calibration/calibration.cpp
```diff
@@ -0,0 +1,57 @@
...
+  benchmark_runner.run_benchmark(BenchmarkType::TCPH, 0.01f, 10);
+
+  for (const auto& table : tables) {
+    Hyrise::get().storage_manager.add_table(table->get_name(), table->get_table());
+
+    lqp_generator.generate(OperatorType::TableScan, table);
+  }
+
+  const auto lqps = lqp_generator.get_lqps();
+  // Execution of lpqs; In the future a good scheduler as replacement for following code would be awesome.
```
Bensk1, line 43: I don't understand that. What functionality should the scheduler offer to be "good" in that case?

src/calibration/calibration.cpp
```diff
@@ -0,0 +1,57 @@
...
+  for (const std::shared_ptr<AbstractLQPNode>& lqp : lqps) {
+    const auto pqp = LQPTranslator{}.translate_node(lqp);
+    const auto tasks = OperatorTask::make_tasks_from_operator(pqp);
+    Hyrise::get().scheduler()->schedule_and_wait_for_tasks(tasks);
+
+    // Execute LQP directly after generation
```
Bensk1, line 49: Execute or export?

src/calibration/calibration.cpp
```diff
@@ -0,0 +1,57 @@
...
+      {1500, 3000, 6000, 10000, 20000, 30000, 60175, 25, 15000, 2000, 8000, 5, 100}});
+  auto table_generator = CalibrationTableGenerator(table_config);
+  const auto tables = table_generator.generate();
+
+  auto const path_train = "./data/train";
+  auto const path_test = "./data/test";
+
+  const auto feature_export = OperatorFeatureExport(path_train);
+  auto lqp_generator = CalibrationLQPGenerator();
+  auto table_export = TableFeatureExport(path_train);
```
Bensk1, line 30: Maybe _TableFeatureExporter_ is a better name? And then I would also call that `table_feature_exporter`

src/calibrationlib/calibration_benchmark_runner.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "hyrise.hpp"
+
+namespace opossum {
+
+CalibrationBenchmarkRunner::CalibrationBenchmarkRunner(const std::string& path_to_dir)
+    : _feature_export(OperatorFeatureExport(path_to_dir)), _table_export(TableFeatureExport(path_to_dir)) {
+  auto config = std::make_shared<BenchmarkConfig>(BenchmarkConfig::get_default_config());
+  config->max_runs = 1;
+  config->enable_visualization = false;
+  config->chunk_size = 100'000;
```
Bensk1, line 17: See above, let's use the DEFAULT_SIZE.

src/calibrationlib/calibration_benchmark_runner.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <tpch/tpch_table_generator.hpp>
+#include "hyrise.hpp"
+
+namespace opossum {
+
+CalibrationBenchmarkRunner::CalibrationBenchmarkRunner(const std::string& path_to_dir)
+    : _feature_export(OperatorFeatureExport(path_to_dir)), _table_export(TableFeatureExport(path_to_dir)) {
+  auto config = std::make_shared<BenchmarkConfig>(BenchmarkConfig::get_default_config());
+  config->max_runs = 1;
+  config->enable_visualization = false;
```
Bensk1, line 16: I am not sure, but isn't that and some of the others a default setting?

src/calibrationlib/calibration_benchmark_runner.cpp
```diff
@@ -0,0 +1,104 @@
...
+  _config = config;
+}
+
+CalibrationBenchmarkRunner::CalibrationBenchmarkRunner(const std::string& path_to_dir,
+                                                       std::shared_ptr<BenchmarkConfig> config)
+    : _feature_export(OperatorFeatureExport(path_to_dir)), _table_export(TableFeatureExport(path_to_dir)) {
+  _config = config;
+}
+
+void CalibrationBenchmarkRunner::run_benchmark(const BenchmarkType type, const float SCALE_FACTOR,
```
Bensk1, line 28: Why is `SCALE_FACTOR` in all caps?

src/calibrationlib/calibration_benchmark_runner.cpp
```diff
@@ -0,0 +1,104 @@
...
+      benchmark_runner = _build_tcpds(SCALE_FACTOR);
+      break;
+    case BenchmarkType::JOB:
+      benchmark_runner = _build_job();
+      break;
+    default:
+      throw std::runtime_error("Provided unknown BenchmarkType.");
+  }
+
+  for (int i = 0; i < number_of_executions; ++i) {
```
Bensk1, line 45: `i` is not a good name.

src/calibrationlib/calibration_benchmark_runner.cpp
```diff
@@ -0,0 +1,104 @@
...
+    benchmark_runner->run();
+
+    auto& pqp_cache = Hyrise::get().default_pqp_cache;
+
+    for (auto pqp_entry = pqp_cache->unsafe_begin(); pqp_entry != pqp_cache->unsafe_end(); ++pqp_entry) {
+      const auto& [query_string, physical_query_plan] = *pqp_entry;
+      _feature_export.export_to_csv(physical_query_plan);
+    }
+
+    // Clear pqp cache for next benchmark
```
Bensk1, line 56: benchmark execution or benchmark run?

src/calibrationlib/calibration_benchmark_runner.cpp
```diff
@@ -0,0 +1,104 @@
...
+    pqp_cache->clear();
+  }
+
+  const std::vector<std::string> table_names = Hyrise::get().storage_manager.table_names();
+  for (const auto& table_name : table_names) {
+    auto table = Hyrise::get().storage_manager.get_table(table_name);
+    _table_export.export_table(std::make_shared<CalibrationTableWrapper>(CalibrationTableWrapper(table, table_name)));
+
+    // Drop table because we do not need it anymore
+    Hyrise::get().storage_manager.drop_table(table_name);
```
Bensk1, line 66: Usually, we don't drop tables for benchmarks because when the binary exits, everything is "dropped". However, it does not harm and there might be cases where we run multiple benchmarks with different scale factors after each other for calibration.

But I think the comment is unnecessary.

src/calibrationlib/calibration_benchmark_runner.cpp
```diff
@@ -0,0 +1,104 @@
...
+                                                            BenchmarkRunner::create_context(*_config));
+
+  return benchmark_runner;
+}
+
+std::shared_ptr<BenchmarkRunner> CalibrationBenchmarkRunner::_build_tcpds(const float SCALE_FACTOR) const {
+  const std::string query_path = "hyrise/resources/benchmark/tpcds/tpcds-result-reproduction/query_qualification";
+
+  auto query_generator =
+      std::make_unique<FileBasedBenchmarkItemRunner>(_config, query_path, std::unordered_set<std::string>{});
```
Bensk1, line 83: Please take a look at the tpcds_benchmark.cpp, there is a query blacklist.

src/calibrationlib/calibration_benchmark_runner.cpp
```diff
@@ -0,0 +1,104 @@
...
+  auto item_runner = std::make_unique<TPCHBenchmarkItemRunner>(_config, false, SCALE_FACTOR);
+  auto benchmark_runner = std::make_shared<BenchmarkRunner>(*_config, std::move(item_runner),
+                                                            std::make_unique<TPCHTableGenerator>(SCALE_FACTOR, _config),
+                                                            BenchmarkRunner::create_context(*_config));
+
+  return benchmark_runner;
+}
+
+std::shared_ptr<BenchmarkRunner> CalibrationBenchmarkRunner::_build_tcpds(const float SCALE_FACTOR) const {
+  const std::string query_path = "hyrise/resources/benchmark/tpcds/tpcds-result-reproduction/query_qualification";
```
Bensk1, line 80: That path does not seem right. We usually execute the binaries from the _hyrise_ folder.

src/calibrationlib/csv_writer.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <vector>
+
+#include "boost/bimap.hpp"
+#include "utils/assert.hpp"
+
+namespace opossum {
+
+class CSVWriter {
+ public:
+  CSVWriter(std::string file_path, const std::vector<std::string> headers, const char delimiter = ',',
```
Bensk1, line 13: Not a strong opinion on that, but does it make sense to be able to change the delimiter here? I would expect the delimiter to stay the same for a complete "calibration project". Therefore, I could imagine having delimiter as a constant.

src/calibrationlib/csv_writer.cpp
```diff
@@ -0,0 +1,114 @@
...
+namespace opossum {
+
+CSVWriter::CSVWriter(
+    std::string file_path,
+    const std::vector<std::string>
+        headers,  // headers for csv file, if the file for file_path already exists, it must have equal headers
+    const char delimiter, bool replace_file)
+    : _headers(headers), _file_path(std::move(file_path)), _delimiter(delimiter) {
+  // Get path from string
+  std::filesystem::path boost_path(_file_path);
```
Bensk1, line 20: boost?

src/calibrationlib/csv_writer.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include <utility>
+
+#include <fstream>
+
+#include <boost/algorithm/string.hpp>
+
+namespace opossum {
+
+CSVWriter::CSVWriter(
+    std::string file_path,
```
Bensk1, line 14: const?

src/calibrationlib/operator_feature_export.hpp
```diff
@@ -0,0 +1,122 @@
...
+  const std::string& _path_to_dir;
+
+  std::optional<const std::vector<std::string>> _get_header(const OperatorType type) const {
+    std::stringstream ss;
+    switch (type) {
+      case OperatorType::TableScan:
+        return std::vector<std::string>({"INPUT_ROWS_LEFT", "OUTPUT_ROWS", "RUNTIME_NS", "SCAN_TYPE", "TABLE_NAME",
+                                         "COLUMN_NAME", "SCAN_IMPLEMENTATION"});
+      default:
+        return {};
```
Bensk1, line 26: `Fail(...)`?

src/calibrationlib/operator_feature_export.hpp
```diff
@@ -0,0 +1,122 @@
...
+      case OperatorType::CreateView:
+        return "CreateView";
+      case OperatorType::DropTable:
+        return "DropTable";
+      case OperatorType::DropView:
+        return "DropView";
+      case OperatorType::Mock:
+        return "Mock";
+      default:
+        throw std::runtime_error("Requested mapping for unknown OperatorType");
```
Bensk1, line 118: Fail()

src/calibrationlib/operator_feature_export.hpp
```diff
@@ -0,0 +1,122 @@
...
+      default:
+        return {};
+    }
+  }
+
+  const std::map<OperatorType, std::shared_ptr<CSVWriter>> _csv_writers = [&]() {
+    auto csv_writers_per_operator = std::map<OperatorType, std::shared_ptr<CSVWriter>>();
+
+    for (int op_type_id = static_cast<int>(OperatorType::Projection);
+         op_type_id != static_cast<int>(OperatorType::Validate); ++op_type_id) {
```
Bensk1, line 34: What is the rationale of that?

src/calibrationlib/operator_feature_export.hpp
Bensk1: Ah, now that I read more of that code, I see that this would not work because you rely on the optional being returned here.

src/calibrationlib/operator_feature_export.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "hyrise.hpp"
+#include "operator_feature_export.hpp"
+#include "storage/table.hpp"
+
+namespace opossum {
+
+OperatorFeatureExport::OperatorFeatureExport(const std::string& path_to_dir) : _path_to_dir(path_to_dir) {}
+
+void OperatorFeatureExport::export_to_csv(std::shared_ptr<const AbstractOperator> op) const {
+  if (op != nullptr) {
```
Bensk1, line 17: You could also write `if (op)` which would have the same effect. However, should that function be ever called intentionally with a `nullptr`? If not, I would put an Assert here.

src/calibrationlib/operator_feature_export.cpp
```diff
@@ -0,0 +1,107 @@
...
+  }
+}
+
+void OperatorFeatureExport::_export_typed_operator(std::shared_ptr<const AbstractOperator> op) const {
+  switch (op->type()) {
+    case OperatorType::TableScan:
+      _export_table_scan(op);
+      break;
+    default:
+      break;
```
Bensk1, line 32: Fail?

src/calibrationlib/operator_feature_export.cpp
```diff
@@ -0,0 +1,107 @@
...
+  for (const auto& el : node->node_expressions) {
+    visit_expression(el, [&](const auto& expression) {
+      if (expression->type == ExpressionType::LQPColumn) {
+        const auto column_expression = std::dynamic_pointer_cast<LQPColumnExpression>(expression);
+        const auto column_reference = column_expression->column_reference;
+        const auto original_node = column_reference.original_node();
+
+        if (original_node->type == LQPNodeType::StoredTable) {
+          // Get values for INPUT_ROWS_LEFT
+          if (op->input_left() != nullptr) {
```
Bensk1, line 53: see above

src/calibrationlib/operator_feature_export.cpp
```diff
@@ -0,0 +1,107 @@
...
+    default:
+      break;
+  }
+}
+
+// Export features of a table scan operator
+void OperatorFeatureExport::_export_table_scan(std::shared_ptr<const AbstractOperator> op) const {
+  DebugAssert(op->type() == OperatorType::TableScan, "Expected operator of type: TableScan but got another one");
+
+  auto csv_writer = _csv_writers.at(op->type());
```
Bensk1, line 40: Since this method might be executed quite often, performance is important. We could use `[]` without bounds checking and use a DebugAssert to be safe.

src/calibrationlib/operator_feature_export.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+  for (const auto& el : node->node_expressions) {
+    visit_expression(el, [&](const auto& expression) {
+      if (expression->type == ExpressionType::LQPColumn) {
+        const auto column_expression = std::dynamic_pointer_cast<LQPColumnExpression>(expression);
+        const auto column_reference = column_expression->column_reference;
+        const auto original_node = column_reference.original_node();
+
+        if (original_node->type == LQPNodeType::StoredTable) {
+          // Get values for INPUT_ROWS_LEFT
```
Bensk1, line 52: The code below is easy to understand, I think we don't need the comments here.

src/calibrationlib/operator_feature_export.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+          // Get values for COLUMN_NAME
+          const auto table = Hyrise::get().storage_manager.get_table(table_name);
+          csv_writer->set_value("COLUMN_NAME", table->column_names()[original_column_id]);
+
+          // Get values for SCAN_IMPLEMENTATION
+          auto description = op->description();
+          std::smatch matches;
+
+          std::regex self_regex("Impl: ([A-Z]\\w+)", std::regex_constants::ECMAScript | std::regex_constants::icase);
```
Bensk1, line 92: @Bouncner do you know a better way for this? 

Also, do we need ti define that again for every TableScan? It should stay the same, shouldn't it?

Bouncner: Why is "ECMAScript" being used here? I'd be fine with a substring taking everything after `Impl: `.

python/prepare_tpch.py
```diff
@@ -0,0 +1,33 @@
...
+                               'ENCODING', 'COMPRESSION', 'SIZE_IN_BYTES']]
+    joined_data = joined_data.rename(columns={"TABLE_NAME_x": "TABLE_NAME", "ROW_COUNT_x": "ROW_COUNT"})
+
+    joined_data['SELECTIVITY'] = (joined_data['OUTPUT_ROWS'] / joined_data['INPUT_ROWS'])
+    joined_data['SELECTIVITY'].fillna(0, inplace=True)
+
+    return joined_data
+
+
+def main():
```
Bouncner, line 32: Necessary?

src/calibrationlib/calibration_benchmark_runner.cpp
boehmchen: So, `resources/benchmark/tpcds/tpcds-result-reproduction/query_qualification` should be correct, right?

src/calibrationlib/calibration_benchmark_runner.cpp
```diff
@@ -0,0 +1,104 @@
...
+  auto query_generator =
+      std::make_unique<FileBasedBenchmarkItemRunner>(_config, query_path, std::unordered_set<std::string>{});
+  auto table_generator = std::make_unique<TpcdsTableGenerator>(SCALE_FACTOR, _config);
+  auto benchmark_runner =
+      std::make_shared<BenchmarkRunner>(*_config, std::move(query_generator), std::move(table_generator),
+                                        opossum::BenchmarkRunner::create_context(*_config));
+  return benchmark_runner;
+}
+
+std::shared_ptr<BenchmarkRunner> CalibrationBenchmarkRunner::_build_job() const {
```
boehmchen: I removed the job benchmark, because we do not have the /imdb_data folder in the master. Martin used the job benchmark in a plugin, which is why I added this benchmark here.

src/calibrationlib/calibration_benchmark_runner.cpp
boehmchen: I adjusted the code to behave like `tpcds_benchmark.cpp`. However I added following code (copied from `tpcds_benchmark.cpp`)

```cpp
namespace {
    const std::unordered_set<std::string> filename_blacklist() {
      auto filename_blacklist = std::unordered_set<std::string>{};
      const auto blacklist_file_path = "resources/benchmark/tpcds/query_blacklist.cfg";
      std::ifstream blacklist_file(blacklist_file_path);

      if (!blacklist_file) {
        std::cerr << "Cannot open the blacklist file: " << blacklist_file_path << "\n";
      } else {
        std::string filename;
        while (std::getline(blacklist_file, filename)) {
          if (filename.size() > 0 && filename.at(0) != '#') {
            filename_blacklist.emplace(filename);
          }
        }
        blacklist_file.close();
      }
      return filename_blacklist;
    }
}  // namespace
```

Is there a better way than copying the code above? Should I simply import `tpcds_benchmark.cpp` (has a main)?

src/test/CMakeLists.txt
```diff
@@ -261,7 +268,7 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 # Configure hyriseTest
 add_executable(hyriseTest ${HYRISE_UNIT_TEST_SOURCES})
 add_dependencies(hyriseTest hyriseTestPlugin hyriseTestNonInstantiablePlugin)
-target_link_libraries(hyriseTest hyrise ${LIBRARIES})
+target_link_libraries(hyriseTest hyrise hyriseCalibrationLib ${LIBRARIES})
```
boehmchen, line 271: Do you want that?

python/prepare_tpch.py
boehmchen: No it's not. I removed it.

src/calibrationlib/calibration_benchmark_runner.cpp
Bouncner: And we don't use `int` in Hyrise. Probably `size_t` here.

src/calibrationlib/calibration_benchmark_runner.cpp
Bouncner: The path has been adapted in the plugin, because the plugin includes Hyrise as a third-party module. Is it really necessary to pass the path here?

src/calibrationlib/calibration_benchmark_runner.cpp
Bouncner: If there is no simple way of doing that, we need to rewrite a few parts of the TPC-DS benchmark. And if that should be the case, I am fine with excluding is in this PR. It's not your fault that the current way to call benchmarks isn't straightforward.

src/calibrationlib/calibration_benchmark_runner.cpp
Bouncner: Same as with TPC-DS. Please don't use the "hacks" of the plugin. If it is not straightforward to call the benchmark, exclude it.

src/test/calibrationlib/csv_writer_test.cpp
```diff
@@ -0,0 +1,184 @@
...
+  EXPECT_EQ(line, "COLUMN_1,COLUMN_2,COLUMN_3,COLUMN_4");  // first line must be the new header
+}
+
+TEST_F(CSVWriterTest, PanicIfCreateNewFileWithDifferentHeaders) {
+  const auto csv_writer = default_writer();
+  const auto new_headers = std::vector<std::string>({"COLUMN_1", "COLUMN_2", "COLUMN_3", "COLUMN_4"});
+  EXPECT_ANY_THROW(CSVWriter(_file_path, new_headers, false));
+}
+
+TEST_F(CSVWriterTest, PanicIfDelimiterInValue) {
```
Bouncner, line 177: Better use Throw over Panic.

src/benchmarklib/synthetic_table_generator.cpp
```diff
@@ -210,4 +210,23 @@ std::shared_ptr<Table> SyntheticTableGenerator::generate_table(
...
+             << "_" << column_data_distribution.min_value << "_" << column_data_distribution.max_value;
+      break;
+    case DataDistributionType::NormalSkewed:
+      stream << "Skewed"
+             << "_" << column_data_distribution.skew_location << "_" << column_data_distribution.skew_scale << "_"
+             << column_data_distribution.skew_shape;
+      break;
+    case DataDistributionType::Pareto:
+      stream << "Pareto"
+             << "_" << column_data_distribution.pareto_scale << "_" << column_data_distribution.pareto_shape;
```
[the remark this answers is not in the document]
boehmchen: Good point. I thought about it too. At the moment, I use the the DataDistributionType as part of the generated column name, which is why I avoid using brackets. Are brackets allowed as part of a column name?

src/benchmarklib/synthetic_table_generator.hpp
```diff
@@ -44,6 +44,12 @@ struct ColumnDataDistribution {
     return c;
   }
 
+  bool operator==(ColumnDataDistribution& rhs) const {
+    return distribution_type == rhs.distribution_type && num_different_values == rhs.num_different_values &&
```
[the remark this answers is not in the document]
boehmchen: Thanks! I changed it to:
```cpp
bool operator==(ColumnDataDistribution& rhs) const {
    return std::tie(distribution_type, num_different_values, pareto_scale, skew_location,
            skew_scale,skew_shape, min_value, max_value) ==
            std::tie(rhs.distribution_type, rhs.num_different_values, rhs.pareto_scale, rhs.skew_location,
                    rhs.skew_scale, rhs.skew_shape, rhs.min_value, rhs.max_value);
  }
```

src/calibration/calibration.cpp
```diff
@@ -0,0 +1,71 @@
...
+  const std::set<DataType> table_data_types = {DataType::Double, DataType::Float,  DataType::Int,
+                                               DataType::Long,   DataType::String, DataType::Null};
+  const std::set<EncodingType> column_encoding_types = {EncodingType::Dictionary};
+  const std::vector<ColumnDataDistribution> column_data_distributions = {
+      ColumnDataDistribution::make_uniform_config(0.0, 1000.0)};
+  const std::set<ChunkOffset> chunk_sizes = {Chunk::DEFAULT_SIZE};
+  const std::set<int> row_counts = {1500, 3000, 6000, 10000, 20000, 30000, 60175, 25, 15000, 2000, 8000, 5, 100};
+
+  // test data generation settings
+  constexpr bool generate_test_data = true;
```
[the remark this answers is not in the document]
boehmchen: Nope. The linter does not trigger on lower case constexpr's. I changed all constexpr in calibration to upper_case values.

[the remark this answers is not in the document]
boehmchen: In addition, I changed all `const` variables for settings to upper case

src/calibration/calibration.cpp
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  const auto lqps = lqp_generator.get_lqps();
+
+  for (const std::shared_ptr<AbstractLQPNode>& lqp : lqps) {
+    const auto pqp = LQPTranslator{}.translate_node(lqp);
+    const auto tasks = OperatorTask::make_tasks_from_operator(pqp);
+    Hyrise::get().scheduler()->schedule_and_wait_for_tasks(tasks);
+
+    // Export LQP directly after generation
```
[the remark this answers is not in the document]
boehmchen: I edited the comment.

src/test/calibrationlib/operator_feature_exporter_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+  std::string line;
+  std::ifstream f(_dir_path + "/TableScan.csv");
+  std::getline(f, line);
+
+  std::vector<std::string> row_values;
+  boost::split(row_values, line, boost::is_any_of(","));
+
+  // Check if we inserted the correct headers
+  const auto num_headers = headers.size();
+  for (u_int64_t header_index = 0; header_index < num_headers; ++header_index) {
```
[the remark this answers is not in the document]
boehmchen: I replaced all usages of uint64_t to size_t

src/test/calibrationlib/operator_feature_exporter_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+  // Generate LQPs to export
+  auto lqp_generator = CalibrationLQPGenerator();
+  lqp_generator.generate(OperatorType::TableScan, _table);
+  auto const lqps = lqp_generator.get_lqps();
+
+  execute_and_export_pqps(lqps);
+
+  std::string line;
+  std::ifstream f(_dir_path + "/TableScan.csv");
```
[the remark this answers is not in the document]
boehmchen: I generate the CSV file name for operators based on the Enum -> string mapping.

Following code is used to create the file name/ path. (`operator_feature_exporter.hpp` `line 38 - 42`)
```cpp
if (const auto headers = _get_header(op_type)) {
        std::stringstream path;
        path << _path_to_dir << "/" << _map_operator_type(op_type) << ".csv";
        csv_writers_per_operator.emplace(op_type, std::make_shared<CSVWriter>(CSVWriter(path.str(), *headers)));
      }
```

```cpp
 const std::string _map_operator_type(const OperatorType op_type) {
    switch (op_type) {
      [...]
      case OperatorType::TableScan:
        return "TableScan";
      [...]
    }
```
Based on this implementation I need to use /TableScan.csv here.

src/test/calibrationlib/operator_feature_exporter_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+                             {100000},
+                             {100}});
+
+    const auto table_generator = CalibrationTableGenerator(table_config);
+    _table = table_generator.generate().at(0);
+    Hyrise::get().storage_manager.add_table(_table->get_name(), _table->get_table());
+  }
+
+  ~OperatorFeatureExporterTest() override {
+    Hyrise::get().storage_manager.drop_table(_table->get_name());
```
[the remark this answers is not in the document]
boehmchen: I removed this line

src/test/calibrationlib/operator_feature_exporter_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+  std::shared_ptr<const CalibrationTableWrapper> _table;
+
+  std::string _dir_path = (std::filesystem::temp_directory_path() / "calibrationTest").string();
+  OperatorFeatureExporter _feature_exporter = OperatorFeatureExporter(_dir_path);
+};
+
+// Check if performance data is added to csv.
+// This does not check if the export of specific table_scan implementations works. (e.g. Table)
+TEST_F(OperatorFeatureExporterTest, TableScanExport) {
+  const auto headers = std::vector<std::string>({"INPUT_ROWS_LEFT", "OUTPUT_ROWS", "RUNTIME_NS", "SCAN_TYPE",
```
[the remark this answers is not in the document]
boehmchen: Yes, it is. The function to get this vector is private.

Should I mark the _get_header() function from OperatorFeatureExport as public because we need it for tests? (We do not use it anywhere else)

src/test/calibrationlib/csv_writer_test.cpp
boehmchen: Ok.

src/test/calibrationlib/operator_feature_exporter_test.cpp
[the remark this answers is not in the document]
boehmchen: This might change if you add magic_enums.

src/test/calibrationlib/operator_feature_exporter_test.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "calibration_table_generator.hpp"
+#include "operator_feature_exporter.hpp"
+#include "storage/table.hpp"
+
+namespace opossum {
+
+class OperatorFeatureExporterTest : public BaseTest {
+ protected:
+  void SetUp() override {
+    auto table_config = std::make_shared<TableGeneratorConfig>(
```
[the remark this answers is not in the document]
boehmchen: True. I had plans to create more complex tests which depended on the Calibration Table Generation, but this is out of scope and not as useful.

Therefore, I changed the table creation with following code

```cpp
const auto table = load_table("resources/test_data/tbl/float_int.tbl", 2);
_table = std::make_shared<CalibrationTableWrapper>(table, "float_int");
```

boehmchen:
I commented and resolved all conversations on files/ code I wrote initially.

src/test/calibrationlib/operator_feature_exporter_test.cpp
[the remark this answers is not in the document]
boehmchen: I changed this for table_feature_exporter_test.cpp, too

src/calibrationlib/calibration_benchmark_runner.cpp
boehmchen: I removed the TCP-DS Benchmark for now.

src/calibrationlib/calibration_benchmark_runner.cpp
boehmchen: Yes, it is, but it doesn't matter for this PR anymore because I removed TPC-DS.

src/calibrationlib/operator_feature_export.cpp
Bouncner: Definitely something to test for or Assert that the previous entry includes "Impl:".

src/test/calibrationlib/operator_feature_exporter_test.cpp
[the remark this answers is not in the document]
Bouncner: I don't see a value in having this header as a private member. And there is no need for the getter. Why not just a public const member?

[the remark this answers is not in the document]
boehmchen: I did this because I followed the rule "Only expose what you really need". Nevertheless, I implemented the headers as const member, now.

src/calibrationlib/calibration_benchmark_runner.cpp
boehmchen: I removed TPC-DS and Jobs Benchmark.

src/calibrationlib/operator_feature_export.cpp
boehmchen: I added an Assert before I set the SCAN_IMPLEMENTATION value.

src/calibrationlib/calibration_lqp_generator.cpp
```diff
@@ -0,0 +1,163 @@
...
+    case OperatorType::TableScan:
+      _generate_table_scans(table);
+      return;
+    default:
+      break;
+  }
+  throw std::runtime_error("Not implemented yet: Only TableScans and JoinHashes are currently supported");
+}
+
+void CalibrationLQPGenerator::_generate_table_scans(const std::shared_ptr<const CalibrationTableWrapper>& table) {
```
Bouncner, line 31: table > table_wrapper?
You access the actual table by `table->get_table()`. Confusing. Maybe `table_wrapper->get_table()`?

src/calibrationlib/calibration_lqp_generator.cpp
```diff
@@ -0,0 +1,163 @@
...
+        // the cursor is a int representation of where the column has to be cut in this iteration
+        const auto step_cursor = static_cast<int>(selectivity_step * step_size);
+
+        const ColumnDataType lower_bound = SyntheticTableGenerator::generate_value<ColumnDataType>(step_cursor);
+
+        auto get_predicate_node_based_on = [column, lower_bound](const std::shared_ptr<AbstractLQPNode>& base) {
+          return PredicateNode::make(greater_than_(column, lower_bound), base);
+        };
+
+        // Baseline
```
Bouncner, line 67: Baseline?

src/calibrationlib/calibration_lqp_generator.cpp
```diff
@@ -0,0 +1,163 @@
...
+        _generated_lpqs.emplace_back(get_predicate_node_based_on(std::shared_ptr<AbstractLQPNode>(stored_table_node)));
+
+        // Add reference scans
+        if (_enable_reference_scans) {
+          // generate reference scans that are executed before the actual predicate
```
Bouncner, line 72: I don't get that.

src/calibrationlib/calibration_lqp_generator.cpp
```diff
@@ -0,0 +1,163 @@
...
+       */
+  const auto column_definitions = table->get_table()->column_definitions();
+  auto column_comparison_pairs = std::vector<ColumnPair>();
+
+  auto singles = std::vector<TableColumnDefinition>();
+
+  for (const TableColumnDefinition& column : column_definitions) {
+    bool matched = false;
+    auto single_iterator = singles.begin();
+    while (single_iterator < singles.end()) {
```
Bouncner, line 130: This is a classic for loop. And for Jan's point: switch to index-based iteration?

src/calibrationlib/calibration_table_generator.cpp
```diff
@@ -0,0 +1,47 @@
+#include "calibration_table_generator.hpp"
+#include "storage/table.hpp"
+
+namespace opossum {
+
+CalibrationTableGenerator::CalibrationTableGenerator(std::shared_ptr<TableGeneratorConfig> config) {
+  // Generate all possible permutations of column types
+  for (DataType data_type : config->data_types) {
```
Bouncner, line 8: const

src/calibrationlib/calibration_table_generator.cpp
```diff
@@ -0,0 +1,47 @@
...
+    for (EncodingType encoding_type : config->encoding_types) {
+      if (encoding_supports_data_type(encoding_type, data_type)) {
+        for (ColumnDataDistribution column_data_distribution : config->column_data_distributions) {
```
Bouncner, line 11: auto

src/calibrationlib/calibration_table_generator.cpp
```diff
@@ -0,0 +1,47 @@
...
+  _config = config;
+}
+
+std::vector<std::shared_ptr<const CalibrationTableWrapper>> CalibrationTableGenerator::generate() const {
+  auto table_wrappers = std::vector<std::shared_ptr<const CalibrationTableWrapper>>();
+  table_wrappers.reserve(_config->chunk_sizes.size() * _config->row_counts.size());
+
+  auto table_generator = std::make_shared<SyntheticTableGenerator>();
+
+  for (int chunk_size : _config->chunk_sizes) {
```
Bouncner, line 34: No int, auto.

src/calibrationlib/calibration_table_wrapper.cpp
```diff
@@ -0,0 +1,22 @@
+#include "calibration_table_wrapper.hpp"
+#include <storage/base_encoded_segment.hpp>
+
+namespace opossum {
+CalibrationTableWrapper::CalibrationTableWrapper(
+    const std::shared_ptr<Table> table, const std::string& table_name,
+    const std::vector<ColumnDataDistribution> column_data_distribution_collection)
+    : _table(table), _name(table_name), _column_data_distribution_collection(column_data_distribution_collection) {
+  assert(table->column_count() == column_data_distribution_collection.size());
```
Bouncner, line 9: As that "our" assert?

Bouncner: I think we use "Assert" which takes an error message.

src/calibrationlib/calibration_table_wrapper.cpp
```diff
@@ -0,0 +1,22 @@
...
+}
+
+CalibrationTableWrapper::CalibrationTableWrapper(const std::shared_ptr<Table> table, const std::string& table_name)
+    : _table(table), _name(table_name), _column_data_distribution_collection() {}
+
+const std::shared_ptr<Table> CalibrationTableWrapper::get_table() const { return _table; }
+
+const std::string& CalibrationTableWrapper::get_name() const { return _name; }
+
+const ColumnDataDistribution CalibrationTableWrapper::get_column_data_distribution(ColumnID id) const {
```
Bouncner, line 19: Looks like a boiler plate class. Wouldn't a method-free struct suffice?

src/calibrationlib/csv_writer.cpp
```diff
@@ -0,0 +1,114 @@
...
+#include <fstream>
+
+#include <boost/algorithm/string.hpp>
+
+namespace opossum {
+
+CSVWriter::CSVWriter(
+    const std::string file_path,
+    const std::vector<std::string>
+        headers,  // headers for csv file, if the file for file_path already exists, it must have equal headers
```
Bouncner, line 16: Does the linter actually check your source code folders? Can you add a `TODO: test` and check please? Such a todo should fail in the linter.

src/calibrationlib/csv_writer.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+                  if (to_test_headers == _headers) {
+                    correct_header = true;
+                  }
+                }
+                check_file.close();
+                return correct_header;
+              })(),
+              "\nCSV Writer ERROR: Provided headers for file '" + _file_path +
+                  "' does not match with existing headers. \n" + "Please validate headers for this CSVFile.");
```
Bouncner, line 60: "CSV file"

Looks like 80% of that method is a DebugAssert. Sure that's a good idea?

src/calibrationlib/calibration_table_generator.cpp
boehmchen: I changed all loop types to const auto

src/calibrationlib/calibration_table_generator.cpp
boehmchen: See above. I changed this to const auto.

src/calibrationlib/operator_feature_exporter.hpp
```diff
@@ -0,0 +1,115 @@
...
+#include <operators/abstract_operator.hpp>
+#include "csv_writer.hpp"
+
+namespace opossum {
+
+class OperatorFeatureExporter {
+ public:
+  explicit OperatorFeatureExporter(const std::string& path_to_dir);
+
+  void export_to_csv(std::shared_ptr<const AbstractOperator> op) const;
```
Bouncner, line 14: const

src/calibrationlib/calibration_table_generator.cpp
boehmchen: Changed this to const auto, too.

src/test/calibrationlib/calibration_lqp_generator_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <synthetic_table_generator.hpp>
+#include "storage/table.hpp"
+
+#include "calibration_lqp_generator.hpp"
+#include "calibration_table_generator.hpp"
+
+namespace opossum {
+class CalibrationLQPGeneratorTest : public BaseTest {
+ protected:
+  void SetUp() override {
```
Bouncner, line 13: Looks like all tables are created for each and every test. How long does that test suite run?

python/cost_models.py
Bouncner: As it is a quick change and we already use it in pretty much all of our python scripts, I'd say we should switch here as well.

python/cost_models.py
```diff
@@ -0,0 +1,220 @@
...
+    # value (cf. Equation (2.14)). By that, LnQ shows structural under- or over- estimation, which may show the quality
+    # of a tested model. A drawback of LnQ is the difficulty in interpreting error values. Doubling the quotient does
+    # not consistently affect the error value. Due to the logarithmic function, the measure penalizes small quotients
+    # relatively hard, whereas it treats large residuals comparably easy.
+    LNQ = 1/len(y_true) * np.sum(np.exp(np.divide(y_pred, y_true)))
+
+    # Percentage
+    perc = np.mean(100 * (np.divide(np.abs(y_true - y_pred), y_true)))
+
+    scores = {'MSE': '%.3f' % mse, 'R2': '%.3f' % R2, 'LNQ': '%.3f' % LNQ, 'Percentage': '%.3f' % perc}
```
Bouncner, line 96: Percentage is MAPE metric?

python/cost_models.py
```diff
@@ -0,0 +1,220 @@
...
+        train_data, test_data = import_data(args)
+    else:
+        train_data = import_train_data(args.train[0])
+        train_data = train_data.dropna()
+        train_data, test_data = train_test_split(train_data)
+
+    if args.out:
+        out = args.out
+
+    if not os.path.exists("{}/Models".format(out)):
```
Bouncner, line 172: Better avoid uppercase letters in folder names.

src/benchmarklib/synthetic_table_generator.cpp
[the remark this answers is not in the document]
Bouncner: Do you use it in multiple places? If not, better avoid defining such "strange" output names globally and use your "string coding" only within your functions.

src/calibrationlib/csv_writer.cpp
boehmchen: The linter triggers on TODO: test

Bouncner: And did you format and that changed something? Maybe format does not include the correct folders?

src/calibrationlib/csv_writer.cpp
boehmchen: The DebugAssert checks whether the headers of an existing file equals the headers provided in the constructor of the class. 

To validate that, I have to do the following steps

- Open the file
- Read the first line
- Comparing headers. This can be done by:
  - Split the line into values and comparing those with the existing vector of headers
  - OR: comparing the entire string of the first line with a computed header line
- Close File

The comparing part does need some additional code in both cases. If you have any idea to shorten this DebugAssert, please let me know.

src/calibrationlib/csv_writer.cpp
boehmchen: The formatter works fine too. Actually I think the formatter added the line break after `std::vector<std::string>` because the following comment is too long

src/calibrationlib/csv_writer.cpp
Bouncner: 1. What is the scenario where we already have a file with the same header and do not want to overwrite it? Do we want to append to this file?
2. Why should that code only be relevant for Debug builds?

Bouncner: And what are "provided headers" for a file?

boehmchen: 1. Yep, this is relevant if we want to append data points to the file. (I have to admit that this is not the case at the moment, but I thought this might be useful in the future, which is why I added it)
2. Good question! I struggled with this decision. The main thought behind it was that the calibration is tested in debug mode beforehand. Asserts, which are not important should be skipped in the release build to improve performance. I agree, that this DebugAssert might be useful in release builds. What do you think?

Provided headers --> `headers` parameter of the CSVWriter constructor

```cpp
 CSVWriter(const std::string file_path, const std::vector<std::string> headers, const bool override_file = true);

```

PS: I wrote DebugAssert's for this class mainly because I ran into a lot of corrupted CSV files because values not being added mainly due to if statements without a default value.

src/calibrationlib/calibration_table_wrapper.cpp
boehmchen: That's another assert. The LQP generation needs a ColumnDataDistribution per Column to generate useful predicates.

src/calibrationlib/calibration_table_wrapper.cpp
boehmchen: I think, if we drop the [assert above](https://github.com/hyrise/hyrise/pull/2009#discussion_r405025220), yes.

src/benchmarklib/synthetic_table_generator.cpp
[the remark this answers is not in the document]
boehmchen: I use it for column name generation only. I decided to remove the DataDistributionType in the column name completely, because we do not use it anywhere at the moment. To ensure uniqueness of column names, I added a simple counter.

A changed the related code in `calibration_table_generator.cpp` to:
```cpp
auto column_count = 0; // Needed to make column names unique
for (const auto& column_data_distribution : config->column_data_distributions) {
  std::stringstream column_name_stringstream;
  column_name_stringstream << data_type << "_" << encoding_type << "_" << column_count;

  auto column_name = column_name_stringstream.str();

  _column_data_distributions.emplace_back(column_data_distribution);
  _column_specs.emplace_back(
  ColumnSpecification(column_data_distribution, data_type, encoding_type, column_name));

  column_count++;
}
```

Therefore, we do not need this change anymore. If you feel like, it might be useful to have the data distribution type encoded in the column name, leave a comment, please.

src/calibrationlib/operator_feature_exporter.hpp
boehmchen: Added const to all std::shared_ptr's for the class OperatorFeatureExporter

src/calibrationlib/csv_writer.cpp
Bouncner: In order to ease the code and ensure that everything that should be supported is actually implemented and tested, I'd vote for removing all capabilities to append to CSVs.
It's all or nothing.

boehmchen: Appending is tested and implemented, just not used in other classes. Nevertheless, I removed the feature to append to existing files in [
315d2e0](https://github.com/hyrise/hyrise/pull/2009/commits/315d2e08b2aaf30628adb3d684c4bf09ad198118)

python/cost_models.py
fefschumann: Yes, it is, I will rename it

python/cost_models.py
fefschumann: Done

run_calibration.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -x
+
+BASE_PATH='./cmake-build-debug'
```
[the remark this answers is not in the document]
fefschumann: Done

python/cost_models.py
```diff
@@ -0,0 +1,220 @@
...
+        diff1 = np.setdiff1d(train[col].unique(), test[col].unique())
+        diff2 = np.setdiff1d(test[col].unique(), train[col].unique())
+        for d1 in diff1:
+            test['{}_{}'.format(col, d1)] = 0
+        for d2 in diff2:
+            train['{}_{}'.format(col, d2)]= 0
+    return [train, test]
+
+
+def parseargs(opt=None):
```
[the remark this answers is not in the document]
fefschumann: I changed the function name

python/cost_models.py
```diff
@@ -0,0 +1,220 @@
...
+
+def calculate_error(test_X, y_true, y_pred, model):
+    # calculate error (ME) for the model
+    mse = mean_squared_error(y_true, y_pred, squared=False)
+
+    # The score function returns the coefficient of determination R^2 of the prediction.
+    # The coefficient R^2 is defined as (1 - u/v), where u is the residual sum of squares ((y_true - y_pred) ** 2).sum()
+    # and v is the total sum of squares ((y_true - y_true.mean()) ** 2).sum(). The best possible score is 1.0 and it can
+    # be negative (because the model can be arbitrarily worse). A constant model that always predicts the expected value
+    # of y, disregarding the input features, would get a R^2 score of 0.0.
```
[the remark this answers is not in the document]
fefschumann: Changed the comment

python/cost_models.py
```diff
@@ -0,0 +1,220 @@
...
+    R2 = model.score(test_X, y_pred)
+
+    # Laut Master Thesis:
+    # logarithm of the accuracy ratio (LnQ). Tofallis et al. define this measure to overcome the limitations of RMSE and
+    # MAPE by providing a unbiased error measure [72]. LnQ is the natural logarithm of the quotient of the predicted
```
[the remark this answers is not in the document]
fefschumann: It's gone

python/cost_models.py
```diff
@@ -0,0 +1,220 @@
...
+    output_path = '{}/Plots/{}_{}_{}'.format(out, method, encoding, scan)
+    plt.savefig(output_path, bbox_inches='tight')
+    plt.close()
+
+    return model_scores
+
+
+def calculate_error(test_X, y_true, y_pred, model):
+    # calculate error (ME) for the model
+    mse = mean_squared_error(y_true, y_pred, squared=False)
```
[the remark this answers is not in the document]
fefschumann: It is the RMSE, I changed the name

python/cost_models.py
```diff
@@ -0,0 +1,220 @@
...
+
+    # The score function returns the coefficient of determination R^2 of the prediction.
+    # The coefficient R^2 is defined as (1 - u/v), where u is the residual sum of squares ((y_true - y_pred) ** 2).sum()
+    # and v is the total sum of squares ((y_true - y_true.mean()) ** 2).sum(). The best possible score is 1.0 and it can
+    # be negative (because the model can be arbitrarily worse). A constant model that always predicts the expected value
+    # of y, disregarding the input features, would get a R^2 score of 0.0.
+    R2 = model.score(test_X, y_pred)
+
+    # Laut Master Thesis:
```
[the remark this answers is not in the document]
fefschumann: I removed that comment in the code

python/cost_models.py
fefschumann: Changed to f-strings

src/calibrationlib/calibration_lqp_generator.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+namespace opossum {
+void CalibrationLQPGenerator::generate(OperatorType operator_type,
+                                       const std::shared_ptr<const CalibrationTableWrapper>& table) {
+  switch (operator_type) {
+    case OperatorType::TableScan:
+      _generate_table_scans(table);
+      return;
+    default:
+      break;
```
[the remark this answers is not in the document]
ramboldio: done

ramboldio:
Re: https://github.com/hyrise/hyrise/pull/2009#discussion_r405023770
I need a classic loop here since I need to remove the element from the vector I'm iterating over (ie. when I matched with another column that column is no longer unmatched). I couldn't find a nicer way of doing that. Maybe there is a way of doing that using range loops?

ramboldio:
Another PR that I came up with while working on the project:
https://github.com/hyrise/hyrise/pull/2081

ramboldio:
An issue that I needed some time to discover: https://github.com/hyrise/hyrise/issues/2082
Don't quite know whether it's a bug tho.

README.md
```diff
@@ -142,6 +142,7 @@ Contact: firstname.lastname@hpi.de
 -   Julian    Menzler
 -   Torben    Meyer
 -   Leander   Neiß
+-   Lukas     Rambold
```
Bouncner, line 145: It's completely up to you, but I am surprised to see only one contributor here.

Bensk1:
As discussed, closing for now but we will continue to work on that and reopen it later.

PR #2013, opened 2020-02-10
Fix recreation of CreateTable and JoinNestedLoop

### LINKED ISSUE #1994: Running CREATE TABLE IF NOT EXIST twice causes a crash
Running this twice causes hyrise to crash:
create table if not exists t1 (num INT)

```
./hyriseConsole
[PERF] Hyrise is running as a debug build. at src/lib/utils/performance_warning.cpp:9

HYRISE SQL Interface
Type 'help' for more information.

Hyrise is running a (debug) build.

(debug)> create table if not exists t1 (num INT)
=== Columns
|      OK|
|     int|
|not null|
===
0 rows total
Execution info: [PARSE: 9 µs 783 ns, SQL TRANSLATE: 386 µs 164 ns, OPTIMIZE: 1 ms 237 µs, LQP TRANSLATE: 307 µs 494 ns, EXECUTE: 1 ms 540 µs (wall time) | QUERY PLAN CACHE HITS: 0/1 statement(s)]
(debug)> create table if not exists t1 (num INT)
  what():  src/lib/operators/abstract_operator.cpp:31 Operator has already been executed
Aborted (core dumped)
```

mrks:
The DYOD set operations group found that JoinNestedLoop drops its secondary predicates when recreated from the PQP cache. A similar issue was found in #1994 for CreateTable.

This PR fixes those two recreations and modifies the SQLite runner to reuse the PQP in the hyriseSystemTest. Hopefully, this helps us identifying those issues earlier. Without static type reflection, I can't think of a good low-level test that makes sure that `_on_deep_copy` is implemented properly.

fixes #1994


src/lib/operators/abstract_operator.cpp
```diff
@@ -28,7 +28,7 @@ void AbstractOperator::execute() {
   DTRACE_PROBE1(HYRISE, OPERATOR_STARTED, name().c_str());
   DebugAssert(!_input_left || _input_left->get_output(), "Left input has not yet been executed");
   DebugAssert(!_input_right || _input_right->get_output(), "Right input has not yet been executed");
-  DebugAssert(!_output, "Operator has already been executed");
+  DebugAssert(_performance_data->walltime.count() == 0, "Operator has already been executed");
```
mrks: This now also covers operators that do not return a table

src/test/sql/sqlite_testrunner/sqlite_testrunner.cpp
```diff
@@ -46,8 +46,11 @@ void SQLiteTestRunner::SetUpTestCase() {
 
   _table_cache_per_encoding.emplace(EncodingType::Unencoded, unencoded_table_cache);
 
-  Hyrise::get().topology.use_numa_topology();
-  Hyrise::get().set_scheduler(std::make_shared<NodeQueueScheduler>());
```
mrks: #2014

mrks:
Trying to distribute the PR workload a bit.

PR #2016, opened 2020-02-10
Some Cleanup Things

### LINKED ISSUE #1983: NULL behaviour in PosList.common_chunk_id()
It is possible that NULL behaviour in `PosList.common_chunk_id()` is faulty similarly to `ReferenceSegmentIterables` in #1421.
If the first value of the PosList references a NULL value, `common_chunk_id()` will return INVALID_CHUNK_ID. Calling code may not be aware of this.

For instance, the ChunkID retrieved in `validate.cpp:69` in  `_validate_chunks()`  would trigger the DebugAssert in `Table::get_chunk()`. The same thing happens in `abstract_dereferenced_column_table_scan_impl.cpp`.

Since guaranteeing to return a valid chunk id might be expensive, it may be ok just to make the caller aware of this problem (e.g. by renaming the function or adding a comment).

(discovered with @He3lixxx and @Xperianer)

mrks:
Couple of things that came up while working on the TPC-H Paper.

- The JSON library was linked pretty much everywhere, but not in the playground. As a result, generating TPC-H data was not possible. I changed its linkage to public. Let me know if you'd prefer for someone else to review the cmake stuff.
- Clarify that PosLists marked as `guarantee_single_chunk` must not contain NULL values, fixes #1983.
- The ChunkPruningRule was executed twice. 
- The Non-FullCI runs did not skip the ODR step.

No relevant performance difference for TPC-H.


src/lib/storage/pos_list.hpp
```diff
@@ -64,10 +64,12 @@ struct PosList final : private pmr_vector<RowID> {
...
             const auto& common_chunk_id = (*this)[0].chunk_id;
-            return std::all_of(cbegin(), cend(),
-                               [&](const auto& row_id) { return row_id.chunk_id == common_chunk_id; });
+            if (common_chunk_id == INVALID_CHUNK_ID) return false;
+            return std::all_of(cbegin(), cend(), [&](const auto& row_id) {
+              return row_id.chunk_id == common_chunk_id && row_id.chunk_offset != INVALID_CHUNK_OFFSET;
+            });
           }(),
-          "Chunk was marked as referencing only a single chunk, but references more");
+          "PosList was marked as referencing a single chunk, but references more or the PosList contains NULL values");
```
j-tr: imo these two checks are rather independent and the null value test has nothing to do with the references_single_chunk property. but having a dedicated check for null values might be too cumbersome as well. 
also: from my understanding, posLists never contain values, so "NULL values" might be confusing. maybe better justin "invalid values"

j-tr:
the cmake stuff looks pretty straight forward but im not an expert for this

src/lib/storage/pos_list.hpp
mrks: PosLists can also contain NULL values. Think of two tables that hold no NULL values being joined using a LEFT JOIN. This may produce NULL values on the right side. Those are not referenceable, as they have just been created. In these cases, the PosList contains NULL values.

Also pinging @Bensk1 because he looked into PosLists lately.

Bensk1: That is related to #1983, isn't it? At some point we could add something like `references_only_NULLs`.

mrks: Yes.

cmake/TargetLinkLibrariesSystem.cmake
```diff
@@ -5,18 +5,7 @@ function(target_link_libraries_system target)
   set(libs ${ARGN})
   foreach(lib ${libs})
     get_target_property(lib_include_dirs ${lib} INTERFACE_INCLUDE_DIRECTORIES)
-    target_include_directories(${target} SYSTEM PRIVATE ${lib_include_dirs})
-    target_link_libraries(${target} ${lib})
+    target_include_directories(${target} SYSTEM PUBLIC ${lib_include_dirs})
```
Bouncner: Do you know what the effects of public include dirs are? Not getting what you changed here.

src/lib/CMakeLists.txt
```diff
@@ -663,7 +663,7 @@ set_source_files_properties(
 # Add our jemalloc build as an interface library so that the build process of hyrise does not have to wait for it
 # -rdynamic tells the linker to export the library's symbols so that plugins can use them.
 target_link_libraries(hyrise_impl PUBLIC ${LIBRARIES} INTERFACE custom_jemalloc -rdynamic)
-target_link_libraries_system_private(hyrise_impl nlohmann_json::nlohmann_json)
```
Bouncner: I don't get what this change from private linking to public linking for `nlohmann_json` does. The description does not help me. The removed line with `private` means that other components linking to `hyrise_impl` do nonetheless not see the linked `nlohmann_json`?

src/lib/storage/pos_list.hpp
```diff
@@ -64,10 +64,12 @@ struct PosList final : private pmr_vector<RowID> {
           [&]() {
             if (size() == 0) return true;
             const auto& common_chunk_id = (*this)[0].chunk_id;
-            return std::all_of(cbegin(), cend(),
-                               [&](const auto& row_id) { return row_id.chunk_id == common_chunk_id; });
+            if (common_chunk_id == INVALID_CHUNK_ID) return false;
+            return std::all_of(cbegin(), cend(), [&](const auto& row_id) {
```
Bouncner: What is the problem with NULL values? Why don't we want them?

cmake/TargetLinkLibrariesSystem.cmake
mrks: PUBLIC makes it visible to whoever includes `target`. As such, whoever includes `hyrise_impl` automatically gets the json library.

src/lib/CMakeLists.txt
mrks: Note that the line was not removed, it was just changed from private to public (see my last reply)

Bouncner: Yeah I know, I think was suprised me was that you can “hide” inclusion/libs for “transitive” includes.

src/lib/storage/pos_list.hpp
mrks: Above, we say

> If all entries in the PosList shares a single ChunkID, it makes sense to explicitly give this guarantee in order to enable some optimizations.

Does changing it to

> If we know that all entries in the PosList share a single ChunkID, we can optimize the indirection by retrieving the respective chunk once and using only the ChunkOffsets to access values. As the consumer will likely call `table->get_chunk(common_chunk_id)`, we require that the common chunk id is valid, i.e., the PosList contains no NULL values.

help?

Bouncner: That does indeed help. Will resolve this issue.

Just for my understanding: is that something that operators actually require right now or is it a preparation for PosList optimization (e.g., storing no chunk_id for every row)?
I'd guess operators wouldn't mind NULL values even when using optimizations for such single-chunk-ref chunks?!??

mrks: We are not talking about NULL values in, e.g., the DictionarySegment being referenced, but about NULL values in the PosList itself. Currently this does not happen, because no operator that guarantees the single chunk property sets such NULL values, but when I played around with it for other operators, I ran into this issue. To avoid other people setting the property falsely, which may break seemingly unrelated downstream operators, I added this check.

mrks:
I sneaked in one more thing: On Mac, we erroneously did not build Hyrise as a shared library. Doing so reduces the time of a rebuild from 0m15.511s to 0m6.861s.

PR #2020, opened 2020-02-12
[DYOD] Add more micro benchmarks for sort operator

birneamstiel:
This PR adds micro benchmarks for the sort operator which are parametrised with regard to:
- row count
- multiple column sort
- null value ratio
- segment type
- column data types

This PR depends on the multi column sort operator PR #2005


src/benchmark/operators/sort_benchmark.cpp
```diff
@@ -3,20 +3,113 @@
 #include "benchmark/benchmark.h"
 
 #include "../micro_benchmark_basic_fixture.hpp"
+#include "SQLParser.h"
+#include "SQLParserResult.h"
```
mrks: Where is this needed?

src/benchmark/operators/sort_benchmark.cpp
```diff
@@ -3,20 +3,113 @@
...
+
+#include "micro_benchmark_utils.hpp"
 
 namespace opossum {
 
-BENCHMARK_F(MicroBenchmarkBasicFixture, BM_Sort)(benchmark::State& state) {
-  _clear_cache();
+static std::shared_ptr<Table> generate_custom_table(const size_t row_count, const ChunkOffset chunk_size,
+                                                    const DataType data_type = DataType::Int,
+                                                    const std::optional<float> null_ratio = std::nullopt) {
```
mrks: Is it necessary to make this optional? Wouldn't `null_ratio = 0.0f` serve the same purpose?

src/benchmark/operators/sort_benchmark.cpp
```diff
@@ -3,20 +3,113 @@
...
+  const auto table_generator = std::make_shared<SyntheticTableGenerator>();
+
+  size_t num_columns = 2;
```
mrks: const?

src/benchmark/operators/sort_benchmark.cpp
```diff
@@ -3,20 +3,113 @@
...
+
+  const int max_different_value = 10'000;
```
mrks: What is it? The largest value or the distinct count?

src/benchmark/operators/sort_benchmark.cpp
```diff
@@ -3,20 +3,113 @@
...
+  const std::vector<DataType> column_data_types = {num_columns, data_type};
+
+  std::vector<ColumnSpecification> column_specifications = std::vector(
+      num_columns, ColumnSpecification(ColumnDataDistribution::make_uniform_config(0.0, max_different_value), data_type,
+                                       {EncodingType::Unencoded}, std::nullopt, null_ratio));
+
+  return table_generator->generate_table(column_specifications, row_count, chunk_size, UseMvcc::Yes);
```
mrks: Any reason you are generating an MVCC table?

src/benchmark/operators/sort_benchmark.cpp
```diff
@@ -3,20 +3,113 @@
...
+  auto limit =
+      std::make_shared<Limit>(table_wrapper, expression_functional::to_expression(std::numeric_limits<int64_t>::max()));
+  limit->execute();
+  table_wrapper = std::make_shared<TableWrapper>(limit->get_output());
+  table_wrapper->execute();
+}
+
+static void BM_Sort(benchmark::State& state, const size_t row_count = 40'000, const ChunkOffset chunk_size = 2'000,
+                    const DataType data_type = DataType::Int, const std::optional<float> null_ratio = std::nullopt,
+                    const bool multi_column_sort = true, const bool use_reference_segment = false) {
```
mrks: No need to change this, just a suggestion for the future: If all arguments have defaults, it might be cleaner to just pass a struct. That way, it is easier to just set the parameters that you want to vary without having to check the parameter order.

src/benchmark/operators/sort_benchmark.cpp
```diff
@@ -3,20 +3,113 @@
...
+  return table_generator->generate_table(column_specifications, row_count, chunk_size, UseMvcc::Yes);
+}
+
+static void make_reference_table(std::shared_ptr<TableWrapper> table_wrapper) {
+  auto limit =
+      std::make_shared<Limit>(table_wrapper, expression_functional::to_expression(std::numeric_limits<int64_t>::max()));
+  limit->execute();
+  table_wrapper = std::make_shared<TableWrapper>(limit->get_output());
+  table_wrapper->execute();
+}
```
mrks: What is this supposed to do? It doesn't return anything and doesn't have any externally observable side effects. My guess is that you wanted to modify `table_wrapper`, but for that, you would have had to pass in a reference. However, what do you need the new table wrapper for? Can't you simply return the limit operator?

src/benchmark/operators/sort_benchmark.cpp
```diff
@@ -3,20 +3,113 @@
...
+      table_wrapper, std::vector<SortColumnDefinition>{SortColumnDefinition{ColumnID{0}, OrderByMode::Ascending}}, 2u);
   warm_up->execute();
   for (auto _ : state) {
-    auto sort = std::make_shared<Sort>(_table_wrapper_a, ColumnID{0} /* "a" */, OrderByMode::Ascending);
-    sort->execute();
+    if (multi_column_sort) {
+      const auto sort_definitions =
+          std::vector<SortColumnDefinition>{{SortColumnDefinition{ColumnID{0}, OrderByMode::Ascending},
+                                             SortColumnDefinition{ColumnID{1}, OrderByMode::Descending}}};
+      auto sort = std::make_shared<Sort>(table_wrapper, sort_definitions, 2u);
```
mrks: Why is 2 a reasonable output chunk size for the benchmarks?

src/lib/operators/sort.cpp
```diff
@@ -33,19 +20,78 @@ const std::string& Sort::name() const {
...
+        // columns -- on their own -- as well).
+        const auto chunk_count = output_mutable->chunk_count();
+        for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+          output_mutable->get_chunk(chunk_id)->set_ordered_by(
+              std::make_pair(sort_definition.column, sort_definition.order_by_mode));
+        }
+        output = output_mutable;
+      } else {
+        previously_sorted_pos_list.reset();
+        previously_sorted_pos_list = std::make_shared<PosList>();
```
Boffmann: Regarding std::shared_ptr::reset:
`For signature (1) the object becomes empty (as if default-constructed).`

So I guess, one of these lines (71 or 72) is enough

Source: [CPP-reference](https://www.cplusplus.com/reference/memory/shared_ptr/reset/)

mrks:
Will integrate into #2005.

PR #2021, opened 2020-02-12
[DYOD] Add parallelization to Sort operator

bakoe:
This is part 4/4 of the planned PRs from 2019's DYOD “Sort operator” group.

> - Adjustments to the SyntheticTableGenerator (PR #1987, already merged)
> - Extension of Sort Operator and LQP Translator to allow multi column sorting in one operator instance (without micro benchmark adjustments) (PR #2005)
> - Extension of the Sort operator’s micro benchmarks incl. NULL ratios created by the updated SyntheticTableGenerator and multi column benchmarks utilizing the updated Sort operator interface from #2005 (PR #2020)
> - **Enabling of parallelization in the Sort operator (to-be-created from our fork’s corresponding branch) (this PR here)** 👈

Accordingly, this PR depends on both #2005 and #2020.

This PR extends the multi-column capable Sort operator with **parallelization**, which allows for better performance in e. g. the TPC-H benchmarks or the Micro Benchmarks (which have been extended in #2020).

### To Do 

- [x] Resolve the TODO comment in sort.cpp regarding code duplication
- [ ] Resolve the TODO comment in pos_list_utils.hpp regarding missing doc comment
- [ ] Add the TPC-H, TPC-DS and Micro Benchmark results (from [our fork's PR ](https://github.com/bakoe/hyrise/pull/13)or newly created) to this PR's description
- [ ] Resolve the Clarification points stated below

### What needs clarification

- At the final presentations, the assumption was raised that our current implementation of merging the parts of the input which have been sorted in parallel (see `MergeImpl`) leads to a decreased performance when sorting tables with a considerable amount of chunks (because currently for e. g. 10000 chunks 10000 elements have to be compared at each iteration of the merging loop). This occurs e. g. in the TPC-H benchmark's warm-up, when the input gets pre-sorted. 
    👉 It should be clarified whether we try out another implementation of the merging step, e. g. using a MergeSort-esque way of recursively merging two parts (maybe involving parallelization as well). 
- We planned to utilize the `sorted_by` information stored in Chunks in the Sort implementation as well – it needs to be decided whether this should happen in this scope or should be addressed later.


Bouncner:
> It should be clarified whether we try out another implementation of the merging step, e. g. using a MergeSort-esque way of recursively merging two parts (maybe involving parallelization as well).

Feel free to approach me for this discussion. I've been working quite a lot on the sort merge join where we faced similar issues (and still do).
To start such a discussion, we would need a simple estimation of the number of operations you are going to do for (let's say) N chunks and M rows per chunk.

For a merge-sort-like approach, you are always iterating two lists (2*M), but one of them ( or both, depending how to determine merge partners) of them will grow over time.
I think that led to something like N*((N/2)*M + M). Can you give a similar estimate for your current approach?

The interesting aspect for the sort-merge part in the sort-merge-join was that is was beneficial to sometimes do not a merging (even though lists have been pre-sorted) but concatenate lists and sort them.

birneamstiel:
> To start such a discussion, we would need a simple estimation of the number of operations you are going to do for (let's say) N chunks and M rows per chunk.

> For a merge-sort-like approach, you are always iterating two lists (2M), but one of them ( or both, depending how to determine merge partners) of them will grow over time.
I think that led to something like N((N/2)*M + M). Can you give a similar estimate for your current approach?

In our current implementation we're iterating over over N lists (`N * M`) and compare with one element of each chunk which results in `N * (N * M) = N^2 * M`. This complexity is indeed highly dependent on the amount of chunks which will lead to rather bad performance when the chunk size is small.

Bouncner:
https://docs.google.com/spreadsheets/d/1gEaZSIXV2lCP6epdjvpi4I3fOgSE_FgJivVPS1dUJDA/edit#gid=0

mrks:
As we haven't heard anything from a while and won't find the time to work on the open points ourselves, we decided to close this PR for now.

PR #2022, opened 2020-02-13
Assert column_expressions / operator output match

### LINKED ISSUE #1960: Assert column_expressions / operator output match.
Now that an operator knows its LQP node, we should assert that the output expressions/columns match.

mrks:
Check that LQP expressions and PQP columns match. If they do not, this is a severe bug as the operators might be operating on the wrong column.

fixes #1960


src/test/sql/sqlite_testrunner/sqlite_testrunner.cpp
```diff
@@ -81,6 +81,14 @@ void SQLiteTestRunner::SetUp() {
 
   auto& table_cache = table_cache_iter->second;
 
+  // In case the previous SQL query was not executed successfully, we always reset all tables
+  // because we cannot be sure that the dirty flags have been properly set.
+  for (auto& [_, table_cache_entry] : table_cache) {
+    if (!_last_run_successful) {
```
j-tr: maybe move if statement outside of the loop

mrks: Yepp, thanks

mrks:
Ready to go?

PR #2025, opened 2020-02-17
[DYOD] Smart Position Lists / adding EntireChunkPosList

XPerianer:


XPerianer:
This is a test

mrks:
Please ping your partner group if you have not already done so.

mrks:
Any idea on when the review will be done?

src/lib/storage/segment_iterables/base_segment_iterators.hpp
```diff
@@ -38,7 +38,16 @@ namespace opossum {
...
-class BaseSegmentIterator : public boost::iterator_facade<Derived, Value, boost::random_access_traversal_tag, Value> {};
+class BaseSegmentIterator
+    : public boost::iterator_facade<Derived, Value, boost::random_access_traversal_tag, Value, std::ptrdiff_t> {
+ public:
+  // boost will not use the random_access_iterator_tag if reference_type is not a c++ reference (which it isn't here)
+  // we still want to use random access (for binary search, distance, ...)
+  /* typedef std::random_access_iterator_tag iterator_category; */
+  // TODO
+  // this makes some tests segfault. We think it's due to some bad distance_to implementation, but we're not sure.
+  // It doesn't give a noticable performance boost, so leaving it out might also be fine?
```
He3lixxx: @Bensk1 Boost will not use the random access iterator tag here, even if we pass `random_access_traversal_tag` to the `iterator_facade`, due to the logic documented [here](https://www.boost.org/doc/libs/1_72_0/libs/iterator/doc/iterator_facade.html) (search for "where iterator-category is defined as follows").
However, in our last benchmarks, we couldn't see a difference between adding this here and using the default. (Keep in mind: The AbstractPosList will use `random_access_iterator_tag` no matter what we set here. We just thought it would make sense to also set the random access tag here.)

Thus, we'd suggest that we simply add a comment linking to the boost documentation and explaining that this Iterator will not be treated as a random access iterator by the STL. Does this sound good / should we do something else?

Bensk1: It's should be no news to us that boost does not use the random access iterator tag here. During our last meeting we were only discussing the use of `/* typedef std::random_access_iterator_tag iterator_category; */` Which tests are segfaulting? @Bouncner might be interested in that. 

Your proposed solution is alright.

He3lixxx: Hmm, turns out the current code doesn't segfault anymore when putting the tag in. I guess some changes on master fixed this?

I guess we'll just put in the tag then and remove the TODO?

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
```diff
@@ -240,7 +240,9 @@ TEST_F(PredicateReorderingTest, PredicatesWithMultipleOutputs) {
   EXPECT_EQ(predicate_b_node->left_input(), table_node);
 }
 
-TEST_F(PredicateReorderingTest, SimpleValidateReorderingTest) {
+// TODO
+// Currently disabled, when MatchesAll gives better performance we should consider removing the rule
+/* TEST_F(PredicateReorderingTest, SimpleValidateReorderingTest) {
```
He3lixxx: To resolve this, I'd say we compare release performance and then decide which ordering is currently better?

However, strangely, master doesn't compile for me on the VM with release so our branch doesn't either:
```
In file included from /home/Richard.Ebeling/hyrise/src/lib/storage/vector_compression/fixed_size_byte_aligned/fixed_size_byte_aligned_vector.hpp:9,
                 from /home/Richard.Ebeling/hyrise/src/lib/storage/vector_compression/resolve_compressed_vector_type.hpp:7,
                 from /home/Richard.Ebeling/hyrise/src/lib/storage/dictionary_segment/dictionary_segment_iterable.hpp:9,
                 from /home/Richard.Ebeling/hyrise/src/lib/storage/create_iterable_from_segment.ipp:3,
                 from /home/Richard.Ebeling/hyrise/src/lib/storage/create_iterable_from_segment.hpp:82,
                 from /home/Richard.Ebeling/hyrise/src/lib/operators/join_nested_loop.cpp:9:
/home/Richard.Ebeling/hyrise/src/lib/storage/vector_compression/fixed_size_byte_aligned/fixed_size_byte_aligned_decompressor.hpp: In function ‘void {anonymous}::join_two_typed_segments(const BinaryFunctor&, LeftIterator, LeftIterator, RightIterator, RightIterator, opossum::ChunkID, opossum::ChunkID, const opossum::JoinNestedLoop::JoinParams&) [with BinaryFunctor = std::greater_equal<void>; LeftIterator = opossum::AnySegmentIterator<int>; RightIterator = opossum::FrameOfReferenceSegmentIterable<int>::PointAccessIterator<opossum::FixedSizeByteAlignedDecompressor<unsigned char> >]’:
/home/Richard.Ebeling/hyrise/src/lib/storage/vector_compression/fixed_size_byte_aligned/fixed_size_byte_aligned_decompressor.hpp:19:46: error: ‘*((void*)(& right_it)+32).opossum::FixedSizeByteAlignedDecompressor<unsigned char>::_data’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
   19 |   uint32_t get(size_t i) final { return _data[i]; }
      |                                         ~~~~~^
/home/Richard.Ebeling/hyrise/src/lib/operators/join_nested_loop.cpp:51:15: note: ‘*((void*)(& right_it)+32).opossum::FixedSizeByteAlignedDecompressor<unsigned char>::_data’ was declared here
   51 |     for (auto right_it = right_begin; right_it != right_end; ++right_it) {
      |               ^~~~~~~~
```
(GCC, release build, same error repeats 17 times)

He3lixxx:
> Any idea on when the review will be done?

@s4pu @tzagorni @carlaterboven

src/lib/storage/segment_iterables/base_segment_iterators.hpp
Bensk1: @Bouncner, what was our decision on using `std::random_access_iterator_tag`?

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
Bensk1: Strange, master should compile without problems. You did not change any of these files, did you?

Bensk1: 41d70ad904 works for me with GCC (g++ (Ubuntu 9.2.1-9ubuntu2) 9.2.1 20191008) in debug and release, `hyriseTest`. `CXXFLAGS=-fdiagnostics-color=always cmake -DENABLE_NUMA_SUPPORT=Off -DCMAKE_BUILD_TYPE=Release -G Ninja ..`.

He3lixxx: I think I didn't -- but maybe I'm missing something. If you have some time, feel free to
```
sudo su -- Richard.Ebeling
cd ~/hyrise/build-release
make
```
to reproduce on vm-nordes.

src/lib/storage/pos_list.hpp
```diff
@@ -62,7 +63,7 @@ struct PosList final : private pmr_vector<RowID> {
   void guarantee_single_chunk() { _references_single_chunk = true; }
 
   // Returns whether the single ChunkID has been given (not necessarily, if it has been met)
-  bool references_single_chunk() const {
+  bool references_single_chunk() const override {
```
mrks: In the other file, you used final.

src/lib/storage/reference_segment/reference_segment_iterable.hpp
```diff
@@ -29,88 +29,93 @@ class ReferenceSegmentIterable : public SegmentIterable<ReferenceSegmentIterable
...
-    const auto end_it = pos_list->end();
+    resolve_pos_list_type(position_filter, [&functor, &referenced_table, &referenced_column_id](auto pos_list) {
+      const auto begin_it = pos_list->begin();
+      const auto end_it = pos_list->end();
 
-    // If we are guaranteed that the reference segment refers to a single non-NULL chunk, we can do some optimizations.
-    // For example, we can use a single, non-virtual segment accessor instead of having to keep multiple and using
-    // virtual method calls. If begin_it is NULL, chunk_id will be INVALID_CHUNK_ID. Therefore, we skip this case.
+      // If we are guaranteed that the reference segment refers to a single non-NULL chunk, we can do some
+      // optimizations.  For example, we can use a single, non-virtual segment accessor instead of having
```
mrks: Double space

src/lib/storage/segment_iterables.hpp
```diff
@@ -143,8 +143,8 @@ class PointAccessibleSegmentIterable : public SegmentIterable<Derived> {
  public:
   using SegmentIterable<Derived>::with_iterators;  // needed because of “name hiding”
 
-  template <typename Functor>
-  void with_iterators(const std::shared_ptr<const PosList>& position_filter, const Functor& functor) const {
+  template <typename Functor, typename PosListType>
```
mrks: Type should be redundant once we renamed the old PosList

src/lib/storage/segment_iterables.hpp
```diff
@@ -153,10 +153,17 @@ class PointAccessibleSegmentIterable : public SegmentIterable<Derived> {
     }
   }
 
+  // Needed, because otherwise the 2 overloads above are ambiguous
```
mrks: two, also could you add half a sentence on why they are ambiguous?

src/lib/storage/segment_iterables/base_segment_iterators.hpp
```diff
@@ -38,7 +38,13 @@ namespace opossum {
  * };
  */
 template <typename Derived, typename Value>
-class BaseSegmentIterator : public boost::iterator_facade<Derived, Value, boost::random_access_traversal_tag, Value> {};
+class BaseSegmentIterator
+    : public boost::iterator_facade<Derived, Value, boost::random_access_traversal_tag, Value, std::ptrdiff_t> {
+ public:
+  // boost will not use the random_access_iterator_tag if reference_type is not a c++ reference (which it isn't here)
+  // we still want to use random access (for binary search, distance, ...)
+  typedef std::random_access_iterator_tag iterator_category;
```
mrks: Wrote something about this above (reference issue, add that this is a hack)

src/test/lib/matches_all_pos_list_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+  std::shared_ptr<TableWrapper> dummy_table_wrapper;
+  std::shared_ptr<CreateTable> create_table;
+};
+
+TEST_F(MatchesAllPosListTest, AddAfterMatchedAllTest) {
+  // This checks if the matchesAllChunk-Flag in the PosList correctly handles rows, that are added to the table
+  // after the posList was created. These later added rows should not be contained in the PosList
+
+  auto table_name = "test_table";
+  auto table = load_table("resources/test_data/tbl/float_int.tbl", 10, false);
```
mrks: When looking just at this file, it is hard to guess what `false` means. In other places, we uses `enum class X : bool` for this. See, e.g., https://www.bfilipek.com/2017/03/on-toggle-parameters.html

src/test/lib/matches_all_pos_list_test.cpp
```diff
@@ -0,0 +1,65 @@
+#include "base_test.hpp"
+#include "gtest/gtest.h"
```
mrks: Should not be necessary

src/test/lib/matches_all_pos_list_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+  auto table_to_add_name = "test_table_to_add";
+  auto table_to_add = load_table("resources/test_data/tbl/float_int.tbl", 10);
+  // Insert Operator works with the Storage Manager, so the test table must also be known to the StorageManager
+  Hyrise::get().storage_manager.add_table(table_name, table);
+  Hyrise::get().storage_manager.add_table(table_to_add_name, table_to_add);
+
+  auto get_table = std::make_shared<GetTable>(table_name);
+  get_table->execute();
+  const auto chunkID = ChunkID{0};
+  const auto matchesAllPosList =
```
mrks: naming

src/test/lib/matches_all_pos_list_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    create_table = std::make_shared<CreateTable>("t", false, dummy_table_wrapper);
+  }
+
+  TableColumnDefinitions column_definitions;
+  std::shared_ptr<TableWrapper> dummy_table_wrapper;
+  std::shared_ptr<CreateTable> create_table;
+};
+
+TEST_F(MatchesAllPosListTest, AddAfterMatchedAllTest) {
```
mrks: Pending the discussion above, this test (and the setup above) might be replaced with an `EXPECT_THROW`. However, I am missing a test that tests the intended use of this type of pos list.

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
mrks: Might not be you: https://stackoverflow.com/questions/14132898/gcc-wuninitialized-wmaybe-uninitialized-issues

Do you have the same compiler version?

src/lib/operators/difference.cpp
```diff
@@ -81,7 +81,7 @@ std::shared_ptr<const Table> Difference::_on_execute() {
     Segments output_segments;
 
     // creating a map to share pos_lists (see table_scan.hpp)
-    std::unordered_map<std::shared_ptr<const PosList>, std::shared_ptr<PosList>> out_pos_list_map;
+    std::unordered_map<std::shared_ptr<const AbstractPosList>, std::shared_ptr<PosList>> out_pos_list_map;
```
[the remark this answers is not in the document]
He3lixxx: Yes. We could rename it. What would be your preferred name? What about `VectorPosList`?

src/lib/optimizer/strategy/predicate_reordering_rule.cpp
```diff
@@ -25,8 +25,10 @@ using namespace opossum;  // NOLINT
 bool is_predicate_style_node(const std::shared_ptr<AbstractLQPNode>& node) {
   if (node->type == LQPNodeType::Predicate) return true;
 
-  // Validate can be seen as a Predicate on the MVCC column
+    // Validate can be seen as a Predicate on the MVCC column
+#if 0
   if (node->type == LQPNodeType::Validate) return true;
+#endif
```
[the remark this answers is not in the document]
He3lixxx: There is a test. We commented it out below, there is a TODO there. The decision on whether we actually want to have this change or not has not yet been made, I'd decide after benchmarking both versions on release.

(see src/test/optimizer/strategy/predicate_reordering_rule_test.cpp)

src/lib/operators/difference.cpp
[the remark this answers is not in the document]
mrks: A PosList that stores only chunk offsets would use a vector, too, so that might conflict.

RowIDPosList?
GenericPosList?

Maybe we'd could try to find two more names before we settle. Also pinging @Bensk1.

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
He3lixxx: I have the exact same compiler version as @Bensk1. If also fails in a completely new build environment:
```bash
mkdir ~/test && cd ~/test
git clone git@github.com:Xperianer/hyrise.git
cd hyrise
mkdir build && cd build
git submodule update --init
git checkout polymorphic-poslist-compiling
git log  # is on 41d70ad
cmake .. -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DENABLE_NUMA_SUPPORT=Off
make -j 32 hyriseTest
```

same thing happens when building with ninja and the command @Bensk1 gave:
```bash
CXXFLAGS=-fdiagnostics-color=always cmake -DENABLE_NUMA_SUPPORT=Off -DCMAKE_BUILD_TYPE=Release -G Ninja ..
ninja hyriseTest
```

`/usr/bin/env` does not report any environment variables that should change gcc's behaviour.
Is there any other place where I could have globally configured gcc to behave differently?

src/lib/operators/difference.cpp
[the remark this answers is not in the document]
He3lixxx: I think `RowIDPosList` is the best candidate. All PosLists differ only in how they internally store the data, so naming them after what they store internally makes sense.

Other names that came to my mind are based on synonyms for "Generic": `DefaultPosList`, `IntuitivePosList`, `NaivePosList` - but I think they are worse.

Any other ideas @XPerianer @leogeier

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
mrks: I'd like to test something in your environment. Let me know when it's free.

He3lixxx: Go ahead :)

mrks: My guess is that it is another gcc unused bug. Reasons for that:
* clang isn't unhappy about it
* `_data` is a reference and, as such, it should be impossible to create an object where it is uninitialized. You should get "error: constructor for 'X' must explicitly initialize the reference member 'foo'" even without any warning messages enabled
* GCC's unused detection is known to be buggy

My suggestion is to add this:

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
...
#pragma GCC diagnostic pop
```

Hopefully, doing this in the get() method should be sufficient.

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -9,6 +9,7 @@
 
 #include "all_type_variant.hpp"
 #include "constant_mappings.hpp"
+#include "storage/pos_list.hpp"
```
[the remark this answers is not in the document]
He3lixxx: `_write_rows_to_matches` is given two instances, one of `AbstractPosList` and one of `PosList` (old name). Vector methods like resize and operator[] with a reference return type will be used on the latter, so it makes sense to keep the type there I think.

src/lib/storage/matches_all_pos_list.hpp
```diff
@@ -0,0 +1,60 @@
...
+    return RowID{_common_chunk_id, static_cast<ChunkOffset>(n)};
+  }
+
+  bool empty() const final { return size() == 0; }
+
+  size_t size() const final { return _common_chunk_size_on_creation; }
+
+  size_t memory_usage(const MemoryUsageCalculationMode) const final { return sizeof *this; }
+
+  PosListIterator<const MatchesAllPosList*, RowID> begin() const {
```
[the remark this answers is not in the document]
He3lixxx: In theory, a random access iterator (which we were aiming for here) should be default constructible. Currently, in our code, this never happens (default ctor is deleted) -> Currently, it could be a reference.

So: should we make it a reference?

src/lib/storage/pos_list.hpp
```diff
@@ -125,24 +124,15 @@ struct PosList final : private pmr_vector<RowID> {
   using Vector::resize;
   using Vector::swap;
 
-  friend bool operator==(const PosList& lhs, const PosList& rhs);
-  friend bool operator==(const PosList& lhs, const pmr_vector<RowID>& rhs);
-  friend bool operator==(const pmr_vector<RowID>& lhs, const PosList& rhs);
+  Vector& as_vector() { return static_cast<Vector&>(*this); }
```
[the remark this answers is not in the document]
He3lixxx: Removed, `as_vector` was an outdated concept.

src/lib/CMakeLists.txt
```diff
@@ -475,6 +476,7 @@ set(
     storage/lz4_segment/lz4_segment_iterable.hpp
     storage/lz4_segment.cpp
     storage/lz4_segment.hpp
+    storage/matches_all_pos_list.hpp
```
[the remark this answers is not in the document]
He3lixxx: done

src/lib/resolve_type.hpp
```diff
@@ -212,6 +214,33 @@ std::enable_if_t<std::is_same_v<BaseSegment, std::remove_const_t<BaseSegmentType
...
+    func(*pos_list);
+  } else if (auto matches_all_pos_list = dynamic_cast<const MatchesAllPosList*>(&untyped_pos_list)) {
+    func(*matches_all_pos_list);
+  } else {
+    Fail("Unrecognized PosList type encountered");
+  }
+}
+
+template <typename Functor>
+void resolve_pos_list_type(const std::shared_ptr<const AbstractPosList>& untyped_pos_list, const Functor& func) {
```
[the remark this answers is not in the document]
He3lixxx: Removed the reference-version of this method as it was unused

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+class PosList;
+
+class AbstractPosList {
+ public:
+  template <typename PosListType = const AbstractPosList*, typename DereferenceReturnType = RowID>
+  class PosListIterator : public boost::iterator_facade<PosListIterator<PosListType, DereferenceReturnType>, RowID,
+                                                        boost::random_access_traversal_tag, DereferenceReturnType> {
+   public:
+    PosListIterator(PosListType pl_pointer, ChunkOffset pos, ChunkOffset max_size) {
```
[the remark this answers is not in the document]
He3lixxx: 1. renamed
2. Removed the parameter. Do we want to keep a DebugAssert like `DebugAssert(_chunk_offset < _max_size, "past-the-end PosListIterator dereferenced");` in the dereference operator? We could by simply using `pl_pointer->size()` there - or, if we think calling `size()` has a performance overhead, we can store the size here, once, and then use that stored value.

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+      _pl_pointer = pl_pointer;
+      _chunk_offset = pos;
+      _max_size = max_size;
+    }
+
+    // boost will not use the random_access_iterator_tag if reference_type is not a c++ reference (which it isn't here)
+    // we still want to use random access (for binary search, distance, ...)
+    typedef std::random_access_iterator_tag iterator_category;
+
+    PosListIterator() = delete;
```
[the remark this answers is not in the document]
He3lixxx: removed.

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+    void increment() { ++_chunk_offset; }
+
+    void decrement() { --_chunk_offset; }
+
+    void advance(std::ptrdiff_t n) { _chunk_offset += n; }
+
+    bool equal(const PosListIterator& other) const {
+      DebugAssert(_pl_pointer == other._pl_pointer, "You trusted me and I failed you.");
```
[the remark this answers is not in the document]
He3lixxx: Fixed.

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+      return other._chunk_offset == _chunk_offset;
+    }
+
+    std::ptrdiff_t distance_to(const PosListIterator& other) const {
+      return static_cast<std::ptrdiff_t>(other._chunk_offset) - _chunk_offset;
+    }
+
+    DereferenceReturnType dereference() const {
+      DebugAssert(_chunk_offset < _max_size, "You foool");
```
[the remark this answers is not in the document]
He3lixxx: See annotation for line 22 - we could keep the assertion without having max_size in the interface.

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+      return (*_pl_pointer)[_chunk_offset];
+    }
+
+    PosListType _pl_pointer;
+    ChunkOffset _chunk_offset;
+    ChunkOffset _max_size;
+  };
+
+  // public:
```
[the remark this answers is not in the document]
He3lixxx: removed.

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+#include "types.hpp"
+#include "utils/assert.hpp"
+#include "utils/performance_warning.hpp"
+
+namespace opossum {
+
+class PosList;
+
+class AbstractPosList {
```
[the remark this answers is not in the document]
He3lixxx: Done as `class AbstractPosList : private Noncopyable` - is this correct in this case? (Not sure about private vs public inheritance in this case)

src/lib/storage/abstract_pos_list.hpp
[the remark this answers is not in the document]
mrks: Sounds good.

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+  // Returns whether it is guaranteed that the PosList references a single ChunkID.
+  // However, it may be false even if this is the case.
+  virtual bool references_single_chunk() const = 0;
+
+  // For chunks that share a common ChunkID, returns that ID.
+  virtual ChunkID common_chunk_id() const = 0;
+
+  virtual RowID operator[](size_t n) const = 0;
+
+  PosListIterator<> begin() const {
```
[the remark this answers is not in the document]
He3lixxx: Done for the 4 iterator methods (`begin`, `end`, `cbegin`, `cend`). The `PosListIterator` methods have to stay here because of templating and the `inline operator==` implementation has to stay here so it's available in every compilation unit, right?

Should we also do that for the implementations of the methods in matches_all_pos_list.hpp / pos_list.hpp (old name)?

Should we benchmark the performance before / after moving? I wouldn't expect these methods to be performance critical.

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+    PerformanceWarning("AbstractPosList::begin() called - dereferencing this iterator will be slow.");
```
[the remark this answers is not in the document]
He3lixxx: done

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+    return PosListIterator<>(this, ChunkOffset{0}, static_cast<ChunkOffset>(size()));
+  }
+
+  PosListIterator<> end() const {
+    PerformanceWarning("AbstractPosList::end() called - dereferencing this iterator will be slow.");
```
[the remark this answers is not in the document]
He3lixxx: makes sense

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+inline bool operator==(const AbstractPosList& lhs, const AbstractPosList& rhs) {
+  PerformanceWarning("Using slow PosList comparison.");
+
+  if (lhs.size() != rhs.size()) {
+    return false;
+  }
+
+  auto lhsIt = lhs.cbegin();
+  auto lhsEnd = lhs.cend();
+  auto rhsIt = rhs.cbegin();
```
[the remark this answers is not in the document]
He3lixxx: replaced by std::equal

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+  // Capacity
+  virtual bool empty() const = 0;
+  virtual size_t size() const = 0;
+
+  virtual size_t memory_usage(const MemoryUsageCalculationMode) const = 0;
+
+  friend bool operator==(const AbstractPosList& lhs, const AbstractPosList& rhs);
+};
+
+inline bool operator==(const AbstractPosList& lhs, const AbstractPosList& rhs) {
```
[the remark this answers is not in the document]
He3lixxx: yup.

src/lib/storage/matches_all_pos_list.hpp
```diff
@@ -0,0 +1,60 @@
+#pragma once
+
+#include "./chunk.hpp"
```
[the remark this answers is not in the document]
He3lixxx: is now `storage/chunk.hpp` as the file moved to `storage/pos_lists/matches_all_pos_list.hpp`

src/lib/storage/matches_all_pos_list.hpp
```diff
@@ -0,0 +1,60 @@
...
+class MatchesAllPosList : public AbstractPosList {
+ public:
+  explicit MatchesAllPosList(std::shared_ptr<const Chunk> common_chunk, const ChunkID common_chunk_id)
+      : _common_chunk(common_chunk),
+        _common_chunk_id(common_chunk_id),
+        _common_chunk_size_on_creation(common_chunk->size()) {}
+
+  MatchesAllPosList& operator=(MatchesAllPosList&& other) = default;
+
+  MatchesAllPosList() = delete;
```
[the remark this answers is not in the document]
He3lixxx: removed

src/lib/storage/matches_all_pos_list.hpp
```diff
@@ -0,0 +1,60 @@
...
+                                                            static_cast<ChunkOffset>(size()));
+  }
+
+  PosListIterator<const MatchesAllPosList*, RowID> cbegin() const { return begin(); }
+
+  PosListIterator<const MatchesAllPosList*, RowID> cend() const { return end(); }
+
+ private:
+  std::shared_ptr<const Chunk> _common_chunk = nullptr;
+  ChunkID _common_chunk_id = INVALID_CHUNK_ID;
```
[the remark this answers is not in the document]
He3lixxx: removed.

src/lib/storage/pos_list.hpp
He3lixxx: replaced all occurences of `override` in this file with `final`.

src/lib/storage/reference_segment/reference_segment_iterable.hpp
He3lixxx: fixed.

src/lib/storage/segment_iterables.hpp
He3lixxx: replaced with
"The two overloads of `with_iterators` are ambiguous if a `shared_ptr<PosListType>` is given.
This will call the overload expecting `shared_ptr<const PosListType>`"

Is that ok?

src/test/lib/matches_all_pos_list_test.cpp
He3lixxx: removed

src/lib/storage/matches_all_pos_list.hpp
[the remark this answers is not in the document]
mrks: Hmm, I just remembered iterators being default-constructible was one of the requirements for the LegacyRandomAccessIterator (inherited from LegacyForwardIterator). If you already deleted the default constructor, changing this to a reference should not make a difference, so yes.

src/lib/storage/abstract_pos_list.hpp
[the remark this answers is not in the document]
mrks: Any chance you forgot to commit a cpp file?

src/lib/storage/segment_iterables.hpp
mrks: Shouldn't the second version match `<PosListType = const WhateverPosList>`, too? Have you tried what happens if you only use the second version?

src/lib/storage/abstract_pos_list.hpp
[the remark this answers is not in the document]
mrks: As we don't use link-time optimization (because it was not beneficial so far and takes a lot of time), I'd expect that having things like `operator[]` in a separate class is bad for performance. Feel free to have look, but for me it would be sufficient to move (c)begin/c(end)/memory_usage in the other classes.

src/lib/operators/difference.cpp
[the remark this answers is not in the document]
mrks: Yepp, makes sense. Just in case this remains the last open item for the PR: Let's go for RowIDPosList then if there is no additional input to the discussion.

src/lib/resolve_type.hpp
```diff
@@ -212,6 +214,33 @@ std::enable_if_t<std::is_same_v<BaseSegment, std::remove_const_t<BaseSegmentType
   }
 }
 
+template <typename Functor>
+void resolve_pos_list_type(const AbstractPosList& untyped_pos_list, const Functor& func) {
```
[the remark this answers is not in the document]
leogeier: 1: A method similar to `with_iterators()` would make sense for AbstractPosList. But we still need to access the resolved PosList directly at least once, so we cannot get rid of `resolve_pos_list_type()`. And having it separate from the AbstractPosList is consistent with `resolve_segment_type()` (as is the naming).

src/lib/storage/pos_lists/abstract_pos_list.hpp
```diff
@@ -83,12 +83,7 @@ class AbstractPosList : private Noncopyable {
 
 inline bool operator==(const AbstractPosList& lhs, const AbstractPosList& rhs) {
   PerformanceWarning("Using slow PosList comparison.");
-
-  if (lhs.size() != rhs.size()) {
-    return false;
-  }
-
-  return std::equal(lhs.cbegin(), lhs.cend(), rhs.cbegin());
+  return std::equal(lhs.cbegin(), lhs.cend(), rhs.cbegin(), rhs.cend());
```
mrks: Hmm, I should have read the documentation until the end. Nice catch.

src/lib/storage/segment_iterables.hpp
He3lixxx: Yes. I think we removed the initially ambiguous version of `with_iterators` that we wrote this resolver for. Using `shared_ptr<PosListType>` instead of `shared_ptr<const PosListType>` allows compiling without it.

Comment and resolve-method removed.

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
He3lixxx: What you suggested works for me. Benchmarks on whether to do validate reordering are still todo.

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+   public:
+    PosListIterator(PosListType pl_pointer, ChunkOffset pos, ChunkOffset max_size) {
+      _pl_pointer = pl_pointer;
+      _chunk_offset = pos;
+      _max_size = max_size;
+    }
+
+    // boost will not use the random_access_iterator_tag if reference_type is not a c++ reference (which it isn't here)
+    // we still want to use random access (for binary search, distance, ...)
+    typedef std::random_access_iterator_tag iterator_category;
```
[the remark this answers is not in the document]
leogeier: Replace with

src/lib/resolve_type.hpp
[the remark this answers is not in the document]
He3lixxx: (1, 2: We don't have strong opinions about the structuring / naming here. Making it a method of AbstractPosList or naming it something else is fine for us)

3: We moved the resolving into the `with_iterators` method. We don't understand why we would need the template parameter to enable / disable resolving.

4: Compile Times (make -j32 hyriseBenchmarkTPCH on new build env without ccache using clang on vm-nordes):
Release:
  our branch:  15min (culprit: column_vs_column_table_scan_impl.cpp takes about 15mins alone)
  master: 8 min (column_vs_column_table_scan_impl takes about 5mins)
Debug:
  our branch: 4min
  master: 4min

src/lib/operators/validate.cpp
```diff
@@ -211,24 +212,30 @@ void Validate::_validate_chunks(const std::shared_ptr<const Table>& in_table, co
       DebugAssert(chunk_in->has_mvcc_data(), "Trying to use Validate on a table that has no MVCC data");
       const auto mvcc_data = chunk_in->get_scoped_mvcc_data_lock();
 
-      temp_pos_list.guarantee_single_chunk();
-
       if (_can_use_chunk_shortcut && _is_entire_chunk_visible(chunk_in, snapshot_commit_id)) {
+#if 1
```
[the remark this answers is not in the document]
leogeier: done

CMakeLists.txt
```diff
@@ -92,10 +92,10 @@ if(${CMAKE_VERSION} VERSION_LESS "3.16.0")
 endif()
 
 # On Linux, use the lld linker, which is twice as fast for incremental debug builds
-if (NOT APPLE)
-    set(CMAKE_SHARED_LINKER_FLAGS "-fuse-ld=lld")
-    set(CMAKE_EXE_LINKER_FLAGS "-fuse-ld=lld")
-endif()
+# if (NOT APPLE)
```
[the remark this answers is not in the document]
leogeier: commented in

src/lib/resolve_type.hpp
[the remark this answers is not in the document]
mrks: 1, 2: Let's leave it the way it is. We already have both ways of doing things and I am not sure if one is better than the other.

3: The template parameter would be helpful in reducing the compile time. If we can opt in to using the AbstractPosList for cases that are slow anyway, we don't need to instantiate as many templates. For example, doing something on LZ4-compressed segments is slow anyway and the cost of the virtual calls in the AbstractPosList will be hidden by the cost of the decompression. Here, setting such a template parameter would help. Have a look at this:

https://github.com/hyrise/hyrise/blob/29b47d5bee4ff3950b72048a22f3263e2197e0f5/src/lib/operators/table_scan/column_vs_column_table_scan_impl.cpp#L37-L56

Unless you really want to do this yourselves, I'll have a look at this once this PR is merged.

src/lib/storage/abstract_pos_list.hpp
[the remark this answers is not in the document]
mrks: Mmh, the original text was not bad. I'd still shortly explain that without this hack, binary_search, distance, and others will not do random but linear accesses.

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
Bouncner: I guess you have an outdated Cmake version. I've seen the problem several times with an old CMake version and a new version fixed it. But I never took a deeper look into that.

mrks: That sounds super weird.

src/lib/storage/segment_iterables/base_segment_iterators.hpp
Bouncner: Sorry, did not realize that I was tagged here. To be honest, I don't like to have this hack in the master. Is it still true that you cannot get the tests running with the tag?

Could you give me a link to the mentioned `distance` method?

Bouncner: And the problem is usually (at least for the iterables currently in the master) not that boost is not using/setting the tag. It actually does. But there is a difference here between boost and the STL tags. The STL talks about random access and has a lot of constraints there, boost talks about random access **traversal**. This traversal works and boost correctly sets the tags for the iterables, but that tag is not something the STL knows or recognizes.

src/lib/storage/pos_lists/abstract_pos_list.hpp
```diff
@@ -0,0 +1,89 @@
...
+class RowIDPosList;
+
+class AbstractPosList : private Noncopyable {
+ public:
+  template <typename PosListType = const AbstractPosList*, typename DereferenceReturnType = RowID>
+  class PosListIterator : public boost::iterator_facade<PosListIterator<PosListType, DereferenceReturnType>, RowID,
+                                                        boost::random_access_traversal_tag, DereferenceReturnType> {
+   public:
+    PosListIterator(PosListType pos_list, ChunkOffset offset) {
+      _pos_list = pos_list;
```
Bouncner: That looks very much like a pos list copy or am I wrong?
Why not use initializer list?
const parameters.

src/test/lib/matches_all_pos_list_test.cpp
XPerianer: I now added ```enum class FinalizeLastChunk : bool { Yes = true, No = false }``` to the ```load_table.hpp```, that should make it clearer

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
Bouncner: It does ... shouldn't be related. But it solved the issue on two machines for us.

src/lib/storage/matches_all_pos_list.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "abstract_pos_list.hpp"
+
+namespace opossum {
+
+class MatchesAllPosList : public AbstractPosList {
+ public:
+  explicit MatchesAllPosList(std::shared_ptr<const Chunk> common_chunk, const ChunkID common_chunk_id)
+      : _common_chunk(common_chunk),
+        _common_chunk_id(common_chunk_id),
+        _common_chunk_size_on_creation(common_chunk->size()) {}
```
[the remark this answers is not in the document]
XPerianer: We would vote for adding a check inside the constructor that ```common_chunk``` is final.
This should catch most of the errors, but we think some of them have to be handled outside the MatchesAllPoslist.

Consider following (constructed) scenario:
---------------
```
                    A                                                B

+------------------------------------------------+-----------------------------------+
                                                 |
 Scans all rows and sees that everything matches |
                                                 |
                                                 |  Adds rows and finalizes chunk
                                                 |
  Create MatchesAllPosList with finalized chunk  |
(containing too many Rows)                       |
                                                 +

```
--------------
This can only be eliminated by A checking before the scan (we think)

src/lib/resolve_type.hpp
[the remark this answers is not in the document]
XPerianer: We'd be very grateful :D

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
He3lixxx: Cmake was already the most recent version in the ubuntu repo (3.13.4). The error still occurs if we remove the pragmas.

src/lib/storage/abstract_pos_list.hpp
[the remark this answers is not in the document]
He3lixxx: `_max_size` removed and replaced with `_pos_list->size()`

src/lib/storage/abstract_pos_list.hpp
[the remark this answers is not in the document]
He3lixxx: Text is fixed.

src/lib/storage/abstract_pos_list.hpp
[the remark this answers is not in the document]
He3lixxx: Moved out all the methods that were not just forwarding the call and not operator[].

src/lib/storage/dictionary_segment/attribute_vector_iterable.hpp
```diff
@@ -88,17 +91,19 @@ class AttributeVectorIterable : public PointAccessibleSegmentIterable<AttributeV
...
   };
 
-  template <typename ZsDecompressorType>
+  template <typename ZsDecompressorType, typename _PosListIteratorType>
   class PointAccessIterator
-      : public BasePointAccessSegmentIterator<PointAccessIterator<ZsDecompressorType>, SegmentPosition<ValueID>> {
+      : public BasePointAccessSegmentIterator<PointAccessIterator<ZsDecompressorType, _PosListIteratorType>,
+                                              SegmentPosition<ValueID>, _PosListIteratorType> {
    public:
+    using PosListIteratorType = _PosListIteratorType;
```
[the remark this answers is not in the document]
He3lixxx: This was an attempt to only pass the PosListIteratorType to base class templates once. Note how the BasePointAccessSegmentIterator is given `_PosListIterator` twice, once indirectly in the first template parameter (which is the current class) and once as the last parameter. We thought that if we made the template parameter publicly available in the current class, we wouldn't have to add that last parameter, because the BasePointAccessSegmentIterator could just use `Derived::PosListIterator`.  However, it didn't work, so currently, we still pass it twice all the time.

Do you knowhow to correctly expose and use the template parameter in the super class?
Otherwise, I guess we will remove all the leading underscores and using directives.

src/lib/storage/matches_all_pos_list.hpp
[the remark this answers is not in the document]
leogeier: We cannot store a reference to the PosList because iterators need to be copy assignable. We have to use a pointer.

src/lib/storage/segment_iterables/base_segment_iterators.hpp
He3lixxx: > Is it still true that you cannot get the tests running with the tag?

The problem was that we could not get the tests running _with_ the tag.

For the `PosListIterator` class, we want all the STL method that can be implemented more efficiently when random access is available to use that (to be concrete: We needed `std::distance` in O(1)). To get to this, we added the random access tag to the PosListIterator class.

Adding it here is not really in our scope. We just thought that it would make sense to apply this everywhere where we define custom iterators that we will with high probability use `std::distance` on as this could easily cause performance penalties.

We are also fine with undoing all the changes we did here.

He3lixxx: related distance calls for the PosListIterator are (non-exhaustive):
- src/lib/storage/chunk_encoder.cpp:60
- src/lib/storage/frame_of_reference_segment/frame_of_reference_encoder.hpp:45
- src/lib/storage/segment_iterables/base_segment_iterators.hpp:81
- src/lib/storage/reference_segment/reference_segment_iterable.hpp:151,210
- src/lib/storage/dictionary_segment/dictionary_encoder.hpp:42
- src/lib/storage/lz4_segment/lz4_encoder.hpp:80

src/lib/storage/pos_lists/abstract_pos_list.hpp
leogeier: PosListType was a pointer, so the PosList is not copied. We made that more explicit. Also did the rest.

src/lib/storage/abstract_pos_list.hpp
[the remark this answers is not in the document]
leogeier: Changed to
```
// Without this hack, binary search, distance, etc. cannot use random accesses and will run in linear time instead.
// For an explanation, see https://github.com/hyrise/hyrise/issues/1531
```

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
```diff
@@ -34,19 +34,21 @@ class FrameOfReferenceSegmentIterable : public PointAccessibleSegmentIterable<Fr
...
-      auto begin = PointAccessIterator<OffsetValueDecompressorT>{&_segment.block_minima(), &_segment.null_values(),
-                                                                 std::move(decompressor), position_filter->cbegin(),
-                                                                 position_filter->cbegin()};
-
-      auto end = PointAccessIterator<OffsetValueDecompressorT>{position_filter->cbegin(), position_filter->cend()};
-
-      functor(begin, end);
+      resolve_pos_list_type(position_filter, [&functor, &decompressor, *this](auto& pos_list) {
+        using PosListIteratorType = std::decay_t<decltype(pos_list->cbegin())>;
+        auto begin = PointAccessIterator<OffsetValueDecompressorT, PosListIteratorType>{
```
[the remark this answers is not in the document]
Bouncner: Will be part of #2062 for FoR.

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
Bouncner: Yes, Ubuntu lacks behind here. See our docker file for an example. In the docker setup, we manually pull the latest cmake version.
@Bensk1, @mrks : any reason not to update vm-nordes and see if that changes something for good?

src/lib/storage/segment_iterables/base_segment_iterators.hpp
Bouncner: If it is only about `std::distance()`, we can do it manually `end - it` and add a todo to use `std::distance()` again, once #1531 is resolved.
`end - it` is the same in the end and should directly call the iterator's `difference_to` method.

However, we need to find the issue that the branch fails when the tag is set. Because I hope that we'll fix the iterators one day.

But any decision here is up to @Bensk1 .

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
mrks: Unless we add that manually build cmake version as a hard requirement (which I don't see a reason too and which will also mess with those IDE things that people seem to use) that won't change anything. I'd stick with the pragmas.

Bouncner: But could we than at least open an issue and add a todo to the pragma to remove it as soon as CMake is updated on Ubuntu?

mrks: Last time I made Ubuntu's cmake version mandatory, some people complained that this would not work with their IDEs.

src/lib/storage/dictionary_segment/attribute_vector_iterable.hpp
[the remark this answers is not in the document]
mrks: Why is passing the parameter twice an issue?

src/lib/storage/matches_all_pos_list.hpp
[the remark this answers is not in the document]
mrks: What if the constructor simply took the chunk size at the time of scanning / validating / ...? If instead of

```
  explicit MatchesAllPosList(std::shared_ptr<const Chunk> common_chunk, const ChunkID common_chunk_id)
      : _common_chunk(common_chunk),
        _common_chunk_id(common_chunk_id),
        _common_chunk_size_on_creation(common_chunk->size()) {}
```

we had

```
  explicit MatchesAllPosList(const ChunkID common_chunk_id, const ChunkOffset common_chunk_size)
      : _common_chunk_id(common_chunk_id),
        _common_chunk_size(chunk_size) {}
```

this should not be an issue. It should also be easier for users. `_common_chunk` seems unused anyway.

src/lib/storage/dictionary_segment/attribute_vector_iterable.hpp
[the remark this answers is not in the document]
He3lixxx: We considered having two parameters that will always be given the same value a code smell.

I'll remove all the underscores and `using PolListIterator = _PosListIterator` statements.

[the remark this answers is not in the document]
mrks: Sounds good. I'm working on other PRs in parallel. In case you have any questions, we can also set up an ad-hoc call. Maybe we can get this done in the next hours?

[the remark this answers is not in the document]
mrks: (Please send me a mail if you think that makes sense. Github notifications seem to be a bit delayed.)

src/lib/storage/segment_iterables/base_segment_iterators.hpp
He3lixxx: > However, we need to find the issue that the branch fails when the tag is set. Because I hope that we'll fix the iterators one day.

The segfault with the tag was fixed when we merged master about 2 weeks ago.

> If it is only about `std::distance()`, we can do it manually `end - it` and add a todo to use `std::distance()` again, once #1531 is resolved.
> `end - it` is the same in the end and should directly call the iterator's `difference_to` method.

I'd estimate that to be 12 places in the code then (for my list above, I went through about half the occurences of std::distance in the code). I'm okay with doing that, but need confirmation that we actually want to do it.

mrks: Do I understand the previous discussion correctly in that we could avoid the `typedef` hack *without losing performance* if we simply replace `std::distance` by `end - it` within the pos list? If yes, by all means, please go for it.

He3lixxx: From what I understood, the replacements wouldn't be _within_ the pos list but at the locations where iterators on pos lists are used, so its scattered around the code base (see the files I listed above).

mrks: Bad phrasing, that's what I meant.

He3lixxx: Alright, will replace by `end - it` and remove the tag.

src/lib/optimizer/strategy/predicate_reordering_rule.cpp
He3lixxx: Current benchmarking results on vm-nordes. with-reordering includes this early return, without-reordering has it commented out.

image

So I guess we will leave the reordering as it is in master? @Bensk1

mrks: My guess is that the performance issues of Validate that led to this being considered have been resolved in the meantime. As such, even if I'm not asked: yes.

src/lib/storage/segment_iterables/base_segment_iterators.hpp
He3lixxx: Done. Turns out it's even fewer places than I initially thought.
Should I add a comment saying  that these subtractions should at some point be replaced with `std::distance` again?

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
He3lixxx: So what do we do here? I'd be fine with
> But could we than at least open an issue and add a todo to the pragma to remove it as soon as CMake is updated on Ubuntu?

src/lib/storage/dictionary_segment/attribute_vector_iterable.hpp
```diff
@@ -88,17 +91,19 @@ class AttributeVectorIterable : public PointAccessibleSegmentIterable<AttributeV
     ChunkOffset _chunk_offset;
   };
 
-  template <typename ZsDecompressorType>
+  template <typename ZsDecompressorType, typename _PosListIteratorType>
```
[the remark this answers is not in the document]
He3lixxx: Underscores removed.

src/lib/storage/dictionary_segment/attribute_vector_iterable.hpp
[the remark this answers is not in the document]
He3lixxx: > I'll remove all the underscores and `using PolListIterator = _PosListIterator` statements.

done.

Thanks for being so responsive :)

src/lib/storage/segment_iterables/base_segment_iterators.hpp
mrks: I don't see any advantage in `std::distance`, so I'd say no

src/test/optimizer/strategy/predicate_reordering_rule_test.cpp
mrks: As mentioned above, this is not solely an issue with the CMake version shipped with Ubuntu, but also that within IDEs. I'd simply leave the pragma it and be done with it. It's not that it really hurts anyone if it remains in there for a while.

src/lib/storage/matches_all_pos_list.hpp
```diff
@@ -0,0 +1,60 @@
+#pragma once
+
+#include "./chunk.hpp"
+#include "abstract_pos_list.hpp"
+
+namespace opossum {
+
+class MatchesAllPosList : public AbstractPosList {
+ public:
+  explicit MatchesAllPosList(std::shared_ptr<const Chunk> common_chunk, const ChunkID common_chunk_id)
```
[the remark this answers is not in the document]
mrks: As far as I know, @Bensk1 is busy with a paper. With three thumbs on on this, just go ahead. ;)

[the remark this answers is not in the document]
He3lixxx: Renamed.

src/test/lib/matches_all_pos_list_test.cpp
He3lixxx: renamed.

src/lib/storage/pos_lists/entire_chunk_pos_list.hpp
```diff
@@ -7,10 +7,9 @@ namespace opossum {
 
 class EntireChunkPosList : public AbstractPosList {
  public:
-  explicit EntireChunkPosList(std::shared_ptr<const Chunk> common_chunk, const ChunkID common_chunk_id)
-      : _common_chunk(common_chunk),
+  explicit EntireChunkPosList(const ChunkID common_chunk_id, const ChunkOffset common_chunk_size) :
         _common_chunk_id(common_chunk_id),
-        _common_chunk_size_on_creation(common_chunk->size()) {}
+        _common_chunk_size(common_chunk_size) {}
```
mrks: Could you add documentation that explains that `common_chunk_size` has to be determined before iterating over the segment? In Validate (which seems to be the only user right now), this is not important because the chunk has to be finalized - still, it should be documented there as well.

src/lib/storage/matches_all_pos_list.hpp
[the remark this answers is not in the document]
He3lixxx: I implemented what you suggested, but I'm not sure about the thread safety (I think either master is broken already or our code that stores the size when the ctor is called should be safe)

In the current master, the algorithm for generating the pos list when taking all shortcuts is:
```c++
if (_can_use_chunk_shortcut && _is_entire_chunk_visible(chunk_in, snapshot_commit_id)) {
  // boom
  const auto chunk_size = chunk_in->size();
  for (auto chunk_offset = 0u; chunk_offset < chunk_size; ++chunk_offset) {
    temp_pos_list[chunk_offset] = RowID{chunk_id, chunk_offset};
  }
}
```
and `temp_pos_list` is then moved to the result PosList. However, where I added `// boom`, we are already in a race condition, or is there some kind of transparent locking going on that I'm missing? If, between checking whether the whole chunk matches, and then getting the size, a new value is added that does not match, then we will include it erroneously, right?

src/lib/storage/pos_lists/entire_chunk_pos_list.hpp
He3lixxx: Ah ok that explains what I asked in https://github.com/hyrise/hyrise/pull/2025#discussion_r399819697

Will do.

src/lib/storage/matches_all_pos_list.hpp
```diff
@@ -0,0 +1,60 @@
...
+  MatchesAllPosList() = delete;
+
+  bool references_single_chunk() const final { return true; }
+
+  ChunkID common_chunk_id() const final {
+    DebugAssert(_common_chunk_id != INVALID_CHUNK_ID, "common_chunk_id called on invalid chunk id");
+    return _common_chunk_id;
+  }
+
+  RowID operator[](size_t n) const final {
```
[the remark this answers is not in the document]
He3lixxx: Renamed to `index`, declared `const`,

src/lib/storage/pos_lists/entire_chunk_pos_list.hpp
He3lixxx: Added:

> The EntireChunkPosList will match the first common_chunk_size elements in the chunk.
If you create an EntireChunkPosList for a mutable chunk, make sure to determine that size
_before_ doing the checks that all elements should match. This way, you prevent the race
condition of an element being added in between doing the checks and getting the size.

mrks: top

src/lib/storage/pos_lists/rowid_pos_list.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+#include "abstract_pos_list.hpp"
+#include "types.hpp"
+#include "utils/assert.hpp"
+
+namespace opossum {
+
+// For a long time, PosList was just a pmr_vector<RowID>. With this class, we want to add functionality to that vector,
+// more specifically, flags that give us some guarantees about its contents. If we know, e.g., that all entries point
+// into the same chunk, we can simplify things in split_pos_list_by_chunk_id.
```
mrks: update comment

src/lib/storage/reference_segment/reference_segment_iterable.hpp
```diff
@@ -29,88 +29,93 @@ class ReferenceSegmentIterable : public SegmentIterable<ReferenceSegmentIterable
     const auto referenced_table = _segment.referenced_table();
     const auto referenced_column_id = _segment.referenced_column_id();
 
-    const auto& pos_list = _segment.pos_list();
+    const auto& position_filter = _segment.pos_list();
 
-    const auto begin_it = pos_list->begin();
-    const auto end_it = pos_list->end();
+    resolve_pos_list_type(position_filter, [&functor, &referenced_table, &referenced_column_id](auto pos_list) {
```
mrks: needed?

src/lib/utils/load_table.cpp
```diff
@@ -80,7 +81,7 @@ std::shared_ptr<Table> load_table(const std::string& file_name, size_t chunk_siz
   }
 
   // All other chunks have been finalized by Table::append() when they reached their capacity.
-  if (!table->empty()) {
+  if (!table->empty() && static_cast<bool>(finalize_last_chunk)) {
```
mrks: For consistency (not necessarily nicer): `finalize_last_chunk == FinalizeLastChunk::Yes`

He3lixxx:
> Is this ready to go or are you aware of anything missing?

I think it's ready to go.

> How much does the ChunkOffsetPosList add?

About +200 loc.

> Can we get this in soon, too? If yes, I'll wait with the compile time work.

I think we could get it merged a lot faster than this one (maybe 1 to 3 days?). Depends on how much time @leogeier and @XPerianer have though.

mrks:
Here are some numbers from nemea:

```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.2757476568222046 | 77    | 1.2424206733703613 | 75    | -3%        |  0.0000 |
+| TPC-H 02       | 79.53157043457031  | 4772  | 84.00621795654297  | 5041  | +6%        |  0.0000 |
 | TPC-H 03       | 8.908345222473145  | 535   | 9.212895393371582  | 553   | +3%        |  0.0000 |
+| TPC-H 04       | 7.828646183013916  | 470   | 8.77935791015625   | 527   | +12%       |  0.0000 |
+| TPC-H 05       | 3.8665037155151367 | 232   | 4.252862930297852  | 256   | +10%       |  0.0000 |
 | TPC-H 06       | 264.18682861328125 | 15852 | 276.4041748046875  | 16585 | +5%        |  0.0000 |
+| TPC-H 07       | 9.634734153747559  | 579   | 10.154862403869629 | 610   | +5%        |  0.0000 |
+| TPC-H 08       | 7.403852462768555  | 445   | 7.904600620269775  | 475   | +7%        |  0.0000 |
 | TPC-H 09       | 1.7570385932922363 | 106   | 1.775230884552002  | 107   | +1%        |  0.0012 |
 | TPC-H 10       | 3.362816333770752  | 202   | 3.2888612747192383 | 198   | -2%        |  0.0310 |
+| TPC-H 11       | 38.01150131225586  | 2281  | 41.566429138183594 | 2495  | +9%        |  0.0000 |
+| TPC-H 12       | 14.9895658493042   | 900   | 17.449411392211914 | 1047  | +16%       |  0.0000 |
+| TPC-H 13       | 2.2825334072113037 | 137   | 2.589919328689575  | 156   | +13%       |  0.0000 |
 | TPC-H 14       | 44.687896728515625 | 2682  | 43.932186126708984 | 2636  | -2%        |  0.0000 |
+| TPC-H 15       | 118.19493103027344 | 7092  | 124.5029067993164  | 7471  | +5%        |  0.0000 |
 | TPC-H 16       | 9.052909851074219  | 544   | 9.007349967956543  | 541   | -1%        |  0.0853 |
+| TPC-H 17       | 5.363152027130127  | 322   | 5.811931133270264  | 349   | +8%        |  0.0000 |
 | TPC-H 18       | 1.2516614198684692 | 76    | 1.310404896736145  | 79    | +5%        |  0.0000 |
+| TPC-H 19       | 9.902627944946289  | 595   | 10.675104141235352 | 641   | +8%        |  0.0000 |
+| TPC-H 20       | 31.53620147705078  | 1893  | 33.400760650634766 | 2005  | +6%        |  0.0000 |
 | TPC-H 21       | 1.3884423971176147 | 84    | 1.4243228435516357 | 86    | +3%        |  0.0000 |
 | TPC-H 22       | 18.706771850585938 | 1123  | 19.187511444091797 | 1152  | +3%        |  0.0000 |
+| geometric mean |                    |       |                    |       | +5%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

First of all: Good job! When we sat down in the villa and discussed different options, I was not certain that we would end up with something that does not reduce the baseline performance.

The performance gains surprise me a bit. I expected to see some gains proportional to the cost of Validate, but for many queries, the gains are bigger than expected. Looks like you cashed in some favors with the gods of compiler optimizations and code alignment. ;)

PR #2026, opened 2020-02-17
[DYOD] Smart Position List: IndexScan PositionList

XPerianer:
This positionlist can only reference a single chunk.
Thats' why we can optimize the storage, not saving whole RowIDs, but only chunk_offsets.

Depends on #2025, cause that adds the AbstractPosList needed for this Subclass.


Bensk1:
Could you explicitly contact the reviewing group? That should be the set operations group: Lukas Ehrig, Teresa Lasarow, Tobias Nack.

Also, could you merge the master?

Bensk1:
> Could you explicitly contact the reviewing group? That should be the set operations group: Lukas Ehrig, Teresa Lasarow, Tobias Nack.
> 
> Also, could you merge the master?

My bad, you should be reviewed by the cache group: Tobias Zagorni, Carla Terboven, Julius Streit, Leonard Pabst.

XPerianer:
> Also, could you merge the master?
Just an update to that, as this depends on #2025, we will first merge master in there and pull that into here afterwards, but currently are dealing with test failures on #2025.

Bensk1:
Alright, that's fine. Please have a look at bensk1/indexscan-poslist, I adjusted the SingleChunkPosList so that it reflects my initial intention to speed up index scans. With the new implementation we get 20 to 30% performance boost on TPC-H 6 and I assume we can get quite a lot for some TPC-H joins.

CMakeLists.txt
```diff
@@ -92,10 +92,10 @@ if(${CMAKE_VERSION} VERSION_LESS "3.16.0")
...
 
 # On Linux, use the lld linker, which is twice as fast for incremental debug builds
-if (NOT APPLE)
-    set(CMAKE_SHARED_LINKER_FLAGS "-fuse-ld=lld")
-    set(CMAKE_EXE_LINKER_FLAGS "-fuse-ld=lld")
-endif()
+# if (NOT APPLE)
+#     set(CMAKE_SHARED_LINKER_FLAGS "-fuse-ld=lld")
+#     set(CMAKE_EXE_LINKER_FLAGS "-fuse-ld=lld")
+# endif()
```
carlaterboven: ?
mkrs already recommended a lot of changes in PR 2025. We will try to avoid duplication.

src/lib/optimizer/strategy/index_scan_rule.cpp
```diff
@@ -71,6 +71,8 @@ bool IndexScanRule::_is_index_scan_applicable(const IndexStatistics& index_stati
 
   if (index_statistics.column_ids[0] != operator_predicate.column_id) return false;
 
+  return true;
```
carlaterboven: What about the rest of the code in line 76-83? Am I wrong or do you always return here?

src/lib/operators/validate.cpp
```diff
@@ -211,24 +212,30 @@ void Validate::_validate_chunks(const std::shared_ptr<const Table>& in_table, co
...
           temp_pos_list[chunk_offset] = RowID{chunk_id, chunk_offset};
         }
+        pos_list_out = std::make_shared<const PosList>(std::move(temp_pos_list));
+#endif
       } else {
+        PosList temp_pos_list;
+        temp_pos_list.guarantee_single_chunk();
         // Generate pos_list_out.
         auto chunk_size = chunk_in->size();  // The compiler fails to optimize this in the for clause :(
         for (auto i = 0u; i < chunk_size; i++) {
```
tzagorni: naming

src/lib/optimizer/strategy/index_scan_rule.cpp
XPerianer: We temporarily always use the indexScan when applicable, and not only when the selectivity rating is good enough, to test the performance. I added a TODO

src/benchmark/tpch_data_micro_benchmark.cpp
```diff
@@ -27,7 +29,7 @@ class TPCHDataMicroBenchmarkFixture : public MicroBenchmarkBasicFixture {
  public:
   void SetUp(::benchmark::State& state) {
     auto& sm = Hyrise::get().storage_manager;
-    const auto scale_factor = 0.01f;
+    const auto scale_factor = 5.f;
```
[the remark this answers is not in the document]
XPerianer: Removed the changed, the magic number is bad, but I think not in the scope of this PR

src/test/CMakeLists.txt
```diff
@@ -37,6 +37,8 @@ set(
     lib/import_export/csv/csv_writer_test.cpp
     lib/fixed_string_test.cpp
     lib/null_value_test.cpp
+    lib/matches_all_pos_list_test.cpp
+    lib/single_chunk_pos_list_test.cpp
```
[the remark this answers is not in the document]
XPerianer: Compilation should not fail, also the both files should exists?

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+inline bool operator==(const AbstractPosList& lhs, const AbstractPosList& rhs) {
+  PerformanceWarning("Using slow PosList comparison.");
+
+  if (lhs.size() != rhs.size()) {
+    return false;
+  }
+
+  auto lhsIt = lhs.cbegin();
+  auto lhsEnd = lhs.cend();
+  auto rhsIt = rhs.cbegin();
```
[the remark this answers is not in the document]
He3lixxx: lhs is for left hand side, so I guess lhsIt was supposed to mean left hand side iterator.
This code was replaced with a call to `std::equal` in the meantime however, so this is resolved I guess?

src/lib/storage/dictionary_segment/attribute_vector_iterable.hpp
```diff
@@ -27,16 +27,20 @@ class AttributeVectorIterable : public PointAccessibleSegmentIterable<AttributeV
...
-                                                           position_filter->cbegin(), position_filter->cbegin()};
-      auto end = PointAccessIterator<ZsDecompressorType>{_null_value_id, nullptr, position_filter->cbegin(),
-                                                         position_filter->cend()};
-      functor(begin, end);
+      resolve_pos_list_type(position_filter, [&, *this](auto& pos_list) {
+        using PosListIteratorType = std::decay_t<decltype(pos_list->cbegin())>;
+
+        auto begin = PointAccessIterator<ZsDecompressorType, PosListIteratorType>{
+            _null_value_id, std::move(decompressor), pos_list->cbegin(), pos_list->cbegin()};
+        auto end = PointAccessIterator<ZsDecompressorType, PosListIteratorType>{_null_value_id, nullptr,
```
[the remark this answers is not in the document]
He3lixxx: ran formatter.

src/lib/storage/dictionary_segment/attribute_vector_iterable.hpp
```diff
@@ -89,16 +93,20 @@ class AttributeVectorIterable : public PointAccessibleSegmentIterable<AttributeV
...
     PointAccessIterator(const ValueID null_value_id, const std::shared_ptr<ZsDecompressorType>& attribute_decompressor,
-                        const PosList::const_iterator position_filter_begin, PosList::const_iterator position_filter_it)
-        : BasePointAccessSegmentIterator<PointAccessIterator<ZsDecompressorType>,
-                                         SegmentPosition<ValueID>>{std::move(position_filter_begin),
-                                                                   std::move(position_filter_it)},
+                        const _PosListIteratorType position_filter_begin, _PosListIteratorType position_filter_it)
+        : BasePointAccessSegmentIterator<PointAccessIterator<ZsDecompressorType, _PosListIteratorType>,
+                                         SegmentPosition<ValueID>, _PosListIteratorType>{std::move(
+                                                                                             position_filter_begin),
+                                                                                         std::move(position_filter_it)},
```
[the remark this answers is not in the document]
He3lixxx: applied auto formatting

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+inline bool operator==(const AbstractPosList& lhs, const AbstractPosList& rhs) {
+  PerformanceWarning("Using slow PosList comparison.");
+
+  if (lhs.size() != rhs.size()) {
+    return false;
+  }
+
+  auto lhsIt = lhs.cbegin();
+  auto lhsEnd = lhs.cend();
```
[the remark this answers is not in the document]
He3lixxx: Code was replaced so this should be resolved(?)

mrks:
Could you please merge the master so that we know where we are?

src/lib/storage/abstract_pos_list.hpp
```diff
@@ -0,0 +1,119 @@
...
+  virtual bool references_single_chunk() const = 0;
+
+  // For chunks that share a common ChunkID, returns that ID.
+  virtual ChunkID common_chunk_id() const = 0;
+
+  virtual RowID operator[](size_t n) const = 0;
+
+  PosListIterator<> begin() const {
+    PerformanceWarning("AbstractPosList::begin() called - dereferencing this iterator will be slow.");
+    return PosListIterator<>(this, ChunkOffset{0}, static_cast<ChunkOffset>(size()));
```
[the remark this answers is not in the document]
He3lixxx: In the code base, so far, `ChunkOffset{}` is used with numeric literals and `static_cast` is used with variables.
Anyway, we don't have this anymore, so its resolved(?)

src/lib/storage/matches_all_pos_list.hpp
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  bool empty() const override final { return size() == 0; }
+
+  size_t size() const override final { return _common_chunk->size(); }
+
+  size_t memory_usage(const MemoryUsageCalculationMode) const final { return sizeof *this; }
+
+  PosListIterator<const MatchesAllPosList*, RowID> begin() const {
+    return PosListIterator<const MatchesAllPosList*, RowID>(this, ChunkOffset{0}, static_cast<ChunkOffset>(size()));
```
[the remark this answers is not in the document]
He3lixxx: should be resolved

src/lib/storage/matches_all_pos_list.hpp
```diff
@@ -0,0 +1,53 @@
...
+  MatchesAllPosList() = delete;
+
+  bool references_single_chunk() const final { return true; }
+
+  ChunkID common_chunk_id() const final {
+    DebugAssert(_common_chunk_id != INVALID_CHUNK_ID, "common_chunk_id called on invalid chunk id");
+    return _common_chunk_id;
+  }
+
+  RowID operator[](size_t n) const final {
```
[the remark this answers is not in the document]
He3lixxx: renamed in the other PR and merged here.

src/lib/storage/pos_list.hpp
```diff
@@ -84,8 +85,10 @@ struct PosList final : private pmr_vector<RowID> {
   using Vector::get_allocator;
 
   // Element access
-  // using Vector::at; - Oh no. People have misused this in the past.
-  using Vector::operator[];
+  RowID operator[](size_t n) const final { return Vector::operator[](n); }
+
+  RowID& operator[](size_t n) { return Vector::operator[](n); }
```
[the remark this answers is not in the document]
He3lixxx: changed.

src/lib/storage/single_chunk_pos_list.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+  virtual bool empty() const override final {
+    return _offsets.empty();
+  }
+  virtual size_t size() const override final {
+    return _offsets.size();
+  }
+  
+  virtual size_t memory_usage(const MemoryUsageCalculationMode) const override final {
+    return sizeof(this) + size() * sizeof(ChunkOffset);
```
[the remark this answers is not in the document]
He3lixxx: Uh, nice catch, thanks!

src/lib/operators/validate.cpp
He3lixxx: Replaced with `chunk_offset`

src/lib/operators/index_scan.cpp
```diff
@@ -111,14 +111,13 @@ void IndexScan::_validate_input() {
...
-RowIDPosList IndexScan::_scan_chunk(const ChunkID chunk_id) {
-  const auto to_row_id = [chunk_id](ChunkOffset chunk_offset) { return RowID{chunk_id, chunk_offset}; };
-
+std::shared_ptr<SingleChunkPosList> IndexScan::_scan_chunk(const ChunkID chunk_id) {
   auto range_begin = AbstractIndex::Iterator{};
   auto range_end = AbstractIndex::Iterator{};
 
   const auto chunk = _in_table->get_chunk(chunk_id);
-  auto matches_out = RowIDPosList{};
+  // Create SingleChunkPosList
```
He3lixxx: We should remove this.

src/benchmark/tpch_data_micro_benchmark.cpp
```diff
@@ -123,6 +125,79 @@ class TPCHDataMicroBenchmarkFixture : public MicroBenchmarkBasicFixture {
...
 };
 
+BENCHMARK_F(TPCHDataMicroBenchmarkFixture, BM_TPCHQ6IndexScan)(benchmark::State& state) {
+  auto& sm = Hyrise::get().storage_manager;
+  auto lineitem_table = sm.get_table("lineitem");
+
+  // Generate an index to benchmark the indexScan
+  std::vector<ColumnID> _indexColumnIDs = {ColumnID{10}};
+  lineitem_table->create_index<GroupKeyIndex>(_indexColumnIDs);
+  // Roughly after _tpchq6_shipdate_less_predicate
```
He3lixxx: ?

src/benchmark/tpch_data_micro_benchmark.cpp
```diff
@@ -123,6 +125,79 @@ class TPCHDataMicroBenchmarkFixture : public MicroBenchmarkBasicFixture {
   LQPColumnReference _lineitem_orderkey, _lineitem_commitdate, _lineitem_receiptdate;
 };
 
+BENCHMARK_F(TPCHDataMicroBenchmarkFixture, BM_TPCHQ6IndexScan)(benchmark::State& state) {
+  auto& sm = Hyrise::get().storage_manager;
+  auto lineitem_table = sm.get_table("lineitem");
+
+  // Generate an index to benchmark the indexScan
```
He3lixxx: Guess we can remove this too as it only summarized the next two lines. @XPerianer

(Again later down)

src/benchmark/tpch_data_micro_benchmark.cpp
```diff
@@ -123,6 +125,79 @@ class TPCHDataMicroBenchmarkFixture : public MicroBenchmarkBasicFixture {
...
+  std::vector<ColumnID> _indexColumnIDs = {ColumnID{10}};
```
He3lixxx: naming style (happens more often in this file)

src/lib/operators/index_scan.cpp
```diff
@@ -134,9 +133,14 @@ RowIDPosList IndexScan::_scan_chunk(const ChunkID chunk_id) {
       range_begin = index->cbegin();
       range_end = index->lower_bound(_right_values);
 
-      matches_out.reserve(std::distance(range_begin, range_end));
-      std::transform(range_begin, range_end, std::back_inserter(matches_out), to_row_id);
+      const auto matches_size = std::distance(range_begin, range_end);
+      auto& chunk_offsets = singleChunkPosList->get_offsets();
+      chunk_offsets.resize(matches_size);
 
+      for (auto matches_position = 0; matches_position < matches_size; ++matches_position) {
```
He3lixxx: `++matches_position, ++range_begin` in the loop header?

(Once more below)

src/lib/operators/index_scan.cpp
```diff
@@ -187,18 +191,19 @@ RowIDPosList IndexScan::_scan_chunk(const ChunkID chunk_id) {
...
 
-  for (auto matches_position = current_matches_size; matches_position < final_matches_size; ++matches_position) {
-    matches_out[matches_position] = RowID{chunk_id, *range_begin};
+  for (auto matches_position = previous_matches_size; matches_position < matches_size; ++matches_position) {
+    chunk_offsets[matches_position] = *range_begin;
     range_begin++;
   }
 
-  return matches_out;
+  // TODO: Maybe it is more easy to make the singleChunkPosList here, and previously only edit a chunk_offset vector
```
He3lixxx: +1, that's how I would have expected it to be designed anyway. Is there a reason not to do this? @XPerianer

src/lib/storage/pos_lists/single_chunk_pos_list.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include "abstract_pos_list.hpp"
+
+namespace opossum {
+
+class SingleChunkPosList final : public AbstractPosList {
+ public:
+  SingleChunkPosList() = delete;
```
He3lixxx: necessary?

src/lib/storage/pos_lists/single_chunk_pos_list.hpp
```diff
@@ -0,0 +1,33 @@
...
+  explicit SingleChunkPosList(ChunkID chunkID) : _chunk_id(chunkID) {}
+
+  bool empty() const final;
+  size_t size() const final;
+
+  size_t memory_usage(const MemoryUsageCalculationMode) const final;
+  bool references_single_chunk() const final;
+  ChunkID common_chunk_id() const final;
+
+  RowID operator[](size_t n) const final { return RowID{_chunk_id, _offsets[n]}; }
```
He3lixxx: I guess if we use `const size_t index` here once, we want it everywhere.

src/test/optimizer/strategy/index_scan_rule_test.cpp
```diff
@@ -108,7 +108,8 @@ TEST_F(IndexScanRuleTest, NoIndexScanWithTwoColumnPredicate) {
   EXPECT_EQ(predicate_node_0->scan_type, ScanType::TableScan);
 }
 
-TEST_F(IndexScanRuleTest, NoIndexScanWithHighSelectivity) {
+// TODO: This is obsolete for a fast IndexScan
```
He3lixxx: todo

src/lib/operators/index_scan.cpp
```diff
@@ -187,18 +191,19 @@ RowIDPosList IndexScan::_scan_chunk(const ChunkID chunk_id) {
...
-  matches_out.resize(final_matches_size);
+  auto& chunk_offsets = singleChunkPosList->get_offsets();
+  const auto previous_matches_size = chunk_offsets.size();
+  const auto matches_size = previous_matches_size + static_cast<size_t>(std::distance(range_begin, range_end));
+  chunk_offsets.resize(matches_size);
 
-  for (auto matches_position = current_matches_size; matches_position < final_matches_size; ++matches_position) {
-    matches_out[matches_position] = RowID{chunk_id, *range_begin};
+  for (auto matches_position = previous_matches_size; matches_position < matches_size; ++matches_position) {
+    chunk_offsets[matches_position] = *range_begin;
```
Bensk1: That way, we will not see much of the performance gains. What about the idea of pointing into index' offsets as we discussed? This would be a less general type of PosList but a much faster one.

mrks: Worth benchmarking before we dive into the discussion, but thinking about making the PosList dependent on `AbstractIndex::Iterator` gives me the shivers. :/

src/lib/operators/table_scan/abstract_dereferenced_column_table_scan_impl.cpp
```diff
@@ -19,11 +19,11 @@ AbstractDereferencedColumnTableScanImpl::AbstractDereferencedColumnTableScanImpl
     const PredicateCondition init_predicate_condition)
     : predicate_condition(init_predicate_condition), _in_table(in_table), _column_id(column_id) {}
 
-std::shared_ptr<PosList> AbstractDereferencedColumnTableScanImpl::scan_chunk(const ChunkID chunk_id) const {
+std::shared_ptr<RowIDPosList> AbstractDereferencedColumnTableScanImpl::scan_chunk(const ChunkID chunk_id) const {
   const auto chunk = _in_table->get_chunk(chunk_id);
   const auto& segment = chunk->get_segment(_column_id);
 
-  auto matches = std::make_shared<PosList>();
+  auto matches = std::make_shared<RowIDPosList>();
```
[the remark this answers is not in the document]
He3lixxx: I like the idea. All _scan_non_reference_segments seem to end up in `_scan_with_iterators` or `SegmentSearch.scan_sorted_segment`. It's a bit tricky though as for some paths, I think all the current interfaces have to continue existing (allow passing a pre-filled RowIDPosList to the _scan_ methods). We could use templating to get this without much duplication, but we would have to move a lot of methods to the .hpp files then, probably increasing compilation and especially recompilation times.

This is also applicable in some operators. For example in the validate operator, there is a path where we know that all `RowID`s have the same `chunk_id` but we're creating a `RowIDPosList`.

src/lib/operators/index_scan.cpp
XPerianer: I just did a test with the current version singleChunkPosList (that is only saving the Row-offset, but does not point into the index) with an index on l_shipdate for the TPCH-6 query. If you look at the [query plan ](https://hyrise-ci.epic-hpi.de/job/hyrise/job/hyrise/job/master/lastStableBuild/artifact/query_plans/tpch/TPC-H_06-LQP.svg), ```l_discount``` might have more effect, but I could not get the indexScan to work with this one. The effects here are really small.

@Bensk1 mentioned an improvement with the PosList pointing into the index of 20%~30% [previously](https://github.com/hyrise/hyrise/pull/2026#issuecomment-595803454), so that might be the better option.
```
+----------------+--------------------+------+--------------------+------+------------+---------+                                                                             │········
| TPC-H 06       | 157.88804626464844 | 9474 | 158.05584716796875 | 9484 | +0%        |  0.6805 |                                                                             │········
| geometric mean |                    |      |                    |      | +0%        |         |                                                                             │········
```

src/lib/operators/table_scan/abstract_dereferenced_column_table_scan_impl.cpp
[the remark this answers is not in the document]
He3lixxx: (See https://github.com/hyrise/hyrise/pull/2026#discussion_r400113202 - @XPerianer had almost zero perfomance impact with the SingleChunkPosList, which I don't really understand)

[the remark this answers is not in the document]
mrks: Could you please commit this (and revert it) so that I can have a look at the diff?

src/lib/operators/table_scan/abstract_table_scan_impl.hpp
```diff
@@ -6,7 +6,7 @@
 
 #include <array>
 
-#include "storage/pos_list.hpp"
+#include "storage/pos_lists/rowid_pos_list.hpp"
```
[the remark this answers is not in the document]
He3lixxx: done

src/lib/operators/index_scan.cpp
Bensk1: > Worth benchmarking before we dive into the discussion, but thinking about making the PosList dependent on `AbstractIndex::Iterator` gives me the shivers. :/

Hm, I am pretty sure that we have discussed that option before. Maybe we did not specifically mention that the PosList has to depend on the AbstractIndex::Iterator though, but certainly that the PosList should somehow "point into the index."

src/lib/operators/table_scan/abstract_dereferenced_column_table_scan_impl.cpp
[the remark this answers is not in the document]
He3lixxx: Perhaps I should have been clearer: he wasn't benchmarking what we are discussing here, he was just testing the performance impact on the index scan, so he was on 08220ad2a

src/lib/operators/index_scan.cpp
mrks: I just had the first look at the AbstractIndex interface since a long time. It seems as if all current index implementations actually hold such an `std::vector<ChunkOffset>`. This is only possible because none of the implementations allows for inserts. For now, this means that we could either use `std::vector<ChunkOffset>` or `std::span<ChunkOffset>` if we want to avoid the copy.

For the long term, when considering modifiable indexes, `AbstractIndex::Iterator` seems somewhat reasonable. Note that this requires a virtual method call for both incrementing and dereferencing. We should consider if resolving the PosList type makes sense if we have virtual method calls anyway.

src/lib/operators/table_scan/abstract_dereferenced_column_table_scan_impl.cpp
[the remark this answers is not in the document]
XPerianer: Actually, on 7aaeb99, with the modification that the `l_shipdate` index is created. I am also pretty sure that the indexScan was executed.

Using a higher scale factor (3) was actually slowing down the version with index scan more, which is why I think we have another bottleneck here that is killing the performance gain.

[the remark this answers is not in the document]
Bensk1: I remember that we took some measurements on where the IndexScan spends its time. I believe it was something like 99% in the loop that writes the PosList. If we only change the loop to add ChunkOffsets instead of RowIDs and do not fundamentally change that behavior, I would not expect any performance improvements.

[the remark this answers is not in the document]
He3lixxx: > I would not expect any performance improvements.

Why? If we reduce the amount of data that has to be written by factor two, I'd also expect this memory-write-loop to speed up by factor two since I would expect this to be bottlenecked by memory write speed.
(I just verified that on vm-nordes, the RowIDPosList uses 8 bytes per element and the SingleChunkPosList uses 4 bytes per element, so there's no padding going on that could slow us down)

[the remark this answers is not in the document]
Bensk1: But during the course of your project we have never seen any performance improvements close to that, have we? In addition, it's not only the writing that consumes time.

In the past we have even seen operations on 32bit data types to be slower than on 64bit data types. Reasons could be manifold, e.g., compiler or hardware optimizations. A small experiment on my machine does not show any significant differences between `uint32_t` and `uint64_t` vectors. By verified you mean the width that is used per element or the speed you can achieve?

I don't even know if a 64bit system would not actually handle a write to a 32bit (or smaller) value as a 64bit operation. Probably, @mrks knows much more about all of that.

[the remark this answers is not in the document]
mrks: No, the system would still have to write 32 bits (otherwise, you would overwrite other data). However, for steps that are not memory-bound (and single-threaded operations often are not entirely memory-bound), a lot of effects come into play. There is a good chance that the benchmarks tell a different story when you use multiple cores and increase the amount of data. In any case, I have given up making any claims about these things. I'd form a hypothesis, run benchmarks, and try to understand the results.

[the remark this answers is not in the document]
He3lixxx: > A small experiment on my machine does not show any significant differences between `uint32_t` and `uint64_t` vectors. By verified you mean the width that is used per element or the speed you can achieve?

@Bensk1 Can you put what you tested in a gist / pastebin / whatever?

I tried this: https://gist.github.com/He3lixxx/bb21cb1f21b48d215e9ecdc123088568
and it gave me exactly what I would expect:
```
g++ -O3 on vm-nordes:
RowIDPosList took 5031ms.
SingleChunkPosList took 1924ms.

clang++ -O3 on vm-nordes:
RowIDPosList took 8085ms. // even 4x slower!
SingleChunkPosList took 1964ms.
```

(For me, just filling a vector with some uint32_t is about twice as fast as filling it with some uint64_t - see http://quick-bench.com/nUHLbcdLoGHVOg6wPZPHqAObJws. Is it possible that some of your code was simply optimized away?)

The "verified" was about the SingleChunkPosList using 4 bytes per element in the offsets vector (no padding is added - if this was padded to 8 bytes, it would make sense to me that we don't see better performance)

With the tests I've done so far, I'd expect that if filling the RowIDPosList really is the slow part, making it a SingleChunkPosList should double the performance of this part of the code.
On the index scan, if replacing it does not yield performance improvements, I'd say either traversing the index is slow or we get a slow down when we later use that PosList.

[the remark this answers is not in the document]
Bensk1: I also just filled a vector and wanted to use quick-bench for that which seems to have some kind of server problem. As the test loop took a couple of hundred milliseconds, I doubt that it was optimized away.

However, your experiment looks promising. I am just wondering that all the measurements (for example, [that one](https://github.com/hyrise/hyrise/pull/2026#discussion_r400113202)) we discussed in the past did not show any performance improvements, am I wrong?

We have measured traversing the index and since this is a simple binary search on ~65000 values, it was never significant. I believe that we also considered a second TableScan after the IndexScan to see if the use of the SingleChunkPosList is slowing everything down. I don't remember how large the impact was.

Good job! I am curious to see the impact on the TPCHMicroBenchmark.

src/lib/operators/index_scan.cpp
XPerianer: To go forward, I would think it is good to follow both approaches: 
1) Make an "Indexscanposlist", that points into the index
2) Make the poslist that only takes chunk_offsets. Here we would start by benchmarking some of the ideas that Markus pointed out [here](https://github.com/hyrise/hyrise/pull/2026#discussion_r399831741)

It might be a good idea to make another merge request for 2), to keep things separated.
We have already tested that 1) can bring a performance boost for TCP-H 6, so I think it is valid to also have an "Indexscanposlist", beside the discussion if a PosList  with offsets (2) is needed or not.

mrks: Please take into account that each additional PosList increases the compile time. We should only add those that bring a measurable benefit.

Bensk1: I just had a discussion with Markus on that. We believe that a second merge request is out of scope for the seminar, it would be too much work. With the new (and amazing) PosList system it should be easy to have a **quick** and dirty benchmark for the two approaches being actually used in an realistic IndexScan, shouldn't it? Thereby, we could verify the results @He3lixxx has seen yesterday and base our decision on which option to implement on that benchmark.

XPerianer: I am currently working on benchmarking TPC-H for the first version, to have some data there.

Bensk1: Great. I think the TPC-H micro benchmark (that uses the resulting PosList for a second scan) is also sufficient.

XPerianer: [Microbenchmarks](https://gist.github.com/XPerianer/47faa74e7933566eb73148d1468f43d3)
In the microBenchmarks, while I get a heavy improvements for the IndexScan, the following tableScan gets a little bit slower, and overall performance does not really increase. This also matches with my observation that I can't get any real benefit in the TPC-H 6 query, (~1% difference).
The code was basically an integration of bensk1/indexscan-poslist in the new master, so I don't understand how you measured the improvements on TPC-H 6.
Maybe I am just missing something here. If you want to look at the code, it's under xperianer/indexscan-poslist (in my fork)

Bensk1: I will try running your code on the server were I did the initial measurements. Also, I am surprised that the old style IndexScan is so much faster than the second scan. Before, we never considered the IndexScan to be a realistic option because it was only marginally faster than a TableScan, this is totally different in your microbenchmarks.

Bensk1: I added some (useless) benchmarks to your gist. The 30% performance improvement for TPC-H Q6 might have occurred when comparing the new IndexScan to a pure TableScan version. In the past, we could not get any gain from IndexScans for that query.

He3lixxx:
Out of curiosity, I changed the `limit`, `table_scan` and `validate` operators to create SingleChunkPosLists wherever possible (code at https://github.com/XPerianer/hyrise/tree/single-chunk-pos-list). This included the places @mrks mentioned (https://github.com/hyrise/hyrise/pull/2026#discussion_r399831741). It did not result in noticable performance improvements:
image

For the PosList that points into the existing index structure, we also couldn't measure any performance improvements except for the ones that we got by simply enabling the index scan. (https://github.com/hyrise/hyrise/pull/2026#discussion_r401718337)

I think this either means that writing the pos list entries does not take significant time or the improvements we gain there are lost on other places where additional costs are added.

So, what do we do now? (cc @Bensk1)

mrks:
That in itself is a valuable finding. Until now, we didn't really have a way to benchmark this. It's unfortunate that it doesn't bring us another breakthrough, but at the same time, we won't take a compile time hit.

@Bensk1 is much more familiar with the index discussion and should comment on that part.

mrks:
One thing that could be interesting to look at is the multi-threading behavior (`--scheduler --clients 50`). A single thread is usually unable to exhaust the memory bandwidth. If the expected benefit comes from less data being transferred, I'd expect it to be more visible in MT environments.

Bensk1:
> Out of curiosity, I changed the `limit`, `table_scan` and `validate` operators to create SingleChunkPosLists wherever possible (code at https://github.com/XPerianer/hyrise/tree/single-chunk-pos-list). This included the places @mrks mentioned ([#2026 (comment)](https://github.com/hyrise/hyrise/pull/2026#discussion_r399831741)). It did not result in noticable performance improvements:
> image

Thanks for making these experiments.

> For the PosList that points into the existing index structure, we also couldn't measure any performance improvements except for the ones that we got by simply enabling the index scan. ([#2026 (comment)](https://github.com/hyrise/hyrise/pull/2026#discussion_r401718337))
> 
> I think this either means that writing the pos list entries does not take significant time or the improvements we gain there are lost on other places where additional costs are added.
> 

For the IndexScan itself, you got an improvement of ~20x by not writing the PosList entries, I would call that significant. I am pretty sure (but of course, we have to measure it) that this is going to be useful when it comes to IndexJoins and/or multi threading because it relieves the pressure on the memory bus (Markus' comment).

> So, what do we do now? (cc @Bensk1)

I am going to send you an email later today.

XPerianer:
Since this won't be merged in near future, we would pull out the renaming part of the row_id_pos_list and the our name entries in the collaborators file, and would do a seperate merge request for this, so this contains only the indexscan poslist experiments. This could then be closed.

Issue #59, opened 2017-06-06
shared_ptr vs Referenz


Bensk1:
Reference as long as object stays on stack, smart pointer when stored in object

mrks:
> The exception to this rule is when the called function does something that removes that "Instance up in the stack". We had this case in our JoinDetectionRule. But I guess that's too exotic to warrant passing shared_ptrs around by value all the time

Was that because you passed around a shared_ptr by reference or was it because you actually stored that reference somewhere. As long as the current method remains active (and everything above it on the stack does too), I can't see where this would become a problem.

mrks:
Please have a look at this and tell me what you think:
[Flowchart Pointer.pdf](https://github.com/hyrise/hyrise/files/1461407/Flowchart.Pointer.pdf)

TwoLaid:
My first thought was: "the poor student who has to go through the code base and check/fix all of that..."

Otherwise, all paths make sense to me, although I still feel like a flow chart is a bit of an overkill for a guideline. Maybe it's more of a "further read" kinda deal

mjendruk:
> Please have a look at this and tell me what you think:
Flowchart Pointer.pdf

I looked at it and found some things to discuss. I’ll come to your office tomorrow.

mrks:
I'll be on my way to TechEd tomorrow. As of now, I am not planning to come to Potsdam tomorrow. We can either discuss that here, on Skype, or once I'm back (Friday).

mjendruk:
> Please have a look at this and tell me what you think:
Flowchart Pointer.pdf

The green part ("Do I need ownership?") makes sense to me. The most fundamental question is probably whether the object is going to be stored somewhere and used outside of the current call stack.

The left side of the red part (trivial types) is correct, although it doesn’t say anything about larger data structures (`std::vector` etc.). The right side is mostly okay. Although it makes sense to say: if you're sure that you are going to share the ownership, just accept a `shared_ptr` without reference. That’s the case for example in constructors.

The main problem I have is with the no-branch. If an object is internally managed via a `unique_ptr` but ownership is never transferred, it doesn’t make sense in my opinion to expose this implementation detail by passing or returning a const-reference to the underlying `unique_ptr`. In this case, it’s better to accept or return a const& to the object itself (or a const* if the value can be null).

[...]

PR #2028, opened 2020-02-20
Avoid unnecessary copies of iterators

Bouncner:
Similar to recent changes in the FrameOfReference iterable, this PR changes the iterator constructors to accept forward references and move iterators.

This PR is open for discussion as I am not too certain that I fully understand all the implications of this change.


src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -136,8 +136,8 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
 
     explicit PointAccessIterator(const pmr_vector<T>* values, const pmr_vector<bool>* null_values,
                                  const pmr_vector<ChunkOffset>* end_positions,
-                                 const PosList::const_iterator position_filter_begin,
-                                 PosList::const_iterator position_filter_it)
+                                 const PosList::const_iterator&& position_filter_begin,
```
mrks: I don't think `const T&&` ever makes any sense.

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -136,8 +136,8 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
+                                 PosList::const_iterator&& position_filter_it)
         : BasePointAccessSegmentIterator<PointAccessIterator, SegmentPosition<T>>{std::move(position_filter_begin),
```
mrks: This was already broken before. I am not sure why `std::move` on a const-qualified type is legal.

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -52,13 +52,13 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
     using EndPositionIterator = typename pmr_vector<ChunkOffset>::const_iterator;
 
    public:
-    explicit Iterator(const ValueIterator& value_it, const NullValueIterator& null_value_it,
-                      const EndPositionIterator& end_position_it, const EndPositionIterator& end_position_begin_it,
+    explicit Iterator(const ValueIterator&& value_it, const NullValueIterator&& null_value_it,
```
mrks: Not sure what you are trying to achieve here. As `Iterator` is always constructed with a prvalue, it should take its argument by value and move them into the members.

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
Bouncner: I read a few SO threads about it and from my understanding, it's the same as a const parameter passed by value. It's not required by helps to state the intend that this is not being changed.

Bouncner: But I have to admit that I did not really consider it in this PR. Would set it on appropriate parameters.
Your opinion on `const T&&`?

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
Bouncner: I personally don't like constructors that potentially create copies or corrupt the caller's data. From my (limited) understanding, we simply better enforce that prvalues are passed and that we cannot run into certain situations.

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
mrks: > Your opinion on const T&&?

Should be illegal.

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
mrks: I'd only address this in cases where it actually becomes an issue. 90%+ of developers don't get universal references right (and I don't necessarily count myself towards the other 10%).

Bouncner: Your call then. You have a much greater experience with students working on Hyrise.
If you think it might confuse people, I will only keep the changes that we always move when possible and through out the `&&`.

mrks: I'd take it by value.

mrks:
This feels like #59 all over again. In the end, I expect this to be irrelevant from a performance standpoint and believe that readability (also for people unfamiliar with universal references) should be more important.

Anyway. Bigger fish to fry.

Bouncner:
I somewhat agree as I have learned how complex such things can easily get. I was about to push a change today that removes the last forwarding reference. Sorry for not marking the PR as a WIP again. There will be another RLE pull request the next days anyways.

I tried to benchmark the changes with Q6 and a chunk size of 100:
```
+----------------+--------------------+------+-------------------+------+------------+---------+
| Benchmark      | prev. iter/s       | runs | new iter/s        | runs | change [%] | p-value |
+----------------+--------------------+------+-------------------+------+------------+---------+
 >> Uncompressed:
| TPC-H 06       | 10.004405975341797 | 6003 | 10.13805866241455 | 6083 | +1%        |  0.0000 |
 >> RLE:
| TPC-H 06       | 6.098711967468262 | 3660 | 6.1403632164001465 | 3685 | +1%        |  0.0854 |
 >> LZ4:
| TPC-H 06       | 5.27440881729126 | 3165 | 5.370595455169678 | 3223 | +2%        |  0.0000 |
 >> Dict with SIMDBP128:
| TPC-H 06       | 6.515908241271973 | 3910 | 6.525854587554932 | 3916 | +0%        |  0.6802 |
| geometric mean |                   |      |                   |      | +0%        |         |
+----------------+-------------------+------+-------------------+------+------------+---------+
```

Can very well be noise, but since it's positive for all cases, I'd assume the few more `std::move`s make a difference (yes, not measurable for an acceptable chunk size, but still ...).

PR #2029, opened 2020-02-21
Delete docker-compose.yml

mrks:
@tobodner confirmed that docker-compose does not make any sense if we only have one container. As this is not tested in CI (and currently broken), it seems safe to remove this.

Also updated https://github.com/hyrise/hyrise/wiki/Installation


Bouncner:
We should adapt the README.md with that PR as well. Is it `dockäh build/up` now?

cH3n7i:
This won't change the outcome, it's still failing with the same behavior with macOS 10.15.3 as Docker host.

mrks:
Interesting. I won't close the issue then - let's discuss the issue there.

PR #2031, opened 2020-02-24
[DYOD] Query Plan Cache Parametrization

s4pu:
WIP


Bensk1:
What's the status on that? Can you give us a heads up when you are ready for a review?

Bensk1:
Could you please post the current state of the TPC-H query 15 issue here as a comment? I am thinking of the example you sent via e-mail to me.

carlaterboven:
> Could you please post the current state of the TPC-H query 15 issue here as a comment? I am thinking of the example you sent via e-mail to me.

The query fails with ```Invalid input error: Failed to resolve Column 'l_extendedprice', LQP is invalid```

The original TPCH query looks like this:
```
create view revenue_view (supplier_no, total_revenue) as SELECT l_suppkey, SUM(l_extendedprice * (1 - l_discount)) FROM lineitem WHERE l_shipdate >= '1996-01-01' AND l_shipdate < '1996-04-01' GROUP BY l_suppkey;
SELECT s_suppkey, s_name, s_address, s_phone, total_revenue FROM supplier, revenue_view WHERE s_suppkey = supplier_no AND total_revenue = (SELECT max(total_revenue) FROM revenue_view) ORDER BY s_suppkey;
```
We are able to create the view but it fails when executing the select statement. Moreover it woks when using ```SUM(l_extendedprice * l_discount)``` instead of ```SUM(l_extendedprice * (1 - l_discount))```.

Additionally we tried to execute the query without the view like that: 
```SELECT s_suppkey, s_name, s_address, s_phone, total_revenue FROM supplier, (SELECT l_suppkey as supplier_no, SUM(l_extendedprice * (1 - l_discount)) as total_revenue FROM lineitem WHERE l_shipdate >= '1996-01-01' AND l_shipdate < '1996-04-01' GROUP BY l_suppkey) as revenue_view WHERE s_suppkey = supplier_no AND total_revenue = (SELECT max(total_revenue)) ORDER BY s_suppkey;``` 
And this query works even when using the **1 -**.

**Comment by Bensk1** [comment]
Thanks, @carlaterboven!
 
> The query fails with `Invalid input error: Failed to resolve Column 'l_extendedprice', LQP is invalid`

@mrks, without looking into it, do you have any idea here? Also, we agreed on not caching `create view` statements, even though the error occurs when using the view.

**Comment by mrks** [comment]
I'd start by checking if the LQPColumnExpression (and the LQPColumnReference inside, see #1893) reference an original_node that is actually part of the sub-LQP. If you copied parts of a plan without adapting the column expressions, you would see the same message.

**Comment by Bensk1** [comment]
The shadow warning can be resolved by renaming the constructor's parameter. Some examples are [here](https://github.com/hyrise/hyrise/pull/2040).

**Comment by carlaterboven** [comment]
Unfortunately, we cannot reproduce the gcc error on our machines. Any idea why the tests might fail?

**Comment by Bensk1** [comment]
> Unfortunately, we cannot reproduce the gcc error on our machines. Any idea why the tests might fail?

What's your setup? I will try to have a look later.

**Comment by carlaterboven** [comment]
> > Unfortunately, we cannot reproduce the gcc error on our machines. Any idea why the tests might fail?
> 
> What's your setup? I will try to have a look later.

Our setup is the vm-nordes and a local docker.

**Comment by Bensk1** [comment]
Okay, I am looking into that. Apart from that, have you considered resolving the linting errors?

**Comment by Bensk1** [comment]
I cleared the ccache files on the CI, it seems to work now. The only thing that is keeping us from running FullCI are the linting errors now.

**Comment by carlaterboven** [comment]
Thanks! :blush:

**Comment by mrks** [comment]
(cancelling the current CI build, which has already failed in gccRelease, to clean up resources)

**Comment by s4pu** [comment]
In `cd clang-release && ../scripts/test/hyriseBenchmarkTPCH_test.py`, we receive the following error:

```- Starting Benchmark...
[PERF] Build relation larger than probe relation in hash join at src/lib/operators/join_hash.cpp:89

[PERF] BETWEEN handled as two table scans because no BETWEEN specialization was available at src/lib/operators/operator_scan_predicate.cpp:106

terminate called after throwing an instance of 'opossum::InvalidInputException'
  what():  Invalid input error: Failed to resolve Column 'l_extendedprice', LQP is invalid
script returned exit code 6 
```

We encountered this error before in TPCH Query 15, where we fixed it by not caching views. This current occurrence is not connected to a view, therefore we don't know where it is coming from. We cannot reproduce this error on our machines nor in the VM. Any ideas?

Bensk1:
It might disappear by the fixes you made. But for the last run the linter was failing again.

s4pu:
How can we clear the ccache files ourselves?

mrks:
I disabled it in the CI a couple of minutes ago and restarted your build.

src/lib/sql/sql_plan_cache.hpp
```diff
@@ -4,13 +4,29 @@
...
+};
+class DeepEquality {
+ public:
+  bool operator()(const std::shared_ptr<AbstractLQPNode>& key1, const std::shared_ptr<AbstractLQPNode>& key2) const {
+    return *key1 == *key2;
+  }
+};
 
+using SQLPhysicalPlanCache = Cache<std::shared_ptr<AbstractOperator>, std::string>;
+using SQLLogicalPlanCache = Cache<std::shared_ptr<PreparedPlan>, std::shared_ptr<AbstractLQPNode>, Hash, DeepEquality>;
```
boehmchen, line 28: At the moment I do not understand, why you open up a new sub namespace here.

src/test/sql/sql_pipeline_statement_test.cpp
```diff
@@ -690,14 +708,13 @@ TEST_F(SQLPipelineStatementTest, PrecheckDDLOperators) {
 TEST_F(SQLPipelineStatementTest, MetaTableNoCaching) {
   const auto meta_table_query = "SELECT * FROM " + MetaTableManager::META_PREFIX + "tables";
 
-  auto sql_pipeline = SQLPipelineBuilder{meta_table_query}
-                          .with_lqp_cache(_lqp_cache)
+  auto sql_pipeline = SQLPipelineBuilder{meta_table_query}  // .with_lqp_cache(_lqp_cache)
```
boehmchen, line 711: Remove commented code

src/test/sql/sql_pipeline_statement_test.cpp
```diff
@@ -690,14 +708,13 @@ TEST_F(SQLPipelineStatementTest, PrecheckDDLOperators) {
...
                           .with_pqp_cache(_pqp_cache)
                           .create_pipeline_statement();
   sql_pipeline.get_result_table();
 
-  EXPECT_EQ(_lqp_cache->size(), 0u);
-  EXPECT_FALSE(_lqp_cache->has(meta_table_query));
+  // EXPECT_EQ(_lqp_cache->size(), 0u);
+  // EXPECT_FALSE(_lqp_cache->has(meta_table_query));
```
boehmchen, line 717: Remove commented code.

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -1184,6 +1184,9 @@ RowIDPosList ExpressionEvaluator::evaluate_expression_to_pos_list(const Abstract
       }
     } break;
 
+    case ExpressionType::Placeholder:
+      Fail("Found the placeholder");
```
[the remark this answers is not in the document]
carlaterboven: I think that was part of CI debugging and can be removed.

src/lib/sql/sql_translator.cpp
```diff
@@ -578,6 +578,7 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_origin(const hsq
 
       } else if (Hyrise::get().storage_manager.has_view(hsql_table_ref.name)) {
         const auto view = Hyrise::get().storage_manager.get_view(hsql_table_ref.name);
+        _cacheable = false;
```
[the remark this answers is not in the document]
s4pu: 1. to avoid a bug when caching (in the pqp cache) selected queries that use views: when you drop the view after the query and send the same query again, hyrise should throw an error, instead it executes the cached pqp.
2. https://github.com/hyrise/hyrise/pull/2031#issuecomment-605879687

src/lib/cache/abstract_cache_impl.hpp
```diff
@@ -7,7 +7,7 @@
 namespace opossum {
 
 // Generic template for a cache implementation.
-template <typename Key, typename Value>
+template <typename Key, typename Value, typename Hash, typename KeyEqual>
```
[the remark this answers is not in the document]
tzagorni: We are now using the unoptimzed LQP as key, which is a pointer to a LQP Node. The default implementations would only check the pointer equality, while we want to compare the whole plans

src/lib/cache/cache.hpp
```diff
@@ -16,13 +16,13 @@ namespace opossum {
 inline constexpr size_t DefaultCacheCapacity = 1024;
 
 // Per-default, uses the GDFS cache as underlying storage.
-template <typename Value, typename Key = std::string>
+template <typename Value, typename Key = std::string, class Hash = std::hash<Key>, class KeyEqual = std::equal_to<Key>>
```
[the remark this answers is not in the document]
tzagorni: No we pass them in sql_plan_cache.hpp for the same reason as above

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -97,45 +102,113 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_unoptimized_lo
...
+                  assert(expression->arguments.empty());
+                  values.push_back(expression);
+                  auto new_expression =
+                      std::make_shared<TypedPlaceholderExpression>(parameter_id, expression->data_type());
+                  expression->replaced_by = new_expression;
+                  expression = new_expression;
+                  parameter_id++;
+                }
+              }
+            }
```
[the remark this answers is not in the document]
tzagorni: this is actually only the way "real expression"-> placeholder. The other way uses PreparedPlan::instanciate. Sightly improved readability now by splitting into two methods. Could possibly still use some comments

Bensk1:
While the performance figures for TPC-C are even better now that they were during the final presentation, the current implementation results in massive performance degradation for other benchmarks. This is due to the way the cache mechanics currently work:

1. ValueExpressions are removed from the unoptimized LQP and replaced by placeholders, this results in a parametrized LQP
1. The resulting LQP serves as cache key and is handed into the optimizer
1. Because the ValueExpressions are removed, some of the optimization rules, e.g., the BetweenComposition, do not work at all which leads to the performance degradation
1. The _optimized_ LQP without ValueExpressions is stored as cache value and its placeholders are replaced.

This could be solved by using the parametrized LQP as cache key and the vanilla LQP as input for the optimizer. The problem here is that the optimizer might shuffle around predicates or even remove them completely. Therefore, it is necessary to reflect these changes for the placeholders in the parametrized LQP, otherwise placeholder replacement will not work properly.

This is not going to be implemented as part of the seminar's group project but can serve as a striating point when we decide to further pursue caching in the future.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -97,45 +102,119 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_unoptimized_lo
   return _unoptimized_logical_plan;
 }
 
+void SQLPipelineStatement::split_expression(std::shared_ptr<AbstractExpression> &expression,
```
Bensk1, line 105: Is split the best name for this method?

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -97,45 +102,119 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_unoptimized_lo
...
+  const auto started_preoptimization_cache = std::chrono::high_resolution_clock::now();
+
+  std::vector<std::shared_ptr<AbstractExpression>> values;
+  const auto unoptimized_lqp = get_split_unoptimized_logical_plan(values);
+
   // Handle logical query plan if statement has been cached
   if (lqp_cache) {
-    if (const auto cached_plan = lqp_cache->try_get(_sql_string)) {
-      const auto plan = *cached_plan;
+    if (const auto cached_plan = *(lqp_cache->try_get(unoptimized_lqp))) {
```
Bensk1, line 162: Why do we need the `*` here?

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -97,45 +102,119 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_unoptimized_lo
...
+      const auto plan = cached_plan->lqp;
       DebugAssert(plan, "Optimized logical query plan retrieved from cache is empty.");
       // MVCC-enabled and MVCC-disabled LQPs will evict each other
       if (lqp_is_validated(plan) == (_use_mvcc == UseMvcc::Yes)) {
         // Copy the LQP for reuse as the LQPTranslator might modify mutable fields (e.g., cached column_expressions)
         // and concurrent translations might conflict.
-        _optimized_logical_plan = plan->deep_copy();
+        _optimized_logical_plan = cached_plan->instantiate(values);
+        auto olqp = _optimized_logical_plan->deep_copy();
```
Bensk1, line 170: naming

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -97,45 +102,119 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_unoptimized_lo
...
-
-  const auto started = std::chrono::high_resolution_clock::now();
-
   // The optimizer works on the original unoptimized LQP nodes. After optimizing, the unoptimized version is also
   // optimized, which could lead to subtle bugs. optimized_logical_plan holds the original values now.
   // As the unoptimized LQP is only used for visualization, we can afford to recreate it if necessary.
   _unoptimized_logical_plan = nullptr;
 
-  _optimized_logical_plan = _optimizer->optimize(std::move(unoptimized_lqp));
+  auto ulqp = unoptimized_lqp->deep_copy();
```
Bensk1, line 186: Why do we need the `deep_copy` here? Also, naming.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -97,45 +102,119 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_unoptimized_lo
...
+}
+
 const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_optimized_logical_plan() {
   if (_optimized_logical_plan) {
     return _optimized_logical_plan;
   }
 
+  const auto started_preoptimization_cache = std::chrono::high_resolution_clock::now();
+
+  std::vector<std::shared_ptr<AbstractExpression>> values;
```
Bensk1, line 157: naming

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -97,45 +102,119 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_unoptimized_lo
...
+  auto prepared_plan = std::make_shared<PreparedPlan>(optimized_without_values, all_parameter_ids);
 
   // Cache newly created plan for the according sql statement
   if (lqp_cache && _translation_info.cacheable) {
-    lqp_cache->set(_sql_string, _optimized_logical_plan);
+    lqp_cache->set(unoptimized_lqp, prepared_plan);
   }
 
+  _optimized_logical_plan = prepared_plan->instantiate(values);
+  auto olqp = _optimized_logical_plan->deep_copy();
```
Bensk1, line 209: See 186.

src/lib/sql/sql_plan_cache.hpp
```diff
@@ -4,13 +4,25 @@
...
+#include "storage/prepared_plan.hpp"
 
 namespace opossum {
 
 class AbstractOperator;
 class AbstractLQPNode;
 
-using SQLPhysicalPlanCache = Cache<std::shared_ptr<AbstractOperator>, std::string>;
-using SQLLogicalPlanCache = Cache<std::shared_ptr<AbstractLQPNode>, std::string>;
+class NodePointerHash {
```
Bensk1, line 15: LQPNode...?

src/lib/sql/sql_pipeline_statement.cpp
s4pu: Yes, we think this is necessary to unwrap the optional we receive from `try_get()`.

src/lib/sql/sql_pipeline_statement.cpp
s4pu: Otherwise, we only get the first node.
We need the complete lqp again in order to move it into the optimizer, where it can be changed without side effects.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -97,45 +102,119 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_unoptimized_lo
   return _unoptimized_logical_plan;
 }
 
+void SQLPipelineStatement::split_expression(std::shared_ptr<AbstractExpression> &expression,
+                                            std::vector<std::shared_ptr<AbstractExpression>>& values,
+                                            ParameterID &next_parameter_id) {
+  if (expression->type == ExpressionType::Value) {
+    if (expression->replaced_by) {
+      const auto valexp = std::dynamic_pointer_cast<ValueExpression>(expression);
+      expression = expression->replaced_by;
```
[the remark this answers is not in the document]
tzagorni: An expression can be referenced from multiple points in the graph, meaning we find them multiple times during the visiting. When we replace a Value with an expression the second pointer still points to the original value. If we don't check this this causes us to create 2 placeholders for what used to be one expression leading to weird crashes.

Bensk1:
As discussed, closing for now but we will continue to work on that and reopen it later.

PR #2037, opened 2020-02-28
Create chunk pruning statistics in StorageManager::AddTable

Bouncner:
This PR create `ChunkPruningStatistics` whenever tables are added to the storage manager.

When tables are added to the storage manager, currently only table statistics are created but no chunk pruning statistics. That leads to cases when data is loaded and no chunk pruning takes place, because the `ChunkPruningRule` cannot find any chunk pruning statistics.

Currently, chunk pruning statistics are generated when data is encoded. Hence, the case of no pruning statistics usually does not happen in our benchmarks.

Moreover, we currently might create the statistics several times as `generate_chunk_pruning_statistics(table)` did not check for their presence. This leads to a reduce encoding times for TPC-H SF 1:
 * Dictionary: from `3 s 857 ms` to `3 s 184 ms` (statistics creation is optimized here using the dictionaries, subtle difference)
 * RLE: from `16 s 743 ms` to `3 s 421 ms`
 * LZ4: from `1 min 3 s` to `45 s 673 ms`

The reason for the clear improvement with RLE is that the first creation is done in parallel during encoding while the second iteration is sequential.

### DIFF
--- a/src/lib/storage/storage_manager.cpp
+++ b/src/lib/storage/storage_manager.cpp
@@ -30,6 +30,7 @@ void StorageManager::add_table(const std::string& name, std::shared_ptr<Table> t
   }
 
   table->set_table_statistics(TableStatistics::from_table(*table));
+  generate_chunk_pruning_statistics(table);
   _tables.emplace(name, std::move(table));
 }
 
--- a/src/lib/storage/storage_manager.cpp
+++ b/src/lib/storage/storage_manager.cpp
@@ -29,8 +29,10 @@ void StorageManager::add_table(const std::string& name, std::shared_ptr<Table> t
     Assert(table->get_chunk(chunk_id)->has_mvcc_data(), "Table must have MVCC data.");
   }
 
+  // Create table statistics and chunk pruning statistics for added table.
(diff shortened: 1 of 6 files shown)


src/lib/storage/storage_manager.cpp
```diff
@@ -29,7 +29,10 @@ void StorageManager::add_table(const std::string& name, std::shared_ptr<Table> t
     Assert(table->get_chunk(chunk_id)->has_mvcc_data(), "Table must have MVCC data.");
   }
 
+  // Create table statistics and chunk pruning statistics for added table.
   table->set_table_statistics(TableStatistics::from_table(*table));
+  generate_chunk_pruning_statistics(table);
```
mrks: Why do we have one version where we check for existing statistics and one without? Should generate_chunk_pruning_statistics fail if statistics are already present?

Bouncner: Good point. Moved the check into the actual `generate_chunk_pruning_statistics(chunk)` method and removed the check from other places.

Does that clear it up?

PR #2038, opened 2020-03-02
Implement advance() for RunLength Segment Iterables

Bouncner:
For a current project (which uses the iterator fix discussed in #1531) we frequently use sorted searches. Currently, the RunLength segment iterables do not support a proper `advance()` but rather iteratively call `++` or `--`.

This PR implements a "proper" `advance()` method. Further, the search for end positions has changed. Previously, the "next" chunk offset was resolved via a linear search, no matter how large the step from the previous to the current offset was (it worked well for most cases, but for large jumps a binary search is much faster, especially for SF 10).
Now, we first estimate a threshold of when to use linear/binary search and then use this threshold to determine which search method shall be used.

We did several benchmarks:
 * The `advance()` method significantly improves the performance of sorted searches. But since this currently does not apply to the master (cf. #1531), I'll skip the results here.
 * The new search approach has been measured with various step sizes for single-threaded TPC-H runs on nemea. A value of 200 turned out to be best and is now used. Results below.

TPC-H on nemea with everything as RLE, current master vs this PR (no iterator fix used):
```diff
 +----------------+--------------------+------+--------------------+------+------------+---------+
 | Benchmark      | prev. iter/s       | runs | new iter/s         | runs | change [%] | p-value |
 +----------------+--------------------+------+--------------------+------+------------+---------+
 | TPC-H 01       | 0.9688083529472351 | 59   | 0.9718952178955078 | 59   | +0%        |  0.8080 |
 | TPC-H 02       | 53.040733337402344 | 3183 | 53.3393669128418   | 3201 | +1%        |  0.3525 |
 | TPC-H 03       | 6.540199279785156  | 393  | 6.499882698059082  | 391  | -1%        |  0.2390 |
 | TPC-H 04       | 5.496410846710205  | 330  | 5.483299732208252  | 329  | -0%        |  0.6706 |
 | TPC-H 05       | 2.611407995223999  | 157  | 2.5479562282562256 | 153  | -2%        |  0.0019 |
 | TPC-H 06       | 34.22892379760742  | 2054 | 34.09223556518555  | 2046 | -0%        |  0.2584 |
 | TPC-H 07       | 6.675305366516113  | 401  | 6.62246036529541   | 398  | -1%        |  0.1663 |
+| TPC-H 08       | 5.45106840133667   | 328  | 5.7662529945373535 | 346  | +6%        |  0.0000 |
 | TPC-H 09       | 0.9445616602897644 | 57   | 0.9435319900512695 | 57   | -0%        |  0.7992 |
 | TPC-H 10       | 2.7358782291412354 | 165  | 2.7698776721954346 | 167  | +1%        |  0.1320 |
 | TPC-H 11       | 23.260658264160156 | 1396 | 23.23638916015625  | 1395 | -0%        |  0.4357 |
 | TPC-H 12       | 4.988798141479492  | 300  | 4.974976062774658  | 299  | -0%        |  0.0479 |
 | TPC-H 13       | 1.6632800102233887 | 100  | 1.6539808511734009 | 100  | -1%        |  0.0044 |
 | TPC-H 14       | 26.685537338256836 | 1602 | 26.686750411987305 | 1602 | +0%        |  0.9539 |
 | TPC-H 15       | 91.73336029052734  | 5505 | 91.94815063476562  | 5517 | +0%        |  0.0004 |
 | TPC-H 16       | 4.892538547515869  | 294  | 4.905280113220215  | 295  | +0%        |  0.4701 |
+| TPC-H 17       | 4.356827259063721  | 262  | 4.683340072631836  | 282  | +7%        |  0.0000 |
 | TPC-H 18       | 1.1319807767868042 | 68   | 1.1755223274230957 | 71   | +4%        |  0.0027 |
+| TPC-H 19       | 7.172689437866211  | 431  | 7.8536906242370605 | 472  | +9%        |  0.0000 |
 | TPC-H 20       | 17.732011795043945 | 1064 | 17.891326904296875 | 1074 | +1%        |  0.4707 |
 | TPC-H 21       | 0.8144749402999878 | 49   | 0.8060727119445801 | 49   | -1%        |  0.2012 |
 | TPC-H 22       | 17.431291580200195 | 1046 | 17.305862426757812 | 1039 | -1%        |  0.0106 |
 | geometric mean |                    |      |                    |      | +1%        |         |
 +----------------+--------------------+------+--------------------+------+------------+---------+
```


src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -153,19 +157,37 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
       auto end_position_it = _end_positions->cend();
 
-      if (current_chunk_offset < _prev_chunk_offset) {
+      const int64_t step_size = static_cast<int64_t>(current_chunk_offset) - _prev_chunk_offset;
+      /**
+       * Depending on the estimated threshold and the step size, a different search approach is used. The threshold
+       * estimates how long a jump needs to be before a binary search is faster than linearly searching.
+       * Three cases are handled:
+       *   - If the chunk offset is smaller then the previous offset (can happen, e.g., after joins), use a binary
+       *     search search from the beginning up to the previous position. Whenever reverse iteration is frequently
```
mrks: renard renard?

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -52,13 +52,13 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
 
    public:
-    explicit Iterator(ValueIterator value_it, NullValueIterator null_value_it,
-                      EndPositionIterator end_position_it, EndPositionIterator end_position_begin_it,
-                      ChunkOffset chunk_offset)
-        : _value_it{std::move(value_it)},
-          _null_value_it{std::move(null_value_it)},
-          _end_position_it{std::move(end_position_it)},
-          _end_position_begin_it{std::move(end_position_begin_it)},
+    explicit Iterator(const pmr_vector<T>* values, const pmr_vector<bool>* null_values,
```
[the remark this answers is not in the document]
Bouncner: Changed to shared pointers. Not done yet though.

Bouncner:
Just to give an update here: the code finished the FullCI before. After some cleanups (comments and var names), so without any significant changes, the PR crashes very often in a test that shouldn't touch the RunLength iterable. Not sure what the problem is.

Apart from that, the code is ready to be reviewed.

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -44,143 +43,189 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
+      const ChunkOffset previous_chunk_offset, const ChunkOffset current_chunk_offset,
+      const size_t previous_end_position_index, const size_t linear_search_threshold,
+      const std::shared_ptr<const pmr_vector<ChunkOffset>>& end_positions) {
+    const int64_t step_size = static_cast<int64_t>(current_chunk_offset) - previous_chunk_offset;
+
+    /**
+     * Depending on the estimated threshold and the step size, a different search approach is used. The threshold
+     * estimates how large the offset between the previously searched chunk offset and the currently searched chunk
+     * offset needs to be before a binary search is faster than linearly searching. Three cases are handled:
+     *   - If the chunk offset is smaller then the previous offset (can happen, e.g., after joins), use a binary
```
j-tr: *than

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -44,143 +43,189 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
+   * determine_linear_search_offset_distance_threshold() estimates the threshold of when to use a linear or a binary
+   * search. Given the previous and the current chunk offset, the threshold is used to determine if a linear search is
+   * faster for the given number of skipped positions.
+   */
+
+  // LINEAR_SEARCH_ELEMENTS_THRESHOLD denotes the size of the range for that a linear search is faster than using a
+  // binary search on a sorted vector. The value of 200 has been found by a set of simple TPC-H measurements (#2038).
+  static constexpr auto LINEAR_SEARCH_VECTOR_DISTANCE_THRESHOLD = 200.0f;
+
+  // This methods estimates the threshold up to which is linear search is faster. Since the end position vector is
```
mrks: But above you have defined the threshold to be 200. For me, the naming is a bit unclear.

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -44,143 +43,189 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
+  // (depending on actual data) a highly compressed vector of offsets, the offset distance up to which is linear search
+  // is beneficial can be much larger than LINEAR_SEARCH_VECTOR_DISTANCE_THRESHOLD.
+  static ChunkOffset determine_linear_search_offset_distance_threshold(
+      const std::shared_ptr<const pmr_vector<ChunkOffset>>& end_positions) {
+    if (end_positions->empty()) {
+      return 0;
+    }
+
+    const ChunkOffset chunk_size = end_positions->back();
```
mrks: Should doc that you are expecting all chunks to be of the same size.

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -44,143 +43,189 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
+    const size_t run_count = end_positions->size();
+
+    const auto avg_elements_per_run = static_cast<float>(chunk_size) / run_count;
+    return static_cast<ChunkOffset>(LINEAR_SEARCH_VECTOR_DISTANCE_THRESHOLD * std::ceil(avg_elements_per_run));
+  }
+
+  using EndPositionIterator = typename pmr_vector<ChunkOffset>::const_iterator;
+  static EndPositionIterator search_end_positions_for_chunk_offset(
```
mrks: This interface looks like a bunch of things just dumped in here. Wouldn't it be easier if this was non-static?

Bouncner: I know it looks ugly and I'll take another look at it. But I am (i) not a big fan of non-static methods and (ii) the function is used in both iterators. I'd guess accessing members within the method isn't possible?!?

mrks: Could you explain your position on (i)? I didn't notice (ii), that makes sense.

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -44,143 +43,192 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
  private:
   const RunLengthSegment<T>& _segment;
 
+  /**
+   * Due to the nature of Run Length encoding, point-access is not in O(1).
+   * Run length segments store the end positions of runs in a sorted vector. To access a particular position, this
+   * vector has to be searched for the end position of the according run. It is possible to find the value of a
+   * position in O(log(n)) by doing a binary search.
+   * In case we have already searched for a nearby position (e.g., when a previous not selective filter removed only
+   * relatively few lines), we can linearly search the end position. Since position lists are often ordered and modern
```
mrks: Minor: previously, search *for*

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -44,143 +43,192 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
+  // use linear searches for offset distances of LINEAR_SEARCH_VECTOR_DISTANCE_THRESHOLD * 6.5. For sorted vectors with
+  // few distinct values (superb cases for run length encoding), binary searches are barely used.
+  static ChunkOffset determine_linear_search_offset_distance_threshold(
+      const std::shared_ptr<const pmr_vector<ChunkOffset>>& end_positions) {
+    if (end_positions->empty()) {
+      return 0;
+    }
+
+    const ChunkOffset chunk_size = end_positions->back();
+    const size_t run_count = end_positions->size();
```
mrks: Not a friend of auto, are you? ;)

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -44,143 +43,192 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
+
+    const auto avg_elements_per_run = static_cast<float>(chunk_size) / run_count;
+    return static_cast<ChunkOffset>(LINEAR_SEARCH_VECTOR_DISTANCE_THRESHOLD * std::ceil(avg_elements_per_run));
+  }
+
+  using EndPositionIterator = typename pmr_vector<ChunkOffset>::const_iterator;
+  static EndPositionIterator search_end_positions_for_chunk_offset(
+      const ChunkOffset previous_chunk_offset, const ChunkOffset chunk_offset,
+      const size_t previous_end_position_index, const size_t linear_search_threshold,
+      const std::shared_ptr<const pmr_vector<ChunkOffset>>& end_positions) {
```
mrks: I now understand better what is happening here. My suggestion to slightly clean up the interface are:

* move end_positions to the front (no idea why, it just feels better to put the thing you operating on first)
* chunk_offset is confusing here. Isn't it a chunk_offset_offset? How do you feel about `distance`?

src/lib/storage/run_length_segment/run_length_segment_iterable.hpp
```diff
@@ -44,143 +43,192 @@ class RunLengthSegmentIterable : public PointAccessibleSegmentIterable<RunLength
...
 
     void decrement() {
       --_chunk_offset;
-
-      // Make sure to only check the previous end position when we are not in the very first run.
-      if (_end_position_it != _end_position_begin_it && _chunk_offset <= *(_end_position_it - 1)) {
-        --_value_it;
-        --_null_value_it;
-        --_end_position_it;
+      if (_end_positions_it != _end_positions_begin_it && _chunk_offset <= *(_end_positions_it - 1)) {
```
mrks: Do we need to check for the begin? If anything decrementing when we have reached the begin should enter a state in which dereferencing is invalid. UB would be fine, too.

mrks: In any case, this invariant should hold:

```
auto it = begin();
--it;
++it;
*it == *begin();
```

Should we test for this?
```

Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Wasn’t aware that this is allowed. Will take a look at it.

File: src/lib/storage/run_length_segment/run_length_segment_iterable.hpp (hunk 6 above)
Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Nothing special, just that reasoning is harder when you access "state" of the object. If it would be a member, we could also modify the state variables directly in the method (e.g., `_prev_chunk_offset`). Because why not, if it's a member, right?
But this would make this method even more hard to understand since it's no longer `const`. Might be my way of understanding code, but I often prefer several ugly parameters over "implicitly imported state variable" (there are probably better terms :D).

File: src/lib/storage/run_length_segment/run_length_segment_iterable.hpp (hunk 10 above)
Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: I think survive "by luck" in most iterators as we use a variable of `ChunkOffset` to store the current position and that one will be decremented from the value of 0 to 2^32-1 and than incremented again. I don't know if this is not actually UB?

Do you know if decrementing from `begin()` is actually valid?

Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: I did not find the actual point in the specification (24.2.6 is somewhat related, but did not find it).
http://www.cplusplus.com/reference/iterator/BidirectionalIterator/ states for a bidirectional iterator and the expression `--a`: `Can be decremented (if a dereferenceable iterator value precedes it).`

Since there is not a dereferencable value that precedes `it` with `it = begin()`, I'd assume that `--it` is not valid.

File: src/lib/storage/run_length_segment/run_length_segment_iterable.hpp (hunk 7 above)
Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Where would you use `previously` over `previous`? Or did I get that wrong?
Added the `for`.

File: src/lib/storage/run_length_segment/run_length_segment_iterable.hpp (hunk 8 above)
Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Changed.

File: src/lib/storage/run_length_segment/run_length_segment_iterable.hpp (hunk 9 above)
Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: `chunk_offset` is rather the `current_chunk_offset` that is being searched. The `distance` is currently called `step_size`. Do you mean this `step_size`? I would be fine with `distance` here.

File: src/lib/storage/run_length_segment/run_length_segment_iterable.hpp (hunk 4 above)
Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Updated the comment. It's about the threshold for the given RLE-end positions vector.

Does that make it clear?

**Comment by Bouncner** [comment]
Added a WIP because I'd like to run a final benchmark removing the shared pointers when the PR is ready to go.

File: src/lib/storage/run_length_segment/run_length_segment_iterable.hpp (hunk 9 above)
Comment (mrks) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Ok, misread it. `old_chunk_offset` and `new_`?

File: src/lib/storage/run_length_segment/run_length_segment_iterable.hpp (hunk 10 above)
Comment (mrks) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: I see. So ++it can be out-of-bounds, but --it can't. Thanks for looking that up. Maybe add a DebugAssert and a comment pointing to the standard?

File: src/lib/storage/run_length_segment/run_length_segment_iterable.hpp (hunk 7 above)
Comment (mrks) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Gnah, I am not happy with the quality of my review. :/

Actually, I believe it should be "when a previous **,** non-selective". Or, maybe better: "a previous filter with a low selectivity resulted in an almost contiguous PosList"?

Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Like the last one. Done.

File: src/lib/storage/run_length_segment/run_length_segment_iterable.hpp (hunk 10 above)
Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Does it make sense to link to a draft? Or do we have access to an actual standard?

Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Added
```
DebugAssert(_chunk_offset > 0, "An iterator pointing at the begin of a segment cannot be decremented, see "
                                     "§24.2.6 of http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf "
                                     "(iterator can be decremented if a dereferencable iterator value precedes it).");
```

But I am not too happy with not having a really good cite of the actual C++ standard.

Comment (mrks) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: The standard is not public. Don't get me started on that. Linking the draft is the normal modus operandi. I like https://eel.is/c++draft/ because it is easier to deep-link into a section.

Comment (Bouncner) [reply src/lib/storage/run_length_segment/run_length_segment_iterable.hpp]: Changed to
```
// According to the C++ standard (see https://eel.is/c++draft/iterator.concept.bidir) a begin() iterator cannot
      // be decremented (an iterator can be decremented if a dereferencable iterator value precedes it).
      DebugAssert(_chunk_offset > 0, "An iterator pointing at the begin of a segment cannot be decremented");
```

Bouncner:
The potential removal of shared_ptr might be tackled in another PR. I don't want to keep this PR open too long.

PR #2043, opened 2020-03-09
Initialize accesses variable with zero

Bouncner:
When working on the segment accessors, I stumpled upon an arbitrary values for the member `_accesses`. This PR simply initializes the variable.

### DIFF
--- a/src/lib/storage/segment_accessor.hpp
+++ b/src/lib/storage/segment_accessor.hpp
@@ -115,7 +115,7 @@ class SingleChunkReferenceSegmentAccessor final : public AbstractSegmentAccessor
   }
 
  protected:
-  mutable uint64_t _accesses{0};   // TODO: block überfrühten Merge
+  mutable uint64_t _accesses{0};
   const PosList& _pos_list;
   const ChunkID _chunk_id;
   const Segment& _segment;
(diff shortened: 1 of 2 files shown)


mrks:
If this is truly uninitialized, please also check why UBsan didn't complain about it. Are we missing a test?

Bouncner:
@mflueggen : can you help to get a better understanding here? I haven't seen a test covering the segement accessors explicitly as a quick first glance. Since the counters I am using here are not directly used as of now but only during destruction, you might catch such cases somewhere else. Is that the case?

mflueggen:
@Bouncner 
I forgot to test the segment accessors explicitly. I am not sure if I understood your question correctly. The segment’s counter is only increased during destruction to minimize the updates on atomics.

Bouncner:
> @Bouncner
> I forgot to test the segment accessors explicitly. I am not sure if I understood your question correctly. The segment’s counter is only increased during destruction to minimize the updates on atomics.

I guess my response was a bit verbose. :D 
I'll add a test and ping you again.

Bouncner:
@mflueggen : can you help to get a better understanding here? I haven't seen a test covering the segement accessors explicitly as a quick first glance. 

> If this is truly uninitialized, please also check why UBsan didn't complain about it. Are we missing a test?

Added a test that failed on my notebook without the fix. How can I test the UBsan thing?

@mflueggen : does the test look sound to you? Did I miss something important?

mflueggen:
The test looks good to me.

PR #2044, opened 2020-03-09
System Information / Utilization Meta Tables

j-tr:
This PR adds meta tables that contain dynamic and static information about the hardware system.


src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,353 @@
...
+}
+
+int MetaSystemUtilizationTable::_get_cpu_count() {
+#if defined __linux__
+
+  std::ifstream cpu_info_file;
+  cpu_info_file.open("/proc/cpuinfo", std::ifstream::in);
+
+  uint32_t processors = 0;
+  for (std::string cpu_info_line; std::getline(cpu_info_file, cpu_info_line);) {
```
mrks: I'm not a fan of for loops that do not follow the for loop pattern (loop variable etc). This looks much more like a while loop to me.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,353 @@
...
+    if (cpu_info_line.rfind("processor", 0) == 0) ++processors;
+  }
+
+  cpu_info_file.close();
```
mrks: Should be unnecessary thanks to RAII, but if you like closing things explicitly, that's fine too. Your call.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,353 @@
...
+  last_system_time = system_time;
+
+  return (100.0f * used) / (total);
+
+#endif
+
+  Fail("Method not implemented for this platform");
+}
+
+MetaSystemUtilizationTable::SystemMemoryUsage MetaSystemUtilizationTable::_get_system_memory_usage() {
```
mrks: How's this different from load above?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+  return memory_usage;
+
+#endif
+
+  Fail("Method not implemented for this platform");
+}
+
+#if defined __linux__
+int64_t MetaSystemUtilizationTable::_int_from_string(std::string input_string) {
```
mrks: Why is this necessary and why is it only implemented for Linux?

j-tr: this is for parsing lines in the /proc/self/status file
under macOS, we can get the values more easily form a system call

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
j-tr: what load do you mean? this function gets the overall memory usage whereas the get_process_memory_usage gets the process specific ram usage

j-tr:
there are not that many platform independent libraries out there. https://github.com/hyperic/sigar seems the most suitable, but since it seems abandonned i thought implementing it on our own would be the better choice

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
mrks: Let's wait for the comment. You might want to reconsider the naming, too. Looks like it does more than just convert a string into an int.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
mrks: Got it

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,353 @@
...
+  mach_timebase_info(&info);
+
+  uint64_t clock_time = mach_absolute_time();
+
+  struct rusage resource_usage;
+  if (getrusage(RUSAGE_SELF, &resource_usage)) {
+    Fail("Unable to access rusage");
+  }
+
+  uint32_t nano = 1'000'000'000, micro = 1'000;
```
Bensk1: These should be constants which should also be indicated by writing their name in all caps. Even better, we should be able to use these, shouldn't we? https://de.cppreference.com/w/cpp/numeric/ratio/ratio

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,353 @@
...
+          static_cast<float>(load_avg.ldavg[1]) / static_cast<float>(load_avg.fscale),
+          static_cast<float>(load_avg.ldavg[2]) / static_cast<float>(load_avg.fscale)};
+
+#endif
+
+  Fail("Method not implemented for this platform");
+}
+
+int MetaSystemUtilizationTable::_get_cpu_count() {
+#if defined __linux__
```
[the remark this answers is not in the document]
j-tr: on the master, we have 24 to 7 in favor for #ifdef. should we unifying everything?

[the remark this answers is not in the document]
mrks: Didn't realize the other option was already wide spread. In the end it's probably not a readability issue, just something I noticed. Your call, feel free to resolve.

src/lib/utils/meta_table_manager.cpp
```diff
@@ -7,23 +7,32 @@
...
+                                                                       std::make_shared<MetaSystemUtilizationTable>()};
 
   _table_names.reserve(_meta_tables.size());
   for (const auto& table : meta_tables) {
     _meta_tables[table->name()] = table;
     _table_names.emplace_back(table->name());
   }
   std::sort(_table_names.begin(), _table_names.end());
+
+  std::static_pointer_cast<MetaSystemUtilizationTable>(_meta_tables.at("system_utilization"))->init();
```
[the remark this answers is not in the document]
mrks: Where has the first part (documentation of init) been resolved?

mrks:
Please let me know when it makes sense to have another look at this :)

src/lib/utils/meta_tables/meta_system_information_table.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include "meta_system_information_table.hpp"
+
+#include "hyrise.hpp"
+
+namespace opossum {
+
+MetaSystemInformationTable::MetaSystemInformationTable()
+    : AbstractMetaSystemTable(TableColumnDefinitions{{"cpu_count", DataType::Int, false},
+                                                     {"ram", DataType::Long, false},
```
Bensk1: I would call that `system_memory_total_bytes`, also could we add the clock speed?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+namespace opossum {
+
+MetaSystemUtilizationTable::MetaSystemUtilizationTable()
+    : AbstractMetaSystemTable(TableColumnDefinitions{{"cpu_system_usage", DataType::Float, false},
+                                                     {"cpu_process_usage", DataType::Float, false},
+                                                     {"load_average_1_min", DataType::Float, false},
+                                                     {"load_average_5_min", DataType::Float, false},
+                                                     {"load_average_15_min", DataType::Float, false},
+                                                     {"system_memory_total", DataType::Long, false},
```
Bensk1: As discussed, we don't need this in both tables.

src/lib/utils/meta_tables/meta_system_information_table.cpp
mrks: @Bensk1 Remember that this might be different on a per-core basis and might change between calls. Sure you need this?

Bensk1: I was talking about the main memory capacity. Why should that change?

mrks: Isn't the base frequency misleading in the presence of (1) frequency scaling / CPU governors, (2) turbo boost, (3) ZMM (AVX-512) slowdown?

Bensk1: I imagined using the base frequency. I haven't seen anybody reporting the ZMM slowdown in her paper yet so it should be fine to leave that out. I could also live with the CPU model name, e.g., `Intel(R) Xeon(R) Platinum 8180 CPU @ 2.50GHz`.

mrks: I'd be more comfortable with the model name, yes

j-tr:
i think someone could have a second look on this @mrks @Bensk1.
i especially don't like that i had to remove the constness of the on_generate function, since the cpu measurement relies on a state. is there a way to keep the constness in the abstract meta table and all the other meta tables but remove it in one subclass? My only idea would be to make the state static.

Bensk1:
I will have a look over the next couple of days

mrks:
> i especially don't like that i had to remove the constness of the on_generate function, since the cpu measurement relies on a state.

Could you give me a pointer to the state, please?

> is there a way to keep the constness in the abstract meta table and all the other meta tables but remove it in one subclass?

mutable? Not really a nice thing though, so let's first have a look at whether it's necessary.

j-tr:
the "state" consits of the last cpu time measurement defined here...
https://github.com/hyrise/hyrise/pull/2044/files#diff-2a6e46de252dfdc972db5d1085c437a2R43
https://github.com/hyrise/hyrise/pull/2044/files#diff-2a6e46de252dfdc972db5d1085c437a2R49

it is set in get_system_cpu_usage https://github.com/hyrise/hyrise/pull/2044/files#diff-17c87cc5779f41598afa7c46c590dba2R124 and get_process_cpu_usage https://github.com/hyrise/hyrise/pull/2044/files#diff-17c87cc5779f41598afa7c46c590dba2R170

this is also why we need the init method.

mrks:
Have you considered not returning the CPU time spent since the last retrieval but the overall time? If the consumer wants to show the CPU time spent in a certain time frame, it needs to do the subtraction itself. This is similar to what we do with the segment access counters. Not only would that solve the const question, but it would also fix the isolation issue that we have right now. The result of me querying the meta table should not be dependent on how long ago I (or even worse, someone else) queried the table the last time.

Bensk1:
> Have you considered not returning the CPU time spent since the last retrieval but the overall time? If the consumer wants to show the CPU time spent in a certain time frame, it needs to do the subtraction itself. This is similar to what we do with the segment access counters. Not only would that solve the const question, but it would also fix the isolation issue that we have right now. The result of me querying the meta table should not be dependent on how long ago I (or even worse, someone else) queried the table the last time.

The current implementation is my fault. I thought it would be really unhandy for the BP to calculate the CPU utilization by themselves. But I see that this creates other problems. We should probably go for the "stateless" version.

Bensk1:
If I understood @j-tr correctly, he will have time to work on this after Tuesday. I will have a look afterwards.

j-tr:
@Bensk1 @mrks feedback highly appreciated

j-tr:
also some thoughts on how to test this:

since we don't know absolute values of how much memory is consumed by a table or how much load an operation produces, i would propose to only test if these values change in a plausible range if something happens. E.g. create a decent amount of dummy data and compare the memory size before and after or launch a bunch of parallelizable tasks/jobs and check the cpu load before, during and after the processing. 

in order to test the overall cpu load and ram usage (not only the load produced by the hyrise processes), we would need an external (python) script that controlls a dummy load and allocates memory and checks if the values change accordingly.

Iidon't see how we can test the SystemInformationTable without machine specific test variables but maybe a machine dependent configuration file could be produced at installation since the hardware specs can be easily obtained by a bash script.

for the numa cpu bindings, we could either manipulate them from within the test with numa_bind() and check if the number of cpus changes or have again an external (python) script that launches tests with different numactl options.

any preferences or additional thoughts?

Bensk1:
You are right, I believe completely testing that would probably be a huge effort.

> E.g. create a decent amount of dummy data and compare the memory size before and after or launch a bunch of parallelizable tasks/jobs and check the cpu load before, during and after the processing.

Sounds reasonable to me.

> in order to test the overall cpu load and ram usage (not only the load produced by the hyrise processes), we would need an external (python) script that controlls a dummy load and allocates memory and checks if the values change accordingly.

Couldn't we just test with above's approach that the overall usage values also increase?

I don't see a good way to test the SystemInformationTable. For NUMA, I could imagine to change the NUMA configuration from within the tests as you suggested.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,337 @@
...
+  Fail("Method not implemented for this platform");
+}
+
+/*
+  * Returns the time in ns since an arbitrary point in the past.
+*/
+int64_t MetaSystemUtilizationTable::_get_total_time() {
+/* CLOCK_MONOTONIC_RAW:
+ * Similar to CLOCK_MONOTONIC, but provides access to a raw hard‐
+ * ware-based time that is not subject to NTP adjustments or the
```
Bensk1, line 96: Is this also the reason why we cannot use C++'s chrono here or is the reason to be more compatible with the system's cpu_time functions?

src/lib/utils/meta_tables/meta_system_utilization_table.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+  struct ProcessMemoryUsage {
+    int64_t virtual_memory;
+    int64_t physical_memory;
+  };
+
+  static LoadAvg _get_load_avg();
+  static int64_t _get_total_time();
+  static int64_t _get_system_cpu_time();
+  static int64_t _get_process_cpu_time();
```
Bensk1, line 38: Hm, these are probably of type `int64_t` because that is the type we need for appending to the table, right? Otherwise, these should not be signed ints, a comment could clarify this.

src/lib/utils/meta_tables/meta_system_information_table.hpp
```diff
@@ -0,0 +1,22 @@
...
+class MetaSystemInformationTable : public AbstractMetaSystemTable {
+ public:
+  MetaSystemInformationTable();
+
+  const std::string& name() const final;
+
+ protected:
+  std::shared_ptr<Table> _on_generate() const final;
+
+  static pmr_string _cpu_model();
```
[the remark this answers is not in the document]
j-tr: since String values in a table must be pmr_strings, i choose to directly return a pmr_string here.

src/lib/utils/meta_tables/meta_system_information_table.cpp
```diff
@@ -0,0 +1,91 @@
...
+  const auto cpus = _get_cpu_count();
+
+  auto numa_cpus = cpus;
+#ifdef __linux__
+  if (numa_available() != -1) {
+    numa_cpus = numa_num_task_cpus();
+  }
+#endif
+
+  int64_t ram;
```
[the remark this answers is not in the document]
j-tr: this is also due to the fact that Long values in a table must be signed. otherwise we would need another conversion before placing the value in the table.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+MetaSystemUtilizationTable::MetaSystemUtilizationTable()
+    : AbstractMetaSystemTable(TableColumnDefinitions{{"cpu_system_time", DataType::Long, false},
+                                                     {"cpu_process_time", DataType::Long, false},
+                                                     {"total_time", DataType::Long, false},
+                                                     {"load_average_1_min", DataType::Float, false},
+                                                     {"load_average_5_min", DataType::Float, false},
+                                                     {"load_average_15_min", DataType::Float, false},
+                                                     {"system_memory_free", DataType::Long, false},
+                                                     {"system_memory_available", DataType::Long, false},
```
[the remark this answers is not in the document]
j-tr: yes, not all memory can be paged out. so system_memory_total_bytes is the size of the physically available ram for the whole system whereas system_memory_available is all the memory that can be made available by paging out unnecessary pages. but it should be always smaller than the total memory

src/lib/utils/meta_tables/meta_system_information_table.hpp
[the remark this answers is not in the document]
mrks: No real guidelines. Usually, "database data" is stored as a pmr_string, "DBMS data" is an std::string. As we can see here, there are grey areas.

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
```diff
@@ -0,0 +1,55 @@
...
+  uint32_t processors = 0;
+  std::string cpu_info_line;
+  while (std::getline(cpu_info_file, cpu_info_line)) {
+    if (cpu_info_line.rfind("processor", 0) == 0) ++processors;
+  }
+
+  cpu_info_file.close();
+
+  return processors;
+#endif
```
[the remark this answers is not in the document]
mrks: Which also works on Mac. Not sure if by default or because of some package though.

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
```diff
@@ -0,0 +1,55 @@
+#ifdef __linux__
+
+#include <fstream>
+
+#endif
```
mrks, line 5: I'd include fstream in any case. Better to keep the divergence as small as possible.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,337 @@
...
+#include <numa.h>
+#include <fstream>
+
+#endif
+
+#ifdef __APPLE__
+
+#include <mach/mach.h>
+
+#endif
```
mrks, line 12: See above. Also, we could reduce the number of empty lines here; they don't really improve the readability.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,337 @@
...
+  * Returns the load average values for 1min, 5min, and 15min.
+*/
+MetaSystemUtilizationTable::LoadAvg MetaSystemUtilizationTable::_get_load_avg() {
+#ifdef __linux__
+  std::ifstream load_avg_file;
+  load_avg_file.open("/proc/loadavg", std::ifstream::in);
+
+  if (!load_avg_file.is_open()) {
+    Fail("Unable to open /proc/loadavg");
+  }
```
mrks, line 62: Assert? Maybe grep the PR for `Fail(`

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+  std::string load_avg_value;
+  std::vector<float> load_avg_values;
+  for (int value_index = 0; value_index < 3; ++value_index) {
```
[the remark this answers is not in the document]
mrks: Oooor use an array if the number is compile-time

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,337 @@
...
+  Fail("Method not implemented for this platform");
+}
+
+/*
+ * Returns the time in ns that ALL processes have spend on the CPU
+ * since an arbitrary point in the past.
+*/
+int64_t MetaSystemUtilizationTable::_get_system_cpu_time() {
+// The amount of time is measured in units of clock ticks.
+// sysconf(_SC_CLK_TCK) can be used to obtain the right value.
```
mrks, line 129: What is the purpose of this comment? Above you say the time is in `ns`. Do you mean the resolution here?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,337 @@
...
+  const auto active_ns = active_ticks * std::nano::den / (sysconf(_SC_CLK_TCK) * _get_cpu_count());
+
+  return active_ns;
+#endif
+
+  Fail("Method not implemented for this platform");
+}
+
+/*
+ * Returns the time in ns that THIS process has spend on the CPU
```
mrks, line 182: spent

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  return active_ns / _get_cpu_count();
+#endif
+
+  Fail("Method not implemented for this platform");
+}
+
+/*
+ * Returns a struct that contains the avaiable and free memory size in bytes.
```
mrks, line 222: available

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,337 @@
...
+ * - Free memory is unallocated memory.
+ * - Availlable memory includes free memory and currently allocated memory that
```
mrks, line 224: available

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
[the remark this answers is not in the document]
Bensk1: Alright, could you add a comment for that if we don't have it yet?

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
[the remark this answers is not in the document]
j-tr: it's part of coreutils. 
however some other key figures could also be obtained by calling an external command but i thought it would be cleaner to not rely on external tools. what's the general opinion here? avload, numactl or free could make a lot of proc file parsing obsolete.

mrks:
Slightly unrelated to this PR, feel free to spin off into its own issue if it isn't done in ten lines.

Quick thought that I had when looking at #2104 and thinking back at how the meta table were generated many times. Should we add a test with a fake, instrumented meta table that tracks how often it is generated, query it, and verify that it was generated only once?

mrks:
Nevermind - I found that #2049 was a better fit for my comment.

j-tr:
> Slightly unrelated to this PR, feel free to spin off into its own issue if it isn't done in ten lines.
> 
> Quick thought that I had when looking at #2104 and thinking back at how the meta table were generated many times. Should we add a test with a fake, instrumented meta table that tracks how often it is generated, query it, and verify that it was generated only once?

How would that be different from SQLTranslatorTest.SelectMetaTableMultipleAccess?

mrks:
The new test may cover a bit more as it also tracks if the meta table was generated, e.g., as part of an optimizer rule.

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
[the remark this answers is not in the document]
Bensk1: Having a widely available solution might not be too bad.

mrks:
Can't find the comment on mobile: when thinking about nproc, we should make sure not to introduce new dependencies.  numactl, for example, is not used on Mac.

mrks:
Also, given how specialized and isolated this is, I personally wouldn't spend more than an hour or so trying to make this more pretty.

Bensk1:
> Can't find the comment on mobile: when thinking about nproc, we should make sure not to introduce new dependencies. numactl, for example, is not used on Mac.

I agree, my intention was to argue for the current solution, not for `nproc`. Apparently, that did not work out too well.

src/test/utils/meta_tables/meta_system_utilization_test.cpp
```diff
@@ -0,0 +1,181 @@
...
+  const auto system_cpu_load_utilization =
+      static_cast<float>(cpu_system_time_load_diff) / static_cast<float>(total_time_load_diff);
+  const auto process_cpu_load_utilization =
+      static_cast<float>(cpu_process_time_load_diff) / static_cast<float>(total_time_load_diff);
+
+  EXPECT_GE(system_cpu_load_utilization, system_cpu_idle_utilization);
+  EXPECT_GE(process_cpu_load_utilization, process_cpu_idle_utilization);
+
+  // EXPECT_GE(system_cpu_load_utilization, process_cpu_load_utilization);
+  // EXPECT_GE(system_cpu_idle_utilization, process_cpu_idle_utilization);
```
Bensk1: ?

j-tr:
as it can be seen in https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-2044/18/pipeline, some tests may fail if other tasks are executed in parallel. We recently discussed that we could exclude the tests in the Jenkins file but then, they could still fail locally under certain conditions. i would rather vote for disabling the tests to prevent confusion if they fail locally. however, we require disabled tests to have a related issue and i don't really like the idea of having such won't-fix-issues. what's your opinion on that? @Bensk1 @mrks

src/test/utils/meta_tables/meta_system_utilization_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+    return table->_generate();
+  }
+
+  static const std::shared_ptr<Table> create_dummy_table() {
+    const auto dummy_table =
+        std::make_shared<Table>(TableColumnDefinitions{{"dummy_column", DataType::Long, false}}, TableType::Data, 5);
+    // Creates some dummy data
+    const auto dummy_table_columns = 1024 * 1024 / 64;
+    for (uint64_t column = 0; column < dummy_table_columns; ++column) {
+      dummy_table->append({static_cast<int64_t>(column)});
```
mrks, line 49: value, not column, isn't it?

mrks:
The allocator might still have memory available, so not every allocation needs to increase the RSS. In any case, I would sleep fine at night if those two are untested.

Bensk1:
Would it be suitable to link to this PR and the discussions in here? If this discussion point and @mrks' last comment is resolved, it would be ready for merge from my side.

Bensk1:
I spoke with @mrks, let's remove these tests.

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "utils/meta_tables/abstract_meta_system_table.hpp"
+
+namespace opossum {
+
+AbstractMetaSystemTable::AbstractMetaSystemTable(const TableColumnDefinitions& column_definitions)
+    : AbstractMetaTable(column_definitions) {}
+
+// Returns the number of logical processors
+size_t AbstractMetaSystemTable::_get_cpu_count() {
```
mrks, line 15: Too little to gain by changing this now, so just as a random thought: The two AbstractMetaSystemTables don't have any observable behavior in common. Building a class hierarchy just to access `_get_cpu_count` feels like overkill, if the method could also have been a helper method somewhere.

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
```diff
@@ -0,0 +1,46 @@
...
+#ifdef __linux__
+
```
mrks, line 17: Empty line here is weird. We don't have empty lines after `if` either

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
```diff
@@ -0,0 +1,46 @@
...
+  std::ifstream cpu_info_file;
+  cpu_info_file.open("/proc/cpuinfo", std::ifstream::in);
+  DebugAssert(cpu_info_file.is_open(), "Failed to open /proc/cpuinfo");
```
mrks, line 20: Just FYI, no need to change: Could also be done by https://en.cppreference.com/w/cpp/io/basic_ios/exceptions

src/lib/utils/meta_tables/meta_system_information_table.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+std::shared_ptr<Table> MetaSystemInformationTable::_on_generate() const {
+  auto output_table = std::make_shared<Table>(_column_definitions, TableType::Data, std::nullopt, UseMvcc::Yes);
+
+  const auto cpus = _get_cpu_count();
+
+  auto numa_cpus = cpus;
+#ifdef __linux__
+  if (numa_available() != -1) {
+    auto* cpu_mask = numa_allocate_cpumask();
```
mrks, line 31: I believe libnuma is optional. Either check HYRISE_NUMA_SUPPORT or make it required (and update CMake accordingly).

src/lib/utils/meta_tables/meta_system_information_table.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include <fstream>
+
+#include "meta_system_information_table.hpp"
+
+namespace opossum {
+
+MetaSystemInformationTable::MetaSystemInformationTable()
+    : AbstractMetaSystemTable(TableColumnDefinitions{{"cpu_count", DataType::Int, false},
+                                                     {"system_memory_total_bytes", DataType::Long, false},
+                                                     {"numa_cpu_count", DataType::Int, false},
```
mrks, line 15: Had to look into the code to find out what this means. This doesn't seem to be related to NUMA. Maybe `cpu_affinity_count`?

src/lib/utils/meta_tables/meta_system_information_table.cpp
mrks: Looking at this again, CPU affinities can also be present without libnuma installed. https://linux.die.net/man/2/sched_getaffinity should work just as well.

Did you check if OS X has scheduling affinities? If it doesn't, we should doc this behavior here.

src/lib/utils/meta_tables/meta_system_information_table.cpp
```diff
@@ -0,0 +1,89 @@
...
+    numa_cpus = numa_bitmask_weight(cpu_mask);
+    numa_free_cpumask(cpu_mask);
+  }
+#endif
+
+  uint64_t ram;
+#ifdef __linux__
+  struct sysinfo memory_info {};
+  [[maybe_unused]] const auto ret = sysinfo(&memory_info);
+  DebugAssert(ret == 0, "Failed to get sysinfo");
```
mrks, line 42: Could also be a regular Assert, which would save us the maybe_unused. It's not like this is executed over and over again. I'd probably change all DebugAsserts here to Asserts

src/lib/utils/meta_tables/meta_system_information_table.cpp
```diff
@@ -0,0 +1,89 @@
...
+#ifdef __APPLE__
+  size_t buffer_size = 256;
+  char buffer[256];
+  const auto ret = sysctlbyname("machdep.cpu.brand_string", &buffer, &buffer_size, nullptr, 0);
+  DebugAssert(ret == 0, "Failed to call sysctl machdep.cpu.brand_string");
+
+  return std::string(buffer);
+#endif
+}
+
```
mrks, line 88: Add Fail()?

src/lib/utils/meta_tables/meta_system_information_table.cpp
```diff
@@ -0,0 +1,89 @@
...
+// Returns the CPU model string
+std::string MetaSystemInformationTable::_cpu_model() {
+#ifdef __linux__
+  std::ifstream cpuinfo_file;
+  cpuinfo_file.open("/proc/cpuinfo", std::ifstream::in);
+  DebugAssert(cpuinfo_file.is_open(), "Failed to open /proc/cpuinfo");
+
+  std::string cpuinfo_line;
+  while (std::getline(cpuinfo_file, cpuinfo_line)) {
+    if (cpuinfo_line.rfind("model name", 0) == 0) {
```
mrks, line 70: Why rfind?

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+  size_t processors = 0;
+  std::string cpu_info_line;
+  while (std::getline(cpu_info_file, cpu_info_line)) {
+    if (cpu_info_line.rfind("processor", 0) == 0) ++processors;
```
mrks, line 25: Minor: Not sure if both of our used libraries already support `string.starts_with`, but that would be slightly easier and more efficient.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,270 @@
...
+    : AbstractMetaSystemTable(TableColumnDefinitions{{"cpu_system_time", DataType::Long, false},
+                                                     {"cpu_process_time", DataType::Long, false},
+                                                     {"total_time", DataType::Long, false},
+                                                     {"load_average_1_min", DataType::Float, false},
+                                                     {"load_average_5_min", DataType::Float, false},
+                                                     {"load_average_15_min", DataType::Float, false},
+                                                     {"system_memory_free", DataType::Long, false},
+                                                     {"system_memory_available", DataType::Long, false},
+                                                     {"process_virtual_memory", DataType::Long, false},
+                                                     {"process_physical_memory_RSS", DataType::Long, false}}) {}
```
mrks, line 26: What does "physical memory" add here? Wouldn't `process_rss` say the same thing?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,270 @@
...
+uint64_t MetaSystemUtilizationTable::_get_total_time() {
+  auto time = std::chrono::steady_clock::now().time_since_epoch();
+  return std::chrono::duration_cast<std::chrono::nanoseconds>(time).count();
+}
+
+/*
+ * Returns the time in ns that ALL processes have spent on the CPU
+ * since an arbitrary point in the past.
+*/
+uint64_t MetaSystemUtilizationTable::_get_system_cpu_time() {
```
mrks, line 75: @Bensk1 What do you use this information for?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,270 @@
...
+  Fail("Method not implemented for this platform");
+}
+
+/*
+ * Returns the time in ns that THIS process has spent on the CPU
+ * since an arbitrary point in the past.
+*/
+uint64_t MetaSystemUtilizationTable::_get_process_cpu_time() {
+  // CLOCK_PROCESS_CPUTIME_ID:
+  // Per-process CPU-time clock (measures CPU time consumed by all threads in the process).
```
mrks, line 132: Not sure what this metric is supposed to mean. How does it differ from user+sys from /proc/self/stat?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,270 @@
...
+  DebugAssert(ret == 0, "Failed in clock_gettime");
+
+  size_t cpu_count;
+  if (numa_available() != -1) {
+    cpu_count = numa_num_task_cpus();
+  } else {
+    cpu_count = _get_cpu_count();
+  }
+
+  const auto active_ns = (time_spec.tv_sec * std::nano::den + time_spec.tv_nsec) / cpu_count;
```
mrks, line 145: Why do we divide by cpu_count here?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+#ifdef __APPLE__
+  const auto active_ns = clock_gettime_nsec_np(CLOCK_PROCESS_CPUTIME_ID);
+  DebugAssert(active_ns != 0, "Failed in clock_gettime_nsec_np");
+
+  return active_ns / _get_cpu_count();
+#endif
+
+  Fail("Method not implemented for this platform");
+}
```
mrks, line 158: Depending on the outcome of the previous discussion, this might be more accurate:

```
struct task_basic_info {
        integer_t       suspend_count;  /* suspend count for task */
        vm_size_t       virtual_size;   /* number of virtual pages */
        vm_size_t       resident_size;  /* number of resident pages */
        time_value_t    user_time;      /* total user run time for
                                           terminated threads */
        time_value_t    system_time;    /* total system run time for
                                           terminated threads */
	policy_t	policy;		/* default policy for new threads */
};
```

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,270 @@
...
+  [[maybe_unused]] auto ret = sysctlbyname("hw.memsize", &physical_memory, &size, nullptr, 0);
+  DebugAssert(ret == 0, "Failed to call sysctl hw.memsize");
+
+  vm_size_t page_size;
+  vm_statistics64_data_t vm_statistics;
+  mach_msg_type_number_t count = sizeof(vm_statistics) / sizeof(natural_t);
+  ret = host_page_size(mach_host_self(), &page_size);
+  DebugAssert(ret == KERN_SUCCESS, "Failed to get page size");
+  ret = host_statistics64(mach_host_self(), HOST_VM_INFO, reinterpret_cast<host_info64_t>(&vm_statistics), &count);
+  DebugAssert(ret == KERN_SUCCESS, "Failed to get host_statistics64");
```
mrks, line 200: Did you get this from somewhere or do you have any resources that you could link?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+#ifdef __APPLE__
+  struct task_basic_info info;
+  mach_msg_type_number_t count = TASK_BASIC_INFO_COUNT;
+  [[maybe_unused]] const auto ret =
+      task_info(mach_task_self(), TASK_BASIC_INFO, reinterpret_cast<task_info_t>(&info), &count);
+  DebugAssert(ret == KERN_SUCCESS, "Failed to get task_info");
+
+  return {info.virtual_size, info.resident_size};
+#endif
```
mrks, line 246: For once, mac's API is less painful :D

src/test/utils/meta_tables/meta_system_information_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+  const uint32_t max_cpu = static_cast<uint32_t>(numa_num_configured_cpus());
+  for (uint32_t cpu_index = 0; cpu_index < max_cpu; ++cpu_index) {
+    if (numa_bitmask_isbitset(default_cpu_mask, cpu_index)) {
+      cpu_mask = numa_bitmask_clearbit(cpu_mask, cpu_index);
+      break;
+    }
+  }
+
+  numa_sched_setaffinity(0, cpu_mask);
```
mrks, line 48: As above, this might not build on a system without libnuma.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
Bensk1: The idea was to be able to differentiate between CPU time consumed by Hyrise and by other processes on the same machine. I am not sure which of the metrics we currently display in the BP's frontend. @Bouncner do you know that?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
Bensk1: That was a handy way to get the time spent on an average CPU, but you are right, the user of the meta tables should probably do this calculation and we should return the raw values.

mrks: In order to divide it by the number of CPUs, it would have to be an aggregate value, wouldn't it? If I am not mistaken, this is simply RDTSC, i.e., the time spent since a random point in time. How does that relate to the number of CPUs?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
mrks: Sounds reasonable to have that. I'd add it as a comment.

src/lib/utils/meta_tables/meta_system_information_table.cpp
j-tr: because it allows to check if the search string is at the beginning by simply checking if the return value is 0. (got it from here https://stackoverflow.com/a/40441240)

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
j-tr: neat

src/lib/utils/meta_tables/meta_system_information_table.cpp
mrks: Interesting, didn't know about that one. :)

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
mrks: Forget about the performance thing though. I didn't know about the rfind trick. This is just syntactic sugar then.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
j-tr: true. this is a leftover from the previous implementation where we measured multiple points in time.

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
j-tr: where would we preferably place such a helper method?

src/lib/utils/meta_tables/meta_system_information_table.cpp
j-tr: rather 
```
const auto ret = whatever();
Assert(ret);
```
or 
```
Assert(whatever());
```
?

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
mrks: Wait a second. How about simply using `Topology::num_cpus`?

src/lib/utils/meta_tables/meta_system_information_table.cpp
mrks: The former. First, because it is more consistent with the rest of the codebase, second because it doesn't break if someone decides to make it a DebugAssert again.

mrks: Generally, assertions should always be completely removable without anything breaking.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
mrks: "If you google around for the patches that went into the kernel to support this (late 2004, by Christoph Lameter), you'll see that it sums the user and system time."

https://arstechnica.com/civis/viewtopic.php?t=1091018

Ok. Maybe add a comment explaining that this is user+sys as known from `time`?

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
mrks: Correction: num_cpus returns the number of used CPUs. Still, I'd consider exposing numa_num_configured_cpus through the Topology class.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
j-tr: https://stackoverflow.com/a/1911863
should that be documented in a code comment as well?

mrks: Doesn't hurt.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
mrks: Guess this is resolved then.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
mrks: The timer_create documentation is better:

>        CLOCK_PROCESS_CPUTIME_ID (since Linux 2.6.12)
              A clock that measures (user and system) CPU time consumed by
              (all of the threads in) the calling process.

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
j-tr: how come that noone ever mentioned `Topology` until now? 
I think we could drop the whole AbstractMetaSystemTable in favor of Topology::num_cpus.
This number is only needed to normalize the runtimes, so we actually want to know the numbers of used cpus and not potentially useable cpus.

mrks: I don't think that `Topology` was touched in the recent past. Only the tales of the elders of Hyrise remind of this mystical class and those elders are sometimes doddery :/

Bensk1: Sorry, haven't used the Topology in years.

mrks: Another thought: The current approach might be a problem if the scheduler is used. As the scheduler spawns worker threads with a certain CPU affinity, `_get_cpu_count` might return the affinity of the worker, not that of the entire process.

Not sure if there is a nice way to automatically test this, but we should at least have a look if the values make sense.

Bensk1:
Is everything addressed now?

j-tr:
there's still the issue about the abstract class and its _get_cpu_count method and the worker issue. 
for the former, i would vote for moving the cpu_affinity_count to the SystemUtilizationTable, integrate get_cpu_count in the SystemInformationTable and drop the abstract table. This way, we would only have static information in the SystemInformationTable and dynamic information in the SystemUtilizationTable. 
For the latter, i still need a little research to find out if thread cpu restictions somehow alter process restrictions. since the workers are restraining their cpu set by pthread_setaffinity_np, this should be only thread related and not on a process level.

Bensk1:
Removing the abstract table and moving the `cpu_affinity_count` sounds reasonable.

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
j-tr: this seems to be a real problem. to solve this, i now return topology.num_cpus as cpu_affinity_count. this is true since the number of cpus in the topology is set according to the available cpus. however, if we change the cpu affinities without reinitializing the topology, this won't reflect in the metatable.

src/lib/utils/meta_tables/meta_system_information_table.cpp
j-tr: this is now solved kind of indirect. Topology cares about HYRISE_NUMA_SUPPORT and "fakes" the numa topology if required.

src/lib/utils/meta_tables/abstract_meta_system_table.cpp
mrks: > however, if we change the cpu affinities without reinitializing the topology, this won't reflect in the metatable.

Shouldn't do that anyway ;)

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -71,16 +71,21 @@ uint64_t MetaSystemUtilizationTable::_get_total_time() {
 /*
  * Returns the time in ns that ALL processes have spent on the CPU
  * since an arbitrary point in the past.
+ * This might be used to differentiate between CPU time consumed by this process 
+ * and by other processes on the same machine.
```
mrks, line 75: Could you please check your editor's line wrapping settings? We only break after 120 characters or to add structure to the text.

PR #2045, opened 2020-03-09
Add missing forward declaration

Bouncner:
Same as with #2042. If people continue to disable PCH (I know I am not the only one), we might want to automatically check for building without it.


Bensk1:
I can imagine that this is really annoying

mrks:
Feel free to add a non PCH build (similar to the ODR one)

Bouncner:
I’ve never touched this Jenkins mess until now. Anybody here who wants?

Bensk1:
I can do it later

Bouncner:
Interestingly, @Boffmann observe the same issue even without disabling PCH.

mrks:
Have a try if `-DCMAKE_DISABLE_PRECOMPILE_HEADERS=On` is sufficient: https://cmake.org/cmake/help/git-stage/variable/CMAKE_DISABLE_PRECOMPILE_HEADERS.html#variable:CMAKE_DISABLE_PRECOMPILE_HEADERS

Bouncner:
`-DCMAKE_DISABLE_PRECOMPILE_HEADERS=On` is sufficient to trigger the `Table not defined` issue. So that would apparently work.

@Bensk1 : would you want that in another PR or in here?

PR #2046, opened 2020-03-11
Make overriden meta table methods final

Bensk1:
I added the `final` keyword to the meta table files and corrected a typo.

### DIFF
--- a/src/lib/utils/settings/abstract_setting.hpp
+++ b/src/lib/utils/settings/abstract_setting.hpp
@@ -22,7 +22,7 @@ namespace opossum {
  */
 class AbstractSetting : public Noncopyable, public std::enable_shared_from_this<AbstractSetting> {
  public:
-  explicit AbstractSetting(const std::string& init_name);
+  explicit AbstractSetting(const std::string& init_name, const std::string& init_value);
 
   virtual ~AbstractSetting() = default;
 
@@ -37,6 +37,8 @@ class AbstractSetting : public Noncopyable, public std::enable_shared_from_this<
   virtual void unregister_at_settings_manager();
 
   const std::string name;
+ protected:
+  std::string _value;
 };
 
 }  // namespace opossum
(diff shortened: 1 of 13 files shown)


src/lib/utils/settings/abstract_setting.hpp
```diff
@@ -37,6 +37,8 @@ class AbstractSetting : public Noncopyable, public std::enable_shared_from_this<
   virtual void unregister_at_settings_manager();
 
   const std::string name;
+ protected:
+  std::string _value;
```
mrks: To me, the entire idea of this settings class was that settings are *not* stored as a "setting" somewhere but are dynamically retrieved whenever they are needed. Why would we need / want to store the value here?

Bensk1: Apparently, I had a totally different idea then. How would that look like for an example? I added a memory budget setting to a compression example plugin: https://github.com/hyrise/hyrise/blob/2ab62566624c6cce15a773c92cdc0bc9becd9d60/src/plugins/compression_plugin.hpp

mrks: My understanding was that the setting's `set` method would convert it to an int and change a member of the plugin itself. Similarly, a setting `SchedulerEnabledSetting` would enable the scheduler if a truthy value is passed and disable it for a falsy value. This would be hard to accomplish if everything the setting did was to store it locally.

Bensk1: Okay, but then the Plugins would need public members or friend all settings, right?

mrks: If by "all" you mean "their", then yes.

Bensk1: Ok.

Bensk1:
Do we want to merge these small improvements?

PR #2048, opened 2020-03-13
Small changes to segment encoding

Bouncner:
This PR mainly introduces a function to the encoding segment utils that allows to receive the segment encoding spec. This function can be used to check if an encoding is really required or if the encoding is already in place (needed for current work on a plugin).

The segment encoding spec has also changed a little bit to make working with it simpler.

### DIFF
--- a/src/lib/storage/chunk_encoder.cpp
+++ b/src/lib/storage/chunk_encoder.cpp
@@ -9,6 +9,7 @@
 #include "resolve_type.hpp"
 #include "statistics/generate_pruning_statistics.hpp"
 #include "storage/base_encoded_segment.hpp"
+#include "storage/base_segment_encoder.hpp"
 #include "storage/reference_segment.hpp"
 #include "storage/segment_encoding_utils.hpp"
 #include "storage/segment_iterables/any_segment_iterable.hpp"
@@ -73,7 +74,12 @@ std::shared_ptr<BaseSegment> ChunkEncoder::encode_segment(const std::shared_ptr<
       });
       result = std::make_shared<ValueSegment<ColumnDataType>>(std::move(values), std::move(null_values));
     } else {
-      result = encode_and_compress_segment(segment, data_type, encoding_spec);
+      auto encoder = create_encoder(encoding_spec.encoding_type);
+      if (encoding_spec.vector_compression_type) {
+        encoder->set_vector_compression(*encoding_spec.vector_compression_type);
+      }
+
+      result = encoder->encode(segment, data_type);
     }
   });
   return result;
--- a/src/lib/storage/chunk_encoder.cpp
+++ b/src/lib/storage/chunk_encoder.cpp
@@ -27,19 +27,17 @@ namespace opossum {
 std::shared_ptr<BaseSegment> ChunkEncoder::encode_segment(const std::shared_ptr<BaseSegment>& segment,
                                                           const DataType data_type,
                                                           const SegmentEncodingSpec& encoding_spec) {
+  Assert(!std::dynamic_pointer_cast<const ReferenceSegment>(segment), "Reference segments cannot be encoded.");
+
   std::shared_ptr<BaseSegment> result;
   resolve_data_type(data_type, [&](const auto type) {
     using ColumnDataType = typename decltype(type)::type;
-
-    if (const auto reference_segment = std::dynamic_pointer_cast<const ReferenceSegment>(segment)) {
-      Fail("Reference segments cannot be encoded.");
-    }
 
     // TODO(anyone): After #1489, build segment statistics in encode_segment() instead of encode_chunk()
     // and store them within the segment instead of a chunk-owned list of statistics.
 
     // Check if early exit is possible when passed segment is already encoded with requested spec.
-    // If case no vector compression is specified, only the correct encoding type is checked and the current vector
+    // In case no vector compression is specified, only the correct encoding type is checked and the current vector
     // compression type is ignored.
     const auto current_segment_encoding_spec = get_segment_encoding_spec(segment);
     if (current_segment_encoding_spec == encoding_spec ||
--- a/src/lib/storage/chunk_encoder.cpp
+++ b/src/lib/storage/chunk_encoder.cpp
@@ -42,7 +42,9 @@ std::shared_ptr<BaseSegment> ChunkEncoder::encode_segment(const std::shared_ptr<
     // If case no vector compression is specified, only the correct encoding type is checked and the current vector
     // compression type is ignored.
     const auto current_segment_encoding_spec = get_segment_encoding_spec(segment);
-    if (current_segment_encoding_spec == encoding_spec || (!encoding_spec.vector_compression_type && current_segment_encoding_spec.encoding_type == encoding_spec.encoding_type)) {
+    if (current_segment_encoding_spec == encoding_spec ||
+        (!encoding_spec.vector_compression_type &&
+         current_segment_encoding_spec.encoding_type == encoding_spec.encoding_type)) {
       result = segment;
       return;
     }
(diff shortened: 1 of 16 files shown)


src/lib/storage/chunk_encoder.cpp
```diff
@@ -28,39 +28,27 @@ std::shared_ptr<BaseSegment> ChunkEncoder::encode_segment(const std::shared_ptr<
...
     }
 
-    // Check if early exit is possible when passed segment is currently unencoded and the same is requested.
-    const auto unencoded_segment = std::dynamic_pointer_cast<const ValueSegment<ColumnDataType>>(segment);
-    if (unencoded_segment && encoding_spec.encoding_type == EncodingType::Unencoded) {
+    // TODO(anyone): After #1489, build segment statistics in encode_segment() instead of encode_chunk()
+    // and store them within the segment instead of a chunk-owned list of statistics.
+
+    // Check if early exit is possible when passed segment is already encoded with requested spec.
+    // If case no vector compression is specified, only the correct encoding type is checked and the current vector
```
Bensk1: In case?

src/lib/storage/segment_encoding_utils.hpp
```diff
@@ -20,15 +20,27 @@ class BaseValueSegment;
...
+ * SegmentEncodingSpec already in place.
  *
  * @return encoded segment if data type is supported, otherwise throws exception.
  */
 std::shared_ptr<BaseEncodedSegment> encode_and_compress_segment(const std::shared_ptr<const BaseSegment>& segment,
                                                                 const DataType data_type,
                                                                 const SegmentEncodingSpec& encoding_spec);
 
+/**
+ * @return the segment encoding spec for thes given segment.
```
Bensk1: `thes`

src/lib/storage/segment_encoding_utils.hpp
```diff
@@ -20,15 +20,27 @@ class BaseValueSegment;
 std::unique_ptr<BaseSegmentEncoder> create_encoder(EncodingType encoding_type);
 
 /**
- * @brief Encodes a segment by the given encoding specification (i.e., the encoding method
- * and -- when given and applicable -- the vector compression type)
+ * @brief Encodes a segment by the given encoding specification (i.e., the encoding method and -- when given and
+ * applicable -- the vector compression type). In general, do not use this method but rather
+ * ChunkEncoder::encode_segment() which supports reencoding to ValueSegments and avoids reencoding with
```
Bensk1: avoids reencoding if the SegmentEncodingSpec is already in place?

Bensk1: The comment sounds like `ChunkEncoder::encode_segment()` can do more and do it better than this method. What do we need it for?

src/lib/storage/segment_encoding_utils.cpp
```diff
@@ -51,11 +51,36 @@ std::shared_ptr<BaseEncodedSegment> encode_and_compress_segment(const std::share
...
- * For the difference of the two, please take a look at compressed_vector_type.hpp.
- */
+SegmentEncodingSpec get_segment_encoding_spec(const std::shared_ptr<const BaseSegment>& segment) {
+  SegmentEncodingSpec result;
+  resolve_data_type(segment->data_type(), [&](const auto type) {
+    using ColumnDataType = typename decltype(type)::type;
+
+    if (const auto reference_segment = std::dynamic_pointer_cast<const ReferenceSegment>(segment)) {
+      Fail("Reference segments cannot be encoded.");
+    }
```
Bensk1: Should we make that an assert?

src/lib/storage/segment_encoding_utils.cpp
```diff
@@ -51,11 +51,36 @@ std::shared_ptr<BaseEncodedSegment> encode_and_compress_segment(const std::share
...
+
+    const auto unencoded_segment = std::dynamic_pointer_cast<const ValueSegment<ColumnDataType>>(segment);
+    if (unencoded_segment) {
+      result = SegmentEncodingSpec{EncodingType::Unencoded};
+      return;
+    }
+
+    const auto encoded_segment = std::dynamic_pointer_cast<const BaseEncodedSegment>(segment);
+    if (encoded_segment) {
+      std::optional<VectorCompressionType> vector_compression = std::nullopt;
```
Bensk1: I believe we do not need this assignment, do we?

src/lib/storage/segment_encoding_utils.cpp
```diff
@@ -51,11 +51,36 @@ std::shared_ptr<BaseEncodedSegment> encode_and_compress_segment(const std::share
...
+      if (!encoded_segment->compressed_vector_type()) {
+        vector_compression = parent_vector_compression_type(*encoded_segment->compressed_vector_type());
+      }
```
Bensk1: I don't understand this. The condition in 72 should only be true if `encoded_segment->compressed_vector_type()` returns `std::nullopt`, right? If so, what's the point in passing that to `parent_vector_compression_type`?

src/lib/storage/segment_encoding_utils.cpp
Bouncner: Probably better, yes. Will change.

src/lib/storage/segment_encoding_utils.hpp
Bouncner: I left in place as most test use the "simple" method, which provides less convenience but always does exactly what is requested.
I don't see a good reason to keep it, but I wasn't sure we can remove the other function.

Remove it?

src/lib/storage/segment_encoding_utils.cpp
Bouncner: You mean `std::optional<VectorCompressionType> vector_compression;`? If so, yes will change it.

src/lib/storage/segment_encoding_utils.cpp
Bouncner: Sorry, the last changes for the remaining tests haven't been pushed yet. The `!` was totally wrong, yes.

src/lib/storage/segment_encoding_utils.cpp
```diff
@@ -51,11 +51,24 @@ std::shared_ptr<BaseEncodedSegment> encode_and_compress_segment(const std::share
...
+
+  if (const auto encoded_segment = std::dynamic_pointer_cast<const BaseEncodedSegment>(segment)) {
+    std::optional<VectorCompressionType> vector_compression;
+    if (encoded_segment->compressed_vector_type()) {
+      vector_compression = parent_vector_compression_type(*encoded_segment->compressed_vector_type());
+    }
+    return SegmentEncodingSpec{encoded_segment->encoding_type(), vector_compression};
+  }
+
+  Fail("Unexpected segment encoding found.");
```
mrks: Guard clause?

src/lib/storage/segment_encoding_utils.hpp
Bensk1: You are the encoding master, I am not the one to judge on that.

src/lib/storage/segment_encoding_utils.cpp
Bouncner: Woud require to check for a couple of possible dynamic_casts, wouldn't it? I don't have an idea how to create a non-ugly guard here. :-/

mrks: Ah right, that's a chain of ifs. Never mind.

src/lib/storage/encoding_type.hpp
```diff
@@ -63,13 +63,19 @@ bool encoding_supports_data_type(EncodingType encoding_type, DataType data_type)
 struct SegmentEncodingSpec {
   constexpr SegmentEncodingSpec() : encoding_type{EncodingType::Dictionary} {}
   constexpr SegmentEncodingSpec(EncodingType encoding_type_) : encoding_type{encoding_type_} {}  // NOLINT
-  constexpr SegmentEncodingSpec(EncodingType encoding_type_, VectorCompressionType vector_compression_type_)
+  constexpr SegmentEncodingSpec(EncodingType encoding_type_,
+                                std::optional<VectorCompressionType> vector_compression_type_)
```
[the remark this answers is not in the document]
Bouncner: Good catch.

src/lib/storage/segment_encoding_utils.hpp
Bouncner: Changed it.

Bouncner:
I think I've addressed all your issues.

Issue #623, opened 2018-01-12
PolymorphicAllocator not propagated


mjendruk:
In order to measure how much memory is allocated by the column encoders, I created a sub-class of the `NUMAMemoryResource` and used it in our `PolymorphicAllocator`. My custom memory resource contained a simple variable counting the allocations and deallocations I was very surprised to find out that in some cases it did not measure anything. In fact, although I passed the memory resource to the allocator, its methods weren’t called at all. In the end, I found the reason in a side note of the standard library’s documentation:

> polymorphic_allocator does not propagate on container copy assignment, move assignment, or swap. As a result, move assignment of a polymorphic_allocator-using container can throw, and swapping two polymorphic_allocator-using containers whose allocators do not compare equal results in undefined behavior.

Basically it means that whatever allocator a container was initialized with cannot be replaced later on. The following code will lead to copy instead of a move of the vector and the polymorphic allocator will not be copied. The consequence is that in some occasions, our columns and chunks may lie on a different node than we expect if we don’t pay attention. There is unfortunately no warning emitted.

```C++
pmr_vector<int> vector_a{};  // could be a member variable

auto alloc = ... // an instance of PolymorphicAllocator
pmr_vector<int> vector_b{alloc};

for (...) { vector_b.push_back(...); }  // fill vector_b

vector_a = vector_b;  // will copy the content of vector_b into vector_a, but not its allocator
vector_a = std::move(vector_b);  // does the same!
```

Instead, this would work:
```C++
auto alloc = ... // an instance of PolymorphicAllocator

pmr_vector<int> vector_a{alloc};  // could be a member variable

pmr_vector<int> vector_b{alloc};

for (...) { vector_b.push_back(...); }  // fill vector_b

vector_a = vector_b;  // will copy the content of vector_b into vector_a, allocator is already the same
vector_a = std::move(vector_b);  // moves the content of vector_b into vector_a
```

I think this needs to be documented somewhere.

PR #2050, opened 2020-03-19
Fix PMR in FixedStringDictionarySegments, include in tests

mrks:
While working on something else, I noticed that 
* FixedStringDictionarySegments did not correctly implement copy_using_allocator (related to #623),
* this was not tested because the FixedStringDictionary was not included in the list of encodings to test, and
* the existing tests did not use external memory for pmr_strings.

All should be fixed now.

Also, this fixes and tests that segment access counters are persistent across migrations.

### DIFF
--- a/src/lib/storage/fixed_string_dictionary_segment.cpp
+++ b/src/lib/storage/fixed_string_dictionary_segment.cpp
@@ -56,8 +56,8 @@ ChunkOffset FixedStringDictionarySegment<T>::size() const {
 template <typename T>
 std::shared_ptr<BaseSegment> FixedStringDictionarySegment<T>::copy_using_allocator(
     const PolymorphicAllocator<size_t>& alloc) const {
+  auto new_dictionary = std::make_shared<FixedStringVector>(*_dictionary, alloc);
   auto new_attribute_vector = _attribute_vector->copy_using_allocator(alloc);
-  auto new_dictionary = std::make_shared<FixedStringVector>(*_dictionary);
   return std::make_shared<FixedStringDictionarySegment<T>>(new_dictionary, std::move(new_attribute_vector));
 }
 
--- a/src/lib/storage/fixed_string_dictionary_segment.cpp
+++ b/src/lib/storage/fixed_string_dictionary_segment.cpp
@@ -58,7 +58,12 @@ std::shared_ptr<BaseSegment> FixedStringDictionarySegment<T>::copy_using_allocat
     const PolymorphicAllocator<size_t>& alloc) const {
   auto new_dictionary = std::make_shared<FixedStringVector>(*_dictionary, alloc);
   auto new_attribute_vector = _attribute_vector->copy_using_allocator(alloc);
-  return std::make_shared<FixedStringDictionarySegment<T>>(new_dictionary, std::move(new_attribute_vector));
+
+  auto copy = std::make_shared<FixedStringDictionarySegment<T>>(new_dictionary, std::move(new_attribute_vector));
+
+  copy->access_counter = access_counter;
+
+  return copy;
 }
 
 template <typename T>
--- a/src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
+++ b/src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
@@ -11,6 +11,12 @@
 
 namespace opossum {
 
+FixedStringVector::FixedStringVector(const FixedStringVector& other, PolymorphicAllocator<char> allocator)
+    : _string_length(other._string_length), _chars(other._chars, allocator), _size(other._size) {
+  // We need to set `_chars` in the initializer list. Otherwise, it would be first created with the default allocator
+  // and the assignment operator would not exchange the allocator.
+}
+
 void FixedStringVector::push_back(const pmr_string& string) {
   Assert(string.size() <= _string_length, "Inserted string is too long to insert in FixedStringVector");
   const auto pos = _chars.size();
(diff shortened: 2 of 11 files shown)


src/lib/storage/value_segment.cpp
```diff
@@ -119,7 +119,6 @@ template <typename T>
 void ValueSegment<T>::resize(const size_t size) {
   DebugAssert(size > _values.size() && size <= _values.capacity(),
               "ValueSegments should not be shrunk or resized beyond their original capacity");
-  access_counter[SegmentAccessCounter::AccessType::Sequential] += _values.size();
```
mrks: @mflueggen Noticed this while working on something else. As the vectors are preallocated, we never move data. In fact, multi-threading on ValueSegment would be completely broken if the data was ever touched as part of resize. As such, this counter increment looks incorrect.

mflueggen: True.

src/lib/storage/value_segment.cpp
```diff
@@ -119,7 +119,6 @@ template <typename T>
 void ValueSegment<T>::resize(const size_t size) {
   DebugAssert(size > _values.size() && size <= _values.capacity(),
```
mflueggen: Shouldn't a `resize` to `size` be ok?
```suggestion
  DebugAssert(size >= _values.size() && size <= _values.capacity(),
```

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
```diff
@@ -11,6 +11,12 @@
 
 namespace opossum {
 
+FixedStringVector::FixedStringVector(const FixedStringVector& other, const PolymorphicAllocator<char>& allocator)
+    : _string_length(other._string_length), _chars(other._chars, allocator), _size(other._size) {
+  // We need to set `_chars` in the initializer list. Otherwise, it would be first created with the default allocator
+  // and the assignment operator would not exchange the allocator.
```
mflueggen:
```suggestion
```

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
```diff
@@ -11,6 +11,12 @@
 
 namespace opossum {
 
+FixedStringVector::FixedStringVector(const FixedStringVector& other, const PolymorphicAllocator<char>& allocator)
+    : _string_length(other._string_length), _chars(other._chars, allocator), _size(other._size) {
+  // We need to set `_chars` in the initializer list. Otherwise, it would be first created with the default allocator
```
mflueggen:
```suggestion
  // '_chars' needs to be initialized in the initializer list to use the given allocator.
```

src/lib/storage/value_segment.cpp
mrks: The message might be slightly misleading, but I don't see the need splitting up the two. While resizing it to the current size is not really wrong, doing so is a good sign that you are doing something wrong.

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
mrks: Not sure what your suggestion is. The point is that `operator=` does not modify the allocator. Maybe that should be added?

mrks: Ah, what Github is slightly confusing. Just saw the comment on the other line. Maybe:

"For pmr_vectors, `operator=` does not change the allocator. As such, we need to set `_chars` in the initializer list. Otherwise, it would be first created with the default allocator and the allocator that is passed in would be ignored."?

src/lib/storage/fixed_string_dictionary_segment.cpp
```diff
@@ -56,8 +56,8 @@ ChunkOffset FixedStringDictionarySegment<T>::size() const {
 template <typename T>
 std::shared_ptr<BaseSegment> FixedStringDictionarySegment<T>::copy_using_allocator(
```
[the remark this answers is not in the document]
mrks: Do we have a test that tests the counters in the context of a segment? I can only find `SegmentAccessCounterTest`.

src/lib/storage/value_segment.cpp
mflueggen: But should you resize to capacity?

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
mflueggen: Maybe:
"For pmr_vectors, operator= does not change the allocator. As such, we need to set _chars in the initializer list. Otherwise, it would be created using the default allocator and ignoring the passed in allocator."

src/lib/storage/fixed_string_dictionary_segment.cpp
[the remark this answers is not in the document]
mflueggen: Partially. In EncodedSegmentTest you will find a few tests. They are in that test suite because I require almost all helper functions of that test. I should refactor those tests and helper functions. In my fork I already extracted the helper functions because I use them for a lot of others tests.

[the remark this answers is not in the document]
mrks: What do you think is easier? I can add it to `SegmentsUsingAllocatorsTest.MigrateSegment`, but if you say that your refactored tests are a better place to test this (also using `get_supporting_segment_encodings_specs`), we can do it there.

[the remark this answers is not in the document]
mflueggen: Implementing the test in SegmentsUsingAllocatorsTest.MigrateSegment seems totally valid to me.

PR #2053, opened 2020-03-23
Size estimation utils: make seed static

Bouncner:
This mini PR changes the seed to be static in the size estimation utils.

The size estimation utils are used to determine the size of string segments, for which random sampling is used. Currently, the seed is random with the effect of steadily changing values for the size of a segment. Since the segments don't change, the size should not change as well.
This effect is especially obvious when the memory usage is steadily requested and visualized (e.g., in the bachelor's project's cockpit).

The potential short comings of this approach are only theoretical in my opinion as the meta table is not meant to be requested multiple times in order to capture a more accurate estimate.
If this would be the case, we could add another Sampling Enum value (SampledStatic, SampledRandom), but I don't see the necessity right now. If anybody sees that, I'll add it.


mrks:
When merging a PR with a single commit, please make sure you update the commit message.

Issue #1902, opened 2019-10-29
Double COUNT(*) is broken


mrks:
When trying to find edge cases, I found that `SELECT COUNT(*) FROM (SELECT COUNT(*) FROM t WHERE a <= 1234) t2` is broken in the current master, returning 2 instead of 1. Any idea if this PR helps us in addressing that? Otherwise, we can just move it to a new issue.

_Originally posted by @mrks in https://github.com/hyrise/hyrise/pull/1895_

mrks:
I'll introduce a PR in a couple of minutes that will detect these cases and throw an exception. As long as this (1) does not produce wrong results and (2) does not get in the way of experiments or benchmarks, it feels more like a cosmetic issue, similar to a couple of others mentioned in the "Expression issues" list.

PR #2057, opened 2020-03-24
Catch duplicate aggregate functions

mrks:
These queries are currently broken because the SQLTranslator incorrectly identifies the aggregate function as having already been calculated (see #1902):

```
SELECT COUNT(*) FROM (SELECT COUNT(*) FROM t WHERE a <= 1234) t2
SELECT COUNT(a) FROM (SELECT a, COUNT(a) FROM t GROUP BY a) t2
SELECT AVG(a) FROM (SELECT a, AVG(a) FROM t GROUP BY a) t3
```

As long as this (1) does not produce wrong results and (2) does not get in the way of experiments or benchmarks, it feels more like a cosmetic issue. This PR introduces a check for these cases.

### DIFF
--- a/src/test/sql/sql_translator_test.cpp
+++ b/src/test/sql/sql_translator_test.cpp
@@ -470,7 +470,8 @@ TEST_F(SQLTranslatorTest, RepeatingAggregates) {
 
   EXPECT_THROW(compile_query("SELECT COUNT(a) FROM (SELECT a, COUNT(a) FROM t GROUP BY a) t2"), InvalidInputException);
 
-  EXPECT_THROW(compile_query("SELECT COUNT(a) FROM (SELECT a, COUNT(a) AS b FROM t GROUP BY a) t2"), InvalidInputException);
+  EXPECT_THROW(compile_query("SELECT COUNT(a) FROM (SELECT a, COUNT(a) AS b FROM t GROUP BY a) t2"),
+               InvalidInputException);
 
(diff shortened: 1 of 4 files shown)


mrks:
WIP again: The tests worked, but TPC-DS doesn't anymore.

mrks:
This looks better, ready for review again.

src/test/sql/sql_translator_test.cpp
```diff
@@ -461,6 +461,21 @@ TEST_F(SQLTranslatorTest, SelectListAliasesDifferentForSimilarAggregatesInSubque
...
+
+  EXPECT_THROW(compile_query("SELECT COUNT(*) FROM (SELECT COUNT(*) FROM t WHERE a <= 1234) t2"),
+               InvalidInputException);
+
+  EXPECT_THROW(compile_query("SELECT COUNT(a) FROM (SELECT a, COUNT(a) FROM t GROUP BY a) t2"), InvalidInputException);
+
+  EXPECT_THROW(compile_query("SELECT COUNT(a) FROM (SELECT a, COUNT(a) AS b FROM t GROUP BY a) t2"),
+               InvalidInputException);
+
+  EXPECT_THROW(compile_query("SELECT AVG(a) FROM (SELECT a, AVG(a) FROM t GROUP BY a) t3"), InvalidInputException);
```
julianmenzler: I would omit the table aliases `t2`, `t3` as they do not contribute to the problem.

mrks: Subqueries in the FROM part without an alias are illegal.

julianmenzler: Oh, I was not aware of that. Thank you!

PR #2058, opened 2020-03-24
Import/Export FixedStringDictionarySegments binary

### LINKED ISSUE #1367: Binary import/export: Support all encoding types

### LINKED ISSUE #2052: Export FixedStringDictionarySegments binary
closes #1367

dey4ss:
closes #2052
closes #1367


src/test/lib/import_export/binary/binary_parser_test.cpp
```diff
@@ -77,10 +75,14 @@ TEST_P(BinaryParserMultiEncodingTest, StringSegment) {
   expected_table->append({"a"});
   expected_table->append({"test"});
 
+  testing::internal::CaptureStdout();
+
   std::string reference_filename =
       _reference_filepath + ::testing::UnitTest::GetInstance()->current_test_info()->name() + ".bin";
   auto table = BinaryParser::parse(reference_filename);
 
+  std::cout << testing::internal::GetCapturedStdout() << std::endl;
```
mrks: What are you doing here?

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.hpp
```diff
@@ -58,14 +58,16 @@ class FixedStringVector {
   ReverseIterator rend() noexcept;
 
   // Return a pointer to the underlying memory
-  char* data();
+  const char* data() const;
```
mrks: As I just had to look this up again: Do you know cdecl.org?

src/lib/import_export/binary/binary_writer.hpp
```diff
@@ -125,12 +125,35 @@ class BinaryWriter {
...
+   * FixedStringDictionary Segments are dumped with the following layout:
+   *
+   * Description           | Type                                  | Size in bytes
+   * -----------------------------------------------------------------------------------------
+   * Column Type           | ColumnType                            |   1
+   * Width of attribute v. | AttributeVectorWidth                  |   1
+   * Size of dictionary v. | ValueID                               |   4
+   * FixedString length    | uint32_t                              |   8
+   * Dictionary Values     | char[]                                |   dict. size * fixed string length
+   * Attribute v. values   | uintX                                 |   rows * width of attribute v.
```
mrks: Would writing out "vector" break the 120 char limit?

src/lib/import_export/binary/binary_writer.hpp
```diff
@@ -125,12 +125,35 @@ class BinaryWriter {
...
    *
-   * @param base_dictionary_segment The segment to export
+   * @param dictionary_segment The segment to export
    * @param ofstream The output stream for exporting
    */
   template <typename T>
   static void _write_segment(const DictionarySegment<T>& dictionary_segment, std::ofstream& ofstream);
 
+  /**
+   * FixedStringDictionary Segments are dumped with the following layout:
```
mrks: FixedStringDictionarySegments to facilitate grepping

src/lib/import_export/binary/binary_writer.cpp
```diff
@@ -200,6 +204,33 @@ void BinaryWriter::_write_segment(const DictionarySegment<T>& dictionary_segment
                             *dictionary_segment.attribute_vector());
 }
 
+template <typename T>
+void BinaryWriter::_write_segment(const FixedStringDictionarySegment<T>& fixed_string_dictionary_segment,
+                                  std::ofstream& ofstream) {
+  Assert(fixed_string_dictionary_segment.compressed_vector_type(),
+         "Expected DictionarySegment to use vector compression for attribute vector");
+  Assert(is_fixed_size_byte_aligned(*fixed_string_dictionary_segment.compressed_vector_type()),
+         "Does only support fixed-size byte-aligned compressed attribute vectors.");
```
mrks: This is also asserted in _compressed_vector_width. Maybe you could harmonize this across the entire file?

src/lib/import_export/binary/binary_writer.hpp
```diff
@@ -125,12 +125,35 @@ class BinaryWriter {
...
+   * -----------------------------------------------------------------------------------------
+   * Column Type           | ColumnType                            |   1
+   * Width of attribute v. | AttributeVectorWidth                  |   1
+   * Size of dictionary v. | ValueID                               |   4
+   * FixedString length    | uint32_t                              |   8
+   * Dictionary Values     | char[]                                |   dict. size * fixed string length
+   * Attribute v. values   | uintX                                 |   rows * width of attribute v.
+   *
+   * Please note that the number of rows are written in the header of the chunk.
+   * The type of the column can be found in the global header of the file.
```
mrks: This is a good comment. I actually looked at other code first, and had that exact question.

This applies to every encoding, right? It might not hurt to copy it.

src/lib/import_export/binary/binary_writer.hpp
```diff
@@ -125,12 +125,35 @@ class BinaryWriter {
...
+   *
+   *
+   * @param fixed_string_dictionary_segment The segment to export
+   * @param ofstream The output stream for exporting
```
mrks: Doesn't really add any information if you ask me. While looking at other examples, I found this outdated one:

```
   * @param reference_segment The segment to export
   * @param base_context A context in the form of an ExportContext. Contains a reference to the ofstream.
   */
  static void _write_segment(const ReferenceSegment& reference_segment, std::ofstream& ofstream);
```

src/lib/import_export/binary/binary_writer.hpp
```diff
@@ -125,12 +125,35 @@ class BinaryWriter {
...
    */
   template <typename T>
   static void _write_segment(const DictionarySegment<T>& dictionary_segment, std::ofstream& ofstream);
 
+  /**
+   * FixedStringDictionary Segments are dumped with the following layout:
+   *
+   * Description           | Type                                  | Size in bytes
+   * -----------------------------------------------------------------------------------------
+   * Column Type           | ColumnType                            |   1
```
mrks: Below, you say "The type of the column can be found in the global header of the file."

Don't you mean EncodingType? Same for other encodings.

src/lib/import_export/binary/binary_writer.cpp
```diff
@@ -200,6 +204,33 @@ void BinaryWriter::_write_segment(const DictionarySegment<T>& dictionary_segment
...
+  // Write the dictionary size and dictionary
+  export_value(ofstream,
+               static_cast<ValueID::base_type>(fixed_string_dictionary_segment.fixed_string_dictionary()->size()));
+  const auto string_length = fixed_string_dictionary_segment.fixed_string_dictionary()->string_length();
+  export_value(ofstream, static_cast<uint32_t>(string_length));
+  export_values(ofstream, *fixed_string_dictionary_segment.fixed_string_dictionary());
+
+  // Write attribute vector
+  Assert(fixed_string_dictionary_segment.compressed_vector_type(),
+         "Expected DictionarySegment to use vector compression for attribute vector");
```
mrks: Should have been caught earlier.

src/lib/import_export/binary/binary_writer.cpp
```diff
@@ -62,6 +62,10 @@ void export_values(std::ofstream& ofstream, const std::vector<T, Alloc>& values)
   ofstream.write(reinterpret_cast<const char*>(values.data()), values.size() * sizeof(T));
 }
 
+void export_values(std::ofstream& ofstream, const FixedStringVector& values) {
+  ofstream.write(reinterpret_cast<const char*>(values.data()), values.size() * values.string_length());
```
mrks: `reinterpret_cast` should not be necessary here.

src/lib/import_export/binary/binary_parser.cpp
```diff
@@ -285,4 +299,19 @@ std::unique_ptr<const BaseCompressedVector> BinaryParser::_import_offset_value_v
...
+    std::ostringstream value;
+    for (uint32_t c = 0; c < string_length; c++) {
+      value << _read_value<char>(file);
+    }
+    values[row] = pmr_string(value.str());
+  }
+
+  return std::make_shared<FixedStringVector>(values.begin(), values.end(), string_length);
+}
+
```
mrks: This looks quite expensive. Feel free to add a new (private and friended?) constructor to FixedStringVector. Something like this:

```
const auto string_length = _read_value<uint32_t>(file);

pmr_vector<char> v(string_length * count);
file.read(&v[0], string_length * count);
return std::make_shared<FixedStringVector>(std::move(v), string_length);
```

src/lib/import_export/binary/binary_parser.cpp
```diff
@@ -133,6 +133,8 @@ std::shared_ptr<BaseSegment> BinaryParser::_import_segment(std::ifstream& file,
       return _import_value_segment<ColumnDataType>(file, row_count, is_nullable);
     case EncodingType::Dictionary:
       return _import_dictionary_segment<ColumnDataType>(file, row_count);
+    case EncodingType::FixedStringDictionary:
```
mrks: This should allow us to remove the `default:` here and in the writer, correct?

src/test/lib/import_export/binary/binary_parser_test.cpp
Bouncner: Wow, that sounds super aggressive. 😲

dey4ss: That's a debugging relict :D

src/lib/import_export/binary/binary_writer.hpp
dey4ss: Actually, I copied it from DictionarySegment comment; this is part of every segment export comment.

src/lib/import_export/binary/binary_parser.cpp
dey4ss: `default` catches and fails if the encoding type in the binary file is invalid, so I think it is okay to keep it.

mrks: How about moving the fail outside of the switch? That way we still handle invalid encoding types but the compiler complains about missing (valid) encoding types.

src/lib/import_export/binary/binary_parser.cpp
dey4ss: When making this constructor non-public, we cannot call `make_shared`, right?

mrks: Mmh, ok. Making it public probably doesn't hurt.

src/lib/import_export/binary/binary_parser.cpp
dey4ss: If we only want to be sure that we crash, we would not need to add anything, as reading the value as EncodingType should fail. If there is no need to add a failure notice (i.e. reading the value as int, check if is not larger than there are values in EncodingType and finally cast it), this could be enough.

mrks: My point is that removing `default` adds compile-time checking that all encodings are covered

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.hpp
dey4ss: Didn't know this :D
The reason for adding this was that I guessed that we do not want to be able to change the data when accessing `data()` and then `const char *` is required.

src/lib/import_export/binary/binary_parser.cpp
dey4ss: Yes, of course, I just wanted to know if we want to have a hint if the EncodingType is correct.

mrks: You might be overcomplicating this. If the switch case gets an invalid value, it will jump to the end without executing any case. By moving the `Fail` outside the switch statement, that is what is being executed. If you search for `Fail("Invalid enum value");`, you will see this already being done in different places.

src/lib/import_export/binary/binary_parser.cpp
```diff
@@ -300,17 +304,9 @@ std::unique_ptr<const BaseCompressedVector> BinaryParser::_import_offset_value_v
...
-    for (uint32_t c = 0; c < string_length; c++) {
-      value << _read_value<char>(file);
-    }
-    values[row] = pmr_string(value.str());
-  }
-
-  return std::make_shared<FixedStringVector>(values.begin(), values.end(), string_length);
+  pmr_vector<char> values(string_length * count);
+  file.read(values.data(), values.size());
+  return std::make_shared<FixedStringVector>(values, string_length);
```
mrks: `std::move(values)`

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.hpp
mrks: Yepp, completely correct.

src/lib/import_export/binary/binary_parser.cpp
dey4ss: Yes, this wa stupid :D

src/lib/import_export/binary/binary_parser.cpp
dey4ss: I had a look at the FixedSizeByteAlignedVector constructor and, like there, did the move inside the value assignment in the constructor.

mrks: Not sure what you are saying. The constructor takes a `pmr_vector<char> chars`, which is a call by value. If you don't move `values` in, it will be copied.

Whatever happens inside the constructor is a completely different thing.

dey4ss: Ah, okay, I added this. So the compressed vectors of the other segments are copied while reading binary then?

mrks: Could you give me a line number, please? If you are talking about this:

https://github.com/hyrise/hyrise/blob/4b842df725c16fe24c536da8ea0ef8994d7e568d/src/lib/import_export/binary/binary_parser.cpp#L295

nothing is copied because _read_values is a temporary (see: guaranteed copy elision).

If you are talking about this:

https://github.com/hyrise/hyrise/blob/4b842df725c16fe24c536da8ea0ef8994d7e568d/src/lib/import_export/binary/binary_parser.cpp#L200-L203

`null_values` is copied, but as it is a shared_ptr, it is by far not as bad as copying the actual vector.

mrks:
Did you cancel the CI run on purpose or is Jenkins acting up?

dey4ss:
There was a Segfault in SQLiteTestRunner and I wanted to know if this was just an incidence, as I did not know why there should be any problems

src/lib/import_export/binary/binary_parser.cpp
dey4ss: Okay, I see :)

mrks:
Is it tracked anywhere?

dey4ss:
> 
> 
> No, shouldn't be related to this. If this does not occur in this branch any more, please make sure that the segfault is tracked somewhere. :)

Seems like it occurred again. On pella, there is no problem (also tested it there with gcc release).

mrks:
Hmm, failed again:

```
[ RUN      ] SQLiteTestRunnerEncodings/SQLiteTestRunner.CompareToSQLite/Line119WithRunLength

script returned exit code 139
```

Paging almost-doctor @Bouncner to the development room. It seems more likely that this is related to #2048 than to the changes in this PR.

dey4ss:
Also the lines doffer somehow: We had 115 and 119

mrks:
But run-length encoding in both cases?

dey4ss:
> 
> 
> But run-length encoding in both cases?

Yes

mrks:
Restarted it again to get a better idea of how prevalent this is

mrks:
Weird. I have build gcc-release and gcc-relwithdebinfo in the docker image on the CI server and cannot reproduce it. Let me try if clearing ccache changes anything.

mrks:
Seems to have helped. Let's keep an eye on this.

PR #2061, opened 2020-03-26
Store row count in performance data, don't keep temporaries

### LINKED ISSUE #1646: PQP visualization relies on CleanupTemporaries::No
Otherwise we lose almost all information  like timings, table sizes - that could be gained from visualization:

CleanupTemporaries::Yes
image

CleanupTemporaries::No
image

Tables sizes should be stored in the OperatorPerformanceData and visualization shouldn't check for get_output_table != nullptr when deciding whether this meta data should be visualized

mrks:
* For each operator, this stores the resulting chunk and row count in OperatorPerformanceData
* As a result, we don't need the result table anymore to visualize the plan (fixes #1646)
* As a result of that, we don't need CleanupTemporaries anymore.
* To reduce the impact of having to calculate the row_count, we now cache the row count for reference tables.

```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.2954007387161255 | 78    | 1.2995855808258057 | 78    | +0%        |  0.4136 |
+| TPC-H 02       | 79.22210693359375  | 4754  | 84.02629089355469  | 5042  | +6%        |  0.0000 |
 | TPC-H 03       | 8.796109199523926  | 528   | 9.104130744934082  | 547   | +4%        |  0.0000 |
 | TPC-H 04       | 7.861326694488525  | 472   | 8.128789901733398  | 488   | +3%        |  0.0000 |
 | TPC-H 05       | 4.098977565765381  | 246   | 4.150041103363037  | 250   | +1%        |  0.0360 |
 | TPC-H 06       | 262.84381103515625 | 15771 | 270.1571044921875  | 16210 | +3%        |  0.0000 |
 | TPC-H 07       | 9.691267967224121  | 582   | 9.99113655090332   | 600   | +3%        |  0.0000 |
 | TPC-H 08       | 7.426480293273926  | 446   | 7.6607770919799805 | 460   | +3%        |  0.0000 |
 | TPC-H 09       | 1.749589204788208  | 105   | 1.7984514236450195 | 108   | +3%        |  0.0000 |
 | TPC-H 10       | 3.3428919315338135 | 201   | 3.394002914428711  | 204   | +2%        |  0.1044 |
+| TPC-H 11       | 37.64109420776367  | 2259  | 39.703086853027344 | 2383  | +5%        |  0.0000 |
 | TPC-H 12       | 14.899829864501953 | 894   | 15.280900001525879 | 917   | +3%        |  0.0000 |
+| TPC-H 13       | 2.3008885383605957 | 139   | 2.657020330429077  | 160   | +15%       |  0.0000 |
 | TPC-H 14       | 43.536373138427734 | 2613  | 42.64398956298828  | 2559  | -2%        |  0.0000 |
 | TPC-H 15       | 118.18135833740234 | 7091  | 117.94392395019531 | 7077  | -0%        |  0.0031 |
 | TPC-H 16       | 9.004341125488281  | 541   | 9.330336570739746  | 560   | +4%        |  0.0000 |
 | TPC-H 17       | 5.367760181427002  | 323   | 5.577611923217773  | 335   | +4%        |  0.0000 |
 | TPC-H 18       | 1.2562159299850464 | 76    | 1.2815972566604614 | 77    | +2%        |  0.0019 |
 | TPC-H 19       | 9.865779876708984  | 592   | 10.181819915771484 | 612   | +3%        |  0.0000 |
+| TPC-H 20       | 31.38033676147461  | 1883  | 33.14265060424805  | 1989  | +6%        |  0.0000 |
 | TPC-H 21       | 1.3725528717041016 | 83    | 1.4100109338760376 | 85    | +3%        |  0.0000 |
 | TPC-H 22       | 18.62116813659668  | 1118  | 18.42943572998047  | 1106  | -1%        |  0.0000 |
 | geometric mean |                    |       |                    |       | +3%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

Not sure if the changes, e.g., in Q13 are really caused by this. At least we don't break anything.

### DIFF
--- a/src/lib/operators/operator_performance_data.hpp
+++ b/src/lib/operators/operator_performance_data.hpp
@@ -13,7 +13,10 @@ namespace opossum {
 struct OperatorPerformanceData : public Noncopyable {
   virtual ~OperatorPerformanceData() = default;
 
+  bool executed{false};
   std::chrono::nanoseconds walltime{0};
+  uint64_t output_row_count{0};
+  uint64_t output_chunk_count{0};
 
   virtual void output_to_stream(std::ostream& stream,
                                 DescriptionMode description_mode = DescriptionMode::SingleLine) const;
--- a/src/lib/operators/operator_performance_data.hpp
+++ b/src/lib/operators/operator_performance_data.hpp
@@ -15,6 +15,10 @@ struct OperatorPerformanceData : public Noncopyable {
 
   bool executed{false};
   std::chrono::nanoseconds walltime{0};
+
+  // Some operators do not return a table (e.g., Insert).
+  // Note: The operator returning an empty table will be expressed as has_output == true, output_row_count == 0
+  bool has_output{false};
   uint64_t output_row_count{0};
   uint64_t output_chunk_count{0};
 
--- a/src/lib/scheduler/operator_task.cpp
+++ b/src/lib/scheduler/operator_task.cpp
@@ -12,8 +12,7 @@
 #include "utils/tracing/probes.hpp"
 
 namespace opossum {
-OperatorTask::OperatorTask(std::shared_ptr<AbstractOperator> op,
-                           SchedulePriority priority, bool stealable)
+OperatorTask::OperatorTask(std::shared_ptr<AbstractOperator> op, SchedulePriority priority, bool stealable)
     : AbstractTask(priority, stealable), _op(std::move(op)) {}
 
 std::string OperatorTask::description() const {
@@ -98,7 +97,7 @@ void OperatorTask::_on_execute() {
   // root (i.e., the final result)
   for (const auto& weak_predecessor : predecessors()) {
     const auto predecessor = std::dynamic_pointer_cast<OperatorTask>(weak_predecessor.lock());
-    DebugAssert(predecessor, "predecessor of OperatorTask is not an OperatorTask itself");
+    DebugAssert(predecessor, "Predecessor of OperatorTask is not an OperatorTask itself");
     auto previous_operator_still_needed = false;
 
     for (const auto& successor : predecessor->successors()) {
--- a/src/lib/scheduler/operator_task.cpp
+++ b/src/lib/scheduler/operator_task.cpp
@@ -12,39 +12,38 @@
 #include "utils/tracing/probes.hpp"
 
 namespace opossum {
-OperatorTask::OperatorTask(std::shared_ptr<AbstractOperator> op, CleanupTemporaries cleanup_temporaries,
+OperatorTask::OperatorTask(std::shared_ptr<AbstractOperator> op,
                            SchedulePriority priority, bool stealable)
-    : AbstractTask(priority, stealable), _op(std::move(op)), _cleanup_temporaries(cleanup_temporaries) {}
+    : AbstractTask(priority, stealable), _op(std::move(op)) {}
 
 std::string OperatorTask::description() const {
   return "OperatorTask with id: " + std::to_string(id()) + " for op: " + _op->description();
 }
 
 std::vector<std::shared_ptr<OperatorTask>> OperatorTask::make_tasks_from_operator(
-    const std::shared_ptr<AbstractOperator>& op, CleanupTemporaries cleanup_temporaries) {
+    const std::shared_ptr<AbstractOperator>& op) {
   std::vector<std::shared_ptr<OperatorTask>> tasks;
   std::unordered_map<std::shared_ptr<AbstractOperator>, std::shared_ptr<OperatorTask>> task_by_op;
-  _add_tasks_from_operator(op, tasks, task_by_op, cleanup_temporaries);
+  _add_tasks_from_operator(op, tasks, task_by_op);
   return tasks;
 }
 
 std::shared_ptr<OperatorTask> OperatorTask::_add_tasks_from_operator(
     const std::shared_ptr<AbstractOperator>& op, std::vector<std::shared_ptr<OperatorTask>>& tasks,
-    std::unordered_map<std::shared_ptr<AbstractOperator>, std::shared_ptr<OperatorTask>>& task_by_op,
-    CleanupTemporaries cleanup_temporaries) {
+    std::unordered_map<std::shared_ptr<AbstractOperator>, std::shared_ptr<OperatorTask>>& task_by_op) {
   const auto task_by_op_it = task_by_op.find(op);
   if (task_by_op_it != task_by_op.end()) return task_by_op_it->second;
 
-  const auto task = std::make_shared<OperatorTask>(op, cleanup_temporaries);
+  const auto task = std::make_shared<OperatorTask>(op);
   task_by_op.emplace(op, task);
 
   if (auto left = op->mutable_input_left()) {
-    auto subtree_root = _add_tasks_from_operator(left, tasks, task_by_op, cleanup_temporaries);
+    auto subtree_root = _add_tasks_from_operator(left, tasks, task_by_op);
     subtree_root->set_as_predecessor_of(task);
   }
 
   if (auto right = op->mutable_input_right()) {
-    auto subtree_root = _add_tasks_from_operator(right, tasks, task_by_op, cleanup_temporaries);
+    auto subtree_root = _add_tasks_from_operator(right, tasks, task_by_op);
     subtree_root->set_as_predecessor_of(task);
   }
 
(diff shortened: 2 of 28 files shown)


Bouncner:
Cpuld you take a TPCH measurement with clang?
I found the current master to be very unstable, even over runs of the exact same binary. Not sure what’s going on ...

mrks:
clang:

```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.6903742551803589 | 102   | 1.6775745153427124 | 101   | -1%        |  0.0417 |
 | TPC-H 02       | 64.87124633789062  | 3893  | 64.64307403564453  | 3879  | -0%        |  0.6089 |
 | TPC-H 03       | 7.802369117736816  | 469   | 7.610366344451904  | 457   | -2%        |  0.0000 |
 | TPC-H 04       | 6.899388790130615  | 414   | 6.844541549682617  | 411   | -1%        |  0.0000 |
 | TPC-H 05       | 3.8702409267425537 | 233   | 3.872727870941162  | 233   | +0%        |  0.8323 |
+| TPC-H 06       | 261.1051330566406  | 15667 | 277.4822692871094  | 16649 | +6%        |  0.0000 |
 | TPC-H 07       | 8.381349563598633  | 503   | 8.320501327514648  | 500   | -1%        |  0.2842 |
 | TPC-H 08       | 6.237124443054199  | 375   | 6.189908981323242  | 372   | -1%        |  0.0103 |
 | TPC-H 09       | 1.7547038793563843 | 106   | 1.7594809532165527 | 106   | +0%        |  0.4051 |
 | TPC-H 10       | 3.307379722595215  | 199   | 3.28890061378479   | 198   | -1%        |  0.5657 |
 | TPC-H 11       | 32.59980010986328  | 1956  | 32.44746780395508  | 1947  | -0%        |  0.0000 |
 | TPC-H 12       | 13.90282917022705  | 835   | 13.764663696289062 | 826   | -1%        |  0.0000 |
+| TPC-H 13       | 2.507786512374878  | 151   | 2.974839210510254  | 179   | +19%       |  0.0000 |
 | TPC-H 14       | 42.65694046020508  | 2560  | 42.59769058227539  | 2556  | -0%        |  0.1963 |
 | TPC-H 15       | 118.458251953125   | 7108  | 117.11412811279297 | 7027  | -1%        |  0.0000 |
 | TPC-H 16       | 9.594493865966797  | 576   | 9.530340194702148  | 572   | -1%        |  0.0000 |
 | TPC-H 17       | 4.339743614196777  | 261   | 4.318838119506836  | 260   | -0%        |  0.0132 |
 | TPC-H 18       | 1.337972640991211  | 81    | 1.3226946592330933 | 80    | -1%        |  0.0735 |
 | TPC-H 19       | 8.26378345489502   | 496   | 8.216631889343262  | 494   | -1%        |  0.2381 |
 | TPC-H 20       | 26.79973030090332  | 1609  | 26.851661682128906 | 1612  | +0%        |  0.3369 |
 | TPC-H 21       | 1.2279857397079468 | 74    | 1.2088176012039185 | 73    | -2%        |  0.0016 |
 | TPC-H 22       | 16.28240203857422  | 977   | 16.162206649780273 | 970   | -1%        |  0.0000 |
 | geometric mean |                    |       |                    |       | +0%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

Interesting. There might even be some truth to the increased performance in Q13.

src/lib/operators/operator_performance_data.hpp
```diff
@@ -13,7 +13,10 @@ namespace opossum {
 struct OperatorPerformanceData : public Noncopyable {
   virtual ~OperatorPerformanceData() = default;
 
+  bool executed{false};
   std::chrono::nanoseconds walltime{0};
+  uint64_t output_row_count{0};
```
Bouncner: Wondering if there is value in using std::optional to make it explicit that some operators do not have an output?

src/lib/scheduler/operator_task.cpp
```diff
@@ -97,21 +96,19 @@ void OperatorTask::_on_execute() {
...
-      DebugAssert(predecessor, "predecessor of OperatorTask is not an OperatorTask itself");
-      auto previous_operator_still_needed = false;
-
-      for (const auto& successor : predecessor->successors()) {
-        if (successor.get() != this && !successor->is_done()) {
-          previous_operator_still_needed = true;
-        }
+  for (const auto& weak_predecessor : predecessors()) {
+    const auto predecessor = std::dynamic_pointer_cast<OperatorTask>(weak_predecessor.lock());
+    DebugAssert(predecessor, "predecessor of OperatorTask is not an OperatorTask itself");
```
Bouncner: /predecessor/Predecessor/ ?

src/lib/storage/table.cpp
```diff
@@ -151,13 +151,27 @@ void Table::append_mutable_chunk() {
...
+  if (_type == TableType::References && _cached_row_count && !HYRISE_DEBUG) {
+    return *_cached_row_count;
+  }
+
+  uint64_t row_count = 0;
   const auto chunk_count = _chunks.size();
   for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
     const auto chunk = get_chunk(chunk_id);
-    if (chunk) ret += chunk->size();
+    if (chunk) row_count += chunk->size();
```
Bouncner: Where do we cover the case of overallocated mutable chunks? At first sight, I did not find it in chunk.cpp or value_segment.cpp (chunk.cpp uses the first segment's size).

mrks: This is not part of `size()` or `row_count()` anywhere. The preallocation happens solely in the `capacity()` of the vectors.

Bouncner: What a dump question that was. 🤦🏻‍♂️

src/lib/scheduler/operator_task.cpp
mrks: legacy, fixed

src/lib/operators/operator_performance_data.hpp
mrks: That optional would also have to cover output_chunk_count. I added an additional flag:

```
  // Some operators do not return a table (e.g., Insert).
  // Note: The operator returning an empty table will be expressed as has_output == true, output_row_count == 0
  bool has_output{false};
  uint64_t output_row_count{0};
  uint64_t output_chunk_count{0};
```

mrks:
Added the flag. I sneaked some tests for this into other the tests for the TableScan and the Delete operators. Not sure if that is really pretty, but a dedicated test where the setup is more complex than the test itself don't seem to make sense either.

Bouncner:
I hope to remember the sneaked tests because I'll add a test suite for that op perf struct.

Any ideas on the performance impact? Do you have the branch built somewhere and can post the two PQP for Q13?

mrks:
Screenshot 2020-03-26 at 17 27 31

mrks:
> I hope to remember the sneaked tests because I'll add a test suite for that op perf struct.

My approach to these things is to add a `// TODO whatever I want to remember` so that lint complains when I try to open the PR ;)

Bouncner:
I would have thought both compiler showing a clear improvement in Q13 should have a clear reason. I can't see one.

The `row_count()` could be faster and the join hash does too many of it, but once execution is around 20 mu seconds on my laptop. That cannot be the reason.
Do you have an idea where the performance comes from?

Bouncner:
I added a lint-crashing todo to my code. I think this PR is zündable, but I am still curious where this performance change comes from.

mrks:
I'd like to fix the column name thing first. Will have a look at the runtime cost of row_count while I'm at it.

mrks:
Measurements on my laptop, as my nemea nodes are busy:

Without caching: 3 ms 451 µs for 108 iterations -> 31.79 µs / iteration -> 0.0056%
With caching: 2 ms 245 µs for 108 iterations -> 20.78 µs / iteration -> 0.0037%

Even if this was off by a factor of 100, it would not explain the behavior of Q13.

mrks:
The other change, removing CleanupTemporaries, does not have an impact either. No idea, but I wasted too much time in the past trying to track these things down.

PR #2062, opened 2020-03-26
Improve advance() for FrameOfReference Iterable

Bouncner:
This PR implements a better version of `advance()` for FrameOfReference iterables and furthermore improves the non-pos-list iterable.

The first implementation of `advance()` revealed that we update too many temporary values which was actually not necessary. Now, the non-pos-list iterator uses (similar to the pos-list iterator of the current master) pointers which avoids several iterator copies.

I know this iterable contains pointers, which we'd like to avoid. But I don't see a real alternative here and I don't think we should ignore the performance advantages.

Since we still don't use `advance()` a lot (#1531), the performance difference is mainly caused by switching to pointers.

Clang
```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.636426568031311  | 295   | 1.6385020017623901 | 295   | +0%        |  0.5418 |
+| TPC-H 02       | 69.05223083496094  | 12430 | 76.53771209716797  | 13777 | +11%       |  0.0000 |
 | TPC-H 03       | 8.12531566619873   | 1463  | 8.178707122802734  | 1473  | +1%        |  0.0000 |
+| TPC-H 04       | 6.891951084136963  | 1241  | 7.3427252769470215 | 1322  | +7%        |  0.0000 |
 | TPC-H 05       | 4.126384735107422  | 743   | 3.961472988128662  | 714   | -4%        |  0.0000 |
 | TPC-H 06       | 276.6238708496094  | 49793 | 277.742431640625   | 49994 | +0%        |  0.0000 |
+| TPC-H 07       | 9.745887756347656  | 1755  | 10.305323600769043 | 1855  | +6%        |  0.0000 |
+| TPC-H 08       | 6.7591094970703125 | 1217  | 7.2683844566345215 | 1309  | +8%        |  0.0000 |
 | TPC-H 09       | 2.0026559829711914 | 361   | 2.0436899662017822 | 368   | +2%        |  0.0000 |
 | TPC-H 10       | 3.2747294902801514 | 590   | 3.3086965084075928 | 596   | +1%        |  0.0682 |
+| TPC-H 11       | 31.51953887939453  | 5674  | 34.217124938964844 | 6160  | +9%        |  0.0000 |
 | TPC-H 12       | 14.126002311706543 | 2543  | 14.692754745483398 | 2645  | +4%        |  0.0000 |
 | TPC-H 13       | 3.127387762069702  | 563   | 3.1879122257232666 | 574   | +2%        |  0.0000 |
 | TPC-H 14       | 43.02265548706055  | 7745  | 43.52595138549805  | 7835  | +1%        |  0.0000 |
 | TPC-H 15       | 114.59480285644531 | 20628 | 115.71728515625    | 20830 | +1%        |  0.0000 |
-| TPC-H 16       | 9.46228313446045   | 1704  | 8.84090518951416   | 1592  | -7%        |  0.0000 |
+| TPC-H 17       | 4.650624752044678  | 838   | 5.099733829498291  | 918   | +10%       |  0.0000 |
+| TPC-H 18       | 1.2112091779708862 | 219   | 1.3071447610855103 | 236   | +8%        |  0.0000 |
+| TPC-H 19       | 8.77545166015625   | 1580  | 9.581247329711914  | 1725  | +9%        |  0.0000 |
+| TPC-H 20       | 27.90374755859375  | 5023  | 30.0426082611084   | 5408  | +8%        |  0.0000 |
 | TPC-H 21       | 1.2700884342193604 | 229   | 1.3275043964385986 | 239   | +5%        |  0.0000 |
 | TPC-H 22       | 16.8425350189209   | 3032  | 16.95630645751953  | 3053  | +1%        |  0.0000 |
 | geometric mean |                    |       |                    |       | +4%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

GCC (one of the worse runs, I've seen many better runs and a few worse runs)
```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.298175573348999  | 234   | 1.2783094644546509 | 231   | -2%        |  0.0000 |
 | TPC-H 02       | 94.96527862548828  | 17094 | 94.50982666015625  | 17012 | -0%        |  0.0758 |
 | TPC-H 03       | 9.682222366333008  | 1743  | 9.825042724609375  | 1769  | +1%        |  0.0000 |
 | TPC-H 04       | 8.320757865905762  | 1498  | 8.707646369934082  | 1568  | +5%        |  0.0000 |
 | TPC-H 05       | 4.495233058929443  | 810   | 4.406675815582275  | 794   | -2%        |  0.0000 |
 | TPC-H 06       | 268.7215270996094  | 48370 | 270.370849609375   | 48667 | +1%        |  0.0000 |
 | TPC-H 07       | 11.963168144226074 | 2154  | 12.17513370513916  | 2192  | +2%        |  0.0001 |
+| TPC-H 08       | 8.492789268493652  | 1529  | 8.959814071655273  | 1613  | +5%        |  0.0000 |
 | TPC-H 09       | 2.077895164489746  | 375   | 2.058915615081787  | 371   | -1%        |  0.0000 |
 | TPC-H 10       | 3.461792230606079  | 624   | 3.420833110809326  | 616   | -1%        |  0.0228 |
 | TPC-H 11       | 40.03861618041992  | 7207  | 41.19872283935547  | 7416  | +3%        |  0.0000 |
 | TPC-H 12       | 16.078540802001953 | 2895  | 16.21451759338379  | 2919  | +1%        |  0.0000 |
 | TPC-H 13       | 2.974128484725952  | 536   | 2.9788687229156494 | 537   | +0%        |  0.0953 |
 | TPC-H 14       | 45.6746826171875   | 8222  | 44.48848342895508  | 8008  | -3%        |  0.0000 |
 | TPC-H 15       | 116.4356918334961  | 20959 | 115.6893081665039  | 20825 | -1%        |  0.0000 |
 | TPC-H 16       | 9.101003646850586  | 1639  | 9.009899139404297  | 1622  | -1%        |  0.0000 |
+| TPC-H 17       | 6.1009697914123535 | 1099  | 6.609766006469727  | 1190  | +8%        |  0.0000 |
 | TPC-H 18       | 1.2001221179962158 | 217   | 1.157517671585083  | 209   | -4%        |  0.0000 |
+| TPC-H 19       | 11.176178932189941 | 2012  | 11.905442237854004 | 2143  | +7%        |  0.0000 |
 | TPC-H 20       | 34.828914642333984 | 6270  | 36.074440002441406 | 6494  | +4%        |  0.0000 |
 | TPC-H 21       | 1.5242425203323364 | 275   | 1.5490636825561523 | 279   | +2%        |  0.0000 |
 | TPC-H 22       | 19.567638397216797 | 3523  | 19.08101463317871  | 3435  | -2%        |  0.0000 |
 | geometric mean |                    |       |                    |       | +1%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

Measurement are unstable. It's mostly between +2 - +5%, but I've also seen -1%. Not sure what the current problem is.


Bouncner:
FYI: I plan to remove that performance gain at a later point in time again, when NULLs are stored in an optional vector. )

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
```diff
@@ -56,79 +55,50 @@ class FrameOfReferenceSegmentIterable : public PointAccessibleSegmentIterable<Fr
...
+      return SegmentPosition<T>{value, is_null, _chunk_offset};
     }
 
    private:
-    ReferenceFrameIterator _block_minimum_it;
-    OffsetValueIteratorT _offset_value_it;
-    NullValueIterator _null_value_it;
-    size_t _index_within_frame;
+    const pmr_vector<T>* _block_minima;
+    const pmr_vector<bool>* _null_values;
```
mrks: Why is this a pointer, not a reference?

Bouncner: Because iterators need to be copyable. And we did it similar in other iterators before. 😁
Am happy to consider alternatives.

mrks: Shouldn't that be possible with references, too? Unless I'm mistaken, references are only a problem if you want to assign the class (which is not possible right now either because the pointer is const)

Bouncner: Afaik, the copy assignment does not work with references. At that's what is being done a lot with iterators. What might work is using a `std::reference_wrapper`.

mrks: Let me have a look at this tomorrow or so

Bouncner: Benchmark is running...

mrks: Ok, we won't get around having something assignable here. reference_wrappers are not really better just because they are not a pointer.

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
```diff
@@ -56,79 +55,50 @@ class FrameOfReferenceSegmentIterable : public PointAccessibleSegmentIterable<Fr
   const FrameOfReferenceSegment<T>& _segment;
 
  private:
-  template <typename OffsetValueIteratorT>
-  class Iterator : public BaseSegmentIterator<Iterator<OffsetValueIteratorT>, SegmentPosition<T>> {
+  template <typename OffsetValueDecompressorT>
```
mrks: Just OffsetValueDecompressor for consistency (entire file)?

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
```diff
@@ -36,13 +36,12 @@ class FrameOfReferenceSegmentIterable : public PointAccessibleSegmentIterable<Fr
...
+    resolve_compressed_vector_type(_segment.offset_values(), [&](const auto& offset_values) {
+      using OffsetValueDecompressorT = std::decay_t<decltype(offset_values.create_decompressor())>;
 
       auto begin = PointAccessIterator<OffsetValueDecompressorT>{&_segment.block_minima(), &_segment.null_values(),
-                                                                 std::move(decompressor), position_filter->cbegin(),
-                                                                 position_filter->cbegin()};
+                                                                 offset_values.create_decompressor(),
+                                                                 position_filter->cbegin(), position_filter->cbegin()};
 
       auto end = PointAccessIterator<OffsetValueDecompressorT>{position_filter->cbegin(), position_filter->cend()};
```
mrks: This made it through the review back then, but is actually pretty awful. There should be no special constructor for the "end iterator", because there is no such thing. There is an iterator pointing to the end. Decrementing and dereferencing that iterator should be possible. Here is a diff for a test:

```
diff --git a/src/test/storage/iterables_test.cpp b/src/test/storage/iterables_test.cpp
index 407a7ae3..a45fe51d 100644
--- a/src/test/storage/iterables_test.cpp
+++ b/src/test/storage/iterables_test.cpp
@@ -167,11 +167,42 @@ TEST_P(EncodedSegmentIterablesTest, IteratorWithIterators) {
     using SegmentType = std::decay_t<decltype(segment)>;

     if constexpr (!std::is_same_v<pmr_string, ColumnDataType>) {
-      auto sum = ColumnDataType{0};
-      auto accessed_offsets = std::vector<ChunkOffset>{};
-      const auto functor = SumUpWithIterator<ColumnDataType>{sum, accessed_offsets};
+      // Test summing up all values in the segment or all values referenced by the PosList
+      {
+        auto sum = ColumnDataType{0};
+        auto accessed_offsets = std::vector<ChunkOffset>{};
+        const auto functor = SumUpWithIterator<ColumnDataType>{sum, accessed_offsets};
+
+        const auto iterable = create_iterable_from_segment<ColumnDataType, false /* no type erasure */>(segment);
+        if (with_position_filter) {
+          if constexpr (!std::is_same_v<SegmentType, ReferenceSegment>) {
+            iterable.with_iterators(position_filter, functor);
+          }
+        } else {
+          iterable.with_iterators(functor);
+        }

+        EXPECT_EQ(sum, expected_sum(nullable, with_position_filter));
+        EXPECT_EQ(accessed_offsets, expected_offsets(with_position_filter));
+      }
+    }
+
+    // Next, test that begin and end iterators are compatible (i.e., that iterating from both ends allows us to meet in
+    // the middle)
+    {
       const auto iterable = create_iterable_from_segment<ColumnDataType, false /* no type erasure */>(segment);
+      const auto functor = [&](auto begin, auto end) {
+        if (with_position_filter) {
+          begin += position_filter->size() - 1;
+        } else {
+          begin += base_segment->size() - 1;
+        }
+        --end;
+
+        EXPECT_EQ(begin->value(), end->value());
+        EXPECT_EQ(begin->chunk_offset(), end->chunk_offset());
+      };
+
       if (with_position_filter) {
         if constexpr (!std::is_same_v<SegmentType, ReferenceSegment>) {
           iterable.with_iterators(position_filter, functor);
@@ -179,9 +210,6 @@ TEST_P(EncodedSegmentIterablesTest, IteratorWithIterators) {
       } else {
         iterable.with_iterators(functor);
       }
-
-      EXPECT_EQ(sum, expected_sum(nullable, with_position_filter));
-      EXPECT_EQ(accessed_offsets, expected_offsets(with_position_filter));
     }
   });
 }
```

Currently, this segfaults.

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
```diff
@@ -56,79 +55,50 @@ class FrameOfReferenceSegmentIterable : public PointAccessibleSegmentIterable<Fr
...
-                      NullValueIterator null_value_it, ChunkOffset chunk_offset)
-        : _block_minimum_it{std::move(block_minimum_it)},
-          _offset_value_it{std::move(offset_value_it)},
-          _null_value_it{std::move(null_value_it)},
-          _index_within_frame{0u},
+    explicit Iterator(const pmr_vector<T>* block_minima, const pmr_vector<bool>* null_values,
+                      OffsetValueDecompressorT attribute_decompressor, ChunkOffset chunk_offset)
+        : _block_minima{block_minima},
+          _null_values{null_values},
+          _offset_value_decompressor{std::move(attribute_decompressor)},
```
mrks: Naming mismatch

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
mrks: > So for the containers listed, not only should the iterator returned from end() be decrementable, the decremented iterator should also be dereferencable. (Unless the container is empty, of course. That invokes undefined behavior.)

https://stackoverflow.com/questions/5322104/how-portable-is-end-iterator-decrement

mrks:
(Let me know if you agree)

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
Bouncner: I removed this end iterator now as well. Did not think too much about it before, but yes. It's quite bad. Sorry for that overlook.
When I tried reference wrapper, it directly forbid to use nullptrs which I would see as an advantage of std::reference_wrapper (and there appears to be no runtime costs).

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
Bouncner: I actually kind of like them, because they "just work" and they cost have a cost here. But when you and Herb Sutter (https://herbsutter.com/2020/02/) don't like them, ok.

Bouncner:
All your changes look good to me.

Bouncner:
Merging the master right now. Give me a couple of minutes. Compiling ...

Bouncner:
I don't fully understand the impact of the new PosLists yet. Any reason to remeasure the TPC-H here? Or should that be completely disconnected?

mrks:
> I don't fully understand the impact of the new PosLists yet.

Neither do I. :D
Theoretically, they should currently have only a very limited benefit, as only the validate operator profits. Apparently, they change the alignment or the optimization in some places for the better. I won't argue with a good thing...

> Any reason to remeasure the TPC-H here? Or should that be completely disconnected?

Could be interesting, yes. Hopefully, the performance variations don't go into the other direction here...

Bouncner:
For completeness, Clang on nemea:
```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.6336297988891602 | 295   | 1.6555933952331543 | 299   | +1%        |  0.0000 |
+| TPC-H 02       | 71.76240539550781  | 12918 | 76.88224792480469  | 13839 | +7%        |  0.0000 |
 | TPC-H 03       | 8.242727279663086  | 1484  | 8.334355354309082  | 1501  | +1%        |  0.0000 |
 | TPC-H 04       | 7.155191421508789  | 1288  | 7.450936317443848  | 1342  | +4%        |  0.0000 |
 | TPC-H 05       | 3.9228553771972656 | 707   | 4.037532806396484  | 727   | +3%        |  0.0000 |
 | TPC-H 06       | 264.0427551269531  | 47528 | 265.5869445800781  | 47806 | +1%        |  0.0000 |
 | TPC-H 07       | 9.590498924255371  | 1727  | 9.919282913208008  | 1786  | +3%        |  0.0000 |
+| TPC-H 08       | 6.807681560516357  | 1226  | 7.179723739624023  | 1293  | +5%        |  0.0000 |
 | TPC-H 09       | 1.968816876411438  | 355   | 2.0019874572753906 | 361   | +2%        |  0.0000 |
 | TPC-H 10       | 3.25930118560791   | 587   | 3.3296267986297607 | 600   | +2%        |  0.0004 |
+| TPC-H 11       | 32.00114059448242  | 5761  | 33.86578369140625  | 6096  | +6%        |  0.0000 |
 | TPC-H 12       | 15.863081932067871 | 2856  | 16.32817840576172  | 2940  | +3%        |  0.0000 |
 | TPC-H 13       | 3.06204891204834   | 552   | 3.070197105407715  | 553   | +0%        |  0.0251 |
 | TPC-H 14       | 43.23070526123047  | 7782  | 42.38031005859375  | 7629  | -2%        |  0.0000 |
 | TPC-H 15       | 108.96278381347656 | 19614 | 110.89640808105469 | 19962 | +2%        |  0.0000 |
 | TPC-H 16       | 10.024027824401855 | 1805  | 10.148356437683105 | 1827  | +1%        |  0.0000 |
+| TPC-H 17       | 4.714269161224365  | 849   | 5.004730224609375  | 901   | +6%        |  0.0000 |
 | TPC-H 18       | 1.19893479347229   | 216   | 1.2259604930877686 | 221   | +2%        |  0.0000 |
+| TPC-H 19       | 8.886797904968262  | 1600  | 9.434701919555664  | 1699  | +6%        |  0.0000 |
 | TPC-H 20       | 28.300615310668945 | 5095  | 29.62635612487793  | 5333  | +5%        |  0.0000 |
 | TPC-H 21       | 1.2929213047027588 | 233   | 1.3237056732177734 | 239   | +2%        |  0.0000 |
 | TPC-H 22       | 17.519725799560547 | 3154  | 17.54766082763672  | 3159  | +0%        |  0.0005 |
 | geometric mean |                    |       |                    |       | +3%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

PR #2063, opened 2020-03-28
Improve colors of scripts/plot_operator_breakdown.py

mrks:
After reviewers rightly complained about this, I made another approach. This uses Cubehelix, which gradually changes all of HSV:

https://ifweassume.blogspot.com/2013/05/cubehelix-or-how-i-learned-to-love.html

Additionally, I added some light hatching. To have a look at how this looks in grayscale, I used

```
\gs  -sOutputFile=output.pdf  -sDEVICE=pdfwrite  -sColorConversionStrategy=Gray  -dProcessColorModel=/DeviceGray  -dCompatibilityLevel=1.4  -dNOPAUSE  -dBATCH -dAutoRotatePages=/None operator_breakdown.pdf
```

I won't say which one is the old and which one is the new. If you can't tell, please simply close this PR. ;)

Screenshot 2020-03-28 at 13 29 49

### DIFF
--- a/scripts/plot_operator_breakdown.py
+++ b/scripts/plot_operator_breakdown.py
@@ -11,6 +11,9 @@ import matplotlib.pyplot as plt
 import pandas as pd
 import re
 import sys
+import seaborn as sns
+import matplotlib.colors as mplcolors
+from matplotlib.colors import LinearSegmentedColormap
 
 if(len(sys.argv) != 1):
     exit("Call without arguments in a folder containing *-PQP.svg files")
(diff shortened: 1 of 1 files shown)


Bouncner:
I wouldn’t say “rightly complained” in the year 2020. It’s not like color printers aren’t widely available yet. And here, where we consume a PDF, it’s even less an argument IMO.
However, the old color scheme was ugly. I see that. The new one is a little bit better (not much IMO), but I hate hatching. It’s just ugly.

But anyhow. I won’t veto the change.
I’ll take a look at the code later.

mrks:
Depending on `sys.argv[1] == 'paper'`, I could (1) change the position of the legend and (2) enable/disable hatching.

Bouncner:
> Depending on `sys.argv[1] == 'paper'`, I could (1) change the position of the legend and (2) enable/disable hatching.

Sounds good to me.

PR #2066, opened 2020-03-30
Remove make_(shared|unique)_by_data_type

### LINKED ISSUE #1894: Kill make_(unique|shared)_by_data_type[s]
While working through the material for the seminar, I realized that we have almost completely switched from `make_shared_by_column_type` to `resolve_type`. In src/lib, we have five uses remaining, all of which should be easy to replace with `resolve_type`.

For the seminar, this is not critical, as the general concept of "do something depending on the data type" is the important part, not the actual implementation (which we do not expect students to look at, anyway). For the next year, however, we should remove it.

mrks:
In most places, we now use `resolve_type` to change a run-time data type into a compile-time type. `resolve_type` is more flexible and does not force people to create unintuitive Impl classes.

This PR removes the remaining uses of `make_(shared|unique)_by_data_type` in order to reduce the amounts of meta magic in the code basis.

fixes #1894.


src/test/storage/segment_accessor_test.cpp
```diff
@@ -14,25 +14,28 @@ namespace opossum {
...
-    vc_str->append(NULL_VALUE);
-
-    dc_int = ChunkEncoder::encode_segment(vc_int, DataType::Int, SegmentEncodingSpec{EncodingType::Dictionary});
-    dc_str = ChunkEncoder::encode_segment(vc_str, DataType::String, SegmentEncodingSpec{EncodingType::Dictionary});
-
-    tbl = std::make_shared<Table>(TableColumnDefinitions{TableColumnDefinition{"vc_int", DataType::Int, false},
-                                                         TableColumnDefinition{"dc_str", DataType::String, false}},
-                                  TableType::Data);
-    tbl->append_chunk({vc_int, dc_str});
+    val_seg_int = std::make_shared<ValueSegment<int>>(true);
```
mrks: Not a big fan of this either, but requiring full names in tests maybe goes a bit too far. Just changed it because "vc" (for ValueColumn, the old name) was outdated.

PR #2070, opened 2020-04-01
NULLs as std::optional in FoR segments

Bouncner:
This PR stores the bool vector for NULL values as an `std::optional<pmr_vector<bool>>`. The performance slightly degrades for GCC (I'd say it's noise because I've seen slight different numbers before) and improves for Clang. I don't think this improvement is too surprising, because the branch should be well predictable and even jumps in a small bool vector don't come for free.
But I am still suprised that the improvement is so "large" for Clang.

The storage overhead is strongly depending on the actual data (unsurprisingly). For TPC-H with all integers as FoR-encoded and FSBA we save a little less than 5%, for SIMDBP128 it's 7%.

It's actually quite significant for sorted columns for which FoR-SIMDBP128 shines (project with @krichly). Here, the NULL values can make up a much larger share. 

GCC:
```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.2217557430267334 | 220   | 1.2203154563903809 | 220   | -0%        |  0.5528 |
 | TPC-H 02       | 104.75822448730469 | 18857 | 103.82511138916016 | 18689 | -1%        |  0.0087 |
 | TPC-H 03       | 10.032971382141113 | 1806  | 9.988512992858887  | 1798  | -0%        |  0.0000 |
 | TPC-H 04       | 8.794621467590332  | 1584  | 8.741357803344727  | 1574  | -1%        |  0.0000 |
 | TPC-H 05       | 4.594256401062012  | 827   | 4.574202060699463  | 824   | -0%        |  0.0007 |
 | TPC-H 06       | 275.9691467285156  | 49675 | 276.3359069824219  | 49741 | +0%        |  0.0592 |
 | TPC-H 07       | 11.866096496582031 | 2136  | 11.738624572753906 | 2113  | -1%        |  0.0160 |
 | TPC-H 08       | 8.620914459228516  | 1552  | 8.584380149841309  | 1546  | -0%        |  0.1299 |
 | TPC-H 09       | 2.00933837890625   | 362   | 2.0243027210235596 | 365   | +1%        |  0.0001 |
 | TPC-H 10       | 3.47220778465271   | 625   | 3.4494662284851074 | 621   | -1%        |  0.2344 |
 | TPC-H 11       | 43.21009826660156  | 7778  | 41.62124252319336  | 7492  | -4%        |  0.0000 |
 | TPC-H 12       | 18.298826217651367 | 3294  | 18.131155014038086 | 3264  | -1%        |  0.0000 |
 | TPC-H 13       | 2.849353313446045  | 513   | 2.8926682472229004 | 521   | +2%        |  0.0000 |
 | TPC-H 14       | 44.88877868652344  | 8080  | 44.72051239013672  | 8050  | -0%        |  0.0000 |
 | TPC-H 15       | 118.3189926147461  | 21298 | 115.08967590332031 | 20717 | -3%        |  0.0000 |
 | TPC-H 16       | 10.21674633026123  | 1840  | 9.788127899169922  | 1762  | -4%        |  0.0000 |
 | TPC-H 17       | 6.26512336730957   | 1128  | 6.230569362640381  | 1122  | -1%        |  0.0001 |
 | TPC-H 18       | 1.356587529182434  | 245   | 1.3056942224502563 | 236   | -4%        |  0.0000 |
 | TPC-H 19       | 11.408391952514648 | 2054  | 11.357592582702637 | 2045  | -0%        |  0.1338 |
 | TPC-H 20       | 35.81550979614258  | 6447  | 35.37551498413086  | 6368  | -1%        |  0.0000 |
 | TPC-H 21       | 1.5605413913726807 | 281   | 1.535070538520813  | 277   | -2%        |  0.0000 |
 | TPC-H 22       | 20.550830841064453 | 3700  | 20.43101692199707  | 3678  | -1%        |  0.0000 |
 | geometric mean |                    |       |                    |       | -1%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

Clang:
```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.6562831401824951 | 299   | 1.6364372968673706 | 295   | -1%        |  0.0000 |
 | TPC-H 02       | 75.24880981445312  | 13545 | 77.59046936035156  | 13967 | +3%        |  0.0000 |
+| TPC-H 03       | 8.131805419921875  | 1464  | 8.540794372558594  | 1538  | +5%        |  0.0000 |
 | TPC-H 04       | 7.198065280914307  | 1296  | 7.52252721786499   | 1355  | +5%        |  0.0000 |
 | TPC-H 05       | 4.02791166305542   | 726   | 4.069015026092529  | 733   | +1%        |  0.0004 |
 | TPC-H 06       | 264.59124755859375 | 47627 | 271.5428466796875  | 48878 | +3%        |  0.0000 |
+| TPC-H 07       | 9.52156925201416   | 1715  | 10.021245002746582 | 1804  | +5%        |  0.0000 |
+| TPC-H 08       | 6.818519115447998  | 1228  | 7.180807113647461  | 1293  | +5%        |  0.0000 |
 | TPC-H 09       | 1.9783483743667603 | 357   | 2.0362472534179688 | 367   | +3%        |  0.0000 |
 | TPC-H 10       | 3.2640419006347656 | 588   | 3.3829407691955566 | 610   | +4%        |  0.0000 |
+| TPC-H 11       | 32.22758483886719  | 5801  | 34.02743911743164  | 6125  | +6%        |  0.0000 |
 | TPC-H 12       | 16.15580177307129  | 2909  | 16.57756805419922  | 2984  | +3%        |  0.0000 |
 | TPC-H 13       | 3.122743606567383  | 563   | 3.114607810974121  | 561   | -0%        |  0.0338 |
 | TPC-H 14       | 43.170448303222656 | 7771  | 43.389732360839844 | 7811  | +1%        |  0.0000 |
 | TPC-H 15       | 112.56536102294922 | 20262 | 112.79100036621094 | 20303 | +0%        |  0.0000 |
 | TPC-H 16       | 10.10153865814209  | 1819  | 10.312785148620605 | 1857  | +2%        |  0.0000 |
+| TPC-H 17       | 4.73165225982666   | 852   | 4.988125801086426  | 898   | +5%        |  0.0000 |
+| TPC-H 18       | 1.174332857131958  | 212   | 1.2543879747390747 | 226   | +7%        |  0.0000 |
+| TPC-H 19       | 8.963787078857422  | 1614  | 9.422157287597656  | 1696  | +5%        |  0.0000 |
 | TPC-H 20       | 28.558895111083984 | 5141  | 29.897592544555664 | 5382  | +5%        |  0.0000 |
 | TPC-H 21       | 1.3034600019454956 | 235   | 1.3487770557403564 | 243   | +3%        |  0.0000 |
 | TPC-H 22       | 17.699033737182617 | 3186  | 18.09370994567871  | 3257  | +2%        |  0.0000 |
 | geometric mean |                    |       |                    |       | +3%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```


src/lib/import_export/binary/binary_writer.hpp
```diff
@@ -192,7 +195,7 @@ class BinaryWriter {
    *
    * ¹: This field is written if the number of blocks is greater than 1. Otherwise "Last Block Size" contains the
    *    size of the single block or 0 if there are no blocks
-   * ²: This field is only written if NULL value size is not 0
+   * ²: This field is only written if NULL values size is not 0
```
mrks, line 198: Null values'?

src/lib/storage/frame_of_reference_segment.cpp
```diff
@@ -52,22 +53,30 @@ template <typename T, typename U>
 std::shared_ptr<BaseSegment> FrameOfReferenceSegment<T, U>::copy_using_allocator(
```
mrks, line 53: We probably should add another test parameter to `SegmentsUsingAllocatorsTest` to test segments with and without null values.

src/lib/storage/frame_of_reference_segment.hpp
```diff
@@ -70,7 +70,7 @@ class FrameOfReferenceSegment : public BaseEncodedSegment {
 
   std::optional<T> get_typed_value(const ChunkOffset chunk_offset) const {
     // performance critical - not in cpp to help with inlining
-    if (_null_values[chunk_offset]) {
+    if (_null_values && (*_null_values)[chunk_offset]) {
```
mrks, line 73: Theoretically, we could also add non-null iterator support to this now. Have a look at `ValueSegmentIterable::_on_with_iterators`. Just a suggestion, not a change request. We should also consider if the compile time costs justify the potential benefits.

src/lib/storage/frame_of_reference_segment/frame_of_reference_encoder.hpp
```diff
@@ -106,6 +109,10 @@ class FrameOfReferenceEncoder : public SegmentEncoder<FrameOfReferenceEncoder> {
 
     auto compressed_offset_values = compress_vector(offset_values, vector_compression_type(), allocator, {max_offset});
 
+    if (!segment_contains_null_values) {
```
mrks, line 112: Minimal code style thing, again not a must have: You could also switch the two cases and get rid of the negation. Personally, I would consider the two cases to be on eye-level, so this wouldn't be a guard clause type of thing for me.

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
```diff
@@ -142,7 +142,7 @@ class FrameOfReferenceSegmentIterable : public PointAccessibleSegmentIterable<Fr
 
    private:
     const pmr_vector<T>* _block_minima;
-    const pmr_vector<bool>* _null_values;
+    const std::optional<pmr_vector<bool>>* _null_values;
```
mrks, line 145: Pointers already have "does not exist" semantics, namely nullptr. Using that would allow us to avoid `**_null_values` above.

src/lib/import_export/binary/binary_writer.hpp
```diff
@@ -158,11 +158,14 @@ class BinaryWriter {
    * Number of Blocks            | uint32_t                            | 4
    * Block minima                | T                                   | Number of blocks * sizeof(T)
    * Size                        | uint32_t                            | 4
-   * NULL values                 | vector<bool> (BoolAsByteType)       | size * 1
+   * NULL values size            | uint32_t                            | 4
```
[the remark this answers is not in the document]
Bouncner: Ok, and get the size of the bool vector via the segment header?

src/lib/storage/frame_of_reference_segment.hpp
Bouncner: To be honest, that looks horrible to me. Shouldn't that be possible with a simple template parameter?
In any case, I would rather only use it for Dictionary and ValueSegments (especially with the increasing TPC-C importance). We just do not use FoR often enough to justify the compile time IMO.

But if you want I can take a look into the two main segments (there should be further improvements possible for value segments as far as I can see) and check if a templated version is possible. But probably only worth it for ValueSegments, right?

mrks: I am not even sure if it has any benefits for value segments. Dictionary segments do not use that optimization, but it would be easy to add. If you are interested in this, you could simply disable the `is_null` check in the two dictionary iterators and run TPC-H.

src/lib/import_export/binary/binary_writer.hpp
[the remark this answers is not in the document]
mrks: Should be the same as "size" in the line before, shouldn't it?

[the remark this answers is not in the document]
Bouncner: Then I'll remove this size. I need to take a deeper look into this. For Dictionary segments, the segmen's row count is sufficient. Should be the same here.

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
Bouncner: Personally, I'd favor std::optional as it clearly states the intend. `nullptr` are bad (IMO). I could measure if `*(_null_values.get())` does have a negative impact if the double star is of concern?

src/lib/storage/frame_of_reference_segment/frame_of_reference_encoder.hpp
Bouncner: Nice catch. Changed.

src/lib/storage/frame_of_reference_segment.hpp
Bouncner: I'll take a look at it in another branch.

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
mrks: Mmmh, I can't find a good canonical answer for that. To me, it opens up more ambiguities. Why is this an `optional<T>*`, i.e., a pointer to an optionally existing value, and not an `optional<T*>`, meaning an optionally existing pointer to a certainly existing value.

Also, having `if (_null_values)` (which always returns true), `if (*_null_values)` (which might return true), and `if ((**_null_values)[x])` (which depends on the data) seems unnecessarily complicated. For funsies, we could even add `if (***_null_values)` to the discussion, which checks if the first value is NULL. ;)

In the great scheme of things, no one will get physically hurt by this, so whatever...

src/lib/import_export/binary/binary_writer.hpp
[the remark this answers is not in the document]
Bouncner: Changed, can you take a look?

src/lib/storage/frame_of_reference_segment.cpp
Bouncner: Modified the tests. Without digging too deep into the allocator tracking I am not really certain that everything is not tested. Can you take a quick look at it?

src/lib/storage/frame_of_reference_segment/frame_of_reference_segment_iterable.hpp
Bouncner: I see your point, but the current way of `optional<T>*` mirrors exactly what it done to all other members of the segment as well. Also, I think "an optionally existing pointer to a certainly existing value" is more confusing. But that might just be me.

PR #2071, opened 2020-04-03
Add AccessCounters to MetaSegmentsTable

dey4ss:
This PR shows the number of point, sequential, monotonic and random accesses of a segment in MetaSegmentsTable and MetaSegmentsAccurateTable.


mrks:
I am not sure where the best place is to do this, but somewhere we should document where the values in the counters come from. This looks like somethings where, e.g., a change in the encoding algorithm could lead to a different access pattern and this test failing.

dey4ss:
>> I am not sure where the best place is to do this, but somewhere we should document where the values in the counters come from. This looks like somethings where, e.g., a change in the encoding algorithm could lead to a different access pattern and this test failing.

Maybe we could reference the access counters the header files of the meta tables and figure out that they are segment dependent? Additionally, there could also be a comment in the test which states that the test files are expected to become outdated on changes to value segments.

mrks:
The comment in the test seems to be the best we can do. For me, the other comment is more confusing that helpful.

PR #2072, opened 2020-04-03
Make storage manager thread-safe

### LINKED ISSUE #1991: Storage Manager is not thread-safe
The bachelor's project is currently discussing the parallel loading of benchmark tables (e.g., TPC-H) via the network interface. @Bensk1 and me just looked into the storage manager and we are not sure if parallel loading is safe.
Bot appear to be fine (insertion/deletion invalidate iterators, which wouldn't be a problem here), but calling `table_names()` can be UB when the iterator is invalidated in parallel.

Since the storage manager is probably not on any critical path, we could potentially use a simple `lock_guard` solution. Any opinions?

Bensk1:
Fixes #1991. This is particularly useful for parallel table loading via our server.

We might be able to use that functionality to speed up benchmark preparation by adding tables to the StorageManager (and thereby generating statistics) in parallel.


src/lib/storage/storage_manager.cpp
```diff
@@ -19,8 +19,12 @@
 namespace opossum {
 
 void StorageManager::add_table(const std::string& name, std::shared_ptr<Table> table) {
-  Assert(_tables.find(name) == _tables.end(), "A table with the name " + name + " already exists");
-  Assert(_views.find(name) == _views.end(), "Cannot add table " + name + " - a view with the same name already exists");
+  const auto table_iter = _tables.find(name);
```
mrks, line 22: PR description: "Solves" is not a magic word. Changed to "fixes"

src/lib/storage/storage_manager.cpp
```diff
@@ -30,123 +34,149 @@ void StorageManager::add_table(const std::string& name, std::shared_ptr<Table> t
...
   // Create table statistics and chunk pruning statistics for added table.
+
   table->set_table_statistics(TableStatistics::from_table(*table));
   generate_chunk_pruning_statistics(table);
 
-  _tables.emplace(name, std::move(table));
+  if (table_iter == _tables.end())
+    _tables.emplace(name, std::move(table));
+  else
+    _tables[name] = std::move(table);
```
mrks, line 44: Wouldn't this line suffice? If not, please add curly braces

mrks, line 44: Do we have a test that adds, drops, and adds again?

src/lib/storage/storage_manager.cpp
```diff
@@ -30,123 +34,149 @@ void StorageManager::add_table(const std::string& name, std::shared_ptr<Table> t
...
 }
 
 void StorageManager::drop_table(const std::string& name) {
-  const auto num_deleted = _tables.erase(name);
-  Assert(num_deleted == 1, "Error deleting table " + name + ": _erase() returned " + std::to_string(num_deleted) + ".");
+  const auto table_iter = _tables.find(name);
+  Assert(table_iter != _tables.end() && table_iter->second, "Error deleting table. No such table named '" + name + "'");
+
+  // The concurrent_unordered_map does not support concurrency-safe erasure. Thus, we simply replace the table.
```
mrks, line 51: ...reset the table pointer?

src/lib/storage/storage_manager.cpp
```diff
@@ -30,123 +34,149 @@ void StorageManager::add_table(const std::string& name, std::shared_ptr<Table> t
...
 }
 
 std::shared_ptr<Table> StorageManager::get_table(const std::string& name) const {
-  const auto iter = _tables.find(name);
-  Assert(iter != _tables.end(), "No such table named '" + name + "'");
+  const auto table_iter = _tables.find(name);
+  Assert(table_iter != _tables.end() && table_iter->second, "No such table named '" + name + "'");
 
-  return iter->second;
+  return table_iter->second;
```
mrks, line 59: This is not thread-safe.

```
Assert(table_iter != _tables.end(), ...);
auto table = table_iter->second;
Assert(table, ...);
return table;
```

should be better.

In any case, we should update the description of these methods that currently says "not thread-safe". Maybe it is a good idea to limit our guarantees to parallel operations on different tables.

src/lib/storage/storage_manager.cpp
```diff
@@ -30,123 +34,149 @@ void StorageManager::add_table(const std::string& name, std::shared_ptr<Table> t
...
-const std::map<std::string, std::shared_ptr<Table>>& StorageManager::tables() const { return _tables; }
+const tbb::concurrent_unordered_map<std::string, std::shared_ptr<Table>>& StorageManager::tables() const {
+  return _tables;
+}
 
 void StorageManager::add_view(const std::string& name, const std::shared_ptr<LQPView>& view) {
-  std::unique_lock lock(*_view_mutex);
-
-  Assert(_tables.find(name) == _tables.end(),
+  const auto table_iter = _tables.find(name);
```
mrks, line 85: Obviously the points from above apply to views and stored plans accordingly

src/lib/storage/storage_manager.hpp
```diff
@@ -62,14 +64,15 @@ class StorageManager : public Noncopyable {
...
-
-  // The map of views is locked because views are created dynamically, e.g., in TPC-H 15
-  std::map<std::string, std::shared_ptr<LQPView>> _views;
-  mutable std::unique_ptr<std::shared_mutex> _view_mutex = std::make_unique<std::shared_mutex>();
+  // We preallocate maps to prevent costly re-allocation.
+  static constexpr size_t _INITIAL_MAP_SIZE = 100;
 
-  std::map<std::string, std::shared_ptr<PreparedPlan>> _prepared_plans;
+  tbb::concurrent_unordered_map<std::string, std::shared_ptr<Table>> _tables{
+      _INITIAL_MAP_SIZE, tbb::tbb_allocator<std::pair<const std::string, std::shared_ptr<Table>>>()};
```
mrks, line 71: Why do we need the tbb_allocator here? Anything about the map that I am not aware of?

src/lib/storage/storage_manager.cpp
Bensk1: Good point, I don't know why I wanted to reuse the old solution.

Bensk1: Sure: https://github.com/hyrise/hyrise/pull/2072/files#diff-6a6d6b4d1981779e9817b29e0e581c5aR75

src/lib/storage/storage_manager.hpp
Bensk1: To be honest: I used the one from the examples, I thought it might have stronger guarantees. Also, preallocating a map without specifying an allocator is not possible.

mrks: Just saw that this wasn't part of the template parameter, so it was used anyway. I'm surprised to see that the first constructor does not work:

https://software.intel.com/en-us/node/506172

Anyway, `{}` instead of talking about the allocator should be sufficient and keeps people from wondering.

src/lib/storage/storage_manager.hpp
```diff
@@ -23,7 +23,7 @@ class AbstractLQPNode;
 class StorageManager : public Noncopyable {
  public:
   /**
-   * @defgroup Manage Tables, not thread-safe
+   * @defgroup Manage Tables, this is only thread-safe for operations on different tables
```
mrks, line 26: tables with with different names?

src/lib/storage/storage_manager.hpp
Bensk1: I think it is safe to say now that the first constructor does not work if the developer is an idiot.

scripts/analyze_ccache_usage.py
```diff
@@ -28,7 +28,8 @@
 
 if len(files) == 0:
   print("No *.ccache-log files found. Did you compile with ccache and the environment variable CCACHE_DEBUG=1?")
-  sys.exit(1)
+  # ToDo(anyone): change back to `sys.exit(1)` as soon as our ccache problem is resolved
```
Bensk1, line 31: @mrks I added that because otherwise the CI fails. We could also remove the call of the analyzer script from the Jenkinsfile as a hot fix. Any preference?

mrks: I found a way to disable ccache without changing any file in the Hyrise repo. Could you please remove it again?

Bensk1: Done.

scripts/test/hyriseConsole_test.py
```diff
@@ -15,7 +15,8 @@ def initialize():
 		sys.exit(1)
 
 	build_dir = sys.argv[1]
-	console = pexpect.spawn(build_dir + "/hyriseConsole", timeout=60, dimensions=(200, 64))
+        console = pexpect.spawn("catchsegv " + build_dir + "/hyriseConsole", timeout=60, dimensions=(200, 64))
+        console.logfile_read = sys.stdout
```
mrks, line 19: Please make sure to undo this change before continuing, as it prevents the test from failing.

Bensk1:
I am looking into that.

Issue #2032, opened 2020-02-25
CMake: third-party includes should not be set in src/CMakeLists.txt


mrks:
According to my (incomplete) understanding of CMake, this is not the correct way to do things:

https://github.com/hyrise/hyrise/blob/64ed84e275501bfc6023657ad50cba9fc4cbd29f/src/CMakeLists.txt#L66-L70

Instead, the libraries should be defined as such and added as a dependency.

PR #2073, opened 2020-04-03
Introduce staged performance counters

Bouncner:
This PR mainly introduces staged performance counters. These extend `OperatorPerformanceData` structs to simply allow for storing the runtime of certain stages. The main reason has been for calibration and learning cases, but in the meanwhile I found them to be extremely helpful when analysing operator performance (visualization is not part of this PR, but will follow).

With this PR, one can analyze in which stages take how much time in the JoinHash. Further, and very often of interest is the trackng of segments scans on sorted segments and the number of skipped segments during execution (i.e., "dynamic" dictionary pruning, cf. Dreseler et al., PVLDB 17).

Further, this PR adds the third party module `enum_magic` which provides of ton of enum helper methods. An open todo is to get rid of some enum hacks (e.g., storing a `count` member to the get the size of enums) and use `enum_magic` where appropriate.
I ran TPC-H benchmarks with very small chunk counts and could not observe any performance degradations.

- [ ] Use `enum_magic` where appropriate


src/lib/visualization/pqp_visualizer.cpp
```diff
@@ -139,15 +139,15 @@ void PQPVisualizer::_build_dataflow(const std::shared_ptr<const AbstractOperator
...
     std::stringstream stream;
-    stream << std::to_string(performance_data.output_row_count) + " row(s)/";
-    stream << std::to_string(performance_data.output_chunk_count) + " chunk(s)";
+    stream << std::to_string(performance_data->output_row_count) + " row(s)/";
+    stream << std::to_string(performance_data->output_chunk_count) + " chunk(s)";
     info.label = stream.str();
   }
 
-  info.pen_width = performance_data.output_row_count;
+  info.pen_width = performance_data->output_row_count;
```
mrks, line 150: Didn't look at the PR yet, just something that came up when I looked at SF=10 graphs and was counting the number of digits in `59986052 row(s)`. Can we add thousands separators? Should be as easy as to throw `std::numpunct` on it.

Bouncner: That’s already a TODO in the following PQP PR. 
Numpunct ain’t that straightforward though because afaik it requires a locale (I think the default C locale might be usable, but I need to take a deeper look into it).

src/lib/operators/operator_performance_data.hpp
```diff
@@ -26,6 +29,18 @@ struct OperatorPerformanceData : public Noncopyable {
                                 DescriptionMode description_mode = DescriptionMode::SingleLine) const;
 };
 
+struct StagedOperatorPerformanceData : public OperatorPerformanceData {
+  StagedOperatorPerformanceData() : OperatorPerformanceData{}, stage_runtimes{10} {}
```
mrks, line 33: Won't this silently blow up if an operator has more than 10 steps?

mrks, line 33: For some reason, something about "stage" doesn't feel right. The dictsch.. dick... wordbook tells me that it is a proper use, but personally, I would prefer "step". If you like stage better, that's fine. If you don't care, maybe we can change it?

src/lib/operators/operator_performance_data.hpp
```diff
@@ -26,6 +29,18 @@ struct OperatorPerformanceData : public Noncopyable {
                                 DescriptionMode description_mode = DescriptionMode::SingleLine) const;
 };
 
+struct StagedOperatorPerformanceData : public OperatorPerformanceData {
```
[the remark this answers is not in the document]
mrks: (This also avoids exposing `OperatorPerformanceData<Stages>` outside of the cpp, which should be good for compile times)

src/CMakeLists.txt
```diff
@@ -70,6 +70,7 @@ include_directories(
     ${PROJECT_SOURCE_DIR}/third_party/cpp-btree
     ${PROJECT_SOURCE_DIR}/third_party/cqf/include
     ${PROJECT_SOURCE_DIR}/third_party/cxxopts/include
+    ${PROJECT_SOURCE_DIR}/third_party/magic_enum/include
```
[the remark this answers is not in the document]
Bouncner: Hmm, can anybody help me here? Tried to do it the same way as `nlohmann_json` is included, but that does not work.

src/lib/operators/aggregate_hash.cpp
```diff
@@ -14,6 +14,7 @@
 #include "constant_mappings.hpp"
 #include "expression/pqp_column_expression.hpp"
 #include "hyrise.hpp"
+#include "magic_enum.hpp"
```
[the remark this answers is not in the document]
Bouncner: Done in d1ca0c3

src/lib/operators/operator_performance_data.hpp
Bouncner: Ah, that's why you use an array in your example? Would that crash during compilation?

Bouncner: I am fine with steps as well. The only thing I can think about being slighly problematic with steps is that steps are one after another in my head. 
In several joins, the "steps" are running in parallel.

What about phase?

Bouncner: Hmm ... then again, parallel steps doesn't sound bad to me.

mrks: If you made 10 a constexpr, you could also do the `static_assert` for the vector. I am using an array so that we don't have a heap allocation for a known, small number of values.

src/CMakeLists.txt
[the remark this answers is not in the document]
mrks: Did you add the dependency to hyrise_impl (src/lib/CMakeLists.txt)?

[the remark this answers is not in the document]
Bouncner: Something like 
```
diff --git a/src/lib/CMakeLists.txt b/src/lib/CMakeLists.txt
index 06786723..9dd975d8 100644
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -703,6 +703,7 @@ set_source_files_properties(
 # -rdynamic tells the linker to export the library's symbols so that plugins can use them.
 target_link_libraries(hyrise_impl PUBLIC ${LIBRARIES} INTERFACE custom_jemalloc -rdynamic)
 target_link_libraries_system(hyrise_impl nlohmann_json::nlohmann_json)
+target_link_libraries_system(hyrise_impl magic_enum::magic_enum)
 target_include_directories(hyrise_impl PUBLIC ${CMAKE_BINARY_DIR})
```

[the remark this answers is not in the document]
Bouncner: `target_link_libraries(hyrise_impl PUBLIC ${LIBRARIES} INTERFACE custom_jemalloc magic_enum -rdynamic)` works, but I thought this would be a system library (don't ask me why, just thought it's a lot like the json library looking where it will eventually be used).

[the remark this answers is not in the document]
Bouncner: I am stupid. I think I slowly start to understand cmake a tiny bit.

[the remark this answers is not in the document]
mrks: Have you tried adding magic_enum to the LIBRARIES variable?

src/lib/operators/operator_performance_data.hpp
Bouncner: What do you mean by static assert?

mrks:
```cpp
static_assert(magic_enum::enum_count<Stages>() < sizeof(stages), "Too many stages");
```

src/CMakeLists.txt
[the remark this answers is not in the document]
Bouncner: This only works if the hpp file is not within a sub directory (works for uninitialized_vector, not for others, I guess that's the reason why their include directories are also listed in `src/CMakeLists.txt`.
I am trying to find a way to make CMake search the sub directories as well, but this whole CMake thing is absolutely ...a

src/lib/operators/operator_performance_data.hpp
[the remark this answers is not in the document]
Bouncner: Have to see how that works out with your CPP comment (have to implement it to understand what you mean I guess), but that looks quite nice! The only thing is the naming. I would like to keep the stage/step thing part of the name as it makes the differences clearer. But a `SteppedOperatorPerformanceData` struct does not sound good to me. Ideas?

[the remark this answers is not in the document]
mrks: > I would like to keep the stage/step thing part of the name as it makes the differences clearer.

What differences? One of the reasons I tried a second approach was because I didn't like the differentiation between two different types of performance data objects.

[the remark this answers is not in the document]
Bouncner: I need to implement it and have a third look at it, I guess.

src/CMakeLists.txt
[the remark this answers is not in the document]
mrks: Having the same issue with a different thing right now. Let's deal with that in #2032.

Bouncner:
I need to make too many changes for which I won't have the time this week. Reopening when I have addressed the issues brought up.

src/lib/operators/operator_performance_data.hpp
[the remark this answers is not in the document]
Bouncner: I am loving it.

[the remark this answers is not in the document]
mrks: One more thought: Couldn't we move the array into `OperatorPerformanceData`? That way, we could (1) use the size of the enum instead of randomly picking 10 and (2) add a `set(Stages stage, uint64_t duration)` method, making the static_cast unnecessary.

src/lib/operators/join_hash.cpp
```diff
@@ -346,21 +363,23 @@ class JoinHash::JoinHashImpl : public AbstractJoinOperatorImpl {
...
-      // Materialize probe column.
+      Timer timer_probe_column;
       if (keep_nulls_probe_column) {
         materialized_probe_column = materialize_input<ProbeColumnType, HashedType, true>(
             _probe_input_table, _column_ids.second, histograms_probe_column, _radix_bits);
       } else {
         materialized_probe_column = materialize_input<ProbeColumnType, HashedType, false>(
             _probe_input_table, _column_ids.second, histograms_probe_column, _radix_bits);
       }
+      clock_materialization_probe_column = timer_probe_column.lap();
```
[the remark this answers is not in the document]
Bouncner: Redone.

src/lib/operators/operator_performance_data.hpp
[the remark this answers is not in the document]
Bouncner: I am not able to do 1.) without further help :D, but the second part makes sense anyways. Will do.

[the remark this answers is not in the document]
mrks:
```
    enum class NoStages {
        Invalid;  // Needed by magic_enum for enum_count
    };
```

Bouncner:
@mweisgut : as you are currently working a lot on the index joins, could you have a quick look at the index join changes?

Bouncner:
@mrks : thanks a lot once again for teaching me C++. Much better know. I am going to self-review with a clear mind tomorrow and remove the WIP in case I am satisfied.

src/lib/operators/join_hash.cpp
```diff
@@ -268,7 +277,7 @@ class JoinHash::JoinHashImpl : public AbstractJoinOperatorImpl {
                                          _mode == JoinMode::AntiNullAsTrue || _mode == JoinMode::AntiNullAsFalse;
 
     // Containers used to store histograms for (potentially subsequent) radix
-    // partitioning phase (in cases _radix_bits > 0). Created during materialization phase.
+    // partitioning stage (in cases _radix_bits > 0). Created during materialization stage.
```
mweisgut, line 280: I see that either "phase" or "stage" should be used consistently, but just out of interest: Is there any particular reason why "stage" should be preferred over "phase"?

mweisgut:
> @mweisgut : as you are currently working a lot on the index joins, could you have a quick look at the index join changes?

JoinIndex changes LGTM

src/lib/operators/join_hash.cpp
Bouncner: Nope. That's simply how the join grew over time. Will change them all to steps. I am fine with Markus's idea of using steps and the hash join was the initial operator why we first thought about such "phases/stages/steps". It's also called `join_hash_steps.hpp`.

Anyways, thanks for pointing it out!

src/lib/operators/aggregate_hash.cpp
```diff
@@ -632,8 +638,7 @@ std::shared_ptr<const Table> AggregateHash::_on_execute() {
       _aggregate<std::vector<AggregateKeyEntry>>();
       break;
   }
-
-  const auto& input_table = left_input_table();
+  step_performance_data.set_step_runtime(OperatorSteps::Aggregating, timer.lap());
```
mrks, line 641: Separate this into the generation of group IDs and the actual aggregation?

src/lib/operators/join_hash.cpp
```diff
@@ -337,15 +347,16 @@ class JoinHash::JoinHashImpl : public AbstractJoinOperatorImpl {
           _probe_input_table, _column_ids.second, histograms_probe_column, _radix_bits, probe_side_bloom_filter,
           build_side_bloom_filter);
     }
+    _performance.set_step_runtime(OperatorSteps::Materialization, timer_materialization.lap());
```
mrks, line 350: I wonder if we want to add a `timer.finish()` method, which would not allow any more calls to lap or finish. Otherwise, `timer_materialization` may falsely be reused. It also gives us some limited thread-safety.

src/lib/operators/join_hash.cpp
```diff
@@ -403,24 +416,30 @@ class JoinHash::JoinHashImpl : public AbstractJoinOperatorImpl {
...
+     *   detecting a NULL value on the build side there.
+     */
     if (_mode == JoinMode::AntiNullAsTrue) {
       for (const auto& build_side_partition : radix_build_column) {
         for (const auto null_value : build_side_partition.null_values) {
           if (null_value) {
-            return _join_hash._build_output_table({});
+            Timer timer_output_writing;
+            const auto result = _join_hash._build_output_table({});
+            _performance.set_step_runtime(OperatorSteps::OutputWriting, timer_output_writing.lap());
```
mrks, line 434: Aren't you overwriting the value here?

src/lib/operators/operator_performance_data.hpp
```diff
@@ -21,11 +27,61 @@ struct OperatorPerformanceData : public Noncopyable {
...
+    }
+  }
+
+  std::chrono::nanoseconds get_step_runtime(const Steps step) const {
+    DebugAssert(magic_enum::enum_integer(step) < magic_enum::enum_count<Steps>(), "Step index is too large.");
+    return step_runtimes[static_cast<size_t>(step)];
+  }
+
+  void set_step_runtime(const Steps step, const std::chrono::nanoseconds duration) {
+    step_runtimes[static_cast<size_t>(step)] = duration;
```
mrks, line 79: Assert for range and `== 0` before? Would have caught what I consider a bug above

src/lib/operators/operator_performance_data.hpp
```diff
@@ -21,11 +27,61 @@ struct OperatorPerformanceData : public Noncopyable {
...
+    if (!has_output) {
+      stream << "Executed, but no output.";
+      return;
+    }
+
+    stream << "Output: " << output_row_count << " row" << (output_row_count > 1 ? "s" : "") << " in "
+           << output_chunk_count << " chunk" << (output_chunk_count > 1 ? "s" : "") << ", "
+           << format_duration(std::chrono::duration_cast<std::chrono::nanoseconds>(walltime)) << ".";
+
+    if constexpr (!std::is_same_v<Steps, NoSteps>) {
```
mrks, line 58: guard clause?

src/lib/operators/operator_performance_data.hpp
```diff
@@ -21,11 +27,61 @@ struct OperatorPerformanceData : public Noncopyable {
...
+      static_assert(magic_enum::enum_count<Steps>() <= sizeof(step_runtimes), "Too many steps.");
+      stream << (description_mode == DescriptionMode::SingleLine ? " " : "\n") << "Operator step runtimes:"
```
mrks, line 60: Store `(description_mode == DescriptionMode::SingleLine ? " " : "\n")` in variable?

src/lib/operators/operator_performance_data.hpp
```diff
@@ -21,11 +27,61 @@ struct OperatorPerformanceData : public Noncopyable {
...
+        }
+        stream << " " << magic_enum::enum_name(static_cast<Steps>(step_index)) << " "
+               << format_duration(step_runtimes[step_index]);
+      }
+      stream << ".";
+    }
+  }
+
+  std::chrono::nanoseconds get_step_runtime(const Steps step) const {
+    DebugAssert(magic_enum::enum_integer(step) < magic_enum::enum_count<Steps>(), "Step index is too large.");
```
mrks, line 74: "Invalid step"?

src/lib/operators/table_scan.hpp
```diff
@@ -48,6 +48,28 @@ class TableScan : public AbstractReadOnlyOperator {
    */
   std::vector<ChunkID> excluded_chunk_ids;
 
+  struct TableScanPerformanceData : public OperatorPerformanceData<AbstractOperatorPerformanceData::NoSteps> {
```
mrks, line 51: Naming: As this is already part of the TableScan class, just "PerformanceData"?

src/lib/operators/table_scan.hpp
```diff
@@ -48,6 +48,28 @@ class TableScan : public AbstractReadOnlyOperator {
...
+    size_t chunk_scans_skipped{0};
+    size_t chunk_scans_sorted{0};
+
+    void output_to_stream(std::ostream& stream, DescriptionMode description_mode) const {
+      const auto separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
```
mrks, line 56: Feel free to spin this off into a new issue: It feels like our lives would be easier if we simply passed in the separator.

src/lib/operators/table_scan.hpp
```diff
@@ -48,6 +48,28 @@ class TableScan : public AbstractReadOnlyOperator {
...
+      if (chunk_scans_skipped > 0 || chunk_scans_sorted > 0) {
```
mrks, line 57: Guard clause

src/lib/operators/table_scan.hpp
```diff
@@ -48,6 +48,28 @@ class TableScan : public AbstractReadOnlyOperator {
...
+        stream << separator << "Chunks: ";
+        if (chunk_scans_skipped > 0) {
+          stream << chunk_scans_skipped << " skipped";
+        }
+        if (chunk_scans_skipped > 0 && chunk_scans_sorted > 0) {
+          stream << ", ";
+        }
+        if (chunk_scans_sorted > 0) {
+          stream << chunk_scans_sorted << " scanned sorted";
```
mrks, line 66: "scanned using binary search", even if it is a bit clunky?

src/lib/operators/table_scan/abstract_table_scan_impl.hpp
```diff
@@ -25,6 +27,9 @@ class AbstractTableScanImpl {
 
   virtual std::shared_ptr<RowIDPosList> scan_chunk(ChunkID chunk_id) const = 0;
 
+  mutable std::atomic<size_t> chunk_scans_skipped{0};
+  mutable std::atomic<size_t> chunk_scans_sorted{0};
```
mrks, line 31: Why does this have to be mutable?

src/test/lib/operators/operator_performance_data_test.cpp
```diff
@@ -1,27 +1,307 @@
...
+TEST_F(OperatorPerformanceDataTest, TableScanPerformanceData) {
+  const TableColumnDefinitions column_definitions{{"a", DataType::Int, false}};
+  auto table = std::make_shared<Table>(column_definitions, TableType::Data, 2);
+  table->append({1});
+  table->append({2});
+  table->append({2});
+  table->append({3});
+
+  // Finalize last chunk
+  table->get_chunk(ChunkID{1})->finalize();
```
mrks, line 66: Redundant comment. You could also use `get_last_chunk` to make your intentions clearer

mrks:
> mrks self-requested a review 29 seconds ago

Wrong button. I'm done with my review.

src/lib/operators/table_scan.hpp
Bouncner: Well, not necessarily. If you look into the PQP tooltip, where we use a multiline output of the OperatorPerformanceData, there is a difference (not necessarily, but anything else is major ugly) between a simple separator and printing list items.

mrks: ok

src/lib/operators/join_hash.cpp
Bouncner: If I don't get this code wrong, it's returning directly afterwards. So it's not executed a second time.

mrks: Yepp, my fault. I'd still add an assertion that prevents overriding the stored step runtime.

src/lib/operators/aggregate_hash.cpp
Bouncner: Good point. Added a new Step `GroupByKeyPartitioning`.

src/lib/operators/join_hash.cpp
Bouncner: I see your point but I am  not sure if isn't a bit too much overhead. It's not a lot added, but I guess the idea of the Timers is to allow timing with the absolute minimum overhead possible.

src/lib/operators/operator_performance_data.hpp
Bouncner: I will reread the again, but I didn't like any solutions that stored separator as we have "different separators" here (it can be `, ` vs `\n` or simply ` ` vs. `\n` or `, ` vs. `\n ` for indented lists).

src/lib/operators/table_scan.hpp
Bouncner: I would be fine with that name, but we would have several structs with the same name in various operators which I didn't like.
Your call.

mrks: As their use outside of the class requires the class name as a qualifier, I don't see an issue here. No strong feelings though.

src/lib/operators/table_scan/abstract_table_scan_impl.hpp
Bouncner: `../src/lib/operators/table_scan/column_between_table_scan_impl.cpp:52:9: error: cannot increment value of type 'const std::atomic<size_t>' (aka 'const atomic<unsigned long>')`

mrks: Why is the Impl const though?

src/lib/operators/table_scan.hpp
Bouncner: Done for `IndexJoinPerformanceData` and `TableScanPerformanceData`.

src/lib/operators/table_scan/abstract_table_scan_impl.hpp
Bouncner: It isn't. It's the const methods of the impls as far as I understand.

Bouncner:
Most parts addressed. Talked to @aloeser who saw significant differences in the hash join's materialization of the two join parts. Thus split the materialization part into build and probe side materialization.

src/lib/operators/table_scan/abstract_table_scan_impl.hpp
mrks: I don't feel strongly about this, but I'd rather remove the consts than add mutables. Resolve as you see fit.

Bouncner:
I removed the `mutable` and `const`s. You're absolutely right, it's cleaner.

Further, I moved the partitioning in `_aggregate()` into a separate function to avoid a NOLINT for too long functions. For this method, I think it's actually really cleaner this way.

PR #2078, opened 2020-04-07
Functional Dependencies & Rewrite of DependentGroupByReductionRule

julianmenzler:
This PR adds basic functional dependency (FD) support to Hyrise. 
FDs are useful for various optimizations purposes. 

As a start, the `DependentGroupByReductionRule` has been rewritten to benefit from the new FD functionality. The code has not only become much more compact. The code also no longer requires access to the `StorageManager` and `TableConstraintDefinition` objects, which has not been ideal in the first place.

### What is a FD?
A FD consists out of two sets of column expressions:
The left column set (determinants) unambigiously identifies the right column set (dependents):
```
{Left} => {Right}
```

**Example A:** Think of a table with three columns: (Semester, CourseID, Lecturer)
The primary key is defined across the first two columns, which leads to the following FD:
``` 
{Semester, CourseID} => {Lecturer}
```

**Example B:** Think of a table with four columns: (ISBN, Genre, Author, Author-Nationality)
The primary key {ISBN} identifies all other columns. Therefore, we get the following FDs:
```
{ISBN} => {Author}
{ISBN} => {Genre}
{ISBN} => {Author}
{ISBN} => {Author-Nationality}
```
Furthermore, knowing an author, we also know his nationality. Hence, we can specify another FD that applies:
```
{Author} => {Author-Nationality}
```

### How are FDs accessed within Hyrise?
FDs can be accessed from within the LQP tree. Every LQPNode inherits the following interface, to return FDs that are valid for itself.
```cpp
std::vector<FunctionalDependency> functional_dependencies() const;
```

The interface of FDs is specified in `functional_dependency.hpp`, which includes the following constructor:
```cpp 
FunctionalDependency(ExpressionUnorderedSet init_determinants, ExpressionUnorderedSet init_dependents);
```

### Where are FDs constructed?
So far, FDs are constructed from unique constraints inside of StoredTableNodes. 
Some forwarding logic, implemented in `AbstractLQPNode`, passes the FDs up in the LQP tree. It ensures, that only valid FDs get forwarded. 

In our implementation, FDs are considered invalid when any of their determinant columns are/become nullable (think of outer joins for example). 
Null-values in conjunction with FDs are not trivial. For more reference, see the following paper: https://arxiv.org/abs/1404.4963

In the future, we may want to create additional FDs based on knowledge like for instance, foreign keys.


src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -245,6 +245,62 @@ bool AbstractLQPNode::is_column_nullable(const ColumnID column_id) const {
...
+  }
+
+  auto fds_in = std::vector<FunctionalDependency>();
+  if (fds_right.empty()) {
+    fds_in = fds_left;
+  } else {
+    // Remove duplicate FDs that might result from e.g. self-joins
+    for (const auto& fd_right : fds_right) {
+      bool duplicate = std::any_of(fds_left.begin(), fds_left.end(), [&fd_right](const auto& fd_left) {
+        return (fd_left.first.size() == fd_right.first.size() && fd_left.second.size() == fd_right.second.size() &&
```
Bouncner, line 266: std::tie?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -245,6 +245,62 @@ bool AbstractLQPNode::is_column_nullable(const ColumnID column_id) const {
...
+    auto fd_expressions = fd.first;
+    fd_expressions.insert(fd.second.begin(), fd.second.end());
+
+    if (std::any_of(fd_expressions.cbegin(), fd_expressions.cend(),
+                    [this, &node_column_expressions_set](const auto& expression) {
+                      // Check for nullability, if possible.
+                      return node_column_expressions_set.contains(expression) && expression->is_nullable_on_lqp(*this);
+                    })) {
+      continue;
+    } else {
```
Bouncner, line 295: Looks like the `else {}` can be removed since the if case continues.

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -92,6 +92,48 @@ bool StoredTableNode::is_column_nullable(const ColumnID column_id) const {
   return table->column_is_nullable(column_id);
 }
 
+std::vector<FunctionalDependency> StoredTableNode::functional_dependencies() const {
+  auto fds = std::vector<FunctionalDependency>();
+  const auto& unique_constraints = Hyrise::get().storage_manager.get_table(table_name)->get_soft_unique_constraints();
+  const auto& column_expressions = this->column_expressions();
+
+  for (auto& constraint : unique_constraints) {
```
Bouncner, line 100: const

src/lib/operators/abstract_join_operator.hpp
```diff
@@ -29,7 +29,7 @@ struct JoinConfiguration {
  * Base class for predicated (i.e., non-cross) join operator implementations. Cross Joins are performed by the Product
  * operator.
  *
- * Find more information about joins in our Wiki: https://github.com/hyrise/hyrise/wiki/Operator-Join
+ * Find more information about joins in our Wiki: https://github.com/hyrise/hyrise/wiki/Operators_JoinSortMerge
```
Bouncner, line 32: Better link our default join? https://github.com/hyrise/hyrise/wiki/Hash-Join-Operator

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -24,155 +24,103 @@ using namespace opossum;  // NOLINT
...
 
-  // Intersect primary key/unique columns and group-by columns. In case a primary key/unique constraint covers
-  // multiple columns, we need to check that all columns are present in order to later remove dependent columns.
-  std::vector<ColumnID> intersection;
-  std::set_intersection(constraint_columns.begin(), constraint_columns.end(), group_by_columns.begin(),
-                        group_by_columns.end(), std::back_inserter(intersection));
-
-  // Skip the current constraint as the primary key/unique constraint is not completely present.
-  if (intersection.size() != constraint_columns.size()) {
+  // To benefit from this rule, the FD's  columns have to be part of the group-by list
```
Bouncner, line 31: double space

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler:
```cpp
return std::tie(fd_left.first.size(), fd_left.second.size()) < std::tie(fd_right.first.size(), fd_right.second.size())
```
The code above does not compile. How would you do it?

Bouncner: Why `<`?

Bouncner: If it's not more compact or you think it's not better to understand, never mind. It think it's much easier to understand for such tasks, but that's just my taste.

julianmenzler: > Why `<`?

Oh, it should be `==`. The `<` operator was the one I found in various examples.

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -101,21 +101,41 @@ TEST_F(DependentGroupByReductionRuleTest, SimpleCases) {
   }
 }
 
-// Test that a removable column is removed when the single primary key column is present.
```
mrks, line 104: ?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -24,155 +24,103 @@ using namespace opossum;  // NOLINT
  *
  * @returns   Boolean value denoting whether at the group-by list of @param aggregate_node changed.
  */
-bool reduce_group_by_columns_for_constraint(const TableConstraintDefinition& table_constraint,
-                                            const std::set<ColumnID>& group_by_columns,
-                                            const std::shared_ptr<const StoredTableNode>& stored_table_node,
-                                            AggregateNode& aggregate_node) {
+bool reduce_group_by_columns_for_fd(const FunctionalDependency& fd, const ExpressionUnorderedSet& group_by_columns,
```
mrks, line 27: How about `remove_dependent_group_by_columns`?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -24,155 +24,103 @@ using namespace opossum;  // NOLINT
...
-  std::set_intersection(constraint_columns.begin(), constraint_columns.end(), group_by_columns.begin(),
-                        group_by_columns.end(), std::back_inserter(intersection));
-
-  // Skip the current constraint as the primary key/unique constraint is not completely present.
-  if (intersection.size() != constraint_columns.size()) {
+  // To benefit from this rule, the FD's columns have to be part of the group-by list
+  if (!std::all_of(fd.first.cbegin(), fd.first.cend(),
+                   [&group_by_columns](std::shared_ptr<AbstractExpression> constraint_col_expr) {
+                     return group_by_columns.contains(constraint_col_expr);
+                   })) {
```
mrks, line 35: If we make FunctionalDependencies a struct (to allow for `operator==`), this check should be done over there.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -24,155 +24,103 @@ using namespace opossum;  // NOLINT
...
-  std::vector<ColumnID> intersection;
-  std::set_intersection(constraint_columns.begin(), constraint_columns.end(), group_by_columns.begin(),
-                        group_by_columns.end(), std::back_inserter(intersection));
-
-  // Skip the current constraint as the primary key/unique constraint is not completely present.
-  if (intersection.size() != constraint_columns.size()) {
+  // To benefit from this rule, the FD's columns have to be part of the group-by list
+  if (!std::all_of(fd.first.cbegin(), fd.first.cend(),
+                   [&group_by_columns](std::shared_ptr<AbstractExpression> constraint_col_expr) {
+                     return group_by_columns.contains(constraint_col_expr);
```
mrks, line 34: `constraint_column_expression`

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -24,155 +24,103 @@ using namespace opossum;  // NOLINT
...
+                   })) {
     return false;
   }
 
-  // Every column that is part of the table but not of the primary key/unique constraint is going to be moved from
-  // the group-by list to the list of aggregates wrapped in an ANY().
+  // Every column that is functionally dependent gets moved from
+  // the group-by list to the aggregate list. For this purpose it is wrapped in an ANY() expression.
```
mrks, line 40: format

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -24,155 +24,103 @@ using namespace opossum;  // NOLINT
...
-      group_by_columns_per_table[stored_table_node].insert(column_id);
-    }
+    // Early exit: If there are no functional dependencies, we can skip this rule.
+    auto fds = aggregate_node.functional_dependencies();
+    if (fds.empty()) return LQPVisitation::VisitInputs;
 
-    // Store copy of the aggregate's column expressions to later check if this order is the query's final column order
+    // --- Preparation ---
+    // Store a copy of the root's column expressions before applying the rule
+    const auto root_column_expressions = root_lqp->column_expressions();
```
mrks, line 88: This is now called more than once. Remember that column_expressions is somewhat expensive.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -24,155 +24,103 @@ using namespace opossum;  // NOLINT
...
+    // Sort the FDs by their left set's column count in hope that the shortest will later form the group-by clause.
+    std::sort(fds.begin(), fds.end(),
+              [](const auto& fd_left, const auto& fd_right) { return fd_left.first.size() < fd_right.first.size(); });
+
+    // --- Main ---
+    // Try to reduce the group-by list one constraint at a time, starting with the shortest constraint. Multiple
+    // constraints might result in different group-by lists (depending on the number of columns of the constraint).
+    // We stop as soon as one constraint successfully reduced the group-by list. There is no advantage in a second
+    // reduction if the first reduction was successful, because no further columns will be removed. Hence, as soon as
+    // one reduction took place, we can ignore the remaining constraints.
```
mrks, line 107: You didn't add this, but is it actually true? Thinking of a table with column a,b,c,d,e, FDs `a->b`, `c->d` (Mr. Codd, please forgive me), and `GROUP BY a,b,c,d`, the rule might actually be applied twice.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -24,155 +24,103 @@ using namespace opossum;  // NOLINT
...
 
-  // Every column that is part of the table but not of the primary key/unique constraint is going to be moved from
-  // the group-by list to the list of aggregates wrapped in an ANY().
+  // Every column that is functionally dependent gets moved from
+  // the group-by list to the aggregate list. For this purpose it is wrapped in an ANY() expression.
   for (const auto& group_by_column : group_by_columns) {
-    if (std::find(constraint_columns.begin(), constraint_columns.end(), group_by_column) != constraint_columns.end()) {
-      // Do not touch primary key/unique constraint columns.
-      continue;
+    DebugAssert(std::dynamic_pointer_cast<LQPColumnExpression>(group_by_column), "Unexpected group-by-column type.");
```
mrks, line 42: This breaks the following:

```
generate_tpch .01
SELECT p_partkey, p_partkey + 1 FROM part GROUP BY p_partkey, p_partkey + 1
```

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
Bouncner: Sounds right. That comment is only true for PKs, not for general FKs. At least your examples sounds rights to me. Can we distinguish between FKs from PKs and from other "sources"?
If not, it should rarely matter if we have multiple reductions runs even though the not-first runs might no longer reduce the list.

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler: I thought it makes sense to compare the `size()`-properties first before checking both unordered sets (left & right) for equality. Unnecessary (micro-)optimization?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -245,6 +245,60 @@ bool AbstractLQPNode::is_column_nullable(const ColumnID column_id) const {
...
+    for (const auto& fd_right : fds_right) {
+      bool duplicate = std::any_of(fds_left.begin(), fds_left.end(), [&fd_right](const auto& fd_left) {
+        return (fd_left.first.size() == fd_right.first.size() && fd_left.second.size() == fd_right.second.size() &&
+                fd_left.first == fd_right.first && fd_left.second == fd_right.second);
+      });
+      if (!duplicate) {
+        fds_in.push_back(fd_right);
+      }
+    }
+    std::move(fds_left.begin(), fds_left.end(), std::back_inserter(fds_in));
```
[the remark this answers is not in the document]
julianmenzler: We could. This would change line 266 to the following:
```cpp
bool duplicate = std::any_of(fds_in.begin(), fds_in.end(), [&fd_right](const auto& fd_left) { .. }
``` 
Since we add to `fds_in` continuously, the duplicate-check needs additional cycles with for-iteration. But I guess, you consider this as micro-optimization that can be omitted, right?

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: Yes. Also, it [happens anyway.](https://github.com/gcc-mirror/gcc/blob/1580b4793d88132747f587ac95ebc12d96fee5b9/libstdc%2B%2B-v3/include/bits/stl_tree.h#L1627)

src/lib/logical_query_plan/abstract_lqp_node.cpp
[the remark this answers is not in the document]
mrks: Can't we simply do `fds_in = fds_left;` in all cases and add elements from the right as needed? That would mean that we don't move anymore, but I think the impact of that is negligible.

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -245,6 +245,60 @@ bool AbstractLQPNode::is_column_nullable(const ColumnID column_id) const {
...
+  }
+
+  // Currently, we do not support FDs in conjunction with null values.
+  // Since previous operators like outer joins might have added null values, we have to check columns for nullability.
```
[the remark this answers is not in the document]
julianmenzler: Good catch! I change it to `Some LQP nodes, like for instance outer joins, might add null values...`

src/lib/logical_query_plan/abstract_lqp_node.cpp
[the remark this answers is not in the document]
mrks: > But I guess, you consider this as micro-optimization that can be omitted, right?

It's good to keep these things in mind. Once readability or debugability are affected, the performance aspect needs to prove its relevance first.

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -92,6 +92,48 @@ bool StoredTableNode::is_column_nullable(const ColumnID column_id) const {
...
+
+    // Gather column expressions for constraint's column ids
+    for (const auto& expression : column_expressions) {
+      const auto lqp_column = std::dynamic_pointer_cast<LQPColumnExpression>(expression);
+      if (std::any_of(constraint.columns.cbegin(), constraint.columns.cend(),
+                      [&lqp_column](const auto constraint_column_id) {
+                        return constraint_column_id == lqp_column->column_reference.original_column_id();
+                      }))
+        left.insert(lqp_column);
+    }
```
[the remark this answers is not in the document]
julianmenzler: This makes life a lot simpler. Thanks!

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -92,6 +92,48 @@ bool StoredTableNode::is_column_nullable(const ColumnID column_id) const {
...
+    Assert(left.size() == constraint.columns.size(),
+           "Failed to collect all column expressions for constraint's column ids");
+
+    // Find column expressions that are functionally dependent
+    std::copy_if(column_expressions.begin(), column_expressions.end(), std::inserter(right, right.begin()),
+                 [&left](std::shared_ptr<AbstractExpression> column_expr) { return !(left.contains(column_expr)); });
+
+    // Create functional dependency
+    if (!right.empty() && std::all_of(right.cbegin(), right.cend(), [this](const auto& right_column_expr) {
+          return !right_column_expr->is_nullable_on_lqp(*this);
```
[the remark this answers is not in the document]
julianmenzler: Interesting. I had to think about this one for some time. You seem to be right. 
Since it is not a bidirectional relationship, our current FD definition (Left => Right) is not hurt by null values on the right hand side.
I have to revisit the forwarding logic regarding this. Big thanks for bringing this up!

julianmenzler:
TPC-H results using:
`numactl -m1 -N1 ./hyriseBenchmarkTPCH --scheduler -s 0.01`

```
| TPC-H 01       | 46.1786003112793   | 2771 | 46.094242095947266 | 2766 | -0%        | (run time too short) |
| TPC-H 02       | 71.76143646240234  | 4306 | 69.05593872070312  | 4144 | -4%        | (run time too short) |
| TPC-H 03       | 96.84539794921875  | 5811 | 96.73564147949219  | 5805 | -0%        | (run time too short) |
| TPC-H 04       | 99.30467224121094  | 5959 | 99.29846954345703  | 5958 | -0%        | (run time too short) |
| TPC-H 05       | 82.54818725585938  | 4953 | 82.1157455444336   | 4927 | -1%        | (run time too short) |
| TPC-H 06       | 99.36840057373047  | 5963 | 99.369873046875    | 5963 | +0%        | (run time too short) |
| TPC-H 07       | 48.9948616027832   | 2940 | 46.28156661987305  | 2777 | -6%        | (run time too short) |
| TPC-H 08       | 22.681163787841797 | 1361 | 16.864768981933594 | 1012 | -26%       | (run time too short) |
| TPC-H 09       | 47.74872970581055  | 2865 | 47.463626861572266 | 2848 | -1%        | (run time too short) |
| TPC-H 10       | 76.67655944824219  | 4601 | 76.07266235351562  | 4565 | -1%        | (run time too short) |
| TPC-H 11       | 99.00297546386719  | 5941 | 98.92937469482422  | 5936 | -0%        | (run time too short) |
| TPC-H 12       | 96.14420318603516  | 5769 | 95.92547607421875  | 5756 | -0%        | (run time too short) |
| TPC-H 13       | 64.04972076416016  | 3843 | 64.62657165527344  | 3878 | +1%        | (run time too short) |
| TPC-H 14       | 99.2937240600586   | 5958 | 99.29084014892578  | 5958 | -0%        | (run time too short) |
| TPC-H 15       | 99.1980972290039   | 5952 | 99.04623413085938  | 5943 | -0%        | (run time too short) |
| TPC-H 16       | 50.644775390625    | 3039 | 50.37803649902344  | 3023 | -1%        | (run time too short) |
| TPC-H 17       | 98.66841125488281  | 5921 | 97.3362808227539   | 5841 | -1%        | (run time too short) |
| TPC-H 18       | 51.541908264160156 | 3093 | 51.49848175048828  | 3090 | -0%        | (run time too short) |
| TPC-H 19       | 46.02850341796875  | 2762 | 45.71327209472656  | 2743 | -1%        | (run time too short) |
| TPC-H 20       | 51.64456558227539  | 3099 | 51.61005401611328  | 3097 | -0%        | (run time too short) |
| TPC-H 21       | 48.11251449584961  | 2887 | 47.512149810791016 | 2851 | -1%        | (run time too short) |
| TPC-H 22       | 80.93133544921875  | 4856 | 78.8481216430664   | 4731 | -3%        | (run time too short) |
| geometric mean |                    |      |                    |      | -2%        |                      |
```

mrks:
Could you run plot_performance_breakdown.py for the old and new versions, please? I wonder if this is in the optimizer.

mrks:
Q8 is sufficient

src/test/logical_query_plan/join_node_test.cpp
```diff
@@ -195,4 +196,81 @@ TEST_F(JoinNodeTest, IsColumnNullableWithOuterJoin) {
   EXPECT_FALSE(lqp_full_join->is_column_nullable(ColumnID{4}));
 }
 
+TEST_F(JoinNodeTest, FunctionalDependenciesDuplicates) {
+  // In case of self-joins, the implementation has to ensure that FDs are not outputted twice.
```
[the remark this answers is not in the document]
julianmenzler: Ok. I wrote this as a test for the following functionality:

```cpp
  // Merge input FDs and remove duplicates, if any
  auto& fds_in = fds_left;
  if (!fds_right.empty()) {
    // Remove duplicate FDs that might result from e.g. self-joins
    for (const auto& fd_right : fds_right) {
      bool duplicate = std::any_of(fds_left.begin(), fds_left.end(),
                                   [&fd_right](const auto& fd_left) { return (fd_left == fd_right); });
      if (!duplicate) {
        fds_in.push_back(fd_right);
      }
    }
  }
``` 

So would you remove the above block from `AbstractLQPNode::functional_dependencies()`?

src/test/logical_query_plan/join_node_test.cpp
```diff
@@ -195,4 +196,81 @@ TEST_F(JoinNodeTest, IsColumnNullableWithOuterJoin) {
...
+TEST_F(JoinNodeTest, FunctionalDependenciesNullabilityFilter) {
+  // Create two tables of 2 columns each
+  TableColumnDefinitions column_definitions_a{{"a", DataType::Int, false}, {"b", DataType::Int, false}};
+  TableColumnDefinitions column_definitions_b{{"x", DataType::Int, false}, {"y", DataType::Int, false}};
+  const auto table_a = std::make_shared<Table>(column_definitions_a, TableType::Data);
+  const auto table_b = std::make_shared<Table>(column_definitions_b, TableType::Data);
+  const auto table_name_a = "table_a";
+  const auto table_name_b = "table_b";
+  Hyrise::get().storage_manager.add_table(table_name_a, table_a);
+  Hyrise::get().storage_manager.add_table(table_name_b, table_b);
```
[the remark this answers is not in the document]
julianmenzler: As of yet, MockNodes do not create FDs from unique constraints as StoredTableNodes do.

But, we could create some dummy FDs and inject them into the MockNodes. I already added some basic container-functionality for FDs in MockNodes.

src/test/logical_query_plan/join_node_test.cpp
```diff
@@ -195,4 +196,81 @@ TEST_F(JoinNodeTest, IsColumnNullableWithOuterJoin) {
...
+
+  // Add unique constraints
+  table_a->add_soft_unique_constraint({ColumnID{0}}, IsPrimaryKey::No);
```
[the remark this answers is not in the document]
julianmenzler: In the next PR, I plan to change the interface of `add_soft_unique_constraint()` making `IsPrimaryKey::No` the default.  So, would omitting `IsPrimaryKey::No` help?

In StoredTableNode a new comment should indicate that we are looking for uniqueness only.

```cpp
for (const auto& constraint : unique_constraints) {
    // We build FDs from two column sets: LeftColumnSet => RightColumnSet
    // The left column set has to be
    //  a) unique (a guarantee already provided by the current unique constraint) and
    //  b) non-nullable
    if (std::any_of(constraint.columns.cbegin(), constraint.columns.cend(),
                    [this](const auto column_id) { return this->is_column_nullable(column_id); })) {
      continue;
    }
...
```

julianmenzler:
> Could you run plot_performance_breakdown.py for the old and new versions, please? I wonder if this is in the optimizer.

**master**
```cpp
Benchmark    Parser  SQLTranslator  Optimizer  LQPTranslator  Execution
TPC-H 08  0.002902       0.014984   0.867647       0.017022   0.097445
```
**This PR**
```cpp
Benchmark    Parser  SQLTranslator  Optimizer  LQPTranslator  Execution
TPC-H 08  0.002985       0.015198   0.863058       0.017772   0.100987
```

src/test/logical_query_plan/join_node_test.cpp
```diff
@@ -195,4 +196,81 @@ TEST_F(JoinNodeTest, IsColumnNullableWithOuterJoin) {
...
+
+  // Prerequisite
+  EXPECT_EQ(stored_table_node_a->functional_dependencies().size(), 1);  // a => b
+  EXPECT_EQ(stored_table_node_b->functional_dependencies().size(), 1);  // x => y
+
+  // Actual tests
+  EXPECT_EQ(inner_join_node->functional_dependencies().size(), 2);
+  EXPECT_EQ(left_join_node->functional_dependencies().size(), 1);
+  EXPECT_EQ(right_join_node->functional_dependencies().size(), 1);
+  EXPECT_EQ(full_outer_join_node->functional_dependencies().size(), 0);
```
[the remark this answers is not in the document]
julianmenzler: Ok, fair point. I'll improve it.

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -48,6 +48,19 @@ struct LQPOutputRelation {
...
+/**
+ * Models a functional dependency (FD) which consists out of two sets of column expressions
+ * (left and right set).
+ * Concrete values for the left set of columns lead to concrete values for the right set of columns.
+ * In other words, the left side unambigiously identifies the right side of values. (Left => Right)
+ * Currently, column expressions are required to be non-nullable to be involved in FDs. As there are
+ * strategies to combine both, null values and FDs (e.g. https://arxiv.org/abs/1404.4963), this might
+ * change in the future.
+ */
+using FunctionalDependency = std::pair<ExpressionUnorderedSet, ExpressionUnorderedSet>;
```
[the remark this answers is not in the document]
mrks: Doesn't seem to make a significant difference

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -48,6 +48,19 @@ struct LQPOutputRelation {
 
 using LQPNodeMapping = std::unordered_map<std::shared_ptr<const AbstractLQPNode>, std::shared_ptr<AbstractLQPNode>>;
 
+/**
+ * Models a functional dependency (FD) which consists out of two sets of column expressions
+ * (left and right set).
+ * Concrete values for the left set of columns lead to concrete values for the right set of columns.
+ * In other words, the left side unambigiously identifies the right side of values. (Left => Right)
```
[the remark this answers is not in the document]
mrks: Much better!

src/lib/logical_query_plan/stored_table_node.cpp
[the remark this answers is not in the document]
mrks: Argh, wait. This might not be the case if we pruned one of the prior columns. Potentially, there is a test missing here.

src/lib/logical_query_plan/stored_table_node.cpp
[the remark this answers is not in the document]
mrks: Mmh? This seems to look just like it was before?

src/lib/logical_query_plan/union_node.cpp
```diff
@@ -24,6 +24,23 @@ std::vector<std::shared_ptr<AbstractExpression>> UnionNode::column_expressions()
...
+    case SetOperationMode::Unique:
+    case SetOperationMode::All:
+      // No guarantees, since UnionAll can break any FD depending on the input.
+      return {};
+    case SetOperationMode::Positions:
+      // By definition, UnionPositions requires both input tables to have the same table origin and structure.
+      // Therefore, we can forward the FDs of either the left or right input node.
+      DebugAssert(left_input()->functional_dependencies() == right_input()->functional_dependencies(),
+                  "Expected "
+                  "both input nodes to have the same FDs.");
```
mrks, line 38: format

src/test/logical_query_plan/join_node_test.cpp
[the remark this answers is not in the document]
mrks: I believe that we currently shouldn't end up with duplicate FDs, no matter what we do. Change it to a DebugAssert and explain that this is only based on our current understanding and might need to change if the need arises?

src/test/logical_query_plan/join_node_test.cpp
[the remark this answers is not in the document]
mrks: However you like

src/test/logical_query_plan/join_node_test.cpp
```diff
@@ -195,4 +196,81 @@ TEST_F(JoinNodeTest, IsColumnNullableWithOuterJoin) {
...
+
+  // Prepare JoinNodes
+  const auto stored_table_node_a = StoredTableNode::make(table_name_a);
+  const auto stored_table_node_b = StoredTableNode::make(table_name_b);
+  const auto join_column_a = lqp_column_(LQPColumnReference(stored_table_node_a, ColumnID{0}));
+  const auto join_column_b = lqp_column_(LQPColumnReference(stored_table_node_b, ColumnID{0}));
+  // clang-format off
+  const auto inner_join_node = JoinNode::make(JoinMode::Inner, equals_(join_column_a, join_column_b),
+                                               stored_table_node_a,
+                                                 stored_table_node_b);
```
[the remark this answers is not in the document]
mrks: No, the indentation is to symbolize the LQP. Both a and b are on the same level, so the should have the same degree of indentation. Line 186, for example, was correct before.

src/test/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -144,4 +145,85 @@ TEST_F(StoredTableNodeTest, GetStatisticsPruneBothColumns) {
...
+  const auto table = Hyrise::get().storage_manager.get_table("t_a");
+  table->add_soft_unique_constraint({_a.original_column_id(), _b.original_column_id(), _c.original_column_id()},
+                                    IsPrimaryKey::No);
+
+  EXPECT_TRUE(_stored_table_node->functional_dependencies().empty());
+}
+
+TEST_F(StoredTableNodeTest, FunctionalDependenciesSingle) {
+  const auto table = Hyrise::get().storage_manager.get_table("t_a");
+  table->add_soft_unique_constraint({_a.original_column_id()}, IsPrimaryKey::No);
```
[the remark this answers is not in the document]
mrks: Maybe "Constraints are not tracked across the PQP."?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
mrks: never mind

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
mrks: In every iteration of `visit_lqp`

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
mrks: Could you explain the problems and what this assertion is supposed to guard against?

Even an aggregate expression could be a group by expression, e.g.:

`SELECT mb, COUNT(*) FROM (SELECT c, MIN(b) as mb FROM mixed GROUP BY c) t GROUP BY mb`

mrks:
Added some comments to my original review. Github makes it look weird. :(

Haven't done a final re-read of everything yet, as at least one point seemed to be still open from the last review.

src/test/logical_query_plan/join_node_test.cpp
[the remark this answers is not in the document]
julianmenzler: Ok! I removed the test and replaced the duplicate-handling code with a DebugAssert

src/test/logical_query_plan/join_node_test.cpp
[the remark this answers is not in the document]
julianmenzler: Got it.

src/test/logical_query_plan/join_node_test.cpp
```diff
@@ -195,4 +196,81 @@ TEST_F(JoinNodeTest, IsColumnNullableWithOuterJoin) {
...
+
+  // Add unique constraints
+  table_a->add_soft_unique_constraint({ColumnID{0}}, IsPrimaryKey::No);
+  table_b->add_soft_unique_constraint({{ColumnID{0}}}, IsPrimaryKey::No);
+
+  // Prepare JoinNodes
+  const auto stored_table_node_a = StoredTableNode::make(table_name_a);
+  const auto stored_table_node_b = StoredTableNode::make(table_name_b);
+  const auto join_column_a = lqp_column_(LQPColumnReference(stored_table_node_a, ColumnID{0}));
+  const auto join_column_b = lqp_column_(LQPColumnReference(stored_table_node_b, ColumnID{0}));
```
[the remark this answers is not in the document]
julianmenzler: True! Is addressed.

src/test/logical_query_plan/join_node_test.cpp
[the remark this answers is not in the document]
julianmenzler: Is addressed.

src/lib/logical_query_plan/stored_table_node.cpp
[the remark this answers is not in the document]
julianmenzler: Thanks! Now it is changed, together with the corresponding test. 
In the course of this, I made `FunctionalDependency` a struct, having a custom `operator==`.

src/test/logical_query_plan/stored_table_node_test.cpp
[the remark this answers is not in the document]
julianmenzler: Would "Constraints are not tracked for reference tables across the PQP." satisfy you?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
julianmenzler: But only for AggregateNodes since we got the following early exit:

```cpp
    // Early exit: If there are no functional dependencies, we can skip this rule.
    auto fds = aggregate_node.functional_dependencies();
    if (fds.empty()) return LQPVisitation::VisitInputs;
```

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
julianmenzler: > Could you explain the problems and what this assertion is supposed to guard against?

Some time ago, I experienced an error in the code above because of an aggregate column that slipped into the group by columns set. 
The root of the problem was an index error in the code that collected the group-by-columns.

I removed the DebugAssert.

src/lib/logical_query_plan/stored_table_node.cpp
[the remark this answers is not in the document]
julianmenzler: Good catch! Tests and new logic is added to support this. Unfortunately, reusing column expressions becomes hideous. So I decided to create them as necessary. See for yourself.

julianmenzler:
Wow, this whole PR has become a mess to overlook. 😅
However, I think every comment should be addressed now.

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -67,7 +67,7 @@ std::vector<std::shared_ptr<AbstractExpression>> StoredTableNode::column_express
     const auto table = Hyrise::get().storage_manager.get_table(table_name);
 
     // Build `_expression` with respect to the `_pruned_column_ids`
-    _column_expressions.emplace(table->column_count() - _pruned_column_ids.size());
+    _column_expressions.emplace(table->column_count() - _pruned_column_ids.size());  // <- does this make sense?
```
julianmenzler, line 70: I don't get this: Why do we are create an `AbstractExpression` from an Int value? Looks strange.

mrks:
> Wow, this whole PR has become a mess to overlook. 😅

Yepp. The GitHub UI makes me angry every time a PR has more than three comments. Last time I checked, I didn't find anything that didn't come with other issues, though. :/

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -246,6 +246,60 @@ bool AbstractLQPNode::is_column_nullable(const ColumnID column_id) const {
...
+  auto& fds_in = fds_left;
+  fds_in.insert(fds_in.end(), fds_right.cbegin(), fds_right.cend());
+
+  // Currently, we do not support FDs in conjunction with null values in their left column set.
+  // Some LQP nodes, however, change column nullability (like for instance outer joins). Therefore, we check input FDs
+  // for compliance and discard them, if necessary.
+  auto fds_out = std::vector<FunctionalDependency>();
+
+  // Collect non-nullable columns
+  auto column_expressions = this->column_expressions();
```
mrks, line 279: const, shadow

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -246,6 +246,60 @@ bool AbstractLQPNode::is_column_nullable(const ColumnID column_id) const {
...
+  if (left_input()) {
+    fds_left = left_input()->functional_dependencies();
+  }
+  if (right_input()) {
+    fds_right = right_input()->functional_dependencies();
+  }
+#if HYRISE_DEBUG
+  if (!fds_right.empty()) {
+    for (const auto& fd_right : fds_right) {
+      bool duplicate = std::any_of(fds_left.begin(), fds_left.end(),
```
mrks, line 262: const

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -50,6 +50,55 @@ struct LQPOutputRelation {
...
+ * {Author} => {Author-Nationality}
+ *
+ *
+ * Currently, column expressions are required to be non-nullable to be involved in FDs. As there are
+ * strategies to combine both, null values and FDs (e.g. https://arxiv.org/abs/1404.4963), this might
+ * change in the future.
+ */
+struct FunctionalDependency : std::pair<ExpressionUnorderedSet, ExpressionUnorderedSet> {
+  FunctionalDependency(ExpressionUnorderedSet left_column_set, ExpressionUnorderedSet right_column_set) {
+    first = left_column_set;
```
mrks, line 80: Inconsistent: left/first

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -50,6 +50,55 @@ struct LQPOutputRelation {
...
+    if (first.size() != other.first.size() || second.size() != other.second.size()) return false;
+
+    // Compare left column sets
+    for (const auto& column_expression : other.first) {
+      if (!first.contains(column_expression)) return false;
+    }
+    // Compare right column sets
+    for (const auto& column_expression : other.second) {
+      if (!second.contains(column_expression)) return false;
+    }
```
mrks, line 95: use op==

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -145,6 +194,11 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
    */
   virtual bool is_column_nullable(const ColumnID column_id) const;
 
+  /**
+   * @return The functional dependencies valid for this node
```
mrks, line 198: by default all are forwarded, nodes may overwrite

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -429,7 +429,6 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_union_node(
 
   switch (union_node->set_operation_mode) {
     case SetOperationMode::Unique:
-      Fail("Currently, only the All and Positions modes are implemented for the union operation");
```
mrks, line 432: ?

src/lib/logical_query_plan/stored_table_node.cpp
mrks: We don't. `_column_expressions` is an `optional<vector<...>>`, so we call `optional::emplace`, which in turn calls the `size_type` constructor of `vector<...>`. Smells a lot though and should be `_column_expressions = std::vector<std::shared_ptr<AbstractExpression>>(table->column_count() - _pruned_column_ids.size());`

mrks, line 70: change

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -92,6 +92,57 @@ bool StoredTableNode::is_column_nullable(const ColumnID column_id) const {
...
+      continue;
+    }
+
+    auto left = ExpressionUnorderedSet{};
+    auto right = ExpressionUnorderedSet{};
+
+    // Create column expressions for both column sets
+    for (auto column_id = ColumnID{0}; column_id < table->column_count(); ++column_id) {
+      // Because of column pruning, we create new column expressions. Some column expressions might not be part
+      // of _column_expressions, but part of the FD's right column set.
```
mrks, line 122: clarify

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -92,6 +92,57 @@ bool StoredTableNode::is_column_nullable(const ColumnID column_id) const {
...
+      const auto column_expression = std::make_shared<LQPColumnExpression>(shared_from_this(), column_id);
+
+      // Check, whether column expression belongs to the left or right column set of the FD
```
mrks, line 125: -,

src/test/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -144,4 +147,127 @@ TEST_F(StoredTableNodeTest, GetStatisticsPruneBothColumns) {
...
+  const auto table = Hyrise::get().storage_manager.get_table("t_a");
+  table->add_soft_unique_constraint({_a->original_column_id}, IsPrimaryKey::No);
+
+  // Prune column "b", which would be part of the right column set in the resulting FD: {a} => {b, c}
+  _stored_table_node->set_pruned_column_ids({ColumnID{1}});
+
+  // Although b is not part of the output column expressions, we want to see an FD returned.
+  const auto fd_expected = FunctionalDependency{{_a}, {_b, _c}};
+  EXPECT_EQ(_stored_table_node->functional_dependencies().size(), 1);
+  EXPECT_EQ(_stored_table_node->functional_dependencies().at(0), fd_expected);
```
mrks, line 193: b should not be contained

src/test/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -144,4 +147,127 @@ TEST_F(StoredTableNodeTest, GetStatisticsPruneBothColumns) {
...
+
+TEST_F(StoredTableNodeTest, FunctionalDependenciesPrunedLeftColumnSet) {
+  const auto table = Hyrise::get().storage_manager.get_table("t_a");
+  table->add_soft_unique_constraint({_a->original_column_id}, IsPrimaryKey::No);
+
+  // Prune unique column "a", which would be part of the left column set in the resulting FD: {a} => {b, c}
+  _stored_table_node->set_pruned_column_ids({ColumnID{0}});
+
+  EXPECT_EQ(_stored_table_node->functional_dependencies().size(), 0);
+}
```
mrks, line 181: add `b->c` test

src/lib/logical_query_plan/union_node.cpp
```diff
@@ -24,6 +24,22 @@ std::vector<std::shared_ptr<AbstractExpression>> UnionNode::column_expressions()
   return left_input()->column_expressions();
 }
 
+std::vector<FunctionalDependency> UnionNode::functional_dependencies() const {
+  switch (set_operation_mode) {
+    case SetOperationMode::Unique:
+    case SetOperationMode::All:
+      // No guarantees, since UnionAll can break any FD depending on the input.
```
mrks, line 31: Clarify for unique

src/test/logical_query_plan/join_node_test.cpp
```diff
@@ -195,4 +196,65 @@ TEST_F(JoinNodeTest, IsColumnNullableWithOuterJoin) {
...
+  mock_node_b->set_functional_dependencies({fd_xy});
+
+  // Prepare JoinNodes
+  const auto join_column_a = a;
+  const auto join_column_x = x;
+  // clang-format off
+  const auto inner_join_node =
+  JoinNode::make(JoinMode::Inner, equals_(join_column_a, join_column_x),
+    mock_node_a,
+      mock_node_b);
```
mrks, line 222: The formatting is still broken :(

src/test/logical_query_plan/stored_table_node_test.cpp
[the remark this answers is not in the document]
mrks: Sounds good

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
mrks: Still less efficient than it could be, but not as bad as I thought

mrks:
Wtf? Those were my temporary notes and I never submitted this. What gives, Github? :(

mrks:
Still on it

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -50,6 +50,55 @@ struct LQPOutputRelation {
...
+ * {ISBN} => {Author-Nationality}
+ * Furthermore, knowing an author, we also know his nationality. Hence, we can specify another FD that applies:
+ * {Author} => {Author-Nationality}
+ *
+ *
+ * Currently, column expressions are required to be non-nullable to be involved in FDs. As there are
+ * strategies to combine both, null values and FDs (e.g. https://arxiv.org/abs/1404.4963), this might
+ * change in the future.
+ */
+struct FunctionalDependency : std::pair<ExpressionUnorderedSet, ExpressionUnorderedSet> {
```
mrks, line 78: Inheriting from STL containers is very dangerous, especially if it is public inheritance.

src/lib/logical_query_plan/lqp_translator.cpp
julianmenzler: No idea, how this came.

mrks:
Give me a second, please. I'm doing some changes right now.

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -253,7 +273,38 @@ TEST_F(DependentGroupByReductionRuleTest, ShortConstraintsFirst) {
...
+  auto c = mock_node->get_column("c");
+  auto d = mock_node->get_column("d");
+  auto e = mock_node->get_column("e");
+  auto fd1 = FunctionalDependency{{a}, {b}};
+  auto fd2 = FunctionalDependency{{c}, {d}};
+  mock_node->set_functional_dependencies({fd1, fd2});
+
+  // clang-format off
+  auto lqp =
+    AggregateNode::make(expression_vector(a, b, c, d), expression_vector(sum_(e)),
```
julianmenzler, line 299: ident

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -101,21 +101,41 @@ TEST_F(DependentGroupByReductionRuleTest, SimpleCases) {
...
+      ProjectionNode::make(expression_vector(column_a_0, column_a_1, sum_(column_a_2)),
+      AggregateNode::make(expression_vector(column_a_0), expression_vector(sum_(column_a_2), any_(column_a_1)),
+        stored_table_node_a));  // NOLINT
 
-  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+    EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+  }
+  {
+    auto lqp =
+      AggregateNode::make(expression_vector(column_a_0, column_a_1), expression_vector(sum_(column_a_0),
```
julianmenzler, line 124: indentation

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -101,21 +101,41 @@ TEST_F(DependentGroupByReductionRuleTest, SimpleCases) {
...
+                                                sum_(column_a_1), sum_(column_a_2)), stored_table_node_a);  // NOLINT
+
+    const auto actual_lqp = apply_rule(rule, lqp);
+
+    const auto expected_lqp =
+      ProjectionNode::make(
```
julianmenzler, line 130: indentation

src/test/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -253,7 +273,38 @@ TEST_F(DependentGroupByReductionRuleTest, ShortConstraintsFirst) {
 
   const auto expected_lqp =
   ProjectionNode::make(expression_vector(column_e_0, column_e_1, column_e_2),
-    AggregateNode::make(expression_vector(column_e_2), expression_vector(any_(column_e_0), any_(column_e_1)), stored_table_node_e));  // NOLINT
+    AggregateNode::make(expression_vector(column_e_2), expression_vector(any_(column_e_1), any_(column_e_0)), stored_table_node_e));  // NOLINT
```
julianmenzler, line 276: indentation

julianmenzler:
> Give me a second, please. I'm doing some changes right now.
Ok.

Adding single comments does not seem to be a good idea. 😑

src/lib/logical_query_plan/functional_dependency.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+FunctionalDependency::FunctionalDependency(ExpressionUnorderedSet init_determinants,
+                                           ExpressionUnorderedSet init_dependents)
+    : determinants(init_determinants), dependents(init_dependents) {
+  DebugAssert(!determinants.empty() && !dependents.empty(), "FunctionalDependency cannot be empty");
+}
+
+bool FunctionalDependency::operator==(const FunctionalDependency& other) const {
+  // Cannot use unordered_set::operator== because it ignores the custom equality function.
+  // https://stackoverflow.com/questions/36167764/can-not-compare-stdunorded-set-with-custom-keyequal
```
mrks, line 13: bloody hell, cpp. :(

julianmenzler: Oh well, that's what I ran into as well. After debugging, making sure that the sets were equal, I built the `operator==` and it simply worked. 
I did not invest any energy to investigate this issue further.

And to be honest: Having read the SO-post and `std::is_permutation`, I still don't understand it. Anyway. As long as it works now. 😅

src/test/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -187,8 +199,7 @@ TEST_F(StoredTableNodeTest, FunctionalDependenciesPrunedRightColumnSet) {
   // Prune column "b", which would be part of the right column set in the resulting FD: {a} => {b, c}
   _stored_table_node->set_pruned_column_ids({ColumnID{1}});
 
-  // Although b is not part of the output column expressions, we want to see an FD returned.
-  const auto fd_expected = FunctionalDependency{{_a}, {_b, _c}};
+  const auto fd_expected = FunctionalDependency{{_a}, {_c}};
```
julianmenzler, line 202: I think, we can not do this. Think of the following example:

```
a  b  c
0  0  0  -> {0} => {0, 0}
1  1  0  -> {1} => {1, 0}
```

After pruning b:
```
a  (b)  c
0  (0)  0  -> {0} => {0}
1  (1)  0  -> {1} => {0}
```

I included the FD anyway because it does not break anything. And in case our pruned column comes back / joins our LQP again, we have the FD still in place.

src/test/logical_query_plan/union_node_test.cpp
```diff
@@ -91,12 +91,6 @@ TEST_F(UnionNodeTest, DiscardFunctionalDependencies) {
   auto validate_node = ValidateNode::make(stored_table_node);
   EXPECT_EQ(validate_node->functional_dependencies().size(), 1);
 
-  // Test UnionAll (discard FDs)
-  auto union_all_node = UnionNode::make(SetOperationMode::All);
-  union_all_node->set_left_input(stored_table_node);
-  union_all_node->set_right_input(stored_table_node);
-  EXPECT_EQ(union_all_node->functional_dependencies().size(), 0);
-
```
julianmenzler: Just curious: Why not have such a test?

src/test/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -156,7 +156,7 @@ TEST_F(StoredTableNodeTest, FunctionalDependenciesNone) {
   table->add_soft_unique_constraint({_a->original_column_id, _b->original_column_id, _c->original_column_id},
                                     IsPrimaryKey::No);
 
-  EXPECT_EQ(_stored_table_node->functional_dependencies().size(), 0);
+  EXPECT_TRUE(_stored_table_node->functional_dependencies().empty());
```
julianmenzler: I decided to use `EXPECT_EQ` for better console output (all values instead of simple fail message). 
No issue though.

src/lib/logical_query_plan/abstract_lqp_node.hpp
julianmenzler: Noted, thanks!

julianmenzler:
By the way: Shall I adapt the PR description to reflect the recent changes?

mrks:
> By the way: Shall I adapt the PR description to reflect the recent changes?

Yes, please

src/test/logical_query_plan/stored_table_node_test.cpp
mrks: What do you mean with "all values"? Does that actually print the FDs?

src/test/logical_query_plan/union_node_test.cpp
mrks: Because I replaced the UnionAll handling with an assertion. We haven't decided what expressions are returned if we run a union with multiple input tables. Writing tests about this pretends that we do.

src/test/logical_query_plan/stored_table_node_test.cpp
mrks: I can't follow. Why is it a problem if we remove columns on the dependent side?

About the example: Any chance you are mixing FDs and uniqueness? The existence of an FD does not guarantee uniqueness on the right side. For example, `zip_code -> {city, state}` might be a valid FD, in which `zip_code -> {state}` holds, too.

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -256,42 +256,42 @@ std::vector<FunctionalDependency> AbstractLQPNode::functional_dependencies() con
...
-                                   [&fd_right](const auto& fd_left) { return (fd_left == fd_right); });
-      DebugAssert(!duplicate,
-                  "Unexpected duplicate functional dependency found in " + this->description(DescriptionMode::Short));
+      const bool duplicate = std::any_of(fds_left.begin(), fds_left.end(),
+                                         [&fd_right](const auto& fd_left) { return (fd_left == fd_right); });
+      Assert(!duplicate,
+             "Unexpected duplicate functional dependency found in " + this->description(DescriptionMode::Short));
     }
   }
-#endif
```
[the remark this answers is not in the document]
mrks: No need to do so. If we really wanted to, `if constexpr (HYRISE_DEBUG)` would be the way to go, but the compiler will remove the branch anyway.

src/lib/logical_query_plan/functional_dependency.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+std::ostream& operator<<(std::ostream& stream, const FunctionalDependency& expression) {
+  stream << "{";
+  auto determinants_vector =
+      std::vector<std::shared_ptr<AbstractExpression>>{expression.determinants.begin(), expression.determinants.end()};
+  stream << determinants_vector.at(0)->as_column_name();
+  for (auto determinant_idx = size_t{1}; determinant_idx < determinants_vector.size(); ++determinant_idx) {
+    stream << ", " << determinants_vector[determinant_idx]->as_column_name();
+  }
```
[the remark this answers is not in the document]
mrks: Because that would start with a comma.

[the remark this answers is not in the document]
julianmenzler: Oh yes, got it!

src/lib/logical_query_plan/abstract_lqp_node.cpp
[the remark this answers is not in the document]
julianmenzler: Ok.

src/test/logical_query_plan/stored_table_node_test.cpp
julianmenzler: > Any chance you are mixing FDs and uniqueness?

Indeed, you are right. Thanks!

src/test/logical_query_plan/stored_table_node_test.cpp
julianmenzler: No, but in case the container is not empty, `EXPECT_EQ` outputs the container's size.

src/test/logical_query_plan/union_node_test.cpp
julianmenzler: Alright, makes sense.

PR #2081, opened 2020-04-10
Add tutorial to setup a development environment only using docker

ramboldio:
I was updating my ubuntu system which broke my dev environment completely. With this PR, I want to prevent future generations from going through the same pain as I did by providing a handy guide that let's you work on hyrise without touching the host system.


ramboldio:
(currently this guid uses rsync to have the same files in the docker container as in the CLion projects, since this is the default and gives CLion control over what is happening. It is probably preffered to use docker volume mounting instead. Wdyt?)

ramboldio:
Do you think, it's a good idea to put together a dedicated Dockerfile for development? That might save some steps from the tutorial.

ramboldio:
@mrks What do you think?

mrks:
I'm currently a bit swamped trying to get the CI server running again. Will look at this soon.

Bouncner:
I have no idea about Docker and what that means for the compiling/execution performance of Hyrise, but since a lot of students appear to use CLion, it sounds like a good idea to put that guide down. Maybe not in the README to keep it concise, but linked in the Wiki?

mrks:
We discussed this in the Fortnightly before you joined. @julianmenzler will test the tutorial just to make sure, after which we can dicuss the best location.

ramboldio:
Nice. Would really love to see getting this merged and supported. It'll be a tremendous time saver for future DYOD students running Linux systems. And for anybody who has to debug issue that are only reproducible in the CI environment using the familiar CLion debugger.

ramboldio:
> I have no idea about Docker and what that means for the compiling/execution performance of Hyrise, but since a lot of students appear to use CLion, it sounds like a good idea to put that guide down. Maybe not in the README to keep it concise, but linked in the Wiki?

When the docker host runs Linux, there will be almost no overhead, since the processes in the docker container processes run directly in the host's kernel. They only use lightweight namespace isolation primitives like cgroups, an isolated file system and isolated networking.

On Windows there will be the overhead of WSL (ie. Linux syscalls are translated to Windows Kernel Syscalls) (only for the latest versions of Docker).
Mac users would pay the full overhead of a Linux VM that Docker for Mac uses under the hood. But I think the projects already supports Mac users quite well in its current configuration ;).

ramboldio:
@julianmenzler let me know if you have any questions. :v:

Bensk1:
In the context of the Bachelor's project, we are currently having a discussion whether docker is the right technology for creating development environments. On the plus side: we already have a docker setup for the CI but setting up a development environment in a docker apparently conflicts with a couple of Docker's dogmas. @Bouncner mentioned that we used Vagrant for Hyrise1 which focuses on a simple, lightweight setup for a development environment.

mrks:
What would be the advantage of that?

> lightweight setup

Doesn't it start a VM?

Bensk1:
I don't know. The Docker disciples just said/wrote that using docker containers for development is against Docker's philosophy. Actually, I don't care.

julianmenzler:
Compile errors under macOS:

https://gist.github.com/julianmenzler/166b64a3fdf0447ba67d7239a97fa79c

mrks:
Could you try building Hyrise in the container without CLion (i.e., creating your own build folder, running cmake, make)

ramboldio:
> Compile errors under macOS:
> https://gist.github.com/julianmenzler/166b64a3fdf0447ba67d7239a97fa79c

Also had that problem before when trying to get hyrise running on my ubuntu (non docker) dev machine. (I recall fixing it by changing GCC to CLang inside the docker). Takes some time to figure this out tho. :v: 

But if you cant build it that it should fail in CI also tho, right?

mrks:
Very weird.

> But if you cant build it that it should fail in CI also tho, right?

That's what I would expect, too. Especially as this runs in a VM and Mac should have no influence.

julianmenzler:
> Could you try building Hyrise in the container without CLion (i.e., creating your own build folder, running cmake, make)
I tried this today. Same issue.

Also, I made a clean checkout from within the container and started compiling using `make -j8`. This time no problems with `libjemalloc`, but other errors occurred.
However, I remembered your suggestion (@mrks) from our meeting:`make -j1`did the trick. Problems with macOS and Clion remain though.

As a next step; I'll try out the tutorial in a Linux setting. I already set up a VM with Linux and Clion today. I'll keep you posted.

mrks:
> Problems with macOS and Clion remain though.

Could you try this: https://youtrack.jetbrains.com/issue/CPP-349#focus=streamItem-27-890460.0-0

julianmenzler:
Update
* Tutorial works for Ubuntu 20.04 and Clion 2020.1
* I give up on getting this to work under macOS Catalina and Clion 2020.1

> Could you try this: https://youtrack.jetbrains.com/issue/CPP-349#focus=streamItem-27-890460.0-0

Clion triggers the following command
```
/cmake-3.16.0/bin/cmake --build /tmp/tmp.dzpGmw5myc/cmake-build-debug_docker --target hyriseTest -j 1
```
So `-j 1` is not a problem. Something else seems to get messed up when Clion copies over the project files I guess.


I suggest we put a disclaimer in front of the tutorial. Something like "Works on Linux systems only."

README.md
```diff
@@ -46,6 +46,52 @@ docker run -it hyrise
 
 Inside the container, you can then checkout Hyrise and run `./install_dependencies.sh` to download the required submodules.
 
+
+## Development inside Docker container using CLion
+
+The recommended way of developing with CLion on unsupported OS versions is to use the remote debugging capabilities of
```
julianmenzler, line 52: Maybe add a reference to the tutorial (which I guess you used?)
https://blog.jetbrains.com/clion/2020/01/using-docker-with-clion/#creating-a-docker-container-for-use-with-clion

README.md
```diff
@@ -46,6 +46,52 @@ docker run -it hyrise
...
+    - Click `+ Add` and `Remote Host`
+    - In the `Configure Remote Credentials` dialog (little gear next to host), enter as host `localhost`, port `2222` and choose as authentication type `OpenSSH config`
+    - Click `Apply` (**Do not yet click `OK`**)
+6. Configure a remote deployment in CLion
+    - Go to `File | Settings | Build, Execution, Deployment | Deployment | Options`
+    - Change `Preserve original file permissions` to `Yes`
+    - Click `OK`
+    - The files of the project will now be 'uploaded' to the docker container 
+7. Run CMake, and build the project and you're good to go.
+8. When you restart your system, you can bring back the development environment with `docker start -d hyrise_dev_env` or
```
julianmenzler, line 91: The `-d` option has no function (at least on my systems)

README.md
```diff
@@ -46,6 +46,52 @@ docker run -it hyrise
...
+[change the restart policy](https://docs.docker.com/config/containers/start-containers-automatically/#use-a-restart-policy) to permanently run in the background.
+
```
julianmenzler, line 93: Also worth mentioning: The ssh service needs to be started manually after restarting the container. Maybe there is also an option to run ssh on boot automatically. But I don't have a command for that right now.

julianmenzler:
Another question to be answered: Shall we make this a separate Wiki-page?

mrks:
> Shall we make this a separate Wiki-page?

Yes, and link it under "Development" in the menu. Could you make the necessary change to the text and move it there? I'd also link it in README.md.

julianmenzler:
> > Shall we make this a separate Wiki-page?
> 
> Yes, and link it under "Development" in the menu. Could you make the necessary changes to the text and move it there? I'd also link it in README.md.

Ok, I take care of that.

julianmenzler:
Ok, I'm done. Here it is: https://github.com/hyrise/hyrise/wiki/Use-Docker-with-CLion

mrks:
Great! Thanks to both of you!

Where does ENABLE_JIT_SUPPORT come from? That has been removed a long time ago. We should grep the code and the wiki for it.

If you change this PR to simply link the wiki page, we can merge and close it, too.

mrks:
Could you merge the current master, please? Otherwise, this might fail because of that server test.

julianmenzler:
> Great! Thanks to both of you!
> 
> Where does ENABLE_JIT_SUPPORT come from? That has been removed a long time ago. We should grep the code and the wiki for it.
> 
> If you change this PR to simply link the wiki page, we can merge and close it, too.

Good to know! I copied that part from my personal config. I removed it.

> If you change this PR to simply link the wiki page, we can merge and close it, too.

Done

---

By the way, is the following paragraph still relevant?
> Inside the container, you can then checkout Hyrise and run `./install_dependencies.sh` to download the required submodules.

From my view, the Dockerfile already ensures that dependencies get set up. Or is there any difference?

mrks:
The submodules are not part of the Docker container. The rest *should* be there, but as the CI run also runs the script, I'd suggest that people do so, too. That way, we can be reasonably sure that everything that works in CI also works for them.

julianmenzler:
Alright, thanks for clarifying. 👍

README.md
ramboldio: you're right. You only need it for `
docker run`

README.md
```diff
@@ -46,6 +46,52 @@ docker run -it hyrise
...
+The recommended way of developing with CLion on unsupported OS versions is to use the remote debugging capabilities of 
+the IDE. Using this approach, the docker container is treated as a remote instance of the system that is controlled by 
+CLion using ssh. You can achieve this by following these steps:
+
+0. Make sure that you checked out the git submodules  by running `git submodule update --init --recursive` (in project root)
+1. Build docker image using `docker build . -t hyrise` (in project root)
+2. Running container and keep it alive in the background, so you can connect later:
+
+```bash
+docker run --name hyrise_dev_env --cap-add sys_ptrace -p 2222:22 -d -it hyrise sleep infinity
```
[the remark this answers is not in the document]
ramboldio: sure, let's do it!

ramboldio:
Why I put the install_dependency section in was because I needed a way of only installing the submodules and not the other dependencies that would be installed in the hosts apt-system.

So maybe just a reminder in the tutorial that the submodules have to be set up to make Docker work could suffice.

README.md
ramboldio: Absolutely. Thought I put this in, but well. :D

README.md
ramboldio: Good point. I think we want to figure out how to do this automatically.

ramboldio:
Have you thought about whether we should switch to a volume-mounting approach instead of rsyncing all the files? I guess that volume mounting has a bunch of advantages. Then there is no overhead for copying the files and you wouldn't need to provide the disc space for 2 copies of hyrise.

julianmenzler:
> So maybe just a reminder in the tutorial that the submodules have to be set up to make Docker work could suffice.

The Dockerfile takes care of fetching all dependencies required for compiling Hyrise. Project files including the git submodules have to be cloned by the developer himself. 

> Have you thought about whether we should switch to a volume-mounting approach instead of rsyncing all the files? I guess that volume mounting has a bunch of advantages. Then there is no overhead for copying the files and you wouldn't need to provide the disc space for 2 copies of hyrise.

Indeed, I thought about this. But it might break CLion's remote workflow and requires further testing. Most people will set up a Hyrise dev environment without Docker anyway. So I guess spending more resources on this guide is not necessary. At least for now.

PR #2084, opened 2020-04-14
Better information during table generation

Bouncner:
I am working on compression schemes and was surprised that (the apparently not necessary) re-encoding (importing LZ4 as binary and using LZ4 as the compression for all columns) took quite a long time (can be easily minutes for large data sets).

Everything worked as expected, but our encoding step also generates the pruning statistics. This PR simply adds that information to the output as it might be confusing to users (as it was to me).


src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -148,7 +148,7 @@ void AbstractTableGenerator::generate_and_store() {
   /**
    * Encode the tables
    */
-  std::cout << "- Encoding tables if necessary" << std::endl;
+  std::cout << "- Encoding tables if necessary and generating pruning statistics" << std::endl;
```
mrks, line 151: "Encoding tables (if necessary) and generating pruning statistics"?

PR #2092, opened 2020-04-17
Make code clang-10 ready

mrks:
I locally installed clang-10 on one of my laptops. Mostly, clang now warns about implicit int to float conversions. While adding static_casts makes the code more verbose, I think that this might make people more aware of potentially lossy casts.

clang-10 support is not yet part of CI, so for now, this is a one time change with (hopefully) no visible effects.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -41,12 +41,15 @@ endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_STANDARD 20)
 
+# Hotfix for boost bug (https://github.com/boostorg/asio/issues/312):
+add_compile_options(-DBOOST_ASIO_DISABLE_CONCEPTS)
+
 # When generating a dtrace header file, symbols containing dollar-signs are created This file needs to be compiled as well.
 # Hence, the `-Wno-dollar-in-identifier-extension` flag is required.
 add_compile_options(-pthread -Wall -Wextra -pedantic -Werror -Wno-unused-parameter -Wno-dollar-in-identifier-extension -Wno-unknown-pragmas -Wno-subobject-linkage)
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-    add_compile_options(-Weverything -Wshadow-all -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-documentation -Wno-padded -Wno-global-constructors -Wno-sign-conversion -Wno-exit-time-destructors -Wno-switch-enum -Wno-weak-vtables -Wno-double-promotion -Wno-covered-switch-default -Wno-unused-macros -Wno-newline-eof -Wno-missing-variable-declarations -Wno-weak-template-vtables -Wno-missing-prototypes -Wno-float-equal -Wno-return-std-move-in-c++11 -Wno-unreachable-code-break -Wno-undefined-func-template -Wno-unknown-warning-option -Wno-pass-failed -Wno-ctad-maybe-unsupported -Wno-header-hygiene)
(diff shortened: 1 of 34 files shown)


src/lib/strong_typedef.hpp
```diff
@@ -20,7 +20,7 @@
 
 #define STRONG_TYPEDEF(T, D)                                                                                      \
   namespace opossum {                                                                                             \
-  struct D : boost::totally_ordered1<D, boost::totally_ordered2<D, T>> {                                          \
+  struct D : boost::totally_ordered1<D> {                                                                         \
```
Bouncner, line 23: Can you explain to me why you changed that?

src/test/server/server_test_runner.cpp
```diff
@@ -431,7 +431,7 @@ TEST_F(ServerTestRunner, TestTransactionConflicts) {
   EXPECT_GE(conflicted_increments, 1);
 
   EXPECT_EQ(successful_increments + conflicted_increments, num_threads * iterations_per_thread);
-  EXPECT_FLOAT_EQ(final_sum - initial_sum, successful_increments);
```
Bouncner, line 434: Is this Gtest macro now no longer working with Clang10?

src/CMakeLists.txt
```diff
@@ -41,12 +41,15 @@ endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_STANDARD 20)
 
+# Hotfix for boost bug (https://github.com/boostorg/asio/issues/312):
+add_compile_options(-DBOOST_ASIO_DISABLE_CONCEPTS)
```
Bouncner, line 45: Is there a way to remind us to remove that option once in a while? Would a TODO help here?

src/CMakeLists.txt
```diff
@@ -41,12 +41,15 @@ endif()
...
+    add_compile_options(-Weverything -Wshadow-all -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-documentation -Wno-padded -Wno-global-constructors -Wno-sign-conversion -Wno-exit-time-destructors -Wno-switch-enum -Wno-weak-vtables -Wno-double-promotion -Wno-covered-switch-default -Wno-unused-macros -Wno-newline-eof -Wno-missing-variable-declarations -Wno-weak-template-vtables -Wno-missing-prototypes -Wno-float-equal -Wno-return-std-move-in-c++11 -Wno-unreachable-code-break -Wno-undefined-func-template -Wno-unknown-warning-option -Wno-pass-failed -Wno-ctad-maybe-unsupported -Wno-header-hygiene -Wno-poison-system-directories)
```
Bouncner, line 52: The docs I found mention that this flag is intended for cross-platform builds. Why do we use it? (not that I am against it, sounds useful)

src/test/server/server_test_runner.cpp
mrks: No, we just shouldn't compare ints with it ;)

src/CMakeLists.txt
mrks: Didn't really look into it. We don't build cross-platform anyway. My first impression is that fixing it is CMake's job, not ours: https://gitlab.kitware.com/cmake/cmake/issues/17483

Bouncner: I just wanted to say "nevermind". I thought it's the other way around and we ensure that we allow cross-platform. I misread the doc.

src/CMakeLists.txt
mrks: TODO sounds like something that can be done right now. Given the age of boost versions in Ubuntu, I'd guess that this won't make it into the standard boost any time soon,.

src/lib/strong_typedef.hpp
mrks: For some reason, clang now thinks that there are two possibilities for `operator==` - the one defined by boost and the one if using the user-defined conversion to the base type. Removing this makes sure that comparisons to the base type always use the second path.

src/CMakeLists.txt
Bouncner: Ack

PR #2096, opened 2020-04-20
Add LogManager

dey4ss:
This PR introduces a basic log which can be accessed via a meta table. 
As an example, the MvccDeletePlugin logs the factor of deleted rows as Info messages.


src/lib/utils/log_manager.cpp
```diff
@@ -0,0 +1,26 @@
+#include "log_manager.hpp"
+
+#include "constant_mappings.hpp"
+#include "utils/settings/log_level_setting.hpp"
+
+namespace opossum {
+
+LogManager::LogManager(SettingsManager& settings_manager, LogLevel log_level) : _log_level(log_level) {
```
mrks, line 8: I'd retrieve the settings_manager from the Hyrise object. We don't pass around pseudo singletons (e.g., the StorageManager) in other places, so we don't need to start here.

src/lib/utils/log_manager.cpp
```diff
@@ -0,0 +1,26 @@
...
+  if (!settings_manager.has_setting(LOG_LEVEL_SETTING_NAME)) {
+    auto log_level_setting = std::make_shared<LogLevelSetting>(LOG_LEVEL_SETTING_NAME);
+    log_level_setting->register_at(settings_manager);
+  }
+}
+
+void LogManager::add_message(const std::string& reporter, const std::string& message, const LogLevel log_level) {
+  if (log_level < _log_level) return;
```
mrks, line 16: What might become a problem is that creating the string could be an expensive operation in itself. Have a look at https://stackoverflow.com/questions/11832960/short-circuit-operator-output-in-c

Could you run a benchmark where you insert an ignored log statement, e.g., `log (this + chunk_id)` in  Table::get_chunk?

src/lib/utils/log_manager.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include <chrono>
+#include <functional>
+
+#include "types.hpp"
+#include "utils/settings_manager.hpp"
+
+namespace opossum {
+
+struct LogEntry {
+  std::chrono::milliseconds timestamp;
```
mrks, line 13: std::chrono::time_point?

src/lib/utils/meta_tables/meta_log_table.cpp
```diff
@@ -0,0 +1,32 @@
+#include "meta_log_table.hpp"
+
+#include "constant_mappings.hpp"
+#include "hyrise.hpp"
+#include "utils/assert.hpp"
+
+namespace opossum {
+
+MetaLogTable::MetaLogTable()
+    : AbstractMetaTable(TableColumnDefinitions{{"timestamp", DataType::Long, false},
```
mrks, line 10: Could you add a `H:M:S` column, please?

src/lib/utils/settings/log_level_setting.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+const std::string& LogLevelSetting::description() const {
+  static const auto description = std::string{"Minimal log level of the LogManager"};
+  return description;
+}
+
+const std::string& LogLevelSetting::get() { return log_level_to_string.left.at(Hyrise::get().log_manager._log_level); }
+
+void LogLevelSetting::set(const std::string& value) {
+  Hyrise::get().log_manager._log_level = log_level_to_string.right.at(value);
```
mrks, line 18: This will kill everything if I mistyped the log level, right? We don't want AssertInput all over the code base, but maybe we need something more lenient here.

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -76,9 +82,16 @@ void MvccDeletePlugin::_logical_delete_loop() {
...
           std::unique_lock<std::mutex> lock(_mutex_physical_delete_queue);
           _physical_delete_queue.emplace(table, chunk_id);
+          removed_row_count += chunk_invalid_rows;
         }
       }
     }
+    if (removed_row_count > 0) {
+      std::ostringstream message;
+      message << "Compressed " << table_name << " by factor "
+              << std::to_string(static_cast<double>(removed_row_count) / row_count);
```
mrks, line 92: "Compressed" is not the perfect term here. Maybe "Consolidated x chunks, freeing up y MB"?

src/lib/utils/log_manager.cpp
dey4ss: The problem is that both Managers get initialized when Hyrise is initialized or reset. On Hyrise init, this causes infinite recursion, as Hyrise::get() keeps on creating new instances, on reset, we register the setting to the old manager that is reset, too.

src/lib/utils/settings/log_level_setting.cpp
dey4ss: I was something I was not sure about, I'm gonna add this

src/plugins/mvcc_delete_plugin.cpp
dey4ss: Hm, do we have a way to get a chunk's or segment's size? I guess this is not trivial when we consider string columns and different segment types, e.g. additional dictionaries (which can be different when we change the table order as the plugin does).

src/lib/utils/log_manager.cpp
mrks: Got it. Maybe add a comment?

src/lib/utils/settings/log_level_setting.cpp
mrks: Could be something to discuss in a more general way. Other settings will see the same.

src/plugins/mvcc_delete_plugin.cpp
mrks: estimate_memory_usage

src/lib/utils/log_manager.cpp
Bensk1, line 16: I think my idea of the LogLevel was a different one: Instead of checking the level during adding a message to the manager, we only return the messages with a certain level. That way all messages would be logged but only certain ones retrieved.

dey4ss: Would you prefer filtering via select (and thus, maybe need to address multiple values in the where clause) or keeping the minimal level as a setting?

src/lib/utils/settings/log_level_setting.cpp
dey4ss: For now, I simply checked if the value is ok. One could also add a try_set method to AbstractSetting that catches errors.

src/lib/utils/log_manager.cpp
dey4ss: I can do this. On the other hand, if we keep every message and filter while retrieving the entries, I think the overhead is something we have to accept. Not logging everything while doing performance-critical stuff is also a good idea for keeping the log small.

It is also the question what we want to log. As far as I understood it, we want to log mostly plugin actions, where we should not need to have to focus on performance too much.

mrks: > Not logging everything while doing performance-critical stuff is also a good idea for keeping the log small.

Give people a chance to do things wrong and someone will find it ;)

mrks: Maybe we should have another look at what we want to accomplish with the log levels. Looking at it again, I don't see a purpose for the `error` level and I question the `warning` level. By design, a Hyrise process does not have any loggable errors, because each detected error leads to std::terminate.

Bensk1: I envisioned it to be filtered via select otherwise I imagine cases where the minimal level is frequently changed back and forth.

src/lib/hyrise.hpp
```diff
@@ -38,6 +38,8 @@ class Hyrise : public Singleton<Hyrise> {
   MetaTableManager meta_table_manager;
   SettingsManager settings_manager;
   Topology topology;
+
+  // We need to pass the SettingsManager here to allow the correct registration of LogManager's setting.
```
mrks, line 42: Rather: We need to explicitly pass `settings_manager` as the Hyrise object is not fully initialized yet.

src/lib/utils/log_manager.cpp
Bensk1: > By design, a Hyrise process does not have any loggable errors

Right, there are nor errors in Hyrise.

Bensk1: Use case for log levels: Our current (BP) plugins run periodically and log that they are running, initialized and stopped. In addition, they add another message if they actually adjust the configuration. Plugins produce a lot of log messages, some of them are more important than others.

dey4ss: Yes, I just used the "common" levels used anywhere and expected to have a discussion about them. And due _to let if fail_ I guess we can keep Debug and Info (and maybe warning for some cases).

If we only use select to filter, also the setting can be removed and the SettingsManager can be removed from LogManager's constructor. I think this would be okay when using not more than three log levels (so filtering via select gets a little more handy).

dey4ss: Would a comment in LogManager be enough should there even be a performance warning?

mrks: The performance warning would simply show up the first time you log something. That's not helpful.

If we log everything, this shouldn't be as bad as you will notice the log being flooded and question the performance impact of that.

mrks: Note that filtering by log level is somewhat painful right now. We would need a numerical representation of the log level, too.

dey4ss: When doing it by select statements, yes, one needs to type every level in the where clause, but this could be okay with just a few levels. This was the reason to introduce a setting with a minimal level, as we can compare log levels internally via their enum values.

dey4ss:
Alright, I removed the LogLevelSetting and the LogManager stores all messages. MetaLogTable also provides a `log_level_id` to filter the log.

src/lib/utils/meta_tables/meta_log_table.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+std::shared_ptr<Table> MetaLogTable::_on_generate() const {
+  auto output_table = std::make_shared<Table>(_column_definitions, TableType::Data, std::nullopt, UseMvcc::Yes);
+
+  for (const auto& entry : Hyrise::get().log_manager.log_entries()) {
+    std::ostringstream timestamp;
+    auto time = std::chrono::system_clock::to_time_t(entry.timestamp);
+    struct tm buffer;
+    timestamp << std::put_time(localtime_r(&time, &buffer), "%F %T");
```
mrks, line 28: Oh god. Maybe link to https://stackoverflow.com/questions/25618702/why-is-there-no-c11-threadsafe-alternative-to-stdlocaltime-and-stdgmtime so that people know that this is ugly because of external constraints.

mrks: (Tidy failed around here, too)

src/plugins/mvcc_delete_plugin.cpp
```diff
@@ -76,9 +80,18 @@ void MvccDeletePlugin::_logical_delete_loop() {
...
           std::unique_lock<std::mutex> lock(_mutex_physical_delete_queue);
           _physical_delete_queue.emplace(table, chunk_id);
+          saved_memory += chunk_memory;
+          num_chunks++;
         }
       }
     }
+    if (saved_memory > 0) {
+      std::ostringstream message;
+      double saved_mb = std::round(static_cast<float>(saved_memory) / (10000.0)) / 100.0;
```
Bensk1, line 90: The intention here is only to scale bytes to megabytes? I would find dividing once by `1000000.0` or twice by `1000.0` easier to understand.

mrks: I think the idea is to round it to two decimal places, but std::setprecision would be the better approach.

PR #2099, opened 2020-04-21
Fix ServerTestRunner

### LINKED ISSUE #2088: Fix TestCacheAndSchedulerInitialization
This has failed in a number of branches. Instead of

```
    // Wait to run the server and set the scheduler
    std::this_thread::sleep_for(std::chrono::milliseconds(500));
```

(which I just increased), we should have a proper programmatic approach.

dey4ss:
fixes #2088


mrks:
> Are you alive, gitHub

"Partial outage 1 hrs 20 mins"

Issue #2074, opened 2020-04-03
Prepare for next-gen toolchain


mrks:
<s>Many recent CI fails disappeared after cleaning ccache. I have reason to believe that this fix could apply to us:

> The opt-in “file_macro sloppiness” mode has been removed so that the input file path now is always included in the direct mode hash. This fixes a bug that could result in false cache hits in an edge case when “file_macro sloppiness” is enabled and several identical source files include a relative header file with the same name but in different directories.

Unfortunately, ccache 3.7.6 is only available with Ubuntu Focal, which will be released later this month. We can start now by working on the release candidate.</s> Nope, still broken.

- [x] Build, tag, and push 20.04 Dockerfile
- [x] Update Jenkinsfile
- [x] <s>Reenable ccache (`rm /mnt/ccache/*` on imresx01, which sets the conf file back to default); check if it now works for clang, make sure that it is on tmpfs.</s> #2108 
- [x] Make builds fit for clang-10, add gcc-10 build in CI
- [ ] Make builds fit for gcc-10, add gcc-10 build in CI, at least glance at static analysis
- [x] Add mode for external users that disables -Werror, -Weverything etc.
- [x] Add that mode to https://github.com/hyrise/hyrise/wiki/Step-by-Step-Guide
- [x] Use newer compilers for asan, clang-tidy, etc.
- [x] Update year in LICENSE

Hopefully, this comes without any breaking changes. Currently, I don't see a reason to completely move to a new compiler.

mrks:
gcc-10 is blocked by https://github.com/boostorg/bimap/pull/15.

Taking into account how long it takes for new boost versions to appear in Ubuntu, we will either have to
* forgo gcc-10 for now
* remove bimap
* add that third-party boost PPA again

:(

PR #2100, opened 2020-04-23
Switch CI to Ubuntu 20.04, add clang-10 build

mrks:
* refs #2074
* Adds a ubuntu/clang-10 build in CI but sticks with clang-9 as the default compiler. Mac is unchanged.
* Adds `cmake -DHYRISE_RELAXED_BUILD=On ..`, which disables most warnings.
* I verified that all scripts still work for Ubuntu 19.04, so while it is not our preferred system anymore, we don't break anything.


mrks:
Interestingly, the clang-10 debug build is much slower...

resources/test_data/sqlite_testrunner_queries.sql
```diff
@@ -287,7 +287,7 @@ sELEcT Sum(b + b) AS sum_b_b from mixed;
 
 -- Aggregates with NULL
 SELECT a, MAX(b) FROM mixed_null GROUP BY a;
-SELECT a, MAX(b) FROM mixed_null GROUP BY a ORDER BY MAX(b) DESC;
+SELECT a, MAX(b) FROM mixed_null GROUP BY a ORDER BY MAX(b), a DESC;
```
Bensk1, line 290: Hm?

Bensk1:
> Interestingly, the clang-10 debug build is much slower...

It takes roughly double the time of clang-9. Any idea at what point we are going to switch to clang-10?

Bensk1:
> * I verified that all scripts still work for Ubuntu 19.04, so while it is not our preferred system anymore, we don't break anything.

Nice, we have never done that before :D

resources/test_data/sqlite_testrunner_queries.sql
mrks: If the query contains "ORDER BY", we expect the results to match line-by-line. However, the output order of GROUP BY is undefined, and the new libc++ changed it.

mrks:
> It takes roughly double the time of clang-9. Any idea at what point we are going to switch to clang-10?

No idea. We should have a good reason to.

install_dependencies.sh
```diff
@@ -53,11 +53,18 @@ if echo $REPLY | grep -E '^[Yy]$' > /dev/null; then
         if [ -f /etc/lsb-release ] && cat /etc/lsb-release | grep DISTRIB_ID | grep Ubuntu >/dev/null; then
             echo "Installing dependencies (this may take a while)..."
             if sudo apt-get update >/dev/null; then
+                sudo apt-get install --no-install-recommends -y software-properties-common lsb-release
```
Bensk1, line 56: Should we add lsb-release to the dependencies.md for the sake of completeness?

mrks: Thought about it but decided against it. We are only using it internally in install_dependencies.sh. If you look into dependencies.md, you probably cannot or don't want to run the install script and are looking for what Hyrise needs. lsb-release is not needed for that.

PR #2104, opened 2020-04-26
BenchmarkRunner: Compute real table size

mflueggen:
Why estimate if we can compute the real table size?

### DIFF
--- a/src/benchmarklib/benchmark_runner.cpp
+++ b/src/benchmarklib/benchmark_runner.cpp
@@ -369,10 +369,10 @@ void BenchmarkRunner::_create_report(std::ostream& stream) const {
     benchmarks.push_back(benchmark);
   }
 
-  // Gather information on the (estimated) table size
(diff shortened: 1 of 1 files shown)


mrks:
Because it takes time. After all, there is a good reason why we have the Sampled mode. Are there any significant differences in the reported numbers?

mflueggen:
I wasn't sure if using Sampled mode was done intentionally or if you just forgot to update that line after Full mode was introduced.

My VM, TPC-H, scale factor = 1:
Sampled table_Size: 1073284941 (computation time: 25 ms)
Full table_size: 964480078 (computation time: 200 ms)
That’s a discrepancy of ~9%. I was confused that the table_size in the json file did not match the size I expected.

Considering that initializing the TPC-H benchmark takes around 30 s one could argue that those additional 200 ms will not be noticeable.

mrks:
That's much less than I expected. Sounds reasonable.

PR #2105, opened 2020-04-27
Prevent Meta Table Modification Caching

### LINKED ISSUE #2049: PlanCache: caching modifications on meta tables leads to crashing

dey4ss:
fixes  #2049

### DIFF
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -343,7 +343,10 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_insert(const hsql::In
   std::shared_ptr<Table> target_table;
   if (is_meta_table) {
     AssertInput(Hyrise::get().meta_table_manager.can_insert_into(table_name), "Cannot insert into " + table_name);
-    target_table = Hyrise::get().meta_table_manager.generate_table(table_name);
+    auto sql_identifier_resolver =
+        _sql_identifier_resolver ? _sql_identifier_resolver : std::make_shared<SQLIdentifierResolver>();
+    _translate_meta_table(table_name, sql_identifier_resolver);
+    target_table = (*_meta_tables)[_trim_meta_table_name(table_name)];
   } else {
     AssertInput(Hyrise::get().storage_manager.has_table(table_name),
                 std::string{"Did not find a table with name "} + table_name);
@@ -491,7 +494,7 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_update(const hsql::Up
   std::shared_ptr<Table> target_table;
   if (is_meta_table) {
     AssertInput(Hyrise::get().meta_table_manager.can_update(table_name), "Cannot update " + table_name);
-    target_table = Hyrise::get().meta_table_manager.generate_table(table_name);
+    target_table = (*_meta_tables)[_trim_meta_table_name(table_name)];
   } else {
     AssertInput(Hyrise::get().storage_manager.has_table(table_name),
                 std::string{"Did not find a table with name "} + table_name);
@@ -733,7 +736,7 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_meta_table(
   // that can change at any time
   _cacheable = false;
 
-  const auto meta_table_name = name.substr(MetaTableManager::META_PREFIX.size());
+  const auto meta_table_name = _trim_meta_table_name(name);
 
   // Meta tables are integrated in the LQP as static table nodes in order to avoid regeneration at every
   // access in the pipeline afterwards.
@@ -1837,6 +1840,10 @@ std::vector<std::shared_ptr<AbstractExpression>> SQLTranslator::_unwrap_elements
     expressions.emplace_back(element.expression);
   }
   return expressions;
+}
+
+std::string SQLTranslator::_trim_meta_table_name(const std::string& name) {
+  return name.substr(MetaTableManager::META_PREFIX.size());
 }
 
 SQLTranslator::SelectListElement::SelectListElement(const std::shared_ptr<AbstractExpression>& init_expression)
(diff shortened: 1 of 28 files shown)


src/test/utils/meta_tables/meta_table_test.cpp
```diff
@@ -138,4 +144,18 @@ TEST_P(MultiMetaTablesTest, SQLFeatures) {
...
+  EXPECT_EQ(mock_table->generate_calls(), 0);
+  SQLPipelineBuilder{"SELECT * FROM meta_mock"}.create_pipeline().get_result_table();
+  EXPECT_EQ(mock_table->generate_calls(), 1);
+  SQLPipelineBuilder{"DELETE FROM meta_mock WHERE mock='abc'"}.create_pipeline().get_result_table();
+  EXPECT_EQ(mock_table->generate_calls(), 2);
+  SQLPipelineBuilder{"INSERT INTO meta_mock VALUES('foo')"}.create_pipeline().get_result_table();
+  EXPECT_EQ(mock_table->generate_calls(), 3);
+  SQLPipelineBuilder{"UPDATE meta_mock SET mock='foo'"}.create_pipeline().get_result_table();
+  EXPECT_EQ(mock_table->generate_calls(), 4);
+}
```
mrks, line 160: Could you (1) make sure that the LQP and PQP caches are enabled and (2) repeat calls to make sure that even with caching, the tables are regenerated?

src/lib/sql/sql_translator.cpp
```diff
@@ -342,8 +342,11 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_insert(const hsql::In
 
   std::shared_ptr<Table> target_table;
   if (is_meta_table) {
+    auto sql_identifier_resolver =
+        _sql_identifier_resolver ? _sql_identifier_resolver : std::make_shared<SQLIdentifierResolver>();
+    _translate_meta_table(table_name, sql_identifier_resolver);
     AssertInput(Hyrise::get().meta_table_manager.can_insert_into(table_name), "Cannot insert into " + table_name);
-    target_table = Hyrise::get().meta_table_manager.generate_table(table_name);
+    target_table = (*_meta_tables)[_trim_meta_table_name(table_name)];
```
mrks, line 349: `operator[]` on maps is a scary thing. Rather use `at()` (more than once in this file).

src/lib/sql/sql_translator.cpp
```diff
@@ -727,13 +730,13 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_stored_table(
 
 std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_meta_table(
     const std::string& name, const std::shared_ptr<SQLIdentifierResolver>& sql_identifier_resolver) {
-  AssertInput(MetaTableManager::is_meta_table_name(name), std::string{"Did not find a meta table with name "} + name);
+  AssertInput(Hyrise::get().meta_table_manager.has_table(name), std::string{"Did not find a table with name "} + name);
```
mrks, line 733: good catch

src/lib/sql/sql_translator.cpp
```diff
@@ -1839,6 +1842,10 @@ std::vector<std::shared_ptr<AbstractExpression>> SQLTranslator::_unwrap_elements
   return expressions;
 }
 
+std::string SQLTranslator::_trim_meta_table_name(const std::string& name) {
+  return name.substr(MetaTableManager::META_PREFIX.size());
```
mrks, line 1846: DebugAssert that the prefix is actually there?

src/lib/utils/meta_tables/meta_tables_table.hpp
```diff
@@ -15,7 +15,7 @@ class MetaTablesTable : public AbstractMetaTable {
  protected:
   friend class MetaTableManager;
 
-  std::shared_ptr<Table> _on_generate() const final;
+  std::shared_ptr<Table> _on_generate() final;
```
mrks, line 18: Where did the removal of const become necessary? I must have missed something in the diff. Generally, I worry that a non-const generate method leads to state being stored in the table and thus, the generations not being independent.

src/test/sql/sql_translator_test.cpp
```diff
@@ -2492,6 +2492,10 @@ TEST_F(SQLTranslatorTest, CatchInputErrors) {
   EXPECT_THROW(compile_query("WITH q AS (SELECT * FROM int_float), q AS (SELECT b FROM q) SELECT * FROM q;"),
                InvalidInputException);
   EXPECT_THROW(compile_query("COPY no_such_table TO 'a_file.tbl';"), InvalidInputException);
+  EXPECT_THROW(compile_query("SELECT * FROM meta_unknown;"), InvalidInputException);
+  EXPECT_THROW(compile_query("UPDATE meta_unknown SET a = 1;"), InvalidInputException);
+  EXPECT_THROW(compile_query("DELETE FROM meta_unknown;"), InvalidInputException);
+  EXPECT_THROW(compile_query("INSERT INTO meta_unknown (a) VALUES(1);"), InvalidInputException);
```
mrks, line 2498: Do we have similar tests for DML on non-modifiable meta tables?

src/lib/utils/meta_tables/meta_tables_table.hpp
dey4ss: I added a count for the generate method of the mock table to verify it is not generated multiple times.

src/test/sql/sql_translator_test.cpp
dey4ss: This should be covered [here](https://github.com/hyrise/hyrise/blob/b6d92331c43d469eb4f07c3df2beb3db44b87c20/src/test/sql/sql_translator_test.cpp#L1951).

src/lib/utils/meta_tables/meta_tables_table.hpp
mrks: So the non-constness is just needed in the test? In that case, simply mark the member as mutable.

PR #2107, opened 2020-04-28
Properly protect against empty chunks

mrks:
In #2085, I found that the aggregate operator created empty chunks, which the sort did not handle properly. This tightens the rules for empty chunks, allowing them only as the last, mutable chunk for data tables. It also fixes the operators that were violating this new rule.

### DIFF
--- a/src/lib/storage/table.cpp
+++ b/src/lib/storage/table.cpp
@@ -238,6 +238,16 @@ void Table::append_chunk(const Segments& segments, std::shared_ptr<MvccData> mvc
       const auto is_reference_segment = std::dynamic_pointer_cast<ReferenceSegment>(segment) != nullptr;
       Assert(is_reference_segment == (_type == TableType::References), "Invalid Segment type");
     }
+
+    // Check that existing chunks are not empty
+    const auto chunk_count = _chunks.size();
+    for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+      const auto chunk = get_chunk(chunk_id);
+      if (!chunk) continue;
+
+      // An empty, mutable chunk at the end is fine, but in that case, append_chunk shouldn't have to be called.
+      DebugAssert(chunk->size() > 0, "append_chunk called on a table that has an empty chunk");
+    }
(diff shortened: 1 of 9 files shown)


src/test/operators/print_test.cpp
```diff
@@ -311,13 +311,6 @@ TEST_F(OperatorsPrintTest, SegmentType) {
 
 TEST_F(OperatorsPrintTest, EmptyTable) {
   auto tab = Hyrise::get().storage_manager.get_table(_table_name);
-  Segments empty_segments;
-  auto column_1 = std::make_shared<opossum::ValueSegment<int>>();
-  auto column_2 = std::make_shared<opossum::ValueSegment<pmr_string>>();
-  empty_segments.push_back(column_1);
-  empty_segments.push_back(column_2);
-  tab->append_chunk(empty_segments);
```
mweisgut, line 319: If the rules for empty chunks shall be tightened, adding two consecutive empty segments should not be allowed, right? Do we expect a failure in this scenario? If so, how about adding an `EXPECT_THROW` test case here?
Currently, this scenario doesn't cause a failure. Adding an assertion in `Table::append_chunk` as it was done in the table constructor could solve this issue.

src/lib/storage/table.cpp
```diff
@@ -236,6 +238,16 @@ void Table::append_chunk(const Segments& segments, std::shared_ptr<MvccData> mvc
       const auto is_reference_segment = std::dynamic_pointer_cast<ReferenceSegment>(segment) != nullptr;
       Assert(is_reference_segment == (_type == TableType::References), "Invalid Segment type");
     }
+
+    // Check that existing chunks are not empty
+    const auto chunk_count = _chunks.size();
+    for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
```
mweisgut, line 244: Wouldn't it be sufficient to check if only the last existing chunk is empty?

mrks: Probably, but since this is done only for debug builds, we don't lose anything.

PR #2110, opened 2020-04-28
Remove NULLS LAST mode

### LINKED ISSUE #2095: Get rid of NULLS LAST sort modes
See this discussion: https://github.com/hyrise/hyrise/pull/2085#discussion_r410450790

mrks:
fixes #2095

Mostly search and replace, hope I didn't miss anything.


mrks:
Blocked by #2000. Well, actually it is not, but merging this would cause additional conflicts over there. After my last review, @Bouncner threatened bodily harm, so I'm not going risk it ;)

Bouncner:
#2000 will add a short todo in chunk.cpp concerning this PR.

src/lib/operators/sort.cpp
```diff
@@ -341,21 +341,17 @@ class Sort::SortImpl {
...
     // 2b. Insert null rows if necessary
     if (!_null_value_rows.empty()) {
-      if (_order_by_mode == OrderByMode::AscendingNullsLast || _order_by_mode == OrderByMode::DescendingNullsLast) {
-        // NULLs last
-        _row_id_value_vector.insert(_row_id_value_vector.end(), _null_value_rows.begin(), _null_value_rows.end());
-      } else {
-        // NULLs first (default behavior)
-        _row_id_value_vector.insert(_row_id_value_vector.begin(), _null_value_rows.begin(), _null_value_rows.end());
-      }
+      // NULLs come before all values. The SQL standard allows for this to be implementation-defined. We used to have
```
Bouncner, line 352: That's kind of a strange definition in my opinion. I'd vote for doing the same as HANA and MySQL (actually does not matter if we join the larger or smaller DB community, I just don't like the handling of HSQLDB) as shown on the site Jan linked: https://docs.mendix.com/refguide/null-ordering-behavior

mrks: The advantage of the NULLS FIRST approach is simply that we need less code. Both in the sort and in code that handles sorted data (e.g., scan_sorted_segment).

Bouncner: Hmm, I don't like it. But than again sorting with NULLs sucks anyways. So when you say it's easier, I trust you here and that weights more.

Bouncner: Add a comment to types.hpp? I think it's rather hidden right now.

mrks:
"Running on rapa in /home/jenkins/workspace/hyrise_hyrise_PR-2110" - I wonder if the TSAN failures are limited to rapa.

PR #2111, opened 2020-04-29
Stop SQLPipeline from leaking the SQLPipelineStatement(s)

### LINKED ISSUE #1893: Refactor LQPColumnExpression
## Do we need both LQPColumnExpression and -Reference?

This came up in a discussion with @mrzzzrm. Independent of the original discussion, we should have a look if there is any point in the LQPColumnReference. When scrolling through the code, I got the feeling that the LQPColumnReferences only exist to be wrapped into LQPColumnExpressions and unwrapped later.

## Rename AbstractLQPNode::column_expressions to ::output_expressions?

The former sounds too close to LQPColumnExpression. Also, the "column" part is misleading / redundant.

### LINKED ISSUE #1905: Refactor create_pipeline_statement
The way it is currently used in the master and in the upcoming server is weird

mrks:
fixes #1905

The whole idea of the SQLPipeline is to have a one stop shop for turning SQL into tables. Handing out individual pipeline statements is unnecessary. With #1993, where the transaction_context may change during the execution of a pipeline, this becomes more important as running pipeline statements individually may lead to inconsistencies.

Also fixes prepared statements in the server not being optimized.


src/test/server/query_handler_test.cpp
```diff
@@ -32,11 +33,17 @@ TEST_F(QueryHandlerTest, CreatePreparedPlan) {
...
 
   const auto result = QueryHandler::bind_prepared_plan(specification);
-  EXPECT_EQ(result->type(), OperatorType::TableScan);
+  EXPECT_EQ(result->type(), OperatorType::Validate);
+
+  const auto get_table = std::dynamic_pointer_cast<const GetTable>(result->input_left()->input_left());
+  ASSERT_TRUE(get_table);
+
+  // Check that the optimizer was executed
+  ASSERT_FALSE(get_table->pruned_chunk_ids().empty());
```
Bensk1, line 46: That seems a bit dirty but is probably the only way to test it without mocking?

src/lib/server/query_handler.cpp
```diff
@@ -81,8 +85,13 @@ std::shared_ptr<AbstractOperator> QueryHandler::bind_prepared_plan(const Prepare
         std::make_shared<ValueExpression>(statement_details.parameters[parameter_idx]);
   }
 
-  const auto lqp = prepared_plan->instantiate(parameter_expressions);
-  return LQPTranslator{}.translate_node(lqp);
+  auto lqp = prepared_plan->instantiate(parameter_expressions);
+  const auto optimizer = Optimizer::create_default_optimizer();
```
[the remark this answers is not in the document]
mrks: This is consistent with `SQLPipelineBuilder::create_pipeline_statement`. Probably not necessary, but hasn't popped up as a performance issue yet.

src/test/server/query_handler_test.cpp
mrks: What part about this feels dirty to you? If there are chunks pruned, the optimizer was called.

Bensk1: The part that we assume that the optimizer was called from a vector not being empty.

mrks: We could run the test with a different parameter and check that a different chunk was pruned. Unless we expect any other component to change the pruned_chunk_ids, that should unambiguously show that the optimizer was executed.

Bensk1: I don't think that we have to do that. The current solution is not too bad, it just seems a bit weird to assert that the optimizer was running based on a data structure that changed in some operator. It feels strange that we would have to adapt that test if we would disable the chunk pruning rule or that the test would fail if we adapted the predicate. But yes, there is probably no better way to do it right now.

mrks: > It feels strange that we would have to adapt that test if we would disable the chunk pruning rule or that the test would fail if we adapted the predicate

Ok, agreed. I'll add a comment about that.

PR #2113, opened 2020-04-30
Remove LQPColumnReference

mrks:
LQPColumnReferences only exist to be wrapped into LQPColumnExpressions and unwrapped later. Seems pointless.

refs #1893, but does not fix it yet. I still want to rename `AbstractLQPNode::column_expressions` to `::output_expressions`, but that should be a separate PR.


src/lib/expression/expression_utils.cpp
```diff
@@ -108,15 +108,13 @@ void expression_adapt_to_different_lqp(std::shared_ptr<AbstractExpression>& expr
 
 std::shared_ptr<LQPColumnExpression> expression_adapt_to_different_lqp(const LQPColumnExpression& lqp_column_expression,
                                                                        const LQPNodeMapping& node_mapping) {
-  const auto node = lqp_column_expression.column_reference.original_node();
+  const auto node = lqp_column_expression.original_node.lock();
+  Assert(node, "LQPColumnExpression is expired");
```
mweisgut, line 112: Just for understanding: in which cases/how can `LQPColumnExpressions` be expired?

src/lib/expression/expression_utils.cpp
```diff
@@ -108,15 +108,13 @@ void expression_adapt_to_different_lqp(std::shared_ptr<AbstractExpression>& expr
 
 std::shared_ptr<LQPColumnExpression> expression_adapt_to_different_lqp(const LQPColumnExpression& lqp_column_expression,
                                                                        const LQPNodeMapping& node_mapping) {
-  const auto node = lqp_column_expression.column_reference.original_node();
+  const auto node = lqp_column_expression.original_node.lock();
```
mweisgut, line 111: `.original_node.lock()` is now frequently present in the code. How about declaring the `LQPColumnExpression`'s member variable `const std::weak_ptr<const AbstractLQPNode> original_node;` as protected and adding `std::shared_ptr<const AbstractLQPNode> LQPColumnExpression::original_node() const { return _original_node.lock(); }` so we can call `original_node()` at these locations?

src/lib/expression/expression_utils.cpp
mrks: > If the original_node is not referenced by any shared_ptr anymore, it is deleted and the weak_ptr is expired. This should never be the case. If `original_node.lock()` is `nullptr`, the LQP is defective.

Would that help as a comment in lqp_column_expression.hpp?

src/lib/expression/expression_utils.cpp
mrks: That's the way it was before. People thinking that it was simply an accessor called `original_node()` more than once (e.g., in lqp_column_expression.cpp). By forcing them to call `lock()`, I hope to avoid that unnecessary overhead.

src/lib/expression/expression_utils.cpp
mweisgut: It's basically the nature of a weak pointer and it's ´lock´ function, however, I think it helps.

src/lib/expression/lqp_column_expression.cpp
```diff
@@ -11,53 +11,74 @@
...
+  if (mode == AbstractExpression::DescriptionMode::Detailed) {
+    output << original_node_locked << ".";
+  }
+
+  if (original_column_id == INVALID_COLUMN_ID) {
+    output << "INVALID_COLUMN_ID";
+    return output.str();
+  }
+
+  if (original_node_locked->type == LQPNodeType::StoredTable) {
```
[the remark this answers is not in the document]
mrks: Good point.

PR #2117, opened 2020-05-04
Fix two expression issues

mrks:
* For BETWEEN expressions where one of the bounds is not a column, we correctly fall back to the ExpressionEvaluator. Before this PR, we still expected both to be PQPColumnExpressions in one place.
* When trying to `SUM(string_column)`, a non-descript exception was thrown later on. We now catch this in the SQLTranslator.

### DIFF
--- a/resources/test_data/sqlite_testrunner_queries.sql
+++ b/resources/test_data/sqlite_testrunner_queries.sql
@@ -81,7 +81,7 @@ SELECT * FROM id_int_int_int_100 WHERE 91 > a AND 20 <= a;
 SELECT * FROM id_int_int_int_100 WHERE 91 >= a AND 20 < a;
 SELECT * FROM id_int_int_int_100 WHERE 91 > a AND 20 < a;
 
--- Scans with BETWEEN that cannot be handled by ColumnBetweenTableScanImpl
+-- Scans with BETWEEN that cannot be handled by ColumnBetweenTableScanImpl, which supports scalar predicates only
 SELECT * FROM mixed WHERE 10 BETWEEN b AND 40
 SELECT * FROM mixed WHERE c BETWEEN b AND 100
 SELECT * FROM mixed WHERE b + 10 BETWEEN b AND c
--- a/resources/test_data/sqlite_testrunner_queries.sql
+++ b/resources/test_data/sqlite_testrunner_queries.sql
@@ -81,6 +81,11 @@ SELECT * FROM id_int_int_int_100 WHERE 91 > a AND 20 <= a;
 SELECT * FROM id_int_int_int_100 WHERE 91 >= a AND 20 < a;
 SELECT * FROM id_int_int_int_100 WHERE 91 > a AND 20 < a;
 
+-- Scans with BETWEEN that cannot be handled by ColumnBetweenTableScanImpl
+SELECT * FROM mixed WHERE 10 BETWEEN b AND 40
+SELECT * FROM mixed WHERE c BETWEEN b AND 100
+SELECT * FROM mixed WHERE b + 10 BETWEEN b AND c
+
 -- Scans with potential for predicate pruning
 SELECT * FROM id_int_int_int_100 WHERE a >= 20 AND a <= 40 OR b >= 50 AND b <= 95;
 SELECT * FROM id_int_int_int_100 WHERE a >= 20 AND a <= 40 AND c <= 35 AND b >= 49 AND a >= 21 AND b <= 95 AND c <= 40 AND c >= 23;
--- a/src/lib/operators/table_scan.cpp
+++ b/src/lib/operators/table_scan.cpp
@@ -300,8 +300,8 @@ std::unique_ptr<AbstractTableScanImpl> TableScan::create_impl() const {
 
     auto lower_bound_value = expression_get_value_or_parameter(*between_expression->lower_bound());
     if (lower_bound_value) {
-      const auto adjusted_predicate_and_value =
-          lossless_predicate_variant_cast(lower_condition, *lower_bound_value, left_column->data_type());
+      const auto adjusted_predicate_and_value = lossless_predicate_variant_cast(
+          lower_condition, *lower_bound_value, between_expression->value()->data_type());
       if (adjusted_predicate_and_value) {
         lower_condition = adjusted_predicate_and_value->first;
         lower_bound_value = adjusted_predicate_and_value->second;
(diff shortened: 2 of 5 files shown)


src/lib/operators/table_scan.cpp
```diff
@@ -300,8 +300,8 @@ std::unique_ptr<AbstractTableScanImpl> TableScan::create_impl() const {
 
     auto lower_bound_value = expression_get_value_or_parameter(*between_expression->lower_bound());
     if (lower_bound_value) {
-      const auto adjusted_predicate_and_value =
-          lossless_predicate_variant_cast(lower_condition, *lower_bound_value, left_column->data_type());
+      const auto adjusted_predicate_and_value = lossless_predicate_variant_cast(
+          lower_condition, *lower_bound_value, between_expression->value()->data_type());
       if (adjusted_predicate_and_value) {
         lower_condition = adjusted_predicate_and_value->first;
         lower_bound_value = adjusted_predicate_and_value->second;
```
julianmenzler, line 307: I know BETWEEN statements, but this implementation seemed Greek to me. 

1. Split up predicate:`auto [lower_condition, upper_condition] = between_to_conditions(predicate_condition);`
2. Some stuff -> **please comment**
3. Bringt it back together again: `conditions_to_between(lower_condition, upper_condition)`

src/lib/sql/sql_translator.cpp
```diff
@@ -1574,8 +1575,16 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_translate_hsql_expr(
               aggregate_expression = std::make_shared<AggregateExpression>(
                   aggregate_function, _translate_hsql_expr(*expr.exprList->front(), sql_identifier_resolver));
             }
+          } break;
         }
 
+        // Check that the aggregate can be calculated on the given expression
+        const auto aggregate_data_type = aggregate_expression->data_type();
+        AssertInput(aggregate_data_type != DataType::Null,
+                    std::string{"Invalid aggregate "} + aggregate_expression->as_column_name() +
```
julianmenzler, line 1584: Why do you wrap "Invalid aggregate " with `std::string{..}`? For " for input data type " you did not do it

src/lib/operators/table_scan.cpp
julianmenzler, line 311: I know BETWEEN statements, but this implementation seemed Greek to me. 

1. Split up predicate:`auto [lower_condition, upper_condition] = between_to_conditions(predicate_condition);`
2. Some modifications -> **please comment what and why**
3. Bringt it back together again: `conditions_to_between(lower_condition, upper_condition)`

src/test/operators/table_scan_test.cpp
```diff
@@ -1020,6 +1020,13 @@ TEST_P(OperatorsTableScanTest, GetImpl) {
     EXPECT_EQ(impl->right_value, AllTypeVariant{4.099999904632568359375f});
   }
 
+  {
+    const auto abstract_impl = TableScan{get_int_float_op(), between_exclusive_(column_a, column_b, 3.1)}.create_impl();
```
julianmenzler, line 1024: Would it make sense to add a swapped version as well?
E.g.: 

```suggestion
     const auto abstract_impl = TableScan{get_int_float_op(), between_exclusive_(column_a, 3.1, column_b)}.create_impl();
```

resources/test_data/sqlite_testrunner_queries.sql
```diff
@@ -81,6 +81,11 @@ SELECT * FROM id_int_int_int_100 WHERE 91 > a AND 20 <= a;
 SELECT * FROM id_int_int_int_100 WHERE 91 >= a AND 20 < a;
 SELECT * FROM id_int_int_int_100 WHERE 91 > a AND 20 < a;
 
+-- Scans with BETWEEN that cannot be handled by ColumnBetweenTableScanImpl
```
julianmenzler, line 84:
```suggestion
-- Scans with BETWEEN that cannot be handled by ColumnBetweenTableScanImpl (because it supports scalar predicates only)
```

src/lib/sql/sql_translator.cpp
mrks: You need an std::string to start using `operator+`. At least it was that way a while ago.

mrks:
Weird. What happened here? Before, I only saw one comment.

src/lib/operators/table_scan.cpp
mrks: Good point. Does this help?

```
The ColumnBetweenTableScanImpl expects both values to be of the same data type as the column that is being
scanned. We retrieve the lower and upper bounds of the BetweenExpression, perform a
lossless_predicate_variant_cast into the column's data type and reassemble the between condition.
```

src/test/operators/table_scan_test.cpp
mrks: Doesn't hurt I guess

src/lib/operators/table_scan.cpp
julianmenzler: > Good point. Does this help?
> 
> ```
> The ColumnBetweenTableScanImpl expects both values to be of the same data type as the column that is being
> scanned. We retrieve the lower and upper bounds of the BetweenExpression, perform a
> lossless_predicate_variant_cast into the column's data type and reassemble the between condition.
> ```

Yes! However, I wonder why we need such a cast?

julianmenzler:
> Weird. What happened here? Before, I only saw one comment.

Well, my fault. I deleted the comment and re-added it as part of the review. You were too fast in replying. Sorry! 😅

src/lib/operators/table_scan.cpp
mrks: Added

> Many implementations require the comparison values to be of the same column type. If we were to cast the values to the column type, `int_column = 16.25` would turn into `int_column = 16`, which is obviously wrong. As such, we use lossless casts to guarantee safe type conversions. This was introduced by #1550.

Does that help?

src/lib/sql/sql_translator.cpp
julianmenzler: Interesting. I assumed today's compilers were clever enough to figure this out themselves.

src/lib/operators/table_scan.cpp
julianmenzler: Perfect!

mrks: Thanks for bringing that up. That would have easily gotten lost over time.

mrks:
Ready to go then?

PR #2118, opened 2020-05-05
Console: Use proper transaction handling

### LINKED ISSUE #2109: hyriseConsole: Use new transaction handling
We should use the transaction handling introduced by #1993 instead of having explicit `begin` etc. commands.

As part of this, remove `SQLPipeline::failed_pipeline_statement` or make it private. We shouldn't expose single SQLPipelineStatements (see #2111).

mrks:
Instead of having its own hacky hacky functions, the console now uses the transaction handling as introduced in #1993.

fixes #2109


mrks:
This was even easier than expected :)

scripts/test/hyriseConsole_test.py
```diff
@@ -62,6 +62,22 @@ def main():
...
+	console.sendline("select sum(a) from test")
+	console.expect("860")
+	console.sendline("txinfo")
+	console.expect("Active transaction")
+	console.sendline("insert into test (a) values (21); rollback;")
+	console.sendline("select sum(a) from test")
+	console.expect("840")
+	console.sendline("txinfo")
+	console.expect("auto-commit mode")
+
```
klauck, line 80: Should we also test invalid schedules of transaction statements, e.g., `begin;begin;`?

mrks:
I will simply remove all the code that deals with transaction conflicts. That cannot occur in the console.

mrks:
@Bensk1 - just remembered that @klauck isn't working right now. Could you take over, please? He already looked at the transaction part, so there shouldn't be too much left over.

PR #2120, opened 2020-05-05
Query plan visualization: use of thousands separators

mweisgut:
This PR integrates the usage of thousands separators in the LQP and PQP visualizations for the row (and chunk) count.

LQP example
TPC-H_07-LQP_excerpt

PQP example
TPC-H_07-PQP_excerpt

### DIFF
--- a/src/lib/visualization/abstract_visualizer.hpp
+++ b/src/lib/visualization/abstract_visualizer.hpp
@@ -46,6 +46,7 @@ struct VizEdgeInfo {
   std::string arrowhead = "normal";
 };
 
+// Custom facet for creating a custom locale with thousands separator.
 struct SeparateThousands : std::numpunct<char> {
   char_type do_thousands_sep() const override { return ','; }  // separate with commas
   string_type do_grouping() const override { return "\3"; }    // groups of 3 digits
--- a/src/lib/visualization/abstract_visualizer.hpp
+++ b/src/lib/visualization/abstract_visualizer.hpp
@@ -48,7 +48,6 @@ struct VizEdgeInfo {
 
 // Custom facet for creating a custom locale with thousands separator.
 struct SeparateThousands : std::numpunct<char> {
-  char_type do_thousands_sep() const override { return ','; }  // separate with commas
   string_type do_grouping() const override { return "\3"; }    // groups of 3 digits
 };
 
--- a/src/lib/visualization/abstract_visualizer.hpp
+++ b/src/lib/visualization/abstract_visualizer.hpp
@@ -47,7 +47,7 @@ struct VizEdgeInfo {
 };
 
 // Custom facet for creating a custom locale with thousands separator.
-struct SeparateThousands : std::numpunct<char> {
+struct SeparateThousandsFacet : std::numpunct<char> {
   string_type do_grouping() const override { return "\3"; }    // groups of 3 digits
 };
 
--- a/src/lib/visualization/abstract_visualizer.hpp
+++ b/src/lib/visualization/abstract_visualizer.hpp
@@ -46,6 +46,11 @@ struct VizEdgeInfo {
   std::string arrowhead = "normal";
 };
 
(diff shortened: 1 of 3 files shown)


src/lib/visualization/abstract_visualizer.hpp
```diff
@@ -46,6 +46,12 @@ struct VizEdgeInfo {
   std::string arrowhead = "normal";
 };
 
+// Custom facet for creating a custom locale with thousands separator.
```
mrks, line 49: Surprised to see that this has to be manually done

src/lib/visualization/lqp_visualizer.cpp
```diff
@@ -113,6 +113,12 @@ void LQPVisualizer::_build_dataflow(const std::shared_ptr<AbstractLQPNode>& from
   }
 
   std::ostringstream label_stream;
+
+  // use a copy of the stream's default locale with thousand separator
+  const auto& separate_thousands_locale =
+      std::locale(label_stream.getloc(), std::make_unique<SeparateThousands>().release());
```
mrks, line 119: This is UB. The object held by the unique_ptr will be destroyed before you access it. Simply make it a stack variable and take its address if the STL insists on getting a pointer.

src/lib/visualization/pqp_visualizer.cpp
```diff
@@ -143,8 +143,14 @@ void PQPVisualizer::_build_dataflow(const std::shared_ptr<const AbstractOperator
   const auto& performance_data = from->performance_data();
   if (performance_data.executed && performance_data.has_output) {
     std::stringstream stream;
-    stream << std::to_string(performance_data.output_row_count) + " row(s)/";
-    stream << std::to_string(performance_data.output_chunk_count) + " chunk(s)";
+
+    // use a copy of the stream's default locale with thousand separator
+    const auto& separate_thousands_locale =
+        std::locale(stream.getloc(), std::make_unique<SeparateThousands>().release());
```
mrks, line 149: same

Bouncner:
The LQP appears to have *.2 rows.

mweisgut:
> The LQP appears to have *.2 rows.
But this is also the case for the current master branch. That was not introduced with this PR.

mrks:
That's just the estimation doing its job.

src/lib/visualization/lqp_visualizer.cpp
mweisgut: This leads to a seg fault:
```c++
SeparateThousands separate_thousands;
const auto& separate_thousands_locale = std::locale(label_stream.getloc(), &separate_thousands);
label_stream.imbue(separate_thousands_locale);
```

This works:
```c++
const auto& separate_thousands_locale = std::locale(label_stream.getloc(), new SeparateThousands);
label_stream.imbue(separate_thousands_locale);
```

I haven't found a better solution yet.

Some resources that also use the `new` solution:
https://en.cppreference.com/w/cpp/locale/numpunct
https://stackoverflow.com/a/7277333
https://stackoverflow.com/a/17530535

mrks: Wow. std::locale::facet does internal reference counting. I am... slightly disgusted.

In that case, `new` seems like the <s>best</s> least awful approach. Please make sure to document that the STL forces this atrocity on us.

mrks: https://www.reddit.com/r/cpp/comments/4aaorj/questions_regarding_stdcodecvt/

mrks:
I like the idea though :)

mrks:
> To https://github.com/mweisgut/hyrise.git
 ! [remote rejected]     feature/pqp_lqp_thousand_separators -> feature/pqp_lqp_thousand_separators (permission denied)
error: failed to push some refs to 'https://github.com/mweisgut/hyrise.git'

Guess you'll have to do the merge yourself then ;)

PR #2123, opened 2020-05-06
CI: Use clang-10

### LINKED ISSUE #2074: Prepare for next-gen toolchain

mrks:
Even if clang-10 is not our default version yet, we should change most builds so that we have newer versions of asan, tsan, etc.

We still have a clang-9 debug build and continue to use clang-9 for Mac.

closes #2074, as we will not be able to support gcc-10 any time soon. :(

TODO:
- [x] Rebuild docker image


mrks:
nope

scripts/clang_tidy_wrapper.sh
```diff
@@ -14,6 +14,9 @@ if grep "$file_relative_to_source" $cmake_source_dir/.clang-tidy-ignore > /dev/n
 	echo "clang-tidy: Ignoring $file_relative_to_source"
 	exit 0
 else
-	exec clang-tidy $file $@
-  exit $?
+	if [ -x /usr/local/bin/brew ]; then
+		exec $(brew --prefix llvm)/bin/clang-tidy $file $@
```
Bouncner, line 18: Does apple ship it's own clang tidy or what is the reason for this change?

Bouncner: Also, why execute v9 when brew is used?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -400,10 +400,10 @@ cxxopts::Options BenchmarkRunner::get_basic_cli_options(const std::string& bench
   // some point. The way this is solved here is not really nice, but as the TPC-C benchmark binary has just a main
   // method and not a class, retrieving this default value properly would require some major refactoring of how
   // benchmarks interact with the BenchmarkRunner. At this moment, that does not seem to be worth the effort.
-  const auto default_mode = (benchmark_name == "TPC-C Benchmark" ? "Shuffled" : "Ordered");
+  const auto* const default_mode = (benchmark_name == "TPC-C Benchmark" ? "Shuffled" : "Ordered");
 
   // TPC-C does not support binary caching
-  const auto default_dont_cache_binary_tables = (benchmark_name == "TPC-C Benchmark" ? "true" : "false");
+  const auto* const default_dont_cache_binary_tables = (benchmark_name == "TPC-C Benchmark" ? "true" : "false");
```
Bouncner, line 406: What? A const pointer to a const object?

src/lib/operators/abstract_join_operator.cpp
```diff
@@ -56,7 +56,7 @@ std::string AbstractJoinOperator::description(DescriptionMode description_mode)
     return "Column #"s + std::to_string(column_id);
   };
 
-  const auto separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
+  const auto* const separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
```
Bouncner, line 59: I don't get why this is a pointer?

src/lib/operators/join_index.cpp
```diff
@@ -64,8 +64,8 @@ const std::string& JoinIndex::name() const {
 }
 
 std::string JoinIndex::description(DescriptionMode description_mode) const {
-  const auto separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
-  const auto index_side_str = _index_side == IndexSide::Left ? "Left" : "Right";
+  const auto* const separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
+  const auto* const index_side_str = _index_side == IndexSide::Left ? "Left" : "Right";
```
Bouncner, line 68: Same as above.

src/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_nodes.cpp
```diff
@@ -141,8 +141,8 @@ AbstractIndex::Iterator ARTNode16::_delegate_to_child(
     const std::function<Iterator(std::iterator_traits<std::array<uint8_t, 16>::iterator>::difference_type, size_t)>&
         function) const {
   auto partial_key = key[depth];
-  auto partial_key_iterator = std::lower_bound(_partial_keys.begin(), _partial_keys.end(), partial_key);
-  auto partial_key_pos = std::distance(_partial_keys.begin(), partial_key_iterator);
+  const auto* const partial_key_iterator = std::lower_bound(_partial_keys.begin(), _partial_keys.end(), partial_key);
```
Bouncner, line 144: Iterator as a pointer?

src/lib/storage/vector_compression/simd_bp128/simd_bp128_packing.hpp
```diff
@@ -21,8 +21,8 @@ class SimdBp128Packing {
   static constexpr auto meta_block_size = block_size * blocks_in_meta_block;
 
  public:
-  static void write_meta_info(const uint8_t* in, uint128_t* out);
-  static void read_meta_info(const uint128_t* in, uint8_t* out);
+  static void write_meta_info(const uint8_t* in, uint128_t* const out);
```
Bouncner, line 24: Isn't in double const and the output just const on the pointer?

scripts/clang_tidy_wrapper.sh
mrks: > Does apple ship it's own clang tidy or what is the reason for this change?

brew does not link `clang-tidy` by default. As such, it is not guaranteed to exist.

> Also, why execute v9 when brew is used?

Where are you seeing that?

src/benchmarklib/benchmark_runner.cpp
mrks: In the end, they both mean the same thing. `const auto` gets deduced to `const char* const`. I have no idea why clang-tidy suddenly wants to make this more explicit, but I think we can't turn it off without turning off other "this could be const" warnings too

src/lib/operators/abstract_join_operator.cpp
mrks: Because "\n" and " " are.

src/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_nodes.cpp
mrks: An iterator does not have to be class or a struct.

> Iterators are a generalization of pointers that allow a C++ program to work with different data structures (for example, containers and ranges) in a uniform manner.

As above, I am not sure what the intention of this is, but I don't see real harm being done.

src/lib/storage/vector_compression/simd_bp128/simd_bp128_packing.hpp
mrks: Yepp

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -13,7 +13,8 @@ template <typename T>
 #ifdef __clang__
 __attribute__((optnone))  // Fixes issues with memcheck. As we are only accessing constant values, shouldn't cost much
 #endif
-size_t string_heap_size(const T& string) {
+size_t
```
Bouncner, line 16: ?

src/lib/storage/vector_compression/simd_bp128/simd_bp128_packing.hpp
Bouncner: Ok, the `in` makes sense to me now. What does `uint128_t* const out` actually mean? Can the pointer we non-const but the content is supposed to be const? Doesn't make sense to me.

scripts/clang_tidy_wrapper.sh
Bouncner: Misread that.

src/lib/utils/size_estimation_utils.hpp
mrks: No idea. clang-format did it.

src/lib/storage/vector_compression/simd_bp128/simd_bp128_packing.hpp
mrks: > Can the pointer we non-const but the content is supposed to be const?

Yes. In that case, you could make the pointer point somewhere else, but you couldn't change the value that is pointed to.

However, in the code above, we have the exact opposite: The pointer cannot be changed, but the value that is pointed to can (const applies to left)

Bouncner: So the `const auto* const separator` we have all over the place is an example where the const applies to the right side (for the left const it's clear as there is not left of it, but the the right side that's "transitive"???).
But then again here in the SIMD code, it applies to the pointer.

Hmpf ...

mrks: It's a const pointer to a const thing, i.e., neither can the pointer be changed to point somewhere else nor can the "thing" be changed itself.

What do you mean with "transitive"?

Did you watch the east const video? I think that clarifies it.

src/lib/utils/size_estimation_utils.hpp
Bouncner: Format is drunk.

src/lib/storage/vector_compression/simd_bp128/simd_bp128_packing.hpp
Bouncner: I've seen it. But I am not sure I fully understood it. :)

What bothers me, is that we have two syntaxes in the same line (if I get it right): `const uint8_t* const in, uint128_t* const out` ... for in, the const is on the left side, for out it's on the right side. West coast, east coast. Or am I again mistaken?

mrks: That's why west const sucks. ;)

If we were using east const only, it would look like this:

`uint8_t const *const in, uint128_t *const out`

Looks like the [clang-format folks are working on it](https://reviews.llvm.org/D69764). It would be interesting to see how many lines would have to change.

Bouncner: Definitely something I have to get used to. If clang-format will solve that for us: fine. But until then we should make sure that at least we don't mix both styles in a single line.

PR #2124, opened 2020-05-06
Fix boost_default_memory_resource alignment

mrks:
testing, WIP

### DIFF
--- a/src/lib/memory/boost_default_memory_resource.cpp
+++ b/src/lib/memory/boost_default_memory_resource.cpp
@@ -9,7 +9,9 @@ namespace boost::container::pmr {
 
 class default_resource_impl : public memory_resource {  // NOLINT
  public:
-  void* do_allocate(std::size_t bytes, std::size_t alignment) override { return std::malloc(bytes); }  // NOLINT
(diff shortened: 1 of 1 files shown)


mrks:
Looks good. We'll see if this fixes the CI builds across all branches, but it is something that should be done in any case.

Bensk1:
Hm, malloc/free should be defined in `stdlib.h`. But `memalign` is in malloc.h which does not exist (at least not where we are looking for it). Apparently, there is `posix_memalign` in `stdlib.h` or `std::aligned_alloc` in `cstdlib` if that helps.

mrks:
Haha. In the art of computer science, subtractive methods are generally underestimated.

mrks:
While it is no longer necessary to ship the default resource ourselves, boost simply continues to ignore that parameter:

https://github.com/boostorg/container/blob/f73fb1442a0d714e97506b6fa86f5a1f16da7ddc/src/global_resource.cpp#L34

mrks:
Ok, as memcheck doesn't seem to like me and as the ignored alignment is equally used in boost, I'll close this.

PR #2126, opened 2020-05-07
Reduce cache size in JoinTestRunner

mrks:
The JoinTestRunner caches the verification tables as generated by the JoinVerification operator. The key for that cache is the test configuration. As more and more test configurations were added, that cache grew in size, leading to out-of-memory issues with TSAN.

Most of the configuration options, however, are not relevant for the JoinVerification operator. Its results will always be the same independent of, e.g, the chunk size. By leaving the options out of the cached key, we can reduce the size of the cache and fix the current CI issues.

As a welcomed side effect, the number of JoinVerification invocations is down from 2219 to 691, reducing the total runtime from 49975 ms to 41540 ms (debug on Mac).

### DIFF
--- a/src/test/operators/join_test_runner.cpp
+++ b/src/test/operators/join_test_runner.cpp
@@ -638,7 +638,26 @@ TEST_P(JoinTestRunner, TestJoin) {
   const auto join_op = configuration.join_operator_factory->create_operator(input_operator_left, input_operator_right,
                                                                             primary_predicate, configuration);
 
-  auto expected_output_table_iter = expected_output_tables.find(configuration);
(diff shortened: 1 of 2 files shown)


mrks:
Added one more thing: TPCHTableGeneratorTest.RowCountsMediumScaleFactor used a chunk size of 100, which made the generation insanely expensive. Using the default chunk size should save us tens of minutes in some CI steps.

mrks:
> fatal: unable to access 'https://github.com/hyrise/sql-parser/': Could not resolve host: github.com


Wazzup, HPI network? :(

PR #2127, opened 2020-05-07
Console: Horizontal Scrolling

### LINKED ISSUE #2106: hyriseConsole: Horizontal Scrolling
When working with wide enterprise tables, the console should not wrap lines but should allow me to scroll horizontally. Vertical scrolling is already implemented, so this should not be a big issue.

j-tr:
closes #2106.
adds horizontal scrolling in console.

### DIFF
--- a/src/bin/console/console.cpp
+++ b/src/bin/console/console.cpp
@@ -358,21 +358,33 @@ void Console::out(const std::shared_ptr<const Table>& table, const PrintFlags fl
   int size_x;
   rl_get_screen_size(&size_y, &size_x);
 
-  const bool fits_on_one_page = table->row_count() < static_cast<uint64_t>(size_y) - 1;
+  const bool fits_on_one_page_y = table->row_count() < static_cast<uint64_t>(size_y) - 1;
+
+  std::stringstream stream;
+  Print::print(table, flags, stream);
+
+  bool fits_on_one_page_x = true;
+  auto stream_backup = stream.str();
+  std::string line;
+  while (std::getline(stream, line, '\n')) {
+    if (line.length() > static_cast<uint64_t>(size_x)) {
+      fits_on_one_page_x = false;
+      break;
+    }
+  }
+  stream.str(stream_backup);
 
   static bool pagination_disabled = false;
-  if (!fits_on_one_page && !std::getenv("TERM") && !pagination_disabled) {
+  if ((!fits_on_one_page_x || !fits_on_one_page_y) && !std::getenv("TERM") && !pagination_disabled) {
     out("Your TERM environment variable is not set - most likely because you are running the console from an IDE. "
         "Pagination is disabled.\n\n");
     pagination_disabled = true;
   }
 
-  // Paginate only if table has more rows that fit in the terminal
-  if (fits_on_one_page || pagination_disabled) {
-    Print::print(table, flags, _out);
+  // Paginate only if table has more rows or printed columns that fit in the terminal
+  if ((fits_on_one_page_x && fits_on_one_page_y) || pagination_disabled) {
+    _out << stream.rdbuf();
   } else {
-    std::stringstream stream;
-    Print::print(table, flags, stream);
     _pagination_active = true;
     Pagination(stream).display();
     _pagination_active = false;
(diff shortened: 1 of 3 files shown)


j-tr:
so far, horizontal scrolling by units of one half screen with is implemented. 
how many interaction options do we want? 

- [ ] Right/Left by on character
- [ ] Right to last column
- [ ] Left to first column.

mrks:
Personally, I'd favor the cursor keys to scroll by a single character. I rather hold the key for a while than scrolling half a page and having to track the line I was in.

j-tr:
catchig Ctrl+... seems difficult with getch(). so jump to begin is now just "a" end end "e".

src/bin/console/pagination.cpp
```diff
@@ -105,6 +106,37 @@ void Pagination::display() {
...
+        break;
+      }
+
+      case 'e': {
+        const auto last_column = _lines[current_line].length() - _size_x + 1;
+        if (_lines[current_line].length() > _size_x && current_column < last_column) {
+          current_column = last_column;
+          reprint = true;
+        }
+        break;
```
mrks, line 137: Using `_lines[current_line].length` is problematic, as the output starts with `=== Columns`, which always fits on the screen. As a result, `[e]` does not work when pressed immediately after a print statement.

src/bin/console/pagination.cpp
```diff
@@ -105,6 +106,37 @@ void Pagination::display() {
...
 
+      case KEY_RIGHT: {
+        ++current_column;
+        reprint = true;
+        break;
+      }
+
+      case KEY_LEFT: {
+        if (current_column > 0) {
+          --current_column;
```
mrks, line 117: Having had the chance to test this, I'd rather make this 5 or so. What do you think?

src/bin/console/console.cpp
```diff
@@ -358,21 +358,33 @@ void Console::out(const std::shared_ptr<const Table>& table, const PrintFlags fl
   int size_x;
   rl_get_screen_size(&size_y, &size_x);
 
-  const bool fits_on_one_page = table->row_count() < static_cast<uint64_t>(size_y) - 1;
+  const bool fits_on_one_page_y = table->row_count() < static_cast<uint64_t>(size_y) - 1;
```
mrks, line 361: This does not take all the additional information into account:

```
(debug)> print test
=== Columns
|       a||        MVCC        |
|     int||_BEGIN|_END  |_TID  |
|not null||      |      |      |
=== Chunk 0 ===
|<ValueS>||
|       1||     0|      |      |
|      24||     0|      |      |
```

Feel free to move this into its own issue, as this has been broken before.

src/bin/console/pagination.cpp
```diff
@@ -148,10 +184,14 @@ void Pagination::_print_help_screen() {
   wprintw(help_screen, "  Available commands:\n\n");
   wprintw(help_screen, "  %-17s- Move down one line.\n", "ARROW DOWN, j");
   wprintw(help_screen, "  %-17s- Move up one line.\n\n", "ARROW UP, k");
+  wprintw(help_screen, "  %-17s- Move right one column.\n", "ARROW RIGHT");
```
mrks, line 187: The grouping in here is weird:

```
  ARROW DOWN, j    - Move down one line.
  ARROW UP, k      - Move up one line.

  ARROW RIGHT      - Move right one column.
  ARROW LEFT       - Move left one column.
  PAGE DOWN, SPACE - Move down one page.
  PAGE UP, b       - Move up one page.

  HOME, g, <       - Go to first line.
  END, G, >        - Go to last line.

  a                - Go to first column.
  e                - Go to last column.
  q                - Quit.
```

Suggestion:
```
  ARROW DOWN, j    - Move down one line.
  ARROW UP, k      - Move up one line.

  ARROW RIGHT      - Move right one column.
  ARROW LEFT       - Move left one column.

  PAGE DOWN, SPACE - Move down one page.
  PAGE UP, b       - Move up one page.

  HOME, g, <       - Go to first line.
  END, G, >        - Go to last line.

  a                - Go to first column.
  e                - Go to last column.
  
  q                - Quit.
```

src/bin/console/pagination.cpp
j-tr: this is how less is doing it. i guess the idea behind is that you dont end up in empty space far to the right in case you have a very long line outside the visible area.
since this can't really happen with our data, i changed it to jump to the global maximum line length.

PR #2129, opened 2020-05-08
Remove unused Cache Implementations

dey4ss:
This PR makes GDFSCache the only cache in Hyrise, and removes the Cache abstraction layer.


src/lib/cache/abstract_cache.hpp
```diff
@@ -1,14 +1,17 @@
 #pragma once
 
+#include <atomic>
 #include <utility>
 
 #include <boost/iterator/iterator_facade.hpp>
 
 namespace opossum {
 
+inline constexpr size_t DefaultCacheCapacity = 1024;
```
mrks, line 10: Shouldn't this be DEFAULT_CACHE_CAPACITY?

src/lib/cache/gdfs_cache.hpp
```diff
@@ -1,19 +1,23 @@
...
+#include "abstract_cache.hpp"
 #include "boost/heap/fibonacci_heap.hpp"
 #include "utils/assert.hpp"
 
 namespace opossum {
 
-// Generic cache implementation using the GDFS policy.
-// Note: This implementation is not thread-safe.
+/*
+ * Generic cache implementation using the GDFS policy.
```
mrks, line 15: Add a comment that previous implementations exist, but have been removed in this PR (link it)?

src/lib/cache/gdfs_cache.hpp
```diff
@@ -1,19 +1,23 @@
...
+ * Note: This implementation is only thread-safe for basic cache operations.
```
mrks, line 16: Not sure if this comments helps a lot. What is a "basic" cache operation?

src/lib/cache/abstract_cache.hpp
```diff
@@ -71,7 +85,7 @@ class AbstractCacheImpl {
   // Remove an element from the cache according to the cache algorithm's strategy
   virtual void _evict() = 0;
 
-  size_t _capacity;
+  std::atomic_size_t _capacity;
```
mrks, line 88: Now that the AbstractCache class has "unsafe" methods and atomic members, we should clearly define its concurrency guarantees.

src/lib/cache/abstract_cache.hpp
```diff
@@ -61,8 +68,15 @@ class AbstractCacheImpl {
   // Resize to the given capacity.
   virtual void resize(size_t capacity) = 0;
 
-  virtual ErasedIterator begin() = 0;
-  virtual ErasedIterator end() = 0;
+  // These methods are named "unsafe_" (similar to tbb's naming) because iterator does not hold a mutex. As such,
+  // modifications to the cache invalidate the iterators. While this is also true for begin()/end() in other data
+  // structures, the Cache class usually deals with concurrency.
```
mrks, line 73: * mutex is implementation-defined
* There is no Cache class anymore

src/lib/cache/gdfs_cache.hpp
```diff
@@ -81,10 +87,7 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
     // If the cache is full, erase the item at the top of the heap
     // so that we can insert the new item.
     if (_queue.size() >= this->_capacity) {
-      auto top = _queue.top();
-      _inflation = top.priority;
-      _map.erase(top.key);
-      _queue.pop();
+      _evict();
```
mrks, line 90: Good catch

src/lib/cache/gdfs_cache.hpp
```diff
@@ -94,9 +97,25 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
...
+
+    Handle handle = it->second;
+    GDFSCacheEntry& entry = (*handle);
+    entry.frequency++;
+    entry.priority = _inflation + static_cast<double>(entry.frequency) / entry.size;
+    _queue.update(handle);
+    return entry.value;
+  }
+
+  Value& get_entry(const Key& key) {
```
mrks, line 115: Make `try_get` return an `optional<reference_wrapper<const Value>>` and simply assert that it is set here?

Also, why is it "get_entry" but just "try_get"? Rename to "get"?

src/lib/cache/gdfs_cache.hpp
```diff
@@ -130,9 +157,9 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
...
   }
 
-  ErasedIterator begin() { return ErasedIterator{std::make_unique<Iterator>(_map.begin())}; }
+  ErasedIterator unsafe_begin() { return ErasedIterator{std::make_unique<Iterator>(_map.begin())}; }
 
-  ErasedIterator end() { return ErasedIterator{std::make_unique<Iterator>(_map.end())}; }
+  ErasedIterator unsafe_end() { return ErasedIterator{std::make_unique<Iterator>(_map.end())}; }
 
   size_t frequency(const Key& key) {
     const auto it = _map.find(key);
```
mrks, line 165: Why is this (as well as priority and potentially others) lock-free?

src/lib/cache/gdfs_cache.hpp
```diff
@@ -30,10 +34,11 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
 
   using Handle = typename boost::heap::fibonacci_heap<GDFSCacheEntry>::handle_type;
   using CacheMap = typename std::unordered_map<Key, Handle>;
+  using CacheSnapshot = typename std::unordered_map<Key, GDFSCacheEntry>;
```
mrks, line 37: Why is this defined here, not in AbstractCache?

src/lib/cache/gdfs_cache.hpp
```diff
@@ -145,15 +172,27 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
...
   boost::heap::fibonacci_heap<GDFSCacheEntry> _queue;
 
   // Map to point towards element in the list.
   CacheMap _map;
 
+  mutable std::shared_mutex _mutex;
+
   // Inflation value that will be updated whenever an item is evicted.
-  double _inflation;
+  std::atomic<double> _inflation;
```
mrks, line 195: We should document how we achieve thread-safety. People might be surprised that we mix locking and atomics. Also, a short comment about why we decided for locked unordered_maps instead of lock-free data structures would be helpful.

src/lib/sql/sql_plan_cache.hpp
```diff
@@ -3,14 +3,14 @@
...
 
 namespace opossum {
 
 class AbstractOperator;
 class AbstractLQPNode;
 
-using SQLPhysicalPlanCache = Cache<std::shared_ptr<AbstractOperator>, std::string>;
-using SQLLogicalPlanCache = Cache<std::shared_ptr<AbstractLQPNode>, std::string>;
+using SQLPhysicalPlanCache = GDFSCache<std::string, std::shared_ptr<AbstractOperator>>;
+using SQLLogicalPlanCache = GDFSCache<std::string, std::shared_ptr<AbstractLQPNode>>;
```
mrks, line 14: Having to explicitly talk about implementation details (GDFS), especially in the absence of alternatives, doesn't feel right here.

src/lib/cache/gdfs_cache.hpp
```diff
@@ -130,9 +157,9 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
     return (*it->second).priority;
   }
 
-  ErasedIterator begin() { return ErasedIterator{std::make_unique<Iterator>(_map.begin())}; }
+  ErasedIterator unsafe_begin() { return ErasedIterator{std::make_unique<Iterator>(_map.begin())}; }
```
mrks, line 160: Now that we have `snapshot()`, do you still need this? @Bensk1 @Bouncner

Bouncner: Too long ago this discussion. But the purpose of `snapshot()` was to avoid the potential problems with the iterators, right?

If so (and I don't oversee something), I am happy with using `snapshopt()` and removing the iterators.

src/lib/cache/gdfs_cache.hpp
Bouncner: Probably related: can we put `Key` and `Value` into the snapshot? Or is there a particular reason to exclude them? (receiving frequency might be one, now that I think about it while writing ...)

I would prefer a solution to use the snapshot that does not require knowledge about the GDFS internals.

mrks: Ah, I didn't even see that the value was GDFSCacheEntry. I guess this only depends on whether you are interested in the frequency values or not. Apparently, you just want to know *what* is in the cache, not *why*.

src/lib/cache/gdfs_cache.hpp
Bensk1: Did we only use these unsafe iterators for our plugins? In that case, that would be fine for me.

src/lib/cache/gdfs_cache.hpp
dey4ss: Okay, I did not know if that information was useful. So I'll change it to an `unordered_map<Key, Value>`.

Bouncner: But that leaves out the frequency.

src/lib/cache/gdfs_cache.hpp
dey4ss: Should this rather be here or in `AbstractCache`?

src/lib/cache/gdfs_cache.hpp
dey4ss: I guess the `optional` is for `try_get`, right?

I just used the names of the old cache class. `get` sounds better.

src/lib/cache/gdfs_cache.hpp
dey4ss: I just copied the behavior of the old implementation, which guaranteed only the functions defined by the cache abstractions to be thread-safe. I can also make everything safe.

src/lib/cache/gdfs_cache.hpp
dey4ss: I kind of answered this below, this means (try-)get, set, has, and snapshot. If I add locks to the other, this will be obsolete.

src/lib/cache/gdfs_cache.hpp
mrks: Doesn't really matter, I guess.

src/lib/cache/gdfs_cache.hpp
mrks: Mmmh... good point. Forget about the reference_wrapper part above then.

src/lib/cache/gdfs_cache.hpp
mrks: Sounds better

src/lib/cache/gdfs_cache.hpp
dey4ss: Right :D
So I guess something like 
```c++
struct SnapshotEntry {
    Value value;
    size_t frequency;
};
...
std::unordered_map<Key, SnapshotEntry> snapshot() {...}
```
would be fine?

mrks: > But then it again is the question of whether to define it in the abstract class or not, as we might assume any future implementations to deal with the frequency.

Correct. That shouldn't go into AbstractCache.

@Bouncner and @Bensk1 should decide - they are the ones using this.

mrks: > I would prefer a solution to use the snapshot that does not require knowledge about the GDFS internals.

Either you want the frequency or you don't... ;)

src/lib/cache/gdfs_cache.hpp
dey4ss: Okay, should I add a comment to the get method or make the Value& const?

mrks: We cannot return a reference, even if it is const. Who guarantees that the referenced value will still be in the cache after we return?

That should be documented.

src/lib/cache/gdfs_cache.hpp
dey4ss: Basically, this could also be a normal double and we lock the inflation() function. If we lock all the other functions as well, this should be fine.

src/lib/sql/sql_plan_cache.hpp
dey4ss: Okay, so what would you suggest?

src/lib/cache/gdfs_cache.hpp
mrks: It looks like most of these getters are just used for tests, anyway. Should we simply friend the test class and get rid of the accessors altogether?

src/lib/sql/sql_plan_cache.hpp
mrks: No idea. Still wanted to point it out. :D

As it's the only place where this happens, I could live with it.

src/lib/cache/gdfs_cache.hpp
dey4ss: I guess this would make the interface of the cache much cleaner. What about priority and frequency? Should they remain at all or public?

mrks: A question for the people that look at the cache, again

dey4ss:
`StressTest.TestTransactionConflicts` and `ServerTestRunner.TestTransactionConflicts` seem to fail quite often (or every time) with TSAN due to timeouts (also on other branches). Did anyone else had the problem? I saw that in both tests we [timeout after a while there if threads did not finish](https://github.com/hyrise/hyrise/blob/master/src/test/server/server_test_runner.cpp#L445). Do we really have a problem here if we just join the threads?

Another thing I recognized was a timeout in `hyriseConsoleTest.py` line 71, like on master currently. Does anybody know why?

mrks:
I wonder if this depends on whether the build is executed on imresx01 or rapa (which is a bit slower). While it's running, Jenkins shows this in the left column if you use the traditional interface. Not sure if you can check that information for a failed build.

dey4ss:
> 
> 
> I wonder if this depends on whether the build is executed on imresx01 or rapa (which is a bit slower). While it's running, Jenkins shows this in the left column if you use the traditional interface. Not sure if you can check that information for a failed build.

[Looks like rapa](https://hyrise-ci.epic-hpi.de/job/hyrise/job/hyrise/job/PR-2129/12/consoleFull). Anyways, do you think simply joining the treads could be a solution? Or setting the timeouts even higher?

mrks:
Feel free to increase the timeouts. For now, I'd try to leave them in so that the tests actually fail, not silently wait forever.

dey4ss:
For now, I'll try it with another 30 s

src/lib/cache/gdfs_cache.hpp
Bouncner: I don't mind removing some getters. For me as a cache user, I am only interested in the frequency in the snapshot. Nowhere else.
And I don't know what the priority really does, so I can't say anything about that.

src/lib/cache/gdfs_cache.hpp
Bouncner: I want the frequency, but I don't care if it's a GDFS or not. A plan cache without a frequency is of really small value to me, but I really don't care if there are changes to the caching at some point in time. When a new caching algorithm does not provide a frequency out of the box, I think it needs to be added or this caching algorithms is simply not a good fit for a database system.

Actually, I'd love even more like HANAs "time of adding entry to cache", but that's another story.

It's hard for me to follow this whole thread here. So sorry if that has been already discussed.

+1 for the SnapshotEntry.

mrks: The SnapshotEntry doesn't make any sense. It pretends to be a common interface for this and other (non-existent) implementations, but those do not necessarily have a frequency field (e.g., LRU).

If you want the snapshot to contain the frequency, I consider the current approach to be correct.

Bouncner: I think it makes sense as it states what we expect of the underlying cache. If the LRU cannot provide it: LRU caches don't fit.
Or do I oversee components in Hyrise that also use a cache and where the frequency does not make any sense?

From a cache user perspective I think having manually get the frequency from some objects types that I don't understand is cumbersome.

mrks: For the sake of convenience, I could live with an `optional<..> frequency`. Still don't think that it's architecturally correct.

src/lib/cache/gdfs_cache.hpp
```diff
@@ -1,19 +1,22 @@
...
 #include "boost/heap/fibonacci_heap.hpp"
 #include "utils/assert.hpp"
 
 namespace opossum {
 
-// Generic cache implementation using the GDFS policy.
-// Note: This implementation is not thread-safe.
+/*
+ * Generic cache implementation using the GDFS policy.
+ * To iterate over a cache copy, use snapshot().
```
mrks, line 15: Could clarify why I want this, e.g., "To iterate over the cache in a thread-safe manner, ..."

src/lib/cache/abstract_cache.hpp
```diff
@@ -1,52 +1,36 @@
...
 
   // Cache the value at the given key.
   // If the new size exceeds the capacity an item will be evicted.
   // Depending on the underlying strategy, the parameters for cost and size may be used.
   // If they are not intended to be used, we specify constant default values here.
   virtual void set(const Key& key, const Value& value, double cost = 1.0, double size = 1.0) = 0;
 
+  // Tries to fetch the cache entry for the query into the result object. Returns true if the entry was found, false
+  // otherwise.
+  virtual std::optional<Value> try_get(const Key& key) = 0;
```
mrks, line 29: > for the query into the result object

Not sure if I understand this correctly, but it sounds like you are simply describing a return value.

> Returns true if the entry was found

No, it returns an optional, which is [contextually convertible](https://en.cppreference.com/w/cpp/language/implicit_conversion) to bool.

"Tries to fetch a cache entry" is probably sufficient.

src/lib/cache/abstract_cache.hpp
```diff
@@ -1,52 +1,36 @@
...
+
   // Get the cached value at the given key.
   // Causes undefined behavior if the item is not in the cache.
+  // This cannot be thread-safe since we return a reference here.
```
mrks, line 33: Is this actually used somewhere outside of the tests or can we get rid of it? In that case, I would add a comment to try_get, explaining why it does not return a reference.

src/lib/cache/gdfs_cache.hpp
```diff
@@ -105,53 +93,60 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
     return entry.value;
   }
 
-  bool has(const Key& key) const { return _map.find(key) != _map.end(); }
+  bool has(const Key& key) const {
+    std::shared_lock<std::shared_mutex> lock(_mutex);
+    return _map.find(key) != _map.end();
```
mrks, line 98: contains?

src/lib/cache/gdfs_cache.hpp
```diff
@@ -105,53 +93,60 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
...
-  ErasedIterator end() { return ErasedIterator{std::make_unique<Iterator>(_map.end())}; }
-
-  size_t frequency(const Key& key) {
+  size_t frequency(const Key& key) const {
+    std::shared_lock<std::shared_mutex> lock(_mutex);
     const auto it = _map.find(key);
     if (it == _map.end()) {
       return size_t{0};
     }
+    return (*it->second).frequency;
```
mrks, line 126: `it->second->frequency`?

src/lib/cache/gdfs_cache.hpp
```diff
@@ -105,53 +93,60 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
...
+  }
 
-    Handle handle = it->second;
-    GDFSCacheEntry& entry = (*handle);
-    return entry.frequency;
+  std::unordered_map<Key, SnapshotEntry> snapshot() const {
+    std::shared_lock<std::shared_mutex> lock(_mutex);
+    std::unordered_map<Key, SnapshotEntry> map_copy(_map.size());
+    for (const auto& [key, entry] : _map) {
+      map_copy[key] = SnapshotEntry{(*entry).value, (*entry).frequency};
```
mrks, line 133: `entry->...`?

src/lib/cache/gdfs_cache.hpp
```diff
@@ -105,53 +93,60 @@ class GDFSCache : public AbstractCacheImpl<Key, Value> {
...
   }
 
  protected:
+  friend class CachePolicyTest;
+
   // Priority queue to hold all elements. Implemented as max-heap.
   boost::heap::fibonacci_heap<GDFSCacheEntry> _queue;
 
   // Map to point towards element in the list.
+  // We use a locked map here since TBB's concurrent maps are not safe for erasing.
```
mrks, line 145: Isn't this also an issue of the heap and the map having to be synchronized?

Suggestion:
* Make this a comment for the _mutex, not for _map. 
* "Locking this data structure is easier than using a concurrent data structure (e.g. TBB) as (1) both the queue and the map would have to be concurrent data structures, (2) their modifications would have to be synchronized, and (3) TBB's concurrent_unordered_map does not provide for deletion and concurrent_hash_map is difficult to use."

src/lib/cache/abstract_cache.hpp
dey4ss: The curse of copy-paste 😅

src/lib/cache/abstract_cache.hpp
dey4ss: After taking a quick look, it is only used in tests. I'll remove `get()`.

src/lib/cache/gdfs_cache.hpp
dey4ss: It looks like the `fibonacci_heap<>::handle_type` does not support `->` rather than being resolved.

dey4ss:
Was it okay to increase the timeout of `hyriseConsoleTest.py` by 15 s?

mrks:
> Was it okay to increase the timeout of hyriseConsole_test.py by 15 s?

Yepp. The exact time for the timeouts doesn't really matter. It's just that the tests should deterministically fail at *some* point.

dey4ss:
Alright.
@Bensk1, @Bouncner, do you have further remarks?

mrks:
(talked to them)

PR #2133, opened 2020-05-11
Fix Export Column Pruning

### LINKED ISSUE #2128: CSV Export does not work
CSV exports only write the last column.

Steps to reproduce:

1. Use a file like
```
a|b
string|string
foo|bar
bar|foo
```
2. Import the file, e.g. `COPY foo FROM './link/to/file.tbl'` (the imported table looks correct).
3. Export the table as CSV, e.g.  `COPY foo TO './link/to/file.csv'`.
4. The CSV file loos like this:
```
"bar"
"foo"
```

dey4ss:
closes #2128


mrks:
Can we modify one of the existing tests to cover this?

mrks:
Interesting find though. How did you end up in that rule?

dey4ss:
> 
> 
> Can we modify one of the existing tests to cover this?

I'm adding tests for the rule at the moment.

dey4ss:
> 
> 
> Interesting find though. How did you end up in that rule?

I noticed we only end up with one column in the export table, so I cout'ed the Table::column_count() whenever it was called and noticed it changed during the execution. Therefore, columns must have been pruned.

mrks:
I was more thinking about an end-to-end CSV test, which seems to be missing.

mrks:
Whoopsie. :(

dey4ss:
> 
> 
> I was more thinking about an end-to-end CSV test, which seems to be missing or using only a single column.

Ah, okay. Would it be enough to change [this test](https://github.com/hyrise/hyrise/blob/master/src/test/server/server_test_runner.cpp#L164) to use a CSV file, too? Or would it also be okay to just compare the tables, as the export itself is affected by the pruning rule?

mrks:
Sounds good.

PR #2135, opened 2020-05-14
Change ValueExpression long suffix to upper case

### LINKED ISSUE #2134: TableScan::create_impl broken for long values on int column
Screenshot 2020-05-13 at 18 24 18

The ExpressionEvaluator being used is expected, as the regular scan expects matching data types. However, I have no idea why 10000000000 becomes 10000000001. That doesn't seem right.

mrks:
Turns out #2134 was a false alarm. The 'l' indicating a long value just looks awfully similar to a '1'. Let's use upper case suffixes instead.

fixes #2134


Bouncner:
A gitmoji would help even more.

PR #2136, opened 2020-05-15
StoredTableColumnAlignmentRule

mweisgut:
## Description
The PQP sub plan memoization for StoredTableNodes is sensitive to the node's table name, set of pruned chunks and set of pruned columns. Consequently, for multiple nodes with the same table name, same pruned chunks and same pruned columns, only one GetTable operator is created and executed. In some queries, the ColumnPruningRule and ChunkPruningRule provide an LQP with multiple StoredTableNodes where the table names and sets of pruned chunks are equal but the sets of pruned columns are different. This leads to the creation and execution of different GetTable operators instead of only one operator.

This rule identifies StoredTableNodes that differ only in the pruned columns. It then intersects the lists of pruned columns. While this means that some columns are left unpruned, it makes the job easier for the memoization in the LQPTranslator. In our experiments, this has led to significant performance improvements and negligible reductions.

## Benchmark Comparison
comparison: d43fb623d vs f7d6246ec
executed on pella, NUMA node 1
build: RelWithDebInfo

### TPC-H
`./hyriseBenchmarkTPCH -s 1 -r 30`
```
| TPC-H 01       | 0.808261215687 | 30   | 0.775613844395 | 30   | -4%        | (run time too short) |
| TPC-H 02       | 39.4202270508  | 30   | 40.5404472351  | 30   | +3%        | (run time too short) |
| TPC-H 03       | 5.11661243439  | 30   | 5.16426706314  | 30   | +1%        | (run time too short) |
| TPC-H 04       | 6.22300100327  | 30   | 6.11033773422  | 30   | -2%        | (run time too short) |
| TPC-H 05       | 1.98856902122  | 30   | 1.98523318768  | 30   | -0%        | (run time too short) |
| TPC-H 06       | 92.9693603516  | 30   | 93.6799240112  | 30   | +1%        | (run time too short) |
| TPC-H 07       | 7.18500375748  | 30   | 7.23907995224  | 30   | +1%        | (run time too short) |
| TPC-H 08       | 7.44914722443  | 30   | 7.46293115616  | 30   | +0%        | (run time too short) |
| TPC-H 09       | 1.05010020733  | 30   | 1.04464066029  | 30   | -1%        | (run time too short) |
| TPC-H 10       | 1.95173037052  | 30   | 1.94143795967  | 30   | -1%        | (run time too short) |
| TPC-H 11       | 51.842124939   | 30   | 51.6629066467  | 30   | -0%        | (run time too short) |
| TPC-H 12       | 10.4508676529  | 30   | 9.9458770752   | 30   | -5%        | (run time too short) |
| TPC-H 13       | 1.52068531513  | 30   | 1.51268649101  | 30   | -1%        | (run time too short) |
| TPC-H 14       | 25.84806633    | 30   | 25.5330581665  | 30   | -1%        | (run time too short) |
| TPC-H 15       | 56.921875      | 30   | 57.5614852905  | 30   | +1%        | (run time too short) |
| TPC-H 16       | 5.95111179352  | 30   | 6.00333356857  | 30   | +1%        | (run time too short) |
| TPC-H 17       | 9.75971412659  | 30   | 16.4921894073  | 30   | +69%       | (run time too short) |
| TPC-H 18       | 0.735659241676 | 30   | 0.73869240284  | 30   | +0%        | (run time too short) |
| TPC-H 19       | 12.1287403107  | 30   | 12.2900209427  | 30   | +1%        | (run time too short) |
| TPC-H 20       | 28.2435932159  | 30   | 28.4328689575  | 30   | +1%        | (run time too short) |
| TPC-H 21       | 0.821533858776 | 30   | 1.16572213173  | 30   | +42%       | (run time too short) |
| TPC-H 22       | 7.78854417801  | 30   | 7.86301755905  | 30   | +1%        | (run time too short) |
| geometric mean |                |      |                |      | +4%        |                      |
```

### TPC-DS
`./hyriseBenchmarkTPCDS -s 1 -r 30`
```
+----------------+---------------+------+---------------+------+------------+----------------------+
| Benchmark      | prev. iter/s  | runs | new iter/s    | runs | change [%] |              p-value |
+----------------+---------------+------+---------------+------+------------+----------------------+
| 01             | 14.8473272324 | 30   | 15.1353216171 | 30   | +2%        | (run time too short) |
| 03             | 57.7059631348 | 30   | 57.7753639221 | 30   | +0%        | (run time too short) |
| 06             | 20.8637065887 | 30   | 20.9599342346 | 30   | +0%        | (run time too short) |
| 07             | 6.79859972    | 30   | 6.51345396042 | 30   | -4%        | (run time too short) |
| 09             | 4.04866456985 | 30   | 4.03576564789 | 30   | -0%        | (run time too short) |
| 10             | 20.539100647  | 30   | 20.6436004639 | 30   | +1%        | (run time too short) |
| 13             | 5.34782171249 | 30   | 5.38502931595 | 30   | +1%        | (run time too short) |
| 15             | 41.746547699  | 30   | 41.9188842773 | 30   | +0%        | (run time too short) |
| 17             | 12.4100160599 | 30   | 12.5039396286 | 30   | +1%        | (run time too short) |
| 19             | 40.1315536499 | 30   | 40.2961769104 | 30   | +0%        | (run time too short) |
| 25             | 18.6570549011 | 30   | 18.9375076294 | 30   | +2%        | (run time too short) |
| 26             | 15.309504509  | 30   | 15.1728496552 | 30   | -1%        | (run time too short) |
| 28             | 7.99202489853 | 30   | 7.92670726776 | 30   | -1%        | (run time too short) |
| 29             | 12.2146492004 | 30   | 12.3524341583 | 30   | +1%        | (run time too short) |
| 31             | 3.84116411209 | 30   | 3.83518838882 | 30   | -0%        | (run time too short) |
| 34             | 17.777009964  | 30   | 17.8109416962 | 30   | +0%        | (run time too short) |
| 35             | 4.74193954468 | 30   | 4.78170204163 | 30   | +1%        | (run time too short) |
| 39a            | 2.4887316227  | 30   | 2.48633623123 | 30   | -0%        | (run time too short) |
| 39b            | 2.48768496513 | 30   | 2.49046421051 | 30   | +0%        | (run time too short) |
| 41             | 17.5549736023 | 30   | 16.9735965729 | 30   | -3%        | (run time too short) |
| 42             | 61.6618843079 | 30   | 61.7899055481 | 30   | +0%        | (run time too short) |
| 43             | 2.19654202461 | 30   | 2.19479179382 | 30   | -0%        | (run time too short) |
| 45             | 43.8250770569 | 30   | 43.9056663513 | 30   | +0%        | (run time too short) |
| 48             | 5.65509653091 | 30   | 5.6732635498  | 30   | +0%        | (run time too short) |
| 50             | 36.4451217651 | 30   | 36.749332428  | 30   | +1%        | (run time too short) |
| 52             | 60.7162437439 | 30   | 60.817199707  | 30   | +0%        | (run time too short) |
| 55             | 62.1214942932 | 30   | 62.0861740112 | 30   | -0%        | (run time too short) |
| 62             | 6.91570949554 | 30   | 6.84029006958 | 30   | -1%        | (run time too short) |
| 65             | 7.27852344513 | 30   | 7.18192815781 | 30   | -1%        | (run time too short) |
| 69             | 16.3616542816 | 30   | 16.3075428009 | 30   | -0%        | (run time too short) |
| 73             | 26.6281795502 | 30   | 26.4427642822 | 30   | -1%        | (run time too short) |
| 79             | 9.3133058548  | 30   | 9.25420951843 | 30   | -1%        | (run time too short) |
| 81             | 28.1001358032 | 30   | 28.116689682  | 30   | +0%        | (run time too short) |
| 83             | 51.9071273804 | 30   | 52.4082145691 | 30   | +1%        | (run time too short) |
| 85             | 6.56966876984 | 30   | 6.73189687729 | 30   | +2%        | (run time too short) |
| 88             | 4.78254747391 | 30   | 4.78843021393 | 30   | +0%        | (run time too short) |
| 91             | 30.44049263   | 30   | 30.6110363007 | 30   | +1%        | (run time too short) |
| 93             | 2.13753294945 | 30   | 2.11107897758 | 30   | -1%        | (run time too short) |
| 96             | 49.883934021  | 30   | 49.6257514954 | 30   | -1%        | (run time too short) |
| 97             | 1.30872070789 | 30   | 1.31454968452 | 30   | +0%        | (run time too short) |
| 99             | 3.34695076942 | 30   | 3.33235883713 | 30   | -0%        | (run time too short) |
| geometric mean |               |      |               |      | -0%        |                      |
+----------------+---------------+------+---------------+------+------------+----------------------+
```

### Testing
- [x] `./hyriseBenchmarkTPCH -s 0.01 -r 1 --verify` (f7d6246ec)
- [x] `./hyriseBenchmarkTPCDS -s 1 -r 1 --verify` (460d187)
- [x] `hyriseTest` (f7d6246ec)

### Query Plan Example: TPC-H Q17

Old LQP
TPC-H_17-LQP_old

Old PQP
TPC-H_17-PQP_old

New LQP
TPC-H_17-LQP_new

New PQP
TPC-H_17-PQP_new

### DIFF
--- a/src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
+++ b/src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
@@ -8,81 +8,57 @@
 namespace opossum {
 
 void StoredTableColumnAlignmentRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root) const {
-  std::cout << "APPLY TO\n";
   StoredTableNodeUnorderedMap<std::vector<std::shared_ptr<StoredTableNode>>> gathered_stored_table_nodes_for_node;
   StoredTableNodeUnorderedMap<std::vector<ColumnID>> aligned_pruned_column_ids_for_node;
-  recursively_gather_stored_table_nodes(root, gathered_stored_table_nodes_for_node,
-  	aligned_pruned_column_ids_for_node);
-  
-  for (const auto& kv : gathered_stored_table_nodes_for_node) {
-  	std::cout << "StoredTableNode: " << kv.first->table_name << "\n";
-  	for (const auto& chunk_id : kv.first->pruned_chunk_ids()) {
-  		std::cout << "  " << chunk_id << "\n";
-  	}
-  	for (const auto& stored_table_node : kv.second) {
-  		std::cout << "pruned columns: " << "\n";
-  		for (const auto& column_id : stored_table_node->pruned_column_ids()) {
-  			std::cout << column_id << " ";
-  		}
-  		std::cout << "\n";
-  	}
-  }
-
-  // iterate over groups of StoredTableNode and set the minimum superset of columns for each StoredTableNode in each
-  // group
+  recursively_gather_stored_table_nodes(root, gathered_stored_table_nodes_for_node, aligned_pruned_column_ids_for_node);
+  // iterate over groups of StoredTableNode and set the maximum superset of columns for each StoredTableNode in each
+  // StoredTableNode group. A group is characterized by the table name and the pruned chunk ids.
   for (const auto& node_nodes_pair : gathered_stored_table_nodes_for_node) {
- 	for (const auto& stored_table_node : node_nodes_pair.second) {
-  		stored_table_node->set_pruned_column_ids(aligned_pruned_column_ids_for_node[stored_table_node]);
-  	}
-  }
-
-  std::cout << "after update" << "\n";
-  for (const auto& kv : gathered_stored_table_nodes_for_node) {
-  	std::cout << "StoredTableNode: " << kv.first->table_name << "\n";
-  	for (const auto& chunk_id : kv.first->pruned_chunk_ids()) {
-  		std::cout << "  " << chunk_id << "\n";
-  	}
-  	for (const auto& stored_table_node : kv.second) {
-  		std::cout << "pruned columns: " << "\n";
-  		for (const auto& column_id : stored_table_node->pruned_column_ids()) {
-  			std::cout << column_id << " ";
-  		}
-  		std::cout << "\n";
-  	}
+    for (const auto& stored_table_node : node_nodes_pair.second) {
+      stored_table_node->set_pruned_column_ids(aligned_pruned_column_ids_for_node[stored_table_node]);
+    }
   }
 }
 
-void StoredTableColumnAlignmentRule::recursively_gather_stored_table_nodes(const std::shared_ptr<AbstractLQPNode>& node,
-  StoredTableNodeUnorderedMap<std::vector<std::shared_ptr<StoredTableNode>>>& gathered_stored_table_nodes_for_node,
-  StoredTableNodeUnorderedMap<std::vector<ColumnID>>& aligned_pruned_column_ids_for_node) const {
+void StoredTableColumnAlignmentRule::recursively_gather_stored_table_nodes(
+    const std::shared_ptr<AbstractLQPNode>& node,
+    StoredTableNodeUnorderedMap<std::vector<std::shared_ptr<StoredTableNode>>>& gathered_stored_table_nodes_for_node,
+    StoredTableNodeUnorderedMap<std::vector<ColumnID>>& aligned_pruned_column_ids_for_node) const {
   const auto& left_input = node->left_input();
   const auto& right_input = node->right_input();
 
   if (left_input) {
-  	recursively_gather_stored_table_nodes(left_input, gathered_stored_table_nodes_for_node,
-  		aligned_pruned_column_ids_for_node);
+    recursively_gather_stored_table_nodes(left_input, gathered_stored_table_nodes_for_node,
+                                          aligned_pruned_column_ids_for_node);
   }
   if (right_input) {
     recursively_gather_stored_table_nodes(right_input, gathered_stored_table_nodes_for_node,
-    	aligned_pruned_column_ids_for_node);	
+                                          aligned_pruned_column_ids_for_node);
   }
   if (node->type == LQPNodeType::StoredTable) {
-  	const auto stored_table_node = std::dynamic_pointer_cast<StoredTableNode>(node);
-  	
-  	auto& gathered_stored_table_nodes = gathered_stored_table_nodes_for_node[stored_table_node];
-  	gathered_stored_table_nodes.emplace_back(stored_table_node);
+    const auto stored_table_node = std::dynamic_pointer_cast<StoredTableNode>(node);
 
-  	// update minimum superset of pruned column ids for a StoredTableNodes with the same table name and same set of
-  	// pruned chunk ids
-  	auto& aligned_pruned_column_ids = aligned_pruned_column_ids_for_node[stored_table_node];
-  	std::vector<ColumnID> updated_pruned_column_ids{};
-  	updated_pruned_column_ids.reserve(aligned_pruned_column_ids.size() +
-  		stored_table_node->pruned_column_ids().size());
-  	std::set_union(aligned_pruned_column_ids.begin(), aligned_pruned_column_ids.end(),
-  		stored_table_node->pruned_column_ids().begin(), stored_table_node->pruned_column_ids().end(),
-  		std::back_inserter(updated_pruned_column_ids));
-  	updated_pruned_column_ids.shrink_to_fit();
-  	aligned_pruned_column_ids_for_node[stored_table_node] = updated_pruned_column_ids;
+    auto& gathered_stored_table_nodes = gathered_stored_table_nodes_for_node[stored_table_node];
+    gathered_stored_table_nodes.emplace_back(stored_table_node);
+
+    // update minimum superset of pruned column ids for a StoredTableNodes with the same table name and same set of
+    // pruned chunk ids
+    if (aligned_pruned_column_ids_for_node.contains(stored_table_node)) {
+      auto& aligned_pruned_column_ids = aligned_pruned_column_ids_for_node[stored_table_node];
+      if (aligned_pruned_column_ids.empty()) {
+      }
+      std::vector<ColumnID> updated_pruned_column_ids{};
+      updated_pruned_column_ids.reserve(aligned_pruned_column_ids.size() +
+                                        stored_table_node->pruned_column_ids().size());
+      std::set_intersection(aligned_pruned_column_ids.begin(), aligned_pruned_column_ids.end(),
+                            stored_table_node->pruned_column_ids().begin(),
+                            stored_table_node->pruned_column_ids().end(),
+                            std::back_inserter(updated_pruned_column_ids));
+      updated_pruned_column_ids.shrink_to_fit();
+      aligned_pruned_column_ids_for_node[stored_table_node] = updated_pruned_column_ids;
+    } else {
+      aligned_pruned_column_ids_for_node[stored_table_node] = stored_table_node->pruned_column_ids();
+    }
   }
 }
 
--- a/src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
+++ b/src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
@@ -4,58 +4,77 @@
 #include <memory>
 
 #include "logical_query_plan/abstract_lqp_node.hpp"
+#include "logical_query_plan/lqp_utils.hpp"
+
+namespace {
+// Modified hash code generation for StoredTableNodes where column pruning information is omitted. Struct is used to
+// enable hash-based containers containing std::shared_ptr<StoredTableNode>.
+struct StoredTableNodeSharedPtrHash final {
+  size_t operator()(const std::shared_ptr<opossum::StoredTableNode>& node) const {
+    size_t hash{0};
+    boost::hash_combine(hash, node->table_name);
+    for (const auto& pruned_chunk_id : node->pruned_chunk_ids()) {
+      boost::hash_combine(hash, static_cast<size_t>(pruned_chunk_id));
+    }
+    return hash;
+  }
+};
+
+// Modified equals evaluation code for StoredTableNodes where column pruning information is omitted. Struct is used to
+// enable hash-based containers containing std::shared_ptr<StoredTableNode>.
+struct StoredTableNodeSharedPtrEqual final {
+  size_t operator()(const std::shared_ptr<opossum::StoredTableNode>& lhs,
+                    const std::shared_ptr<opossum::StoredTableNode>& rhs) const {
+    return lhs == rhs || (lhs->table_name == rhs->table_name && lhs->pruned_chunk_ids() == rhs->pruned_chunk_ids());
+  }
+};
+
+template <typename Value>
+using ColumnPruningAgnosticStoredTableNodeMap =
+    std::unordered_map<std::shared_ptr<opossum::StoredTableNode>, Value, StoredTableNodeSharedPtrHash,
+                       StoredTableNodeSharedPtrEqual>;
+}  // namespace
 
 namespace opossum {
 
 void StoredTableColumnAlignmentRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root) const {
-  StoredTableNodeUnorderedMap<std::vector<std::shared_ptr<StoredTableNode>>> gathered_stored_table_nodes_for_node;
-  StoredTableNodeUnorderedMap<std::vector<ColumnID>> aligned_pruned_column_ids_for_node;
-  recursively_gather_stored_table_nodes(root, gathered_stored_table_nodes_for_node, aligned_pruned_column_ids_for_node);
+  ColumnPruningAgnosticStoredTableNodeMap<std::vector<std::shared_ptr<StoredTableNode>>>
+      gathered_stored_table_nodes_for_node;
+  ColumnPruningAgnosticStoredTableNodeMap<std::vector<ColumnID>> aligned_pruned_column_ids_for_node;
+
+  visit_lqp(root, [&](const auto& node) {
+    if (node->type == LQPNodeType::StoredTable) {
+      const auto stored_table_node = std::dynamic_pointer_cast<StoredTableNode>(node);
+
+      auto& gathered_stored_table_nodes = gathered_stored_table_nodes_for_node[stored_table_node];
+      gathered_stored_table_nodes.emplace_back(stored_table_node);
+
+      // update maximum superset of pruned column ids for a StoredTableNodes with the same table name and same set of
+      // pruned chunk ids
+      if (aligned_pruned_column_ids_for_node.contains(stored_table_node)) {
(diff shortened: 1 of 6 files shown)


mrks:
Mmmh, I guess I was too fixated on the perfect solution within the LQP cosmos :D

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <vector>
+
+#include "abstract_rule.hpp"
+#include "logical_query_plan/stored_table_node.hpp"
+
+namespace opossum {
+
+// Modified hash code generation for StoredTableNodes where column pruning information is omitted. Struct is used to
+// enable hash based containers containing std::shared_ptr<StoredTableNode>.
+struct StoredTableNodeSharedPtrHash final {
```
mrks, line 13: These strucs, as well as StoredTableNodeUnorderedMap, are not used outside the rule, are they? Move into an anonymous namespace within the cpp?

Also, `StoredTableNodeUnorderedMap` sounds a bit incorrect. We have specific equality semantics here and the name should cover that. ColumnPruningAgnosticSomething?

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
```diff
@@ -0,0 +1,54 @@
...
+    boost::hash_combine(hash, node->table_name);
+    for (const auto& pruned_chunk_id : node->pruned_chunk_ids()) {
+      boost::hash_combine(hash, static_cast<size_t>(pruned_chunk_id));
+    }
+    return hash;
+  }
+};
+
+// Modified equals evaluation code for StoredTableNodes where column pruning information is omitted. Struct is used to
+// enable hash based containers containing std::shared_ptr<StoredTableNode>.
```
mrks, line 25: hash-based

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+template <typename Value>
+using StoredTableNodeUnorderedMap = std::unordered_map<std::shared_ptr<StoredTableNode>, Value,
+                                                       StoredTableNodeSharedPtrHash, StoredTableNodeSharedPtrEqual>;
+
+// The PQP sub plan memoization for StoredTableNodes is sensitive to the node's table name, set of pruned chunks and
+// set of pruned columns. Consequently, for multiple nodes with the same table name, chunk and column pruning, only one
+// GetTable operator is created and executed. In some queries, the ColumnPruningRule and ChunkPruningRule provide an
+// LQP with multiple StoredTableNodes where the table names and sets of pruned chunks are equal but the sets of pruned
+// columns are different. This leads to the creation and execution of different GetTable operators.
```
mrks, line 40: I'm not sure if the comment is helpful for someone who has not seen the "broken" LQP. I wonder if an example would be helpful.

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
```diff
@@ -0,0 +1,54 @@
...
+//
+// This rule identifies StoredTableNodes with equal table names and sets of pruned chunks and aligns the pruned columns
+// so that the sets of pruned columns are equal which means that the PQP sub plan memoization is effective for theses
+// StoredTableNodes and only one GetTable operator is created.
```
mrks, line 44: Suggestion, but i am not yet sure about it: This rule identifies StoredTableNodes that differ only in the pruned columns. It then intersects the lists of pruned columns. While this means that some columns are left unpruned, it makes the job easier for the memoization in the LQPTranslator. In our experiments, this has led to significant performance improvements and negligible reductions.

src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <memory>
+
+#include "logical_query_plan/abstract_lqp_node.hpp"
+
+namespace opossum {
+
+void StoredTableColumnAlignmentRule::apply_to(const std::shared_ptr<AbstractLQPNode>& root) const {
+  StoredTableNodeUnorderedMap<std::vector<std::shared_ptr<StoredTableNode>>> gathered_stored_table_nodes_for_node;
+  StoredTableNodeUnorderedMap<std::vector<ColumnID>> aligned_pruned_column_ids_for_node;
+  recursively_gather_stored_table_nodes(root, gathered_stored_table_nodes_for_node, aligned_pruned_column_ids_for_node);
```
mrks, line 13: Calling a method recursively is not the preferred method of traversing an LQP, as nodes in diamond structures will be visited more than once. Instead, we prefer `visit_lqp`. An additional advantage is that you can simply capture your map(s) by reference.

mrks, line 13: The semantics of the map are a bit unclear to me. What is the key, what is the value?

Have you considered this approach:
1. Iterate over the LQP, store all StoredTableNodes in a column-pruning-agnostic unordered_set
2. For each set:
2.1 Iterate over the nodes and calculate the set intersection of the pruned column_ids.
2.2 Iterate over the nodes and set the new pruning list

src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
```diff
@@ -0,0 +1,63 @@
...
+    gathered_stored_table_nodes.emplace_back(stored_table_node);
+
+    // update maximum superset of pruned column ids for a StoredTableNodes with the same table name and same set of
+    // pruned chunk ids
+    if (aligned_pruned_column_ids_for_node.contains(stored_table_node)) {
+      auto& aligned_pruned_column_ids = aligned_pruned_column_ids_for_node[stored_table_node];
+      std::vector<ColumnID> updated_pruned_column_ids{};
+      updated_pruned_column_ids.reserve(aligned_pruned_column_ids.size() +
+                                        stored_table_node->pruned_column_ids().size());
+      std::set_intersection(aligned_pruned_column_ids.begin(), aligned_pruned_column_ids.end(),
```
mrks, line 51: set_intersection requires that both input containers are sorted. Right now, this is the case just by chance. We shouldn't rely on that being the case. I'd add an assertion for `is_sorted`.

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
Bouncner: "creation and execution of different GetTable operators." > leads to GetTables not being merged?

Also, does this rule only apply to pruned columns or also pruned chunks? The Q21 case would probably still be worth merging the nodes even if one GetTable node has "only a few" chunks more to load. Or am I completely on the wrong path here?

src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
Bouncner: (haven't read the code) If it is sorted just by chance: why not sort the vectors before intersecting?

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
mrks: My gut feeling is that the case of different chunk pruning lists is much more unlikely. The problem tackled in this PR is that we have two operators operating on the same data, but the translator is unable to identify it. If the chunk pruning lists differ, we are not working on the same data any more.

Not sure if this explanation makes perfect sense. In any case, I would only look at it once we actually see it in PQPs.

src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
mrks: is_sorted is cheaper than sort. "just by chance" was probably not the best choice of words. It is sorted because that is how the ColumnPruningRule is written (at least I think so). That won't change any time soon, so we can use that fact as long as we don't silently rely on it.

mweisgut: `StoredTableNode::set_pruned_column_ids(...)` expects a sorted vector:
https://github.com/hyrise/hyrise/blob/32e59e009fae0f2c6497532816eea5cd7cb435e4/src/lib/logical_query_plan/stored_table_node.cpp#L33-L48

Bouncner: Ah ok. Got it.

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
Bouncner: Ok

src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
mrks: I see. In that case, I'd simply add a comment that the sortedness is guaranteed by set_pruned_column_ids.

mrks:
For the tests, I'd additionally check that the hash values and `operator==` are equal after the modifications. After all, that's what matters

src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
mweisgut: I thought about a similar approach where I use vectors instead of unordered sets. But sure, using sets makes more sense. I have abandoned the idea since I wanted to avoid the two iterations over the nodes for each vector/set. However, I think the code would be easier to understand using your mentioned approach. The computation costs for the two iterations should be negligible.

>What is the key, what is the value?

All StoredTableNodes with the same table name and pruned chunks are stored in a vector. `ColumnPruningAgnosticStoredTableNodeMap<std::vector<std::shared_ptr<StoredTableNode>>>
      gathered_stored_table_nodes_for_node;` is used to achieve that. To avoid your first mentioned iteration over all vectors/sets (2.1), the "final pruning list" for each visited StoredTableNode group is updated using `ColumnPruningAgnosticStoredTableNodeMap<std::vector<ColumnID>> aligned_pruned_column_ids_for_node`. As a final step, your mentioned 2.2-iteration is executed.

>Iterate over the LQP, store all StoredTableNodes in a column-pruning-agnostic unordered_set

But we somehow have to group the StoredTableNodes, i.e., put each StoredTableNode in the correct set. We actually don't need the current specialized hash generation and equals comparison. It would be sufficient to have a map with {table_name, pruned_chunks} as key and set<StoredTableNodes> as value. I'll try something like that soon.

Sorry for the long message.

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
mweisgut: I'd agree with that statement.

src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
mrks: Ok, so instead of doing two passes where the first one results in the intersection, you perform the intersection on-the-fly. I'm not sure if this is more efficient. As you mentioned, the iteration itself shouldn't be a cost factor and the actual work is more or less the same, independent of where it happens.

> But we somehow have to group the StoredTableNodes, i.e., put each StoredTableNode in the correct set. We actually don't need the current specialized hash generation and equals comparison. It would be sufficient to have a map with {table_name, pruned_chunks} as key and sets of StoredTableNodes as values. I'll try something like that soon.

The specialized hash/equals is a good approach. My suggestion is simply to use a set with these specializations instead of a map. That way, we have a bucket of all StoredTableNodes that should be combined.

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
mweisgut: > I'm not sure if the comment is helpful for someone who has not seen the "broken" LQP. I wonder if an example would be helpful.

We have an example now. But maybe the .hpp is a little long now. Alternatively, if you think it's too long now, we could put the example on a wiki page and reference it in the .hpp file?

src/lib/optimizer/strategy/stored_table_column_alignment_rule.cpp
```diff
@@ -0,0 +1,89 @@
...
+    for (const auto& pruned_chunk_id : node->pruned_chunk_ids()) {
+      boost::hash_combine(hash, static_cast<size_t>(pruned_chunk_id));
+    }
+    return hash;
+  }
+};
+
+// Modified equals evaluation code for StoredTableNodes where column pruning information is omitted. Struct is used to
+// enable hash-based containers containing std::shared_ptr<StoredTableNode>.
+struct StoredTableNodeSharedPtrEqual final {
```
mrks, line 28: Add a warning to `StoredTableNode::_on_shallow_equals`?

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
mrks: I think comments are more likely to be read and updated than wiki pages. Having to scroll just a bit more is a small price to pay.

src/lib/optimizer/strategy/stored_table_column_alignment_rule.hpp
```diff
@@ -0,0 +1,67 @@
...
+       | Mode: Semi             |     | Mode: Semi             |
+       | predicate: Li_a = Pa_a |     | predicate: Li_a = Pa_a |
+       +------------------------+     +------------------------+
+                   /          \           /           \
+ +---------------------+  +-----------------------+  +-----------------------+
+ | StoredTable (1)     |  | StoredTable (2)       |  | StoredTable (3)       |
+ | table name:     Li  |  | table name:     Pa    |  | table name:     Li    |
+ | pruned chunks:  { } |  | pruned chunks:  {1,2} |  | pruned chunks:  { }   |
+ | pruned columns: {3} |  | pruned columns: { }   |  | pruned columns: {3,4} |
+ +---------------------+  +-----------------------+  +-----------------------+
```
mrks, line 41: Are you sure the rule applies here? I was tempted to make this change:

```
       +------------------------+     +------------------------+
       | Join (2)               |     | Join (3)               |
       | Mode: Semi             |     | Mode: Semi             |
       | predicate: Li_a = Pa_a |     | predicate: Li_a = Pa_a |
       +------------------------+     +------------------------+
             left /          \   right   /           \ left
```

but that would mean that table Li is on the left side of the semi join in both cases. As a result, Pa would not be available for the final join.

If, however, the left and right inputs for join (2) and (3) were as they are in the image, it would mean that the joins would not compare equal even after the rule is applied.

mweisgut: The example was inspired by TPC-H Q17 and the lineitem table was at the left side for both join nodes. So yes, your descriptions of the sides are correct but it's also true that Join (1) is not correct then. How about changing Join (2) and (3) to Inner and Join (1) to Semi?

PR #2137, opened 2020-05-18
Finalize operator results

### LINKED ISSUE #2131: Tables created by operators should be immutable
Non-data tables should be immutable. Right now, we neither finalize() all chunks in several operators nor enforce it.

With my current work on #2000, I am already working on this issue. There will be another PR tackling it.

Bouncner:
To set sort orders and table clustering in #2000, several operators needed to first finalize their results. With that we realized that all read-only operators should always `finalize()` their results and these are immutable.

This PR adds the last few `finalize()` calls to operators and adds a DebugAssert in `abstract_operators.cpp` to check for it.

Fixes #2131

### DIFF
--- a/src/lib/operators/abstract_operator.cpp
+++ b/src/lib/operators/abstract_operator.cpp
@@ -70,8 +70,8 @@ void AbstractOperator::execute() {
   // Chunks from operators other than GetTable should be immutable
   if constexpr (HYRISE_DEBUG) {
     // We exclude the TableWrapper (heavily used in tests with arbitrary tables) and modifiying operators.
-    if (const auto ro_operator = std::dynamic_pointer_cast<const AbstractReadOnlyOperator>(_input_left) &&
-                                 _input_left->type() != OperatorType::TableWrapper) {
+    if (std::dynamic_pointer_cast<const AbstractReadOnlyOperator>(_input_left) &&
+        _input_left->type() != OperatorType::TableWrapper) {
       const auto check_input_table = [&](const std::shared_ptr<const Table>& input_table) {
         const auto chunk_count = input_table->chunk_count();
         for (auto chunk_id = ChunkID{0u}; chunk_id < chunk_count; ++chunk_id) {
--- a/src/lib/operators/abstract_operator.cpp
+++ b/src/lib/operators/abstract_operator.cpp
@@ -67,6 +67,30 @@ void AbstractOperator::execute() {
                 _output ? _output->row_count() : 0, _output ? _output->chunk_count() : 0,
                 reinterpret_cast<uintptr_t>(this));
 
+  // Chunks from operators other than GetTable should be immutable
+  if constexpr (HYRISE_DEBUG) {
+    // We exclude the TableWrapper (heavily used in tests with arbitrary tables) and modifiying operators.
+    if (const auto ro_operator = std::dynamic_pointer_cast<const AbstractReadOnlyOperator>(_input_left) &&
+                                 _input_left->type() != OperatorType::TableWrapper) {
+      const auto check_input_table = [&](const std::shared_ptr<const Table>& input_table) {
+        const auto chunk_count = input_table->chunk_count();
+        for (auto chunk_id = ChunkID{0u}; chunk_id < chunk_count; ++chunk_id) {
+          const auto chunk = input_table->get_chunk(chunk_id);
+          Assert(chunk, "Physically deleted chunk should not reach this point, see get_chunk / #1686.");
+          DebugAssert(input_table->type() == TableType::Data || !chunk->is_mutable(),
+                      std::string{"Operator "} + _input_left->name() + " received an unexpected mutable chunk.");
+        }
+      };
+
+      if (_input_left) {
+        check_input_table(_input_left->get_output());
+      }
+      if (_input_right) {
+        check_input_table(_input_right->get_output());
+      }
+    }
+  }
+
   // Verify that LQP (if set) and PQP match.
   if constexpr (HYRISE_DEBUG) {
     if (lqp_node) {
--- a/src/lib/operators/abstract_join_operator.cpp
+++ b/src/lib/operators/abstract_join_operator.cpp
@@ -103,6 +103,15 @@ std::shared_ptr<Table> AbstractJoinOperator::_build_output_table(std::vector<std
     }
   }
 
+  const auto chunk_count = chunks.size();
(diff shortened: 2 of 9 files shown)


Bouncner:
Would like to get your opinion on fc6bb3e, @mrks . Do you know which queries require this change? I did not expect pruned tables to be mutable.

mrks:
Ah, that's an issue. You are taking the data chunk from a table that is still mutable and mark it as mutable. As such, you are modifying the actual data table and essentially break inserts. I'm a bit unhappy that this isn't caught as part of the regular tests. :/

Bouncner:
GetTable is used in various paths with modifying operators. So it must also emit mutable chunks.

The actual problem above is that I thought that all chunks that are column-pruned are per se chunks which can are immutable and I thus finalized them. This crashed. Apparently, we apply column pruning to chunks which are later modified.

As we emit mutable chunks anyways, I'd vote to keep it simple and check in the column-pruned part whether the base chunk is immutable and only finalize in this case.

mrks:
What about AggregateSort and Projection? Both produce ValueSegments and they don't seem to be touched here.

The fact that (1) no other code had to be changed in this PR and (2) these oversights did not pop up anywhere makes me wonder if we chose the best approach.

Off the top of my head, we have the following combinations:

|   | Chunk containing...    | inserts? | invalidations? | set_ordered_by? | max_begin_cid?                           |
|---|------------------------|----------|----------------|-----------------|------------------------------------------|
| 1 | mutable ValueSegment   | yes      | yes            | no              | no                                       |
| 2 | immutable ValueSegment | no       | yes            | yes             | yes                                      |
| 3 | temporary ValueSegment | no       | no             | yes             | no                                       |
| 4 | EncodedSegment         | no       | yes            | yes             | yes                                      |
| 5 | ReferenceSegment       | no       | yes            | yes             | no (indirect if references single chunk) |

This does not really seem to fit the current `_is_mutable` / `finalize()` approach.

We run into additional issues once we actually want to set the max_begin_cid after an insert. The transaction that reserves the last row should set the mutable flag before it releases the append lock. The last transaction that commits its inserts on the table should calculate the max_begin_cid. Currently, both steps are combined in `finalize()`.

mrks:
In other words: We should clarify the lifecycle of a chunk and clearly define who is responsible for the state transitions.

Bouncner:
> What about AggregateSort and Projection? Both produce ValueSegments and they don't seem to be touched here.

This PR tackles the "left overs" from #2000 + adds the test. Projection etc. already finalize. hyriseTest fails if they wouldn't. So that should cover (1) and (2) I guess.

I have to think about the max cid stuff. Haven't worked with that yet.

mrks:
Screenshot 2020-05-20 at 13 15 38

src/lib/operators/abstract_operator.cpp
```diff
@@ -67,6 +67,30 @@ void AbstractOperator::execute() {
...
 
+  // Chunks from operators other than GetTable should be immutable
+  if constexpr (HYRISE_DEBUG) {
+    // We exclude the TableWrapper (heavily used in tests with arbitrary tables) and modifiying operators.
+    if (std::dynamic_pointer_cast<const AbstractReadOnlyOperator>(_input_left) &&
+        _input_left->type() != OperatorType::TableWrapper) {
+      const auto check_input_table = [&](const std::shared_ptr<const Table>& input_table) {
+        const auto chunk_count = input_table->chunk_count();
+        for (auto chunk_id = ChunkID{0u}; chunk_id < chunk_count; ++chunk_id) {
+          const auto chunk = input_table->get_chunk(chunk_id);
```
mweisgut, line 78:
```suggestion
          const auto& chunk = input_table->get_chunk(chunk_id);
```

Bouncner: That wouldn't work with the following assert as references cannot be null. `get_chunk()` however can return a nullptr (see `remove_chunk()`).

mrks: Nah, it would be a reference to a shared_ptr, which can be null. For data tables, we would have a lifetime extension, but for reference tables, we could save a shared_ptr copy

Bouncner: Ah I see. Then again: is there a reason we do it differently pretty much everywhere? Should the operator really be part of ensuring that a chunk is not unwillingly freed?

mrks: Chunks should not suddenly disappear, so it probably is a small inconsistency.

Bouncner: Do we a common understanding/guide which component ensure that chunks are kept alive during execution of an operator?

mrks:
After some thinking, this might be easier than I originally thought:

* We do not need to track whether a chunk accepts deletes. If it has MVCC data, it does. Otherwise, the MVCC data would be pointless.
* For the same reason, inserts into temporary chunks are already impossible.
* Now that we pre-allocate the ValueSegments, we can simply use that information to track whether a chunk still accepts inserts. If there is capacity left, it does, otherwise it doesn't.

As such, how about this?

* `is_mutable` becomes `accepts_inserts`, which is false if (1) we have no MvccData, (2) the first segment is not a ValueSegment, or (3) its size is equal to its capacity.
* This is sufficient as a precondition for `set_sorted_by` for both data and temporary tables
* For temporary ValueSegments, which call the `ValueSegment(pmr_vector<T>&& values)` constructor, we could either call shrink_to_fit or we store the capacity explicitly.

* As a result, we do not explicitly finalize chunks anymore.
* Identifying the max_begin_cid is done explicitly as `identify_max_begin_cid`. Currently, this would only be called in the benchmarks. In the future, this is done asynchronously, just like the encoding.
* `store_max_begin_cid` checks for (1) `!accepts_inserts` and (2) `MAX_COMMIT_ID` (same as right now)

Bouncner:
Will look at it again, when I find the time.

Bouncner:
Will look at it again, when I find the time.

Bouncner:
I'll need to take a look into this max commit id thing before I can comment.

While it's not needed for the sorting question, I wonder if we also need `accepts_modifications` for chunks that are full or not and also store MVCC information.

mrks:
How would it be different from just checking for the presence of MVCC information?

Bouncner:
That's probably enough. I just wanted to mention it because I am not sure if we need it or not. From a "is the interface conclusive and complete"-stand point, the function (as simple as it might be) would make sense.

Bouncner:
Check #2609 when re-evaluating this topic.

PR #2139, opened 2020-05-19
Console: Streamline exception handling

### LINKED ISSUE #1406: Console style things
As noted [here](https://github.com/hyrise/hyrise/pull/1400#discussion_r247517757), I take issue with the console overusing try/catch. 

Also, this:

```
    Assert(!_sql_pipeline->failed_pipeline_statement(),
           "The transaction has failed. This should never happen in the console, where only one statement gets "
           "executed at a time.");
```

implies `_sql_pipeline->failed_pipeline_statement(),` is always false in the console, yet `Console::_handle_rollback()` thinks it might be true. Speaking of `_handle_rollback`, the name implies mvcc rollback(?) but is called even after optimization threw an exception (`Console::_visualize`) or after PQP visualization failed. 

My takeaway would be to get rid of all non-InvalidInputException handling, because it evidently just leads into confusion. I'd rather have the console exit more than necessary than it continuing to operate with hyrise in undefined state or it performing totally unrelated error handling (`_handle_rollback`) after arbitrary exceptions.

mrks:
fixes #1406

As was rightly said over there, we should not unnecessarily handle exceptions in the console. The most common issues (broken SQL statement, invalid table name, ...) are InvalidInputExceptions. The runtime_errors that we might see in other places should kill the console.


src/bin/console/console.cpp
```diff
@@ -718,40 +719,30 @@ int Console::_visualize(const std::string& input) {
...
-      } catch (const std::exception& exception) {
-        out(std::string(exception.what()) + "\n");
-        return ReturnCode::Error;
+      const auto& lqps = (plan_type == PlanType::LQP) ? _sql_pipeline->get_optimized_logical_plans()
+                                                      : _sql_pipeline->get_unoptimized_logical_plans();
+
+      lqp_roots.reserve(lqps.size());
+
+      for (const auto& lqp : lqps) {
+        lqp_roots.push_back(lqp);
```
mweisgut, line 728: Let's use `emplace_back` to avoid the extra copy or move operation

mrks:
No idea what LSAN is doing :/

src/bin/console/console.cpp
Bouncner: Out of curiosity: does that change anything? We need to add a new copy of the inital lqp shrptr anyways and the loop just holds a reference. So we cannot move it.

Or is there more behind it?

mrks: Probably not.

https://stackoverflow.com/questions/10890653/why-would-i-ever-use-push-back-instead-of-emplace-back

mweisgut: Both `push_back` and `emplace_back` call the constructor but only `emplace_back` construct the element in-place at the location provided by the container. `push_back` creates a temporary object which is then moved to the new position in the container.

Please correct me if I'm wrong.

mrks: https://wandbox.org/permlink/Q9TV2fIzhRASFOHv

mrks:
That's a CI issue in any case.

Did you also look at the original issue? It would be helpful to know if you have the same understanding and consider it resolved.

mrks:
> Were these intentionally omitted?

Should have written something about them I guess.

Yes, I would expect that these are somewhat likely to fail in the normal use of the console but are unlikely to leave Hyrise in an undefined state. When working with different data sets, I saw exceptions from there from time to time, for example when I was building an appropriate csv.json.

PR #2145, opened 2020-05-27
Cleanup optimizer tests

### LINKED ISSUE #1024: Cleanup optimizer tests
As discussed in #1022, we should change the *_rule_test.cpp to use `EXPECT_LQP_EQ` instead of simply checking for moved pointers.

> [W]hen implementing optimizer rules you have to make sure the rule doesn't break the LQP. Just testing the presence of nodes can easily miss bugs.
> 
> Second point: If you test the entire LQP, then the reader of the code can be sure the test actually tests everything the rule might or might not do to the LQP. If you just test a selected number of node types, the reader really has to study your test to figure out whether it might miss anything. Which the reader won't do, cause he lazy. Enter error proneness.

mweisgut:
fixes #1024 

#### TODO
- [x] use `EXPECT_LQP_EQ`
- [x] load tables only once in `SetUpTestSuite`
- [x] fix [sub-query node mapping issue](https://github.com/hyrise/hyrise/pull/2145#discussion_r431164962) (previously built input_lqp was incorrect which was not caught by the test cases)


src/test/lib/optimizer/strategy/predicate_placement_rule_test.cpp
```diff
@@ -247,59 +254,76 @@ TEST_F(PredicatePlacementRuleTest, ComplexBlockingPredicatesPushdownTest) {
...
+  auto actual_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
 
-  EXPECT_EQ(reordered, _projection_pushdown_node);
-  EXPECT_EQ(reordered->left_input(), predicate_node);
-  EXPECT_EQ(reordered->left_input()->left_input(), _table_a);
+  // TODO(Marcel) If a node mapping is created in `lqp_find_subplan_mismatch` which is called in `EXPECT_LQP_EQ`, the
+  // nodes of sub-queries are not mapped. Consequently, `expression_adapt_to_different_lqp` cannot find the
+  // StoredTableNode of the sub-query in the node mapping. Therefore, an assertion fails: "Couldn't find referenced
+  // node x in NodeMapping."
+  EXPECT_LQP_EQ(actual_lqp, expected_lqp);
```
mweisgut, line 276: This is the issue that I tried to explain during today's Fortnightly. If anyone had this issue before and knows a good way to solve it, it would be a pleasure to get feedback.

mrks:
> This is the issue that I tried to explain during today's Fortnightly. If anyone had this issue before and knows a good way to solve it, it would be a pleasure to get feedback.

This is what Github sent me via mail, but I can't see the comment here. Did you remove it?

src/test/lib/optimizer/strategy/predicate_placement_rule_test.cpp
mweisgut, line 276: This is the issue that I tried to explain during today's Fortnightly. If anyone had this issue before and knows a good way to solve it, it would be a pleasure to get feedback.

mrks: It's somewhat surprising, but it seems like EXPECT_LQP_EQ was never used with subplans.
 PreparedPlanTest.InstantiateHashEqual might be an exception, but it doesn't use StoredTableNodes. As far as I understand it, those are the reason why we need the mapping at all.

Maybe (!!!) something like this?

```
LQPNodeMapping lqp_create_node_mapping(const std::shared_ptr<AbstractLQPNode>& lhs,
                                       const std::shared_ptr<AbstractLQPNode>& rhs) {
  LQPNodeMapping mapping;
  lqp_create_node_mapping_impl(mapping, lhs, rhs);

  auto left_roots = lqp_find_subplan_roots(lhs);
  auto right_roots = lqp_find_subplan_roots(rhs);
  // Somehow map entry in left_roots to right_roots
  lqp_create_node_mapping_impl(mapping, lhs_subplan, rhs_subplan);

  return mapping;
}
```

mweisgut: It was actually the `input_lqp` that was built incorrectly.

mweisgut:
During the review, please check in particular whether the input and the expected lqps are built correctly.

src/test/lib/optimizer/strategy/predicate_placement_rule_test.cpp
```diff
@@ -21,121 +21,133 @@ namespace opossum {
...
-    _d_b = _table_d->get_column("b");
-
-    Hyrise::get().storage_manager.add_table("e", load_table("resources/test_data/tbl/int_int4.tbl"));
-    _table_e = std::make_shared<StoredTableNode>("e");
-    _e_a = _table_e->get_column("a");
-    _e_b = _table_e->get_column("b");
+    Hyrise::get().storage_manager.add_table("a", _table_a);
+    _stored_table_a = StoredTableNode::make("a");
+    _a_a = _stored_table_a->get_column("a");
+    _a_b = _stored_table_a->get_column("b");
```
mrks, line 36: Now that I am seeing this again, I realize that the proper way to work around the reset of the StorageManager would have been to either use a MockNode with statistics or a StaticTableNode. No need to change this again, just wanted to let you know.

src/test/lib/optimizer/strategy/predicate_placement_rule_test.cpp
```diff
@@ -21,121 +21,133 @@ namespace opossum {
...
-
-      auto parameter_c = correlated_parameter_(ParameterID{0}, a);
-      auto lqp_c = AggregateNode::make(expression_vector(), expression_vector(max_(add_(a, parameter_c))),
-                                       ProjectionNode::make(expression_vector(add_(a, parameter_c)), int_float_node_a));
-
-      _subquery_c = lqp_subquery_(lqp_c, std::make_pair(ParameterID{0}, a));
+      auto parameter_c = correlated_parameter_(ParameterID{0}, _a_a);
+      auto lqp_c =
+          AggregateNode::make(expression_vector(), expression_vector(max_(add_(_a_a, parameter_c))),
+                              ProjectionNode::make(expression_vector(add_(_a_a, parameter_c)), _stored_table_a));
```
mrks, line 65: Format with clang-format and indentation, just as you did below

src/test/lib/optimizer/strategy/predicate_placement_rule_test.cpp
```diff
@@ -21,121 +21,133 @@ namespace opossum {
...
   }
 
+  inline static std::shared_ptr<Table> _table_a, _table_b, _table_c, _table_d, _table_e;
   std::shared_ptr<CorrelatedParameterExpression> _parameter_a_a;
   std::shared_ptr<AbstractLQPNode> _subquery_lqp;
   std::shared_ptr<PredicatePlacementRule> _rule;
-  std::shared_ptr<StoredTableNode> _table_a, _table_b, _table_c, _table_d, _table_e;
+  std::shared_ptr<StoredTableNode> _stored_table_a, _stored_table_b, _stored_table_c, _stored_table_d, _stored_table_e;
   std::shared_ptr<LQPColumnExpression> _a_a, _a_b, _b_a, _b_b, _c_a, _c_b, _d_a, _d_b, _e_a, _e_b;
   std::shared_ptr<ProjectionNode> _projection_pushdown_node;
```
mrks, line 81: Is this one still used? Same for others.

src/test/lib/optimizer/strategy/predicate_reordering_rule_test.cpp
```diff
@@ -114,30 +114,30 @@ TEST_F(PredicateReorderingTest, SameOrderingForStoredTable) {
...
-  auto reordered_1 = StrategyBaseTest::apply_rule(_rule, predicate_node_3);
+  const auto first_actual_lqp = StrategyBaseTest::apply_rule(_rule, first_input_lqp);
+  const auto second_actual_lqp = StrategyBaseTest::apply_rule(_rule, second_input_lqp);
 
-  EXPECT_EQ(reordered, predicate_node_1);
-  EXPECT_EQ(reordered->left_input(), predicate_node_0);
-  EXPECT_EQ(reordered_1, predicate_node_2);
-  EXPECT_EQ(reordered_1->left_input(), predicate_node_3);
+  EXPECT_LQP_EQ(first_actual_lqp, first_expected_lqp);
+  EXPECT_LQP_EQ(second_actual_lqp, second_expected_lqp);
```
mrks, line 140: These two tests seem independent. Reorder and put into scopes?

src/test/lib/optimizer/strategy/predicate_reordering_rule_test.cpp
```diff
@@ -149,11 +149,11 @@ TEST_F(PredicateReorderingTest, PredicatesAsRightInput) {
...
      *             _______Cross________
      *            /                    \
-     *  Predicate_0(a > 80)     Predicate_2(a > 90)
+     *  Predicate(a > 80)     Predicate(a > 90)
      *           |                     |
-     *  Predicate_1(a > 60)     Predicate_3(a > 50)
+     *  Predicate(a > 60)     Predicate(a > 50)
      *           |                     |
-     *        Table_0           Predicate_4(a > 30)
+     *        Table_0           Predicate(a > 30)
```
mrks, line 156: Formatting

src/test/lib/optimizer/strategy/predicate_reordering_rule_test.cpp
```diff
@@ -218,25 +218,33 @@ TEST_F(PredicateReorderingTest, PredicatesWithMultipleOutputs) {
...
-  EXPECT_EQ(reordered, union_node);
-  EXPECT_EQ(reordered->left_input(), predicate_a_node);
-  EXPECT_EQ(reordered->right_input(), predicate_c_node);
-  EXPECT_EQ(predicate_a_node->left_input(), predicate_c_node);
-  EXPECT_EQ(predicate_c_node->left_input(), predicate_b_node);
-  EXPECT_EQ(predicate_b_node->left_input(), table_node);
+
+  // clang-format off
+  const auto sub_lqp =
+  PredicateNode::make(greater_than_(lqp_column_(table_node, ColumnID{0}), 10),
```
mrks, line 224: We also have StoredTableNode::get_column (same for mock and static nodes)

src/test/lib/optimizer/strategy/predicate_reordering_rule_test.cpp
mweisgut: In what way? What's the issue?

mrks: "Predicate" is not aligned with the Predicates above anymore. Really minor though. ;)

src/test/lib/optimizer/strategy/predicate_reordering_rule_test.cpp
mrks: (No need to change this. We use `lqp_column_` in other places too)

PR #2146, opened 2020-05-27
Clarify benchmark information text

Bouncner:
When using 1 client, the benchmark states "1 simulated clients is scheduling items in parallel".

For one client, it's now "1 simulated client is scheduling items".

### DIFF
--- a/scripts/test/hyriseBenchmarkFileBased_test.py
+++ b/scripts/test/hyriseBenchmarkFileBased_test.py
@@ -29,7 +29,7 @@ def main():
 
   benchmark.expect_exact("Writing benchmark results to 'json_output.txt'")
   benchmark.expect_exact("Running in single-threaded mode")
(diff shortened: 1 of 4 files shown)


scripts/test/hyriseBenchmarkFileBased_test.py
```diff
@@ -29,7 +29,7 @@ def main():
 
   benchmark.expect_exact("Writing benchmark results to 'json_output.txt'")
   benchmark.expect_exact("Running in single-threaded mode")
-  benchmark.expect_exact("1 simulated clients are scheduling items in parallel")
+  benchmark.expect_exact("1 simulated clients is scheduling items")
```
mweisgut, line 32:
```suggestion
  benchmark.expect_exact("1 simulated client is scheduling items")
```

PR #2147, opened 2020-05-31
Streamline implementation of Table Constraints

julianmenzler:
Currently, Hyrise has support for soft unique constraints in data tables (including PRIMARY_KEY and UNIQUE key constraints). Latter constraints are called "soft" since they are not enforced during inserts, updates, and deletes. However, they can still serve various optimization purposes.

This PR streamlines both, interface and implementation of the aforementioned unique constraints. 

### Constraint Specification

Before this PR, the `TableConstraintDefinition` struct served as a container for the specification of unique constraints. The name could imply the modeling of various SQL constraints, which it did not.

This PR adds more clarity and extensibility to Hyrise's table constraints implementation. It

* introduces a new class called `AbstractTableConstraint`, serving as a basis for various table constraints. (e.g. for referential constraints and check constraints we may want to add in the future)
* replaces the struct TableConstraintDefinition with `TableKeyConstraint`, a subclass of AbstractTableConstraint.

Instead of using a vector for modeling column sets, AbstractTableConstraint uses an unordered_set which lead to the removal of some sorting and duplicate prevention logic.

### Table interface changes 

Class Table no longer constructs constraints on its own. In addition, a new getter has been added:

```cpp 
  void add_soft_key_constraint(const TableKeyConstraint& table_key_constraint);
```


src/lib/logical_query_plan/mock_node.hpp
```diff
@@ -24,7 +25,8 @@ class MockNode : public EnableMakeForLQPNode<MockNode>, public AbstractLQPNode {
  public:
   using ColumnDefinitions = std::vector<std::pair<DataType, std::string>>;
 
-  explicit MockNode(const ColumnDefinitions& column_definitions, const std::optional<std::string>& init_name = {});
+  explicit MockNode(const ColumnDefinitions& column_definitions, const std::optional<std::string>& init_name = {},
+                    const TableUniqueConstraints& constraints = {});
```
mrks, line 29: Can we stick with `set_unique_constraints`? It feels more consistent to have only one way of setting these

src/lib/storage/constraints/table_unique_constraint.cpp
```diff
@@ -0,0 +1,14 @@
+#include "table_unique_constraint.hpp"
+
+namespace opossum {
+
+TableUniqueConstraint::TableUniqueConstraint() : TableKeyConstraint({}, KeyConstraintType::NONE) {}
```
mrks, line 5: Why is "none" a valid constraint type?

src/lib/storage/constraints/table_unique_constraint.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include "table_key_constraint.hpp"
+
+namespace opossum {
+
+/**
+ * Container to define a unique constraint for a given set of column ids.
+ * Validity checks take place when unique constraints are added to tables. For example, checks for nullability in
+ * case of PRIMARY_KEY constraints.
+ */
+class TableUniqueConstraint final : public TableKeyConstraint {
```
mrks, line 12: What is the purpose of this class? It doesn't seem to add anything to TableKeyConstraint.

src/lib/logical_query_plan/mock_node.hpp
julianmenzler: ok

src/lib/storage/constraints/table_unique_constraint.cpp
julianmenzler: Simple declarations like this
```cpp
  TableUniqueConstraint _unique_constraint_a;
  TableUniqueConstraint _unique_constraint_b_c;
  TableUniqueConstraint _unique_constraint_x;
  TableUniqueConstraint _unique_constraint_y;
```
require me to define a default constructor. Full initialization can not take place on startup because the SetUp() routine has to create tables first.

I did not want to create a dummy PRIMARY KEY or UNIQUE constraint. Therefore, I added NONE as another type. How would you do it?

mrks: Tests should usually not impact the design of the main code. If it's only those four lines, maybe we can just use optional<>?

julianmenzler:
> Could you please describe the design rationale for adding inheritance to the constraints? I am not yet seeing where this will help us.

I decided on this for clarity. 

* I do not like `TableConstraintDefinition` because it sounds too general.
* I do not like having `TableUniqueConstraint` only because it misses the relation to (other) SQL constraints.

Further considerations:

* Functions like `Table::add_soft_unique_constraint` require a certain subset of table constraints (Primary Key and Unique). A specialized type, such as `TableUniqueConstraint` expresses that requirement. Also, the compiler assists with type checks. We do not have to filter input vectors for certain constraint types.

* Different to SQL constraints, the upcoming expression-based constraints in LQPs are no longer bound to tables. However, since they model uniqueness, I'd like to give them such as LQPUniqueConstraint or ExpressionUniqueConstraint. Latter names would go well with `TableUniqueConstraint`.

I knew this discussion would come up. But, I hope the above points outline my reasoning  😄

src/lib/storage/constraints/table_unique_constraint.hpp
julianmenzler: see https://github.com/hyrise/hyrise/pull/2147#issuecomment-637642816

src/lib/storage/constraints/table_unique_constraint.cpp
julianmenzler: Good suggestion! I'll go with optional<>.

mrks:
Got it. When thinking about additional constraints, including foreign keys and checks, this seems to make sense in general. Some thoughts for discussion:

* Yesterday, we had the discussion about "default constraints". The SQL standard does not seem to consider default values as constraints (and I can't see why it would):

```
         <column definition> ::=
              <column name> { <data type> | <domain name> }
              [ <default clause> ]
              [ <column constraint definition>... ]
              [ <collate clause> ]

         <column constraint definition> ::=
              [ <constraint name definition> ]
              <column constraint>
                [ <constraint attributes> ]

         <column constraint> ::=
                NOT NULL
              | <unique specification>
              | <references specification>
              | <check constraint definition>
```

* `NOT NULL` is treated as a column constraint, however, in Hyrise, it is already stored in TableColumnDefinition. I would keep it that way. The documentation of the constraints should reflect this.

* SQL makes a difference between column and table constraints. I don't see an advantage in treating unique (single-)column constraints different from multi-column constraints. Some time in the future, it should be the job of the SQLTranslator to convert single-column unique constraints to the more general multi-column version.

```
         <table constraint> ::=
                <unique constraint definition>
              | <referential constraint definition>
              | <check constraint definition>

         <unique constraint definition> ::=
              <unique specification>
                <left paren> <unique column list> <right paren>

         <unique specification> ::=
              UNIQUE | PRIMARY KEY

         <unique column list> ::= <column name list>
```

* I agree that subsuming UNIQUE and PRIMARY KEY under "unique specification" might not be the best naming. 

* Looking at these multi-column constraints, I wonder if `TableKeyConstraint` is the optimal name for the super class. A check constraint does not have to be on a key column. A referential constraint *references* a key column, but the affected columns in the table in question are not keys. How about simply *AbstractTableConstraint* for the super class?

* The super class should be abstract. If you want to use OOP for defining the type, it should not have a KeyConstraintType field, as that duplicates information. Otherwise, we would have to rule out a TableUniqueConstraint with ConstraintType::Foreign.

* "Validity checks take place when unique constraints are added to tables." - currently, they are not.

mweisgut:
>But, as of now, the struct models unique constraints (PRIMARY KEY and UNIQUE) only.
>- To avoid confusion, the struct's name is changed to TableUniqueConstraint

Regarding the SQL standard, a _table constraint_ is either a _unique constraint_, _primary key constraint_, _referential constraint_, or a _check constraint_. With the current proposed approach, a `TableUniqueConstraint` could be both PRIMARY KEY and UNIQUE, right? Modeling it closer to the standard might reduce the potential of confusion.

Having an `AbstractTableConstraint` class and `Table<Specification>Constraint` subclasses seems to be a clean structure and would be consistent with other classes in Hyrise that have specialized implementations.

>We could also think about TableKeyConstraint, as both PRIMARY and UNIQUE define a key.

`TableKeyConstraint` as a superclass for `TableUniqueConstraint` and `TablePrimaryKeyConstraint` or as a struct/class that internally holds a type (PRIMARY or UNIQUE)?

mrks:
> TableKeyConstraint as a superclass for TableUniqueConstraint and TablePrimaryKeyConstraint or as a struct/class that internally holds a type (PRIMARY or UNIQUE)?

I think I'd go for the latter, as they are treated the same for most cases and their implementation does not differ.

julianmenzler:
At first, thanks a lot for your extensive feedback @mrks.
Too bad, there is no official SQL spec easily accessible on the web. 

> How about simply _AbstractTableConstraint_ for the super class? You mention that you "do not like TableConstraintDefinition because it sounds too general", but I guess the abstract constraint simply is that general.

AbstractTableConstraint is great.  

> You say "I do not like having TableUniqueConstraint only because it misses the relation to (other) SQL constraints". By "other constraints", do you mean primary keys? We could also think about TableKeyConstraint, as both PRIMARY and UNIQUE define a key.

I also thought of other types such as FOREIGN KEY and CHECK constraints. Name-wise, AbstractTableConstraint leaves room for those, whereas TableUniqueConstraint seems kind of restricted.

> "Validity checks take place when unique constraints are added to tables." - currently, they are not.

Well, at least we have the following check in place:
```cpp
    // PRIMARY KEY requires non-nullable columns
    if (table_key_constraint.type() == KeyConstraintType::PRIMARY_KEY) {
      Assert(!column_is_nullable(column_id), "Column must be non-nullable to comply with PRIMARY KEY.");
    }
```

> Do we still track PRIMARY KEY in the LQP? Does it make a difference? If not, we should avoid naming symmetries with the information stored in the table (which tracks PRIMARY information).

No, PRIMARY KEY becomes irrelevant in LQPs.

julianmenzler:
Ready for the next review.

src/lib/storage/table.cpp
```diff
@@ -323,38 +323,36 @@ void Table::set_table_statistics(const std::shared_ptr<TableStatistics>& table_s
...
-                        [&new_constraint](const auto& existing_constraint) {
-                          return new_constraint.columns == existing_constraint.columns;
-                        }) == _constraint_definitions.end(),
-           "Another constraint on the same columns already exists.");
+    for (const auto& existing_constraint : _table_key_constraints) {
+      // Ensure that no other PRIMARY KEY is defined
+      Assert(existing_constraint.key_type() == KeyConstraintType::UNIQUE ||
+                 table_key_constraint.key_type() == KeyConstraintType::UNIQUE,
+             "Another primary key already exists for "
+             "this table.");
```
mrks, line 349: format

src/lib/storage/table.cpp
```diff
@@ -323,38 +323,36 @@ void Table::set_table_statistics(const std::shared_ptr<TableStatistics>& table_s
...
 
-    Assert(std::find_if(_constraint_definitions.begin(), _constraint_definitions.end(),
-                        [&new_constraint](const auto& existing_constraint) {
-                          return new_constraint.columns == existing_constraint.columns;
-                        }) == _constraint_definitions.end(),
-           "Another constraint on the same columns already exists.");
+    for (const auto& existing_constraint : _table_key_constraints) {
+      // Ensure that no other PRIMARY KEY is defined
+      Assert(existing_constraint.key_type() == KeyConstraintType::UNIQUE ||
+                 table_key_constraint.key_type() == KeyConstraintType::UNIQUE,
```
mrks, line 347: Why are you doing this check over and over again? Move outside the for loop?

src/lib/storage/constraints/table_key_constraint.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include "abstract_table_constraint.hpp"
+
+namespace opossum {
+
+enum class KeyConstraintType : bool { PRIMARY_KEY, UNIQUE };
```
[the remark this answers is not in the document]
julianmenzler: I thought of memory here. But your argument makes sense. Thanks!

src/lib/storage/constraints/abstract_table_constraint.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+  bool operator==(const AbstractTableConstraint& rhs) const;
+  bool operator!=(const AbstractTableConstraint& rhs) const;
+
+ protected:
+  /**
+   * Compare two table constraints of the same type. Only additional fields have to be compared since column ids are
+   * already compared by the caller.
+   */
+  virtual bool _on_shallow_equals(const AbstractTableConstraint& table_constraint) const = 0;
```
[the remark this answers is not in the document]
julianmenzler: Good to know. I thought "shallow" is a naming convention for subclass equals-helper-functions.

src/lib/storage/table.cpp
julianmenzler: To check for existing primary keys we need to iterate over all existing constraints. The implementation above requires only one for loop. Would you prefer two for loops anyway?

mrks: Are we talking about the same thing? If `table_key_constraint.key_type() == KeyConstraintType::UNIQUE`, the entire check loop seems pointless.

src/lib/logical_query_plan/mock_node.cpp
```diff
@@ -118,14 +124,17 @@ size_t MockNode::_on_shallow_hash() const {
 std::shared_ptr<AbstractLQPNode> MockNode::_on_shallow_copy(LQPNodeMapping& node_mapping) const {
   const auto mock_node = MockNode::make(_column_definitions, name);
   mock_node->set_table_statistics(_table_statistics);
+  mock_node->set_soft_key_constraints(_table_key_constraints);
+  mock_node->set_functional_dependencies(_functional_dependencies);
```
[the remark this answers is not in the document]
julianmenzler: The constraints will be required for test cases in the upcoming LQP unique constraints PR. However, it is not planned to generate FDs based on table constraints as it is done by StoredTableNodes. MockNodes should only provide container functionality for FDs and table constraints.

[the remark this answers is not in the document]
julianmenzler: Shall I remove it anyway?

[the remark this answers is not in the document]
mrks: How about we allow only one of the two to be set (to avoid inconsistencies) and prevent the FDs to be retrieved when soft key constraints are set (to avoid the user expecting the FDs to be derived from the constraints)?

src/lib/storage/table.cpp
julianmenzler: Correct, in that case, the above Assert will always return true.  But we still need the Assert below, checking for equal column sets.

mrks: ok

src/lib/logical_query_plan/mock_node.cpp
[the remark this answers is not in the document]
julianmenzler: Hmm, is this still simple enough for MockNode? Since MockNode is for testing purposes only, I could live with pure container functionality. 

I just noticed: In case of MockNode, it does not make sense to use the term "soft" for constraints. There is no data anyway. Nothing to check for.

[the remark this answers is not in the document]
mrks: > Hmm, is this still simple enough for MockNode? Since MockNode is for testing purposes only, I could live with pure container functionality.

We wouldn't add any logic, we just add assertions to prevent some constellations that are probably prone to errors.

> I just noticed: In case of MockNode, it does not make sense to use the term "soft" for constraints. There is no data anyway. Nothing to check for.

Good point. I can see arguments for both changing it and keeping it (to mirror the existing calls in a StoredTableNode). Your call.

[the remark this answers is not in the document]
julianmenzler: > We wouldn't add any logic, we just add assertions to prevent some constellations that are probably prone to errors.

Good idea. I added one.

> Good point. I can see arguments for both changing it and keeping it (to mirror the existing calls in a StoredTableNode). Your call.

Ok. I decided against "soft" to avoid confusion.

src/lib/logical_query_plan/mock_node.cpp
```diff
@@ -103,11 +97,22 @@ void MockNode::set_table_statistics(const std::shared_ptr<TableStatistics>& tabl
...
 
-const TableKeyConstraints& MockNode::soft_key_constraints() const { return _table_key_constraints; }
+const TableKeyConstraints& MockNode::key_constraints() const { return _table_key_constraints; }
+
+void MockNode::set_functional_dependencies(const std::vector<FunctionalDependency>& fds) {
+  _functional_dependencies = fds;
+}
+
+std::vector<FunctionalDependency> MockNode::functional_dependencies() const {
+  Assert(_table_key_constraints.empty() || !_functional_dependencies.empty(),
```
mrks, line 111: Why is `!_table_key_constraints.empty() && !_functional_dependencies.empty()` a valid case. Shouldn't we avoid having potentially mismatching constraints and dependencies?

julianmenzler: I thought of two cases:

* Whenever FDs are set manually (which is only possible for MockNodes), one seems to be aware of the different logic to StoredTableNodes.

* No key constraints being set means no resulting FDs in StoredTableNodes. Same for MockNode.

julianmenzler: `!_table_key_constraints.empty() && !_functional_dependencies.empty()` being true means, the FDs have been set manually before.

PR #2152, opened 2020-06-08
Rename column_expressions to output_expressions

### LINKED ISSUE #1893: Refactor LQPColumnExpression

mrks:
Renames AbstractLQPNode::column_expressions to ::output_expressions

The former sounds too close to LQPColumnExpression. Also, the "column" part is misleading / redundant.

fixes #1893.


julianmenzler:
Good change!

PR #2153, opened 2020-06-08
Removal of random values in tests

### LINKED ISSUE #1321: Don't use random values in tests

mweisgut:
fixes #1321:
- removal of random value generation in tests
- extension of the lint script so that using random values in tests will be reported as errors (taken from @mrks)
- renaming (leading underscore for private/protected variables/functions)
- formatting

### DIFF
--- a/src/test/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index_test.cpp
+++ b/src/test/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index_test.cpp
@@ -53,38 +53,16 @@ class AdaptiveRadixTreeIndexTest : public BaseTest {
 
     std::set<std::optional<int32_t>> distinct_values(values.begin(), values.end());
 
-    const auto delta = 1332;
-    const auto int32_max = std::numeric_limits<int32_t>::max();
-    auto additional_values = std::array<int32_t, 2>{0, int32_max};
-    auto reset_counts = std::array<int32_t, 2>{0, 0};
+    // 0, 1, the maximum int32_t value and 20 randomly generated positive int32_t values
+    const auto additional_value_pool = std::array<int32_t, 23>{
+        0,          1,          2147483647, 359982793,  2079147571, 386510393,  1762525179, 664094870,
+        1937011651, 784616030,  1000301793, 1891243447, 1654409109, 1476995155, 1110747240, 1726885375,
+        400463838,  2100392285, 308227485,  616812231,  997147072,  2121290667, 986682158};
 
-    // Create search values from given values + additional ones. Add alternately the first value (additional_values[0])
-    // and the second value (additional_values[1]). The first value increases, the second value decreases over time.
     std::set<std::optional<int32_t>> search_values = distinct_values;
-    while (search_values.size() < distinct_values.size() * 2) {
-      const auto value_index = search_values.size() % 2;
-      auto& value = additional_values[value_index];
-      search_values.insert(value);
-
-      // update recently used value
-      auto& reset_count = reset_counts[value_index];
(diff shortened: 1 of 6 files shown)


src/test/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index_test.cpp
```diff
@@ -44,28 +44,29 @@ class AdaptiveRadixTreeIndexTest : public BaseTest {
...
         std::make_shared<AdaptiveRadixTreeIndex>(std::vector<std::shared_ptr<const AbstractSegment>>({segment}));
 
     std::set<std::optional<int32_t>> distinct_values(values.begin(), values.end());
 
-    // create search values from given values + randomly chosen ones
+    // 0, 1, the maximum int32_t value and 20 randomly generated positive int32_t values
+    const auto additional_value_pool = std::array<int32_t, 23>{
+        0,          1,          2147483647, 359982793,  2079147571, 386510393,  1762525179, 664094870,
+        1937011651, 784616030,  1000301793, 1891243447, 1654409109, 1476995155, 1110747240, 1726885375,
+        400463838,  2100392285, 308227485,  616812231,  997147072,  2121290667, 986682158};
```
mrks, line 60: Good solution. Why only positive values? I feel like negative values are underrepresented in our tests. Same for the EncodedSegmentTest.

src/test/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index_test.cpp
```diff
@@ -219,24 +234,65 @@ TEST_F(AdaptiveRadixTreeIndexTest, SimpleTest) {
...
+  const auto cluster_number = 10u;  // 10 clusters with 9 gaps distributed over the range of [0, int32_max]
+  const auto cluster_size = test_size / cluster_number;
+  const auto gap_size = (std::numeric_limits<int32_t>::max() - cluster_number * cluster_size) / (cluster_number - 1);
+
+  std::vector<std::optional<int32_t>> values{};
+  values.reserve(test_size);
+
+  std::array<std::array<int32_t, cluster_size>, cluster_number> clustered_values{};
+
+  // Create clusters of values within the range [0, int32_t::max]. Each cluster has a start value and subsequent values
```
mrks, line 250: Could you describe the rationale for creating these clusters?

src/test/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index_test.cpp
mweisgut: I used the interval that was used before. But of course, we can extend it.

src/test/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index_test.cpp
mweisgut: Creating these clusters is an approach to meet the requirement of large gaps between values in particular. But since std::uniform_int_distribution was used previously, we can also use the value pool approach that is used above. That would be less complex.

mrks: Not saying this needs to change in any way. Just add the comment in the code instead of here.

PR #2157, opened 2020-06-15
Meta tables: Show allocated memory, fix line endings

mrks:
This returns the actually allocated memory. It differs from _get_process_memory_usage in that it only returns memory that has been allocated. The reported virtual memory consumption is usually higher than the amount of allocated memory as free'd memory is not immediately returned to the system (either due to internal page fragmentation or because jemalloc keeps empty pages for future use). In rare cases, it can also be higher than the amount of virtual memory if memory has been allocated but not committed yet.

Here is an example where the lineitem table is changed from unencoded to dictionary encoding after 15 seconds. You can see (1) the internal fragmentation and (2) the delay with which the RSS is reduced.

q6

Remember to hide whitespace changes when reviewing.


scripts/lint.sh
```diff
@@ -85,4 +85,14 @@ for probe in $(grep -r --include=*.[ch]pp --exclude=probes.hpp --exclude=provide
     fi
 done
 
+# Check for Windows-style line endings
+output=$(find src -type f -exec dos2unix -ic {} +)
```
Bensk1, line 89: How do Windows-style line endings get into our code?

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
```diff
@@ -1,268 +1,320 @@
...
+    auto error_code = mallctl("config.stats", &stats_enabled, &stats_enabled_size, nullptr, 0);
+    Assert(!error_code, "Cannot check if jemalloc was built with --stats_enabled");
+    Assert(stats_enabled, "Hyrise's jemalloc was not build with --stats_enabled");
+  }
+
+  // Before retrieving the statistics, we need to update jemalloc's epoch to get current values. See the mallctl
+  // documentation for details.
+  {
+    static uint64_t epoch = 1;
+    epoch++;
```
Bensk1, line 281: To me, the documentation sounds like it is incrementing the epoch itself: 

> "epoch" (uint64_t) rw
If a value is passed in, refresh the data from which the mallctl* functions report values, and increment the epoch.

scripts/lint.sh
mrks: @j-tr?

mrks: Probably some default in the editor or alike

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
mrks: Mmh, what is the purpose of the value being passed in then?

Bensk1: I don't know if it has to be passed in or should be `nullptr` in that case.

Bensk1: Also, I have never seen these functions before and was only taking a short look at the documentation.

scripts/lint.sh
j-tr: guilty

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
j-tr: i think `epoch++` isn't needed. 
see how jemalloc uses it for it's own stats printing:
[https://github.com/jemalloc/jemalloc/blob/ea6b3e973b477b8061e0076bb257dbd7f3faa756/src/stats.c#L1412](https://github.com/jemalloc/jemalloc/blob/ea6b3e973b477b8061e0076bb257dbd7f3faa756/src/stats.c#L1412)

mrks:
We now have to wait for jemalloc to be build before we can build Hyrise. This makes the build process slightly longer in a clean build folder, but since jemalloc does not regularly change, I'd consider it acceptable.

src/lib/utils/meta_tables/meta_system_utilization_table.cpp
mrks: Thanks for looking that up!

mrks:
Still fighting with different behavior on Linux and Mac.

mrks:
The new way of linking jemalloc slightly hurts our throughput, but I'd consider this to be one of those freak fluctuations.

```diff
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs  | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
 | TPC-H 01       | 1.2366394996643066 | 75    | 1.2730578184127808 | 77    | +3%        |  0.0000 |
 | TPC-H 02       | 206.37286376953125 | 12383 | 203.45697021484375 | 12208 | -1%        |  0.0885 |
 | TPC-H 03       | 11.922367095947266 | 716   | 11.564188003540039 | 694   | -3%        |  0.0000 |
 | TPC-H 04       | 14.69411563873291  | 882   | 14.621509552001953 | 878   | -0%        |  0.0000 |
 | TPC-H 05       | 4.3962578773498535 | 264   | 4.562040328979492  | 274   | +4%        |  0.0000 |
 | TPC-H 06       | 273.3237609863281  | 16400 | 271.3006286621094  | 16279 | -1%        |  0.0000 |
 | TPC-H 07       | 16.255126953125    | 976   | 16.19300079345703  | 972   | -0%        |  0.7215 |
 | TPC-H 08       | 13.099245071411133 | 786   | 12.700176239013672 | 763   | -3%        |  0.0000 |
 | TPC-H 09       | 1.9061306715011597 | 115   | 1.9030835628509521 | 115   | -0%        |  0.6921 |
-| TPC-H 10       | 3.8660147190093994 | 232   | 3.595914363861084  | 216   | -7%        |  0.0000 |
 | TPC-H 11       | 115.55313110351562 | 6934  | 115.31096649169922 | 6919  | -0%        |  0.0554 |
 | TPC-H 12       | 22.16092300415039  | 1330  | 21.701824188232422 | 1303  | -2%        |  0.0000 |
 | TPC-H 13       | 2.7387921810150146 | 165   | 2.7539327144622803 | 166   | +1%        |  0.0001 |
-| TPC-H 14       | 46.59079360961914  | 2796  | 36.87356185913086  | 2213  | -21%       |  0.0000 |
 | TPC-H 15       | 123.82136535644531 | 7430  | 123.33146667480469 | 7400  | -0%        |  0.0000 |
-| TPC-H 16       | 10.900328636169434 | 655   | 9.980742454528809  | 599   | -8%        |  0.0000 |
 | TPC-H 17       | 45.92328643798828  | 2756  | 43.923954010009766 | 2636  | -4%        |  0.0000 |
 | TPC-H 18       | 1.3331670761108398 | 80    | 1.3072922229766846 | 79    | -2%        |  0.0029 |
 | TPC-H 19       | 29.34632682800293  | 1761  | 28.1800537109375   | 1691  | -4%        |  0.0000 |
 | TPC-H 20       | 55.896934509277344 | 3354  | 55.275108337402344 | 3317  | -1%        |  0.0001 |
 | TPC-H 21       | 2.360239267349243  | 142   | 2.348756790161133  | 141   | -0%        |  0.2828 |
-| TPC-H 22       | 19.004470825195312 | 1141  | 18.028423309326172 | 1082  | -5%        |  0.0000 |
 | geometric mean |                    |       |                    |       | -3%        |         |
 +----------------+--------------------+-------+--------------------+-------+------------+---------+
```

I confirmed that the performance drops more significantly if jemalloc is not used.

Bensk1:
Have we seen freak fluctuations of -21% yet?

mrks:
Yes. :/

Bouncner:
Maybe add a clang and MT run to check if it’s really a fluctuation?

mrks:
MT would be expected to show the same fluctuation. I'll run it with MT.

mrks:
Turns out it was an actual issue with the way we replaced the default allocator with jemalloc. 

On Mac, where jemalloc is now removed:

```diff
 +----------------+--------------------+-------+--------------------+------+------------+---------+
 | Benchmark      | prev. iter/s       | runs  | new iter/s         | runs | change [%] | p-value |
 +----------------+--------------------+-------+--------------------+------+------------+---------+
-| TPC-H 01       | 1.6123777627944946 | 97    | 1.4390076398849487 | 87   | -11%       |  0.0000 |
-| TPC-H 02       | 166.97422790527344 | 10019 | 103.29190826416016 | 6198 | -38%       |  0.0000 |
-| TPC-H 03       | 10.019548416137695 | 602   | 6.659781455993652  | 400  | -34%       |  0.0000 |
-| TPC-H 04       | 12.108245849609375 | 727   | 10.488577842712402 | 630  | -13%       |  0.0000 |
-| TPC-H 05       | 3.5893115997314453 | 216   | 2.7103049755096436 | 163  | -24%       |  0.0000 |
-| TPC-H 06       | 132.93008422851562 | 7976  | 116.22422790527344 | 6974 | -13%       |  0.0000 |
-| TPC-H 07       | 11.877799034118652 | 713   | 8.75653076171875   | 526  | -26%       |  0.0000 |
-| TPC-H 08       | 9.970952987670898  | 599   | 8.104108810424805  | 487  | -19%       |  0.0000 |
-| TPC-H 09       | 1.5544050931930542 | 94    | 1.341689944267273  | 81   | -14%       |  0.0000 |
-| TPC-H 10       | 3.336468458175659  | 201   | 2.4665026664733887 | 148  | -26%       |  0.0000 |
-| TPC-H 11       | 78.61650848388672  | 4717  | 60.04450225830078  | 3603 | -24%       |  0.0000 |
 | TPC-H 12       | 17.68667221069336  | 1062  | 17.1570987701416   | 1030 | -3%        |  0.0000 |
-| TPC-H 13       | 1.7374308109283447 | 105   | 1.3590434789657593 | 82   | -22%       |  0.0000 |
-| TPC-H 14       | 35.374732971191406 | 2123  | 27.128955841064453 | 1628 | -23%       |  0.0000 |
-| TPC-H 15       | 84.89486694335938  | 5094  | 69.80027770996094  | 4189 | -18%       |  0.0000 |
-| TPC-H 16       | 9.79648494720459   | 588   | 6.737063884735107  | 405  | -31%       |  0.0000 |
 | TPC-H 17       | 28.8576602935791   | 1732  | 28.517093658447266 | 1712 | -1%        |  0.0067 |
+| TPC-H 18       | 0.8431762456893921 | 51    | 0.9561082124710083 | 58   | +13%       |  0.0000 |
-| TPC-H 19       | 20.68048667907715  | 1241  | 18.88153648376465  | 1134 | -9%        |  0.0000 |
-| TPC-H 20       | 36.05957794189453  | 2164  | 33.07355499267578  | 1985 | -8%        |  0.0000 |
-| TPC-H 21       | 1.8845421075820923 | 114   | 1.6086194515228271 | 97   | -15%       |  0.0000 |
-| TPC-H 22       | 13.647863388061523 | 819   | 10.946023941040039 | 657  | -20%       |  0.0000 |
-| geometric mean |                    |       |                    |      | -18%       |         |
 +----------------+--------------------+-------+--------------------+------+------------+---------+
```

On nemea with clang (20s only):
```diff
 +----------------+--------------------+------+--------------------+------+------------+----------------------+
 | Benchmark      | prev. iter/s       | runs | new iter/s         | runs | change [%] |              p-value |
 +----------------+--------------------+------+--------------------+------+------------+----------------------+
 | TPC-H 01       | 1.6073918342590332 | 33   | 1.6075549125671387 | 33   | +0%        | (run time too short) |
 | TPC-H 02       | 122.08198547363281 | 2442 | 123.39434051513672 | 2468 | +1%        | (run time too short) |
 | TPC-H 03       | 10.541433334350586 | 211  | 10.680342674255371 | 214  | +1%        | (run time too short) |
 | TPC-H 04       | 13.270790100097656 | 266  | 13.143855094909668 | 263  | -1%        | (run time too short) |
 | TPC-H 05       | 4.063416004180908  | 82   | 4.043587684631348  | 81   | -0%        | (run time too short) |
 | TPC-H 06       | 264.9241027832031  | 5299 | 263.4817810058594  | 5270 | -1%        | (run time too short) |
 | TPC-H 07       | 13.725506782531738 | 275  | 13.794315338134766 | 276  | +1%        | (run time too short) |
 | TPC-H 08       | 12.477733612060547 | 250  | 12.554290771484375 | 252  | +1%        | (run time too short) |
 | TPC-H 09       | 1.9473185539245605 | 39   | 1.9574837684631348 | 40   | +1%        | (run time too short) |
 | TPC-H 10       | 3.934159517288208  | 79   | 3.9346039295196533 | 79   | +0%        | (run time too short) |
 | TPC-H 11       | 115.48187255859375 | 2310 | 116.7041244506836  | 2335 | +1%        | (run time too short) |
 | TPC-H 12       | 20.407533645629883 | 409  | 21.061622619628906 | 422  | +3%        | (run time too short) |
 | TPC-H 13       | 2.9104177951812744 | 59   | 2.9395275115966797 | 59   | +1%        | (run time too short) |
 | TPC-H 14       | 46.214134216308594 | 925  | 46.83518981933594  | 937  | +1%        | (run time too short) |
 | TPC-H 15       | 114.81187438964844 | 2297 | 115.65462493896484 | 2314 | +1%        | (run time too short) |
 | TPC-H 16       | 11.244185447692871 | 225  | 11.286304473876953 | 226  | +0%        | (run time too short) |
 | TPC-H 17       | 41.86561584472656  | 838  | 42.819984436035156 | 857  | +2%        | (run time too short) |
 | TPC-H 18       | 1.4934589862823486 | 30   | 1.482218861579895  | 30   | -1%        | (run time too short) |
 | TPC-H 19       | 28.3396053314209   | 567  | 29.01471519470215  | 581  | +2%        | (run time too short) |
 | TPC-H 20       | 54.691741943359375 | 1094 | 54.896968841552734 | 1098 | +0%        | (run time too short) |
 | TPC-H 21       | 2.2318966388702393 | 45   | 2.1762843132019043 | 44   | -2%        | (run time too short) |
 | TPC-H 22       | 17.812557220458984 | 357  | 17.911623001098633 | 359  | +1%        | (run time too short) |
 | geometric mean |                    |      |                    |      | +1%        |                      |
 +----------------+--------------------+------+--------------------+------+------------+----------------------+
```

Bouncner:
I am not familiar at all with the whole topic of allocators. So I would trust your decision here.

From a plugin writer perspective, I am see plugins as something outside of Hyrise, which tries to come up with some good tweaks but does itself not allocate vast amounts of memory. But that might be just true for me.

Is there a difference with allocating between the "external" plugins and the ones that are part of Hyrise (e.g., MVCC)?

mrks:
> From a plugin writer perspective, I am see plugins as something outside of Hyrise, which tries to come up with some good tweaks but does itself not allocate vast amounts of memory. But that might be just true for me.

If they decide, e.g., on a different encoding, they might allocate significant amounts of memory. I am not 100% sure where that memory would come from on Mac.

> Is there a difference with allocating between the "external" plugins and the ones that are part of Hyrise (e.g., MVCC)?

No.

mrks:
@Bensk1

Bensk1:
> If they decide, e.g., on a different encoding, they might allocate significant amounts of memory. I am not 100% sure where that memory would come from on Mac.

@Bouncner, that should be the case for the BP's compression plugin, shouldn't it?


> Personally, I care about consistent allocations more than about the performance on Mac, but this is something that should be discussed.

I agree, we are not going to make any important benchmarks on Macs anyway. It would be nice to have competitive/comparable performance on macOS but I don't know if its worth more effort.

> Here's the Mac problem described: https://github.com/exabytes18/jemalloc-example

Could the tcmalloc zone discussion help us with the recently discovered plugin problem where accesses to memory that was allocated by plugins lead to errors after plugins were unloaded?

Bouncner:
> @Bouncner, that should be the case for the BP's compression plugin, shouldn't it?

I'd thought that calling internals of Hyrise such as the `ChunkEncoder` should not cause such allocations problems. But if they apparently do, yes that should be a problem.

mrks:
> Could the tcmalloc zone discussion help us with the recently discovered plugin problem where accesses to memory that was allocated by plugins lead to errors after plugins were unloaded?

Is that reproducable? If yes, we should test it on this branch.

mrks:
Results from Fortnightly:

- [ ] Add a notice somewhere that Mac is supported for development only

mrks:
> ## Supported Systems
> Hyrise is developed for Linux (preferrably the most current Ubuntu version) and optimized to run on server hardware. We support Mac to facilitate the local development of Hyrise, but do not recommend it for benchmarking.

@Bensk1?

Bensk1:
Works for me.

PR #2161, opened 2020-06-16
Fix storage manager accessors

Bensk1:


src/lib/storage/storage_manager.hpp
```diff
@@ -31,7 +31,7 @@ class StorageManager : public Noncopyable {
   std::shared_ptr<Table> get_table(const std::string& name) const;
   bool has_table(const std::string& name) const;
   std::vector<std::string> table_names() const;
-  const tbb::concurrent_unordered_map<std::string, std::shared_ptr<Table>>& tables() const;
+  std::unordered_map<std::string, std::shared_ptr<Table>> tables() const;
```
mrks, line 34: Can/should we return a `const shared_ptr` so that people don't think they are actually modifying the list of tables?

Bensk1: Tidy was complaining about the const here. Since this is not a reference, how should they think they could modify it?

mrks: If you read the prototype properly, you shouldn't. Did tidy complain about the const inside or about a const on the unordered_map?

Bensk1: I didn't understand it either:

```
/var/lib/jenkins/workspace/hyrise_hyrise_PR-2161/src/lib/storage/storage_manager.cpp:81:1: error: return type 'const std::unordered_map<std::string, std::shared_ptr<Table> >' (aka 'const unordered_map<basic_string<char>, shared_ptr<opossum::Table> >') is 'const'-qualified at the top level, which may reduce code readability without improving const correctness [readability-const-return-type,-warnings-as-errors]
const std::unordered_map<std::string, std::shared_ptr<Table>> StorageManager::tables() const {
```

mrks: Try to move the const in front of the shared_ptr please

Bensk1: That does not directly work because the shared_ptr in the actual map is also not const. Making that shared_ptr const as well leads to errors in other places. I need to look at that properly.

mrks: Thinking about it, I'm not 100% sure if unordered_map supports const values at all.

mrks:
https://github.com/hyrise/hyrise/blob/08ee7eb9e2da11202573210d0c2cce29dbd33231/src/plugins/mvcc_delete_plugin.cpp#L37

https://en.cppreference.com/w/cpp/language/range-for#Temporary_range_expression

http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_closed.html#900

I didn't want to stare into C++'s abyss that early in the morning. Good chance we are making that mistake in other places, too.

Bensk1:
Great thanks. It was not only that, but also (mostly?) that the plugin's queue stored references.

Bensk1:
> I didn't want to stare into C++'s abyss that early in the morning. Good chance we are making that mistake in other places, too.

Is there any way to automatically detect cases where we use range-based for loops on functions that return by value?

Good that we have tools like asan, we would never have found that one.

mrks:
No, I would have hoped for some type of SCA to catch that :/

Bensk1:
```
==236962==LeakSanitizer has encountered a fatal error.
==236962==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==236962==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
script returned exit code 1
```

I cannot reproduce that.

mrks:
Any chance this ran on rapa? I think we are seeing this from time to time when running on rapa. Haven't found out why yet. :/

Bensk1:
No, this is from the CI log.

mrks:
I mean if jenkins used the rapa SSH slave

Bensk1:
Yes, rapa is used all the time.

Jenkinsfile
```diff
@@ -52,7 +52,7 @@ try {
     def oppossumCI = docker.image('hyrise/opossum-ci:20.04');
     oppossumCI.pull()
 
-    oppossumCI.inside("-u 0:0") {
+    oppossumCI.inside("--cap-add SYS_PTRACE -u 0:0") {
```
mrks, line 55: Doc the PTRACE thing? Otherwise someone will find that CI seems to work without it...

Jenkinsfile
```diff
@@ -45,10 +45,14 @@ try {
   }
 
   node('linux') {
+    stage("hostname") {
+      sh "hostname"
```
mrks, line 49: "Print the hostname so that we can reproduce what node the docker image was executed on"

mrks:
Took the liberty of slightly adding to it, instead of asking for a change. Hope that's ok.

PR #2163, opened 2020-06-17
Fix `show tables` and subqueries involving meta tables (#2162)

### LINKED ISSUE #2162: Some queries involving Metatables give wrong results
To reproduce in hyriseConsole:

```
SHOW tables
generate_tpch 0.01
SHOW tables
```

The second `SHOW tables` should display the tables of the TPC-H dataset. Instead, it displays an empty table because the plan (metatable) of the initial `SHOW tables` has been cached.

The same applies if `SHOW tables` is replaced with `SELECT table_name FROM (SELECT table_name, column_count FROM meta_tables) as subquery`.

Bensk1:
Fixes #2162.

There are two things I don't really like, they are up for discussion:

- ~~Solving the subquery involving non-cacheable metatables problem by making `SQLTranslator's` `_cacheable` a public member~~
- ~~The way this is tested: adding an optional parameter to `SQLTranslatorTest's` `compile_query` method. We could also change the function's return type but then we would need to adapt almost all 147 SQLTranslatorTests.~~

### DIFF
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -640,6 +640,9 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_origin(const hsq
                                         _with_descriptions, _meta_tables};
       lqp = subquery_translator._translate_select_statement(*hsql_table_ref.select);
 
+      // If a statement of the subquery is not cacheable (e.g., for meta tables), this statement should not be cacheable either.
+      cacheable = subquery_translator.cacheable;
+
       std::vector<std::vector<SQLIdentifier>> identifiers;
       for (const auto& element : subquery_translator._inflated_select_list_elements) {
         identifiers.emplace_back(element.identifiers);
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -142,7 +142,7 @@ SQLTranslator::SQLTranslator(const UseMvcc use_mvcc)
                     std::make_shared<std::unordered_map<std::string, std::shared_ptr<Table>>>()) {}
 
 SQLTranslationResult SQLTranslator::translate_parser_result(const hsql::SQLParserResult& result) {
-  cacheable = true;
+  _cacheable = true;
 
   std::vector<std::shared_ptr<AbstractLQPNode>> result_nodes;
   const std::vector<hsql::SQLStatement*>& statements = result.getStatements();
@@ -159,7 +159,7 @@ SQLTranslationResult SQLTranslator::translate_parser_result(const hsql::SQLParse
     parameter_ids[value_placeholder_id] = parameter_id;
   }
 
-  return {result_nodes, {cacheable, parameter_ids}};
+  return {result_nodes, {_cacheable, parameter_ids}};
 }
 
 SQLTranslator::SQLTranslator(
@@ -740,7 +740,7 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_meta_table(
 
   // MetaTables are non-cacheable because they might contain information about the general system state
   // that can change at any time
-  cacheable = false;
+  _cacheable = false;
 
   const auto meta_table_name = _trim_meta_table_name(name);
 
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -1227,6 +1227,8 @@ void SQLTranslator::_translate_limit(const hsql::LimitDescription& limit) {
 
 // NOLINTNEXTLINE - while this particular method could be made static, others cannot.
 std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_show(const hsql::ShowStatement& show_statement) {
+  cacheable = false;
+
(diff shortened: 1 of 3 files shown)


src/lib/sql/sql_translator.hpp
```diff
@@ -203,12 +203,13 @@ class SQLTranslator final {
 
   static std::string _trim_meta_table_name(const std::string& name);
 
+  // The current LQP might not be cacheable if it involves a meta table
+  bool cacheable{true};
```
Bensk1, line 207: See PR discussion.

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -79,15 +79,22 @@ class SQLTranslatorTest : public BaseTest {
     int_int_int_c = stored_table_node_int_int_int->get_column("c");
   }
 
-  std::shared_ptr<opossum::AbstractLQPNode> compile_query(const std::string& query,
-                                                          const UseMvcc use_mvcc = UseMvcc::No) {
+  std::shared_ptr<opossum::AbstractLQPNode> compile_query(
+      const std::string& query, const UseMvcc use_mvcc = UseMvcc::No,
+      const std::shared_ptr<SQLTranslationInfo>& translation_info = nullptr) {
```
Bensk1, line 84: See PR discussion.

src/lib/sql/sql_translator.cpp
```diff
@@ -640,6 +640,10 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_origin(const hsq
                                         _with_descriptions, _meta_tables};
       lqp = subquery_translator._translate_select_statement(*hsql_table_ref.select);
 
+      // If a statement of the subquery is not cacheable (e.g., for meta tables),
+      // this statement should not be cacheable either.
+      cacheable = subquery_translator.cacheable;
```
mrks, line 645: Shouldn't it be `|=`?

mrks: If yes, there is probably a test missing.

src/lib/sql/sql_translator.hpp
mrks: > Solving the subquery involving non-cacheable metatables problem by making SQLTranslator's _cacheable a public member

Why does it have to be public? Where are you accessing it from the outside?

src/test/lib/sql/sql_translator_test.cpp
mrks: I don't like `compile_query` too much. Maybe `sql_to_lqp_helper`?

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -79,15 +79,22 @@ class SQLTranslatorTest : public BaseTest {
...
 
     const auto translation_result = SQLTranslator{use_mvcc}.translate_parser_result(parser_result);
     const auto lqps = translation_result.lqp_nodes;
 
+    if (translation_info) {
+      translation_info->cacheable = translation_result.translation_info.cacheable;
+      translation_info->parameter_ids_of_value_placeholders =
+          translation_result.translation_info.parameter_ids_of_value_placeholders;
+    }
+
```
mrks, line 97: Any reason you can't return a pair instead of having an output parameter?

src/lib/sql/sql_translator.cpp
Bensk1: Agreed!

src/lib/sql/sql_translator.hpp
Bensk1: The SubqueryTranslator is another SQLTranslator, isn't it?

src/test/lib/sql/sql_translator_test.cpp
Bensk1: Nope. Just that we have to change all tests then.

src/lib/sql/sql_translator.cpp
mrks: Ah wait. It should be `&=`, right?

src/lib/sql/sql_translator.hpp
mrks: So?

> A private member of a class is only accessible to the members and friends of that class, regardless of whether the members are on the same or different instances

Am I missing something?

src/lib/sql/sql_translator.cpp
Bensk1: Right, thank you! Only if outer and inner plan are cacheable, it should be set to `true`.

src/lib/sql/sql_translator.hpp
Bensk1: No, but I did, apparently.

src/test/lib/sql/sql_translator_test.cpp
mrks: Looks like 90% can be covered with a single Regexp

mrks:
As the final commit message includes the PR number, I'd suggest leaving "(#2162)" out of the PR title. For context, people will easily find the "fixes ..." in the text.

PR #2167, opened 2020-06-17
Explicitly install lld-9

Bouncner:
Depending on the linux system used, lld might not be available. Since we are very explicit with `clang-9`, `clang-cov-9`, we should also use `lld-9`. With that, adding the llvm9 repositories is sufficient to get the install script running.


mrks:
Please update the Dockerfile, too.

Bouncner:
Done.

PR #2169, opened 2020-06-18
Add formatting and linting for Python.

Bensk1:
- [x] Maybe document that somewhere
- [x] pexpect is part of the install script and requirements.txt
- [x] Do we want sorted imports? -> Not necessary
- [x] Decide on mentioning python packages in dependencies file
- [x] Change Docker image tag in Jenkinsfile


Bensk1:
`pexpect` is part of the install script and requirements.txt

Bensk1:
Do we need pexpect as part of the [install_dependencies script](https://github.com/hyrise/hyrise/blob/e136084ddd303e41463489bef49dee9d71335d84/install_dependencies.sh#L66) and in the [requirements.txt](https://github.com/hyrise/hyrise/blob/e136084ddd303e41463489bef49dee9d71335d84/requirements.txt#L4)?

README.md
```diff
@@ -72,10 +72,10 @@ Usually debug binaries are created.
 To configure a build directory for a release build make sure it is empty and call CMake like `cmake -DCMAKE_BUILD_TYPE=Release`
 
 ### Lint
-`./scripts/lint.sh` (Google's cpplint is used which needs python 2.7)
```
Bensk1, line 75: I believe this is not true anymore.

scripts/benchmark_multithreaded.py
```diff
@@ -1,18 +1,17 @@
...
-import matplotlib.pyplot as plt
 
-import os
 import argparse
+from datetime import datetime
+import json
+import matplotlib
 import multiprocessing
+import os
 import subprocess
```
mrks, line 10: > Do we want sorted imports?

I don't care. If you ask me, this whole `from ... import ... as ...` makes for a blood mess anyway.

scripts/benchmark_multithreaded.py
```diff
@@ -69,98 +130,109 @@ def get_core_counts(args):
...
 
     results = {}
     for _, _, files in os.walk(result_dir):
-        json_files = [f for f in files if f.split('.')[-1] == 'json']
-        num_cores_from_filename = lambda filename: int(filename.split('-')[0])
+        json_files = [f for f in files if f.split(".")[-1] == "json"]
         # Add the results in sorted order (low core count -> high core count) for plotting later
-        for file in sorted(json_files, key=num_cores_from_filename):
-            with open(os.path.join(result_dir, file), 'r') as json_file:
+        for file in sorted(json_files, key=lambda filename: int(filename.split("-")[0])):
```
mrks, line 219: Was this something that the linter made you do?

scripts/compare_benchmarks_multithreaded.py
```diff
@@ -93,74 +103,77 @@ def plot_scaleup(flipped_results, n_rows_cols, numa_borders, max_cores):
...
         one_result = results[label]
 
         for _, _, files in os.walk(result_dir):
-            json_files = [f for f in files if f.split('.')[-1] == 'json']
-            num_cores_from_filename = lambda filename: int(filename.split('-')[0])
+            json_files = [f for f in files if f.split(".")[-1] == "json"]
             # Add the results in sorted order (low core count -> high core count) for plotting later
-            for file in sorted(json_files, key=num_cores_from_filename):
-                with open(os.path.join(result_dir, file), 'r') as json_file:
+            for file in sorted(json_files, key=lambda filename: int(filename.split("-")[0])):
```
mrks, line 152: see above

scripts/plot_operator_breakdown.py
```diff
@@ -15,48 +15,48 @@
 import matplotlib.colors as mplcolors
 from matplotlib.colors import LinearSegmentedColormap
 
-if(not len(sys.argv) in [1,2]):
+if not len(sys.argv) in [1, 2]:
```
mrks, line 18: I saw changes from `not x in y` to `x not in y` above. Were those manual? If not, why didn't the linter complain here?

scripts/plot_performance_over_time.py
```diff
@@ -28,40 +28,43 @@ def lighten_color(input_color, amount=0.5):
     """
     try:
         c = colors.cnames[input_color]
-    except:
+    except Exception:
```
mrks, line 31: What's the reasoning here? (Also in other files)

scripts/plot_performance_over_time.py
```diff
@@ -88,25 +91,34 @@ def lighten_color(input_color, amount=0.5):
...
-                data.plot(ax=ax, kind='scatter', x='begin', y='duration', c=single_run_color,
-                          figsize=(figw, figh), s=5, marker=('o' if success else 'x'), linewidth=1)
+                data.plot(
+                    ax=ax,
+                    kind="scatter",
+                    x="begin",
+                    y="duration",
+                    c=single_run_color,
+                    figsize=(figw, figh),
+                    s=5,
```
mrks, line 107: replace with long form? I can't remember what s stands for

Dockerfile
```diff
@@ -45,4 +46,6 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
     && ln -sf /usr/bin/llvm-symbolizer-3.8 /usr/bin/llvm-symbolizer
 
+RUN pip3 install flake8
```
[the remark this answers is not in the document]
Bensk1: But then it is only available in the Docker container.

scripts/benchmark_multithreaded.py
Bensk1: Yes, lambdas should not be assigned to variables. If we want to give it a name to make it more descriptive it should be a function.

scripts/plot_performance_over_time.py
Bensk1: The linter does not like `except` without specifying any Exception type, but yes, `Exception` is not much better. I think this particular case could be solved without exceptions. The other ones are more painful.

Dockerfile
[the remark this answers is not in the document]
mrks: This is all about the Dockerfile?!

scripts/benchmark_multithreaded.py
mrks: I'd do that or add a comment containing the filename magic. Right now, it is unclear what we are doing here.

Dockerfile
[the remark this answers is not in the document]
Bensk1: No, we have pexpect in the requirements.txt and the install script.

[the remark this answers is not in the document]
mrks: python3-pexpect is installed using aptitude above. Shouldn't we install pexpect and flake similarly?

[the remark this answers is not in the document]
Bensk1: Yes. Then, there are two pexpect installations remaining: I would probably remove it from install_dependencies.sh and keep it in requirements.txt to have all python packages together.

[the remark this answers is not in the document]
mrks: In that case, make sure that requirements.txt is used as part of the docker setup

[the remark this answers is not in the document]
Bensk1: Let's discuss that in person.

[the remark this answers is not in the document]
mrks: Result: Make dependencies.md part of install_dependencies.sh. Change pexpect from apt to pip.

Bensk1:
Also, pexpect is the only python package that is listed in DEPENDENCIES.md. For me, this is fine because it is the only package that is necessary for CI/running all tests. Any other opinions?

Jenkinsfile
```diff
@@ -50,7 +50,7 @@ try {
       sh "hostname"
     }
 
-    def oppossumCI = docker.image('hyrise/opossum-ci:20.04');
+    def oppossumCI = docker.image('hyrise/opossum-ci:20.04_2');
```
Bensk1, line 53: TODO: change back as soon as everything else is done here.

install_dependencies.sh
```diff
@@ -63,13 +68,18 @@ if echo $REPLY | grep -E '^[Yy]$' > /dev/null; then
...
                 if ! git submodule update --jobs 5 --init --recursive; then
                     echo "Error during installation."
                     exit 1
                 fi
 
+                if ! pip3 install -r requirements.txt; then
+                    echo "Error during installation of python requirements."
+                    exit 1
+                fi
+
```
Bensk1, line 82: Hm, I don't like that this adds a couple of minutes to the CI's setup step. Pandas and numpy take quite some time to build.

scripts/plot_operator_breakdown.py
Bensk1: According to that [rule](https://www.flake8rules.com/rules/E713.html), all occurrences should have been changed. Apparently, it does not recognize that case, maybe because of the call to `len`. I will replace it manually.

scripts/plot_performance_over_time.py
Bensk1: Apparently, it's the marker size. I couldn't find a proper long version for scatter plots and added a comment.

install_dependencies.sh
Bensk1: How about adding the requirements.txt during the docker build process to the container, installing the requirements in the Dockerfile, and only install them with install_dependencies if in non-`OPOSSUM_HEADLESS_SETUP` mode?

Bouncner: If there are components build from source, we should update them. Had the same issues with my compare benchmark script, take a look there.
Precompiled wheels are not available for too old versions.

mrks: Hmm... The problem with adding the requirements.txt is that we would add a dependency on the Hyrise git to the Docker image. In that case, we could forgo most of the Dockerfile and call install_dependencies right away.

Alternatively, we could only add pexpect to the Dockerfile and skip the requirements.txt using OPOSSUM_HEADLESS_SETUP.

Bensk1: Martin's proposal works.

Bensk1:
> Also, pexpect is the only python package that is listed in DEPENDENCIES.md. For me, this is fine because it is the only package that is necessary for CI/running all tests. Any other opinions?

@mrks?

Dockerfile
```diff
@@ -29,13 +29,13 @@ RUN apt-get update \
         libreadline-dev \
         libsqlite3-dev \
         libtbb-dev \
-        lld-9 \
+        lld \
```
mrks, line 32: Was this reverted on purpose?

install_dependencies.sh
mrks: Could you explain the current solution please? Doesn't this mean that we install the packages every time?

Dockerfile
mrks, line 32: Did you already push a new Docker image?

Bensk1: Yes

Bensk1: Something must have gone wrong during merge. I will change it back.

install_dependencies.sh
Bensk1: We install the python packages every time but do not build them from source. If we don't want to install them every time, we could do it in the docker file but then 

> The problem with adding the requirements.txt is that we would add a dependency on the Hyrise git to the Docker image

mrks: Got it. Doesn't seem to add too much to the CI runtime anyway.

Bensk1:
This does not build with lld-9: https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-2169/18/pipeline/46

I was surprised that it does not work because @Bouncner's initial PR was building without errors. But may that was only the case because no new Dockerfile was built and pushed?

CMakeLists.txt
```diff
@@ -94,8 +94,8 @@ endif()
 
 # On Linux, use the lld linker, which is twice as fast for incremental debug builds
 if (NOT APPLE)
-    set(CMAKE_SHARED_LINKER_FLAGS "-fuse-ld=lld")
-    set(CMAKE_EXE_LINKER_FLAGS "-fuse-ld=lld")
+    set(CMAKE_SHARED_LINKER_FLAGS "-fuse-ld=lld-9")
+    set(CMAKE_EXE_LINKER_FLAGS "-fuse-ld=lld-9")
```
Bensk1, line 98: This works for clang but not for g++: https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-2169/19/pipeline/48/

@Bouncner did you verify that lld-9 works with g++? Otherwise should we #2167?

Bouncner: Worked on 18.04 for me, but it fixed a clang issue and I relied on the CI. :-/
So no, haven't explicitly checked GCC.  :(

As it is unnecessary work to dive deep into this, just go back to `lld` and I'll take a look at the 18.04 LTS issue another time (still working ARM benchmarks).

Bensk1: @mrks fine with reverting #2167?

mrks: Sure

mrks: One more thing to try is to explicitly install it using update-alternatives. I don't care if we do it here or separately.

Dockerfile
```diff
@@ -1,3 +1,5 @@
+# IMPORTANT: If you change this file in any way, chances are high that the docker
+# image used by our CI server must be rebuilt. See the wiki for further details:
```
mrks, line 2: That's a bit unspecific. Maybe: "Changes in this file do not automatically affect the Docker image used by the CI server. You need to build and push it manually, see the wiki for details:"?

CMakeLists.txt
Bensk1: Separately!

PR #2170, opened 2020-06-18
Benchmark optimizer rules

j-tr:
- Tracks the runtime of each individual optimizer rule as part of the SQLPipelineStatementMetrics.

- Adds average runtime for each query to the performance breakdown plot.

- Adds seperate performance breakdown plot for the most relevant optimizer rules that looks like this:

tpcc_result_optimizer_breakdown


scripts/plot_performance_breakdown.py
```diff
@@ -35,31 +40,66 @@
...
+column_names = ['Benchmark', 'Parser', 'SQLTranslator']
+if options.plot_rules:
+    column_names.extend(sum_optimizer_rule_durations.keys())
+    column_names.append("Other Rules")
+else:
+    column_names.append('Optimizer')
+column_names.extend(['LQPTranslator', 'Execution'])
+df = pd.DataFrame(benchmarks, columns=column_names)
 
+total_time = df.iloc[:,1:].apply(lambda x: x.sum(), axis=1)
```
mrks, line 74: What's happening here?

scripts/plot_performance_breakdown.py
```diff
@@ -35,31 +40,66 @@
...
 print(df)
 
-ax = df.plot.bar(x='Benchmark', stacked=True)
+if options.plot_rules:
+    # aggregate all rule durations below the threshold
+    threshold = 0.05
+    for index, benchmark in df[sum_optimizer_rule_durations.keys()].iterrows():
+        for rule_name, rule_duration in benchmark.iteritems():
+            if rule_duration < threshold:
+                df.loc[index, "Other Rules"] += rule_duration
```
mrks, line 85: Right now, the order of rules seems random. Can we change that and move "other" to the bottom?

src/lib/optimizer/optimizer.cpp
```diff
@@ -172,11 +174,31 @@ std::shared_ptr<AbstractLQPNode> Optimizer::optimize(std::shared_ptr<AbstractLQP
   const auto root_node = LogicalPlanRootNode::make(std::move(input));
   input = nullptr;
 
-  if constexpr (HYRISE_DEBUG) validate_lqp(root_node);
+  if constexpr (HYRISE_DEBUG) {
```
mrks, line 177: Could you explain what you added here?

src/lib/optimizer/optimizer.cpp
```diff
@@ -172,11 +174,31 @@ std::shared_ptr<AbstractLQPNode> Optimizer::optimize(std::shared_ptr<AbstractLQP
...
 
-  for (const auto& rule : _rules) {
-    _apply_rule(*rule, root_node);
-    if constexpr (HYRISE_DEBUG) validate_lqp(root_node);
+    Timer rule_timer{};
+    for (const auto& rule : _rules) {
+      _apply_rule(*rule, root_node);
+      auto rule_duration = rule_timer.lap();
+
+      auto& r = *rule.get();
```
mrks, line 185: Naming

src/lib/optimizer/optimizer.cpp
```diff
@@ -172,11 +174,31 @@ std::shared_ptr<AbstractLQPNode> Optimizer::optimize(std::shared_ptr<AbstractLQP
...
+      auto rule_name = std::string(typeid(r).name());
+      rule_name = rule_name.substr(
+          rule_name.rend() - std::find_if(rule_name.rbegin(), rule_name.rend(), [](auto c) { return std::isdigit(c); }),
+          rule_name.size());
+      rule_name = rule_name.substr(0, rule_name.find("Rule") + 4);
+
+      if (rule_durations) {
+        rule_durations->push_back({rule_name, rule_duration});
```
mrks, line 193: This looks like it's not done for release builds?

src/lib/optimizer/optimizer.hpp
```diff
@@ -10,6 +10,11 @@
 
 namespace opossum {
 
+struct OptimizerRuleMetrics {
+  std::string name;
```
mrks, line 14: Name of what?

src/lib/optimizer/optimizer.hpp
```diff
@@ -10,6 +10,11 @@
...
+  std::chrono::nanoseconds duration;
```
mrks, line 15: Is this overwritten when a rule is executed twice (e.g., PredicateSplitUpRule)?

src/lib/optimizer/optimizer.hpp
```diff
@@ -32,7 +37,9 @@ class Optimizer final {
    */
   void add_rule(std::unique_ptr<AbstractRule> rule);
 
-  std::shared_ptr<AbstractLQPNode> optimize(std::shared_ptr<AbstractLQPNode> input) const;
+  std::shared_ptr<AbstractLQPNode> optimize(
+      std::shared_ptr<AbstractLQPNode> input,
+      std::shared_ptr<std::vector<OptimizerRuleMetrics>> rule_durations = nullptr) const;
```
mrks, line 42: Doc?

mrks, line 42: Why do we need a shared_ptr?

scripts/plot_performance_breakdown.py
```diff
@@ -6,14 +6,18 @@
 import json
 import matplotlib.pyplot as plt
 import pandas as pd
-import sys
+import argparse
```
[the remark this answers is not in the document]
j-tr: see the figure above: the average runtime is printed below the bars for each query. maybe i should add a hint that the time is the average runtime.

scripts/plot_performance_breakdown.py
j-tr: indeed. would you prefer alphabetic ordering? ordering them in the actual execution order might be difficult since some rules are applied more than once.

src/lib/optimizer/optimizer.cpp
j-tr: i thought we don't want the timer and rule name deduction in there for release builds. before, we had the validation steps before and after the rule execution in debug builds. now, it's either clean rule execution or validation, timed rule execution, validation.

src/lib/optimizer/optimizer.cpp
j-tr: indeed. that's why we check for HYRISE_DEBUG earlier

src/lib/optimizer/optimizer.hpp
j-tr: no. a new instance is added to the vector of metrics. later in the python script, multiple runtimes for one rule are added up.

src/lib/optimizer/optimizer.hpp
j-tr: references and default parameters seem not to like each other... what's your suggestion here?

scripts/plot_performance_breakdown.py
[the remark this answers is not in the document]
j-tr: depends on the usecase i guess, but would be no effort to add a --optimizer_only flag.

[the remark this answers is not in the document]
mrks: Oups. Adding a ⌀ wouldn't hurt, but I think it is clear the way it is. I just didn't map the text to it.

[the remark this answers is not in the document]
mrks: What is the advantage of having two graphs or being able to disable a subplot?

scripts/plot_performance_breakdown.py
mrks: Add this as a comment?

src/lib/optimizer/optimizer.cpp
mrks: I'd say that release builds are much more interesting. Maybe you could show a comparison of the breakdown for debug and release builds? I'd expect there to be some bigger differences.

src/lib/optimizer/optimizer.hpp
mrks: Did you test this? The following looks like you would override a value in a dictionary:

```
for (auto rule_duration : sql_statement_metrics->optimizer_rule_durations) {
  rule_metrics_json[rule_duration.name] = rule_duration.duration.count();
}
```

mrks: If, instead of using the rule's name as the key, we kept a list of applied rules, we could also display the rules in the order the optimizer executed them. However, we need to keep in mind that the upcoming changes to the plan cache might mean that the second execution of a query executes different rules, potentially even in a different order. I have no good solution to that yet.

scripts/plot_performance_breakdown.py
Bensk1: In my opinion, the actual execution order does not matter. So anything that's not random is fine.

src/lib/optimizer/optimizer.hpp
mrks: Ah, makes sense. No good solution yet. I'm not a fan of output parameters, but changing the return type to a more complicated one just to include information that does not pertain to the actual job sounds worse.

src/lib/optimizer/optimizer.cpp
Bensk1: Agreed! Performance measurements for debug builds have usually a low significance. But on the other hand, it would be interesting to see if this timing has any measurable impact on benchmark performance.

j-tr: I compared the TPCC benchmarks in release build with and without optimizer rule timing and it seems that it has indeed an impact. 5% seem a little bit too much for enabeling it by default. what do you think about another benchmark configuration that enables the timing only if needed?

```
| Delivery       | 0.911102354527 | 164  | 0.866361558437 | 156  | -5%        | (run time too short) |
| New-Order      | 10.2332353592  | 1842 | 9.7576751709   | 1757 | -5%        |               0.0000 |
| Order-Status   | 0.90554690361  | 163  | 0.866361558437 | 156  | -4%        | (run time too short) |
| Payment        | 9.78879547119  | 1762 | 9.32449436188  | 1679 | -5%        | (run time too short) |
| Stock-Level    | 0.90554690361  | 163  | 0.866361558437 | 156  | -4%        | (run time too short) |
| geometric mean |                |      |                |      | -5%        |                      |

```

mrks: `--metrics`?

Bensk1: `--optimizer-metrics`?

mrks: Do we want/need different metrics levels?

mrks: Result of the fortnightly: Let's check if we can reduce the tracking cost first.

j-tr: it was indeed the string demangling that added so much time. if we do that in the python script, we get only -1% performance overhead. could we live with that?

```
| Delivery       | 0.911095440388 | 164  | 0.899798333645 | 162  | -1%        | (run time too short) |
| New-Order      | 10.2220458984  | 1840 | 10.1588344574  | 1829 | -1%        |               0.6388 |
| Order-Status   | 0.905539989471 | 163  | 0.899798333645 | 162  | -1%        | (run time too short) |
| Payment        | 9.76094341278  | 1757 | 9.71448898315  | 1749 | -0%        | (run time too short) |
| Stock-Level    | 0.911095440388 | 164  | 0.899798333645 | 162  | -1%        | (run time too short) |
| geometric mean |                |      |                |      | -1%        |                      |

```

scripts/plot_performance_breakdown.py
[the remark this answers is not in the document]
j-tr: i went for an optional seperate plot without the non-optimizer steps. exporting subplots seems cumbersome and since sharing axes is not very helpful in our case i decided against subplots.

src/lib/optimizer/optimizer.hpp
j-tr: you were right, in the json export, the duration was overwritten. now it's appended to a list and summed up in the python script

src/lib/optimizer/optimizer.cpp
mrks: Could you run TPC-H with SF 0.01 just to get a second data point, please?

scripts/plot_performance_breakdown.py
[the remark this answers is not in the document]
mrks: In that case, I'd simply plot it in a second file, skipping the need for a CLI parameter. Not sure what that parameter adds. @Bensk1?

src/lib/optimizer/optimizer.cpp
j-tr: For TPC-H also -1% mean:

```
+----------------+---------------+--------+---------------+--------+------------+---------+
| Benchmark      | prev. iter/s  | runs   | new iter/s    | runs   | change [%] | p-value |
+----------------+---------------+--------+---------------+--------+------------+---------+
| TPC-H 01       | 69.6706771851 | 4181   | 66.8618774414 | 4012   | -4%        |  0.0000 |
| TPC-H 02       | 517.577331543 | 31055  | 520.63079834  | 31238  | +1%        |  0.4273 |
| TPC-H 03       | 593.663513184 | 35620  | 595.178283691 | 35711  | +0%        |  0.0072 |
| TPC-H 04       | 886.466186523 | 53188  | 888.989257812 | 53340  | +0%        |  0.0000 |
| TPC-H 05       | 355.508605957 | 21331  | 356.893920898 | 21414  | +0%        |  0.1542 |
| TPC-H 06       | 3368.53833008 | 202113 | 3361.73608398 | 201705 | -0%        |  0.0000 |
| TPC-H 07       | 236.379425049 | 14183  | 230.260604858 | 13816  | -3%        |  0.1846 |
| TPC-H 08       | 24.7375068665 | 1485   | 24.7477493286 | 1485   | +0%        |  0.9699 |
| TPC-H 09       | 190.273300171 | 11417  | 189.314483643 | 11359  | -1%        |  0.3475 |
| TPC-H 10       | 324.644439697 | 19479  | 323.588653564 | 19416  | -0%        |  0.0000 |
| TPC-H 11       | 1094.73828125 | 65685  | 1099.55517578 | 65974  | +0%        |  0.0000 |
| TPC-H 12       | 726.192687988 | 43572  | 728.618591309 | 43718  | +0%        |  0.0010 |
| TPC-H 13       | 256.077880859 | 15365  | 240.425704956 | 14426  | -6%        |  0.0000 |
| TPC-H 14       | 1389.7442627  | 83385  | 1394.77392578 | 83687  | +0%        |  0.0000 |
| TPC-H 15       | 391.153015137 | 23470  | 373.581420898 | 22415  | -4%        |  0.0000 |
| TPC-H 16       | 241.167785645 | 14471  | 236.514419556 | 14191  | -2%        |  0.0000 |
| TPC-H 17       | 1198.90344238 | 71935  | 1203.57421875 | 72215  | +0%        |  0.0504 |
| TPC-H 18       | 200.514389038 | 12031  | 200.445114136 | 12027  | -0%        |  0.3096 |
| TPC-H 19       | 99.2068862915 | 5953   | 98.2062072754 | 5893   | -1%        |  0.0000 |
| TPC-H 20       | 189.089447021 | 11346  | 186.923522949 | 11216  | -1%        |  0.0000 |
| TPC-H 21       | 104.742706299 | 6285   | 104.421081543 | 6266   | -0%        |  0.3499 |
| TPC-H 22       | 340.831390381 | 20450  | 336.041107178 | 20163  | -1%        |  0.0000 |
| geometric mean |               |        |               |        | -1%        |         |
+----------------+---------------+--------+---------------+--------+------------+---------+

```

scripts/plot_performance_breakdown.py
```diff
@@ -1,65 +1,143 @@
...
         for metrics in run['metrics']:
             sum_parse_duration += metrics['parse_duration']
 
             for statement in metrics['statements']:
                 sum_sql_translation_duration += statement['sql_translation_duration']
                 sum_optimization_duration += statement['optimization_duration']
+                if statement['optimizer_rule_durations']:
+                    for rule_name, rule_duration in statement['optimizer_rule_durations'].items():
+                        # crop typeid().name down to the actual rule name. This might be compiler dependent
+                        rule_name = rule_name[11:-1]
```
mrks, line 49: Did you check gcc and clang? We could also use cxxfilt if it is an issue, but that might be overkill.

j-tr: astonishingly this works with clang-9 on mac and linux

scripts/plot_performance_breakdown.py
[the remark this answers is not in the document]
j-tr: i we have the optimizer in a seperate plot, i think it an extra parameter doesn't add that much. removed it.

mrks:
ready to merge?

PR #2171, opened 2020-06-19
Remove Counting Quotient Filters

Bouncner:
A few months ago, I was the only one arguing to keep counting quotient filters (CQF) in Hyrise as I planned to evaluate them. As usual, I haven't found the time to do that and I don't think I'll find the time in the upcoming months.

In case they are of interest again, reverting this PR should be very simple.

The reason to remove it right now is that they don't compile on ARM servers. But that's just a side note.

@mrks, @Bensk1 : I don't know what your plans concerning indexes and transactional processing are. That's something where CQF might be a very good fit. Your opinions?


Bensk1:
> @mrks, @Bensk1 : I don't know what your plans concerning indexes and transactional processing are. That's something where CQF might be a very good fit. Your opinions?

As we were until now not planning to evaluate them again, I would probably not keep them just because of that possibility. As you said, we could just revert the PR.

mrks:
Agreed. 

I take it that you already grepped for CQF and quotient in the code and the wiki?

Bouncner:
I searched for `cqf` as well as `counting.*q` and haven't found anything left in this PR. Found one reference in the wiki and removed it.

PR #2172, opened 2020-06-19
Shallow clone for third party module nlohmann

Bouncner:
Our third party module `nlohmann_json` has recently deleted its test files and shrunk from ~260 MB to ~20 MB.
This PR uses the shallow submodule flag to only pull the most recent version (same as a clone with `--depth 1`).

```
$ git clone --single-branch --branch master --recursive https://github.com/hyrise/hyrise.git master_clone
$ git clone --single-branch --branch martin/shallow_nlohmann_submodule --recursive https://github.com/hyrise/hyrise.git shallow_branch_clone
$ du -h -d1
679M	./master_clone
408M	./shallow_branch_clone
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -40,6 +40,7 @@
 [submodule "third_party/nlohmann_json"]
 	path = third_party/nlohmann_json
 	url = https://github.com/nlohmann/json.git
+	shallow = true
 [submodule "third_party/sql-parser"]
 	path = third_party/sql-parser
 	url = https://github.com/hyrise/sql-parser


mrks:
Thank god they did. Any reason not to do the same for all of the submodules?

Bouncner:
I was wondering the same and thought "there might be people working more with submodules and that would hinder them while saving single digit MBs".

I work on submodule twice a year, so I wouldn't bother. Any objections by anybody to apply that to all submodules?

mrks:
Thinking about it again, nlohmann/json was by far the biggest submodule. The time spent discussing and testing this for other modules (and potentially tracking down errors in the future) is probably not worth the split seconds of checkout time that we save.

If you want to add it for consistency, go ahead. If not, don't.

PR #2181, opened 2020-07-08
PluginManager: Fix loading of two plugins

mrks:


src/test/lib/utils/plugin_manager_test.cpp
```diff
@@ -91,6 +91,15 @@ TEST_F(PluginManagerTest, LoadingNonInstantiableLibrary) {
...
 }
 
+TEST_F(PluginManagerTest, LoadingDifferentPlugins) {
+  auto& pm = Hyrise::get().plugin_manager;
+  auto& plugins = get_plugins();
+
+  EXPECT_EQ(plugins.size(), 0u);
+  pm.load_plugin(build_dylib_path("libhyriseTestPlugin"));
+  pm.load_plugin(build_dylib_path("libhyriseMVCCDeletePlugin"));
+}
```
Bensk1, line 101: This test is considered successful if the loading of the second plugin does not crash, right? For an outsider the test might become clearer with another `EXPECT_EQ(get_plugins().size(), 2u);` in the end. What do you think?

src/lib/utils/plugin_manager.cpp
```diff
@@ -10,8 +10,10 @@
 namespace opossum {
 
 bool PluginManager::_is_duplicate(const std::unique_ptr<AbstractPlugin>& plugin) const {
+  const auto& plugin_ref = *plugin;
   for (const auto& [_, plugin_handle_wrapper] : _plugins) {
-    if (typeid(plugin_handle_wrapper.plugin) == typeid(plugin)) {
+    const auto& loaded_plugin = *plugin_handle_wrapper.plugin;
```
Bensk1, line 15: For consistency `existing_plugin_ref` instead of `loaded_plugin`?

PR #2188, opened 2020-07-28
Export Chunk Sort Information to Binaries

dey4ss:
This PR adds the ColumnSortDefinitions of each chunk to the exported binary files.


src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -275,4 +280,21 @@ AbstractTableGenerator::SortOrderByTable AbstractTableGenerator::_sort_order_by_
...
+  for (ChunkID chunk_id{0}; chunk_id < table->chunk_count(); ++chunk_id) {
+    bool chunk_sorted = false;
+    for (const auto& sort_column_definition : table->get_chunk(chunk_id)->sorted_by()) {
+      if (sort_column_definition.column == sort_definition.column) {
+        Assert(sort_column_definition.sort_mode == sort_definition.sort_mode,
+               "Column is already sorted by another SortMode");
+        chunk_sorted = true;
+      }
+    }
+    table_sorted &= chunk_sorted;
```
Bensk1, line 295: I am fine with this implementation as it is easy to understand. It could be made more efficient by returning `false` as early as the first chunk is not sorted.

resources/test_data/csv/float.csv.json
```diff
@@ -2,7 +2,18 @@
     "columns": [
         {
             "name": "a",
+            "nullable": false,
             "type": "float"
         }
-    ]
+    ],
```
Bensk1, line 8: Hm?

dey4ss: I just exported this somewhere, is seems we export the standard CSV config now :D
But I guess this can be removed

src/benchmarklib/abstract_table_generator.cpp
Bensk1: That was actually not what I meant but it's another optimization. I meant if `chunk_sorted` in 297 turns `table_sorted` to `false`, all further chunks will still be evaluated while `table_sorted` cannot turn true anymore. But you don't have to change the current implementation, it's your call.

dey4ss: OK, got it :D

src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -275,4 +280,22 @@ AbstractTableGenerator::SortOrderByTable AbstractTableGenerator::_sort_order_by_
...
     std::unordered_map<std::string, BenchmarkTableInfo>& table_info_by_name) const {}
 
+bool AbstractTableGenerator::_table_is_sorted_by(const std::shared_ptr<Table>& table,
+                                                 const SortColumnDefinition& sort_definition) {
+  for (ChunkID chunk_id{0}; chunk_id < table->chunk_count(); ++chunk_id) {
+    const auto& sort_column_definitions = table->get_chunk(chunk_id)->sorted_by();
+    if (sort_column_definitions.empty()) return false;
+    bool chunk_sorted = false;
+    for (const auto& sort_column_definition : sort_column_definitions) {
+      if (sort_column_definition.column == sort_definition.column) {
```
mrks, line 290: Can you find better names for `sort_column_definition` and `sort_definition`?

src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -275,4 +280,22 @@ AbstractTableGenerator::SortOrderByTable AbstractTableGenerator::_sort_order_by_
 void AbstractTableGenerator::_add_constraints(
     std::unordered_map<std::string, BenchmarkTableInfo>& table_info_by_name) const {}
 
+bool AbstractTableGenerator::_table_is_sorted_by(const std::shared_ptr<Table>& table,
+                                                 const SortColumnDefinition& sort_definition) {
```
mrks, line 284: The way this looks to me is as if you are only checking if the individual chunks are sorted. That, however, does not mean that the entire table is sorted, too.

mrks, line 284: The "sorted" property of tables and chunks is somewhat incompatible. If a table is sorted on multiple columns using the `Sort` operator, it is first sorted according to one column, then to the other. For a chunk, it means this:

```
   * In a chunk, multiple segments may be sorted independently. For example, in a table storing orders, both the order
   * id and date of incoming orders might have incrementing values. In this case, sorted_by has two entries (assuming
   * this knowledge is available to the database). However, for cases where the data is first orderered by one column,
   * then by another (e.g. ORDER_BY last_name, first_name), only the primary order is stored.
```

I can see how this is misleading. Maybe we should rename the chunk method to `individually_sorted_by`?

src/lib/import_export/binary/binary_parser.cpp
```diff
@@ -101,6 +101,16 @@ std::pair<std::shared_ptr<Table>, ChunkID> BinaryParser::_read_header(std::ifstr
 void BinaryParser::_import_chunk(std::ifstream& file, std::shared_ptr<Table>& table) {
   const auto row_count = _read_value<ChunkOffset>(file);
 
+  // import sort column definitions
```
mrks, line 104: Really minor: capitalize the first letter of a comment

src/lib/import_export/binary/binary_writer.cpp
```diff
@@ -128,6 +128,16 @@ void BinaryWriter::_write_chunk(const Table& table, std::ofstream& ofstream, con
   Assert(chunk, "Physically deleted chunk should not reach this point, see get_chunk / #1686.");
   export_value(ofstream, static_cast<ChunkOffset>(chunk->size()));
 
+  // export sort column definitions
+  const auto& sort_column_definitions = chunk->sorted_by();
+  const auto sorted_columns = sort_column_definitions.size();
+  export_value(ofstream, static_cast<uint32_t>(sorted_columns));
+  for (ColumnID sorted_column_id{0}; sorted_column_id < sorted_columns; ++sorted_column_id) {
+    const auto sort_column_definition = sort_column_definitions.at(sorted_column_id);
```
mrks, line 136: `for (const auto [column, sort_mode] : sorted_columns) {` ?

src/lib/import_export/binary/binary_parser.cpp
```diff
@@ -101,6 +101,16 @@ std::pair<std::shared_ptr<Table>, ChunkID> BinaryParser::_read_header(std::ifstr
...
+  const auto sorted_columns = _read_value<uint32_t>(file);
```
mrks, line 105: `num_sorted_columns`?

src/test/lib/operators/export_test.cpp
```diff
@@ -70,8 +70,8 @@ TEST_P(OperatorsExportMultiFileTypeTest, ExportWithFileType) {
   auto table_wrapper = std::make_shared<TableWrapper>(std::move(table));
   table_wrapper->execute();
 
-  std::string reference_filename = reference_filepath + reference_filenames.at(GetParam());
-  auto exporter = std::make_shared<opossum::Export>(table_wrapper, test_filename, GetParam());
+  const auto reference_filename = reference_filepath + reference_filenames.at(GetParam());
+  auto exporter = std::make_shared<Export>(table_wrapper, test_filename, GetParam());
```
mrks, line 74: :)

src/benchmarklib/abstract_table_generator.cpp
dey4ss: Yes, this is an incorrect naming here. As it is already checked whether the table is sorted, we only make sure here that each chunk has the correct SortColumnDefinition as a consequence.

dey4ss: Seems like good idea

PR #2156, opened 2020-06-14
Add support for Unique Constraints in LQP nodes

julianmenzler:
This PR adds support for unique constraints in LQP nodes. They express uniqueness guarantees for subsets of LQP output expressions. They are meant to be used by the optimizer.

In tables, Hyrise already provides soft key constraints, which also model certain uniqueness guarantees. However, LQP-based unique constraints offer several benefits over table key constraints:

* Strict isolation: Optimizer rules can access unique constraints directly from LQP nodes by calling  `unique_constraints()`.
* Each LQP node can add its own, additional unique constraints.
* LQP nodes can remove invalid unique constraints when they are no longer valid.

## Definition
We use a container struct `LQPUniqueConstraint` to define uniqueness for a subset of LQP output expressions. 

We define output expressions as "unique" when expecting distinct value sets. Similar to table key constraints with `KeyConstraintType::UNIQUE`, our uniqueness definition allows for any number of NULL values.

## Implementation Overview
Most unique constraints are derived from table key constraints in `StoredTableNode`, `MockNode` and `StaticTableNode`.

Other nodes in the LQP tree simply forward unique constraints from their input nodes. Invalid constraints become discarded appropriately (column pruning etc.).

The following table provides a rough implementation-overview of all LQP nodes touched by this PR:

| LQP Node | unique_constraints() impl |
| :------- | :------------------------ |
| AggregateNode | Forward Selected & Create from Group-By Columns |
| AliasNode | Forward All |
| ExceptNode | Forward All |
| IntersectNode | Forward All |
| JoinNode (JoinMode::AntiNull*) | Forward All |
| JoinNode (JoinMode::Cross) | Discard All |
| JoinNode (JoinMode::Semi) | Forward All |
| JoinNode (Multi-Predicate) | Discard All |
| JoinNode (Single-Predicate) | Forward Selected |
| LimitNode | Forward All |
| MockNode | Create from Table Key Constraints |
| PredicateNode | Forward All |
| ProjectionNode | Forward Selected |
| SortNode | Forward All |
| StaticTableNode | Create from Table Key Constraints |
| StoredTableNode | Create from Table Key Constraints |
| UnionNode (SetOperationMode::Positions) | Forward All |
| UnionNode (SetOperationMode::All) | Discard All |
| ValidateNode | Forward All |

PR #2189, opened 2020-07-29
Derive FDs from LQP unique constraints

julianmenzler:
We added functional dependencies to Hyrise with #2078. This PR improves the implementation to profit from LQP unique constraints introduced via #2156.

**Previous implementation**
Before this PR, we created FDs from table key constraints in StoredTableNode. To make use of them in upper nodes, we provided some forwarding logic to `functional_dependencies()` in AbstractLQPNode.

**New implementation**
When calling `functional_dependencies()`, we return FDs from two different sources:

1. We translate the current LQP node's unique constraints into FDs, one-by-one. We call them **trivial FDs** because we can easily create them from several LQP nodes (as long as the unique constraints stay valid).
In comparison to the previous approach, we profit from unique constraints created by other nodes than StoredTableNode. Furthermore, MockNode returns trivial FDs based on its unique constraints, just like StoredTableNode.

2. We collect **non-trivial FDs** from child nodes. They can be generated from single nodes only, which is different from trivial FDs. By overriding `non_trivial_functional_dependencies()`, each LQP node can
  * add its own non-trivial FDs
  * remove invalid non-trivial FDs
  * specify forwarding of non-trivial FDs in the case of two input nodes.


src/lib/logical_query_plan/intersect_node.cpp
```diff
@@ -34,11 +34,20 @@ std::shared_ptr<LQPUniqueConstraints> IntersectNode::unique_constraints() const
...
+  DebugAssert(left_input()->unique_constraints() == right_input()->unique_constraints(),
+              "Merging of unique constraints should be implemented.");
   return _forward_left_unique_constraints();
 }
 
+std::vector<FunctionalDependency> IntersectNode::pass_functional_dependencies() const {
+  /**
+   * Future Work: Merge FDs from left and right input node.
+   */
+  DebugAssert(left_input()->pass_functional_dependencies() != right_input()->pass_functional_dependencies(),
```
mrks, line 46: I'd use Assert instead of DebugAssert as things could go really wrong here

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -172,9 +172,23 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
...
+
+  /**
+   * Helper function that passes FDs to upper nodes. By default, it passes FDs from the left input node.
+   * Some nodes may override this function to
+   *  - add additional FDs or
+   *  - to specify FD passing in case of two input nodes.
+   *
+   *  Please note: Do not pass FDs that can be generated from unique constraints because functional_dependencies()
+   *               already takes care of that. However, whenever unique constraints are discarded, it makes sense to
+   *               do so since FDs cannot be generated later on anymore.
```
mrks, line 189: Could you rephrase the second sentence?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,60 +274,80 @@ bool AbstractLQPNode::has_matching_unique_constraint(const ExpressionUnorderedSe
...
-    fds_right = right_input()->functional_dependencies();
+  // (1) Gather FDs from previous nodes
+  auto fds_in = pass_functional_dependencies();
+  if constexpr (HYRISE_DEBUG) {
+    auto distinct_fds = std::vector<FunctionalDependency>();
+    for (const auto& fd : fds_in) {
+      Assert(std::none_of(distinct_fds.cbegin(), distinct_fds.cend(),
+                          [&fd](const auto& distinct_fd) { return fd == distinct_fd; }),
+             "Did not expect LQP nodes to pass duplicate FDs.");
+      distinct_fds.push_back(fd);
```
mrks, line 285: `DebugAssert(std::set(fds_in.begin(), fds_in.end()).size() == fds_in.size())`?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,60 +274,80 @@ bool AbstractLQPNode::has_matching_unique_constraint(const ExpressionUnorderedSe
...
   }
 
-  auto& fds_in = fds_left;
-  fds_in.insert(fds_in.end(), fds_right.cbegin(), fds_right.cend());
-
-  // Currently, we do not support FDs in conjunction with null values in their determinant set.
-  // Some LQP nodes, however, change column nullability (like for instance outer joins). Therefore, we check input FDs
-  // for compliance and discard them, if necessary.
+  /**
+   * (3) Remove invalid FDs:
```
mrks, line 311: This only applies to the FDs created from (2), correct?

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -172,9 +172,23 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
...
+   *         They are collected as follows:
+   *          (1) FDs are derived from the current node's unique constraints.
+   *          (2) FDs are provided by previous nodes by passing them up the LQP tree.
    */
-  virtual std::vector<FunctionalDependency> functional_dependencies() const;
+  std::vector<FunctionalDependency> functional_dependencies() const;
+
+  /**
+   * Helper function that passes FDs to upper nodes. By default, it passes FDs from the left input node.
+   * Some nodes may override this function to
```
mrks, line 183: Have you considered a non-virtual method `_fds_from_unique_constraints`, which is then called by `virtual functional_dependencies`? For some reason, that feels much more intuitive to me.

julianmenzler: Not really. It could be an improvement code-wise. But, I do not see a big architectural change here.

Calling `_fds_from_unique_constraints` over and over again from within the LQP tree means a lot of overhead. It only makes sense when we lose unique constraints (e.g. JoinNode). Otherwise, it should be called once as the last step before returning to the user.

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler: No, this is a general rule. It also applies for FDs from (1).

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler: Thanks, very clever!
When having an `if constexpr (HYRISE_DEBUG) { }`-block, I always use regular Asserts. Shall I use DebugAssert anyway?

src/lib/logical_query_plan/abstract_lqp_node.hpp
julianmenzler: Changed to
```
*  Please note: Do not pass FDs that can be generated from unique constraints because functional_dependencies()
*               already takes care of that.
*               However, there is an exception: Some nodes, such as JoinNode, discard unique constraints.
*               Consequently, we lose FDs that can no longer be derived in upper nodes. To preserve them, we might
*               want to generate FDs from unique constraints prematurely to manually pass them up the LQP tree.
```

src/lib/logical_query_plan/abstract_lqp_node.hpp
mrks: Could you describe where this would be an overhead in more detail? Good chance that I'm missing something here

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: No hard feelings about it. Assert might be better because you might run into unused variables if you comment out the `if constexpr`. Is it still necessary here though?

julianmenzler: No, it is just my personal practice.

src/lib/logical_query_plan/abstract_lqp_node.hpp
julianmenzler: Think of an LQP as follows: 
```
  A
  |
  B
  |
  C (JoinNode)
 /  \
D    E
```
**Naive implementation**
Calling `functional_dependencies()` on A leads to:
* Call `unique_constraints()` on A
  * Call `unique_constraints()` on B
    * Call `unique_constraints()` on C
      * Call `unique_constraints()` on D
      * Call `unique_constraints()` on E
* Call `functional_dependencies()` on B
  * Call `unique_constraints()` on B
    * Call `unique_constraints()` on C
      * Call `unique_constraints()` on D
      * Call `unique_constraints()` on E
* Call `functional_dependencies()` on C
    * Call `unique_constraints()` on C
      * Call `unique_constraints()` on D
      * Call `unique_constraints()` on E
* Call `functional_dependencies()` on D
  * Call `unique_constraints()` on D
* Call `functional_dependencies()` on E
  * Call `unique_constraints()` on E

**Optimized implementation**
Calling `functional_dependencies()` on A leads to:
* Call `unique_constraints()` on A
  * Call `unique_constraints()` on B
    * Call `unique_constraints()` on C
      * Call `unique_constraints()` on D
      * Call `unique_constraints()` on E
* Call `pass_functional_dependencies()` on B
* Call `pass_functional_dependencies()` on C (JoinNode)
  * Call `functional_dependencies()` on D
    * Call `unique_constraints()` on D
  * Call `functional_dependencies()` on E
    * Call `unique_constraints()` on E


Hope it helps. Regarding the special Join-handling, see https://github.com/hyrise/hyrise/pull/2189#discussion_r464344365

mrks: Yepp, that makes sense. Good call. I don't like the method's name too much though. Maybe `_non_unique_functional_dependencies` instead of `pass_functional_dependencies`?

mrks: Generally, I don't think that "pass" is the best word here, neither in the method name nor in the comment.

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -172,9 +172,23 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
...
+   *  - add additional FDs or
```
mrks, line 184: additional FDs that are not based on unique constraints (e.g., `a -> a+1`, which is not implemented yet)

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: Which is good. My point was that the entire block should become redundant with the std::set hack from above

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,60 +274,80 @@ bool AbstractLQPNode::has_matching_unique_constraint(const ExpressionUnorderedSe
...
 std::vector<FunctionalDependency> AbstractLQPNode::functional_dependencies() const {
-  // Gather input FDs from all input nodes
-  auto fds_left = std::vector<FunctionalDependency>();
-  auto fds_right = std::vector<FunctionalDependency>();
-  if (left_input()) {
-    fds_left = left_input()->functional_dependencies();
-  }
-  if (right_input()) {
-    fds_right = right_input()->functional_dependencies();
+  // (1) Gather FDs from previous nodes
```
mrks, line 277: Explain that these exclude FDs based on unique constraints?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,60 +274,80 @@ bool AbstractLQPNode::has_matching_unique_constraint(const ExpressionUnorderedSe
...
+  for (const auto& unique_constraint : *unique_constraints) {
+    auto determinants = unique_constraint.expressions;
+    auto dependents = ExpressionUnorderedSet();
+    for (const auto& output_expression : output_expressions) {
+      if (!determinants.contains(output_expression)) {
+        dependents.insert(output_expression);
+      }
+    }
+
+    // Add the FD to the output if not already existing
```
mrks, line 301: Maybe I haven't fully understood the concept yet. In which case would the FD already exist? Wouldn't that mean that some input node passed an FD that is based on a unique constraint?

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: But a column should not suddenly become nullable. If an FD from (1) acts on a nullable column, isn't the input faulty?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,60 +274,80 @@ bool AbstractLQPNode::has_matching_unique_constraint(const ExpressionUnorderedSe
...
 
   return fds_out;
 }
 
+std::vector<FunctionalDependency> AbstractLQPNode::pass_functional_dependencies() const {
+  if (left_input()) {
+    Assert(!right_input(), "Expected single input node for implicit FD forwarding. Please override this function.");
+    return left_input()->pass_functional_dependencies();
+  } else {
+    return {};
```
mrks, line 347: When is this called? Should we `Assert(left_input && !right_input)`?

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -22,8 +22,8 @@ AggregateNode::AggregateNode(const std::vector<std::shared_ptr<AbstractExpressio
       aggregate_expressions_begin_idx{group_by_expressions.size()} {
 #if HYRISE_DEBUG
   for (const auto& aggregate_expression : aggregate_expressions) {
-    DebugAssert(aggregate_expression->type == ExpressionType::Aggregate,
-                "Expression used as aggregate expression must be of type AggregateExpression.");
+    Assert(aggregate_expression->type == ExpressionType::Aggregate,
```
mrks, line 25: Change to if constexpr?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -123,6 +123,27 @@ std::shared_ptr<LQPUniqueConstraints> JoinNode::unique_constraints() const {
   return unique_constraints;
 }
 
+std::vector<FunctionalDependency> JoinNode::pass_functional_dependencies() const {
+  /**
+   * Due to the logic of joins, we might lose several unique constraints in this node. Consequently, upper nodes
+   * will have less unique constraints to derive functional dependencies from.
+   * Consequently, we create functional dependencies from both input nodes to manually pass them up the LQP tree.
```
mrks, line 130: Can you remove one "consequently"?

mrks, line 130: Can we create FDs in the case of 1:1 and 1:n joins? Not saying that we have to add it right away, but if yes, we probably should point it out.

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler:
```cpp
  if constexpr (HYRISE_DEBUG) {
    const auto distinct_fds = std::unordered_set<FunctionalDependency>(fds_in.cbegin(), fds_in.cend());
    Assert(distinct_fds.size() == fds_in.size(), "Did not expect LQP nodes to pass duplicate FDs.");
  }
```

I stick with the block to remove the variable from release builds.

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler: Yes, it is. Since `pass_functional_dependencies()` is a pure helper function to pass FDs, it is not required to filter FDs. It is only required to check for duplicates before adding to it.

By design, validity-checks are performed at the end, in the top node.

src/lib/logical_query_plan/abstract_lqp_node.hpp
julianmenzler: I already had a hard time thinking about a good, short name for this type of helper function. 

To summarize: The function returns FDs that cannot be derived from unique constraints in upper nodes. 
* Some FDs might have become invalid due to projection or nullability changes
* No duplicate FDs in the output

To be honest, I do not like `_non_unique_functional_dependencies` either. What is a non-unique FD? :/

julianmenzler: Maybe `non_derivable_functional_dependencies()` ?

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler: _(1) Gather FDs from previous nodes that cannot be derived from unique constraints_

Better?

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler: This block is due to the special handling of joins. 
In JoinNode, we prematurely create FDs from unique constraints because we might lose some in upper nodes.
To simplify the implementation, JoinNode creates FDs from all unique constraints. 

However, not all unique constraints get lost. Therefore, we might end up with FD duplicates here.

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler: For example, in StoredTableNode, MockNode and StaticTableNode.

src/lib/logical_query_plan/join_node.cpp
julianmenzler: Do you think of the following:

Two tables: 
* A, B
* C, D

Join on "B = C" should result in
* B -> C
* C -> B

src/lib/operators/union_positions.cpp
```diff
@@ -267,47 +267,48 @@ std::shared_ptr<const Table> UnionPositions::_prepare_operator() {
     _referenced_column_ids.emplace_back(ref_segment->referenced_column_id());
   }
 
-#if HYRISE_DEBUG
-  /**
+  if constexpr (HYRISE_DEBUG) {
+    /**
```
julianmenzler, line 271: indent

src/lib/logical_query_plan/abstract_lqp_node.hpp
mrks: non-trivial?

src/lib/logical_query_plan/abstract_lqp_node.hpp
mrks: Yepp. :)

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: ok ¯\\\_(ツ)\_/¯

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: Mmmh, I'm not sure if I like this. For me, it would be more natural to (1) debug-assert that FDs coming from `pass_functional_dependencies` are not based on nullable columns and (2) skip the FD generations in the `for (const auto& output_expression : output_expressions)` block if the expression is nullable.

I won't block the PR because of this, but please consider this.

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: Yepp. Maybe include the trivial/derivable terminology from above.

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: Hmm, but doesn't that violate the "non-derivable/non-trivial FD" contract discussed above? Having a method that does not forward unique constraints *unless there was a join somewhere below that node* seems hard to deal with.

src/lib/logical_query_plan/abstract_lqp_node.cpp
mrks: Hmm, I guess overriding the method over there does not really help with anything.

src/lib/logical_query_plan/join_node.cpp
mrks: Correct. Also, if C is unique, B->D should also hold. Which, in turn makes me realize that we have not talked about transitive FDs yet. Again - not something that we should do in this PR, but we should highlight this concrete case (B->C and possible B->D) for the Join and in `functional_dependency.hpp`. For the latter, we might suggest that a future method `fds_apply(fds, dependent, dependee)` could take a set of FDs and two expressions to see if `dependee` is dependent on `dependent`. Basically what we are doing by hand for now. That could include checking for transitive FDs.

src/lib/logical_query_plan/abstract_lqp_node.hpp
julianmenzler: "non-trivial" works great for me 👍

src/lib/utils/performance_warning.cpp
```diff
@@ -3,11 +3,11 @@
 namespace opossum {
 
 bool PerformanceWarningClass::_disabled = []() {  // NOLINT
-// static initializer hack to print some warnings in various binaries
+  // static initializer hack to print some warnings in various binaries
 
-#if HYRISE_DEBUG
-  PerformanceWarning("Hyrise is running as a debug build.");
-#endif
+  if constexpr (HYRISE_DEBUG) {
```
mrks, line 8: Thanks!

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -123,6 +123,44 @@ std::shared_ptr<LQPUniqueConstraints> JoinNode::unique_constraints() const {
   return unique_constraints;
 }
 
+std::vector<FunctionalDependency> JoinNode::non_trivial_functional_dependencies() const {
+  if (join_mode != JoinMode::Semi && join_mode != JoinMode::AntiNullAsTrue && join_mode != JoinMode::AntiNullAsFalse) {
```
mrks, line 127: Guard clause

src/lib/logical_query_plan/functional_dependency.cpp
```diff
@@ -51,3 +53,13 @@ std::ostream& operator<<(std::ostream& stream, const FunctionalDependency& expre
 }
 
 }  // namespace opossum
+
+namespace std {
+
+size_t hash<opossum::FunctionalDependency>::operator()(const opossum::FunctionalDependency& fd) const {
+  auto hash = boost::hash_range(fd.determinants.cbegin(), fd.determinants.cend());
+  boost::hash_combine(hash, boost::hash_range(fd.dependents.cbegin(), fd.dependents.cend()));
```
mrks, line 61: If one set of determinants determines multiple dependents, they should be grouped in a single FD. It's not an issue if they are not, but I'd expect the common case to be that they are. As such, including the dependents in the hash might be expensive without adding too much to the collision safety.

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -222,6 +242,18 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
...
+   * @return TODO
+   */
+  std::vector<FunctionalDependency> _fds_from_unique_constraints() const;
+
+  /**
+   * This is a helper method to TODO(Julian) removes FDs when expressions are missing
+   * @return TODO
+   */
+  void _remove_invalid_fds(std::vector<FunctionalDependency>& fds) const;
+
```
mrks, line 256: ?

src/lib/logical_query_plan/intersect_node.cpp
```diff
@@ -34,11 +34,20 @@ std::shared_ptr<LQPUniqueConstraints> IntersectNode::unique_constraints() const
...
   return _forward_left_unique_constraints();
 }
 
+std::vector<FunctionalDependency> IntersectNode::non_trivial_functional_dependencies() const {
+  /**
+   * Future Work: Merge FDs from left and right input node.
+   */
+  Assert(left_input()->non_trivial_functional_dependencies() != right_input()->non_trivial_functional_dependencies(),
+         "Merging of FDs should be implemented.");
+  return left_input()->non_trivial_functional_dependencies();
```
mrks, line 48: Why `!=`?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,58 +276,88 @@ bool AbstractLQPNode::has_matching_unique_constraint(const ExpressionUnorderedSe
...
-    fds_right = right_input()->functional_dependencies();
-  }
-
-  if (HYRISE_DEBUG && !fds_right.empty()) {
-    for (const auto& fd_right : fds_right) {
-      const bool duplicate = std::any_of(fds_left.begin(), fds_left.end(),
-                                         [&fd_right](const auto& fd_left) { return (fd_left == fd_right); });
-      Assert(!duplicate,
-             "Unexpected duplicate functional dependency found in " + this->description(DescriptionMode::Short));
+  // (1) Gather non-trivial FDs from previous, which cannot be derived from unique constraints
```
mrks, line 279: "previous"? Maybe "Gather non-trivial FDs and perform sanity checks"?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,58 +276,88 @@ bool AbstractLQPNode::has_matching_unique_constraint(const ExpressionUnorderedSe
...
-  // for compliance and discard them, if necessary.
-  auto fds_out = std::vector<FunctionalDependency>();
-
-  // Collect non-nullable columns
-  const auto expressions = output_expressions();
-  auto output_expressions_non_nullable = ExpressionUnorderedSet{};
-  for (auto column_id = ColumnID{0}; column_id < expressions.size(); ++column_id) {
-    if (!is_column_nullable(column_id)) {
-      output_expressions_non_nullable.insert(expressions.at(column_id));
+  // (2) Generate FDs from current node's unique constraints
```
mrks, line 303: Generate trivial FDs?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -403,10 +435,55 @@ std::shared_ptr<LQPUniqueConstraints> AbstractLQPNode::_forward_left_unique_cons
...
+                     })) {
+      continue;
+    }
+    // (2) Collect the dependent output expressions
+    auto dependents = ExpressionUnorderedSet();
+    for (const auto& output_expression : output_expressions) {
+      if (!determinants.contains(output_expression)) {
+        dependents.insert(output_expression);
+      }
+    }
```
mrks, line 472: Guard clause

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -123,6 +123,44 @@ std::shared_ptr<LQPUniqueConstraints> JoinNode::unique_constraints() const {
...
+    /**
+     * In the course of inner/outer joins,
+     *  (a) some unique constraints and therefore some trivial FDs of the JoinNode's input nodes might become lost.
+     *  (b) some unique constraints might stay valid. Due to the increased number of output expressions, however,
+     *      derived FDs become longer. For example: - Table A: (UNIQUE(a), b) leads to the FD {a} -> {b}
+     *                                              - Table B: (x, y)
+     *                                              After joining both tables, we may end up with
+     *                                              - (UNIQUE(a), b, x, y) which leads to the FD {a} -> {b, x, y}
```
mrks, line 135: Isn't this only the case if x is unique too? Also, is this something that we already do? I can't find it in the code.

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,58 +276,88 @@ bool AbstractLQPNode::has_matching_unique_constraint(const ExpressionUnorderedSe
...
+
+                             // Remove dependents that are not part of the node's output expressions
+                             auto part_of_output_expressions =
+                                 [&output_expressions_set](const auto& fd_dependent_expression) {
+                                   return !output_expressions_set.contains(fd_dependent_expression);
+                             };
+                             std::erase_if(fd.dependents, part_of_output_expressions);
+
+                             // If there are no dependents left, we can discard the FD altogether
+                             return fd.dependents.size() == 0;
```
mrks, line 348: Calling `erase_if` in a nested way feels like something where things could go bad really quickly. Would it make a difference to first iterate over the FDs and remove the dependents and then remove empty FDs in a separate step?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,58 +276,88 @@ bool AbstractLQPNode::has_matching_unique_constraint(const ExpressionUnorderedSe
...
+}
 
-    // FD remains valid, so add it to the output vector
-    fds_out.push_back(fd);
+void AbstractLQPNode::_remove_invalid_fds(std::vector<FunctionalDependency>& fds) const {
+  if (fds.empty()) return;
+  const auto& output_expressions = this->output_expressions();
+  const auto& output_expressions_set = ExpressionUnorderedSet{output_expressions.cbegin(), output_expressions.cend()};
+  fds.erase(std::remove_if(fds.begin(), fds.end(),
+                           [this, &output_expressions_set](const auto& fd) {
```
mrks, line 327: Remove const to fix the build

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,58 +276,88 @@ bool AbstractLQPNode::has_matching_unique_constraint(const ExpressionUnorderedSe
...
+                             };
+                             std::erase_if(fd.dependents, part_of_output_expressions);
+
+                             // If there are no dependents left, we can discard the FD altogether
+                             return fd.dependents.size() == 0;
+                           }),
+            fds.end());
+}
+
+std::vector<FunctionalDependency> AbstractLQPNode::non_trivial_functional_dependencies() const {
```
mrks, line 353: I can replace this with `return {};` without any tests failing. :/

We should have a handful of more complex LQPs in the tests.

src/lib/logical_query_plan/intersect_node.cpp
mrks: Fail!

src/lib/logical_query_plan/join_node.cpp
mrks: Clarify that this is not yet done

src/lib/logical_query_plan/functional_dependency.cpp
julianmenzler: Both has been implemented now.

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler: It leads to a bit more runtime. Nonetheless, we get the same result and it looks cleaner. I changed it ;)

src/lib/logical_query_plan/abstract_lqp_node.cpp
julianmenzler: I have added several tests for
* AggregateNode
* ProjectionNode
* JoinNode

Not sure though, if we need another one.

src/lib/logical_query_plan/join_node.cpp
julianmenzler: Very useful ideas. I added notes for both of them.

src/lib/logical_query_plan/functional_dependency.hpp
```diff
@@ -39,4 +39,33 @@ struct FunctionalDependency {
 
 std::ostream& operator<<(std::ostream& stream, const FunctionalDependency& expression);
 
+/**
+ * @return A merger FD set from the given input @param fds_a and @param fds_b. FDs with the same determinant
+ *         expressions are merged into single objects by merging their dependent expressions.
+ */
+std::vector<FunctionalDependency> merge_fds(const std::vector<FunctionalDependency>& fds_a,
+                                            const std::vector<FunctionalDependency>& fds_b);
```
julianmenzler, line 47: I am not sure, where to put this helper. Some other FD-related helper functions are in `lqp_utils.cpp`. However, since this merge function does not involve any LQPs, putting it into `lqp_utils.cpp` also seems wrong.

src/lib/logical_query_plan/functional_dependency.cpp
```diff
@@ -50,4 +52,43 @@ std::ostream& operator<<(std::ostream& stream, const FunctionalDependency& expre
...
+    auto existing_fd = std::find_if(fds_merged.begin(), fds_merged.end(),
+                                    [&fd_to_add](auto& fd) { return fd.determinants == fd_to_add.determinants; });
+    if (existing_fd == fds_merged.end()) {
+      fds_merged.push_back(fd_to_add);
+    } else {
+      // An FD with the same determinant expressions already exists. Therefore, we only have to add to the dependent
+      // expressions set
+      existing_fd->dependents.insert(fd_to_add.dependents.begin(), fd_to_add.dependents.end());
+    }
+  }
```
julianmenzler, line 81: The merge logic for FDs with duplicate determinant expressions seems necessary. But to be honest, I could not come up with an LQP scenario that requires the merging of such FDs. 
Should we stay with the above logic or replace it with pure vector-copying and an Assert?

src/lib/logical_query_plan/functional_dependency.hpp
mrks: This is fine. In the long term, I wonder if the FD container should become its own class.

src/lib/logical_query_plan/functional_dependency.cpp
mrks: Not sure; I can't come up with anything right now either. Your call. If it stays, it should be tested though.

src/lib/logical_query_plan/functional_dependency.hpp
```diff
@@ -39,4 +39,33 @@ struct FunctionalDependency {
 
 std::ostream& operator<<(std::ostream& stream, const FunctionalDependency& expression);
 
+/**
+ * @return A merger FD set from the given input @param fds_a and @param fds_b. FDs with the same determinant
```
mrks, line 43: merged?

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -208,69 +208,157 @@ TEST_F(JoinNodeTest, IsColumnNullableWithOuterJoin) {
...
+      join_node = JoinNode::make(JoinMode::Cross, _mock_node_a, _mock_node_b);
+    } else {
+      join_node = JoinNode::make(join_mode, equals_(join_column_a, join_column_x), _mock_node_a, _mock_node_b);
+    }
+
+    // Left input Node has non-trivial FDs
+    _mock_node_a->set_key_constraints({});
+    _mock_node_b->set_key_constraints({});
+    _mock_node_a->set_non_trivial_functional_dependencies({fd_a});
+    _mock_node_b->set_non_trivial_functional_dependencies({});
```
mrks, line 259: This looks like you are resetting the test a couple of times. Split this into several test cases to improve isolation?

src/lib/logical_query_plan/join_node.hpp
```diff
@@ -38,11 +38,34 @@ class JoinNode : public EnableMakeForLQPNode<JoinNode>, public AbstractLQPNode {
    */
   std::shared_ptr<LQPUniqueConstraints> unique_constraints() const override;
 
+  /**
+   * (a) Semi- & Anti-Joins: Forwards left input node's non-trivial FDs
+   * (b) Inner-/Outer-/Cross-Joins:
+   *      - Forwards non-trivial FDs from both input nodes that stay valid.
+   *      - Returns derived, non-trivial FDs from input unique constraints that become discarded due to the join.
```
mrks, line 45: Can you clarify "that stay valid"? Do you mean that they are not nullable?
Also, turn the second item into "Previously trivial FDs turn into non-trivial FDs as the determinant columns become non-unique, but the FDs still hold true"?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -92,35 +101,89 @@ std::shared_ptr<LQPUniqueConstraints> JoinNode::unique_constraints() const {
...
+  }
+
+  auto fds_left = left_input()->non_trivial_functional_dependencies();
+  auto fds_right = right_input()->non_trivial_functional_dependencies();
+
+  /**
+   * When joining tables, we usually lose some or even all unique constraints from both input tables. This leads to
+   * less trivial FDs that we can generate from unique constraints in upper nodes. To preserve all FDs possible, we
+   * generate and return all FDs from both input nodes.
+   * In the following, we determine the unique constraints that become discarded and built FDs from them.
```
mrks, line 150: build

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -92,35 +101,89 @@ std::shared_ptr<LQPUniqueConstraints> JoinNode::unique_constraints() const {
...
+  if (join_mode == JoinMode::Semi || join_mode == JoinMode::AntiNullAsTrue || join_mode == JoinMode::AntiNullAsFalse) {
+    return left_input()->non_trivial_functional_dependencies();
+  }
+
+  auto fds_left = left_input()->non_trivial_functional_dependencies();
+  auto fds_right = right_input()->non_trivial_functional_dependencies();
+
+  /**
+   * When joining tables, we usually lose some or even all unique constraints from both input tables. This leads to
+   * less trivial FDs that we can generate from unique constraints in upper nodes. To preserve all FDs possible, we
```
mrks, line 148: fewer

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -92,35 +101,89 @@ std::shared_ptr<LQPUniqueConstraints> JoinNode::unique_constraints() const {
...
+   * generate and return all FDs from both input nodes.
+   * In the following, we determine the unique constraints that become discarded and built FDs from them.
+   */
+  const auto& left_unique_constraints = left_input()->unique_constraints();
+  const auto& right_unique_constraints = right_input()->unique_constraints();
+  const auto& valid_unique_constraints = _valid_unique_constraints(left_unique_constraints, right_unique_constraints);
+  if (valid_unique_constraints->empty()) {
+    // All unique constraints become discarded
```
mrks, line 156: ..., so the FDs need to be added manually as non-trivial FDs

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -92,35 +101,89 @@ std::shared_ptr<LQPUniqueConstraints> JoinNode::unique_constraints() const {
...
+    if (!left_unique_constraints->empty()) {
```
mrks, line 157: Do we need the `x->empty()` checks? `merge_fds` has a shortcut and I don't think they add anything to the clarity here.

src/lib/logical_query_plan/join_node.hpp
```diff
@@ -38,11 +38,34 @@ class JoinNode : public EnableMakeForLQPNode<JoinNode>, public AbstractLQPNode {
...
+   *         right_unique_constraints that remains valid despite the join operation.
+   *         Depending on the join columns,
+   *          (a) the left,
+   *          (b) the right,
+   *          (c) both or
+   *          (d) none
+   *         of the given unique constraint sets are returned.
+   * Please note: This helper function can be called for all joins, except for Semi- and Anti-Join types.
+   */
+  std::shared_ptr<LQPUniqueConstraints> _valid_unique_constraints(
```
mrks, line 65: Why not _output_unique_constraints?

src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -354,4 +350,85 @@ bool contains_matching_unique_constraint(const std::shared_ptr<LQPUniqueConstrai
...
+  return fds;
+}
+
+void remove_invalid_fds(const std::shared_ptr<const AbstractLQPNode>& lqp, std::vector<FunctionalDependency>& fds) {
+  if (fds.empty()) return;
+  const auto& output_expressions = lqp->output_expressions();
+  const auto& output_expressions_set = ExpressionUnorderedSet{output_expressions.cbegin(), output_expressions.cend()};
+
+  // Adjust FDs: Remove dependents that are not part of the node's output expressions
+  auto part_of_output_expressions = [&output_expressions_set](const auto& fd_dependent_expression) {
```
mrks, line 405: `not_part_of_output_expressions`

src/lib/logical_query_plan/join_node.cpp
julianmenzler: Yes, it is necessary to fulfill an Assert in `fds_from_unique_constraints()`.

But, I found a better solution:

```cpp
if (valid_unique_constraints->empty()) {
    // All unique constraints become discarded, so we have to manually forward all FDs from both input nodes.
    fds_left = left_input()->functional_dependencies();
    fds_right = right_input()->functional_dependencies();
  } else if (valid_unique_constraints == right_unique_constraints) {
    // Left input node's unique constraints become discarded
    fds_left = left_input()->functional_dependencies();
    fds_right = right_input()->non_trivial_functional_dependencies();
  } else if (valid_unique_constraints == left_unique_constraints) {
    // Right input node's unique constraints become discarded
    fds_left = left_input()->non_trivial_functional_dependencies();
    fds_right = right_input()->functional_dependencies();
  } else {
    // All unique constraints from input nodes remain valid. Therefore, we only have to forward the non-trivial FDs.
    DebugAssert(
        valid_unique_constraints->size() == (left_unique_constraints->size() + right_unique_constraints->size()),
        "Unexpected number of unique constraints.");
    fds_left = left_input()->non_trivial_functional_dependencies();
    fds_right = right_input()->non_trivial_functional_dependencies();
  }
```

Looks more intuitive to me.

src/lib/logical_query_plan/join_node.hpp
julianmenzler: > Can you clarify "that stay valid"? Do you mean that they are not nullable?

Yes. They become invalid as soon as one determinant expression becomes nullable. Is it necessary to have this level of detail in the function description?

> Also, turn the second item into "Previously trivial FDs turn into non-trivial FDs as the determinant columns become non-unique, but the FDs still hold true"?

Not exactly: If any determinant column becomes non-nullable, the FD is discarded. The uniqueness of determinant expressions does not impact the validity of FDs. 

What about the following?

```cpp
 /**
   * (a) Semi- & Anti-Joins: 
   *      - Forwards left input node's non-trivial FDs
   * (b) Cross-Joins: 
   *      - Forwards non-trivial FDs from both input nodes.
   * (c) Inner-/Outer-Joins:
   *      - Forwards non-trivial FDs from both input nodes whose determinant expressions stay non-nullable.
   *      - Returns derived, trivial FDs from the left and/or right input node as non-trivial FDs if associated unique 
   *        constraints become discarded.
   */
  std::vector<FunctionalDependency> non_trivial_functional_dependencies() const override;
```

mrks: Sounds good. Minor suggestion for the last point:

Turns derived, trivial FDs from the left and/or right input node into non-trivial FDs if the underlying unique constraints do not survive the join.

src/test/lib/logical_query_plan/join_node_test.cpp
julianmenzler: Legit point. I split it into four tests.

julianmenzler:
> Just final naming / phrasing suggestions. Did you check if the TPC-H performance is affected on low scale factors?

Great. Here are the numbers for `./hyriseBenchmarkTPCH --scheduler -s 0.01`:

```diff
 +----------------+--------------------+------+--------------------+------+------------+----------------------+
 | Benchmark      | prev. iter/s       | runs | new iter/s         | runs | change [%] |              p-value |
 +----------------+--------------------+------+--------------------+------+------------+----------------------+
 | TPC-H 01       | 45.49365234375     | 2730 | 44.96186828613281  | 2698 | -1%        | (run time too short) |
 | TPC-H 02       | 70.26488494873047  | 4216 | 69.64547729492188  | 4179 | -1%        | (run time too short) |
 | TPC-H 03       | 96.57759094238281  | 5795 | 94.02224731445312  | 5642 | -3%        | (run time too short) |
 | TPC-H 04       | 99.20668029785156  | 5953 | 96.92048645019531  | 5816 | -2%        | (run time too short) |
 | TPC-H 05       | 79.4949722290039   | 4770 | 79.5609130859375   | 4774 | +0%        | (run time too short) |
 | TPC-H 06       | 99.33406829833984  | 5961 | 98.10153198242188  | 5887 | -1%        | (run time too short) |
 | TPC-H 07       | 46.28016662597656  | 2777 | 47.9783821105957   | 2879 | +4%        | (run time too short) |
+| TPC-H 08       | 17.064729690551758 | 1024 | 22.54984474182129  | 1353 | +32%       | (run time too short) |
 | TPC-H 09       | 47.380069732666016 | 2843 | 47.13097381591797  | 2828 | -1%        | (run time too short) |
 | TPC-H 10       | 76.073974609375    | 4565 | 76.43897247314453  | 4587 | +0%        | (run time too short) |
 | TPC-H 11       | 98.8437728881836   | 5931 | 98.87598419189453  | 5933 | +0%        | (run time too short) |
 | TPC-H 12       | 95.68460083007812  | 5742 | 95.94143676757812  | 5757 | +0%        | (run time too short) |
+| TPC-H 13       | 58.04072952270508  | 3483 | 63.777408599853516 | 3827 | +10%       | (run time too short) |
 | TPC-H 14       | 99.19422149658203  | 5952 | 99.19347381591797  | 5952 | -0%        | (run time too short) |
 | TPC-H 15       | 98.91523742675781  | 5935 | 99.09671020507812  | 5946 | +0%        | (run time too short) |
 | TPC-H 16       | 50.22915267944336  | 3014 | 50.47877502441406  | 3029 | +0%        | (run time too short) |
 | TPC-H 17       | 97.541259765625    | 5853 | 98.53816223144531  | 5913 | +1%        | (run time too short) |
 | TPC-H 18       | 51.482017517089844 | 3089 | 51.332889556884766 | 3080 | -0%        | (run time too short) |
 | TPC-H 19       | 46.196746826171875 | 2772 | 46.228736877441406 | 2774 | +0%        | (run time too short) |
 | TPC-H 20       | 51.37692642211914  | 3083 | 51.69187545776367  | 3102 | +1%        | (run time too short) |
 | TPC-H 21       | 46.87885665893555  | 2813 | 47.032997131347656 | 2822 | +0%        | (run time too short) |
 | TPC-H 22       | 79.18091583251953  | 4751 | 81.41436004638672  | 4885 | +3%        | (run time too short) |
 | geometric mean |                    |      |                    |      | +2%        |                      |
 +----------------+--------------------+------+--------------------+------+------------+----------------------+
```

src/lib/logical_query_plan/functional_dependency.cpp
julianmenzler: Alright, I added one in JoinNodeTest and called it FunctionalDependenciesMerge.

src/lib/logical_query_plan/functional_dependency.cpp
```diff
@@ -69,8 +69,20 @@ std::vector<FunctionalDependency> merge_fds(const std::vector<FunctionalDependen
   fds_merged.insert(fds_merged.end(), fds_a.begin(), fds_a.end());
 
   for (const auto& fd_to_add : fds_b) {
-    auto existing_fd = std::find_if(fds_merged.begin(), fds_merged.end(),
-                                    [&fd_to_add](auto& fd) { return fd.determinants == fd_to_add.determinants; });
+    auto existing_fd = std::find_if(fds_merged.begin(), fds_merged.end(), [&fd_to_add](auto& fd) {
+      // Cannot use unordered_set::operator== because it ignores the custom equality function.
+      // https://stackoverflow.com/questions/36167764/can-not-compare-stdunorded-set-with-custom-keyequal
```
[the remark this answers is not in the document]
mrks: https://stackoverflow.com/questions/63500137/static-code-analysis-linting-warn-about-being-called-on-unordered-setk-h

mrks (approved):

src/lib/logical_query_plan/functional_dependency.cpp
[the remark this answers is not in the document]
julianmenzler: > Good catch. I hate this behavior so much.
Very painful, indeed.

> https://stackoverflow.com/questions/63500137/static-code-analysis-linting-warn-about-being-called-on-unordered-setk-h

Thanks! I registered myself as a follower.

PR #2195, opened 2020-08-26
JoinOrderingRule shortcut

j-tr:
By skipping the join ordering algorithm if the join graph has just one vertex, the otherwise very expensive JoinOrderingRule becomes neglectable for TPC-C queries resulting in a substatial performance boost.

```+--------------++----------+---------+--------++----------+----------+--------+----------------------+
| Item         || Latency (ms/iter)  | Change || Throughput (iter/s) | Change |              p-value |
|              ||      old |     new |        ||      old |      new |        |                      |
+--------------++----------+---------+--------++----------+----------+--------+----------------------+
| Delivery     ||    139.6 |    62.1 |  -56%  ||     0.92 |     1.57 |  +71%  | (run time too short) |
| New-Order    ||     62.8 |    38.1 |  -39%  ||    10.26 |    17.64 |  +72%  | (run time too short) |
| Order-Status ||      3.2 |     2.7 |  -18%  ||     0.88 |     1.55 |  +76%  | (run time too short) |
| Payment      ||      7.5 |     5.4 |  -27%  ||     9.79 |    16.84 |  +72%  | (run time too short) |
| Stock-Level  ||    165.4 |    85.7 |  -48%  ||     0.92 |     1.57 |  +71%  | (run time too short) |
+--------------++----------+---------+--------++----------+----------+--------+----------------------+
| Sum          ||    378.6 |   194.1 |  -49%  ||          |          |        |                      |
| Geomean      ||          |         |        ||          |          |  +72%  |                      |
+--------------++----------+---------+--------++----------+----------+--------+----------------------+
```

The effect of this shortcut is not really observeable for TPC-H queries.

```+----------++----------+---------+--------++----------+----------+--------+---------+
| Item     || Latency (ms/iter)  | Change || Throughput (iter/s) | Change | p-value |
|          ||      old |     new |        ||      old |      new |        |         |
+----------++----------+---------+--------++----------+----------+--------+---------+
| TPC-H 01 ||   1823.1 |  1772.7 |   -3%  ||     0.55 |     0.56 |   +3%  |  0.1548 |
| TPC-H 02 ||     12.6 |    13.4 |   +6%  ||    79.16 |    74.88 |   -5%  |  0.0000 |
| TPC-H 03 ||    182.3 |   192.1 |   +5%  ||     5.49 |     5.20 |   -5%  |  0.0000 |
| TPC-H 04 ||    158.2 |   158.7 |   +0%  ||     6.32 |     6.30 |   -0%  |  0.4412 |
| TPC-H 05 ||    475.3 |   477.1 |   +0%  ||     2.10 |     2.10 |   -0%  |  0.5140 |
| TPC-H 06 ||      9.6 |     9.7 |   +1%  ||   103.67 |   103.04 |   -1%  |  0.0510 |
| TPC-H 07 ||    131.0 |   132.1 |   +1%  ||     7.63 |     7.57 |   -1%  |  0.4602 |
| TPC-H 08 ||    138.4 |   141.0 |   +2%  ||     7.22 |     7.09 |   -2%  |  0.0050 |
| TPC-H 09 ||   1009.9 |  1004.4 |   -1%  ||     0.99 |     1.00 |   +1%  |  0.2055 |
| TPC-H 10 ||    515.3 |   531.0 |   +3%  ||     1.94 |     1.88 |   -3%  |  0.0053 |
| TPC-H 11 ||     16.5 |    18.2 |  +10%  ||    60.57 |    54.86 |   -9%  |  0.0000 |
| TPC-H 12 ||     93.7 |    94.7 |   +1%  ||    10.67 |    10.56 |   -1%  |  0.0046 |
| TPC-H 13 ||    779.7 |   753.2 |   -3%  ||     1.28 |     1.33 |   +4%  |  0.0000 |
| TPC-H 14 ||     40.4 |    38.0 |   -6%  ||    24.74 |    26.34 |   +6%  |  0.0000 |
| TPC-H 15 ||     18.6 |    17.2 |   -7%  ||    53.82 |    58.09 |   +8%  |  0.0000 |
| TPC-H 16 ||    206.0 |   185.5 |  -10%  ||     4.86 |     5.39 |  +11%  |  0.0000 |
| TPC-H 17 ||     59.4 |    60.6 |   +2%  ||    16.82 |    16.51 |   -2%  |  0.0000 |
| TPC-H 18 ||   1344.2 |  1357.5 |   +1%  ||     0.74 |     0.74 |   -1%  |  0.1089 |
| TPC-H 19 ||     83.4 |    81.8 |   -2%  ||    11.98 |    12.23 |   +2%  |  0.0000 |
| TPC-H 20 ||     36.1 |    33.6 |   -7%  ||    27.69 |    29.73 |   +7%  |  0.0000 |
| TPC-H 21 ||    849.1 |   829.3 |   -2%  ||     1.18 |     1.21 |   +2%  |  0.0108 |
| TPC-H 22 ||    138.0 |   127.3 |   -8%  ||     7.25 |     7.86 |   +8%  |  0.0000 |
+----------++----------+---------+--------++----------+----------+--------+---------+
| Sum      ||   8120.9 |  8029.0 |   -1%  ||          |          |        |         |
| Geomean  ||          |         |        ||          |          |   +1%  |         |
+----------++----------+---------+--------++----------+----------+--------+---------+
```

Repetitive measurement showed that the apparent performance drop of e.g. query 11 might be caused by noise.

```
| TPC-H 11 ||     17.5 |    17.6 |   +1%  ||    57.10 |    56.79 |   -1%  |  0.0285 |
```

### DIFF
--- a/src/lib/optimizer/strategy/join_ordering_rule.cpp
+++ b/src/lib/optimizer/strategy/join_ordering_rule.cpp
@@ -69,7 +69,11 @@ std::shared_ptr<AbstractLQPNode> JoinOrderingRule::_perform_join_ordering_recurs
    */
   // TODO(anybody) Increase X once our costing/cardinality estimation is faster/uses internal caching
   auto result_lqp = std::shared_ptr<AbstractLQPNode>{};
-  if (join_graph->vertices.size() < 9) {
+  DebugAssert(join_graph->vertices.size() != 0, "There should be nodes in the join graph.");
+  if (join_graph->vertices.size() == 1) {
+    // a join graph with only one vertex is no actual join and needs no ordering
+    result_lqp = lqp;
+  } else if (join_graph->vertices.size() < 9) {
     result_lqp = DpCcp{}(*join_graph, caching_cost_estimator);  // NOLINT - doesn't like `{}()`
   } else {
     result_lqp = GreedyOperatorOrdering{}(*join_graph, caching_cost_estimator);  // NOLINT - doesn't like `{}()`
--- a/src/lib/optimizer/strategy/join_ordering_rule.cpp
+++ b/src/lib/optimizer/strategy/join_ordering_rule.cpp
@@ -69,7 +69,7 @@ std::shared_ptr<AbstractLQPNode> JoinOrderingRule::_perform_join_ordering_recurs
    */
   // TODO(anybody) Increase X once our costing/cardinality estimation is faster/uses internal caching
   auto result_lqp = std::shared_ptr<AbstractLQPNode>{};
-  DebugAssert(join_graph->vertices.size() != 0, "There should be nodes in the join graph.");
+  DebugAssert(!join_graph->vertices.empty(), "There should be nodes in the join graph.");
   if (join_graph->vertices.size() == 1) {
     // a join graph with only one vertex is no actual join and needs no ordering
     result_lqp = lqp;
--- a/resources/test_data/tbl/meta_tables/meta_segments_accurate_libcpp_updated.tbl
+++ b/resources/test_data/tbl/meta_tables/meta_segments_accurate_libcpp_updated.tbl
@@ -2,7 +2,7 @@ table_name|chunk_id|column_id|column_name|column_data_type|distinct_value_count|
 string|int|int|string|string|long|string_null|string_null|long|long|long|long|long
 int_int|0|0|a|int|2|null|null|200|3|8|0|0
 int_int|0|1|b|int|2|null|null|200|3|6|0|0
-int_int|1|0|a|int|1|null|null|200|2|3|0|0
+int_int|1|0|a|int|1|null|null|200|1|4|0|0
 int_int|1|1|b|int|1|null|null|200|1|3|0|0
 int_int|2|0|a|int|1|null|null|200|0|1|0|0
 int_int|2|1|b|int|1|null|null|200|0|1|0|0
--- a/resources/test_data/tbl/meta_tables/meta_segments_accurate_libstdcpp_updated.tbl
+++ b/resources/test_data/tbl/meta_tables/meta_segments_accurate_libstdcpp_updated.tbl
@@ -2,7 +2,7 @@ table_name|chunk_id|column_id|column_name|column_data_type|distinct_value_count|
 string|int|int|string|string|long|string_null|string_null|long|long|long|long|long
 int_int|0|0|a|int|2|null|null|192|3|8|0|0
 int_int|0|1|b|int|2|null|null|192|3|6|0|0
-int_int|1|0|a|int|1|null|null|192|2|3|0|0
+int_int|1|0|a|int|1|null|null|192|1|4|0|0
 int_int|1|1|b|int|1|null|null|192|1|3|0|0
 int_int|2|0|a|int|1|null|null|192|0|1|0|0
 int_int|2|1|b|int|1|null|null|192|0|1|0|0
--- a/resources/test_data/tbl/meta_tables/meta_segments_libcpp_updated.tbl
+++ b/resources/test_data/tbl/meta_tables/meta_segments_libcpp_updated.tbl
@@ -2,7 +2,7 @@ table_name|chunk_id|column_id|column_name|column_data_type|encoding_type|vector_
 string|int|int|string|string|string_null|string_null|long|long|long|long|long
 int_int|0|0|a|int|null|null|200|3|6|0|0
 int_int|0|1|b|int|null|null|200|3|4|0|0
-int_int|1|0|a|int|null|null|200|2|2|0|0
+int_int|1|0|a|int|null|null|200|1|3|0|0
 int_int|1|1|b|int|null|null|200|1|2|0|0
 int_int|2|0|a|int|null|null|200|0|0|0|0
 int_int|2|1|b|int|null|null|200|0|0|0|0
--- a/resources/test_data/tbl/meta_tables/meta_segments_libstdcpp_updated.tbl
+++ b/resources/test_data/tbl/meta_tables/meta_segments_libstdcpp_updated.tbl
@@ -2,7 +2,7 @@ table_name|chunk_id|column_id|column_name|column_data_type|encoding_type|vector_
 string|int|int|string|string|string_null|string_null|long|long|long|long|long
 int_int|0|0|a|int|null|null|192|3|6|0|0
 int_int|0|1|b|int|null|null|192|3|4|0|0
-int_int|1|0|a|int|null|null|192|2|2|0|0
+int_int|1|0|a|int|null|null|192|1|3|0|0
 int_int|1|1|b|int|null|null|192|1|2|0|0
 int_int|2|0|a|int|null|null|192|0|0|0|0
 int_int|2|1|b|int|null|null|192|0|0|0|0
--- a/scripts/plot_performance_breakdown.py
+++ b/scripts/plot_performance_breakdown.py
@@ -87,7 +87,7 @@ ax.legend(reversed(handles), reversed(labels), bbox_to_anchor=(1.0, 1.0))
 # Add total runtime to labels
 xlabels = ax.get_xticklabels()
 for label_id, label in enumerate(xlabels):
-    label.set_text(label.get_text() + "\n" + r"$\emptyset$ " + f"{total_time[label_id]/10e6:.2f} ms")
+    label.set_text(label.get_text() + r" $\emptyset$ " + f"{total_time[label_id]/10e6:.2f} ms")
 ax.set_xticklabels(xlabels)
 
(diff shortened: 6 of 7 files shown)


mrks (approved):

Bensk1 (approved):

j-tr:
Optimizer rule execution times for TPC-C:

**Master**
tpcc_master_optimizer_breakdown

**This Branch**
tpcc_new_banch_optimizer_breakdown

Bensk1:
@Bouncner, @j-tr posted the graphics.

Bouncner:
This graph does not show me how much time in spent in optimizing and how much is spent in execution, right? I assume the value in the x axis is the transaction avg. runtime?

Bensk1:
No, you are right. But this graph can also be generated.

j-tr:
@Bouncner

>  I assume the value in the x axis is the transaction avg. runtime?
In the two previous graphs the value was the avg. optimizer runtime

In the following graphs, it's the avg. transaction runtime.

**Master**
tpcc_master_breakdown

**This Branch**
tpcc_new_banch_breakdown

src/lib/optimizer/strategy/join_ordering_rule.cpp
```diff
@@ -69,7 +69,11 @@ std::shared_ptr<AbstractLQPNode> JoinOrderingRule::_perform_join_ordering_recurs
    */
   // TODO(anybody) Increase X once our costing/cardinality estimation is faster/uses internal caching
   auto result_lqp = std::shared_ptr<AbstractLQPNode>{};
-  if (join_graph->vertices.size() < 9) {
+  DebugAssert(join_graph->vertices.size() != 0, "There should be nodes in the join graph.");
+  if (join_graph->vertices.size() == 1) {
```
Bouncner, line 73: I don't know how this whole mechanism works, but what happens if we have 2 vertexes? Isn't that just a simple join that cannot be reordered? So can't we take the short cut `if vertices.size < 3`?

mrks: Not necessarily, as JoinOrdering is also responsible for bringing cross joins and their predicates together. Good point though, we should probably add a comment to that regard.

mrks:
Are you sure the graphs are labeled correctly? It looks like the relative cost of the optimizer grew, when it should have decreased.

julianmenzler (commented):
> Are you sure the graphs are labeled correctly? It looks like the relative cost of the optimizer grew, when it should have decreased.

Considering the average transaction runtime share, they are labeled correctly. The increased optimizer runtime share is weird though.

Here is a GIF-version:
out

mrks:
@j-tr - Could you investigate that, please? As a first step, maybe it helps to see the absolute numbers for the different steps instead of the relative share.

j-tr:
> Are you sure the graphs are labeled correctly? It looks like the relative cost of the optimizer grew, when it should have decreased.

> As a first step, maybe it helps to see the absolute numbers for the different steps instead of the relative share.

Indeed the share of the optimizer is now bigger. When looking at the absolute numbers, we see that the LQPTranslator and the Execution becomes faster as well... so leaving out the join ordering when there is only one vertex apparently isn't the same as applying the join ordering on one vertex.

**Master**
```
      Benchmark        Parser  SQLTranslator     Optimizer  LQPTranslator     Execution
0      Delivery  1.066367e+06   2.626234e+06  5.407624e+07   5.362755e+06  7.094362e+07
1     New-Order  7.834367e+05   1.993030e+06  3.126973e+07   2.353836e+06  2.189347e+07
2  Order-Status  5.958095e+04   1.811801e+05  1.970567e+06   1.968854e+05  5.492647e+05
3       Payment  1.337620e+05   2.913726e+05  4.439625e+06   5.312678e+05  1.497298e+06
4   Stock-Level  1.780292e+05   2.084743e+05  1.649203e+08   1.524784e+05  1.218679e+06
```

**This Branch**
```
      Benchmark        Parser  SQLTranslator     Optimizer  LQPTranslator     Execution
0      Delivery  1.082253e+06   2.599785e+06  3.535663e+07   4.999469e+06  1.282814e+07
1     New-Order  7.876769e+05   1.966207e+06  2.282855e+07   2.226331e+06  6.403790e+06
2  Order-Status  5.890047e+04   1.774259e+05  1.375260e+06   1.922038e+05  5.589882e+05
3       Payment  1.345675e+05   2.887187e+05  3.046870e+06   4.990437e+05  9.535868e+05
4   Stock-Level  1.770227e+05   2.015905e+05  8.277321e+07   1.433496e+05  1.506193e+06
```

When looking at the LQPs, for TPC-H all LQPs are the same with or without joinordering on only one vertex.
For TPC-C however, the LQPs are not the same. This is due to the fact that TPC-C LQPs contain Recurring Nodes and the DpCcp exchanges recurring nodes equivalently in the LQP. E.g.:

```
 [0] [Update] Table: 'ORDER'
  \_[1] [Predicate] .O_W_ID = 1
  |  \_[2] [Predicate] .O_D_ID = 1
  |     \_[3] [Predicate] .O_ID = 2101
  |        \_[4] [Validate]
  |           \_[5] [StoredTable] Name: 'ORDER' pruned: 0/1 chunk(s), 0/8 column(s)
  \_[6] [Projection] .O_ID, .O_D_ID, .O_W_ID, .O_C_ID, .O_ENTRY_D, 6, .O_OL_CNT, .O_ALL_LOCAL
    \_Recurring Node --> [1]
```
 becomes 

```
 [0] [Update] Table: 'ORDER'
  \_[1] [Predicate] .O_W_ID = 1
  |  \_[2] [Predicate] .O_D_ID = 1
  |     \_[3] [Predicate] .O_ID = 2101
  |        \_[4] [Validate]
  |           \_[5] [StoredTable] Name: 'ORDER' pruned: 0/1 chunk(s), 0/8 column(s)
  \_[6] [Projection] .O_ID, .O_D_ID, .O_W_ID, .O_C_ID, .O_ENTRY_D, 6, .O_OL_CNT, .O_ALL_LOCAL
     \_[7] [Predicate] .O_W_ID = 1
       \_[8] [Predicate] .O_D_ID = 1
           \_[9] [Predicate] .O_ID = 2101
              \_Recurring Node --> [4]
```

This change eventually also causes the subsequent PredicateReorderingRule to behave differently and generate another predicate ordering. Im not sure yet if the better performance in the LQPTranslator and Execution comes from the fact that DpCcp always excludes some nodes from the Recurring Node, making less parts of the LQP reusable, or because of the different predicate ordering.

I'm still investigating why and how DpCcp is changing these Recurrent Nodes. Avoiding this could potentially make plans with a DpCcp join ordering more compact and execute faster. Does anyone knows more about our DpCcp implementation in this matter?

So it seems that doing no join ordering on just one vertex doesn't only saves the time for the actual ordering but also avoids unnecessary changes in the plan.

mrks:
Interesting.

> This is due to the fact that TPC-C LQPs contain Recurring Nodes and the DpCcp exchanges recurring nodes equivalently in the LQP

My first intuition is to blame this part:

https://github.com/hyrise/hyrise/blob/master/src/lib/optimizer/strategy/join_ordering_rule.cpp#L85-L88

Different from other rules, this does not use `visit_lqp`. As such, it will treat the left and right input of the `Update` operator independently. This hasn't popped up yet as there are not too many diamond structures at the time the `JoinOrderingRule` is run. Most are added later by other rules. `Update` is one of the few exceptions. This also explains why this hasn't popped up yet.

> This change eventually also causes the subsequent PredicateReorderingRule to behave differently and generate another predicate ordering.

How so? I take it that it blocks the validate node (which is not very selective) from being moved upwards as it is used in more than one subgraph?

> I'm still investigating why and how DpCcp is changing these Recurrent Nodes. Avoiding this could potentially make plans with a DpCcp join ordering more compact and execute faster. Does anyone knows more about our DpCcp implementation in this matter?

Not a DpCcp thing, but a general property of our join ordering rule. I've described part of it above. The other part is that we transform the input into join graphs (see `JoinGraph::build_from_lqp`). These contain tables (nodes) and predicates (edges) and are the base for both join ordering algorithms. The LQP is then reconstructed from these join graphs. In this process, the initial predicate ordering is lost.

In any case, this explains the numbers seen above. We can move the inefficient recursion into its own issue and merge this once CI is green. My guess on the CI failure is that the new (non-duplicated) PQP for Update queries accesses the segments in a different, more efficient way, causing the access counters to return different numbers.

mrks:
Moved the issue discussed above to our long-term list. I don't think that this can occur anywhere outside of UDPATEs right now.

Bensk1:
Quite interesting to see that the runtime improvement was not caused by spending less time in the optimizer (as we initially thought) but by executing a more efficient plan. The new "share of query runtime" chart gives even more motivation to make the optimizer more efficient.

PR #2201, opened 2020-08-27
Fix problems with flake8 linter and E203

### LINKED ISSUE #2200: Problems with Python E203 during formatting and linting
As documented here, E203 is broken. Thus, when `black` correctly changes colons (e.g., in an array slice) flake marks this as an error.

Bouncner:
E203 is broken and it is recommended to ignored it (see https://github.com/psf/black/issues/280).
Currently, `black` correctly adds spaces around colons in some cases and `flake8` marks them as errors due to the broken E203.

This PR extends the ignore list by adding E203 (note, the default settings of `flake8` already ignore several warnings).

Resolves #2200.

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -96,7 +96,7 @@ if [[ ${output} ]]; then
 fi
 
 # Python linting
-output=$(flake8 --max-line-length 120 scripts)
+output=$(flake8 --max-line-length 120 --extend-ignore=E203 scripts)
 if [ ! -z "$output" ]; then
 	echo "$output"
 	exitcode=1


mweisgut (approved):

julianmenzler (commented):
> E203 is broken and it is recommended to ignored it (see psf/black#280).

I wonder why nobody has fixed this yet since it is a rather old issue.

Bouncner:
These are two conflicting styles recommendations. Since `black` adheres to Python's PEP8 "standard", people recommend to disable E203, but I don't think one can say that E203 is "wrong".
They are in fact working on a change (https://github.com/PyCQA/pycodestyle/pull/914) but the discussion is quite long and the PR is months old. Moreover, it doesn't sound as if it is clear how they will eventually solve the issue. So I don't think we should wait for a merge.

julianmenzler:
> These are two conflicting styles recommendations. Since `black` adheres to Python's PEP8 "standard", people recommend to disable E203, but I don't think one can say that E203 is "wrong".
> They are in fact working on a change ([PyCQA/pycodestyle#914](https://github.com/PyCQA/pycodestyle/pull/914)) but the discussion is quite long and the PR is months old. Moreover, it doesn't sound as if it is clear how they will eventually solve the issue. So I don't think we should wait for a merge.

Got it! Thanks for clarifying

PR #2211, opened 2020-09-09
Update sql-parser

mrks:
This updates the sql-parser repo. Nothing more, nothing less.

Changes were caused by external PRs. Having kExprExtract as an own expression type instead of treating it as a function sure makes sense.

### DIFF
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -1717,9 +1717,13 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_translate_hsql_expr(
       Fail("Star expression should have been handled earlier");
 
     case hsql::kExprArrayIndex:
+      FailInput("Array indexes are not yet supported");
+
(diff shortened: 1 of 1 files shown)


src/lib/sql/sql_translator.cpp
```diff
@@ -1713,8 +1717,8 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_translate_hsql_expr(
       Fail("Star expression should have been handled earlier");
 
     case hsql::kExprArrayIndex:
-    case hsql::kExprDatetimeField:
     case hsql::kExprHint:
+    case hsql::kExprCast:
       FailInput("Can't translate this hsql expression into a Hyrise expression");
```
Bouncner, line 1722: Not the most helpful error? Can't we output different failures for `HINT` and `CAST`?

Bouncner (approved):
Just this small comment. Everything else looks good to me.

PR #2215, opened 2020-09-13
Sort : keep nullable property for sorted segments

### LINKED ISSUE #2158: Cached binary tables broken for unencoded data
When you execute `./hyriseBenchmarkTPCH -e Unencoded` twice, the second run fails with:

```
-  Loading table "region" from cached binary "tpch_cached_tables/sf-1.000000/region.bin" (1 ms 72 µs)
-  Loading table "partsupp" from cached binary "tpch_cached_tables/sf-1.000000/partsupp.bin" (4 s 711 ms)
-  Loading table "part" from cached binary "tpch_cached_tables/sf-1.000000/part.bin" (1 s 223 ms)
libc++abi.dylib: terminating with uncaught exception of type std::logic_error: src/lib/import_export/binary/binary_parser.cpp:156 Invalid EncodingType
-  Loading table "orders" from cached binary "tpch_cached_tables/sf-1.000000/orders.bin"Abort trap: 6
```

j-tr:
fixes #2158

When executing `./hyriseBenchmarkTPCH -e Unencoded` twice, the second run failed while loading the cached tables.
This is not due to the binary reader/writer but due to the sorting of the order and lineitem table.
The sort operator kept the nullability property for the columns but always set all segments to be nullable. This missmatch between column and semgent nullability caused the binary reader/writer to mess up.

### DIFF
--- a/src/test/lib/operators/sort_test.cpp
+++ b/src/test/lib/operators/sort_test.cpp
@@ -70,6 +70,36 @@ TEST_P(SortTest, Sort) {
   }
 }
 
+bool column_nullablility_equals_segment_nullability(const std::shared_ptr<Table>& table) {
+  for (auto chunk_id = ChunkID{0}; chunk_id < table->chunk_count(); ++chunk_id) {
+    const auto chunk = table->get_chunk(chunk_id);
+    for (auto column_id = ColumnID{0}; column_id < chunk->column_count(); ++column_id) {
+      auto column_is_nullable = table->column_is_nullable(column_id);
+      auto abstract_segment = chunk->get_segment(column_id);
+      resolve_data_and_segment_type(*abstract_segment, [&](const auto data_type_t, const auto& segment) {
+        using ColumnDataType = typename decltype(data_type_t)::type;
+        using SegmentType = std::decay_t<decltype(segment)>;
+        if constexpr (std::is_same_v<SegmentType, ValueSegment<ColumnDataType>>) {
+          if (segment.is_nullable() != column_is_nullable) {
+            std::cout << "column " << column_id << " nullable " << column_is_nullable << " segment in chunk " << chunk_id << " nullable " << segment.is_nullable() << std::endl;
+          }
+        }
+      });
+    }
+  }
+}
+
+TEST_P(SortTest, UnchangedNullability) {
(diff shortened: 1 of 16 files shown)


j-tr:
out of curiosity, i added a debug assert to the append_chunk method in table.cpp that checks if nullable segments are added if the corresponding column is set non-nullable and found out that the AggregateSort suffers from the same problem as the Sort operator.
so exporting and importing tables that have been a result of an aggregate could potentially fail as well. don't know if thats critical

src/test/lib/operators/sort_test.cpp
```diff
@@ -70,6 +70,40 @@ TEST_P(SortTest, Sort) {
   }
 }
 
+bool column_nullablility_equals_segment_nullability(const std::shared_ptr<const Table>& table) {
```
mrks, line 73: If it's only used for one test, move it into a lambda?

mrks (commented):
Thanks for tracking this down!

> This missmatch between column and semgent nullability caused the binary reader/writer to mess up.

Any chance we can add an assertion there that is more speaking than the current error?

> I added a debug assert to the append_chunk method in table.cpp that checks if nullable segments are added if the corresponding column is set non-nullable

Sounds like an assertion that we should add. Either here or a new issue. I don't think all operators use `append_chunk`, maybe make it part of `AbstractOperator::execute`? The other way (column nullable but segment is not) should still be allowed.

j-tr:
> The other way (column nullable but segment is not) should still be allowed.

in this case, we need to adapt the BinaryParser. Currently, the parser assumes that all segments have the same nullability as the column.

mrks:
Ok, let's not touch this for now and require the segment and column nullability to match. I'd still add a comment explaining that it would be possible to loosen this in the future if we think that it makes sense.

mrks:
Are you planning to look into the assertion, too? Otherwise, please make an issue out of it so that we can merge this PR without losing track of it.

j-tr:
i implemented the assertion but it failed surprisingly often. now investigating how to fix this. moved it to a new issue since this might be a little bit more complicated and think that this PR can be merged

mrks (approved):

PR #2217, opened 2020-09-14
run_benchmarks_over_commit_range.sh: Some Fixes

### LINKED ISSUE #2216: run_benchmark_over_commit_range does not use Ninja
In a ninja build folder, the script tries to call make and does not fail.

mrks:
* Use ninja where ninja is used, fixes #2216 
* Check that the SHA-1 hash is part of the json
* Fix output if incomplete hash or other commit reference (e.g., HEAD^) is used

### DIFF
--- a/scripts/run_benchmarks_over_commit_range.sh
+++ b/scripts/run_benchmarks_over_commit_range.sh
@@ -11,8 +11,8 @@ then
 	exit 1
 fi
 
-start_commit=$1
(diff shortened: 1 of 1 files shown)


mrks:
@Bensk1 ping

Bensk1 (approved):
Haven't tried it and bash makes me feel sick every time I look at it but the changes look good to me.

PR #2218, opened 2020-09-15
Compression plugin change memory budget setting to mega byte

Alexander-Dubrawski:
This Pr changes the memory budget setting in the `CompressionPlugin` from bytes to megabytes. 

## To do:

- [x] take care of linter complaints.


src/plugins/compression_plugin.cpp
```diff
@@ -154,7 +154,8 @@ int64_t CompressionPlugin::_compress_column(const std::string table_name, const
 void CompressionPlugin::_optimize_compression() {
   _keep_alive_stash.clear();
   const auto initial_system_memory_usage = get_all_segments_memory_usage();
-  const auto memory_budget = std::stoll(_memory_budget_setting->get());
+  const auto memory_budget_mb = std::stoll(_memory_budget_setting->get());
+  const auto memory_budget = memory_budget_mb * 1024 * 1024;
```
Bensk1, line 158: Better change that quickly to `* 1000 * 1000` before @Bouncner freaks out.

src/plugins/compression_plugin.hpp
```diff
@@ -30,13 +30,13 @@ class CompressionPlugin : public AbstractPlugin {
...
     const std::string& description() const final {
-      static const auto description = std::string{"The memory budget to target for the CompressionPlugin."};
+      static const auto description = std::string{"The memory budget (MB) to target for the CompressionPlugin."};
       return description;
     }
     const std::string& get() { return _value; }
     void set(const std::string& value) final { _value = value; }
 
-    std::string _value = "9999999999";
+    std::string _value = "9999";
```
Bensk1, line 39: For bytes that looked fine, for MB I would somehow go for `10000`.

src/plugins/compression_plugin.hpp
```diff
@@ -30,13 +30,13 @@ class CompressionPlugin : public AbstractPlugin {
    public:
     explicit MemoryBudgetSetting() : AbstractSetting("Plugin::Compression::MemoryBudget") {}
     const std::string& description() const final {
-      static const auto description = std::string{"The memory budget to target for the CompressionPlugin."};
+      static const auto description = std::string{"The memory budget (MB) to target for the CompressionPlugin."};
```
Bensk1, line 33: How does the frontend determine the description text for the form field? Is it just using the last part of the setting's name:  Plugin::Compression::**MemoryBudge**?

Bensk1 (changes requested):

src/plugins/compression_plugin.hpp
Alexander-Dubrawski: Yes. In the backend, we are sending the SQL query `SELECT name, value, description FROM meta_settings WHERE name LIKE 'Plugin::%';` to the Hyrise. That way we are getting the names of the plugins as well as there value and descriptions. Afterwards, the backend is giving this information to the frontend. 
You can find the method in the backend under: 

https://github.com/hyrise/Cockpit/blob/dev/hyrisecockpit/database_manager/database.py#L286-#L309

Bensk1: Thanks! If I understand it correctly, the setting's name is set to `MemoryBudget`. Can you confirm that this is displayed in the frontend? Ideally we would would display `Memory Budget (MB). The only more or less elegant way I can think of is adding a new field to the settings table: "display name".

Do you know whether the description is displayed in the frontend, maybe when hovering the field?

Alexander-Dubrawski: The Description is displayed in the frontend. If you want to display `Memory Budget (MB)` instead of `Memory Budget` I think to add a new field in the table is a good idea, because then everything stays generic. We wouldn't need to differentiate between different plug-ins in the backend.   

Screenshot 2020-09-15 at 12 20 02

Bensk1: Exactly that was the intention. @Bouncner any thoughts?

Also, I find the leading `?` looks weird. What's your opinion on that?

Bensk1:
@Alexander-Dubrawski do you want to take care of the [linter complaints](https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-2218/2/pipeline)?

Alexander-Dubrawski:
> @Alexander-Dubrawski do you want to take care of the [linter complaints](https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-2218/2/pipeline)?

Yes, I can do that.

src/plugins/compression_plugin.hpp
Alexander-Dubrawski: I think it is weird as well. For me, it would be enough information to just see `Memory Budget (MB)` without a `?`.

Bouncner: If we are already talking about adding a "title shown" field to the meta table, could we also add a description and change the frontend icon to an "i" (I think that's mostly used for some kind of "information")?

Pretty soon we'll need the descriptions anyway I'd guess. So I wouldn't remove it now.

src/plugins/compression_plugin.cpp
Bouncner: (ノಠ益ಠ)ノ彡┻━┻

src/plugins/compression_plugin.hpp
Bensk1: I am not sure if I understand correctly but we already have `description` field in the meta table.

Bouncner: Sorry, I should read the whole thread next time. Yes, the question mark looks strange. I'd vote for some kind of this icon : https://www.flaticon.com/free-icon/info_786024?term=information&page=1&position=3 .

Still, a display name for the setting would be nice.

Bensk1: @Alexander-Dubrawski, do you feel confident adding such a column to the [meta table](https://github.com/hyrise/hyrise/blob/bp1920/src/lib/utils/meta_tables/meta_settings_table.cpp)?

Bensk1: I am also in favor of such an icon. If you use a similar icon elsewhere, we should recycle it. Also, the position of it, before the text, seems strange.

Alexander-Dubrawski: > @Alexander-Dubrawski, do you feel confident adding such a column to the metatable?

Yes. I can try it. I can also change the icon in the frontend.

Bensk1: Perfect.

src/lib/storage/vector_compression/simd_bp128/simd_bp128_decompressor.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include <array>
 #include <memory>
+#include <mutex>
```
Bensk1, line 5: @Bouncner why is that not necessary in the master branch?

Bensk1 (approved):

src/lib/storage/vector_compression/simd_bp128/simd_bp128_decompressor.hpp
Bouncner: The master does not use a lock. There is an issue about potential MT problems, but we decided against a lock a couple of months ago.

PR #2220, opened 2020-09-15
[Hyrise Cockpit] Add display name column to meta settings

Alexander-Dubrawski:
This PR adds the `display name` column to the `meta_settings` table. Also, it sets the display name of the `CompressionPlugin` to `Memory Budget (MB)`. I added the `get_display_name()` method to the `AbstrectSetting` class. It has a default return of `NULL`. A result of that is that the class is not Pure-virtual anymore. So I would be happy about feedback if that implementation is ok.


src/lib/utils/meta_tables/meta_settings_table.cpp
```diff
@@ -8,7 +8,8 @@ namespace opossum {
 MetaSettingsTable::MetaSettingsTable()
     : AbstractMetaTable(TableColumnDefinitions{{"name", DataType::String, false},
                                                {"value", DataType::String, false},
-                                               {"description", DataType::String, false}}) {}
+                                               {"description", DataType::String, false},
+                                               {"display name", DataType::String, false}}) {}
```
Bouncner, line 12: Please don't use spaces in column names.

Bouncner: Can you help me to understand this. Do we insert a new row to settings every time an update is triggered in the frontend? Or is this the place where we register settings?

src/lib/utils/settings/abstract_setting.hpp
```diff
@@ -30,6 +30,8 @@ class AbstractSetting : public Noncopyable, public std::enable_shared_from_this<
 
   virtual const std::string& get() = 0;
 
+  virtual const std::string& get_display_name() {return NULL;}
```
Bouncner, line 33: Either return "" or the previous string we have used. Don't return NULL.

Bouncner (changes requested):

src/lib/utils/settings/abstract_setting.hpp
Bensk1: That was maybe a misunderstanding. By returning `NULL` I meant a SQL NULL, it should be the output of the table in case there is no display name. Empty string would also be fine for me.

src/lib/utils/meta_tables/meta_settings_table.cpp
Bensk1: No, meta tables do not follow the insert-only approach. Instead, updates are handled by the addressed table's `_on_update()` method. In this case the value of the Setting in the SettingsManager is changed.

src/lib/utils/settings/abstract_setting.hpp
Alexander-Dubrawski: Ok. I will change it to an empty string then 🙃 .

Bouncner: Why not default so the string that was used before?

Bensk1: The one used before was determined in the python backend, wasn't it? Also fine for me.

Alexander-Dubrawski: In the backend, we selected the name column in the `meta_setting` tables and used some string operation to get the last part of the name. That isn't a really nice solution. Here is the code:

https://github.com/hyrise/Cockpit/blob/dev/hyrisecockpit/database_manager/database.py#L286-#L309

So should I return the setting name by default? But then the field in the `display_name` column could have the same value as in the `name` column.

Bouncner: We either enforce all plugins to provide a display name (fine with me), or we should have a "somewhat acceptable" default and here I would rather vote for the setting name instead of an empty string. Because empty strings are hard to debug.

So maybe keep it pure virtual and enforce plugins to provide it?

Bensk1: We are still talking about **settings** not **plugins**, right? Do we really need to enforce it? We only need the display name for the frontend and not all settings need to be displayed there.

Alexander-Dubrawski: I would like to keep it pure virtual, but if we do so isn't that all `settings` need to implement this method `display_name`?

Bensk1: If I understand that correctly, yes, if it's a pure virtual method in `AbstractSetting`, all child settings have to implement it.

Bouncner: Yes, but I would consider it a good thing to force Plugin developers to think about an understandable and concise setting name.

Bouncner (commented):
Does this compile? Doesn’t the clustering plugin need the function as well? 
The CI does not build these targets.

Bensk1:
Not yet. I am working on it.

No, ClusteringPlugin does not need it. That's what I meant yesterday: we are talking about Settings here, not Plugins.

Bensk1:
Also, I don't know why but the CI is building these targets.

Bouncner:
Missed that the clustering does not have settings, right.

Bensk1:
JENKINS!

mrks:
Seems to be an issue on rapa. Maybe check if docker is running there.

Bensk1 (approved):

PR #2221, opened 2020-09-16
Aggregate: Write ANY as part of grouping, not as pseudo aggregate function

mrks:
Currently we treat ANY as a regular aggregate function, which means that we look at each value individually. By writing it as if it was a group-by column (but not using it for grouping), we gain quite a bit:

```diff
 +Configuration Overview----+------------------------------------------------+------------------------------------------------+
 | Parameter                | /home/Markus.Dreseler/master_node1.json        | agg.json                                       |
 +--------------------------+------------------------------------------------+------------------------------------------------+
 |  GIT-HASH                | ae811a4f022d58a54c40c8b49eeb2146fb01e9db-dirty | e228c01d829714a3027a484a903c8bbba7550023-dirty |
 |  benchmark_mode          | Ordered                                        | Ordered                                        |
 |  build_type              | release                                        | release                                        |
 |  chunk_size              | 65535                                          | 65535                                          |
 |  clients                 | 1                                              | 1                                              |
 |  compiler                | gcc 9.2                                        | gcc 9.2                                        |
 |  cores                   | 0                                              | 0                                              |
 |  date                    | 2020-09-16 13:04:38                            | 2020-09-16 12:15:33                            |
 |  encoding                | {'default': {'encoding': 'Dictionary'}}        | {'default': {'encoding': 'Dictionary'}}        |
 |  indexes                 | False                                          | False                                          |
 |  max_duration            | 60000000000                                    | 60000000000                                    |
 |  max_runs                | -1                                             | -1                                             |
 |  scale_factor            | 1.0                                            | 1.0                                            |
 |  time_unit               | ns                                             | ns                                             |
 |  use_prepared_statements | False                                          | False                                          |
 |  using_scheduler         | False                                          | False                                          |
 |  verify                  | False                                          | False                                          |
 |  warmup_duration         | 0                                              | 0                                              |
 +--------------------------+------------------------------------------------+------------------------------------------------+

 +----------++----------+---------+--------++----------+----------+--------+---------+
 | Item     || Latency (ms/iter)  | Change || Throughput (iter/s) | Change | p-value |
 |          ||      old |     new |        ||      old |      new |        |         |
 +----------++----------+---------+--------++----------+----------+--------+---------+
 | TPC-H 01 ||    822.9 |   822.4 |   -0%  ||     1.22 |     1.22 |   +0%  |  0.9315 |
 | TPC-H 02 ||      4.5 |     4.6 |   +2%  ||   220.20 |   215.56 |   -2%  |  0.0069 |
 | TPC-H 03 ||     80.9 |    81.3 |   +1%  ||    12.36 |    12.29 |   -1%  |  0.0001 |
 | TPC-H 04 ||     65.3 |    65.7 |   +1%  ||    15.32 |    15.22 |   -1%  |  0.0000 |
 | TPC-H 05 ||    222.1 |   225.6 |   +2%  ||     4.50 |     4.43 |   -2%  |  0.0003 |
 | TPC-H 06 ||      3.2 |     3.2 |   +0%  ||   313.27 |   312.78 |   -0%  |  0.1279 |
 | TPC-H 07 ||     56.8 |    57.1 |   +0%  ||    17.59 |    17.53 |   -0%  |  0.7006 |
 | TPC-H 08 ||     72.0 |    72.9 |   +1%  ||    13.90 |    13.71 |   -1%  |  0.0063 |
 | TPC-H 09 ||    502.3 |   510.8 |   +2%  ||     1.99 |     1.96 |   -2%  |  0.0000 |
+| TPC-H 10 ||    252.7 |   164.9 |  -35%  ||     3.96 |     6.06 |  +53%  |  0.0000 |
 | TPC-H 11 ||      8.2 |     8.2 |   +1%  ||   122.33 |   121.29 |   -1%  |  0.0000 |
 | TPC-H 12 ||     42.2 |    43.0 |   +2%  ||    23.71 |    23.26 |   -2%  |  0.0000 |
 | TPC-H 13 ||    351.1 |   357.7 |   +2%  ||     2.85 |     2.80 |   -2%  |  0.0000 |
 | TPC-H 14 ||     21.0 |    21.3 |   +1%  ||    47.63 |    46.96 |   -1%  |  0.0000 |
 | TPC-H 15 ||      7.7 |     7.8 |   +2%  ||   130.42 |   127.50 |   -2%  |  0.0000 |
 | TPC-H 16 ||     89.9 |    91.3 |   +2%  ||    11.12 |    10.95 |   -2%  |  0.0000 |
 | TPC-H 17 ||     16.3 |    16.5 |   +1%  ||    61.21 |    60.62 |   -1%  |  0.0000 |
 | TPC-H 18 ||    726.7 |   727.7 |   +0%  ||     1.38 |     1.37 |   -0%  |  0.8958 |
 | TPC-H 19 ||     27.8 |    28.2 |   +2%  ||    36.02 |    35.47 |   -2%  |  0.0000 |
 | TPC-H 20 ||     15.6 |    15.6 |   +0%  ||    63.97 |    63.92 |   -0%  |  0.7599 |
 | TPC-H 21 ||    410.2 |   413.3 |   +1%  ||     2.44 |     2.42 |   -1%  |  0.1112 |
 | TPC-H 22 ||     49.5 |    49.7 |   +0%  ||    20.22 |    20.13 |   -0%  |  0.0000 |
 +----------++----------+---------+--------++----------+----------+--------+---------+
 | Sum      ||   3848.8 |  3789.0 |   -2%  ||          |          |        |         |
 | Geomean  ||          |         |        ||          |          |   +1%  |         |
 +----------++----------+---------+--------++----------+----------+--------+---------+
```

### DIFF
--- a/src/lib/operators/aggregate_hash.cpp
+++ b/src/lib/operators/aggregate_hash.cpp
@@ -617,8 +617,8 @@ void AggregateHash::_aggregate() {
                   chunk_id, aggregate_idx, *abstract_segment, keys_per_chunk);
               break;
             case AggregateFunction::Any:
-              _aggregate_segment<ColumnDataType, AggregateFunction::Any, AggregateKey>(
-                  chunk_id, aggregate_idx, *abstract_segment, keys_per_chunk);
+              //  ANY is a pseudo-function and is handled by _write_groupby_output
+              break;
           }
         });
 
@@ -651,8 +651,9 @@ std::shared_ptr<const Table> AggregateHash::_on_execute() {
       break;
   }
 
-  auto& step_performance_data = static_cast<OperatorPerformanceData<OperatorSteps>&>(*performance_data);
-  Timer timer;  // _aggregate above has its own, internal timer. Start measuring once _aggregate is done.
+  const auto num_output_columns = _groupby_column_ids.size() + _aggregates.size();
+  _output_column_definitions.resize(num_output_columns);
+  _output_segments.resize(num_output_columns);
 
   /**
    * Write group-by columns.
@@ -672,7 +673,9 @@ std::shared_ptr<const Table> AggregateHash::_on_execute() {
     }
     _write_groupby_output(pos_list);
   }
-  step_performance_data.set_step_runtime(OperatorSteps::GroupByColumnsWriting, timer.lap());
+
+  auto& step_performance_data = static_cast<OperatorPerformanceData<OperatorSteps>&>(*performance_data);
+  Timer timer;  // _aggregate and _write_groupby_output have their own, internal timer. Start measuring once they are done.
 
   /*
   Write the aggregated columns to the output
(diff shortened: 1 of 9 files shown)


src/lib/operators/aggregate_hash.cpp
```diff
@@ -651,18 +656,15 @@ std::shared_ptr<const Table> AggregateHash::_on_execute() {
...
+  _output_segments.resize(num_output_columns);
 
   /**
-   * Write group-by columns.
-   *
-   * 'results_per_column' always contains at least one element, since there are either GroupBy or Aggregate columns.
-   * However, we need to look only at the first element, because the keys for all columns are the same.
-   *
-   * The following loop is used for both, actual GroupBy columns and DISTINCT columns.
+   * If only GROUP BY columns (including ANY pseudo-aggregates) are written, we need to call _write_groupby_output.
```
Bouncner, line 664: What are example of that case?

src/lib/operators/aggregate_hash.cpp
```diff
@@ -816,15 +821,39 @@ write_aggregate_values(pmr_vector<AggregateType>& values, pmr_vector<bool>& null
 }
 
 void AggregateHash::_write_groupby_output(RowIDPosList& pos_list) {
+  auto& step_performance_data = static_cast<OperatorPerformanceData<OperatorSteps>&>(*performance_data);
+  Timer timer;  // _aggregate above has its own, internal timer. Start measuring once _aggregate is done.
+
   auto input_table = left_input_table();
 
+  auto unaggregated_columns = std::vector<std::pair<ColumnID, ColumnID>>{};
```
Bouncner, line 829: Is there a reason not calling these `unaggregate_columns` simply `group_by_columns`?

src/lib/optimizer/optimizer.cpp
```diff
@@ -107,6 +105,11 @@ std::shared_ptr<Optimizer> Optimizer::create_default_optimizer() {
   // SubqueryToJoinRule. As such, we run the JoinOrderingRule before the SubqueryToJoinRule.
   optimizer->add_rule(std::make_unique<JoinOrderingRule>());
 
+  // Run Group-By Reduction after the JoinOrderingRule ran. The actual join order is not important, but the matching
```
Bouncner, line 108: Was that a problem before? How did you catch that?

Bouncner (commented):
Changes look good to me. But I am not an aggregation expert. :-/

src/lib/operators/aggregate_hash.cpp
mrks: Added an example.

src/lib/operators/aggregate_hash.cpp
mrks: Yes. `group_by_columns` is already used for the columns that we actually group on. The fact that the ANY columns used to be part of the group by is nothing that the aggregate operator should care about.

src/lib/optimizer/optimizer.cpp
mrks: > Was that a problem before?

It became a "problem" (rather: a missed optimization opportunity) when @julianmenzler changed the rule to use FDs instead of the unique constraints defined in the StoredTable. Now, we can exploit the FD that comes from the 1:N join. For that, however, the join has to be identified first.

> How did you catch that?

Remembered you mentioning that we couldn't exploit the FD introduced by the join.

mrks: (To clarify: That's nothing that @julianmenzler missed, his changes just allow us to do more than before.)

mrks:
@Bouncner - ping?

Bouncner:
Sorry, missed that one here.

Bouncner (approved):

PR #2228, opened 2020-09-23
nullable Segment implies nullable column

### LINKED ISSUE #2223: non-nullable columns sometimes contain nullable segments
Are you planning to look into the assertion, too? Otherwise, please make an issue out of it so that we can merge this PR without losing track of it.

_Originally posted by @mrks in https://github.com/hyrise/hyrise/pull/2215#issuecomment-692270728_

Nullable segments in non-nullable columns are currently generated by the AggregateSort and -Hash Operator (and maybe more operators).

Adding an Assertion to AbstractOperator::execute() and fixing all the resulting assertion fails would fix this.

j-tr:
fixes #2223
as discussed in #2215

Non-nullable columns should never contain nullable segments, whereas nullable columns can contain nullable and non-nullable segments.

In this PR:

- A debug assert in the abstract operator ensures that this is always the case
- The aggregate operators were adapted since they put nullable segments in non-nullable columns
- The binary Writer/Parser was adapted so that segments do not have to be nullable if the column is nullable
- A little bit unrelated: an export test was added for an existing import test


src/lib/operators/abstract_operator.cpp
```diff
@@ -96,6 +96,23 @@ void AbstractOperator::execute() {
...
+            using SegmentType = std::decay_t<decltype(segment)>;
+            if constexpr (std::is_same_v<SegmentType, ValueSegment<ColumnDataType>>) {
+              // If segment is nullable, the column must be nullable as well
+              Assert(!segment.is_nullable() || _output->column_is_nullable(column_id),
+                     "Nullable segment found in non-nullable column");
+            }
+          });
+        }
+      }
+    }
```
Bensk1, line 115: Do I see it correctly that this loop iterating over all chunks and all columns is executed for every executed operator that has an output?

j-tr: Correct. In debug mode.

Bensk1: Makes sense, didn't see it, never mind.

src/lib/operators/aggregate_sort.cpp
```diff
@@ -554,12 +559,14 @@ std::shared_ptr<const Table> AggregateSort::_on_execute() {
...
   const auto write_groupby_column = [&](const ColumnID input_column_id, const ColumnID output_column_id) {
+    const auto column_is_nullable = _output_column_definitions.at(output_column_id).nullable;
     auto group_boundary_iter = group_boundaries.cbegin();
     auto data_type = input_table->column_data_type(input_column_id);
     resolve_data_type(data_type, [&](auto type) {
       using ColumnDataType = typename decltype(type)::type;
       auto values = pmr_vector<ColumnDataType>(group_boundaries.size() + 1);
-      auto null_values = pmr_vector<bool>(group_boundaries.size() + 1);
+      pmr_vector<bool> null_values;
+      if (column_is_nullable) null_values = pmr_vector<bool>(group_boundaries.size() + 1);
```
mrks, line 569: Personally, I think that a ternary operator is slightly easier to read, but that might be just my opinion. No need to change.

src/lib/operators/aggregate_sort.cpp
```diff
@@ -583,16 +590,24 @@ std::shared_ptr<const Table> AggregateSort::_on_execute() {
...
+        if (column_is_nullable) {
+          null_values[value_index] = variant_is_null(value);
+          if (!null_values[value_index]) {
+            // Only store non-null values
+            values[value_index] = boost::get<ColumnDataType>(value);
+          }
+        } else {
           values[value_index] = boost::get<ColumnDataType>(value);
         }
       }
```
mrks, line 602: Why make it this complicated? The vector has been initialized to false anyway.

```
if (column_is_nullable && variant_is_null(value)) {
  null_values[...] = true;
} else {
  values[...] = ...;
}
```

Am I missing a case?

mrks (commented):
Hmm, this PR really brings to light that NULL values were added later on. ;)

src/lib/import_export/binary/binary_writer.cpp
```diff
@@ -138,16 +138,28 @@ void BinaryWriter::_write_chunk(const Table& table, std::ofstream& ofstream, con
...
-        [&](const auto data_type_t, const auto& resolved_segment) { _write_segment(resolved_segment, ofstream); });
+    resolve_data_and_segment_type(*chunk->get_segment(column_id),
+                                  [&](const auto data_type_t, const auto& resolved_segment) {
+                                    using ColumnDataType = typename decltype(data_type_t)::type;
+                                    using SegmentType = std::decay_t<decltype(resolved_segment)>;
+                                    if constexpr (std::is_same_v<SegmentType, ValueSegment<ColumnDataType>>) {
+                                      _write_segment(resolved_segment, table.column_is_nullable(column_id), ofstream);
+                                    } else {
+                                      _write_segment(resolved_segment, ofstream);
+                                    }
```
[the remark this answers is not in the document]
j-tr: i now added the column_is_nullable to all the write_segment calls so that we don't need a distinction between segments here. i didn't liked having the column_is_nullable for other segments than value segment because it is not used for the export. but since this part is not performance relevant, i think we can go for simplicity here.

src/lib/operators/aggregate_sort.cpp
j-tr: what would be the else part of the ternary operator in this case? am i missing something?

src/lib/operators/aggregate_sort.cpp
j-tr: we can even drop the column_is_nullable check, since this PR makes sure that variant_is_null can only be true if column_is_nullable ;)

src/lib/operators/abstract_operator.cpp
```diff
@@ -96,6 +96,23 @@ void AbstractOperator::execute() {
...
     }
+    // Verify that nullability of columns and segments match for value segments
+    if (_output) {
+      for (auto chunk_id = ChunkID{0}; chunk_id < _output->chunk_count(); ++chunk_id) {
+        for (auto column_id = ColumnID{0}; column_id < _output->column_count(); ++column_id) {
+          const auto& abstract_segment = _output->get_chunk(chunk_id)->get_segment(column_id);
+          resolve_data_and_segment_type(*abstract_segment, [&](const auto data_type_t, const auto& segment) {
+            using ColumnDataType = typename decltype(data_type_t)::type;
+            using SegmentType = std::decay_t<decltype(segment)>;
+            if constexpr (std::is_same_v<SegmentType, ValueSegment<ColumnDataType>>) {
```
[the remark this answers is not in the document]
j-tr: doc at the beginning of the logical block

src/lib/operators/aggregate_sort.cpp
mrks: `auto null_values = pmr_vector<bool>(column_is_nullable ? group_boundaries.size() + 1 : 0);`

mrks (approved):

PR #2229, opened 2020-09-24
Fix segfault when requesting vector compression of LZ4 segment of an empty string column

Bouncner:
For empty string columns (not the case for TPC-H, but for TPC-DS), LZ4 does not store the string vector offsets as every query is of length 0.
When requesting the vector compression type, a check was missing and Hyrise crashed when the vector compression type was requested for an empty string column.

I thought about testing it, but empty string columns are tested and would be needed is now testing the return value of the compressed vector type. I am not sure if that test makes sense.
In case anybody likes to see a test, I'll add one.

### DIFF
--- a/src/lib/storage/lz4_segment.cpp
+++ b/src/lib/storage/lz4_segment.cpp
@@ -23,7 +23,7 @@ LZ4Segment<T>::LZ4Segment(pmr_vector<pmr_vector<char>>&& lz4_blocks, std::option
       _lz4_blocks{std::move(lz4_blocks)},
       _null_values{std::move(null_values)},
       _dictionary{std::move(dictionary)},
-      _string_offsets{std::nullopt},
+      _string_offsets{nullptr},
       _block_size{block_size},
       _last_block_size{last_block_size},
       _compressed_size{compressed_size},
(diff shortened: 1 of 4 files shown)


src/lib/storage/lz4_segment.cpp
```diff
@@ -488,8 +488,8 @@ std::optional<CompressedVectorType> LZ4Segment<T>::compressed_vector_type() cons
 template <>
 std::optional<CompressedVectorType> LZ4Segment<pmr_string>::compressed_vector_type() const {
   std::optional<CompressedVectorType> type;
-  if (_string_offsets) {
-    resolve_compressed_vector_type(*(*_string_offsets), [&](const auto& vector) { type = vector.type(); });
+  if (_string_offsets && *_string_offsets) {
+    return (*_string_offsets)->type();
```
mrks, line 492: Why is `_string_offsets && !*_string_offsets` a valid case? We had a discussion about [optional<pointer_type>](https://github.com/hyrise/hyrise/pull/2070/files/cb7b16eb700d1bbef5746ee38bfc05f3c547bd00#r402193890) a while ago. To me, the bug fixed in this PR looks like a good example on why that ambiguity causes problems.

Bouncner: Good point. Tbh, I haven't thought about to much and just cared about the segfault.

Personally, I'd remove the unique pointer as this vector is a member of the segment and shouldn't be shared or passed somewhere else. So ownership should be clear.
What do you think about `std::optional<std::vector__offsets>`?

Bouncner: Ah, it's a base pointer. Doesn't work. 

I am wondering if I should delete the PR altogether, because I need to rework this part anyways. But who knows when this will be done. :(

What's your opinion on a unique ptr member which defaults to a nullptr. I don't like it, because I am afraid of pointers. They keep me awake at night.

mrks: Yes. I haven't looked at the code around this, but judging from the PR, a unique_ptr that might be null would be cleaner.

Bouncner: Changed it. I agree that it's better this way. Even though I still hate pointers. :D

mrks (approved):

PR #2241, opened 2020-10-13
JSON serializer for Hyrise expressions

CAJan93:
**Disclaimer:**

In the context of this PR the word "serialize" referees both to serialization and deserialization unless explicitly stated otherwise.

**Overview:**
This PR introduces a JSON serializer for Hyrise expressions. Serialization is needed for [Skyrise](https://github.com/hpi-epic/skyrise), which can only communicate with its workers via JSON. 

**Goals:**

- Keep the Hyrise expression intact

We are trying to avoid any unnecessary change to the existing Hyrise expressions so that Skyrise can benefit from the existing expression processing done in Hyrise. However, all classes used in the PQP have to contain some meta-information about which members to serialize. We further want to move all Hyrise Expression tests to Skyrise.

- Full TPC-H coverage

The goal of this PR is to cover all TPC-H queries. 

- Testing

We aim to include the serializer in the benchmark suite of Hyrise. We assume that our serializer has a decent quality if we are able to run all TPC-H queries via the serializer. Like so: `for (query : tpc_h_queries) run_query(deserialize(serialize(query)))`. 

**Limitations:**

- We do not store any information regarding the LQP. 

Information about the LQP is unnecessary for Skyrise. 

- The serializer is domain-specific. 

A domain-independent serializer would be possible, but that would require more changes to the Hyrise expressions, which we are trying to avoid. Because of this, the serializer has to know how to construct a node during deserialization. 

**TODO:**

- [ ] Support TPC-H Q6 
- [ ] Split up serializer class in serializer and deserializer
- [ ] Move `as_any` and `with_any` to separate files
- [ ] Check if Hyrise can execute `deserialize(serialize(Q6))`


Bouncner:
Why does Hyrise need adaptions for another database that is (to my knowledge) completely independent and has no connections to Hyrise? I cannot follow the general idea of this PR.

Is this PR meant to be merged in Hyrise at some point in time or is this for discussion?

tobodner:
> Why does Hyrise need adaptions for another database that is (to my knowledge) completely independent and has no connections to Hyrise? I cannot follow the general idea of this PR.
> 
> Is this PR meant to be merged in Hyrise at some point in time or is this for discussion?

@CAJan93 accidentally pushed this here. It was meant to go against a Hyrise fork that we use to experiment with designs.

CAJan93:
@tobodner Confirmed. The new PR is at https://github.com/hpi-epic/hyrise/pull/1.

tobodner:
> We should not have another repository with the name “hyrise” in one of the epic groups. Could you guys rename that one please?

I just kept the name for the fork. Any suggestions if you do not like it? hyrise-fork? Or would you rather have such design discussions here (I did not assume so)?

Bouncner:
Does it need to be public?

Bouncner:
Let's discuss in the next roundtable (shouldn't matter that there might not be one for a couple of weeks).

tobodner:
> Does it need to be public?

There is no direct way to turn forks of public repos private. Will move the repo to my account.

PR #2245, opened 2020-10-16
Make hyriseBenchmarkX loading faster

mrks:
* For the generation of the table histograms, use the robin_map. It allows us to cache the hash of strings, which makes generation much faster. We will likely replace the current bytell hash map with this one for other things, too.
* Parallelize some steps of the benchmark start. Hide whitespaces when reviewing this.

Old:
- Sorting tables done (5 s 957 ms)
- Encoding tables and generating pruning statistic done (10 s 692 ms)
- Adding tables to StorageManager and generating table statistics done (2 min 41 s)
- Total time from binary start to first query: 3 min 21 s

New:
- Sorting tables done (5 s 238 ms)
- Encoding tables and generating pruning statistic done (8 s 778 ms)
- Adding tables to StorageManager and generating table statistics done (1 min 23 s)
- Total time from binary start to first query: 1 min 50 s

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -47,3 +47,6 @@
 [submodule "third_party/zstd"]
 	path = third_party/zstd
 	url = https://github.com/facebook/zstd.git
+[submodule "third_party/robin-map"]
+	path = third_party/robin-map
+	url = https://github.com/Tessil/robin-map.git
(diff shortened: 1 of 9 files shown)


dey4ss (approved):
Looks good, I see nothing suspicious if the captures work with clang.

mrks:
Thanks for the quick review.

Clang is right, this is a [defect](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#2313) in C++.

Issue #1254, opened 2018-11-01
Consider using final override


mrzzzrm:
Tim & me just noticed that, e.g., the Histograms might profit from declaring virtual functions as `override final`. Maybe we should have a pass over the codebase, drop in those declarations and see whether it makes a difference.

lanice:
Profit in what way?

mrks:
Do you have any numbers on this? I expect this to only be significant if we call a virtual method on a class where we already knew the type.

Also, is marking the class final equivalent?

mrzzzrm:
> expect this to only be significant if we call a virtual method on a class where we already knew the type

precisely

> Also, is marking the class final equivalent?

I would think so

No numbers yet, that's why i am opening this issue

mrks:
If we go around tagging methods to see how it affects performance, should we hand out a couple of `nothrow`s as well? While we do throw exceptions, we almost never handle them, so `std::terminate` would be appropriate. I wouldn't do this for one or two percent of performance, but maybe it's more?

mrks:
Quite relevant: https://quuxplusone.github.io/blog/2021/02/15/devirtualization/

PR #2248, opened 2020-10-21
Add support for clang-11 and gcc-10

mrks:
To enable `brew upgrade`, I fixed the issues that came up with clang-11. Nothing big, details can be found in the comments.

This does not yet add gcc-10/clang-11 to any dependencies. It also does not fix any issues that clang-tidy-11 might complain about.

This was tested on OS X 10.15.6 and Ubuntu 20.10 dev.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/google/benchmark.git
 [submodule "third_party/cpp-btree"]
 	path = third_party/cpp-btree
-	url = https://github.com/dmeister/cpp-btree
+	url = https://github.com/mrks/cpp-btree
 [submodule "third_party/cpplint"]
 	path = third_party/cpplint
 	url = https://github.com/cpplint/cpplint.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/google/benchmark.git
 [submodule "third_party/cpp-btree"]
 	path = third_party/cpp-btree
-	url = https://github.com/algorithm-ninja/cpp-btree
+	url = https://github.com/dmeister/cpp-btree
 [submodule "third_party/cpplint"]
 	path = third_party/cpplint
 	url = https://github.com/cpplint/cpplint.git
--- a/src/benchmark/tpch_data_micro_benchmark.cpp
+++ b/src/benchmark/tpch_data_micro_benchmark.cpp
@@ -29,7 +29,7 @@ class TableWrapper;
 // Defining the base fixture class
 class TPCHDataMicroBenchmarkFixture : public MicroBenchmarkBasicFixture {
  public:
-  void SetUp(::benchmark::State& state) {
+  void SetUp(::benchmark::State& state) override {
     auto& sm = Hyrise::get().storage_manager;
     const auto scale_factor = 0.01f;
     const auto default_encoding = EncodingType::Dictionary;
@@ -87,7 +87,7 @@ class TPCHDataMicroBenchmarkFixture : public MicroBenchmarkBasicFixture {
   }
 
   // Required to avoid resetting of StorageManager in MicroBenchmarkBasicFixture::TearDown()
-  void TearDown(::benchmark::State&) {}
+  void TearDown(::benchmark::State&) override {}
 
   std::map<std::string, std::shared_ptr<TableWrapper>> create_table_wrappers(StorageManager& sm) {
     std::map<std::string, std::shared_ptr<TableWrapper>> wrapper_map;
--- a/src/lib/cache/gdfs_cache.hpp
+++ b/src/lib/cache/gdfs_cache.hpp
@@ -35,7 +35,7 @@ class GDFSCache : public AbstractCache<Key, Value> {
 
   explicit GDFSCache(size_t capacity = DEFAULT_CACHE_CAPACITY) : AbstractCache<Key, Value>(capacity), _inflation(0.0) {}
 
-  void set(const Key& key, const Value& value, double cost = 1.0, double size = 1.0) {
+  void set(const Key& key, const Value& value, double cost = 1.0, double size = 1.0) final {
     std::unique_lock<std::shared_mutex> lock(_mutex);
     if (this->_capacity == 0) return;
     auto it = _map.find(key);
@@ -66,7 +66,7 @@ class GDFSCache : public AbstractCache<Key, Value> {
     _map[key] = handle;
   }
 
-  std::optional<Value> try_get(const Key& query) {
+  std::optional<Value> try_get(const Key& query) final {
     std::unique_lock<std::shared_mutex> lock(_mutex);
     auto it = _map.find(query);
     if (it == _map.end()) return std::nullopt;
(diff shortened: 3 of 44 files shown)


.gitmodules
```diff
@@ -6,7 +6,7 @@
 	url = https://github.com/google/benchmark.git
 [submodule "third_party/cpp-btree"]
 	path = third_party/cpp-btree
-	url = https://github.com/algorithm-ninja/cpp-btree
+	url = https://github.com/mrks/cpp-btree
```
mrks, line 9: The original repo is outdated. My fork is based on https://github.com/dmeister/cpp-btree and fixes the use of allocator::rebind, which was deprecated in C++17 and removed in C++20.

src/benchmark/tpch_data_micro_benchmark.cpp
```diff
@@ -29,7 +29,7 @@ class TableWrapper;
 // Defining the base fixture class
 class TPCHDataMicroBenchmarkFixture : public MicroBenchmarkBasicFixture {
  public:
-  void SetUp(::benchmark::State& state) {
+  void SetUp(::benchmark::State& state) override {
```
mrks, line 32: clang-11 finally complains about virtual overrides not marked as either override or final :)

src/lib/cache/gdfs_cache.hpp
```diff
@@ -35,7 +35,7 @@ class GDFSCache : public AbstractCache<Key, Value> {
 
   explicit GDFSCache(size_t capacity = DEFAULT_CACHE_CAPACITY) : AbstractCache<Key, Value>(capacity), _inflation(0.0) {}
 
-  void set(const Key& key, const Value& value, double cost = 1.0, double size = 1.0) {
+  void set(const Key& key, const Value& value, double cost = 1.0, double size = 1.0) final {
```
mrks, line 38: No real pattern in whether I use override or final. I just went with what was predominant in the files.

#1254

src/lib/operators/abstract_join_operator.hpp
```diff
@@ -58,14 +58,6 @@ class AbstractJoinOperator : public AbstractReadOnlyOperator {
   std::shared_ptr<Table> _build_output_table(std::vector<std::shared_ptr<Chunk>>&& chunks,
                                              const TableType table_type = TableType::References) const;
 
-  // Some operators need an internal implementation class, mostly in cases where
-  // their execute method depends on a template parameter. An example for this is
-  // found in join_hash.hpp.
-  class AbstractJoinOperatorImpl : public AbstractReadOnlyOperatorImpl {
-   public:
-    virtual ~AbstractJoinOperatorImpl() = default;
-    virtual std::shared_ptr<const Table> _on_execute() = 0;
```
mrks, line 67: Completely pointless.

src/lib/storage/table_key_constraint.cpp
```diff
@@ -8,8 +8,6 @@ TableKeyConstraint::TableKeyConstraint(std::unordered_set<ColumnID> init_columns
 KeyConstraintType TableKeyConstraint::key_type() const { return _key_type; }
 
 bool TableKeyConstraint::_on_equals(const AbstractTableConstraint& table_constraint) const {
-  DebugAssert(dynamic_cast<const AbstractTableConstraint*>(&table_constraint),
```
mrks, line 11: dynamic_cast to the identical types doesn't check anything

src/lib/types.hpp
```diff
@@ -10,6 +10,11 @@
 #include <tuple>
 #include <vector>
 
+#include <boost/version.hpp>
+#if BOOST_VERSION < 107400                      // TODO(anyone): remove this block once Ubuntu ships boost 1.74
+#include "utils/boost_bimap_core_override.hpp"  // NOLINT
```
mrks, line 15: Again, the ancient boost version in Ubuntu becomes an issue. Even groovy (20.10) will come with 1.71. Use the same hack as we already successfully use with boost_curry_override.hpp.

mrks:
@julianmenzler Not as bad as it looks like. Almost 500 of the modified lines are just copied from boost. Most of this can just be scrolled over.

src/lib/types.hpp
julianmenzler: 1. `#if BOOST_VERSION < 107400`  Call: `#include "utils/boost_bimap_core_override.hpp"`
2. Call: `#include <boost/bimap.hpp>` (l. 18)

Looks redundant to me. Shouldn't we put `#include <boost/bimap.hpp>` into an `#else`-clause?

julianmenzler (changes requested):

src/lib/cache/gdfs_cache.hpp
mrks: It does. Non-virtual functions are final by definition. `virtual final` does not make sense as you would declare your method virtual on the top-level, but not allow anyone to override it. Thus, whenever you mark something as final, it is clear that you are overriding something.

I was trying to find a source for this, and found this, which seems to be wrong: https://stackoverflow.com/questions/29412412/does-final-imply-override

Clang complains about final being used on something that does not override a virtual method:

https://godbolt.org/z/MarY1G

src/lib/storage/table_key_constraint.cpp
mrks: Ah, should have read the line below. :/

src/lib/types.hpp
mrks: `bimap_core` is something different than `bimap`. `bimap.hpp` would usually include `bimap_core. By including it ourselves before `bimap.hpp` had a chance to do it, the backported version gets loaded and sets its header guards. As such, when `bimap.hpp` include the `bimap_core.hpp` that comes with the system's boost version, the header guard is already set, and the file will be ignored.

Fun fact: This hack would not be possible with `#pragma once`. Not sure if that is a good or a bad thing though ;)

src/lib/storage/index/group_key/composite_group_key_index.hpp
```diff
@@ -51,7 +51,6 @@ class CompositeGroupKeyIndex : public AbstractIndex {
   static size_t estimate_memory_consumption(ChunkOffset row_count, ChunkOffset distinct_count, uint32_t value_bytes);
 
   CompositeGroupKeyIndex(CompositeGroupKeyIndex&&) = default;
-  ~CompositeGroupKeyIndex() = default;
```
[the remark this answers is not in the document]
mrks: Destructor. Removed because it is already defined in AbstractIndex and we are not adding anything here.

src/test/lib/operators/difference_test.cpp
```diff
@@ -19,7 +19,7 @@ using namespace opossum::expression_functional;  // NOLINT
 namespace opossum {
 class OperatorsDifferenceTest : public BaseTest {
  protected:
-  virtual void SetUp() {
```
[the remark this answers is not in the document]
mrks: Yepp, I like that this is finally enforced.

src/test/lib/sql/sqlite_testrunner/sqlite_testrunner.hpp
```diff
@@ -68,7 +68,7 @@ class SQLiteTestRunner : public BaseTestWithParam<SQLiteTestRunnerParam> {
   inline static bool _last_run_successful{true};
 };
 
-auto sqlite_testrunner_formatter = [](const ::testing::TestParamInfo<SQLiteTestRunnerParam>& info) {
+inline auto sqlite_testrunner_formatter = [](const ::testing::TestParamInfo<SQLiteTestRunnerParam>& info) {
```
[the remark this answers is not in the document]
mrks: This is a hpp file. Without the `inline`, the lambda would get instantiated twice, violating ODR. This is required by `-fno-common`, which is now default. Here's the matching discussion for gcc:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85678

src/lib/cache/gdfs_cache.hpp
julianmenzler: Learned something. Thanks!

julianmenzler (approved):

PR #2255, opened 2020-11-03
Make AbstractExpression Noncopyable

Bouncner:
I was not able to build with Clang 9 and disabled precompiled headers.
Error was:
```
In file included from ../src/lib/logical_query_plan/abstract_lqp_node.hpp:8:
../src/lib/expression/abstract_expression.hpp:50:11: error: definition of implicit copy constructor for 'AbstractExpression' is deprecated because it has a user-declared destructor [-Werror,-Wdeprecated]
  virtual ~AbstractExpression() = default;
```

Not sure what I am doing here, but it fixed the issue. Not sure why this was not caught by our CI.

### DIFF
--- a/src/lib/expression/abstract_expression.hpp
+++ b/src/lib/expression/abstract_expression.hpp
@@ -47,6 +47,7 @@ class AbstractExpression : public std::enable_shared_from_this<AbstractExpressio
  public:
   explicit AbstractExpression(const ExpressionType init_type,
                               const std::vector<std::shared_ptr<AbstractExpression>>& init_arguments);
+  AbstractExpression(const AbstractExpression&) = default;
   virtual ~AbstractExpression() = default;
 
   /**
--- a/src/lib/expression/abstract_expression.hpp
+++ b/src/lib/expression/abstract_expression.hpp
@@ -43,11 +43,10 @@ enum class ExpressionType {
  *
  * Expressions are evaluated (typically for all rows of a Chunk) using the ExpressionEvaluator.
  */
-class AbstractExpression : public std::enable_shared_from_this<AbstractExpression> {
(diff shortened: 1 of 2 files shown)


src/lib/expression/abstract_expression.hpp
```diff
@@ -47,6 +47,7 @@ class AbstractExpression : public std::enable_shared_from_this<AbstractExpressio
  public:
   explicit AbstractExpression(const ExpressionType init_type,
                               const std::vector<std::shared_ptr<AbstractExpression>>& init_arguments);
+  AbstractExpression(const AbstractExpression&) = default;
```
mrks, line 50: In which case would we want to copy an expression? if we actually need a second version, we probably want to use `deep_copy`. Should expressions maybe even be `Noncopyable`?

mrks (changes requested):

src/lib/expression/abstract_expression.hpp
Bouncner: Haven't worked with Expressions a lot. Whatever makes sense.

mrks: Could you paste the compiler error that comes without your change?

Bouncner: That's the one in the description.

mrks:
@Bouncner - does this one work for you, too?

Bouncner:
> @Bouncner - does this one work for you, too?

Builds like a charm. No opinion about the copy-behaviour change though. Good to go from my site.

mrks:
If there is only one place in the code base that needed to be changed, my gut feeling that there is no reason to ever shallow-copy one of those seems to be true.

Disallowing copies by default is just another thing that Rust seems to do better.

mrks (approved):

PR #2257, opened 2020-11-11
plot_operator_breakdown: Remove relative share, fix warning

### LINKED ISSUE #2250: plot_operator_breakdown uses arithmetic mean on percentages
```
df.loc["Relative"] = df.head(-1).sum() / df.head(-1).count()
```

Should be the geomean instead? Not as clear cut as the compare_benchmarks case.

mrks:
fixes #2250

### DIFF
--- a/scripts/plot_operator_breakdown.py
+++ b/scripts/plot_operator_breakdown.py
@@ -13,9 +13,10 @@ import re
 import sys
 import seaborn as sns
 import matplotlib.colors as mplcolors
+import matplotlib.ticker as ticker
 from matplotlib.colors import LinearSegmentedColormap
 
-if len(sys.argv) not in [1, 2]:
+if len(sys.argv) not in [1, 2] or len(glob.glob("*-PQP.svg")) == 0:
     exit("Call in a folder containing *-PQP.svg files, pass `paper` as an argument to change legend and hatching")
 paper_mode = len(sys.argv) == 2 and sys.argv[1] == "paper"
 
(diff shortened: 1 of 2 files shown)


Bouncner:
`ax.set_yticklabels(["{:,.0%}".format(x) for x in ax.get_yticks()])`

Since the removed line looks reasonable to me (it's percentages, right?): could you add a screenshot of the change?

scripts/plot_operator_breakdown.py
```diff
@@ -117,7 +118,7 @@
         bar.set_linewidth(0)
 
 # Set labels
-ax.set_yticklabels(["{:,.0%}".format(x) for x in ax.get_yticks()])
```
Bouncner, line 120: ax.set_yticklabels(["{:,.0%}".format(x) for x in ax.get_yticks()])

mrks: ?

Bouncner: Copy/paste fail ...

I wanted to ask for a screenshot as the removed line looks pretty good to me (it's percent and that's what the formatting does, right?).

mrks: Does the same thing, but if the yticks ever change (which they don't), the formatting would be broken. That's why we were getting a warning there. Also, using the PercentFormatter is just the cleaner, canonical way of doing this.

mrks:
Mmh, now that we have a better assertion in place, that spurious fail does not show up again. It's unrelated to this PR anyway.

Bouncner (approved):

PR #2262, opened 2020-11-13
Adjust CI stage to generate TPC-H Q15 query plans

### LINKED ISSUE #2261: CI stages: Separate stages for generating query plans and verification?
For TPC-H Q15, we skip the verification in the CI since #1378 as we cannot support it. Therefore, we also do not get any query plans for Q15 [as build artifacts](https://hyrise-ci.epic-hpi.de/job/hyrise/job/hyrise/job/master/lastStableBuild/artifact/query_plans/tpch/).

By making two stages out of it, we could ensure that query plans are generated for all queries.

julianmenzler:
Fixes #2261.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -281,20 +281,12 @@ try {
         }, tpchQueryPlansAndVerification: {
           stage("tpchQueryPlansAndVerification") {
             if (env.BRANCH_NAME == 'master' || full_ci) {
-              sh "mkdir -p query_plans/tpch; cd query_plans/tpch; ln -s ../../resources; ../../clang-release/hyriseBenchmarkTPCH -r 1 --visualize --verify"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -287,6 +287,14 @@ try {
               Utils.markStageSkippedForConditional("tpchQueryPlansAndVerification")
             }
           }
+        }, tpchQ15QueryPlans: {
+          stage("tpchQ15QueryPlans") {
+            if (env.BRANCH_NAME == 'master' || full_ci) {
+              sh "mkdir -p query_plans/tpch; cd query_plans/tpch; ln -s ../../resources; ../../clang-release/hyriseBenchmarkTPCH -r 1 -q 15 --visualize"
```
mrks, line 293: This might cause concurrency issues with a second binary running in the same folder. Why not just add the `sh` command to the existing stage?

julianmenzler: Ok, good idea ;)

mrks (approved):

PR #2268, opened 2020-11-20
Fix operator performance counters in hash aggregate

### LINKED ISSUE #2266: Incorrect performance counters in hash aggregate
image

Here, it's 35 + 528 + 31 + 46 ms. Which is 640 ms, while the operator runtime is 615 ms.
~~Moreover, step "GroupByColumnsWriting" is often written twice, which overwrites the first runtime. So the error would be even larger, if we wouldn't overwrite.~~

I looked into the aggregate, but couldn't find a reason for the mismatch. @mrks : you did a lot lately on the aggregate, any ideas?

Bouncner:
Resolves #2266.

`_write_groupby_output` can be called at two places. In one of the (`write_aggregate_output()`), its self-recorded step runtime is included in the runtime of the AggregateColumnWriting step, leading to cumulative step runtimes that are larger than the actual operator runtime.

### DIFF
--- a/src/lib/operators/aggregate_hash.cpp
+++ b/src/lib/operators/aggregate_hash.cpp
@@ -654,7 +654,7 @@ std::shared_ptr<const Table> AggregateHash::_on_execute() {
 
   // _aggregate has its own internal timer. As groupby/aggregate column writing can be interleaved, the runtime is
   // stored in members and later written to the operator performance data struct.
-  auto& step_performance_data = static_cast<OperatorPerformanceData<OperatorSteps>&>(*performance_data);
+  auto& step_performance_data = dynamic_cast<OperatorPerformanceData<OperatorSteps>&>(*performance_data);
   step_performance_data.set_step_runtime(OperatorSteps::OutputWriting, timer.lap());
 
   step_performance_data.set_step_runtime(OperatorSteps::GroupByColumnsWriting, groupby_columns_writing_duration);
@@ -884,8 +884,9 @@ void AggregateHash::_write_aggregate_output(boost::hana::basic_type<ColumnDataTy
 
 template <typename ColumnDataType, AggregateFunction aggregate_function>
 void AggregateHash::write_aggregate_output(ColumnID aggregate_index) {
-  // Used to substract groupby writing in case it needs to be done.
-  auto write_groupby_output_duration = std::chrono::nanoseconds{};
+  // Used to track the duration of groupby columns writing, which is done for the first aggregate only. Value is
+  // substraced from the runtime of this method (thus, it's either 0 or non-zero for the first aggregate).
+  auto first_aggregate_column_groupby_writing_time = std::chrono::nanoseconds{};
   Timer timer;
 
   // retrieve type information from the aggregation traits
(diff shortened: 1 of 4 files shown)


src/lib/operators/aggregate_hash.cpp
```diff
@@ -624,10 +628,7 @@ std::shared_ptr<const Table> AggregateHash::_on_execute() {
     }
     _write_groupby_output(pos_list);
   }
-
-  // _aggregate and _write_groupby_output have their own, internal timer. Start measuring once they are done.
-  auto& step_performance_data = static_cast<OperatorPerformanceData<OperatorSteps>&>(*performance_data);
-  Timer timer;
+  step_performance_data.set_step_runtime(OperatorSteps::GroupByColumnsWriting, timer.lap());
```
mrks, line 631: Have a look at what your timer measures. If `_has_aggregate_functions` is true, no code is executed between the creation of the timer and `timer.lap()`.

`_write_groupby_output` is also called from `_write_aggregate_output`.

mrks (changes requested):

src/lib/operators/aggregate_hash.cpp
Bouncner: Changed that. Not sure how happy I am with this solution.

mrks: Have a look at #2270, where I ran into a similar problem. We should decide for one solution.

Bouncner: Doesn't look the same to me as you are adding step times within another function, aren't you? I don't want to pass runtimes between functions.

Bouncner: Just reviewed and saw that you are using members. I'll think about it...

src/lib/operators/aggregate_hash.cpp
```diff
@@ -946,6 +952,8 @@ void AggregateHash::write_aggregate_output(ColumnID aggregate_index) {
         std::make_shared<ValueSegment<decltype(aggregate_type)>>(std::move(values), std::move(null_values));
   }
   _output_segments[output_column_id] = output_segment;
+
+  aggregate_columns_writing_duration += timer.lap() - write_groupby_output_duration;
```
mrks, line 956: `_write_groupby_output` is only called for `aggregate_index == 0`. Did you test this?

Bouncner: Not sure what you are asking for. `write_groupby_output_duration` does only store non-zero runtimes, when  `aggregate_index == 0` was true. Was this the question?

mrks: I confused it with `groupby_columns_writing_duration`. Maybe, we can find clearer names? Also, should we a comment explaining that `write_groupby_output_duration` is 0 for all but the first aggregate?

Bouncner: Like this?

```
// Used to track the duration of groupby columns writing, which is done for the first aggregate only. Value is
  // substraced from the runtime of this method (thus, it's either 0 or non-zero for the first aggregate).
  auto first_aggregate_groupby_column_writing_time = std::chrono::nanoseconds{};
```

Don't like the var name.

mrks: Typo: subtracted

Maybe just call it `excluded_time`?

Bouncner: Changed.

mrks (approved):

PR #2270, opened 2020-11-25
Sort: Make performance steps more precise

mrks:
With this PR, we split up the "Sort" step into different phases:

Screenshot 2020-11-25 at 16 32 38

### DIFF
--- a/src/lib/operators/sort.cpp
+++ b/src/lib/operators/sort.cpp
@@ -245,7 +245,6 @@ std::shared_ptr<AbstractOperator> Sort::_on_deep_copy(
 void Sort::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeVariant>& parameters) {}
 
 std::shared_ptr<const Table> Sort::_on_execute() {
-  Timer timer;
   const auto& input_table = left_input_table();
 
   for (const auto& column_sort_definition : _sort_definitions) {
@@ -269,6 +268,10 @@ std::shared_ptr<const Table> Sort::_on_execute() {
   // ReferenceSegments.
   auto previously_sorted_pos_list = std::optional<RowIDPosList>{};
 
+  auto total_materialization_time = std::chrono::nanoseconds{};
+  auto total_temporary_result_writing_time = std::chrono::nanoseconds{};
+  auto total_sort_time = std::chrono::nanoseconds{};
+
   for (auto sort_step = static_cast<int64_t>(_sort_definitions.size() - 1); sort_step >= 0; --sort_step) {
     const auto& sort_definition = _sort_definitions[sort_step];
     const auto data_type = input_table->column_data_type(sort_definition.column);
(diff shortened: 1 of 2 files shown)


src/lib/operators/sort.cpp
```diff
@@ -278,17 +281,24 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
       auto sort_impl = SortImpl<ColumnDataType>(input_table, sort_definition.column, sort_definition.sort_mode);
       previously_sorted_pos_list = sort_impl.sort(previously_sorted_pos_list);
+
+      total_materialization_time += sort_impl.materialization_time;
+      total_temporary_result_writing_time += sort_impl.temporary_result_writing_time;
+      total_sort_time += sort_impl.sort_time;
     });
   }
 
   auto& step_performance_data = dynamic_cast<OperatorPerformanceData<OperatorSteps>&>(*performance_data);
```
Bouncner, line 291: Is there a reason to use a `dynamic_cast` over `static_cast`?

mrks: I had a case where I forgot to change the `OperatorPerformanceData` inheritance and everything went down south. This at least segfaults if something is wrong. Adding an assertion would also add noise.

We could also consider introducing a `checked_cast` (e.g., [this](https://github.com/Schorny/checked_cast)). I believe Hyrise1 had something similar.

Bouncner: Wouldn't a proper test for the sort operator catch the wrong the `static_cast`? If we could do the `checked_cast` in a few lines and as port of Hyrise, I would like it. But not sure if we should add another third party library just for that.

mrks: No. An invalid static_cast is UB and, as such, by definition untestable. We might hope for UBSan to catch it.

Adding a `checked_cast` would be a bigger thing and we should probably look at each `dynamic_cast + (Debug)Assert` combination throughout the code. Not sure if this is worth it.

I honestly don't have too strong of an opinion about this. If you prefer, we can change it to `static_cast` (both here and in the other operators, too).

Bouncner: Hmm, UB sucks. That's for sure.

We could use Boost's `polymorphic_cast` which appears to do the same as the `check_cast` lib. Wouldn't require another third party library. And we probably don't have to start immediately and replace all occurrences.
We could also try to use the pattern apparently used in the expression system where we DebugAssert a dynamic_cast and subsequently use a static_cast (see `cast_expresssion.cpp:27`).

Anyhow: I haven't found a place apart from the expression system (can't tell right now if that might be a measurable impact) where the `dynamic_cast` is within a loop and could be avoided. So I am fine with it, I guess.

Bouncner (approved):

Bouncner (approved):
Looks good to me. I started another PR run (bus error on Mac again).

Bouncner:
Zündung?

PR #2275, opened 2020-12-01
Make MetaTableManager::_add() public

Bouncner:
I am currently working on a plugin and found the meta tables to be an excellent way of storing information.
This PR simply makes the add table method public, so that a plugin can add meta tables to Hyise.

A sample use case (also what I am currently working on) is a plugin that extracts all operators from the query plan cache and makes it queryable. The Hyrise cockpit can then simply do a `SELECT * FROM meta_cool_new_list_of_operator_instances`.

### DIFF
--- a/src/test/lib/utils/meta_table_manager_test.cpp
+++ b/src/test/lib/utils/meta_table_manager_test.cpp
@@ -55,7 +55,7 @@ class MetaTableManagerTest : public BaseTest {
  protected:
   std::shared_ptr<const Table> mock_manipulation_values;
 
-  void SetUp() override {
+  void SetUp() {
     Hyrise::reset();
 
     const auto column_definitions = MetaMockTable().column_definitions();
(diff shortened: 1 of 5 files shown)


src/test/lib/utils/meta_table_manager_test.cpp
```diff
@@ -50,7 +50,7 @@ class MetaTableManagerTest : public BaseTest {
 
   // We need this as the add method of MetaTableManager is protected.
   // Won't compile if add is not called by test class, which is a friend of MetaTableManager.
-  static void add_meta_table(const MetaTable& table) { Hyrise::get().meta_table_manager._add(table); }
+  static void add_meta_table(const MetaTable& table) { Hyrise::get().meta_table_manager.add(table); }
```
mrks, line 53: The comment reads as if we can get rid of this?

mrks (approved):

Bouncner:
Re-requested a review. Just for the `get_table` function. This function allows to later interact with the added meta table.

mrks:
I guess we should make `add` and `get_table` consistent (i.e., `add_table`)?

Bouncner:
Just did exactly this. 👍

Bouncner:
Added a test.

Bouncner:
@mrks: Don't want to merge since I added code since acceptance. It's only the changes of ce63fa2 I'd like to get an ok for.

PR #2277, opened 2020-12-07
Hash join: track which input table is build side

Bouncner:
Track the build side of the hash join by setting a flag in the operator performance struct. This information is important for join runtime estimations. @dey4ss, @aloeser, and me currently solve it by copying the if statement of the join, which is not an ideal solution.
Moreover, it's convenient to have this information in the PQP visualization.

### DIFF
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -652,9 +652,7 @@ void JoinHash::PerformanceData::output_to_stream(std::ostream& stream, Descripti
 
   const auto *const separator = description_mode == DescriptionMode::SingleLine ? " " : "\n";
   stream << separator << "Radix bits:" << separator << radix_bits;
-  if (!left_input_is_build_side) {
-    stream << "." << separator <<  "Input sides have been flipped.";
-  }
+  stream << "." << separator <<  "Build side is " << (left_input_is_build_side ? "left." : "right.");
 }
 
 }  // namespace opossum
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -46,7 +46,7 @@ JoinHash::JoinHash(const std::shared_ptr<const AbstractOperator>& left,
                    const std::vector<OperatorJoinPredicate>& secondary_predicates,
                    const std::optional<size_t>& radix_bits)
     : AbstractJoinOperator(OperatorType::JoinHash, left, right, mode, primary_predicate, secondary_predicates,
-                           std::make_unique<OperatorPerformanceData<OperatorSteps>>()),
+                           std::make_unique<PerformanceData>()),
       _radix_bits(radix_bits) {}
 
 const std::string& JoinHash::name() const {
@@ -178,7 +178,7 @@ std::shared_ptr<const Table> JoinHash::_on_execute() {
     output_column_order = OutputColumnOrder::BuildFirstProbeSecond;
   }
 
-  auto& join_hash_performance_data = static_cast<PerformanceData&>(*performance_data);
+  auto& join_hash_performance_data = dynamic_cast<PerformanceData&>(*performance_data);
   resolve_data_type(build_column_type, [&](const auto build_data_type_t) {
     using BuildColumnDataType = typename decltype(build_data_type_t)::type;
     resolve_data_type(probe_column_type, [&](const auto probe_data_type_t) {
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -648,7 +648,7 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
 void JoinHash::PerformanceData::output_to_stream(std::ostream& stream, DescriptionMode description_mode) const {
   OperatorPerformanceData<OperatorSteps>::output_to_stream(stream, description_mode);
 
-   const auto *const separator = description_mode == DescriptionMode::SingleLine ? " " : "\n";
+  const auto *const separator = description_mode == DescriptionMode::SingleLine ? " " : "\n";
   stream << separator << "Radix bits:" << separator << radix_bits;
   if (!left_input_is_build_side) {
     stream << "." << separator <<  "Input sides have been flipped.";
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -94,8 +94,8 @@ size_t JoinHash::calculate_radix_bits(const size_t build_relation_size, const si
 
   // We assume an L2 cache of 1024 KB for an Intel Xeon Platinum 8180. For local deployments or other CPUs, this size
   // might be different (e.g., an AMD EPYC 7F72 CPU has an L2 cache size of 512 KB and Apple's M1 has 16 MB).
-  constexpr auto l2_cache_size = 1'024'000;                   // bytes
-  constexpr auto l2_cache_max_usable = l2_cache_size * 0.75;  // use 75% of the L2 cache size
+  constexpr auto L2_CACHE_SIZE = 1'024'000;                   // bytes
+  constexpr auto L2_CACHE_MAX_USABLE = L2_CACHE_SIZE * 0.75;  // use 75% of the L2 cache size
 
   // For information about the sizing of the bytell hash map, see the comments:
   // https://probablydance.com/2018/05/28/a-new-fast-hash-table-in-response-to-googles-new-fast-hash-table/
@@ -108,7 +108,7 @@ size_t JoinHash::calculate_radix_bits(const size_t build_relation_size, const si
       // key + value (and one byte overhead, see link above)
       static_cast<double>(sizeof(T) + 1) / 0.85;
 
-  const auto cluster_count = std::max(1.0, complete_hash_map_size / l2_cache_max_usable);
+  const auto cluster_count = std::max(1.0, complete_hash_map_size / L2_CACHE_MAX_USABLE);
 
   return static_cast<size_t>(std::ceil(std::log2(cluster_count)));
 }
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -127,13 +127,21 @@ std::shared_ptr<const Table> JoinHash::_on_execute() {
   auto probe_column_id = ColumnID{};
 
   /**
-   * Build and probe side are assigned as follows (depending only on JoinMode, except for inner joins where input
-   * relation sizes are considered):
+   * The hash join works best when the table being probed is larger than the side for which the hash table is built
+   * (i.e., the build side). As consequence, when we are to freely determine the build side, we chose the smaller
+   * input table. For other cases, we cannot freely decide.
    *
-   * JoinMode::Inner        The smaller relation becomes the build side, the bigger the probe side
-   * JoinMode::Left/Right   The outer relation becomes the probe side, the inner relation becomes the build side
-   * JoinMode::FullOuter    Not supported by JoinHash
-   * JoinMode::Semi/Anti*   The left relation becomes the probe side, the right relation becomes the build side
+   * Build and probe side are assigned as follows:
+   *   JoinMode::Inner        The smaller relation becomes the build side, the bigger the probe side
+   *   JoinMode::Left/Right   The outer relation becomes the probe side, the inner relation becomes the build side
+   *   JoinMode::FullOuter    Not supported by JoinHash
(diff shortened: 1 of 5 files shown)


src/lib/operators/join_hash.cpp
```diff
@@ -74,41 +74,41 @@ void JoinHash::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeV
...
+      - each entry in the hash map is a pair of the actual hash key and the SmallPosList storing uint32_t offsets (see
+        hash_join_steps.hpp)
   */
   if (build_relation_size > probe_relation_size) {
     /*
-      Hash joins perform best when the build relation is small. In case the
-      optimizer selects the hash join due to such a situation, but neglects that the
-      input will be switched (e.g., due to the join mode), the user will be warned.
+      Hash joins perform best when the build relation is small. In case the inputs are switched (e.g., due to the join
+      mode) making the build partition larger than the probe partition, the user will be warned.
```
mrks, line 90: I read that this is only for CI, but just thought of this:

This sounds as if there was a right and a wrong way for the optimizer to specify the join. Someone new might try to find the code that is responsible for avoiding the flipping.

How do you feel about getting rid of the "flip" terminology altogether and simply reporting whether the left or the right side is the build side?

Bouncner: Maybe something we should discuss on Wednesday.

Bouncner: Just notes for Friday. Handling this in the optimizer is something I would like to work on, but I don't have the time right now. It's a simple heuristic. Nothing fancy.

The problem is that the only alternative is the sort-merge join and we have broke that join's performance with the introduction of the secondary attributes.

mrks: Could you remind me why we even care about this? I'm having a hard time remembering in which cases this makes a difference.

Bouncner: Mainly because hash joins suck when the build side is much larger. Even right now with the sub-optimal sort-merge join we could easily speed up several of these joins.

Bouncner:
Will close it for now, since I am still working on the L2 sizing. That takes a while. However, happy to further discuss this PR and whether is makes sense.

src/lib/operators/join_hash.cpp
mrks: My point was that I wonder why the optimizer has to deal with it.

Bouncner: I don't get where this question comes from. There are no changes to the optimizer.

The place that I would currently change is rather the LQP translator.

mrks: The LQPTranslator should not touch any statistics. Even if - what advantage do we have if this is done in the LQP rather than in the operator?

Bouncner: I don't plan to touch neither the LQP translator, nor optimizer, nor the internals of the join hash. I don't see where this discussion is going.

If statistics are needed, then yes, I should not end up in the LQP translator. In this case, I'd vote for a LQPNode marker similar to the PredicateNode.
My personal favorite would be a proxy join operator that decides during runtime, but it wasn't received very well back in the days when we first built that with @SvenLehmann.

Does that answer your question? Maybe discuss via phone?

mrks: Let's do it just before you pick this PR up again.

Bouncner: I misunderstood your first comment and totally lost track of what we discussed afterwards.

Changed the output for the visualization to` "Build side is " << (left_input_is_build_side ? "left." : "right.");"`. So no mention of "flip".

Do you know what we have discussed afterwards?

Bouncner:
Reopened the PR. There are no performance changes, only the tracking of the build side.
Changing the radix partitioning will be part of a follow up PR.

src/test/lib/operators/join_hash_test.cpp
```diff
@@ -109,12 +109,13 @@ TEST_F(OperatorsJoinHashTest, DeepCopy) {
 }
 
 TEST_F(OperatorsJoinHashTest, RadixBitCalculation) {
-  // Simple cases: handle minimal inputs and very large inputs
-  EXPECT_EQ(JoinHash::calculate_radix_bits<int>(1, 1), 0ul);
-  EXPECT_EQ(JoinHash::calculate_radix_bits<int>(0, 1), 0ul);
-  EXPECT_EQ(JoinHash::calculate_radix_bits<int>(1, 0), 0ul);
-  EXPECT_TRUE(JoinHash::calculate_radix_bits<int>(std::numeric_limits<size_t>::max(),
-                                                  std::numeric_limits<size_t>::max()) > 0ul);
+  // Simpe tests to check that side switching and zero-sizes work.
```
mrks, line 112: Simple

src/lib/operators/join_hash.hpp
```diff
@@ -46,6 +46,14 @@ class JoinHash : public AbstractJoinOperator {
     OutputWriting
   };
 
+  struct PerformanceData : public OperatorPerformanceData<OperatorSteps> {
+    void output_to_stream(std::ostream& stream, DescriptionMode description_mode) const;
+
+     size_t radix_bits{0};
+    // Initially, the left input is the build side and the right side is the probe side.
+    bool left_input_is_build_side{true};
```
mrks, line 54: Format

src/lib/operators/join_hash.cpp
```diff
@@ -206,14 +208,17 @@ std::shared_ptr<const Table> JoinHash::_on_execute() {
...
-            dynamic_cast<OperatorPerformanceData<JoinHash::OperatorSteps>&>(*performance_data),
-            std::move(adjusted_secondary_predicates));
+            join_hash_performance_data, std::move(adjusted_secondary_predicates));
       } else {
         Fail("Cannot join String with non-String column");
       }
     });
   });
 
+  DebugAssert(_radix_bits, "Radix bits are not set.");
```
mrks, line 218: Assert?

src/lib/operators/join_hash.cpp
mrks: Hmm, I can't recall all of it. From looking at the previous discussion and the code, I think I was missing a distinction between two cases:

1. The optimizer passes the tables so that the build table is on the left. I still don't understand how that is any issue at all. Previously, we said that we "flip" the table. Instead, we should go for "build"/"probe" side as early as possible. That might already be the case. 
2. In the second case, we had to pick the larger table as the probe table. This is the case where we should show a warning, saying that the hash join might not be the most efficient join operator for this. The text and the comment about it should be clearer. Not just "e.g., due to the join mode". As far as I understand it, this is not an example but the only case where this warning should be printed. We should explain that certain join modes require the a certain table to be the build table - even if it is bigger.

Also, is "build partition" really the correct term here? Have we already partitioned the input? Also, do we need to talk about relations? I always find that term unnecessarily academic if everyone knows that we are talking about tables.

Bouncner:
> The optimizer passes the tables so that the build table is on the left. I still don't understand how that is any issue at all.

Still not sure, what we are talking about. Just comments?
This PR does not change anything in the optimizer. But ideally, the optimizer would not chose the hash join for joins which have the smaller relation as the probe table (roughly said). But that's not something I am touching with this PR.
Or do you ask me to extend the comment and the performance warning?

I'll add the other points in the comments. Definitely some good hints. I just explained the hash join to a student, the comments are probably not sufficient right now.

mrks:
> But ideally, the optimizer would not chose the hash join for joins which have the smaller relation as the probe table (roughly said).

Let me rephrase that. Do you mean:

> But ideally, the optimizer would not chose the hash join for join types THAT* require the smaller relation to be used as the probe table because that is the only way the hash join implementation can execute the join.

If yes, we are on the same page and this is only about comments. I felt that, in the past, this was mixed up with the flipping that simply happens if the probe side is not on the left.

*) Yes, I'll keep bugging you with that ;)

Bouncner:
> Also, is "build partition" really the correct term here? Have we already partitioned the input? Also, do we need to talk about relations? I always find that term unnecessarily academic if everyone knows that we are talking about tables.

I think build partitions is correct here. Before that, we determine the radix bits and this is follow up check to ensure we don't overflow the offsets. 
I removed all mentions of "relation".

> *) Yes, I'll keep bugging you with that ;)

Isn't that a matter of british vs. american english? My fellow brexiteers like to use which where ever possible.

src/lib/operators/join_hash.cpp
```diff
@@ -72,34 +72,41 @@ std::shared_ptr<AbstractOperator> JoinHash::_on_deep_copy(
 void JoinHash::_on_set_parameters(const std::unordered_map<ParameterID, AllTypeVariant>& parameters) {}
 
 template <typename T>
-size_t JoinHash::calculate_radix_bits(const size_t build_relation_size, const size_t probe_relation_size) {
+size_t JoinHash::calculate_radix_bits(const size_t build_relation_size, const size_t probe_relation_size,
+                                      const JoinMode mode) {
```
Bouncner, line 76: `mode` will be used in an upcoming PR as well.

mrks:
> I think build partitions is correct here. Before that, we determine the radix bits and this is follow up check to ensure we don't overflow the offsets.

But in the formerly referenced code, we compared `build_side_size > probe_side_size`, both of which refer to entire tables. Not an issue anymore, anyway.

> Isn't that a matter of british vs. american english? My fellow brexiteers like to use which where ever possible.

Interesting:

> Note that in British English, the word which is often used interchangeably with the restrictive that. [...] This common British construction is not strictly incorrect in American English, but it is generally avoided, especially in formal writing.

I can't think of too many cases where AE is more restrictive than BE.

https://www.lexico.com/grammar/that-or-which

src/lib/operators/join_hash.cpp
```diff
@@ -72,43 +72,50 @@ std::shared_ptr<AbstractOperator> JoinHash::_on_deep_copy(
...
-      optimizer selects the hash join due to such a situation, but neglects that the
-      input will be switched (e.g., due to the join mode), the user will be warned.
+      Hash joins perform best when the build side is small. For inner joins, we can simply select the smaller input
+      table as the build side. For other joins, such as semi or outer joins, the build side is fixed. In this case,
+      other join operators might be more efficient. We emit performance warning in this case. Note, as of now we do not
+      consider such cases of inefficient hash joins and do not switch to other join algorithms.
     */
-    PerformanceWarning("Build relation larger than probe relation in hash join");
+    std::stringstream performance_warning;
+    performance_warning << "Build side larger than probe side in hash join (join mode: " << magic_enum::enum_name(mode)
```
mrks, line 97: neat

src/lib/operators/join_hash.cpp
```diff
@@ -641,4 +654,12 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
   }
 };
 
+void JoinHash::PerformanceData::output_to_stream(std::ostream& stream, DescriptionMode description_mode) const {
+  OperatorPerformanceData<OperatorSteps>::output_to_stream(stream, description_mode);
+
+  const auto* const separator = description_mode == DescriptionMode::SingleLine ? " " : "\n";
+  stream << separator << "Radix bits:" << separator << radix_bits;
```
mrks, line 661: Why do you want a new line in here?

src/lib/operators/join_hash.cpp
```diff
@@ -72,43 +72,50 @@ std::shared_ptr<AbstractOperator> JoinHash::_on_deep_copy(
...
-  if (build_relation_size > probe_relation_size) {
+  if (build_side_size > probe_side_size) {
     /*
-      Hash joins perform best when the build relation is small. In case the
-      optimizer selects the hash join due to such a situation, but neglects that the
-      input will be switched (e.g., due to the join mode), the user will be warned.
+      Hash joins perform best when the build side is small. For inner joins, we can simply select the smaller input
+      table as the build side. For other joins, such as semi or outer joins, the build side is fixed. In this case,
+      other join operators might be more efficient. We emit performance warning in this case. Note, as of now we do not
+      consider such cases of inefficient hash joins and do not switch to other join algorithms.
```
mrks, line 94: "As of now" does not really add any information. Maybe "In the future, the optimizer could identify these cases of potentially inefficient hash joins and switch to other join algorithms"?

mrks (approved):

src/lib/operators/join_hash.cpp
mrks: (i.e., Why not "Radix bits: ")

PR #2280, opened 2020-12-11
Disable binary caches for CI

mrks:
For the join order benchmark, binary tables are written into the table_path. In CI, this path is shared by different targets that are potentially executed concurrently. This sometimes led to issues with corrupted binary files.

I am also sneaking in some things that are too minor to have their own PR.

### DIFF
--- a/scripts/test/hyriseBenchmarkJoinOrder_test.py
+++ b/scripts/test/hyriseBenchmarkJoinOrder_test.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-import glob
+# import glob
 import json
 import os
 import sys
@@ -98,9 +98,9 @@ def main():
         output["context"]["clients"], int(arguments["--clients"]), "Client count doesn't match with JSON:", return_error
     )
 
-    #if not glob.glob(arguments["--table_path"].replace("'", "") + "*.bin"):
-    #    print("ERROR: Cannot find binary tables in " + arguments["--table_path"])
-    #    return_error = True
+    # if not glob.glob(arguments["--table_path"].replace("'", "") + "*.bin"):
+    #     print("ERROR: Cannot find binary tables in " + arguments["--table_path"])
+    #     return_error = True
 
     os.system(f'rm -rf {arguments["--table_path"]}/*.bin')
 
--- a/scripts/test/hyriseBenchmarkJoinOrder_test.py
+++ b/scripts/test/hyriseBenchmarkJoinOrder_test.py
@@ -98,9 +98,9 @@ def main():
         output["context"]["clients"], int(arguments["--clients"]), "Client count doesn't match with JSON:", return_error
     )
 
-    if not glob.glob(arguments["--table_path"].replace("'", "") + "*.bin"):
-        print("ERROR: Cannot find binary tables in " + arguments["--table_path"])
-        return_error = True
+    #if not glob.glob(arguments["--table_path"].replace("'", "") + "*.bin"):
+    #    print("ERROR: Cannot find binary tables in " + arguments["--table_path"])
+    #    return_error = True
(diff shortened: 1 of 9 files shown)


scripts/test/hyriseBenchmarkJoinOrder_test.py
```diff
@@ -29,6 +28,10 @@ def main():
     arguments["--clients"] = "1"
     arguments["--scheduler"] = "false"
 
+    # Binary tables would be written into the table_path. In CI, this path is shared by different targets that are
+    # potentially executed concurrently. This sometimes led to issues with corrupted binary files.
```
Bensk1, line 32: This problem should occur for all benchmarks that cache binary tables, right? The other benchmark script files don't contain such a comment, so sad.

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -211,11 +211,12 @@ class PosHashTable {
 
 // The bloom filter (with k=1) is used during the materialization and build phases. It contains `true` for each
 // `hash_function(value) % BLOOM_FILTER_SIZE`. Much of how the bloom filter is used in the hash join could be improved:
-// (1) Dynamically check whether bloom filters are worth the memory and computational costs This could be based on the
+// (1) Dynamically check whether bloom filters are worth the memory and computational costs. This could be based on the
 //     input table sizes, the expected cardinalities, the hardware characteristics, or other factors.
 // (2) Choosing an appropriate filter size. 2^20 was experimentally found to be good for TPC-H SF 10, but is certainly
 //     not optimal in every situation.
-// (3) Check whether using multiple hash functions (k>1) brings any improvement.
+// (3) Check whether using multiple hash functions (k>1) brings any improvement. For single-threaded TPC-H SF 10 (which
```
Bensk1, line 218: `where many are under 2^20`: does this mean that most of these integers have values < 1M?

Bensk1: Also, `it however appears as if this has no immediate benefits.`: multiple hash functions have no effect or the check has no effect?

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -54,6 +54,8 @@ TEST_F(StressTest, TestTransactionConflicts) {
   };
 
   // Create the async objects and spawn them asynchronously (i.e., as their own threads)
+  // Note that async has a bunch of issues (cf. Mastering the C++17 STL, pages 205f).
```
Bensk1, line 57: Hm, I don't really know what to make of that and if the reader of this comment will order the book to understand the comment. Is there an online version of it?

scripts/test/hyriseBenchmarkJoinOrder_test.py
mrks: Not an issue for the TPC-* benchmarks, which place the cached files in the current directory. You are right about the file-based benchmark though.

src/lib/operators/join_hash/join_hash_steps.hpp
mrks: > Evaluate whether using multiple hash functions (k>1) brings any improvement. In a first experiment, however, we saw no benefits for single-threaded TPC-H SF 10.

Better? I don't think we need to be too specific here.

src/test/lib/concurrency/stress_test.cpp
mrks: > Note that async has a bunch of issues:
> - https://stackoverflow.com/questions/12508653/what-is-the-issue-with-stdasync
> - Mastering the C++17 STL, pages 205f

better?

Bensk1 (approved):

PR #2281, opened 2020-12-11
Improve PQP tooltip of table scan

Bouncner:
I was wondering, why table scans did not show any skipping (e.g., due to value not existing in the dictionary) for scans that should skip (at least that was what I thought).
Turns out, this information is only shown when there have been at least one chunk that was skipped or scanned binary.
This PR now fixes that (i) there has not been a tooltip before when no chunks were skipped and (ii) the skip information is always shown now.

Before Q19 (no information for non-skipped):
image

Now:
image

image

Further changed the table scan operator as well as the tests to use `dynamic_cast` over `static_cast` to ensure that wrong casts are catched. It's not in a hot loop, so that should not matter.

One potentially open question is how to handle table scan's splitting of reference table into their actual physical chunks. The table scan skips 92 chunks, even though just 30 chunks are passed into it.

### DIFF
--- a/src/lib/operators/table_scan.hpp
+++ b/src/lib/operators/table_scan.hpp
@@ -53,22 +53,11 @@ class TableScan : public AbstractReadOnlyOperator {
     size_t chunk_scans_sorted{0};
 
     void output_to_stream(std::ostream& stream, DescriptionMode description_mode) const override {
-      if (chunk_scans_skipped == 0 && chunk_scans_sorted == 0) {
-        return;
-      }
+      OperatorPerformanceData<AbstractOperatorPerformanceData::NoSteps>::output_to_stream(stream, description_mode);
 
       const auto* const separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
-      stream << separator << "Chunks: ";
-      if (chunk_scans_skipped > 0) {
-        stream << chunk_scans_skipped << " skipped";
-      }
-      if (chunk_scans_skipped > 0 && chunk_scans_sorted > 0) {
-        stream << ", ";
-      }
-      if (chunk_scans_sorted > 0) {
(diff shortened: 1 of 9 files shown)


src/lib/operators/table_scan.hpp
```diff
@@ -53,22 +53,11 @@ class TableScan : public AbstractReadOnlyOperator {
...
-        stream << chunk_scans_skipped << " skipped";
-      }
-      if (chunk_scans_skipped > 0 && chunk_scans_sorted > 0) {
-        stream << ", ";
-      }
-      if (chunk_scans_sorted > 0) {
-        stream << chunk_scans_sorted << " scanned using binary search";
-      }
-      stream << ". ";
+      stream << separator << "Chunks: " << chunk_scans_skipped << " skipped, ";
```
mrks, line 59: How do you feel about "early out" instead of skipped?

mrks (approved):

src/lib/operators/table_scan.hpp
Bouncner: Fine with me. Both for variable and printing? Can't think of a good name ... `chunk_scans_early_out`?

mrks: `num_early_out_chunks`?

Only if you agree with it though. I have no strong opinion about it, just thought it might be slightly clearer.

Bouncner: I am superbad at naming and when that makes it for somebody else clearer, I am happy with it.

And `num_sorted_scan_chunks`?

mrks: Forgot about that one. Maybe `num_chunks_with_binary_search` and `num_chunks_with_early_out`?

Bouncner: Done.

mrks (approved):

PR #2283, opened 2020-12-15
TableScan: track number of chunks where all rows match

aloeser:
While working with @dey4ss on our research seminar topic (what-if clustering cost models), we noticed that the table scan currently tracks only the number of chunks for which the scan was skipped because no row matches. However, this number is often 0 because those chunks get pruned before the scan is executed.
This PR adds an additional counter to track the number of chunks for which the scan was skipped because all rows match. This does happen, e.g., for `BETWEEN` scans on `l_shipdate`.

Those two metrics are not identical because for the latter we still need to write the output.

EDIT: As @Bensk1 pointed out, this PR changes the way sorted scans are tracked: If sorted BETWEEN scans use an early exit (no match) or an all-match-shortcut, the scan now increases the respective no/all-match counter, rather than the sorted scan counter, as the scan is not really executed with binary search. ColumnVsValue scans are not affected because they do not have a similar early-exit code path (unless I missed it).

### DIFF
--- a/src/lib/operators/table_scan.hpp
+++ b/src/lib/operators/table_scan.hpp
@@ -50,6 +50,7 @@ class TableScan : public AbstractReadOnlyOperator {
 
   struct PerformanceData : public OperatorPerformanceData<AbstractOperatorPerformanceData::NoSteps> {
     std::atomic<size_t> num_chunks_with_early_out{0};
+    std::atomic<size_t> num_chunks_with_all_rows_matching{0};
     std::atomic<size_t> num_chunks_with_binary_search{0};
 
     void output_to_stream(std::ostream& stream, DescriptionMode description_mode) const override {
@@ -57,6 +58,7 @@ class TableScan : public AbstractReadOnlyOperator {
 
       const auto* const separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
       stream << separator << "Chunks: " << num_chunks_with_early_out.load() << " skipped, ";
+      stream << separator << num_chunks_with_all_rows_matching.load() << " matched all rows,";
       stream << num_chunks_with_binary_search.load() << " scanned using binary search.";
     }
   };
--- a/src/lib/operators/table_scan.hpp
+++ b/src/lib/operators/table_scan.hpp
@@ -57,8 +57,8 @@ class TableScan : public AbstractReadOnlyOperator {
       OperatorPerformanceData<AbstractOperatorPerformanceData::NoSteps>::output_to_stream(stream, description_mode);
 
       const auto* const separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
-      stream << separator << "Chunks: " << num_chunks_with_early_out.load() << " skipped, ";
-      stream << separator << num_chunks_with_all_rows_matching.load() << " matched all rows,";
+      stream << separator << "Chunks: " << num_chunks_with_early_out.load() << " skipped with no results, ";
+      stream << separator << num_chunks_with_all_rows_matching.load() << " skipped with all matching,";
       stream << num_chunks_with_binary_search.load() << " scanned using binary search.";
     }
   };
--- a/src/lib/operators/table_scan.hpp
+++ b/src/lib/operators/table_scan.hpp
@@ -58,7 +58,7 @@ class TableScan : public AbstractReadOnlyOperator {
 
       const auto* const separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
       stream << separator << "Chunks: " << num_chunks_with_early_out.load() << " skipped with no results, ";
-      stream << separator << num_chunks_with_all_rows_matching.load() << " skipped with all matching,";
+      stream << separator << num_chunks_with_all_rows_matching.load() << " skipped with all matching, ";
       stream << num_chunks_with_binary_search.load() << " scanned using binary search.";
     }
   };
--- a/src/lib/operators/table_scan/column_between_table_scan_impl.cpp
+++ b/src/lib/operators/table_scan/column_between_table_scan_impl.cpp
@@ -49,7 +49,6 @@ void ColumnBetweenTableScanImpl::_scan_non_reference_segment(
     for (const auto& sorted_by : chunk_sorted_by) {
       if (sorted_by.column == _column_id) {
         _scan_sorted_segment(segment, chunk_id, matches, position_filter, sorted_by.sort_mode);
-        ++num_chunks_with_binary_search;
         return;
       }
     }
@@ -121,6 +120,7 @@ void ColumnBetweenTableScanImpl::_scan_dictionary_segment(
       });
     } else {
       // No NULLs, all entries match.
+      ++num_chunks_with_all_rows_matching;
       const auto output_size = position_filter ? position_filter->size() : segment.size();
       const auto output_start_offset = matches.size();
       matches.resize(matches.size() + output_size);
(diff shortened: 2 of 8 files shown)


src/lib/operators/table_scan.hpp
```diff
@@ -50,13 +50,15 @@ class TableScan : public AbstractReadOnlyOperator {
...
     std::atomic<size_t> num_chunks_with_early_out{0};
+    std::atomic<size_t> num_chunks_with_all_rows_matching{0};
     std::atomic<size_t> num_chunks_with_binary_search{0};
 
     void output_to_stream(std::ostream& stream, DescriptionMode description_mode) const override {
       OperatorPerformanceData<AbstractOperatorPerformanceData::NoSteps>::output_to_stream(stream, description_mode);
 
       const auto* const separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
       stream << separator << "Chunks: " << num_chunks_with_early_out.load() << " skipped, ";
+      stream << separator << num_chunks_with_all_rows_matching.load() << " matched all rows,";
```
Bouncner, line 61: This line might be misunderstood. It sounds like the info `input_size == output_size`, but not like there has been work skipped. I don't have an idea how to state this however. :-/

aloeser: maybe rephrase this line and the one above in `X skipped because no/all rows match?`

mrks: "X skipped with no results, Y skipped with all matching"?

mrks: Or maybe reuse the "matches_all/matches_none" terminology from the code? That makes it easy to find it in the code.

aloeser: Don't have a strong opinion here, I'm happy with whatever you prefer

Bensk1: I think both are fine. "X skipped with no results (matches_none), Y skipped with all matching (matches_all)"?

Bouncner: Reusing the terminology is a good idea, but I wouldn't directly it.

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -49,7 +49,6 @@ void ColumnBetweenTableScanImpl::_scan_non_reference_segment(
     for (const auto& sorted_by : chunk_sorted_by) {
       if (sorted_by.column == _column_id) {
         _scan_sorted_segment(segment, chunk_id, matches, position_filter, sorted_by.sort_mode);
-        ++num_chunks_with_binary_search;
```
Bensk1, line 52: With this PR, this metric will be different because it is not incremented for every `_scan_sorted_segment` but only for the ones without early out. Is anybody relying on this metric (@Bouncner maybe?).

Bouncner: Fine with me.

src/lib/operators/table_scan.hpp
aloeser: adressed

Bensk1 (approved):

src/test/lib/operators/operator_performance_data_test.cpp
```diff
@@ -107,7 +112,8 @@ TEST_F(OperatorPerformanceDataTest, TableScanPerformanceData) {
     const auto& performance_data = dynamic_cast<TableScan::PerformanceData&>(*table_scan->performance_data);
     EXPECT_GT(performance_data.walltime.count(), 0ul);
     EXPECT_EQ(performance_data.num_chunks_with_early_out, 0ul);
-    EXPECT_EQ(performance_data.num_chunks_with_binary_search, 2ul);
+    EXPECT_EQ(performance_data.num_chunks_with_all_rows_matching, 0ul);
```
Bouncner, line 115: So we don't have a short cut for chunk 1 with all 2's and the predicate `=2`?

src/test/lib/operators/operator_performance_data_test.cpp
```diff
@@ -116,13 +122,63 @@ TEST_F(OperatorPerformanceDataTest, TableScanPerformanceData) {
...
 
     const auto table_scan = std::make_shared<TableScan>(
-        table_wrapper, between_inclusive_(get_column_expression(table_wrapper, ColumnID{0}), 1, 2));
+        table_wrapper, between_inclusive_(get_column_expression(table_wrapper, ColumnID{0}), 1, 4));
     table_scan->execute();
 
     const auto& performance_data = dynamic_cast<TableScan::PerformanceData&>(*table_scan->performance_data);
+    /*
+     * TODO(anyone) actually, there could be 3 chunks with all rows matching, and none with binary search.
+     * However, the all-match shortcurt of the sorted segment search currently always assumes an exclusive between.
```
Bouncner, line 131: Assumes sounds kind of problematic, doesn't it? Can this break with other betweens?

Bouncner: There are two tests ` // Between scan`. Maybe change second comment to ` // Between scan on nullable columns`.

aloeser: > Assumes sounds kind of problematic, doesn't it? Can this break with other betweens?
Not really. The worst thing that can happen is that we perform binary searches where an all-match-shortcut would work too, i.e., some unnecessary work. But i'd expect the performance impact of this to be rather small.
The same is true for the no-match-shortcut (an inclusive between is assumed, which may cause unnecessary scans, but should not break anything)
https://github.com/hyrise/hyrise/blob/e3bb51cb6b93503eb3dcca5411a36344b2b226f9/src/lib/operators/table_scan/sorted_segment_search.hpp#L156

src/test/lib/operators/operator_performance_data_test.cpp
aloeser: Exactly, as far as I can tell, there is no such shortcut for ColumnVsValue scans in the sorted segment search, only for between scans. For Equals scans, we directly call `_get_first/last_bound()`, which I consider to be the search operation because after that we only write the output.

https://github.com/hyrise/hyrise/blob/e3bb51cb6b93503eb3dcca5411a36344b2b226f9/src/lib/operators/table_scan/sorted_segment_search.hpp#L80

Maybe you could consider some parts of the NotEquals ColumnVsValue scans as shortcuts, but those are not included in the tests

mrks:
mergemergemerge?

aloeser:
> mergemergemerge?

If you meant a master merge, done.
If you meant whether this is ready to be merged, probably @Bouncner should have a final look

src/lib/operators/table_scan.hpp
```diff
@@ -50,13 +50,15 @@ class TableScan : public AbstractReadOnlyOperator {
...
+    std::atomic<size_t> num_chunks_with_all_rows_matching{0};
     std::atomic<size_t> num_chunks_with_binary_search{0};
 
     void output_to_stream(std::ostream& stream, DescriptionMode description_mode) const override {
       OperatorPerformanceData<AbstractOperatorPerformanceData::NoSteps>::output_to_stream(stream, description_mode);
 
       const auto* const separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
-      stream << separator << "Chunks: " << num_chunks_with_early_out.load() << " skipped, ";
+      stream << separator << "Chunks: " << num_chunks_with_early_out.load() << " skipped with no results, ";
+      stream << separator << num_chunks_with_all_rows_matching.load() << " skipped with all matching,";
```
Bouncner, line 61: Looks like there is a missing space after the last comma of this line.

Bouncner (approved):
Looks good to me. But please look at this fundamental issue that I have found before merging.

Bensk1 (approved):

PR #2285, opened 2020-12-17
Enable TPC-DS Query Q64

mrks:


julianmenzler:
According to our feature matrix in Google Docs, Q64 was blocked by 
* join_graph_statistics_cache.cpp: Wrong number in requested column order

What was the fix?

mrks:
```
terminate called after throwing an instance of 'boost::wrapexcept<boost::bad_get>'

  what():  boost::bad_get: failed value get using boost::get

Aborted (core dumped)
```

None - I just wanted to check if the blacklist was still up-to-date. On my machine, that one executed without issues. :D

PR #2293, opened 2021-01-03
implement FastPFor encoding

benrobby:


benrobby:
was a mistake, this pr contains only WIP changes

PR #2298, opened 2021-01-10
Allow benchmark data generation in hyrise server

Bouncner:
This PR introduces the capability of generating benchmark data by simply passing a parameter to Hyrise.

- [x] Add .py test using pexpect


mrks:
Just an idea, not even a suggestion: Should we move the parsing of `benchmark_data` into `hyriseBenchmarkLib` and make the console use it, too? `generate_tpch 10` would then become `generate tpch:10` (or we could make the colon optional).

Bouncner:
The consistency makes sense. Just implemented it and bothered that we don't have an enum with benchmark names nor a default benchmark scale factor (the benchmark binary have that, but not the benchmark table generators).

Not sure how clean that is, but I also don't think introducing enums and scale factors is really that useful.

mrks:
No point in overengineering that. The benchmark generation facilities are just a (very useful) add-on. I don't think that we are building up any technical debt here.

Bouncner:
`out("  generate_tpch SCALE_FACTOR [CHUNK_SIZE] - Generate all TPC-H tables\n");`

Would a `generate tpch:10 100` be better than a `generate tpch 10 1000`? I was not aware of the chunk size as another parameter. But in this case, it's probably better to not use the `:` notation?

Then again I am not sure if we should also use a second parameter in the server. E.g. `hyriseServer --benchmark_name TPC-H --sizing_factor 10`.

mrks:
Argh. Forgot about that. I don't particularly like `generate tpch:10 100` either. We *could* come up with some benchmark configuration string that includes those options, too. Feels like we are overcomplicating things though. If you really wanted to fine-tune things, you can still modify the configuration in the `server.cpp`.

src/bin/server.cpp
```diff
@@ -10,6 +18,10 @@ cxxopts::Options get_server_cli_options() {
     ("help", "Display this help and exit") // NOLINT
     ("address", "Specify the address to run on", cxxopts::value<std::string>()->default_value("0.0.0.0"))  // NOLINT
     ("p,port", "Specify the port number. 0 means randomly select an available one. If no port is specified, the the server will start on PostgreSQL's official port", cxxopts::value<uint16_t>()->default_value("5432"))  // NOLINT
+    ("benchmark_data", "Specify the benchmark name and sizing factor to generate at server start (e.g., " // NOLINT
```
mrks, line 21: Something with "optional"? This could be understood as if this was something normal for the server to do.

src/bin/server.cpp
```diff
@@ -27,6 +39,43 @@ int main(int argc, char* argv[]) {
...
+      config->cache_binary_tables = false;  // Not yet supported for TPC-C
+      opossum::TPCCTableGenerator{static_cast<uint32_t>(sizing_factor), config}.generate_and_store();
+    } else if (benchmark_name == "tpcds") {
+      opossum::TPCDSTableGenerator{static_cast<uint32_t>(sizing_factor), config}.generate_and_store();
+    } else if (benchmark_name == "tpch") {
+      opossum::TPCHTableGenerator{sizing_factor, config}.generate_and_store();
+    } else {
+      Fail("Unexpected benchmark name passed in parameter 'benchmark_data'.");
+    }
+  }
```
mrks, line 77: Move this into a separate function? You wouldn't want someone who just wants to understand the server to read this.

src/bin/server.cpp
```diff
@@ -27,6 +39,43 @@ int main(int argc, char* argv[]) {
...
+    * facilitate easy evaluation of hyrise, we decided to integrate the data generation in to the server nonetheless.
+    */
+  if (parsed_options.count("benchmark_data")) {
+    auto benchmark_data_arg = parsed_options["benchmark_data"].as<std::string>();
+    std::vector<std::string> bechmark_data_config;
+
+    boost::trim_if(benchmark_data_arg, boost::is_any_of(":"));
+    boost::replace_all(benchmark_data_arg, "-", "");
+    boost::to_lower(benchmark_data_arg);
+    boost::split(bechmark_data_config, benchmark_data_arg, boost::is_any_of(":"), boost::token_compress_on);
```
mrks, line 54: Could you give one-line comments (or just examples) on what these transformations do?

src/bin/server.cpp
```diff
@@ -27,6 +39,43 @@ int main(int argc, char* argv[]) {
...
+    Assert(bechmark_data_config.size() < 3,
+           "Malformed input for benchmark data generation. Expecting a benchmark "
+           "name and a sizing factor.");
```
mrks, line 57: Did clang-format put that break at this weird position?

src/bin/server.cpp
```diff
@@ -27,6 +39,43 @@ int main(int argc, char* argv[]) {
...
+
+    const auto benchmark_name = bechmark_data_config[0];
```
mrks, line 59: typo

src/bin/server.cpp
```diff
@@ -27,6 +39,43 @@ int main(int argc, char* argv[]) {
     return 0;
   }
 
+  /**
+    * The following code handles the parameter `benchmark_data` which allows users to generate benchmark data when
+    * starting the hyrise server. This is not an ideal solution, but due to several users requests and our goal to
+    * facilitate easy evaluation of hyrise, we decided to integrate the data generation in to the server nonetheless.
```
mrks, line 45: > We do not plan on exposing other parameters, such as the encoding or the chunk size via this facility. You can change the modify the `config` object as needed.

Not sure though

src/bin/server.cpp
Bouncner: That was me. Copied it from TPCH. Since we use NOLINT, I could add one very long line, but that doesn't real help reading I think. Or do you mean something else?

src/bin/server.cpp
Bouncner: Works for me.

src/bin/server.cpp
mrks: I am just surprised that it doesn't fill up the first line.

src/bin/server.cpp
Bouncner: Added.

src/bin/server.cpp
Bouncner: I had the parsing stuff in CLIConfigParser, but leaves all the benchmark data generation still in the server. Putting the generation there as well doesn't make sense IMO.
Shall I create a benchmark helper? 

I'd prefer moving into an anonymous namespace, but that would leave it in the server.cpp.

src/bin/server.cpp
Bouncner: Changed.

src/bin/server.cpp
mrks: Anonymous namespace is more than sufficient.

Bouncner:
I hope everything is addressed. Expanded the test a little bit.

src/bin/server.cpp
```diff
@@ -10,6 +10,46 @@
 #include "tpcds/tpcds_table_generator.hpp"
 #include "tpch/tpch_table_generator.hpp"
 
+namespace {
+
+void generate_benchmark_data(std::string argument_string) {
+  std::vector<std::string> benchmark_data_config;
```
mrks, line 16: auto to stick?

src/bin/server.cpp
```diff
@@ -10,6 +10,46 @@
...
+
+  // Remove dashes and convert to lower case to unify different writings of benchmarks ("TPC-H", "tpch", or "tpc-h")
+  boost::replace_all(argument_string, "-", "");
+  boost::to_lower(argument_string);
+
+  // Split benchmark name and sizing factor
+  boost::split(benchmark_data_config, argument_string, boost::is_any_of(":"), boost::token_compress_on);
+  Assert(benchmark_data_config.size() < 3,
+         "Malformed input for benchmark data generation. Expecting a benchmark "
+         "name and a sizing factor.");
```
mrks, line 29: You are listing exactly two things - why are you comparing against 3?

src/bin/server.cpp
mrks: Also, move it down to where it's actually used first?

src/bin/server.cpp
```diff
@@ -27,6 +79,18 @@ int main(int argc, char* argv[]) {
     return 0;
   }
 
+  /**
+    * The optional parameter `benchmark_data` allows users to generate benchmark data when starting the hyrise server.
+    * This is not an ideal solution, but due to several users requests and our goal to facilitate easy evaluation of
```
mrks, line 84: "user requests" or "users' requests"

mrks (changes requested):

src/bin/server.cpp
Bouncner: Oh sorry, leftover from `size` being optional for the console.

mrks (approved):

PR #2306, opened 2021-01-22
Fix compilation of release cmake build type for ppc64le

pkoutsov:
This extra tiny modification of CMakeLists.txt is necessary so Hyrise can compile successfully for ppc64le arch and ```-DCMAKE_BUILD_TYPE=RelWithDebInfo```. Otherwise compilation fails with this message

```console
unrecognized command line option '-march=native'; did you mean '-mcpu=native'
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,7 +84,7 @@ set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${FLAGS_ALL}")
 # https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu
 # We check specifically for arm64 (e.g., Apple's M1) or aarch64 (e.g., Raspberry Pi 4), as these are the two ARM architectures we have tested so far.
 # Match for "arm" on your own risk to include other ARM architures.
-if (CMAKE_SYSTEM_PROCESSOR MATCHES "arm64|aarch64")
+if (CMAKE_SYSTEM_PROCESSOR MATCHES "arm64|aarch64|ppc64le")
     set(FLAGS_RELEASE "-mcpu=native")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${FLAGS_ALL}")
 # https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu
 # We check specifically for arm64 (e.g., Apple's M1) or aarch64 (e.g., Raspberry Pi 4), as these are the two ARM architectures we have tested so far.
 # Match for "arm" on your own risk to include other ARM architures.
-if (CMAKE_SYSTEM_PROCESSOR MATCHES "arm64|aarch64")
+if (CMAKE_SYSTEM_PROCESSOR MATCHES "arm64|aarch64|ppc64le")
```
Bouncner, line 87: Could you add a comment/extend the one on ARM builds? Is it similar to ARM and arch native works for GCC but not for clang?

mrks:
@Bouncner - should we just move to https://stackoverflow.com/questions/46724267/cmake-march-hardware ?

Bouncner:
Looks cleaner. Sure, why not.

mrks:
Do you want to look into that or should I?

Bouncner:
The M1 commit has been my first contact with CMake so far. I guess you're faster with that, so I'd be happy if you could do that one.

pkoutsov:
Since there is a pending PR (https://github.com/hyrise/hyrise/pull/2313) that addresses this exact issue, please feel free to close this PR.

mrks:
That was the plan for when we have verified that this works for all architectures. Can you confirm that the solution from #2313 also fixes the issue for ppc64le?

PR #2308, opened 2021-01-22
Fix Jenkins, allow admin override

mrks:
Github changed from HTTP/1.1 to HTTP/2 so the header check failed. Fixing that and also adding a quick hack that allows the Jenkins admin to run PRs of non-members.


Bouncner (approved):
I am not familiar with Jenkins at all, but the parts I understand look good to me.

PR #2310, opened 2021-01-23
Fix Jenkins 2.0

mrks-test:
Forgot that the master has no PR info :/

Also, by using a different user, I want to verify that non-Hyrise-Members cannot immediately have their code executed by Jenkins.


mrks:
"User unknown" - perfect. Now rerunning as admin.

Bouncner (approved):

PR #2313, opened 2021-01-27
Fix -march=native once and for all (?)

mrks:
This should enable `-march=native` for exactly those architecture/compiler combinations that support it:

https://github.com/hyrise/hyrise/pull/2306#issuecomment-767388729

- [x]  M1: builds with clang and no performance regression
- [x]  Raspberry PI: builds with clang/gcc and no performance regressions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,15 +79,13 @@ set(FLAGS_ALL "-fopenmp-simd")  # enables loop vectorization hints, but does not
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} ${FLAGS_ALL}")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${FLAGS_ALL}")
 
-# Build the binary optimized for the current system, ignoring older systems. For x86 systems, we use -march=native.
-# Clang does not support -march=native on ARM. ARM recommends setting -mcpu for both GCC and Clang instead of march:
-# https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu
-# We check specifically for arm64 (e.g., Apple's M1) or aarch64 (e.g., Raspberry Pi 4), as these are the two ARM architectures we have tested so far.
-# Match for "arm" on your own risk to include other ARM architures.
-if (CMAKE_SYSTEM_PROCESSOR MATCHES "arm64|aarch64")
-    set(FLAGS_RELEASE "-mcpu=native")
-else()
-    set(FLAGS_RELEASE "-march=native")  
+# Build the binary optimized for the current system, ignoring older systems.
+# Not all environments support march=native - check before we use it.
+set(FLAGS_RELEASE "-mcpu=native -mtune=native")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -79,15 +79,13 @@ set(FLAGS_ALL "-fopenmp-simd")  # enables loop vectorization hints, but does not
...
-# https://community.arm.com/developer/tools-software/tools/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu
-# We check specifically for arm64 (e.g., Apple's M1) or aarch64 (e.g., Raspberry Pi 4), as these are the two ARM architectures we have tested so far.
-# Match for "arm" on your own risk to include other ARM architures.
-if (CMAKE_SYSTEM_PROCESSOR MATCHES "arm64|aarch64")
-    set(FLAGS_RELEASE "-mcpu=native")
-else()
-    set(FLAGS_RELEASE "-march=native")  
+# Build the binary optimized for the current system, ignoring older systems.
+# Not all environments support march=native - check before we use it.
+set(FLAGS_RELEASE "-mcpu=native -mtune=native")
```
Bouncner, line 84: The reason for not setting all three was the following summary of the linked article (it's an ARM article):
```
TL;DR: Whenever possible use only -mcpu=native. Avoid -march and -mtune.
```

Bouncner:
Works with Clang and GCC on a recent Raspberry Pi with Ubuntu 20.04 (did not wait for the full compilation, takes quite some time).
On the M1, cmake runs for both compilers, but GCC does not compile (probably expected).

CMakeLists.txt
mrks: My understand is that this summary is limited to ARM. As far as I know, march is the way to go for x86. That contradicts the tldr.

Bouncner: Hmm, what you said has been done before this PR. Why does this contradict the tldr?

mrks: Are you talking about the addition of mtune? As far as I understand it, it doesn't hurt, but `mcpu` is more specific in some cases.

mrks:
Hmpf.

> clang: warning: argument unused during compilation: '-mcpu=native' [-Wunused-command-line-argument]

mrks:
Removed mtune, fixed clang unused warning.

Bouncner (approved):

Bouncner:
Zünden?

mrks:
Did you re-run it on the machines? I'd rather double check this, possibly even the performance, as any error that we introduce here will likely not be caught any time soon. I'm also hoping on feedback from @pkoutsov.

Bouncner:
> Did you re-run it on the machines? I'd rather double check this, possibly even the performance, as any error that we introduce here will likely not be caught any time soon. I'm also hoping on feedback from @pkoutsov.

M1 works fine. There has been a 9% improvement on Q9. When I ran it again it was 6% but then  Q22 improved by 7%. So I would assume is just flaky with Q9 right now on M1.

The RaspPi killed the compiler because of a lack of swap space. Compiling with `-j 1` now, that's going to take a while.

Bouncner:
Everything's fine on a Raspberry Pi 4.

pkoutsov:
Everything's fine on ```ppc64le``` as well

Issue #2014, opened 2020-02-10
(Really) run SQLite tests with scheduler


mrks:
Currently, the scheduler is reset by BaseTest. The solution is to run

```
  Hyrise::get().topology.use_numa_topology();
  Hyrise::get().set_scheduler(std::make_shared<NodeQueueScheduler>());
```

as part of SQLiteTestRunner::SetUp, not SQLiteTestRunner::SetUpTestCase. However, this currently fails. I suspect #1643 to be the culprit.

Bouncner:
Can't say for sure that test 106 (see #1643) is no longer a problem after #2315, but at least this test does not crash.

What crashes with an enabled scheduler is the following query:

Note: Google Test filter = *Line391WithUnencoded*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SQLiteTestRunnerUnencoded/SQLiteTestRunner
[ RUN      ] SQLiteTestRunnerUnencoded/SQLiteTestRunner.CompareToSQLite/Line391WithUnencoded
1
0x7fa33120fed8
libc++abi.dylib: terminating with uncaught exception of type std::logic_error: src/lib/operators/abstract_operator.cpp:72 Operator is already being executed.
[1]    12410 abort      ./hyriseTest --gtest_filter="*Line391WithUnencoded*"
Play around branch:
martin/#1643. Looks like the aggregate in the subquery is already executed?

Bouncner:
Pinging @julianmenzler here. Sorry for the confusion, I posted this text in the wrong issue and Markus pinged you there.

Tests 34 and 390 fail as well, but less often. 391 triggers it every time on my laptop.

julianmenzler:
<img width="706" alt="Untitled" src="https://user-images.githubusercontent.com/9380917/111286531-958de500-8642-11eb-8874-2c3e746fe357.png">

The scheduler schedules both inputs of the JoinHash for parallel execution. 
The TableScan uses the ExpressionEvaluator to spontaneously execute the first Projection which is already being executed as part of the JoinHash's right input. 

To fix the issue, we need to implement some waiting logic, maybe in conjunction with a state model for operators.

mrks:
Yes please. This should have priority over #2312, as it looks as if we broke something here. Maybe we could discuss options tomorrow?

julianmenzler:
> Maybe we could discuss options tomorrow?

Ok, so then I don't start with the code before our discussion.

[...]

PR #2315, opened 2021-01-29
Fix map synchronization in  join sort merge

Alexander-Dubrawski:
Currently, multiple jobs inside the join sort-merge are accessing the same maps `_right_row_ids_emitted`, `_left_row_ids_emitted`. This can lead to potential synchronization problems. This PR tries to fix this issue the following way:

* A vector for the right and the left side is created with maps for every cluster
* The vectors for the right and the left side are merged to one map (`_right_row_ids_emitted` and `_left_row_ids_emitted`)

### DIFF
--- a/src/lib/operators/join_sort_merge.cpp
+++ b/src/lib/operators/join_sort_merge.cpp
@@ -222,63 +222,63 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
   **/
   void _join_runs(TableRange left_run, TableRange right_run, CompareResult compare_result,
                   std::optional<MultiPredicateJoinEvaluator>& multi_predicate_join_evaluator,
-                  size_t cluster_number) {
-    // size_t cluster_number = left_run.start.cluster;
+                  size_t cluster_id) {
+
     switch (_primary_predicate_condition) {
       case PredicateCondition::Equals:
         if (compare_result == CompareResult::Equal) {
-          _emit_qualified_combinations(cluster_number, left_run, right_run, multi_predicate_join_evaluator);
+          _emit_qualified_combinations(cluster_id, left_run, right_run, multi_predicate_join_evaluator);
         } else if (compare_result == CompareResult::Less) {
           if (_mode == JoinMode::Left || _mode == JoinMode::FullOuter) {
-            _emit_right_primary_null_combinations(cluster_number, left_run);
+            _emit_right_primary_null_combinations(cluster_id, left_run);
           }
         } else if (compare_result == CompareResult::Greater) {
           if (_mode == JoinMode::Right || _mode == JoinMode::FullOuter) {
-            _emit_left_primary_null_combinations(cluster_number, right_run);
+            _emit_left_primary_null_combinations(cluster_id, right_run);
           }
         }
         break;
       case PredicateCondition::NotEquals:
         if (compare_result == CompareResult::Greater) {
-          _emit_qualified_combinations(cluster_number, left_run.start.to(_end_of_left_table), right_run,
+          _emit_qualified_combinations(cluster_id, left_run.start.to(_end_of_left_table), right_run,
                                        multi_predicate_join_evaluator);
         } else if (compare_result == CompareResult::Equal) {
-          _emit_qualified_combinations(cluster_number, left_run.end.to(_end_of_left_table), right_run,
+          _emit_qualified_combinations(cluster_id, left_run.end.to(_end_of_left_table), right_run,
                                        multi_predicate_join_evaluator);
-          _emit_qualified_combinations(cluster_number, left_run, right_run.end.to(_end_of_right_table),
+          _emit_qualified_combinations(cluster_id, left_run, right_run.end.to(_end_of_right_table),
                                        multi_predicate_join_evaluator);
         } else if (compare_result == CompareResult::Less) {
-          _emit_qualified_combinations(cluster_number, left_run, right_run.start.to(_end_of_right_table),
+          _emit_qualified_combinations(cluster_id, left_run, right_run.start.to(_end_of_right_table),
                                        multi_predicate_join_evaluator);
         }
         break;
       case PredicateCondition::GreaterThan:
         if (compare_result == CompareResult::Greater) {
-          _emit_qualified_combinations(cluster_number, left_run.start.to(_end_of_left_table), right_run,
+          _emit_qualified_combinations(cluster_id, left_run.start.to(_end_of_left_table), right_run,
                                        multi_predicate_join_evaluator);
         } else if (compare_result == CompareResult::Equal) {
-          _emit_qualified_combinations(cluster_number, left_run.end.to(_end_of_left_table), right_run,
+          _emit_qualified_combinations(cluster_id, left_run.end.to(_end_of_left_table), right_run,
                                        multi_predicate_join_evaluator);
         }
         break;
       case PredicateCondition::GreaterThanEquals:
         if (compare_result == CompareResult::Greater || compare_result == CompareResult::Equal) {
-          _emit_qualified_combinations(cluster_number, left_run.start.to(_end_of_left_table), right_run,
+          _emit_qualified_combinations(cluster_id, left_run.start.to(_end_of_left_table), right_run,
                                        multi_predicate_join_evaluator);
         }
         break;
       case PredicateCondition::LessThan:
         if (compare_result == CompareResult::Less) {
-          _emit_qualified_combinations(cluster_number, left_run, right_run.start.to(_end_of_right_table),
+          _emit_qualified_combinations(cluster_id, left_run, right_run.start.to(_end_of_right_table),
                                        multi_predicate_join_evaluator);
         } else if (compare_result == CompareResult::Equal) {
-          _emit_qualified_combinations(cluster_number, left_run, right_run.end.to(_end_of_right_table),
+          _emit_qualified_combinations(cluster_id, left_run, right_run.end.to(_end_of_right_table),
                                        multi_predicate_join_evaluator);
         }
         break;
       case PredicateCondition::LessThanEquals:
         if (compare_result == CompareResult::Less || compare_result == CompareResult::Equal) {
-          _emit_qualified_combinations(cluster_number, left_run, right_run.start.to(_end_of_right_table),
+          _emit_qualified_combinations(cluster_id, left_run, right_run.start.to(_end_of_right_table),
                                        multi_predicate_join_evaluator);
         }
         break;
@@ -512,10 +512,10 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
   * Performs the join on a single cluster. Runs of entries with the same value are identified and handled together.
   * This constitutes the merge phase of the join. The output combinations of row ids are determined by _join_runs.
   **/
-  void _join_cluster(size_t cluster_number,
+  void _join_cluster(size_t cluster_id,
                      std::optional<MultiPredicateJoinEvaluator>& multi_predicate_join_evaluator) {
-    auto& left_cluster = (*_sorted_left_table)[cluster_number];
-    auto& right_cluster = (*_sorted_right_table)[cluster_number];
+    auto& left_cluster = (*_sorted_left_table)[cluster_id];
+    auto& right_cluster = (*_sorted_right_table)[cluster_id];
 
     size_t left_run_start = 0;
     size_t right_run_start = 0;
@@ -532,9 +532,9 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
 
       auto compare_result = _compare(left_value, right_value);
 
-      TableRange left_run(cluster_number, left_run_start, left_run_end);
-      TableRange right_run(cluster_number, right_run_start, right_run_end);
-      _join_runs(left_run, right_run, compare_result, multi_predicate_join_evaluator, cluster_number);
+      TableRange left_run(cluster_id, left_run_start, left_run_end);
+      TableRange right_run(cluster_id, right_run_start, right_run_end);
+      _join_runs(left_run, right_run, compare_result, multi_predicate_join_evaluator, cluster_id);
 
       // Advance to the next run on the smaller side or both if equal
       if (compare_result == CompareResult::Equal) {
@@ -555,12 +555,12 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
     }
 
     // Join the rest of the unfinished side, which is relevant for outer joins and non-equi joins
-    auto right_rest = TableRange(cluster_number, right_run_start, right_size);
-    auto left_rest = TableRange(cluster_number, left_run_start, left_size);
+    auto right_rest = TableRange(cluster_id, right_run_start, right_size);
+    auto left_rest = TableRange(cluster_id, left_run_start, left_size);
     if (left_run_start < left_size) {
-      _join_runs(left_rest, right_rest, CompareResult::Less, multi_predicate_join_evaluator, cluster_number);
+      _join_runs(left_rest, right_rest, CompareResult::Less, multi_predicate_join_evaluator, cluster_id);
     } else if (right_run_start < right_size) {
-      _join_runs(left_rest, right_rest, CompareResult::Greater, multi_predicate_join_evaluator, cluster_number);
+      _join_runs(left_rest, right_rest, CompareResult::Greater, multi_predicate_join_evaluator, cluster_id);
     }
   }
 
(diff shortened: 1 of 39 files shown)


mrks:
Wheee, I thought this would never ever ever be fixed. Thanks for looking into it!

src/lib/operators/join_sort_merge.cpp
```diff
@@ -132,6 +132,9 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
   std::map<RowID, bool> _left_row_ids_emitted{};
   std::map<RowID, bool> _right_row_ids_emitted{};
 
+  std::vector<std::map<RowID, bool>> _left_row_ids_emitted_per_chunks;
```
Bouncner, line 135: We are slowly moving to auto to stick: `auto _left_row_ids_emitted_per_chunks std::vector<std::map<RowID, bool>>{};

src/lib/operators/join_sort_merge.cpp
```diff
@@ -510,10 +510,9 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
   * Performs the join on a single cluster. Runs of entries with the same value are identified and handled together.
   * This constitutes the merge phase of the join. The output combinations of row ids are determined by _join_runs.
   **/
-  void _join_cluster(size_t cluster_number,
-                     std::optional<MultiPredicateJoinEvaluator>& multi_predicate_join_evaluator) {
-    auto& left_cluster = (*_sorted_left_table)[cluster_number];
-    auto& right_cluster = (*_sorted_right_table)[cluster_number];
+  void _join_cluster(size_t cluster_id, std::optional<MultiPredicateJoinEvaluator>& multi_predicate_join_evaluator) {
```
Bouncner, line 513: const cluster_id

src/lib/operators/join_sort_merge.cpp
```diff
@@ -814,14 +818,24 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
     // Note: Equi outer joins can be integrated into the main algorithm, while these can not.
     if ((_mode == JoinMode::Left || _mode == JoinMode::FullOuter) &&
         _primary_predicate_condition != PredicateCondition::Equals) {
+      // We need to merge the maps from every cluster to one map.
+      for (auto const& map : _left_row_ids_emitted_per_chunks) {
+        for (auto const& submap : map) {
```
Bouncner, line 823: Structured bindings should be easier to read here.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -814,14 +818,24 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+          _left_row_ids_emitted[submap.first] = _left_row_ids_emitted[submap.first] || submap.second;
```
mrks, line 824: `|=?

src/lib/operators/join_sort_merge.cpp
mrks, line 135: Does this fix #1643 and does it allow us to tackle #2014 next?

Alexander-Dubrawski: If I do that the compiler is throwing the following error: `non-static data member declared with placeholder ‘auto’`. The reason is explained in: https://stackoverflow.com/questions/11302981/c11-declaring-non-static-data-members-as-auto .

src/lib/operators/join_sort_merge.cpp
```diff
@@ -132,6 +132,9 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
   std::map<RowID, bool> _left_row_ids_emitted{};
   std::map<RowID, bool> _right_row_ids_emitted{};
```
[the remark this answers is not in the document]
Alexander-Dubrawski: I moved it down to the `_perform_join` function. I also moved the `left_row_ids_emitted_per_chunk` and `right_row_ids_emitted_per_chunk` vectors there. As a result, we pass them through a few functions. I hope it is still ok 🙃 .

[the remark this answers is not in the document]
Alexander-Dubrawski: I replaced the `std::map` with `ska::bytell_hash_map` .

src/lib/operators/join_sort_merge.cpp
Alexander-Dubrawski: I moved the attributes to the `_perform_join` function. In there I used auto to stick. 🙃

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
mrks: Argh, I didn't realize that we were in the list of class members, not in some variables for the methods. I am not sure if the new version is actually better. :/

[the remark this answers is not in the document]
mrks: Do you have some performance numbers for that? Generally, I suggest to only use those "special" hashmaps if they have a measurable performance benefit over `unordered_map`.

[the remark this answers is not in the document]
mrks: (And yes, I probably should have included that information in the first comment. I'm somewhat unhappy with my own review here, sorry.)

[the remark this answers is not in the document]
Alexander-Dubrawski: No worries. I just reverted the commit.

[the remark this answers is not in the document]
Alexander-Dubrawski: I replaced `ska::bytell_hash_map` with `std::unordered_map `. I think to compare the performance of these two, it could make sense to write a little benchmark so that both hash maps are using the same hash function. Do you think that would be helpful? Or is a `std::unordered_map ` sufficient, since the use of the maps is an edge case anyway?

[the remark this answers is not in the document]
mrks: Works for me. If we come back and see that this is an issue, we can still look into other maps.

src/lib/operators/join_sort_merge.cpp
Alexander-Dubrawski: I wasn't able to reproduce the error from #1643 (master) on my machine 😞 . Sadly I found another error on master. If we use SMJ as the default join and are running the test suit the test `SQLiteTestRunnerUnencoded/SQLiteTestRunner.CompareToSQLite/Line162WithUnencoded` fails with the exception: `C++ exception with description "src/lib/operators/alias_operator.cpp:85 Chunk is sorted by an invalid column ID." thrown in the test body.`.

mrks: Feel free to move this into a new issue.

mrks: Also, there is a good chance that #1643 is easier to reproduce on machines with many cores.

Alexander-Dubrawski: I create a new Issue for the error #2321 . Still, I have some problems reproducing #1643 . So just to double-check:  

* I use clang by setting up CMake/Ninja with: `CC=/usr/bin/clang CXX=/usr/bin/clang++ cmake -DENABLE_NUMA_SUPPORT=Off -G Ninja ..`
* In https://github.com/hyrise/hyrise/blob/37d36653e0c6c2977f9b188381aa99cae2e2edc9/src/test/lib/sql/sqlite_testrunner/sqlite_testrunner.cpp#L56 I add the lines `Hyrise::get().topology.use_numa_topology(); Hyrise::get().set_scheduler(std::make_shared<NodeQueueScheduler>());`
* Then I build the `hyriseTest` executable and rin `./hyriseTest --gtest_filter=SQLiteTestRunnerUnencoded/SQLiteTestRunner.CompareToSQLite/Line192WithUnencoded` (https://github.com/hyrise/hyrise/blob/8827a25805e866ee7cc393f1b55698ddf6ef4d23/resources/test_data/sqlite_testrunner_queries.sql#L192)

I run the test on Aptera with 80 cores (If I check htop while running the test I can see that I am using all the cores).
Am I doing something wrong, or is it just bad luck? 😞

mrks: > I use clang by setting up CMake/Ninja with: CC=/usr/bin/clang CXX=/usr/bin/clang++ cmake -DENABLE_NUMA_SUPPORT=Off -G Ninja ..

Any reason you turn off NUMA? Also, I believe the canonical cmake method is `-DCMAKE_C(XX)_COMPILER=clang(++)`. Neither of those *should* make a difference though.

Maybe you could try running the test with `--gtest_repeat=-1` and simply hammer the test until it breaks.

Otherwise, have you understood the issue enough to write a query on the TPC-H data that could trigger the bug? Maybe we are too fast by now for the bug to occur on the sqlite data.

Alexander-Dubrawski: > Any reason you turn off NUMA?  

Not really, that's my mistake. 

> Otherwise, have you understood the issue enough to write a query on the TPC-H data that could trigger the bug? Maybe we are too fast by now for the bug to occur on the sqlite data.

I already wrote a little python script awhile ago with one query that should trigger the bug on a hyrise server that utilizes multiple cores. But it is not using the TPC-H data though. I could try to adjust the query to use the TPC-H data. The query was: 

```sql
SELECT * FROM table_1 LEFT JOIN table_2 ON table_1.a > table_2.a AND table_1.b >= table_2.b AND table_1.c < table_2.c
```

<details><summary>script</summary>
<p>

```python
#!/usr/bin/env python3

from psycopg2 import connect
from random import randrange
from multiprocessing import Process

def insert_task(host, port, rows, id):
    conn = connect(host=host, port=port)
    conn.set_session(autocommit=True)
    cur = conn.cursor()
    number_range = 5

    for _ in range(rows):
        cur.execute(f"INSERT INTO table_1 (a, b, c) VALUES ({randrange(number_range)}, {randrange(number_range)}, {randrange(number_range)});")
        cur.execute(f"INSERT INTO table_2 (a, b, c) VALUES ({randrange(number_range)}, {randrange(number_range)}, {randrange(number_range)});")

    print(f"-> Worker {id} done")

    cur.close()
    conn.close()

def load_table(host, port):
    conn = connect(host=host, port=port)
    conn.set_session(autocommit=True)
    cur = conn.cursor()

    cur.execute("CREATE TABLE table_1 (a int, b int, c int);")
    cur.execute("CREATE TABLE table_2 (a int, b int, c int);")

    cur.close()
    conn.close()

def insert(host, port, rows, workers):
    worker_processes = []
    for worker_id in range(workers):
        process = Process(target=insert_task, args=(host, port, rows, worker_id,))
        process.start()
        worker_processes.append(process)

    print("Waiting for workers to finish:")
    for process in worker_processes:
        process.join()
        process.terminate()


def execute_sql(host, port, sql, times):
    conn = connect(host=host, port=port)
    conn.set_session(autocommit=True)
    cur = conn.cursor()
    results = []

    print("Executing sql query:")
    for i in  range(times):
        print(f"-> Execute number: {i}")
        cur.execute(sql)
        result = cur.fetchall()
        results.append(result)
        if i > 0 and results[0] != result:
            print("Results do not match")
            return

    # for result in results:
    #     if results[0] != result:
    #         print("Results do not match")

def main():
    host = "0.0.0.0"
    port = "5432"
    rows_per_worker = 100_000
    workers = 10
    sql = "SELECT * FROM table_1 LEFT JOIN table_2 ON table_1.a > table_2.a AND table_1.b >= table_2.b AND table_1.c < table_2.c;"

    load_table(host, port)
    insert(host, port, rows_per_worker, workers)
    execute_sql(host, port, sql, 6)

if __name__ == '__main__':
    main()
```

</p>
</details>

> Maybe you could try running the test with --gtest_repeat=-1 and simply hammer the test until it breaks.

I think I will try this method for now.

mrks: It's not about the TPC-H data in particular, I just wonder if we should use more data to trigger the bug more reproducibly.

Bouncner: I am working on verifying that we no longer crash here. More or less by chance, we made it really hard to trigger the bug with #2004. This PR introduced checking for the size of jobs and only spawned tasks when the jobs have been big enough.
With that, we never had sufficiently large joins in the tests.

Still not super easy to trigger because once we have larger joins, we also use radix clustering ~that~ which kind of works against larger clusters. As soon as we enforce several radix clusters *and* always spawn jobs, it more or less crashes immediately.
But the point that is not easy to trigger obviously does not mean that we will never see it happen.

I'll ping you here, once I confirmed that it does not longer crash with this PR.

mrks: > we made it really hard to trigger the bug with #2004

Good catch!

Bouncner: So it ran for 2 hours without any problems.

I wanted to push and revert the code to have a reproducible example in the PR. But I think I failed. Sorry for that.

Bouncner:
Looking at the code I cannot find a place where we ever "unset" an emitted rowid. Consequently, I don't see a reason why we cannot use an unordered_set.
What am I missing here?

PS: would even allow using `merge()`.

Alexander-Dubrawski:
> Looking at the code I cannot find a place where we ever "unset" an emitted rowid. Consequently, I don't see a reason why we cannot use an unordered_set.
> What am I missing here?

I thought here: https://github.com/hyrise/hyrise/blob/alexander-dubrawski/Fix_map_synchronization_join_sort_merge/src/lib/operators/join_sort_merge.cpp#L460-L467
But I found this in the docs of std::map `The insertion only takes place if no other element in the container has a key equivalent to the one being emplaced (keys in a map container are unique).`

So I think you are right. 

> PS: would even allow using `merge()`.
> PPS: sorry for the commit fuck up. I totally failed and I am too stupid to correctly revert it. :(

No worries I will try to revert it later.

src/lib/operators/join_sort_merge.cpp
Alexander-Dubrawski: I try to revert it, but it kind of failed. Is it okay if we somehow force push 50c8249 and then commit the code (that tests the bug) without any merges in it, or even better in a separate branch where this branch is the base? We could refer to this branch here.

mrks: > even better in a separate branch where this branch is the base? We could refer to this branch here.

That would make it harder to use `git blame` later to understand the history of the PR.

I don't see too much wrong with force-pushing into a PR. If you are the only developer on that branch, it's fine in virtually any case; otherwise, just sync with the other people who use this branch.

Bouncner:
I think we decided against using `std::set`, right? If so, could you add a short explaination why it does not work with sets? Just in case somebody later revisits this PR and wonders.

Otherwise, anything that's blocking us here?

Alexander-Dubrawski:
> I think we decided against using `std::set`, right? If so, could you add a short explaination why it does not work with sets? Just in case somebody later revisits this PR and wonders.
> 
> Otherwise, anything that's blocking us here?

@Bouncner I had the plan to implement it the following way:
1. Add the emitted combination to the `set`
2. Merge all the sets from the partitions
3. Run over the hole Table Range and check if the Row ID is in the set. If not emit the null combination for this Row ID.

Somehow Step 3 isn't working correctly (Tests are failing because the output table has more Rows than it should have) and I'm not sure what exactly I am missing 😞 .

src/lib/operators/join_sort_merge.cpp
```diff
@@ -139,6 +139,24 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
   std::vector<std::shared_ptr<RowIDPosList>> _output_pos_lists_left;
   std::vector<std::shared_ptr<RowIDPosList>> _output_pos_lists_right;
 
+  struct RowHasher {
+    size_t operator()(const RowID& row) const {
+      size_t seed = 0;
```
Bouncner, line 144: `auto seed = size_t{0}`;

src/lib/operators/join_sort_merge.cpp
```diff
@@ -139,6 +139,24 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
   std::vector<std::shared_ptr<RowIDPosList>> _output_pos_lists_left;
   std::vector<std::shared_ptr<RowIDPosList>> _output_pos_lists_right;
 
+  struct RowHasher {
+    size_t operator()(const RowID& row) const {
```
Bouncner, line 143: For eight byte I wouldn't use a reference. But I don't have numbers to support that.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -139,6 +140,24 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+      auto seed = size_t{0};
+      boost::hash_combine(seed, row.chunk_id);
+      boost::hash_combine(seed, row.chunk_offset);
+      return seed;
+    }
+  };
+
+  using RowHashSet = std::unordered_set<RowID, RowHasher>;
+  // these are used for outer joins where the primary predicate is not Equals.
```
Bouncner, line 153: Should be a line below and that comment is not fully correct, right? It's about Outer with secondary predicates, isn't it?

mrks: Also please make sure to proper capitalize comments.

Bouncner (approved):

PR #2202, opened 2020-08-31
Hash Join: Merge output chunks <500 rows

mrks:
If a hash join's input is heavily pre-filtered or the join results in very few matches, we might end up with a high number of chunks that contain only few rows. This involves a trade-off: A lower number of output chunks reduces the overhead, especially when multi-threading is used. However, merging chunks destroys a potential references_single_chunk property of the PosList that would have been emitted otherwise.

@mweisgut - Did we already talk about the impact of chunk sizes on the efficiency of multi-threading? This popped up when I was looking at MT for my thesis, so I wrote a quick fix for it. `500` is a number that seems to work well for TPC-H, but that is just an educated guess. I suspect that there is more to the chunk+MT discussion than just this.

---------------

**System**
<details>
<summary>nemea - click to expand</summary>

| property | value |
| -- | -- |
| Hostname | nemea |
| CPU | Intel(R) Xeon(R) Platinum 8180 CPU @ 2.50GHz |
| Memory | 1.0TB |
| numactl | nodebind: 0  |
| numactl | membind: 0  |
</details>

**Commit Info and Build Time**
| commit | date | message | build time |
| -- | -- | -- | -- |
| f8e4cd4f8 | 27.08.2020 18:18 | Table Scan: Improve all-rows-match shortcut (#2194) | real 282.58 user 2451.64 sys 142.82|
| 9ae936be1 | 28.08.2020 17:31 | Hash Join: Merge output chunks <500 rows | real 282.50 user 2449.63 sys 142.54|

**hyriseBenchmarkTPCH - single-threaded**
<details>
<summary>
Sum of avg. item runtimes: +0%
 ||
Geometric mean of throughput changes: -1%
</summary>

```diff
 +Configuration Overview----+----------------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------+
 | Parameter                | /home/Markus.Dreseler/hyrise5/build-release/benchmark_all_results/hyriseBenchmarkTPCH_f8e4cd4f8c9426864a00a5a7ef2eda42a6c5bd1f_st.json | /home/Markus.Dreseler/hyrise5/build-release/benchmark_all_results/hyriseBenchmarkTPCH_9ae936be1ed8740d379b1743cff2540bd24559cd_st.json |
 +--------------------------+----------------------------------------------------------------------------------------------------------------------------------------+----------------------------------------------------------------------------------------------------------------------------------------+
[...]

PR #2316, opened 2021-01-29
Improve join merge sort output writing

Alexander-Dubrawski:
This PR improves the output writing of the join sort-merge. To do that it now uses the output writing the logic of the join hash. To minimize code duplication this PR creates a new folder `join_helper` and an inside it a `join_output_writing.hpp` where all the output writing logic is.


Alexander-Dubrawski:
> Maybe add a comment that we don't merge since we would lose the sort orders?
> Or is there another reason?

Will do. From my understanding that should be the reason.

src/lib/operators/join_hash.cpp
```diff
@@ -15,20 +15,15 @@
...
 #include "utils/timer.hpp"
 
-namespace {
-
-// Depending on which input table became the build/probe table we have to order the columns of the output table.
-// Semi/Anti* Joins only emit tuples from the probe table
-enum class OutputColumnOrder { BuildFirstProbeSecond, ProbeFirstBuildSecond, ProbeOnly };
-
-}  // namespace
+namespace {}  // namespace
```
Bouncner, line 26: Remove?

src/lib/operators/join_hash.cpp
```diff
@@ -527,114 +523,18 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
...
-                  std::back_inserter(*build_side_pos_list));
-        build_side_pos_lists[partition_id + 1] = {};
-
-        std::copy(probe_side_pos_lists[partition_id + 1].begin(), probe_side_pos_lists[partition_id + 1].end(),
-                  std::back_inserter(*probe_side_pos_list));
-        probe_side_pos_lists[partition_id + 1] = {};
-
-        ++partition_id;
-      }
+    auto allow_partition_merge = true;
```
Bouncner, line 532: `constexpr` and add a comment? Might be good to reference the initial PR for the result merging.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -911,96 +921,68 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+      if (!null_output_left.empty()) {
+        _output_pos_lists_left.push_back(std::move(null_output_left));
+        _output_pos_lists_right.push_back(std::move(null_output_right));
       }
     }
     _performance.set_step_runtime(OperatorSteps::Merging, timer.lap());
 
-    // Intermediate structure for output chunks (to avoid concurrent appending to table)
-    std::vector<std::shared_ptr<Chunk>> output_chunks(_output_pos_lists_left.size());
+    auto create_left_side_pos_lists_by_segment = (_left_input_table->type() == TableType::References);
```
Bouncner, line 952: const

Bouncner:
Comparison against the master. Both used the SMJ as the default join.

mrks:
Impressive :)

src/lib/operators/join_hash.cpp
Alexander-Dubrawski: I referenced #2202 as a PR. Is the reference okay?

Bouncner: Yes.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -911,96 +921,68 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
-        output_chunks[pos_list_id] = output_chunk;
-      };
 
-      if (write_output_concurrently) {
-        auto job = std::make_shared<JobTask>(write_output_chunk);
-        output_jobs.push_back(job);
-        output_jobs.back()->schedule();
-      } else {
-        write_output_chunk();
+    for (auto chunk : output_chunks) {
```
Bouncner, line 966: It's "just" a shared ptr, but I still prefer `auto chunk&`.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -911,96 +921,68 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
-    std::vector<std::shared_ptr<Chunk>> output_chunks(_output_pos_lists_left.size());
+    const auto create_left_side_pos_lists_by_segment = (_left_input_table->type() == TableType::References);
+    const auto create_right_side_pos_lists_by_segment = (_right_input_table->type() == TableType::References);
 
-    // Determine if writing output in parallel is necessary.
-    // As partitions ought to be roughly equally sized, looking at the first should be sufficient.
-    const auto first_output_size = _output_pos_lists_left[0]->size();
-    const auto write_output_concurrently = _cluster_count > 1 && first_output_size > JOB_SPAWN_THRESHOLD;
+    // Do not allow merge of the partitions, to keep the sort order.
+    constexpr auto allow_partition_merge = false;
```
Bouncner, line 956: Tidy will ask you to use ALL_CAPS for that.

src/lib/operators/join_hash.cpp
```diff
@@ -534,114 +528,20 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
...
-    }
+    const auto create_left_side_pos_lists_by_segment =
+        (_build_input_table->type() == TableType::References && _output_column_order != OutputColumnOrder::RightOnly);
+    const auto create_right_side_pos_lists_by_segment = (_probe_input_table->type() == TableType::References);
 
-    // probe_side_pos_lists_by_segment will only be needed if right is a reference table
-    if (_probe_input_table->type() == TableType::References) {
-      probe_side_pos_lists_by_segment = setup_pos_lists_by_chunk(_probe_input_table);
-    }
+    // A hash join's input can be heavily pre-filtered or the join results in very few matches. To counteract this the
```
Bouncner, line 537: It's the same for the sort-merge join. Not just the hash join.

src/lib/operators/join_hash.cpp
```diff
@@ -170,15 +163,16 @@ std::shared_ptr<const Table> JoinHash::_on_execute() {
...
   const auto probe_column_type = probe_input_table->column_data_type(probe_column_id);
 
-  // Determine output column order
+  // Depending on which input table became the build/probe table we have to order the columns of the output table.
+  // Semi/Anti* Joins only emit tuples from the probe table
   auto output_column_order = OutputColumnOrder{};
 
   if (_mode == JoinMode::Semi || _mode == JoinMode::AntiNullAsTrue || _mode == JoinMode::AntiNullAsFalse) {
-    output_column_order = OutputColumnOrder::ProbeOnly;
+    output_column_order = OutputColumnOrder::RightOnly;
```
Bouncner, line 171: Hmm, this will add a little bit of confusion again. We might want to add a few comments here.

@mrks : what's your take on the namings?

src/lib/operators/join_helper/join_output_writing.hpp
```diff
@@ -0,0 +1,269 @@
...
+/**
+ * Returns a vector where each entry with index i references a PosLists object. The PosLists object
+ * contains the position list of every segment/chunk in column i.
+ * @param input_table
+ */
+// See usage in _on_execute() for doc.
+inline PosListsByChunk setup_pos_lists_by_chunk(const std::shared_ptr<const Table>& input_table) {
+  Assert(input_table->type() == TableType::References, "Function only works for reference tables");
+
+  std::map<PosLists, std::shared_ptr<PosLists>> shared_pos_lists_by_pos_lists;
```
Bouncner, line 22: Are we sure that we don't want an `unordered_map`?

src/lib/operators/join_helper/join_output_writing.hpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+  }
+
+  std::vector<std::shared_ptr<Chunk>> output_chunks{};
+  output_chunks.reserve(expected_output_chunk_count);
+
+  // For every partition, create a reference segment.
+  auto partition_id = size_t{0};
+  auto output_chunk_id = size_t{0};
+  while (partition_id < pos_lists_left.size()) {
```
Bouncner, line 202: Reuse `pos_lists_left.size()` which has been pulled out of for loop.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -62,6 +63,11 @@ std::shared_ptr<const Table> JoinSortMerge::_on_execute() {
                    !_secondary_predicates.empty(), left_input_table()->type(), right_input_table()->type()}),
          "JoinSortMerge doesn't support these parameters");
 
+  std::shared_ptr<const Table> left_input_table_ptr;
```
Bouncner, line 66: Why in four lines?

Bouncner: I think (correct me if we use somewhere differently) we use the ending `_ptr` only for raw pointers.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -911,96 +921,68 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
-        output_chunks[pos_list_id] = output_chunk;
-      };
 
-      if (write_output_concurrently) {
-        auto job = std::make_shared<JobTask>(write_output_chunk);
-        output_jobs.push_back(job);
-        output_jobs.back()->schedule();
-      } else {
-        write_output_chunk();
+    for (auto &chunk : output_chunks) {
```
Bouncner, line 966: `auto& chunk`

src/lib/operators/join_hash.cpp
mrks: I remember that we got rid of the entire left/right terminology and probably had a good reason for it. One of it was probably that this becomes confusing once we flip input sides. I presume that this renaming is so that we can share OutputColumnOrder with both join implementations?

src/lib/operators/join_sort_merge.cpp
mrks: Nope. We just generally try to avoid it because it does not really add any valuable information.

src/lib/operators/join_hash.cpp
Bouncner: Yes, but thinking again about it, it might be ok here, since it's only about the output order. You only have to consider the input order for that and that should be quite clear.
Still think we need to add a good comment here, but it should not be the same confusing terminilogy as it was in the hash join internals.

mrks: Agreed.

src/lib/operators/join_hash.cpp
Alexander-Dubrawski: How about I add the following comment to the sort-merge join: 

```
    // A sort merge join's input can be heavily pre-filtered or the join results in very few matches. In contrast to
    // the hash join we can not allow a merge of the partitions, since we need to keep the sort order.
```

src/lib/operators/join_helper/join_output_writing.hpp
Alexander-Dubrawski: @Bouncner I replaced the `std:.map` with `std::unordered_map`. Can you please have a look if you are happy with the implementation especially with the hasher.

src/lib/operators/join_hash.cpp
Bouncner: I'd rather say:
```
// A sort merge join's input can be heavily pre-filtered or the join results in very few matches. In contrast to the hash join, we do not (for now) merge small partitions to keep the sorted chunk guarantees, which could be exploited by subsequent operators.

File: src/lib/operators/join_hash.cpp (hunk 7 above)
Comment (Bouncner) [reply src/lib/operators/join_hash.cpp:171]: What about extending the current comment.

Current:
```
// Depending on which input table became the build/probe table we have to order the columns of the output table.
// Semi/Anti* Joins only emit tuples from the probe table
```
New:
```
// Depending on which input table became the build/probe table we have to order the columns of the output table.
// Semi/Anti* Joins only emit tuples from the probe table, which is given by the right input table in Hyrise.
// For other join modes, we need to check which side has been chosen as the probe side (see variable `build_hash_table_for_right_input` for more details).
```

@mrks?

Comment (mrks) [reply src/lib/operators/join_hash.cpp:171]: Works for me.

**Comment by Bouncner** [comment]
Running a final benchmark (no modified join order in LQP translator) to ensure we have accidentally introduced any regressions for the hash join.

**Comment by Bouncner** [comment]
No performance changes for the hash join. Ready to go from my side. Anything to add @mrks?

File: src/lib/operators/join_hash.cpp [hunk 12]
```diff
@@ -171,15 +164,20 @@ std::shared_ptr<const Table> JoinHash::_on_execute() {
   const auto build_column_type = build_input_table->column_data_type(build_column_id);
   const auto probe_column_type = probe_input_table->column_data_type(probe_column_id);
 
-  // Determine output column order
+  // Depending on which input table became the build/probe table we have to order the columns of the output table.
+  // Semi/Anti* Joins only emit tuples from the probe table
   auto output_column_order = OutputColumnOrder{};
 
+  // Depending on which input table became the build/probe table we have to order the columns of the output table.
+  // Semi/Anti* Joins only emit tuples from the probe table, which is given by the right input table in Hyrise.
```
mrks, line 172: Why is the text duplicated?

src/lib/operators/join_hash.cpp
```diff
@@ -535,114 +533,21 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
...
+    const auto create_right_side_pos_lists_by_segment = (_probe_input_table->type() == TableType::References);
 
-    // probe_side_pos_lists_by_segment will only be needed if right is a reference table
-    if (_probe_input_table->type() == TableType::References) {
-      probe_side_pos_lists_by_segment = setup_pos_lists_by_chunk(_probe_input_table);
-    }
+    // A sort merge join's input can be heavily pre-filtered or the join results in very few matches. In contrast to
+    // the hash join, we do not (for now) merge small partitions to keep the sorted chunk guarantees, which could be
+    // exploited by subsequent operators.
+    constexpr auto ALLOW_PARTITION_MERGE = true;
```
mrks, line 545: Why are you talking about the sort-merge join in the hash-join file?

src/lib/operators/join_helper/join_output_writing.hpp
```diff
@@ -0,0 +1,278 @@
...
+     *          PosLists Chunk by Chunk
+     *      - Avoid creating the std::vector<const RowIDPosList*> for each Partition over and over again.
+     *
+     * They hold one entry per column in the table, not per AbstractSegment in a single chunk
+     */
+
+  PosListsByChunk left_side_pos_lists_by_segment;
+  PosListsByChunk right_side_pos_lists_by_segment;
+
+  // left_side_pos_lists_by_segment will only be needed if build is a reference table and being output
```
mrks, line 186: What do you mean with "if build"?

src/lib/operators/join_helper/join_output_writing.hpp
```diff
@@ -0,0 +1,278 @@
...
+ * @param input_table Table which all the position lists reference
+ * @param input_pos_list_ptrs_sptrs_by_segments Contains all position lists to all columns of input table
+ * @param pos_list contains the positions of rows to use from the input table
+ */
+inline void write_output_segments(Segments& output_segments, const std::shared_ptr<const Table>& input_table,
+                                  const PosListsByChunk& input_pos_list_ptrs_sptrs_by_segments,
+                                  std::shared_ptr<RowIDPosList> pos_list) {
+  std::map<std::shared_ptr<PosLists>, std::shared_ptr<RowIDPosList>> output_pos_list_cache;
+
+  // We might use this later, but want to have it outside of the for loop
```
mrks, line 76: Not yours, but this comment is pointless.

mrks (commented):
I trust both your review on this and our join test suite. As such, I only scrolled over the changes. Just a handful of remarks.

src/lib/operators/join_helper/join_output_writing.hpp
Bouncner: Thanks for the review.
I guess only reading the changes isn't sufficient. :( I should have seen these ...

src/lib/operators/join_helper/join_output_writing.hpp
```diff
@@ -183,12 +182,13 @@ inline std::vector<std::shared_ptr<Chunk>> write_output_chunks(
   PosListsByChunk left_side_pos_lists_by_segment;
   PosListsByChunk right_side_pos_lists_by_segment;
 
-  // left_side_pos_lists_by_segment will only be needed if build is a reference table and being output
+  // left_side_pos_lists_by_segment will only be needed if the left input table is a reference table and being an
```
Bouncner, line 185: I don't get "and being an output table."

Alexander-Dubrawski: Sorry, but I didn't write the code. So I'm not really sure about a more precise comment.

mrks: Before this PR, this was

> build_side_pos_lists_by_segment will only be needed if build is a reference table and being output

Knowing that the build side might not be part of the output and reading "output" not as a noun but as a participle adjective (read: an [*outputted*](https://english.stackexchange.com/questions/35418/past-tense-of-to-output-output-or-outputted) table)

Honestly, I think this comment is pointless right now.

Alexander-Dubrawski: ok, then I will remove it.

Bouncner (approved):

PR #2328, opened 2021-02-22
stateless

noamaviv:
A PR that allows Hyrise to run with the StorageManager and other objects given as parameters. This removes the dependence on a global object, possibly allowing different namespaces/schemas, an alternate transaction model, and several different implementations of StorageManager.

Note: While most tests pass, a few still don't. I am posting this PR to see if you are interested, if you are, I will fix the few tests that do not pass.

### DIFF
--- a/src/benchmarklib/abstract_benchmark_item_runner.cpp
+++ b/src/benchmarklib/abstract_benchmark_item_runner.cpp
@@ -9,8 +9,9 @@
 
 namespace opossum {
 
-AbstractBenchmarkItemRunner::AbstractBenchmarkItemRunner(const std::shared_ptr<BenchmarkConfig>& config)
-    : _config(config) {}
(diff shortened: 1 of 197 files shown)


mrks:
Thank you for sharing your modfications. The team has discussed this extensively. As of now, we believe that the drawbacks outweigh the benefits. Most importantly, the PR would introduce code that is effectively "dead". The master would not contain any code path that uses multiple `Hyrise` instances. Not only does this make the code (slightly) harder to understand and maintain, but it also makes testing and debugging more complicated. If there is no code that uses multiple instances, we will not notice if something gets broken.

Additionally, we do not see clear semantics for the lines between different instances. The PluginManager, which in this PR is also duplicated, can not support a plugin being loaded twice because of the internal dynamic linking of plugins. For the scheduler, one could argue that instances should schedule their tasks independently; however, the scheduler currently assumes that it owns the entire system (or at least all assigned cores). Similar questions arise for the meta tables.

We would have ended up at the same point if this was an internal contribution. In fact, for similar reasons, many modifications that were made of master's theses etc. have not been merged either.

PR #2329, opened 2021-02-25
Fix 2321

### LINKED ISSUE #2321: After SortMergeJoin Chunk is sorted by an invalid column ID
If we use SMJ as the default join (change the order in https://github.com/hyrise/hyrise/blob/8827a25805e866ee7cc393f1b55698ddf6ef4d23/src/lib/logical_query_plan/lqp_translator.cpp#L359) and are running the test  
`./hyriseTest --gtest_filter=SQLiteTestRunnerUnencoded/SQLiteTestRunner.CompareToSQLite/Line162WithUnencoded`  
, it failes with the following error message:

```
C++ exception with description "src/lib/operators/alias_operator.cpp:85 Chunk is sorted by an invalid column ID." thrown in the test body.
[  FAILED  ] SQLiteTestRunnerUnencoded/SQLiteTestRunner.CompareToSQLite/Line162WithUnencoded, where GetParam() = ((162, "SELECT \"left\".a1, \"left\".a2, \"right\".a1 FROM (SELECT a AS a1, a AS a2 FROM mixed) AS \"left\" JOIN (SELECT a AS a1, a AS a2 FROM mixed_null) AS \"right\" ON \"left\".a1 = \"right\".a2;"), Unencoded) (164 ms)
```

Bouncner:
In cases where a sorted column is projected twice (e.g., `SELECT a as a1, a as a2 FROM table`), the following alias receives the full sorting information, but only one of the sorted columns (and that twice). Thus, I cannot validate if both columns (by which the chunk should be sorted) are actually sorted. This is caused by the LQP translator (see #2321).
Instead of changing the LQP translator, which would require quite a lot of changes. We modified to the alias operator in this PR.

Resolves #2321.


Bouncner:
@mrks: You can savely ignore the tests and the changes on the install file. A full CI runs passes with the changes.
It's just about the 10 lines of actual change.

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -356,7 +356,7 @@ std::shared_ptr<AbstractOperator> LQPTranslator::_translate_join_node(
   // Lacking a proper cost model, we assume JoinHash is always faster than JoinSortMerge, which is faster than
   // JoinNestedLoop and thus check for an operator compatible with the JoinNode in that order
   constexpr auto JOIN_OPERATOR_PREFERENCE_ORDER =
-      hana::to_tuple(hana::tuple_t<JoinHash, JoinSortMerge, JoinNestedLoop>);
+      hana::to_tuple(hana::tuple_t<JoinSortMerge, JoinHash, JoinNestedLoop>);
```
mrks, line 359: ?

src/test/lib/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -609,7 +609,7 @@ TEST_F(LQPTranslatorTest, ProjectionNode) {
   EXPECT_EQ(*projection_op->expressions[0], *PQPColumnExpression::from_table(*table_int_float, "a"));
 }
 
-TEST_F(LQPTranslatorTest, JoinNodeToJoinHash) {
+TEST_F(LQPTranslatorTest, DISABLED_JoinNodeToJoinHash /* TMP: #2321 */) {
```
mrks, line 612: ?

src/lib/operators/alias_operator.cpp
```diff
@@ -79,12 +79,17 @@ std::shared_ptr<const Table> AliasOperator::_on_execute() {
       auto sort_definitions = std::vector<SortColumnDefinition>{};
       sort_definitions.reserve(sorted_by.size());
 
-      // Adapt column ids
+      // Adapt column ids of chunk sort definitions
       for (const auto& sort_definition : sorted_by) {
+        // In some edge cases an input table might be sorted by a column that is not included in the list of columns.
```
mrks, line 84: cases,

src/lib/operators/alias_operator.cpp
```diff
@@ -79,12 +79,17 @@ std::shared_ptr<const Table> AliasOperator::_on_execute() {
...
+        // This can happen when an expression occurs repeatedly (e.g., `SELECT a as a1, a as a2`) and the LQP
+        // translator references the first occurrence twice (leaving the second (sorted) occurrence unreferenced, see
+        // issue #2321 for more details). We thus check, if the sorted column is part of the input table.
```
mrks, line 87: no comma

src/lib/operators/alias_operator.cpp
```diff
@@ -79,12 +79,17 @@ std::shared_ptr<const Table> AliasOperator::_on_execute() {
...
         const auto it = std::find(_column_ids.cbegin(), _column_ids.cend(), sort_definition.column);
-        Assert(it != _column_ids.cend(), "Chunk is sorted by an invalid column ID.");
-        const auto index = ColumnID{static_cast<uint16_t>(std::distance(_column_ids.cbegin(), it))};
-        sort_definitions.emplace_back(SortColumnDefinition(index, sort_definition.sort_mode));
+        if (it != _column_ids.cend()) {
+          const auto index = ColumnID{static_cast<uint16_t>(std::distance(_column_ids.cbegin(), it))};
+          sort_definitions.emplace_back(SortColumnDefinition(index, sort_definition.sort_mode));
+        }
```
mrks, line 92: Assert that the number of output sort definitions matches that of input definitions?

mrks, line 92: Where is this tested?

src/lib/operators/projection.cpp
```diff
@@ -272,10 +272,12 @@ std::shared_ptr<const Table> Projection::_on_execute() {
...
       transformed.reserve(sorted_by.size());
       for (const auto& [column_id, mode] : sorted_by) {
-        if (!input_column_to_output_column.count(column_id)) {
+        const auto iter = std::find_if(output_column_to_input_column.begin(), output_column_to_input_column.end(),
+                                       [column_id = column_id](const auto entry) { return column_id == entry.second; });
+        if (iter == output_column_to_input_column.end()) {
           continue;  // column is not present in output expression list
         }
-        const auto projected_column_id = input_column_to_output_column[column_id];
+        const auto projected_column_id = iter->first;
```
mrks, line 280: Where is this tested?

src/lib/logical_query_plan/lqp_translator.cpp
Bouncner: Just to trigger everything in the CI.

src/test/lib/logical_query_plan/lqp_translator_test.cpp
Bouncner: Just to trigger everything in the CI. Does not pass as we need to have SMJ for the bug. This test checks for the hash join. Will be reverted.

src/lib/operators/alias_operator.cpp
Bouncner: It does not equal. Since the alias operator never sees the second sorted columns (it's removed in the LQP translator), we lose the second sorted column.

We could iterate the columns and check if they are part of the sorted definition instead of the current way of iterating the sorted information. This way, we should have the same length.

Is this was you recommend?

mrks: > we lose the second sorted column

Right, but won't we find the first column twice?

Bouncner: Changed the code to find the first column twice, added an assert.

src/lib/operators/projection.cpp
Bouncner: Tested with extended test in the projection tests.

src/lib/operators/alias_operator.cpp
Bouncner: Added test.

Bouncner:
Full CI went through. Will remove the temporary changes (join order and disabled tests) once we agree the fix is ok.

src/lib/operators/projection.cpp
```diff
@@ -272,11 +272,12 @@ std::shared_ptr<const Table> Projection::_on_execute() {
       std::vector<SortColumnDefinition> transformed;
       transformed.reserve(sorted_by.size());
       for (const auto& [column_id, mode] : sorted_by) {
-        if (!input_column_to_output_column.count(column_id)) {
-          continue;  // column is not present in output expression list
+        // Full iteration as input columns can be projected repeatedly
+        for (const auto& [output_column_id, input_column_id] : output_column_to_input_column) {
+          if (column_id == input_column_id) {
```
mrks, line 277: Any reason you are not using find_if here?

mrks (approved):

src/lib/operators/projection.cpp
Bouncner: I do not only care about the first hit. Another output column might be sorted as well with the sorting information of the same input column.

mrks: Wouldn't it make sense to swap the two for loops then and to use `find_if`? I won't block the PR either way.

Bouncner: Ah, got you. Changed. Thanks.

PR #2335, opened 2021-03-11
Improve operator breakdown plot

Bouncner:
This PR extends the `plot_operator_breakdown.py` graph:
- in addition to the relative performance, we now also plot the absolute operator runtimes for each query
- we split the total column from the query overview as we cannot use the same axis for the absolute plot; that keeps it consistent
  - both summary plots look quite the same, but I still like them both as they give a good overview and we can directly plot the values (% and runtimes)
- of course, the numbers have to be taken with a little bit of care since this is a CI run; but similar to the PQP artifacts it's an easy and fast way to get a good insight in a benchmark's performance

Moreover, Jenkins is adapted to no longer run the TPC-H verifications with SF 10 (takes forever with SQLite). To still have SF 10 PQP artifacts, we execute an additional TPC-H SF 10 run.

### Exemplary outputs:
[TPC-H](https://hyrise-ci.epic-hpi.de/job/hyrise/job/hyrise/job/PR-2335/18/artifact/query_plans/tpch/operator_breakdown.pdf)
[TPC-DS](https://hyrise-ci.epic-hpi.de/job/hyrise/job/hyrise/job/PR-2335/13/artifact/query_plans/tpcds/operator_breakdown.pdf)
[JOB - hatched](https://github.com/hyrise/hyrise/files/6135614/operator_breakdown.pdf)

- [x] Check why we build wheels for some very recent packages (e.g., `black`)
- [x] Shadow does not help a lot, see if we can adapt the hatch code
- [x] Change x axis of relative summary to show percentages


Bouncner:
Ready for reviewing.

Jenkinsfile
```diff
@@ -285,20 +285,30 @@ try {
               Utils.markStageSkippedForConditional("memcheckReleaseTest")
             }
           }
-        }, tpchQueryPlansAndVerification: {
-          stage("tpchQueryPlansAndVerification") {
+        }, tpchVerification: {
```
mrks, line 288: (random line)
Just a thought: Should we split up `TableScan` into `TableScan (simple)` and `TableScan (complex)` to make it clearer that those expensive scans are actually ExpressionEvaluator invocations?

scripts/plot_operator_breakdown.py
```diff
@@ -15,6 +16,34 @@
...
+    stack_sum = sum([p.get_height() for p in ax.patches])
+    for patch in ax.patches:
+        value = patch.get_height()
+        if value < (stack_sum * 0.1):  # don't add text labels to tiny bars
+            continue
+        x = patch.get_x() + patch.get_width() / 2
+        y = patch.get_y() + patch.get_height() / 2
+        value_str = format_string.format(value)
+        patch_color = patch.get_facecolor()[:3]
+        ax.text(x, y, value_str, ha="center", color="white" if color_is_dark(patch_color) else "black")
```
mrks, line 45: Also set `va` so that it is centered in the box?

scripts/plot_operator_breakdown.py
```diff
@@ -65,25 +94,70 @@
...
 # Setup colorscheme - using cubehelix, which provides a color mapping that gracefully degrades to grayscale
-colors = sns.cubehelix_palette(n_colors=len(df), rot=2, reverse=True, light=0.9, dark=0.1, hue=1)
+colors = sns.cubehelix_palette(n_colors=len(df_norm), rot=2, reverse=True, light=0.9, dark=0.1, hue=1)
 cmap = LinearSegmentedColormap.from_list("my_colormap", colors)
 
-# Plot it
-ax = df.plot.bar(stacked=True, figsize=(2 + len(df) / 4, 4), colormap=cmap)
+# Factor to adapt the sizes and ratios to large benchmarks with many queries (e.g., Join Order Benchmark).
+# The sqrt() somewhat accomodates that we don't create extremely wide charts but still adapt.
+query_count_apaption = int(round(math.sqrt(len(df_norm))))
```
mrks, line 121: No idea what you are doing here.

scripts/plot_operator_breakdown.py
```diff
@@ -65,25 +94,70 @@
...
+
+fig = plt.figure(figsize=(20 + query_count_apaption, 6))
+plt.subplots(constrained_layout=True)
+# Create a grid with two rows and ensure at least a ratio for the plots of the queries and the summary plot of
+# 4:1. If there are many queries, we increase the share of the query plots.
+gs = GridSpec(2, 5 + query_count_apaption)
```
mrks, line 127: Typo: adaptation

scripts/plot_operator_breakdown.py
```diff
@@ -65,25 +94,70 @@
...
+
+ax_1 = fig.add_subplot(gs[0, :-1])  # Queries, relative
+ax_2 = fig.add_subplot(gs[0, -1])  # Summary, relative
+ax_3 = fig.add_subplot(gs[1, :-1])  # Queries, absolute
+ax_4 = fig.add_subplot(gs[1, -1])  # Summary, absolute
```
mrks, line 132: Why don't you label the variables instead?

scripts/plot_operator_breakdown.py
```diff
@@ -65,25 +94,70 @@
...
+
+df_norm_queries.plot.bar(ax=ax_1, stacked=True, colormap=cmap)
+ax_1.yaxis.set_major_formatter(ticker.PercentFormatter(xmax=1.0))
+ax_1.set_ylabel("Share of run time\n(hiding operators <1%)")
```
mrks, line 136: Nitpick: This is inconsistent with how it is calculated. `1%` is excluded, too.

scripts/plot_operator_breakdown.py
```diff
@@ -102,36 +176,15 @@
...
+    for axis in [ax_1, ax_2, ax_3, ax_4]:
+        column_count = len(axis.get_xticks())
+        hatches = [p for p in patterns for i in range(column_count)]
+        for bar, hatch in zip(reversed(axis.patches), hatches):
+            bar.set_edgecolor(get_hatch_color(bar.get_facecolor()[:3]))
+            bar.set_hatch(hatch)
+            bar.set_linewidth(0)
+
+handles, labels = ax_1.get_legend_handles_labels()
+fig.legend(reversed(handles), reversed(labels), loc=9, ncol=10)
```
mrks, line 188: Seeing it in action, I would rather place the legend at the bottom. What do you think?

Jenkinsfile
Bouncner: You are talking about JOB?
It’s certainly an interesting information, but inner vs semi join is also super interesting or simple aggregates vs complex aggregates (multiple group columns or whatever). So I don’t know if we should really split this rather clear separation into operators and where that would lead to.

mrks: Not only JOB.

The difference between TableScan and the named examples is that in your cases, those are just different variations of the same operator. In the case of the TableScan with the ExpressionEvaluator backend, we are effectively executing an entirely different operator. Remember that an EE scan could internally execute entire query plans. In the past, we have seen a lot of confusion with people wondering why the scan was so expensive.

Bouncner: Hmm, I see your point but I am still not sure where this would end up (personally, I'd be more interested in the joins and the PQPs clearly state the scan implementation). I wouldn't veto though.

scripts/plot_operator_breakdown.py
Bouncner: VLDB Errata? :D

Bouncner: Filtering with `<=` now.

Jenkinsfile
mrks: No strong opinion about this - just a suggestion.

scripts/plot_operator_breakdown.py
mrks: I feel like `<1%` would be more common/expected than `<=1%`

scripts/plot_operator_breakdown.py
Bouncner: Personally, I don't like to see multiple plots without a legend on top. It's different if it is only a single plot. I can live with both positions.

scripts/plot_operator_breakdown.py
Bouncner: Changed the comment. Better?

scripts/plot_operator_breakdown.py
Bouncner: I agree, but it's worse to read on the axis title. :)

Your call.

scripts/plot_operator_breakdown.py
mrks: Won't argue.

scripts/plot_operator_breakdown.py
mrks: I'd leave the label the way it is in the master and adjust the script accordingly.

Bouncner: Hmpf, that's what I meant and what I did (it's actually `>=` in the filtering). Too late ...

scripts/plot_operator_breakdown.py
mrks: I think I already got lost at the first sentence. What sizes and ratios? Is your goal to make the figure wider if we have more queries? If yes, that should be said first. Also, assuming that I understood the purpose correctly, why don't we make this linearly dependent on the number of queries? A wide chart is better than an illegible one, isn't it?

Bouncner: I started with a linear scaling but that was already "suboptimal" for DS, but for JOB it did not work at all (I think). Either you really have tiny bars for "small" workloads as TPC-H with looks a little bit stupid as there is sufficient space. Or you have "nice" TPC-H and you need to zoom into the JOB graph every time to see the results.
That sqrt() somewhat solved that for me and created okay'ish graphs for all three benchmarks.

Bouncner: (I don't want to promise is looks good for an ERP system with thousands of queries though :) )

mrks: > We want the figure to be slightly wider if more queries are printed. `added_figure_width` grows sublinearly, using sqrt just happens to work well for the benchmarks that we tested it with.

mrks (approved):

Jenkinsfile
Bouncner: Open a ticket about it and gather some opinions? I should not be more than the aggregation within the PQP visualization, right?

mrks: Shouldn't be more than two lines or so, correct.

Your call how we want to proceed. Maybe just suggest it in the next fortnightly?

PR #2337, opened 2021-03-14
Enable binary exports/imports for TPC-C

Bouncner:
Enable binary exports/imports for TPC-C, which was the only benchmark left that did not support binary export and imports.

The data creation for TPC-H does not scale linearly and larger tests with 10 warehouses, take a good amount of time.

| Warehouse count | Time to generate data (s) |
| --:            | --: |
| 1 |   6 |
| 2 |  19 |
| 3 |  36 |
| 5 |  86 |
| 10 | 312 |

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -131,7 +131,7 @@ try {
               sh "./scripts/test/hyriseServer_test.py clang-release"
               sh "./scripts/test/hyriseBenchmarkJoinOrder_test.py clang-release"
               sh "./scripts/test/hyriseBenchmarkFileBased_test.py clang-release"
-              sh "cd clang-release && ../scripts/test/hyriseBenchmarkTPCC_test.py ." // Own folder to isolate visualization
+              sh "cd clang-release && ../scripts/test/hyriseBenchmarkTPCC_test.py ." // Own folder to isolate binary export tests
               sh "cd clang-release && ../scripts/test/hyriseBenchmarkTPCH_test.py ." // Own folder to isolate visualization
 
             } else {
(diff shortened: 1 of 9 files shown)


scripts/plot_operator_breakdown.py
```diff
@@ -76,7 +76,7 @@
 print(df)
 
 # Drop all operators that do not exceed 1% in any query
-df = df[df > 0.01].dropna(axis="columns", how="all")
+df = df[df >= 0.01].dropna(axis="columns", how="all")
```
mrks, line 79: ?

src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -354,4 +356,25 @@ bool AbstractTableGenerator::_all_chunks_sorted_by(const std::shared_ptr<Table>&
   return true;
 }
 
+std::unordered_map<std::string, BenchmarkTableInfo> AbstractTableGenerator::_load_binary_tables_from_path(
+    std::string cache_directory) {
```
mrks, line 360: const ref?

mrks (approved):

scripts/plot_operator_breakdown.py
Bouncner: Whooopsie. It really confused me that I needed to make this change twice. That explains it .... will revert.

Jenkinsfile
```diff
@@ -131,7 +131,7 @@ try {
               sh "./scripts/test/hyriseServer_test.py clang-release"
               sh "./scripts/test/hyriseBenchmarkJoinOrder_test.py clang-release"
               sh "./scripts/test/hyriseBenchmarkFileBased_test.py clang-release"
-              sh "./scripts/test/hyriseBenchmarkTPCC_test.py clang-release"
+              sh "cd clang-release && ../scripts/test/hyriseBenchmarkTPCC_test.py ." // Own folder to isolate visualization
```
mrks, line 134: to isolate export tests?

Bouncner: This time I can't even blame on the time of day. :(

PR #2338, opened 2021-03-15
Better error message for --verify

### LINKED ISSUE #2295: --verify fails if run from other directory
```
-  Adding indexes to SQLite table regionterminate called after throwing an instance of 'std::logic_error'
  what():  src/lib/utils/sqlite_wrapper.cpp:233 Failed to execute query (INSERT INTO "region" SELECT * FROM region_unindexed). SQL error: no such table: region
```

`add_indices_to_sqlite` should assert that the passed file names are valid.

mrks:
fixes #2295

### DIFF
--- a/src/lib/utils/sqlite_add_indices.cpp
+++ b/src/lib/utils/sqlite_add_indices.cpp
@@ -36,12 +36,17 @@ void add_indices_to_sqlite(const std::string& schema_file_path, const std::strin
 
   // Recreate tables using the passed schema sql file
   std::ifstream schema_file(schema_file_path);
+  Assert(schema_file.good(), std::string{"Schema file "} + schema_file_path +
+                                 " not found - are you calling the binary from a different working directory?");
   std::string schema_sql((std::istreambuf_iterator<char>(schema_file)), std::istreambuf_iterator<char>());
   sqlite_wrapper->main_connection.raw_execute_query(schema_sql);
 
   // If indices are not part of the schema file, add them here
   if (!create_indices_file_path.empty()) {
     std::ifstream create_indices_file(create_indices_file_path);
+    Assert(create_indices_file.good(),
+           std::string{"Index file "} + create_indices_file_path +
+               " not found - are you calling the binary from a different working directory?");
     std::string create_indices_sql((std::istreambuf_iterator<char>(create_indices_file)),
                                    std::istreambuf_iterator<char>());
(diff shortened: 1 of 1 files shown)


src/lib/utils/sqlite_add_indices.cpp
```diff
@@ -36,12 +36,17 @@ void add_indices_to_sqlite(const std::string& schema_file_path, const std::strin
 
   // Recreate tables using the passed schema sql file
   std::ifstream schema_file(schema_file_path);
+  Assert(schema_file.good(), std::string{"Schema file "} + schema_file_path +
+                                 " not found - are you calling the binary from a different working directory?");
```
Bouncner, line 40: Do we need to call from the binary root? 

“Different work dir” is not clear to me

mrks: Doesn't really matter, as long as the file exists. This could be the root dir or the build dir.

I believe that the full message is clearer, but maybe you have a better idea:

> libc++abi.dylib: terminating with uncaught exception of type std::logic_error: src/lib/utils/sqlite_add_indices.cpp:40 Schema file resources/benchmark/tpch/schema.sql not found - are you calling the binary from a different working directory?

Bouncner: Even this confuses me. The working directory is the one I am currently in. Can I be in another working directory? Or is that supposed to also cover cases of sub shells etc.?

Maybe we can directly say, that we cannot access the `resources` directory which we expect to be in the "working directory"?

mrks: Mmh, but this file does not really know anything about `resources/`.

Maybe "try running the binary from the Hyrise root or the build directory"?

mrks: (btw, *working directory* is a POSIX term: https://www.gnu.org/software/libc/manual/html_node/Working-Directory.html)

Bouncner:
I don't find any information on the Posix site that would help me understand the "different working dir" information.

Anyhow, I like your "try running the binary from the Hyrise root or the build directory". It's much clearer to me and tells me what to do.

mrks:
> I don't find any information on the Posix site that would help me understand the "different working dir" information.

The idea was to say "maybe you are in the wrong working directory - try going somewhere else" without specifying where that might be.

> Anyhow, I like your "try running the binary from the Hyrise root or the build directory". It's much clearer to me and tells me what to do.

Changed.

PR #2339, opened 2021-03-17
Added mac-arm to Jenkinsfile

### LINKED ISSUE #2342: Remove 'cmake/EmbedLLVM.cmake'?
Found it while looking at the current  M1 build problems.

Remove it? As far as I understand the comments, it's a relict of the JIT times.

mweisgut:
Posted by @mrks:
- [x] Avoid code duplication between the two Mac configurations.
- [x] Run the Mac configurations in parallel.
- [x] Doc the Java/Rosetta thing. Maybe reference @mrks' [SO post](https://stackoverflow.com/questions/66705777/jenkins-does-not-work-properly-on-mac-m1-agent/66705778#66705778).
- [x] Should Jenkins verify that ARM binaries are built?

fixes #2342


mrks:
@Bouncner - I believe that it has *something* to do with the environment used by Jenkins. Not sure what though.

Bouncner:
I don't want to fiddle with the files while you guys are working on it. Have you tried to add `LDFLAGS` etc. to the `cmake` as well?

However, no idea how to fix this problem properly.

mrks:
I'm not touching it at the moment. Haven't looked at LDFLAGS yet - I am still surprised that it works with a regular ssh connection but not if Jenkins opens the ssh connection.

mrks:
@Bouncner - you can also set environment variables in the configuration of the Jenkins node.

Bouncner:
I set a few but it does not make a change. I don't know what else it could be. :-/

Bouncner:
We could use our own FindPostgreSQL.cmake with the added /opt paths, but that does not sound clean to me. And I am surprised nobody else has this issue with M1 builds (at least google does not help me here).

mrks:
> I set a few but it does not make a change. I don't know what else it could be. :-/

Did you set PATH etc. in Jenkins?

> We could use our own FindPostgreSQL.cmake with the added /opt paths

Have you tested setting `-DPOSTGRESQL_BIN=...`?

> And I am surprised nobody else has this issue with M1 builds (at least google does not help me here).

It's even more surprising that it works outside of Jenkins.

Bouncner:
> Did you set PATH etc. in Jenkins?

In the admin backend as a path for the Jenkins slave.
 
> Have you tested setting `-DPOSTGRESQL_BIN=...`?

Not bin, but the library dir. If we execute in regularly, cmake finds `/opt/homebrew/lib/libpq.dylib`. Do we also need the binary?

> It's even more surprising that it works outside of Jenkins.

Are we using a non-standard way to run the CI agent on mac?

mrks:
Screenshot 2021-03-19 at 10 03 18

Good chance that it has something to do with Rosetta. I just remembered that we had to install Java (wurgs). Can we install a Rosetta version of Postgres?

Bouncner:
We could use these two aliases here (https://stackoverflow.com/a/65424574/1147726, I used them a couple of weeks ago and they worked as expected) to remove Postgres and reinstall with Intel.

Bouncner:
No with pretty much all the Postgres paths given to Cmake, it no longer complains about Postgres. Now it's TBB (should be fixable, that's our own cmake file).

The problem with the (potential) rosetta fix, it's not what actual M1 users will do.

mrks:
> We could use these two aliases here (https://stackoverflow.com/a/65424574/1147726, I used them a couple of weeks ago and they worked as expected) to remove Postgres and reinstall with Intel.

This was a good hint. Using `arch` does the trick with virtually no additional hacks needed. We still have some boost warnings, but at least the tests are executed.

I'll have to work on my other things now; feel free to look into this.

mrks:
Even better, by installing a different (native) JDK, I was able to fix almost all problems.

Bensk1:
@Bouncner any updates on that one?

Bouncner:
I think @mrks fix with the JVM hickup was everything we needed. The only thing left might be how to get this running in parallel. The last comments here sounded like it's not easy to do with Jenkins, but I have absolutely no idea. Never worked with Jenskins.

Bouncner:
> Should Jenkins verify that ARM binaries are built?

I'd say yes.

mrks:
> The only thing left might be how to get this running in parallel.

You mean M1 Mac jobs running in parallel with x86 jobs?

Bouncner:
> > The only thing left might be how to get this running in parallel.
> 
> You mean M1 Mac jobs running in parallel with x86 jobs?

I was referring to this comment `  // I have not found a nice way to run this in parallel with the steps above, as those are in a "docker.inside" block and this is not.`.
I don't know what it means or on which level Jenkins parallelizes which steps. :-/

mrks:
That refers to running M1 together with the Linux-based tests. Running M1 in parallel with x86 should be way easier, just copy one of the `parallel {}` blocks as a template.

Bensk1:
> * [ ]  Avoid code duplication between the two Mac configurations.

I don't know if we can save a lot here due to the fact that the code runs on two different nodes and compiler versions and paths vary.

> * [ ]  Doc the Java/Rosetta thing. Maybe reference @mrks' [SO post](https://stackoverflow.com/questions/66705777/jenkins-does-not-work-properly-on-mac-m1-agent/66705778#66705778).

What/where could we document that?

> * [ ]  Should Jenkins verify that ARM binaries are built?

I verified that locally, I believe we don't have to do it in Jenkins then, what do you think?

Apart from that, it's working now.

Bensk1:
@mweisgut I cannot assign you for a review since you opened the PR. Could you have a look?

Jenkinsfile
```diff
@@ -3,6 +3,11 @@ import org.jenkinsci.plugins.pipeline.modeldefinition.Utils
 full_ci = env.BRANCH_NAME == 'master' || pullRequest.labels.contains('FullCI')
 tests_excluded_in_sanitizer_builds = '--gtest_filter=-SQLiteTestRunnerEncodings/*:TPCDSTableGeneratorTest.GenerateAndStoreRowCounts:TPCHTableGeneratorTest.RowCountsMediumScaleFactor'
 
+unity = '-DCMAKE_UNITY_BUILD=ON'
```
Bouncner, line 6: It's defined again later. Can you remove that one?

Bouncner: Same for the other ones.

Bouncner (approved):
Just a small thing. Looks good to me, but I am not experienced with Jenkins.

Jenkinsfile
```diff
@@ -12,7 +17,7 @@ try {
 
       if (jenkinsUserName != "admin" && env.BRANCH_NAME != "master") {
         try {
-          withCredentials([usernamePassword(credentialsId: 'github', usernameVariable: 'GITHUB_USERNAME', passwordVariable: 'GITHUB_TOKEN')]) {
+          withCredentials([usernamePassword(credentialsId: '5fe8ede9-bbdb-4803-a307-6924d4b4d9b5', usernameVariable: 'GITHUB_USERNAME', passwordVariable: 'GITHUB_TOKEN')]) {
```
Bensk1, line 20: @mweisgut do you know anything about this change?

Bouncner: Nope, no idea.

mweisgut: Me neither

Bensk1: Looking through Jenkins, I believe `GitHub` is correct.

Bensk1:
> > * [ ]   Avoid code duplication between the two Mac configurations.
> 
> I don't know if we can save a lot here due to the fact that the code runs on two different nodes and compiler versions and paths vary.
> 
> > * [ ]   Doc the Java/Rosetta thing. Maybe reference @mrks' [SO post](https://stackoverflow.com/questions/66705777/jenkins-does-not-work-properly-on-mac-m1-agent/66705778#66705778).
> 
> What/where could we document that?
> 
> > * [ ]   Should Jenkins verify that ARM binaries are built?
> 
> I verified that locally, I believe we don't have to do it in Jenkins then, what do you think?
> 
> Apart from that, it's working now.

@mweisgut do you want to comment on that? If not, I will merge this as soon as the CI is green.

mweisgut:
> > * [ ]   Avoid code duplication between the two Mac configurations.
> 
> I don't know if we can save a lot here due to the fact that the code runs on two different nodes and compiler versions and paths vary.

Agreed. 
> > * [ ]   Doc the Java/Rosetta thing. Maybe reference @mrks' [SO post](https://stackoverflow.com/questions/66705777/jenkins-does-not-work-properly-on-mac-m1-agent/66705778#66705778).
> 
> What/where could we document that?

The referenced SO post suggests two solutions: `Preferred Option (install native JDK)`, `Hotfix (without changing the JDK)`. I could not find the hotfix solution in the diff of this PR, so I assume we followed the preferred option. A comment that the native JDK is required on the CI machine could be helpful in the `clangReleaseMacArm` code branch.

> > * [ ]   Should Jenkins verify that ARM binaries are built?
> 
> I verified that locally, I believe we don't have to do it in Jenkins then, what do you think?
> 
> Apart from that, it's working now.

Do we verify that x64 binaries are built? Iff so, I suggest being consistent and also check it for ARM binaries.

Bensk1:
> The referenced SO post suggests two solutions: `Preferred Option (install native JDK)`, `Hotfix (without changing the JDK)`. I could not find the hotfix solution in the diff of this PR, so I assume we followed the preferred option. A comment that the native JDK is required on the CI machine could be helpful in the `clangReleaseMacArm` code branch.

Done.

> Do we verify that x64 binaries are built? Iff so, I suggest being consistent and also check it for ARM binaries.

No, we don't.

mrks:
>> Do we verify that x64 binaries are built? Iff so, I suggest being consistent and also check it for ARM binaries.

> No, we don't.

Why would we? There is no risk of an x64 machine suddenly building a different binary. For M1, we accidentally could end up with an x64 binary that is executed via Rosetta. Checking that we actually built an ARM binary makes sense to me.

Bensk1:
> > > Do we verify that x64 binaries are built? Iff so, I suggest being consistent and also check it for ARM binaries.
> 
> > No, we don't.
> 
> Why would we? There is no risk of an x64 machine suddenly building a different binary. 

I agree, that was a "No, we don't [verify that x64 binaries are built and don't have to]."

> For M1, we accidentally could end up with an x64 binary that is executed via Rosetta. Checking that we actually built an ARM binary makes sense to me.

As far as I can tell we would need to explicitly configure the build for a non-ARM binary on the M1 system for that to happen. We could still check for that. For `sysctl.proc_translated` we would need to integrate that into Hyrise since it must be run from the process. [Apparently](https://apple.stackexchange.com/questions/407467/how-to-determine-if-a-macos-program-on-apple-silicon-is-emulated-x86-or-native-a), `file` on the binary is sufficient.

mrks:
Yes, but we should check for the return value of file, not just print it.

Bensk1:
> Yes, but we should check for the return value of file, not just print it.

Sure!

PR #2341, opened 2021-03-17
Update Contributor guidelines

### LINKED ISSUE #2190: Clean up contributor guidelines
@julianmenzler replaced some `#if HYRISE_DEBUG` with `if constexpr ...`. He also changed the `DebugAssert`s within to regular asserts. I think regular asserts make sense if we already have the debug check above.

Didn't want to add it to CONTRIBUTING.md just yet, as I feel like the list is becoming harder to read and to remember. Should we clean it up a bit and try to put the Top 10 rules on the top?

mrks:
@Bensk1 and I tried to clean up the contributor guidelines. Our goal was to move the most important aspects to top and to add some more structure. Some minor points have been removed. As the retentiveness when reading the guidelines is limited, we need to balance the desire to have guidelines for all situations with the conciseness of the document.

fixes #2190 

- [x] Test if clang-tidy covers the following cases - if yes, we do not need to list them here:
  - [x] Use override/final for the respective methods, do not repeat virtual
  > Clang complains, gcc doesn't. Good enough:
  > `/var/lib/jenkins/workspace/hyrise_hyrise_PR-2341/src/lib/expression/abstract_predicate_expression.hpp:21:8: error: '_shallow_equals' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]`
  - [x] Add a virtual destructor for abstract classes
  > Clang complains, gcc doesn't. Good enough:
  > `/var/lib/jenkins/workspace/hyrise_hyrise_PR-2341/src/lib/cache/abstract_cache.hpp:14:7: error: 'opossum::AbstractCache<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::shared_ptr<opossum::AbstractLQPNode> >' has virtual functions but non-virtual destructor [-Werror,-Wnon-virtual-dtor]`
  - [x] Check if this is also covered in hpp files
- [x] Update README
  - [x] Mention --gtest_filter
- [x] Add line breaks to this document so that it is easier to read in a text editor. Do this after we have reached consensus on the text

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@
    you. Try to improve it as part of your PRs and do not hesitate to ask if anything is unclear. Chances are that it can
    be improved.
 4. Thou shalt properly test thy code. This includes unit *and* integration tests. Try to isolate parts that can be
-   individually tested.
+   independently tested.
 
 # C++
 * Use automatic memory management (RAII, smart pointers). `new` and `malloc` are evil words.
@@ -29,7 +29,7 @@
   * Use forward declarations instead of full header includes wherever possible.
 
 * Loops
-  * Use range-based for loops when possible: `for (const auto& item : items) {...}`.
+  * Use range-based `for` loops when possible: `for (const auto& item : items) {...}`.
   * If you have to use old-style loops, keep in mind that the loop condition is evaluated every time: Instead of
     `for (auto offset = size_t{0}; offset < something.size(); ++offset)`, the size should be retrieved just once. See
     also [this document](http://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop).
@@ -47,7 +47,7 @@
       from Noncopyable to avoid these potentially expensive copies.
 
 * Miscellaneous
-  * Prefer `if (object) {` over `if (object != nullptr) {` or `if (object.has_value()) {`.
+  * Prefer `if (object)` over `if (object != nullptr)` or `if (object.has_value())`.
   * Don't write `this->` if you don't have to.
   * Be explicit with types: Use [u]int(8|16|32|64)_t instead of `int, long, uint` etc.
   * Use [auto-to-stick](https://www.fluentcpp.com/2018/09/28/auto-stick-changing-style/): `auto x = 17;` or
@@ -60,7 +60,7 @@
   clang-tidy is a great help, do not rely on it.
 * Call ./scripts/format.sh before pushing.
 * Choose clear and concise names, and avoid, e.g., `i`, `j`, `ch_ptr`.
-* Formatting details: 2 spaces, 120 columns, comments above code.
+* Formatting details: 2 spaces for indentation, 120 columns, comments above code.
 * Use empty lines to structure your code.
 * Naming conventions:
     * Files: lowercase separated by underscores, e.g., abstract_operator.cpp, usually corresponding to a class, e.g.,
@@ -80,7 +80,7 @@
 * Maintain correct orthography and grammar. Comments should start with a capital letter, sentences should be finished
   with a full stop.
   * Class names within comments are written in PascalCase - e.g., "As the input may be a ReferenceSegment, a valid RowID
-    may *point to* a row that is NULL."
+    may point to a row that is NULL."
 
 # Pull Requests
 ## Opening PRs
@@ -89,7 +89,7 @@
     costs.
   * We do not do this automatically as the CI server is not sufficiently isolated and the performance results would
     vary. Similarly, your personal laptop is likely to produce unreliable results.
-* If your PR is related to an existing issue, reference it in the PR's description (e.g., `fixes #123`).
+* If your PR is related to an existing issue, reference it in the PR's description (e.g., `fixes #123` or `refs #123`).
 * If you are not a member of the Hyrise organization, your PR will not be built by our CI server. Contact a maintainer
   for this. They can add you to the organization or manually trigger builds from within Jenkins.
 * For your PR to be merged, it must pass the FullCI run. Set the FullCI tag in Github before committing to trigger the
@@ -98,9 +98,9 @@
   message of the squash merge should NOT contain the individual commit messages from your branch.
 
 ## Reviewing PRs
-* Check if the PR sufficiently adds tests both for happy and unhappy cases.
 * Keep the whole picture in mind. Often, it makes sense to make two passes: One for the code style and line-level
   modifications; one for checking how it fits into the overall picture.
+* Check if the PR sufficiently adds tests both for happy and unhappy cases.
 * Do not shy away from requesting changes on surrounding code that was not modified in the PR. Remember that after a PR,
   the code base should be better than before.
 * Verify the CI results, including the change in procentual coverage, and check if the compile time or query performance
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,8 @@
     `auto y = std::vector<size_t>{};`.
   * Namespaces: Do not create nested namespaces, do not import namespaces.
   * Consider structured bindings: `const auto& [iterator, added] = unordered_map.emplace(...);`
+  * Prefer braced control statements (`if (...) { x(); }` over `if (...) x();`). For multi-line statements, this is
+    mandatory.
 
 # Formatting and Naming
 * Much of this is enforced by clang-tidy. However, clang-tidy does not yet cover hpp files (see #1901). Also, while
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 1. Thy code shalt be the primary method of documentation. Part of this is to choose concise but descriptive names.
    Comments should be used to explain the concept and usage of classes (in the hpp file) and the structure of the
    algorithms (in the implementation).
-2. Thou shalt program defensively. Use Assert wherever it makes sense, use DebugAssert in performance-critical parts
+2. Thou shalt program defensively. Use `Assert` wherever it makes sense, use `DebugAssert` in performance-critical parts
    (e.g., within hot loops). Also, we do not handle exceptions in Hyrise. If an invalid state is reached, we crash
    immediately. This makes debugging easier. An exception to this is user-facing code where we handle, e.g., typos in
    SQL queries.
@@ -16,7 +16,7 @@
 * Use automatic memory management (RAII, smart pointers). `new` and `malloc` are evil words.
 * Be mindful of ownership. Not everything needs to be a smart pointer. Consider passing around references to the object
   or references to a shared_ptr instead of copying the shared_ptr. Remember that this might not be safe when passing
-  shared_ptrs into, i.e., JobTasks.
+  shared_ptrs into, e.g., JobTasks.
 * Use `const` whenever possible. Consider variables, methods, pointers, and their pointees.
 
 * Header files
@@ -38,13 +38,13 @@
   * When creating a vector where you know the size beforehand, use `reserve` to avoid unnecessary resizes and
     allocations.
   * Hash-based data structures are usually faster than tree-based data structures. Unless you have a reason to use the
-    latter, prefer unordered_(map|set) over map and set.
+    latter, prefer `unordered_(map|set)` over `map` and `set`.
 
 * Copies
-    * C++ makes it too easy to inadvertently copy objects.
-    * For larger elements (e.g., vectors), pass a (const) reference instead.
-    * When your implemented class does not need to be copied (e.g., a Table should never exist more than once), inherit
-      from Noncopyable to avoid these potentially expensive copies.
+    * Avoid unnecessary copies, C++ makes it too easy to inadvertently copy objects.
+    * For larger elements (e.g., vectors), pass a (`const`) reference instead.
+    * If your implemented class does not need to be copied (e.g., a `Table` should never exist more than once), inherit
+      from `Noncopyable` to avoid these potentially expensive copies.
 
 * Miscellaneous
   * Prefer `if (object)` over `if (object != nullptr)` or `if (object.has_value())`.
@@ -60,7 +60,7 @@
 # Formatting and Naming
 * Much of this is enforced by clang-tidy. However, clang-tidy does not yet cover hpp files (see #1901). Also, while
   clang-tidy is a great help, do not rely on it.
-* Call ./scripts/format.sh before pushing.
+* Call ./scripts/format.sh before committing your code.
 * Choose clear and concise names, and avoid, e.g., `i`, `j`, `ch_ptr`.
 * Formatting details: 2 spaces for indentation, 120 columns, comments above code.
 * Use empty lines to structure your code.
@@ -87,14 +87,14 @@
 # Pull Requests
 ## Opening PRs
 * When you submit a non-trivial PR, include the results of benchmark_all.sh.
-  * These help in understanding potential performance changes as well as document potential changes to the compilation
+  * These results help in understanding potential performance changes as well as document potential changes to the compilation
     costs.
   * We do not do this automatically as the CI server is not sufficiently isolated and the performance results would
     vary. Similarly, your personal laptop is likely to produce unreliable results.
 * If your PR is related to an existing issue, reference it in the PR's description (e.g., `fixes #123` or `refs #123`).
 * If you are not a member of the Hyrise organization, your PR will not be built by our CI server. Contact a maintainer
   for this. They can add you to the organization or manually trigger builds from within Jenkins.
-* For your PR to be merged, it must pass the FullCI run. Set the FullCI tag in Github before committing to trigger the
+* For your PR to be merged, it must pass a FullCI run. Set the FullCI tag in GitHub before committing to trigger the
   complete (but more expensive!) run.
 * When merging your PR, copy your PR description (excluding the benchmark results) into the commit message. The commit
   message of the squash merge should NOT contain the individual commit messages from your branch.
@@ -105,6 +105,6 @@
 * Check if the PR sufficiently adds tests both for happy and unhappy cases.
 * Do not shy away from requesting changes on surrounding code that was not modified in the PR. Remember that after a PR,
   the code base should be better than before.
-* Verify the CI results, including the change in procentual coverage, and check if the compile time or query performance
+* Verify the CI results, including how the reported coverage changed, and check if the compile time or query performance
   have been negatively affected. For this, the author should have posted the results of benchmark_all.sh into the PR
   description.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,11 +1,12 @@
 # The Four Commandments
-1. Thy code (or the beginning of the header file) shalt be the primary method of documentation. Comments should be used to explain higher level concepts, but primarily, the code should be explaining itself. Part of this is to choose concise, descriptive names.
-1. Thou shalt program defensively. Use Assert wherever it makes sense, use DebugAssert in performance-critical parts (e.g., within hot loops). Also, we do not handle exceptions in Hyrise. If an invalid state is reached, we crash immediately. This makes debugging easier. An exception (hihi) to this is user-facing code where we handle, e.g., typos in SQL queries.
-1. Thou shalt reflect on existing code. Just as your code is not perfect, neither is the code that people wrote before you. Try to improve it as part of your PRs and do not hesitate to ask if anything is unclear. Chances are that it can be improved.
-1. Thou shalt properly test thy code. This includes unit *and* integration tests. Try to isolate parts that can be individually tested.
+1. Thy code shalt be the primary method of documentation. Part of this is to choose concise but descriptive names. Comments should be used to explain the concept and usage of classes (in the hpp file) and the structure of the algorithms (in the implementation).
+2. Thou shalt program defensively. Use Assert wherever it makes sense, use DebugAssert in performance-critical parts (e.g., within hot loops). Also, we do not handle exceptions in Hyrise. If an invalid state is reached, we crash immediately. This makes debugging easier. An exception (hihi) to this is user-facing code where we handle, e.g., typos in SQL queries.
+3. Thou shalt reflect on existing code. Just as your code is not perfect, neither is the code that people wrote before you. Try to improve it as part of your PRs and do not hesitate to ask if anything is unclear. Chances are that it can be improved.
+4. Thou shalt properly test thy code. This includes unit *and* integration tests. Try to isolate parts that can be individually tested.
 
 # C++
 * Use automatic memory management (RAII, smart pointers). `new` and `malloc` are evil words.
+* Be mindful of ownership. Not everything needs to be a smart pointer. Consider passing around references to the object or references to a shared_ptr instead of copying the shared_ptr. Remember that this might not be safe when passing shared_ptrs into, i.e., JobTasks.
 * Use `const` whenever possible. Consider variables, methods, pointers, and their pointees.
 * Reduce the size of your hpp files, both in terms of the number of lines and the code complexity. This keeps the compilation duration low.
   * Code in hpp files is compiled for every cpp file that includes the header. As such, move code to cpp files where possible. This often includes templated classes, where it is sometimes possible to implement their code in cpp files.
@@ -13,7 +14,7 @@
   * Use forward declarations instead of full header includes wherever possible.
 * Loops
   * Use range-based for loops when possible: `for (const auto& item : items) {...}`.
-  * If you have to use old-style loops, keep in mind that the loop condition is evaluated every time: Instead of `for (auto i = 0; i < expensive_function(); ++i)`, the comparison value should be calculated just once. http://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop
+  * If you have to use old-style loops, keep in mind that the loop condition is evaluated every time: Instead of `for (auto offset = size_t{0}; offset < something.size(); ++offset)`, the size should be retrieved just once. See also [here](http://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop)
 * Data structures
   * When creating a vector where you know the size beforehand, use `reserve` to avoid unnecessary resizes and allocations.
   * Hash-based data structures are usually faster than tree-based data structures. Unless you have a reason to use the latter, prefer unordered_(map|set) over map and set.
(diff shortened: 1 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -1,73 +1,76 @@
...
+
+# C++
+* Use automatic memory management (RAII, smart pointers). `new` and `malloc` are evil words.
+* Use `const` whenever possible. Consider variables, methods, pointers, and their pointees.
+* Reduce the size of your hpp files, both in terms of the number of lines and the code complexity. This keeps the compilation duration low.
+  * Code in hpp files is compiled for every cpp file that includes the header. As such, move code to cpp files where possible. This often includes templated classes, where it is sometimes possible to implement their code in cpp files.
+  * Anonymous namespaces are a good way to define local helper methods.
+  * Use forward declarations instead of full header includes wherever possible.
+* Loops
+  * Use range-based for loops when possible: `for (const auto& item : items) {...}`.
```
Bouncner, line 15: Since I remarked this quite often the last weeks: also name structured bindings?

CONTRIBUTING.md
```diff
@@ -1,73 +1,76 @@
...
+# The Four Commandments
+1. Thy code (or the beginning of the header file) shalt be the primary method of documentation. Comments should be used to explain higher level concepts, but primarily, the code should be explaining itself. Part of this is to choose concise, descriptive names.
+1. Thou shalt program defensively. Use Assert wherever it makes sense, use DebugAssert in performance-critical parts (e.g., within hot loops). Also, we do not handle exceptions in Hyrise. If an invalid state is reached, we crash immediately. This makes debugging easier. An exception (hihi) to this is user-facing code where we handle, e.g., typos in SQL queries.
+1. Thou shalt reflect on existing code. Just as your code is not perfect, neither is the code that people wrote before you. Try to improve it as part of your PRs and do not hesitate to ask if anything is unclear. Chances are that it can be improved.
+1. Thou shalt properly test thy code. This includes unit *and* integration tests. Try to isolate parts that can be individually tested.
+
+# C++
+* Use automatic memory management (RAII, smart pointers). `new` and `malloc` are evil words.
```
Bouncner, line 8: Since we had a ton of (unnecessary) shared_ptr's in the sort merge join, I am wondering if there good guidelines by Herb and co. when to use them and when not. I haven't seen this overuse lately though.

README.md
```diff
@@ -122,6 +122,7 @@ To configure a build directory for a release build make sure it is empty and cal
 ### Test
 Calling `make hyriseTest` from the build directory builds all available tests.
 The binary can be executed with `./<YourBuildDirectory>/hyriseTest`.
+Subsets of all available tests can be selected via `--gtest_filter=`.
 Note, that the tests/sanitizers/etc need to be executed from the project root in order for table files to be found.
```
Bouncner, line 126: The tests run just fine from the build dir, right? Is that outdated?

Bouncner, line 126: The tests run fine from the build dir. Is that comment outdated?

CONTRIBUTING.md
```diff
@@ -1,73 +1,76 @@
...
+## Opening PRs
+* When you submit a non-trivial PR, include the results of benchmark_all.sh.
+  * These help in understanding potential performance changes as well as document potential changes to the compilation costs.
+  * We do not do this automatically as the CI server is not sufficiently isolated and the performance results would vary. Similarly, your personal laptop is likely to produce unreliable results.
+* If your PR is related to an existing issue, reference it in the PR's description (e.g., `fixes #123`).
+* If you are not a member of the Hyrise organization, your PR will not be built by our CI server. Contact a maintainer for this. They can add you to the organization or manually trigger builds from within Jenkins.
+* When merging your PR, copy your PR description (excluding the benchmark results) into the commit message. The commit message of the squash merge should NOT contain the individual commit messages from your branch.
+* For your PR to be merged, it must pass the FullCI run. Set the FullCI tag in Github before committing to trigger the complete (but more expensive!) run.
+
+## Reviewing PRs
```
Bouncner, line 72: PRs > Pull Requests

CONTRIBUTING.md
```diff
@@ -1,73 +1,76 @@
...
+# C++
+* Use automatic memory management (RAII, smart pointers). `new` and `malloc` are evil words.
+* Use `const` whenever possible. Consider variables, methods, pointers, and their pointees.
+* Reduce the size of your hpp files, both in terms of the number of lines and the code complexity. This keeps the compilation duration low.
+  * Code in hpp files is compiled for every cpp file that includes the header. As such, move code to cpp files where possible. This often includes templated classes, where it is sometimes possible to implement their code in cpp files.
+  * Anonymous namespaces are a good way to define local helper methods.
+  * Use forward declarations instead of full header includes wherever possible.
+* Loops
+  * Use range-based for loops when possible: `for (const auto& item : items) {...}`.
+  * If you have to use old-style loops, keep in mind that the loop condition is evaluated every time: Instead of `for (auto i = 0; i < expensive_function(); ++i)`, the comparison value should be calculated just once. http://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop
```
Bouncner, line 16: `i = 0` -> `i = uint_32{0}`?

mrks: Haha, when typing this I was wondering if this would come up. I am not a big fan of the "int vs int32_t" rule anyway. We know that an int is at least 32 bits, so I don't see how this makes any difference. I wonder if this rule was established before our compiler settings complained about mismatching types.

Bouncner: I am not well enough informed to discuss this. For me, being precise makes a lot of sense and it' supposed to be safer as we deploy to ARM, IBM, and Intel. I guess. But I might be totally wrong here. :)

If it does not matter, we should remove the other recommendation to use the more precise types.

CONTRIBUTING.md
Bouncner: Haven't seen the title above where we write "Pull Request". Maybe just drop the "PRs"?

CONTRIBUTING.md
Bensk1: > `i = 0` -> `i = uint_32{0}`?

Maybe we should also get rid of the `i` as we request to avoid it in the same document.

CONTRIBUTING.md
```diff
@@ -1,73 +1,76 @@
...
-	- Did the author update documentation and dependencies (Wiki, README.md, DEPENDENCIES.md, Dockerfile, install_dependencies.sh, Jenkinsfile)
+# The Four Commandments
+1. Thy code (or the beginning of the header file) shalt be the primary method of documentation. Comments should be used to explain higher level concepts, but primarily, the code should be explaining itself. Part of this is to choose concise, descriptive names.
+1. Thou shalt program defensively. Use Assert wherever it makes sense, use DebugAssert in performance-critical parts (e.g., within hot loops). Also, we do not handle exceptions in Hyrise. If an invalid state is reached, we crash immediately. This makes debugging easier. An exception (hihi) to this is user-facing code where we handle, e.g., typos in SQL queries.
+1. Thou shalt reflect on existing code. Just as your code is not perfect, neither is the code that people wrote before you. Try to improve it as part of your PRs and do not hesitate to ask if anything is unclear. Chances are that it can be improved.
+1. Thou shalt properly test thy code. This includes unit *and* integration tests. Try to isolate parts that can be individually tested.
```
julianmenzler, line 5: I never read "Thy" and "Thou" before.

julianmenzler: "(hihi)"?

CONTRIBUTING.md
```diff
@@ -1,73 +1,76 @@
...
+* Much of this is enforced by clang-tidy. However, clang-tidy does not yet cover hpp files (see #1901). Also, while clang-tidy is a great help, do not rely on it.
+* Call ./scripts/format.sh before pushing.
+* Choose clear and concise names, and avoid, e.g., `i`, `j`, `ch_ptr`.
+* Formatting details: 2 spaces, 120 columns, comments above code.
+* Use empty lines to structure your code.
+* Naming conventions:
+
+    * Files: lowercase separated by underscores, e.g., abstract_operator.cpp, usually corresponding to a class, e.g., AbstractOperator.
+
+    * Types (classes, structs, enums, typedefs, using): CamelCase starting with uppercase letter, e.g., `TableScan`.
```
julianmenzler, line 45: I thought this naming style was referred to as PascalCase, not camelCase (https://en.wikipedia.org/wiki/Camel_case).

CONTRIBUTING.md
```diff
@@ -1,73 +1,76 @@
...
+    * Private / protected members / methods: like variables / functions with leading underscore, e.g., `_on_execute()`.
+
+    * Classes that are used only to have a non-templated base class are named `BaseXY` (e.g., `BaseValueSegment`, while classes that have multiple differing implementations are named `AbstractXY` (e.g., `AbstractOperator`).
+
+    * In cases where a constructor parameter would have the same name as the member it initializes, prefix it with `init`: `C(int init_foo) : foo(init_foo) {}`.
+
+    * If an identifier contains a verb or an adjective in addition to a noun, the schema [verb|adjective]\[noun] is preferred, e.g., use `left_input` rather than ~~`input_left`~~ and `set_left_input()` rather than ~~`set_input_left()`~~.
+
+* Maintain correct orthography and grammar. Comments should start with a capital letter, sentences should be finished with a full stop.
+  * Class names within comments are written in CamelCase - e.g., "As the input may be a ReferenceSegment, a valid RowID may *point to* a row that is NULL."
```
julianmenzler, line 60: Maybe add that we use `//` for single line comments, but prefer

```cpp
/**
 * Topic XY:
 *  ...
 */
```

for comment blocks.

julianmenzler (commented):
Nice work!

CONTRIBUTING.md
```diff
@@ -1,73 +1,76 @@
...
+* Miscellaneous
+  * Prefer `if (object) {` over `if (object != nullptr) {` or `if (object.has_value()) {`.
+  * Don't write `this->` if you don't have to.
+  * Be explicit with types: Use [u]int(8|16|32|64)_t instead of `int, long, uint` etc.
+  * Use auto-to-stick (https://www.fluentcpp.com/2018/09/28/auto-stick-changing-style/): `auto x = 17;` or `auto y = std::vector<size_t>{};`.
+  * Namespaces: Do not create nested namespaces, do not import namespaces.
+
+# Formatting and Naming
+* Much of this is enforced by clang-tidy. However, clang-tidy does not yet cover hpp files (see #1901). Also, while clang-tidy is a great help, do not rely on it.
+* Call ./scripts/format.sh before pushing.
```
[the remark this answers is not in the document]
Bensk1: Linting and test coverage is covered in the README

CONTRIBUTING.md
julianmenzler: I'm fine with both.

CONTRIBUTING.md
mrks: Added

> Consider structured bindings: `const auto& [iterator, added] = unordered_map.emplace(...);`

CONTRIBUTING.md
mrks: Is this something that can be boiled down into a proper guideline though? I don't think you can teach the concepts of ownership in this document. I'm happy to add

> Be mindful of ownership. Not everything needs to be a smart pointer. Consider passing around references to the object or references to a shared_ptr instead of copying the shared_ptr. Remember that this might not be safe when passing shared_ptrs into, i.e., JobTasks.

Anything beyond that is, in my opinion, the job of the seminar.

CONTRIBUTING.md
mrks: Not sure what your point is. We use "PR" in every second line here.

CONTRIBUTING.md
mrks: > If you have to use old-style loops, keep in mind that the loop condition is evaluated every time: Instead of `for (auto offset = size_t{0}; offset < something.size(); ++offset)`, the size should be retrieved just once. See also [here](http://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop)

CONTRIBUTING.md
mrks: I was just amused by talking about exceptions to the exception handling policy. I'm not attached to that joke - leave it or remove it?

CONTRIBUTING.md
mrks: Changed to PascalCase.

CONTRIBUTING.md
mrks: This is, at least, applied inconsistently. I don't really like the `/* */` style, but would be fine with using it. Before we make this a rule however, we should make sure that the codebase follows it to a large degree.

Honestly, I don't see a big advantage in prescribing something here. For me, code highlighting does the job.

Any other opinions on this?

CONTRIBUTING.md
```diff
@@ -1,73 +1,76 @@
...
-
-**Things to look for:**
-
-	- Guidelines (see above)
-	- Is the copy constructor deleted where it makes sense?
-	- Is the destructor virtual for base classes?
-	- Are unnecessary copies of heavy elements made? (prefer vector& over vector, but not int& over int)
-	- Did the author update documentation and dependencies (Wiki, README.md, DEPENDENCIES.md, Dockerfile, install_dependencies.sh, Jenkinsfile)
+# The Four Commandments
+1. Thy code (or the beginning of the header file) shalt be the primary method of documentation. Comments should be used to explain higher level concepts, but primarily, the code should be explaining itself. Part of this is to choose concise, descriptive names.
```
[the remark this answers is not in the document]
mrks: > Thy code shalt be the primary method of documentation. Part of this is to choose concise but descriptive names. Comments should be used to explain the concept and usage of classes (in the hpp file) and the structure of the algorithms (in the implementation).

Better?

CONTRIBUTING.md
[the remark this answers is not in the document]
mrks: In my opinion, the guidelines should document what we have done so far to keep the code base clean. If we want to introduce something new, it should first be realized in the code base and then added to the guidelines, not the other way round. Otherwise, we risk that people see the guidelines as wishful thinking.

CONTRIBUTING.md
Bouncner: That's fine with me. I link would probably help though (I don't like any of the links above tbh ... Herb has a post, but that one does not reason about when to use pointers at all ... https://herbsutter.com/2013/05/29/gotw-89-solution-smart-pointers/)

CONTRIBUTING.md
Bouncner: I don't like unnecessary abbr in titles. That's all. But that's probably just me. Resolved.

CONTRIBUTING.md
Bouncner: I smiled about the exceptions as well as the "thy", but at least for me it's often the case that I regret writing such things a year from now. So maybe keep the guidelines "boring"?

CONTRIBUTING.md
Bouncner: I don't care. If we continue to use the `*` in every line of the comment, I don't see a reason to use multi line comments. It saves one char and is still annoying to write. :)

I am fine with using `//` everywhere and would prefer it (a little bit). I would definitely like to see a strong recommendation in the guidelines for one of both.

mrks: > I would definitely like to see a strong recommendation in the guidelines for one of both.

> In my opinion, the guidelines should document what we have done so far to keep the code base clean. If we want to introduce something new, it should first be realized in the code base and then added to the guidelines, not the other way round. Otherwise, we risk that people see the guidelines as wishful thinking.

Discussion over here: https://github.com/hyrise/hyrise/pull/2341#discussion_r606756175

CONTRIBUTING.md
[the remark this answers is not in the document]
mweisgut: Agreed

CONTRIBUTING.md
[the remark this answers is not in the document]
mweisgut: yes

CONTRIBUTING.md
```diff
@@ -1,73 +1,78 @@
...
+    * When your implemented class does not need to be copied (e.g., a Table should never exist more than once), inherit from Noncopyable to avoid these potentially expensive copies.
+
+* Miscellaneous
+  * Prefer `if (object) {` over `if (object != nullptr) {` or `if (object.has_value()) {`.
+  * Don't write `this->` if you don't have to.
+  * Be explicit with types: Use [u]int(8|16|32|64)_t instead of `int, long, uint` etc.
+  * Use [auto-to-stick](https://www.fluentcpp.com/2018/09/28/auto-stick-changing-style/): `auto x = 17;` or `auto y = std::vector<size_t>{};`.
+  * Namespaces: Do not create nested namespaces, do not import namespaces.
+  * Consider structured bindings: `const auto& [iterator, added] = unordered_map.emplace(...);`
+
```
julianmenzler, line 36: Add some notes on templated methods? Maybe here or add another bullet point. 

These are the notes from the previous guidelines:

> If your templated methods/classes are used outside of your file, they have to be in the header. But if you only use them internally, you should place them in the cpp file.

> Try to keep the size of your templated methods as small as possible (Thin Template idiom). If only a small part of your method depends on the template parameter, consider moving the rest into a non-templated method. This reduces compile times.

mrks: Move Thin Template to Ressource Page

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
...
+4. Thou shalt properly test thy code. This includes unit *and* integration tests. Try to isolate parts that can be
+   independently tested.
 
-# Coding Style
-Avoid exception handling. Because Hyrise is not a product, we do not have to recover from errors. Instead, fail loud (i.e., terminate the program) so that developers immediately notice that something is wrong and can fix the problem at its root.
+# C++
+* Use automatic memory management (RAII, smart pointers). `new` and `malloc` are evil words.
+* Be mindful of ownership. Not everything needs to be a smart pointer. Consider passing around references to the object
+  or references to a shared_ptr instead of copying the shared_ptr. Remember that this might not be safe when passing
+  shared_ptrs into, i.e., JobTasks.
```
Bensk1, line 19: `i.e.` -> `e.g.`

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
...
-- Don't write `this->` if you don't have to
-- Use C++11 for loops when possible: `for (const auto& item : items) {...}`
-- When creating a vector where you know the size beforehand, use `reserve` to avoid unnecessary resizes and allocations
-- Don’t evaluate end() every time through a loop: http://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop
-- Use `if (object) {` over `if (object != nullptr) {` or `if (object.has_value()) {`
+* Data structures
+  * When creating a vector where you know the size beforehand, use `reserve` to avoid unnecessary resizes and
+    allocations.
+  * Hash-based data structures are usually faster than tree-based data structures. Unless you have a reason to use the
+    latter, prefer unordered_(map|set) over map and set.
```
Bensk1, line 41: `unordered_`...?

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
...
 
+* Copies
+    * C++ makes it too easy to inadvertently copy objects.
```
Bensk1, line 44: Even though this information is valuable, this is the only (or one of the few lines) that does not contain any kind of commandment. Maybe, we could change it to "Avoid unnecessary copies, C++ makes it..."

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
...
+    * For larger elements (e.g., vectors), pass a (const) reference instead.
```
Bensk1, line 45: `const`?

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
...
+    * When your implemented class does not need to be copied (e.g., a Table should never exist more than once), inherit
+      from Noncopyable to avoid these potentially expensive copies.
```
Bensk1, line 47: `Noncopyable`

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
...
+* Data structures
+  * When creating a vector where you know the size beforehand, use `reserve` to avoid unnecessary resizes and
+    allocations.
+  * Hash-based data structures are usually faster than tree-based data structures. Unless you have a reason to use the
+    latter, prefer unordered_(map|set) over map and set.
 
+* Copies
+    * C++ makes it too easy to inadvertently copy objects.
+    * For larger elements (e.g., vectors), pass a (const) reference instead.
+    * When your implemented class does not need to be copied (e.g., a Table should never exist more than once), inherit
```
Bensk1, line 46: `Table`

Bensk1, line 46: When -> If?

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
...
+  * Namespaces: Do not create nested namespaces, do not import namespaces.
+  * Consider structured bindings: `const auto& [iterator, added] = unordered_map.emplace(...);`
+  * Prefer braced control statements (`if (...) { x(); }` over `if (...) x();`). For multi-line statements, this is
+    mandatory.
 
-### Naming convention for gtest macros:
+# Formatting and Naming
+* Much of this is enforced by clang-tidy. However, clang-tidy does not yet cover hpp files (see #1901). Also, while
+  clang-tidy is a great help, do not rely on it.
+* Call ./scripts/format.sh before pushing.
```
Bensk1, line 63: "...before committing your code"?

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
...
+* Maintain correct orthography and grammar. Comments should start with a capital letter, sentences should be finished
+  with a full stop.
+  * Class names within comments are written in PascalCase - e.g., "As the input may be a ReferenceSegment, a valid RowID
+    may point to a row that is NULL."
 
-If you want to test a single module, class or test you have to execute the test binary and use the `gtest_filter` option:
+# Pull Requests
+## Opening PRs
+* When you submit a non-trivial PR, include the results of benchmark_all.sh.
+  * These help in understanding potential performance changes as well as document potential changes to the compilation
```
Bensk1, line 90: These results

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
...
+    costs.
+  * We do not do this automatically as the CI server is not sufficiently isolated and the performance results would
+    vary. Similarly, your personal laptop is likely to produce unreliable results.
+* If your PR is related to an existing issue, reference it in the PR's description (e.g., `fixes #123` or `refs #123`).
+* If you are not a member of the Hyrise organization, your PR will not be built by our CI server. Contact a maintainer
+  for this. They can add you to the organization or manually trigger builds from within Jenkins.
+* For your PR to be merged, it must pass the FullCI run. Set the FullCI tag in Github before committing to trigger the
```
Bensk1, line 97: **a** FullCI run?

Bensk1: Super minor: GitHub

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
...
-	- Is the destructor virtual for base classes?
-	- Are unnecessary copies of heavy elements made? (prefer vector& over vector, but not int& over int)
-	- Did the author update documentation and dependencies (Wiki, README.md, DEPENDENCIES.md, Dockerfile, install_dependencies.sh, Jenkinsfile)
+## Reviewing PRs
+* Keep the whole picture in mind. Often, it makes sense to make two passes: One for the code style and line-level
+  modifications; one for checking how it fits into the overall picture.
+* Check if the PR sufficiently adds tests both for happy and unhappy cases.
+* Do not shy away from requesting changes on surrounding code that was not modified in the PR. Remember that after a PR,
+  the code base should be better than before.
+* Verify the CI results, including the change in procentual coverage, and check if the compile time or query performance
```
Bensk1, line 108: procentual -> percentaged, percental

Bensk1 (changes requested):

CONTRIBUTING.md
```diff
@@ -1,73 +1,110 @@
-# Contribution Guidelines
-Do not commit/push directly to the master branch. Instead, create a feature branch/fork and file a merge request.
+# The Four Commandments
+1. Thy code shalt be the primary method of documentation. Part of this is to choose concise but descriptive names.
+   Comments should be used to explain the concept and usage of classes (in the hpp file) and the structure of the
+   algorithms (in the implementation).
+2. Thou shalt program defensively. Use Assert wherever it makes sense, use DebugAssert in performance-critical parts
```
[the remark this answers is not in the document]
mrks: Mmmh? You mean 

> Use Assert and DebugAssert wherever it makes sense, use DebugAssert in performance-critical parts

? I think the idea was to prioritize Assert.

CONTRIBUTING.md
mrks: ?

CONTRIBUTING.md
mrks: ?

CONTRIBUTING.md
mrks: Ah, are you talking about adding backticks? Don't care.

CONTRIBUTING.md
mrks: I'm not a fan of following companies' weird spelling preferences.

CONTRIBUTING.md
mrks: Maybe just "how the reported coverage changed"?

CONTRIBUTING.md
mrks: (Don't care enough to argue about it though.)

CONTRIBUTING.md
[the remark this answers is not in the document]
Bensk1: No, I just meant that we **could** use quotes to indicate code.

CONTRIBUTING.md
Bensk1: Same as above, indicating code.

CONTRIBUTING.md
Bensk1: Same as above, indicate code. I believe we used `const` over const in other places, too.

CONTRIBUTING.md
Bensk1: Yes.

CONTRIBUTING.md
Bensk1: Works for me.

CONTRIBUTING.md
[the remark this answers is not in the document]
mrks: Whatever you prefer.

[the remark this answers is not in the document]
Bensk1: I like the backtick version but it's not that important to me.

Bensk1 (approved):

PR #2343, opened 2021-03-22
Fix jemalloc includes for Plugin builds on Linux

Bouncner:
Currently, we cannot build Hyrise's example plugin on Linux. In general, we build jemalloc first and then link and include the generated files.

Now, if we have import hyrise from the path `hyrise`, we do no longer find the jemalloc includes under `third_party/jemalloc/include` but rather `hyrise/third_party/jemalloc/include`.

This PR solves this issue by using `CMAKE_CURRENT_BINARY_DIR` to obtain the path `cmake-build-debug/hyrise/third_party` from which we can simply go upwards. `CMAKE_BINARY_DIR` does not tell us that Hyrise lives within the `hyrise` folder.
I have not found a nicer way to obtain the information, that we need to add `hyrise` to the include path.

This error does not occur on MaOS as we do not use jemalloc on MacOS.

### DIFF
--- a/src/lib/CMakeLists.txt
+++ b/src/lib/CMakeLists.txt
@@ -719,7 +719,9 @@ if(NOT APPLE)
     target_link_libraries(hyrise_impl PUBLIC custom_jemalloc -rdynamic)
     target_compile_definitions(hyrise_impl PUBLIC HYRISE_WITH_JEMALLOC)
 
-    # Add the jemalloc headers to Hyrise so that we can use jemalloc-specific functions
+    # Add the jemalloc headers to Hyrise so that we can use jemalloc-specific functions.
+    # We need to use a path relative to CMAKE_CURRENT_BINARY_DIR instead of using CMAKE_BINARY_DIR because Hyrise might
+    # not be the top-level CMake project (e.g., if it is used as a submodule in a Hyrise plugin repository).
     target_include_directories(hyrise_impl PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/../../third_party/jemalloc/include)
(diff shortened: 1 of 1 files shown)


mrks:
I don't see anything wrong with it. If you add a comment explaining that `CMAKE_BINARY_DIR` could be different from `hyrise/` if Hyrise is a submodule, that should be sufficient.

I'm not an expert for CMake prettiness though. Usually I can get it to do what I want it to do, but this might not be the perfect solution.

Bouncner:
```
# Add the jemalloc headers to Hyrise so that we can use jemalloc-specific functions

>>

# Add the jemalloc headers to Hyrise so that we can use jemalloc-specific functions
# Using CMAKE_CURRENT_BINARY_DIR and going two levels over CMAKE_CURRENT_BINARY_DIR directly. The reason is that we can obtain the folder in which Hyrise is built. In case a plugin loads Hyrise as a submodule, Hyrise is not necessarily built in the root directory.
```
?

mrks:
> Using CMAKE_CURRENT_BINARY_DIR and going two levels over CMAKE_CURRENT_BINARY_DIR directly. The reason is that we can obtain the folder in which Hyrise is built. In case a plugin loads Hyrise as a submodule, Hyrise is not necessarily built in the root directory.

Suggestion:

> We need to use a path relative to CMAKE_CURRENT_BINARY_DIR instead of using CMAKE_BINARY_DIR because Hyrise might not be the top-level CMake project (e.g., if it is used as a submodule in a Hyrise plugin repository)

mrks (approved):

PR #2344, opened 2021-03-23
Enable table data loading from directory in server.cpp

Bouncner:
This PR enables loading data from a directory. Initially, my goal was to enable creating the data for JOB data, but JOB only uses the `FileBasedTableGenerator` anyways.

This way one can easily import data from "any" directory *and* gains the advantage of automatically loading binary data when possible.

### DIFF
--- a/src/bin/server.cpp
+++ b/src/bin/server.cpp
@@ -1,10 +1,13 @@
 #include "cxxopts.hpp"
 
+#include <filesystem>
+
 #include <boost/algorithm/string.hpp>
 #include <boost/lexical_cast.hpp>
 
 #include "benchmark_config.hpp"
 #include "cli_config_parser.hpp"
+#include "file_based_table_generator.hpp"
 #include "server/server.hpp"
 #include "tpcc/tpcc_table_generator.hpp"
 #include "tpcds/tpcds_table_generator.hpp"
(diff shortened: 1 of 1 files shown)


mrks:
Any reasons you don't use the import commands?

Bouncner:
Mainly convenience for working on the JOB (not having twenty queries to load and also not checking if a binary is loadable or not).

If that convenience is considered "not worth it", I don't mind to close this PR.

mrks:
One could easily put this into a script rather than the server binary. Personally, I feel like this isn't the job of the server and the TPC-H convenience commands should not be taken as precedent.

mrks:
We could put a `load_join_order_benchmark.sql` into the resources folder for convenience though.

Bouncner:
Forgot to mention another factor that is pretty significant: load time. When using the `FileBasedTableGenerator`, we get the parallelization of encoding and histogram generation of the `AbstractTableGenerator`. That's a huge difference (mostly for TPC-H).
I haven't tested what happens if I add tables in parallel though, but I remember that it was likely to crash when tables are added in parallel to the storage manager.

mrks:
Sounds like something we should fix in general though, not just by adding convenience methods to the server though. :/

Bouncner:
Yeah, absolutely. This was just a note to have this issue written down. Not meant to reopen this PR.

PR #2345, opened 2021-03-30
Hash join: keep track of skipped values from Bloom filtering

Bouncner:
This PR adds three integers to the hash join's performance operator struct. They keep track of the "success" of the bloom filtering:
 * size of materialized probe side (might or might not be filtered by a bloom filter, see added test)
 * size of hash table
 * size of positions stored for hash table (can be empty for certain joins)

The reason to add these counters is to enable better join runtime predictions (might be of interest for @dey4ss, @aloeser, and @Bensk1; even though it does not interact with clustering decisions). For certain joins, the bloom filters remove 90% of the data, almost removing the runtime of the probe phase from join.

I thought about adding these values to the visualization as well. But for now I have not, because:
- a sole number is cumbersome, to calculate percentage (input table vs. materialized) we would need the input size, which we don't have
- this number works for the probe side and a "normal" join, where we can count the positions stored for the hash table
- BUT I have no idea what to print for semi/anti joins; ideally we could compare the number of distinct values of the input table and the size of the hash table after the build phase, but we don't have the distinct count at hand

### DIFF
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -366,8 +366,8 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
       _performance_data.set_step_runtime(OperatorSteps::ProbeSideMaterializing, timer_materialization.lap());
     } else {
       // Here, we first materialize the probe side and use the resulting Bloom filter in the materialization of the
-      // build side. Consequently, the passed Bloom filter in build() will have no effect as it has been already used
-      // during the materialization to filter non-matching values.
+      // build side. Consequently, the Bloom filter later passed into build() will have no effect as it has already
+      // been used here to filter non-matching values.
       materialize_probe_side(ALL_TRUE_BLOOM_FILTER);
       _performance_data.set_step_runtime(OperatorSteps::ProbeSideMaterializing, timer_materialization.lap());
       materialize_build_side(probe_side_bloom_filter);
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -371,6 +371,12 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
       _performance.set_step_runtime(OperatorSteps::BuildSideMaterializing, timer_materialization.lap());
     }
 
+    // Store the number of probe values after the probe side has been materialized. Depending on the input sizes, the
+    // probe side has potentially been filtered using the Bloom filter.
+    for (const auto& partition : materialized_probe_column) {
+      _performance.probe_side_materialized_values += partition.elements.size();
+    }
+
     /**
      * 2. Perform radix partitioning for build and probe sides. The bloom filters are not used in this step. Future work
      *    could use them on the build side to exclude them for values that are not seen on the probe side. That would
@@ -439,6 +445,16 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
                                                        probe_side_bloom_filter);
     }
     _performance.set_step_runtime(OperatorSteps::Building, timer_hash_map_building.lap());
+
+    // Store the element counts of the built hash tables. Depending on the bloom filter, we might have significantly
+    // less values stored than in the initial input table.
+    for (const auto& hash_table : hash_tables) {
+      if (hash_table) {
+        const auto [hash_table_size, position_count] = hash_table->get_hash_table_size_and_position_count();
+        _performance.build_side_hash_table_size += hash_table_size;
+        _performance.build_side_position_count += position_count;
+      }
+    }
 
     /**
      * Short cut for AntiNullAsTrue:
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -458,7 +458,14 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
       if (!hash_table) continue;
 
       _performance_data.hash_tables_distinct_value_count += hash_table->distinct_value_count();
-      _performance_data.hash_tables_position_count += hash_table->position_count() ? *hash_table->position_count() : 0;
+      const auto position_count = hash_table->position_count();
+      if (position_count) {
+        // Update or set hash_tables_position_count when hash table stores positions.
+        _performance_data.hash_tables_position_count =
+            _performance_data.hash_tables_position_count
+                ? *_performance_data.hash_tables_position_count + *position_count
+                : *position_count;
+      }
     }
 
     /**
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -237,7 +237,7 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
         _mode(mode),
         _column_ids(column_ids),
         _predicate_condition(predicate_condition),
-        _performance(performance_data),
+        _performance_data(performance_data),
         _output_column_order(output_column_order),
         _secondary_predicates(std::move(secondary_predicates)),
         _radix_bits(radix_bits) {}
@@ -248,7 +248,7 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
   const JoinMode _mode;
   const ColumnIDPair _column_ids;
   const PredicateCondition _predicate_condition;
-  JoinHash::PerformanceData& _performance;
+  JoinHash::PerformanceData& _performance_data;
 
   OutputColumnOrder _output_column_order;
 
@@ -361,27 +361,26 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
       // of code paths low, materialize_*_side always expects a Bloom filter. For the first step, we thus pass in a
       // Bloom filter that returns true for every probe.
       materialize_build_side(ALL_TRUE_BLOOM_FILTER);
-      _performance.set_step_runtime(OperatorSteps::BuildSideMaterializing, timer_materialization.lap());
+      _performance_data.set_step_runtime(OperatorSteps::BuildSideMaterializing, timer_materialization.lap());
       materialize_probe_side(build_side_bloom_filter);
-      _performance.set_step_runtime(OperatorSteps::ProbeSideMaterializing, timer_materialization.lap());
+      _performance_data.set_step_runtime(OperatorSteps::ProbeSideMaterializing, timer_materialization.lap());
     } else {
-      std::cout << "probe larger build" << std::endl;
       // Here, we first materialize the probe side and use the resulting Bloom filter in the materialization of the
       // build side. Consequently, the passed Bloom filter in build() will have no effect as it has been already used
       // during the materialization to filter non-matching values.
       materialize_probe_side(ALL_TRUE_BLOOM_FILTER);
-      _performance.set_step_runtime(OperatorSteps::ProbeSideMaterializing, timer_materialization.lap());
+      _performance_data.set_step_runtime(OperatorSteps::ProbeSideMaterializing, timer_materialization.lap());
       materialize_build_side(probe_side_bloom_filter);
-      _performance.set_step_runtime(OperatorSteps::BuildSideMaterializing, timer_materialization.lap());
+      _performance_data.set_step_runtime(OperatorSteps::BuildSideMaterializing, timer_materialization.lap());
     }
 
     // Store the number of materialized values. Depending on the order of materialization (which depends on the input
     // sizes), each side might or might not be filtered by the Bloom filter.
     for (const auto& partition : materialized_build_column) {
-      _performance.build_side_materialized_value_count += partition.elements.size();
+      _performance_data.build_side_materialized_value_count += partition.elements.size();
     }
     for (const auto& partition : materialized_probe_column) {
-      _performance.probe_side_materialized_value_count += partition.elements.size();
+      _performance_data.probe_side_materialized_value_count += partition.elements.size();
     }
 
     /**
@@ -427,7 +426,7 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
       histograms_build_column.clear();
       histograms_probe_column.clear();
 
-      _performance.set_step_runtime(OperatorSteps::Clustering, timer_clustering.lap());
+      _performance_data.set_step_runtime(OperatorSteps::Clustering, timer_clustering.lap());
     } else {
       // short cut: skip radix partitioning and use materialized data directly
       radix_build_column = std::move(materialized_build_column);
@@ -451,15 +450,15 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
       hash_tables = build<BuildColumnType, HashedType>(radix_build_column, JoinHashBuildMode::AllPositions, _radix_bits,
                                                        probe_side_bloom_filter);
     }
-    _performance.set_step_runtime(OperatorSteps::Building, timer_hash_map_building.lap());
+    _performance_data.set_step_runtime(OperatorSteps::Building, timer_hash_map_building.lap());
 
     // Store the element counts of the built hash tables. Depending on the Bloom filter, we might have significantly
     // less values stored than in the initial input table.
     for (const auto& hash_table : hash_tables) {
       if (!hash_table) continue;
 
-      _performance.hash_table_distinct_value_count += hash_table->get_distinct_value_count();
-      _performance.hash_table_position_count += hash_table->get_position_count();
+      _performance_data.hash_tables_distinct_value_count += hash_table->distinct_value_count();
+      _performance_data.hash_tables_position_count += hash_table->position_count() ? *hash_table->position_count() : 0;
     }
 
     /**
@@ -475,7 +474,7 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
           if (null_value) {
             Timer timer_output_writing;
             const auto result = _join_hash._build_output_table({});
-            _performance.set_step_runtime(OperatorSteps::OutputWriting, timer_output_writing.lap());
+            _performance_data.set_step_runtime(OperatorSteps::OutputWriting, timer_output_writing.lap());
             return result;
           }
         }
@@ -537,7 +536,7 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
       default:
         Fail("JoinMode not supported by JoinHash");
     }
-    _performance.set_step_runtime(OperatorSteps::Probing, timer_probing.lap());
+    _performance_data.set_step_runtime(OperatorSteps::Probing, timer_probing.lap());
 
     radix_probe_column.clear();
     hash_tables.clear();
@@ -566,7 +565,7 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
                             create_left_side_pos_lists_by_segment, create_right_side_pos_lists_by_segment,
                             _output_column_order, ALLOW_PARTITION_MERGE);
 
-    _performance.set_step_runtime(OperatorSteps::OutputWriting, timer_output_writing.lap());
+    _performance_data.set_step_runtime(OperatorSteps::OutputWriting, timer_output_writing.lap());
 
     return _join_hash._build_output_table(std::move(output_chunks));
   }
--- a/src/lib/operators/join_hash.cpp
+++ b/src/lib/operators/join_hash.cpp
@@ -365,6 +365,7 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
       materialize_probe_side(build_side_bloom_filter);
       _performance.set_step_runtime(OperatorSteps::ProbeSideMaterializing, timer_materialization.lap());
     } else {
+      std::cout << "probe larger build" << std::endl;
       // Here, we first materialize the probe side and use the resulting Bloom filter in the materialization of the
       // build side. Consequently, the passed Bloom filter in build() will have no effect as it has been already used
       // during the materialization to filter non-matching values.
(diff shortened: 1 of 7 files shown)


src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -191,6 +191,16 @@ class PosHashTable {
     return _offset_hash_table.find(casted_value) != _offset_hash_table.end();
   }
 
+  // Return the size of the hash table and the number of positions stored (in case the mode is AllPositions). We use
+  // the unified position list that is created in finalize(). For semi/anti* joins the number of positions is 0.
```
mrks, line 195: if you use the unified positions list, maybe assert that it was already created?

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -191,6 +191,16 @@ class PosHashTable {
...
+  std::pair<size_t, size_t> get_hash_table_size_and_position_count() const {
```
mrks, line 196: To me, it doesn't look like we are saving anything by writing an `a_and_b` method. Any reason not to just use two methods?

src/lib/operators/join_hash.cpp
```diff
@@ -371,6 +371,12 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
       _performance.set_step_runtime(OperatorSteps::BuildSideMaterializing, timer_materialization.lap());
     }
 
+    // Store the number of probe values after the probe side has been materialized. Depending on the input sizes, the
+    // probe side has potentially been filtered using the Bloom filter.
```
mrks, line 375: bloom vs Bloom (I'd use the former)

src/lib/operators/join_hash.cpp
```diff
@@ -440,6 +446,16 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
     }
     _performance.set_step_runtime(OperatorSteps::Building, timer_hash_map_building.lap());
 
+    // Store the element counts of the built hash tables. Depending on the bloom filter, we might have significantly
+    // less values stored than in the initial input table.
+    for (const auto& hash_table : hash_tables) {
+      if (hash_table) {
```
mrks, line 452: Guard clause?

src/lib/operators/join_hash.hpp
```diff
@@ -52,6 +52,13 @@ class JoinHash : public AbstractJoinOperator {
     size_t radix_bits{0};
     // Initially, the left input is the build side and the right side is the probe side.
     bool left_input_is_build_side{true};
+
+    // Due to the used Bloom filters, the number of actually joined tuples can significantly differ from the sizes of
+    // the input tables. For the build side, the bloom filter reduces the number of values in the hash table (i.e., the
+    // size of the hash table) and potentially the number of rows (in case of non-semi/anti* joins).
+    size_t probe_side_materialized_values{0};
+    size_t build_side_hash_table_size{0};
+    size_t build_side_position_count{0};
```
mrks, line 61: What is the semantic difference between "values", "size", and "count"?

mrks, line 61: Are you making the assumption that the build side is always materialized first and the probe side is the first that is filtered by a bloom filter? If yes, this does not seem to match the JoinHash implementation:

https://github.com/hyrise/hyrise/blob/adaa9d5305eb952d4125cd66b20f5e58ef62a9eb/src/lib/operators/join_hash.cpp#L359-L372

src/test/lib/operators/operator_performance_data_test.cpp
```diff
@@ -202,6 +202,48 @@ TEST_F(OperatorPerformanceDataTest, JoinHashStepRuntimes) {
...
+  // partner. Thus, we only check for the existence of the minimum number of values and positions.
+  const auto inner_join = std::make_shared<JoinHash>(
+      table_wrapper_a, table_wrapper_b, JoinMode::Inner,
+      OperatorJoinPredicate{ColumnIDPair(ColumnID{0}, ColumnID{0}), PredicateCondition::Equals});
+  inner_join->execute();
+
+  const auto& inner_perf = dynamic_cast<JoinHash::PerformanceData&>(*inner_join->performance_data);
+  EXPECT_GE(inner_perf.probe_side_materialized_values, 3ul);  // values 2,6,2
+  EXPECT_GE(inner_perf.build_side_hash_table_size, 2ul);      // values 2,6
+  EXPECT_GE(inner_perf.build_side_position_count, 3ul);       // positions 1,2,3
```
mrks, line 228: Wouldn't this test pass because of `_GE` if we either did no bloom filtering or the bloom filtering was not reported?

src/lib/operators/join_hash.hpp
```diff
@@ -52,6 +52,13 @@ class JoinHash : public AbstractJoinOperator {
     size_t radix_bits{0};
     // Initially, the left input is the build side and the right side is the probe side.
     bool left_input_is_build_side{true};
+
```
mrks, line 55: >I thought about adding these values to the visualization as well. But for now I have not, because:
> - a sole number is cumbersome, to calculate percentage (input table vs. materialized) we would need the input size, which we don't have

Why not?

> - this number works for the probe side and a "normal" join, where we can count the positions stored for the hash table
> - BUT I have no idea what to print for semi/anti joins; ideally we could compare the number of distinct values of the input table and the size of the hash table after the build phase, but we don't have the distinct count at hand

Have you considered simply tracking the number of values that didn't make it through the bloom filter? I'd hope for the cost of that to be minimal if we only synchronize at the end of the task.

src/lib/operators/join_hash.hpp
mrks: I just realized that the remainder of the hash join code largely works under the assumption that the if statement from line 359 is true.

Unless I am mistaken, if we materialize the probe side first, it does not make any sense to use a bloom filter when building the hash table. On the other side, we *could* use a bloom filter in the probe methods if we have not yet filtered the probe side when materializing it. This seems to catch some additional cases, but at least for TPC-H, I did not see a significant gain.

Bouncner: Good point. Maybe `probe_side_materialized_value_count`, `build_side_hash_table_item_count`, and `build_side_position_count`?

src/lib/operators/join_hash.cpp
Bouncner: Since its a name it should actually be capitalized. I'll change it to be "Bloom" everywhere, objection?

src/lib/operators/join_hash.hpp
Bouncner: > Are you making the assumption that the build side is always materialized first and the probe side is the first that is filtered by a bloom filter?

Nope. That's also what's being tested (next comment).

Concerning your other comment: I thought about it. But since we invest a lot of work for L2 local hash tables, I thought (might be totally wrong) that the saved hash lookup does not make up the costs of the Bloom lookup. To be honest, I don't have the time to look into it. I could add a comment about it though.

src/test/lib/operators/operator_performance_data_test.cpp
Bouncner: Yes to "no bloom filtering". But I don't get your second point.

src/lib/operators/join_hash.hpp
Bouncner: > Why not?

If I am not mistaken, the operator performance does not know it's parent when being printed. And we don't store the input tuple count. So I (i) have not found  a nice way to do it and (ii) don't think storing the input counts would be worth it.

> Have you considered simply tracking the number of values that didn't make it through the bloom filter? I'd hope for the cost of that to be minimal if we only synchronize at the end of the task.

Do you mean tracking it in the actual loops?

src/test/lib/operators/operator_performance_data_test.cpp
mrks: Pretty much the same thing. If, instead of reporting the "bloom-filtered" values, we just reported the input values, this test would also pass.

src/lib/operators/join_hash.hpp
mrks: Let's discuss this once we figured out what to do about the next review comment.

mrks: Ah, are you talking about this?

> As we first materialize the smaller side (i.e., the probe side in this case) and create the initial bloom filter with that, there will be no reduction due to bloom filtering on the probe side.

But the build side would be bloom-filtered, right? Why don't we have a `build_side_materialized_value_count`?

mrks: > Concerning your other comment: I thought about it. But since we invest a lot of work for L2 local hash tables, I thought (might be totally wrong) that the saved hash lookup does not make up the costs of the Bloom lookup. To be honest, I don't have the time to look into it. I could add a comment about it though.

This was my impression, too. In some cases, it seemed to hurt the performance. A comment would certainly help.

Similarly, I wonder if we go down the `else` part of the quoted statement (i.e., materialize the probe side first and bloom-filter the build side), it makes any sense to pass in the bloom filter here:

https://github.com/hyrise/hyrise/blob/adaa9d5305eb952d4125cd66b20f5e58ef62a9eb/src/lib/operators/join_hash.cpp#L435-L436

Again, no need to change anything, maybe just add a comment if you agree.

src/lib/operators/join_hash.hpp
mrks: > If I am not mistaken, the operator performance does not know it's parent when being printed. And we don't store the input tuple count. So I (i) have not found a nice way to do it and (ii) don't think storing the input counts would be worth it.

If, instead of storing absolute numbers, we calculate the bloom filter efficiency while writing the OperatorPerformanceData, we would still have the input sizes at hand.

> Do you mean tracking it in the actual loops?

Yes. Maybe that's worth testing.

src/lib/operators/join_hash.hpp
Bouncner: > Why don't we have a build_side_materialized_value_count?

The current "naming scheme" is meant to clearly differentiate between the two kind of "values" we are interested in:
  1. Distinct values in the hash map (i.e., size), when no positions are stored (semi/anti case)
  1. Distinct positions in the values of the hash table's entries, that's basically the number of remaining table tuples after bloom-filtering (inner case).

`build_side_materialized_value_count` does not clearly state what is means, I think.

Will add comment about the two issues mentioned.

src/test/lib/operators/operator_performance_data_test.cpp
Bouncner: That's right. Would you prefer to simply check for values being > 0 (ensure they have been set **somehow**)?

Since we have no handle on the Bloom filtering as long as we use `std::` stuff everywhere, an exact check shouldn't be possible (at least if we assume there is at least one value being filtered out).

src/lib/operators/join_hash.hpp
Bouncner: > If, instead of storing absolute numbers, we calculate the bloom filter efficiency while writing the OperatorPerformanceData, we would still have the input sizes at hand.

That's actually how I started. At had two floats like `build_side_bloom_reduction_ratio` which was in [0,1].
But I there is no way of having such a value for the semi case (we would need the distinct counts of the input table) and thus I found the current way is better (even though we might not be able to put it into the visualization).

> Yes. Maybe that's worth testing.

Really hesitant to introduce such a tracking in the most important operator. Might turn out to be fine on nemea, but we don't know about what is costs on AMD, ARM; we might only test the current page layout; and so on.
That's always true obviously, but for adding something in a hot loop, there needs to be a very good reason IMO. And if these counters are currently used by me three Hyrise users with very specific problems, I tend to not doing it.

mrks: Fair.

src/test/lib/operators/operator_performance_data_test.cpp
mrks: It's really safe to assume that std::hash of an integer is idempotent. It doesn't say so anywhere in the standard, but afaik that's what every implementation does. I wouldn't let that keep us from using an `EQ` test. Add a comment and we're good.

src/lib/operators/join_hash.hpp
Bouncner: Added
```
// Here, we first materialize the probe side and use the resulting Bloom filter in the build side
// materialization. Consequently, the passed Bloom filter in build() will have no effect.
```

src/test/lib/operators/operator_performance_data_test.cpp
Bouncner: Done.

src/lib/operators/join_hash.cpp
```diff
@@ -357,22 +357,35 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
...
+      // of code paths low, materialize_*_side always expects a Bloom filter. For the first step, we thus pass in a
+      // Bloom filter that returns true for every probe.
       materialize_build_side(ALL_TRUE_BLOOM_FILTER);
       _performance.set_step_runtime(OperatorSteps::BuildSideMaterializing, timer_materialization.lap());
       materialize_probe_side(build_side_bloom_filter);
       _performance.set_step_runtime(OperatorSteps::ProbeSideMaterializing, timer_materialization.lap());
     } else {
+      std::cout << "probe larger build" << std::endl;
+      // Here, we first materialize the probe side and use the resulting Bloom filter in the materialization of the
+      // build side. Consequently, the passed Bloom filter in build() will have no effect as it has been already used
```
mrks, line 370: Could this be clearer?

> the Bloom filter later passed into ... will have no effect as it has already been used here to filter non-matching values

src/lib/operators/join_hash.cpp
```diff
@@ -357,22 +357,35 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
...
       materialize_probe_side(ALL_TRUE_BLOOM_FILTER);
       _performance.set_step_runtime(OperatorSteps::ProbeSideMaterializing, timer_materialization.lap());
       materialize_build_side(probe_side_bloom_filter);
       _performance.set_step_runtime(OperatorSteps::BuildSideMaterializing, timer_materialization.lap());
     }
 
+    // Store the number of materialized values. Depending on the order of materialization (which depends on the input
+    // sizes), each side might or might not be filtered by the Bloom filter.
+    for (const auto& partition : materialized_build_column) {
+      _performance.build_side_materialized_value_count += partition.elements.size();
```
mrks, line 381: Slightly surprised about `_performance`. Didn't seem to have bugged me before. I wonder if `_performance_counters` might be easier to recognize. No strong opinion about it though.

src/lib/operators/join_hash.cpp
```diff
@@ -357,22 +357,35 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
...
-      // bloom filter that returns true for every probe.
+      // When materializing the first side (here: the build side), we do not yet have a Bloom filter. To keep the number
+      // of code paths low, materialize_*_side always expects a Bloom filter. For the first step, we thus pass in a
+      // Bloom filter that returns true for every probe.
       materialize_build_side(ALL_TRUE_BLOOM_FILTER);
       _performance.set_step_runtime(OperatorSteps::BuildSideMaterializing, timer_materialization.lap());
       materialize_probe_side(build_side_bloom_filter);
       _performance.set_step_runtime(OperatorSteps::ProbeSideMaterializing, timer_materialization.lap());
     } else {
+      std::cout << "probe larger build" << std::endl;
```
mrks, line 368: .

src/lib/operators/join_hash.cpp
```diff
@@ -440,6 +453,15 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
     }
     _performance.set_step_runtime(OperatorSteps::Building, timer_hash_map_building.lap());
 
+    // Store the element counts of the built hash tables. Depending on the Bloom filter, we might have significantly
+    // less values stored than in the initial input table.
+    for (const auto& hash_table : hash_tables) {
+      if (!hash_table) continue;
+
+      _performance.hash_table_distinct_value_count += hash_table->get_distinct_value_count();
```
mrks, line 461: `hash_tableS_...`?

src/lib/operators/join_hash.cpp
```diff
@@ -440,6 +453,15 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
...
+      _performance.hash_table_position_count += hash_table->get_position_count();
```
mrks, line 462: I don't entirely recall the `get_...` naming discussion. I believe we decided against `get_` unless it sounds weird (e.g., `chunk` instead of `get_chunk`).

src/lib/operators/join_hash.hpp
```diff
@@ -52,6 +52,19 @@ class JoinHash : public AbstractJoinOperator {
...
+    // Due to the used Bloom filters, the number of actually joined tuples can significantly differ from the sizes of
+    // the input tables. To enable analyses of the Bloom filter efficiency, we stored number of values that were
```
mrks, line 57: we store?

src/lib/operators/join_hash.hpp
```diff
@@ -52,6 +52,19 @@ class JoinHash : public AbstractJoinOperator {
...
+    // eventually materialized; i.e., "input_row_count - filtered_values_by_Bloom_filter".
+    size_t build_side_materialized_value_count{0};
+    size_t probe_side_materialized_value_count{0};
+
+    // For the build side, the Bloom filter reduces the distinct values in the hash table (i.e., the size of the hash
```
mrks, line 62: s/For the build side/In build()/?

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -191,6 +191,21 @@ class PosHashTable {
...
+  // Return the number of distinct values (i.e., the size of the hash table).
+  size_t get_distinct_value_count() const {
+    return _offset_hash_table.size();
+  }
+
+  // Return the number of positions stored. For semi/anti joins, no positions are stored and thus 0 is returned.
+  // Otherwise, we use the unified position list that is created in finalize().
+  size_t get_position_count() const {
+    if (_mode == JoinHashBuildMode::AllPositions) {
+      Assert(_unified_pos_list, "");
```
mrks, line 203: Text?

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -191,6 +191,21 @@ class PosHashTable {
...
+      return _unified_pos_list->pos_list.size();
+    }
+    return 0ul;
```
mrks, line 206: Could it be less ambiguous to return `std::optional<size_t>`?

mrks (approved):
Just nitpicky stuff

src/lib/operators/join_hash/join_hash_steps.hpp
Bouncner: `Assert(_unified_pos_list, "Expected unified position list to be set (i.e., hash table is finalized).");`

mrks: `(i.e., finalize() has been called)`? Just to remove one level of textual indirection.

src/lib/operators/join_hash.cpp
Bouncner: Changed to `_performance_data`, since it also stored bools etc.

src/lib/operators/join_hash/join_hash_steps.hpp
Bouncner: Changed.

Bouncner:
Zündable (saw your approve but there might be issues remaining in the unresolved comments?!?)?

src/lib/operators/join_hash.hpp
```diff
@@ -52,6 +52,19 @@ class JoinHash : public AbstractJoinOperator {
...
+    // eventually materialized; i.e., "input_row_count - filtered_values_by_Bloom_filter".
+    size_t build_side_materialized_value_count{0};
+    size_t probe_side_materialized_value_count{0};
+
+    // In build(), the Bloom filter potentially reduces the distinct values in the hash table (i.e., the size of the
+    // hash table) and the number of rows (in case of non-semi/anti* joins).
+    // Note, depending on the order of materialization, build_side_materialized_value_count is not necessarily equal to
+    // build_side_position_count (see order of materialization in hash_join.cpp).
+    size_t hash_tables_distinct_value_count{0};
+    size_t hash_tables_position_count{0};
```
mrks, line 67: Why don't you pass the optional all the way to the consumer? The whole idea was that the consumer can identify the difference between 0 and N/A.

src/lib/operators/join_hash.hpp
```diff
@@ -52,6 +52,19 @@ class JoinHash : public AbstractJoinOperator {
     size_t radix_bits{0};
     // Initially, the left input is the build side and the right side is the probe side.
     bool left_input_is_build_side{true};
+
+    // Due to the used Bloom filters, the number of actually joined tuples can significantly differ from the sizes of
+    // the input tables. To enable analyses of the Bloom filter efficiency, we store the number of values that were
+    // eventually materialized; i.e., "input_row_count - filtered_values_by_Bloom_filter".
+    size_t build_side_materialized_value_count{0};
```
mrks, line 59: You said

> build_side_materialized_value_count does not clearly state what is means, I think.

Are you fine with it after all? If you are happy, I am happy. ;)

mrks (approved):

src/lib/operators/join_hash.hpp
mrks: (No need to re-review once this is addressed)

src/lib/operators/join_hash.hpp
Bouncner: Happy with it. I think the problem was that I misunderstood you and though you wanted to use the name for the size of the hash table.

src/lib/operators/join_hash.hpp
Bouncner: Hmm, any ideas how to do that cleaner than this?
```
auto hash_tables_position_count = size{0};
    for (const auto& hash_table : hash_tables) {
      if (!hash_table) continue;

      _performance_data.hash_tables_distinct_value_count += hash_table->distinct_value_count();
      _performance_data.hash_tables_position_count += hash_table->position_count() ? *hash_table->position_count() : 0;
    }
    if (hash_tables_position_count > 0) {
      _performance_data.hash_tables_position_count = hash_tables_position_count;
    }
```

mrks:
````_performance_data.hash_tables_position_count =
  _performance_data.hash_tables_position_count.value_or(0) + *position_count;
```
?

Also, just noticed that it should say "IF hash table stores", not "when".

PR #2353, opened 2021-04-13
BenchmarkRunner: Write report idempotently

mrks:
I am currently running multi-hour benchmarks. Waiting until the end to realize that something is wrong sucks. With this PR, `BenchmarkRunner::write_report_to_file` can be
* called more than once, with the resulting file simply being overwritten, and
* called from the outside so that my plugin can generate a report when something has been done.

This PR does not change the default behavior of the BenchmarkRunner. I decided against generating a report after each executed item in the `Ordered` mode as this may incur significant costs with `--metrics`. If you want to profit from this change, you thus need to call `Hyrise::get().benchmark_runner.lock()->write_report_to_file()` yourself.

### DIFF
--- a/src/benchmarklib/benchmark_runner.cpp
+++ b/src/benchmarklib/benchmark_runner.cpp
@@ -453,7 +453,7 @@ void BenchmarkRunner::write_report_to_file() const {
                         {"summary", std::move(summary)},
                         {"table_generation", _table_generator->metrics}};
 
-  // Add information that was gathered asynchronously to the output data
+  // Add information that was temporarily stored in the `benchmark_...` tables during the benchmark execution
   if (Hyrise::get().storage_manager.has_table("benchmark_system_utilization_log")) {
     report["system_utilization"] = _sql_to_json("SELECT * FROM benchmark_system_utilization_log");
   }
--- a/src/benchmarklib/benchmark_runner.cpp
+++ b/src/benchmarklib/benchmark_runner.cpp
@@ -295,8 +295,6 @@ void BenchmarkRunner::_benchmark_ordered() {
     // into account, too. In light of the significant amount of data added by the snapshots to the JSON file and the
     // unclear advantage of excluding those runs, we only take one snapshot here.
     _snapshot_segment_access_counters(name);
-
-    write_report_to_file();
   }
 }
 
--- a/src/benchmarklib/benchmark_runner.cpp
+++ b/src/benchmarklib/benchmark_runner.cpp
@@ -133,14 +133,10 @@ void BenchmarkRunner::run() {
     }
   }
 
-  auto benchmark_end = std::chrono::system_clock::now();
-  _total_run_duration = benchmark_end - _benchmark_start;
-
   // Create report
   if (_config.output_file_path) {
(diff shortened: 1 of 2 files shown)


mrks:
Just to clarify: This isn't something that I *need*. Since this isn't adding any code, I figured it might not hurt to merge this into the master and gain a little bit of flexibility.

Bouncner:
Does the ability to write intermittent results make it idempotent?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -454,6 +453,7 @@ void BenchmarkRunner::_create_report(std::ostream& stream) const {
                         {"summary", std::move(summary)},
                         {"table_generation", _table_generator->metrics}};
 
+  // Add information that was gathered asynchronously to the output data
```
Bouncner, line 456: How and where was that gathered asynchronously?

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -462,7 +462,8 @@ void BenchmarkRunner::_create_report(std::ostream& stream) const {
     report["segments"] = _sql_to_json("SELECT * FROM benchmark_segments_log");
   }
 
-  stream << std::setw(2) << report << std::endl;
+  // Write the output file
+  std::ofstream{_config.output_file_path.value()} << std::setw(2) << report << std::endl;
```
Bouncner, line 466: You're writing the whole JSON report every time, right? Just to be sure I get this right.

src/benchmarklib/benchmark_runner.hpp
```diff
@@ -51,6 +51,11 @@ class BenchmarkRunner : public Noncopyable {
   // If the query execution should be validated, this stores a pointer to the used SQLite instance
   std::shared_ptr<SQLiteWrapper> sqlite_wrapper;
 
+  // Create a report in roughly the same format as google benchmarks do when run with --benchmark_format=json.
+  // This is idempotent, i.e., you can call it multiple times and the result will be updated. Be aware that this may
```
Bouncner, line 55: That "be updated" confused me a bit, since you "just overwrite" the whole file, don't you?

src/benchmarklib/benchmark_runner.cpp
mrks: Grep for `INSERT INTO`. Just realized that only the system utilization is gathered asynchronously.

> Add information that was temporarily stored during the benchmark execution?

src/benchmarklib/benchmark_runner.hpp
mrks: `and the result file will be overwritten`?

Bouncner: 👍

src/benchmarklib/benchmark_runner.cpp
Bouncner: Looking at the `INSERT INTO` part makes it clear. Thanks. In this case, the current comment is fine.

Bouncner (approved):

PR #2354, opened 2021-04-14
Jenkinsfile: Use credentials for docker.com

mrks:
Fixes

> Error response from daemon: toomanyrequests: You have reached your pull rate limit.
> You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ try {
 
       if (jenkinsUserName != "admin" && env.BRANCH_NAME != "master") {
         try {
-          withCredentials([usernamePassword(credentialsId: '5fe8ede9-bbdb-4803-a307-6924d4b4d9b5', usernameVariable: 'GITHUB_USERNAME', passwordVariable: 'GITHUB_TOKEN')]) {
+          withCredentials([usernamePassword(credentialsId: 'github', usernameVariable: 'GITHUB_USERNAME', passwordVariable: 'GITHUB_TOKEN')]) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -52,7 +52,8 @@ try {
       // Print the hostname to let us know on which node the docker image was executed for reproducibility.
       sh "hostname"
     }
-
+  
+  docker.withRegistry('https://registry.hub.docker.com', 'docker') {
```
mrks, line 56: I'll format this properly after your review.

mrks:
@Bensk1 - we might want to make the seminar students aware of this. If Jenkins can't pull from docker.com, chances are that on-campus students can't, either. [I doc'ed the issue here.](https://stackoverflow.com/questions/67086917/jenkins-toomanyrequests-with-docker-com/67086918#67086918)

Jenkinsfile
Bensk1, line 56: Do I understand it correctly that this is functionality offered by Jenkins? Are we logged in via the Jenkins UI and the necessary docker user data/token is taken from there?

mrks: Jenkins offers a credentials store. This is where we already store, e.g., the Github credentials. I added a docker account to that store (a new, read-only user). `'docker'` is the ID of that credential. Jenkins internally does magic to retrieve the login information when needed. This is so that we don't have to store any credentials in the Jenkinsfile.

Bensk1 (approved):

PR #2340, opened 2021-03-17
Add OperatorState & TaskState, reuse OperatorTasks, run SQLite tests w/ scheduler

julianmenzler:
This PR implements two state machines to make our code more robust and debuggable. Also, it implements the reuse of OperatorTasks and enables the scheduler for our SQLite tests (fixes #2014).

### OperatorState for AbstractOperator

```
 +---------+
 | Created |
 +---------+
      |
      | execute()
      v
 +---------+               +----------------------+                         +--------------------+
 | Running | ------------> | ExecutedAndAvailable | ----------------------> | ExecutedAndCleared |
 +---------+   execute()   +----------------------+     clear_output()      +--------------------+
               finishes                          (e.g., if consumer_count == 0)
```

### TaskState for AbstractTask

```
 +-----------+
 |  Created  |
 +-----------+
       | schedule()
       v
 +-----------+    try_mark_as_enqueued()    +----------+
 | Scheduled | ---------------------------> | Enqueued |
 +-----------+                              +----------+
       |  |                                       |
       |  +---------------------------------------+
       |                                          | try_mark_as_assigned_to_worker()
       |                                          v
       | Direct path taken by the        +------------------+
       | ImmediateExecutionScheduler     | AssignedToWorker |
       |                                 +------------------+
       |                                          |
       +------------------------------------------+
       | execute()
       v
 +-----------+
 |  Started  |
 +-----------+
       | execute() finishes
       v
 +-----------+
 |   Done    |
 +-----------+
```

### Other changes

Also, `OperatorTask::make_tasks_from_operator` now uses an unordered set to create tasks. As a side effect, it no longer returns a vector of ordered tasks. Instead, it returns a pair consisting of a vector with unordered tasks and a shared pointer to the root OperatorTask.

---

## Benchmark Results

**System**
<details>
<summary>pella - click to expand</summary>

| property | value |
| -- | -- |
| Hostname | pella |
| CPU | Intel(R) Xeon(R) CPU E7- 8870  @ 2.40GHz |
| Memory | 1.3TB |
| numactl | nodebind: 0 1 2 3  |
| numactl | membind: 0 1 2 3  |
</details>

**Commit Info and Build Time**
| commit | date | message | build time |
| -- | -- | -- | -- |
| e1389f276 | 26.04.2021 19:18 | Markus' retirement (#2359) | real 772.32 user 3791.82 sys 183.49|
[...]

PR #2356, opened 2021-04-14
Exclude correlated subqueries from diamond shapes

julianmenzler:
This PR adjusts the LQP translation so correlated subquery PQPs become separate from the PQPs that own them. It also adds two new tests and fixes an issue where correlated subqueries could block the automatic clearing of operators in the root PQP. 

Currently, the PQPs of correlated subqueries reuse operators from the root PQP. For example:

image

```sql
SELECT * FROM id_int_int_int_100 t1 WHERE (SELECT MIN(t2.id + 10) FROM id_int_int_int_100 t2 WHERE t2.id = t1.id) > 20;
```
(SQLiteTestRunner, Line395)

In contrast to uncorrelated subqueries, correlated subqueries cannot reuse operator results from their root PQPs. Consequently, we should not model it. 

After this PR, the above plan changes to the following:

image

### DIFF
--- a/src/lib/logical_query_plan/lqp_translator.cpp
+++ b/src/lib/logical_query_plan/lqp_translator.cpp
@@ -592,13 +592,15 @@ std::shared_ptr<AbstractExpression> LQPTranslator::_translate_expression(
 
       /**
        * Notes on generating subquery PQPs:
-       *  a) For uncorrelated subqueries, this LQPTranslator's cache is used to create subquery PQPs. By using the
-       *     cache, operator results can be shared between identical parts in subquery PQPs and owning PQPs.
+       *  a) For uncorrelated subqueries, operator results can be shared between identical parts in uncorrelated
+       *     subqueries and outer queries. Therefore, this LQPTranslator instance is used to deduplicate subquery PQPs
+       *     with _operator_by_lqp_node.
        *
-       *  b) For correlated subqueries, a new LQPTranslator instance is used to create subquery PQPs.
-       *     In contrast to uncorrelated subqueries, operator results cannot be shared easily between subquery PQPs
-       *     and owning PQPs because correlated subqueries imply changing parameters on a row-by-row basis.
-       *     See ExpressionEvaluator::_evaluate_subquery_expression_for_row for evaluation details.
+       *  b) In contrast to uncorrelated subqueries, correlated subqueries cannot share identical parts with outer
+       *     queries because ExpressionEvaluator::_evaluate_subquery_expression_for_row always deep-copies the whole PQP
+       *     at evaluation time. The deep copy includes both correlated and uncorrelated parts.
+       *     Consequently, a new LQPTranslator instance is used for correlated subqueries to prevent deduplication
+       *     with outer queries.
        */
       auto subquery_pqp = std::shared_ptr<AbstractOperator>();
       if (subquery_expression->is_correlated()) {
--- a/src/lib/logical_query_plan/lqp_translator.cpp
+++ b/src/lib/logical_query_plan/lqp_translator.cpp
@@ -590,7 +590,6 @@ std::shared_ptr<AbstractExpression> LQPTranslator::_translate_expression(
       const auto subquery_expression = std::dynamic_pointer_cast<LQPSubqueryExpression>(expression);
       Assert(subquery_expression, "Expected LQPSubqueryExpression");
 
-
       /**
        * Notes on generating subquery PQPs:
        *  a) For uncorrelated subqueries, this LQPTranslator's cache is used to create subquery PQPs. By using the
--- a/src/lib/logical_query_plan/lqp_translator.cpp
+++ b/src/lib/logical_query_plan/lqp_translator.cpp
@@ -92,7 +92,7 @@ std::shared_ptr<AbstractOperator> LQPTranslator::translate_node(const std::share
    * would result in multiple operators created from predicate_c and thus in performance drops.
    *
    * Deduplication:
-   * _operator_by_lqp_node compares entries by value (i.e., AbstractOperator::operator==), not by identity
+   * _operator_by_lqp_node_cache compares entries by value (i.e., AbstractOperator::operator==), not by identity
    * (shared_ptr::operator==). As a result, two separate, but equal LQP nodes will be translated into a single PQP
    * node. This prevents us from executing the same operation twice.
    *   Excursus: You would be right to wonder why this is not done on the LQP by some type of optimizer rule. That would
@@ -101,8 +101,8 @@ std::shared_ptr<AbstractOperator> LQPTranslator::translate_node(const std::share
    *   instances would also become indistinguishable. That breaks things left and right.
    */
 
-  const auto operator_iter = _operator_by_lqp_node.find(node);
-  if (operator_iter != _operator_by_lqp_node.end()) {
+  const auto operator_iter = _operator_by_lqp_node_cache.find(node);
+  if (operator_iter != _operator_by_lqp_node_cache.end()) {
     return operator_iter->second;
   }
 
@@ -112,7 +112,7 @@ std::shared_ptr<AbstractOperator> LQPTranslator::translate_node(const std::share
   // _translate_predicate_node_to_index_scan() as well, because the function creates two scans operators and returns
   // only the merging union node (all three PQP nodes share the same originating LQP node).
   pqp->lqp_node = node;
-  _operator_by_lqp_node.emplace(node, pqp);
+  _operator_by_lqp_node_cache.emplace(node, pqp);
 
   return pqp;
 }
--- a/src/lib/logical_query_plan/lqp_translator.cpp
+++ b/src/lib/logical_query_plan/lqp_translator.cpp
@@ -600,7 +600,8 @@ std::shared_ptr<AbstractExpression> LQPTranslator::_translate_expression(
        *     queries because ExpressionEvaluator::_evaluate_subquery_expression_for_row always deep-copies the whole PQP
        *     at evaluation time. The deep copy includes both correlated and uncorrelated parts.
        *     Consequently, a new LQPTranslator instance is used for correlated subqueries to prevent deduplication
-       *     with outer queries.
+       *     with outer queries. Besides, we also prevent correlated subqueries from increasing the consumer count of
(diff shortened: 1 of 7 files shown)


mrks:
Two questions:
* How did this pop up? Is this broken right now in any way or why did this change become necessary?
* In your example, there is no harm in sharing the GetTable operator, as the subplan does not have any correlated parameters. Could we identify these cases? Do we want to?

julianmenzler:
> How did this pop up?

In #2340, I tried to locate the root OperatorTask in an unordered set of executed OperatorTasks. The root OperatorTask should be the only one with an operator in `OperatorState::ExecutedAndAvailable`. Therefore, I added a DebugAssert that counts the number of operators in `OperatorState::Executed` after query execution.
For some SQLiteTestRunner queries, the Assert failed. I wondered, why the GetTable operator from the screenshot above had still one registered consumer, although the query had already executed. Then, I noticed that "skeleton PQPs" of correlated subqueries can register themselves as consumers at parts of the root PQP. But, in contrast to all other operators, the ones in the "skeleton PQP" never get executed. (because the ExpressionEvaluation deep-copies them for evaluation) Consequently, the "skeleton" operators from correlated subqueries cannot deregister themselves from operators in the root query.

> Is this broken right now in any way or why did this change become necessary?

No. Nothing is logically broken. But the automatic clearing of operators may be hindered in case of correlated subqueries.

> In your example, there is no harm in sharing the GetTable operator, as the subplan does not have any correlated parameters. Could we identify these cases? Do we want to?

True. But, since the ExpressionEvaluator deep-copies GetTable anyway, it is not useful to model it as a diamond shape.

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -92,7 +92,7 @@ std::shared_ptr<AbstractOperator> LQPTranslator::translate_node(const std::share
    * would result in multiple operators created from predicate_c and thus in performance drops.
    *
    * Deduplication:
-   * _operator_by_lqp_node compares entries by value (i.e., AbstractOperator::operator==), not by identity
+   * _operator_by_lqp_node_cache compares entries by value (i.e., AbstractOperator::operator==), not by identity
```
mrks, line 95: For me, a "cache" is something that simply stores the result of an idempotent computation. Clearing the cache at any point should have no effect on the result of the algorithm (here: the LQP translation) other than its execution duration. As such, in my understanding, this is not a cache.

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -590,7 +590,22 @@ std::shared_ptr<AbstractExpression> LQPTranslator::_translate_expression(
       const auto subquery_expression = std::dynamic_pointer_cast<LQPSubqueryExpression>(expression);
       Assert(subquery_expression, "Expected LQPSubqueryExpression");
 
-      const auto subquery_pqp = translate_node(subquery_expression->lqp);
+      /**
+       * Notes on generating subquery PQPs:
+       *  a) For uncorrelated subqueries, this LQPTranslator's cache is used to create subquery PQPs. By using the
```
mrks, line 595: see above re. "cache".

src/lib/logical_query_plan/lqp_translator.cpp
```diff
@@ -590,7 +590,22 @@ std::shared_ptr<AbstractExpression> LQPTranslator::_translate_expression(
...
+       *     cache, operator results can be shared between identical parts in subquery PQPs and owning PQPs.
+       *
+       *  b) For correlated subqueries, a new LQPTranslator instance is used to create subquery PQPs.
+       *     In contrast to uncorrelated subqueries, operator results cannot be shared easily between subquery PQPs
+       *     and owning PQPs because correlated subqueries imply changing parameters on a row-by-row basis.
```
mrks, line 600: Someone reading this might have the same question that I had in the previous discussion: What about uncorrelated parts of the subquery? Maybe explain that correlated subqueries are executed using `deep_copy`, which does not identify uncorrelated parts of the copied plan? As a result, those parts of the plan that would be deduplicated here would be duplicated, anyway, correct?

src/test/lib/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -748,6 +748,108 @@ TEST_F(LQPTranslatorTest, LimitNode) {
   EXPECT_EQ(*limit_op->row_count_expression(), *value_(2));
 }
 
+TEST_F(LQPTranslatorTest, CreateTable) {
```
mrks, line 751: What happend here?

src/test/lib/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -796,6 +898,95 @@ TEST_F(LQPTranslatorTest, DiamondShapeSimple) {
...
+  ASSERT_EQ(greater_than_predicate->right_operand()->type, ExpressionType::PQPSubquery);
+  const auto pqp_subquery_expression =
+      std::static_pointer_cast<const PQPSubqueryExpression>(greater_than_predicate->right_operand());
+  ASSERT_FALSE(pqp_subquery_expression->is_correlated());
+  ASSERT_EQ(pqp_subquery_expression->pqp->type(), OperatorType::Projection);
+  const auto subquery_projection = std::static_pointer_cast<const Projection>(pqp_subquery_expression->pqp);
+  ASSERT_EQ(subquery_projection->left_input()->type(), OperatorType::JoinHash);
+  const auto subquery_join = std::static_pointer_cast<const JoinHash>(subquery_projection->left_input());
+  ASSERT_EQ(subquery_join->left_input()->type(), OperatorType::GetTable);
+  ASSERT_EQ(subquery_join->right_input()->type(), OperatorType::GetTable);
```
mrks, line 939: This could really use some newlines (same below).

src/test/lib/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -796,6 +898,95 @@ TEST_F(LQPTranslatorTest, DiamondShapeSimple) {
   EXPECT_EQ(pqp->left_input()->left_input()->left_input(), pqp->right_input()->left_input()->left_input());
 }
 
+TEST_F(LQPTranslatorTest, DiamondShapeIncludeUncorrelatedSubqueries) {
+  // Tests if uncorrelated subqueries become part of the root diamond shape.
```
mrks, line 902: > Tests that PQP parts that are shared between an (un)correlated subquery and the outer plan are (not) de-duplicated? (same below)

src/test/lib/logical_query_plan/lqp_translator_test.cpp
```diff
@@ -796,6 +898,95 @@ TEST_F(LQPTranslatorTest, DiamondShapeSimple) {
...
+  ASSERT_FALSE(pqp_subquery_expression->is_correlated());
+  ASSERT_EQ(pqp_subquery_expression->pqp->type(), OperatorType::Projection);
+  const auto subquery_projection = std::static_pointer_cast<const Projection>(pqp_subquery_expression->pqp);
+  ASSERT_EQ(subquery_projection->left_input()->type(), OperatorType::JoinHash);
+  const auto subquery_join = std::static_pointer_cast<const JoinHash>(subquery_projection->left_input());
+  ASSERT_EQ(subquery_join->left_input()->type(), OperatorType::GetTable);
+  ASSERT_EQ(subquery_join->right_input()->type(), OperatorType::GetTable);
+
+  // Compare addresses to check if the uncorrelated PQP subquery reuses the GetTable instance from its owning PQP.
+  EXPECT_EQ(subquery_join->left_input().get(), get_table_int_float.get());
```
mrks, line 942: Why do you call `get()`? `operator==` works for `shared_ptr` just fine.

src/lib/logical_query_plan/lqp_translator.cpp
julianmenzler: > Clearing the cache at any point should have no effect on the result of the algorithm (here: the LQP translation) other than its execution duration.

Good point. I revert this change.

src/test/lib/logical_query_plan/lqp_translator_test.cpp
julianmenzler: Changed to 

> Tests that PQP parts that are shared between an uncorrelated subquery and the outer plan are deduplicated.

and 

> Tests that PQP parts that are shared between a correlated subquery and the outer plan are NOT deduplicated.

src/lib/logical_query_plan/lqp_translator.cpp
julianmenzler: Changed to 

```cpp
/**
 * Notes on generating subquery PQPs:
 *  a) For uncorrelated subqueries, operator results can be shared between identical parts in uncorrelated
 *     subqueries and outer queries. Therefore, this LQPTranslator instance is used to deduplicate subquery PQPs
 *     with _operator_by_lqp_node.
 *
 *  b) In contrast to uncorrelated subqueries, correlated subqueries cannot share identical parts with outer
 *     queries because ExpressionEvaluator::_evaluate_subquery_expression_for_row always deep-copies the whole PQP
 *     at evaluation time. The deep copy includes both correlated and uncorrelated parts.
 *     Consequently, a new LQPTranslator instance is used for correlated subqueries to prevent deduplication
 *     with outer queries.
 */
```

mrks: The last sentence will leave people wondering why we do extra work to avoid the deduplication. Add a sentence about the consumer count and I'm happy.

julianmenzler: Added 

>  Besides, we also prevent correlated subqueries from increasing the consumer count of outer query operators, which would otherwise block the automatic clearing of results.

mrks: "Besides"? Isn't that the whole reason why this PR exists?

This might seem nitpicky, but those are things that people might get confused about in a year from now.

julianmenzler: "This also prevents correlated subqueries ..." ?

mrks: Haha, I was just about to complain about the "also". Sounds good. ;)

mrks (approved):
Pending the one open item as discussed.

PR #2362, opened 2021-05-18
Remove TPC-DS Metadata Table Files

mweisgut:
This PR removes the unused TPC-DS metadata table files (`csv.json`).

Ref. #1716 (metadata table files were added)
Ref. #1750 (introduction of the TPC-DS table generator)


Bouncner:
Can you explain we had these files in the first place and what we are doing now that they are no longer used?

mweisgut:
Initially, the metadata files were required for the file-based TPC-DS table generation (#1750) since the `CsvParser` was used that requires the metadata. With the in-memory table generator introduced in #1750, we don't need these metadata files anymore for the execution of the TPC-DS benchmark because the corresponding table schemas are now created by the table generator using hardcoded schema information.

However, I just realized that the metadata files are used for `hyriseSystemTest`.

Bouncner:
Is there a good reason to use these files over the table generator? I’d hope the CSV parser is sufficiently tested anyways.

mweisgut:
The metadata files are used together with the generated data (`.csv` files) of the official TPC-DS tool (`dsdgen`) to verify the correctness of the `TPCDSTableGenerator` (see https://github.com/hyrise/hyrise/blob/54c2c122957f77ac0606ca2f031dd9d8ba76f9f8/src/test/benchmarklib/tpcds/tpcds_db_generator_test.cpp).

Bouncner:
Ah ok, I see.

PR #2367, opened 2021-06-01
Add semi and anti join to join sort merge

Alexander-Dubrawski:
This PR adds the Semi and Anti join to the SMJ for the following case:

* Equal case with multiple predicates

Moreover, the Anti join needs to be differentiated between Null as true and Null as false. These two cases and the Semi case need to be handled differently.

## Quick SMJ recap

The values of the left and right join columns are grouped into runs (rows of a run have the same value). The values of the runs are compared and then we determine if we advanced the right run, left run, or both runs. If the comparison result is equal, we advance both runs. If it is less, we advance the left one. If it is greater, we advance the right one (the arrows are representing the runs (entries with the same value) we are currently looking at):

```
 Compare result equal:   | After advancing both runs:
        a  | b           |         a  | b
      -----|----         |       -----|----
    -> [2] | [2] <-      |        [2] | [2]
       [3] | [4]         |     -> [3] | [4] <-
       [6] | [7]         |        [6] | [7]
Compare result less:     | After advancing left run:
       a  | b            |         a  | b
       ---|---           |       -----|----
   -> [3] | [4] <-       |        [3] | [4] <-
      [6] | [5]          |     -> [6] | [5]
      [7] | [9]          |        [7] | [9]
compare result greater:  | After advancing right run:
       a  | b            |         a  | b
       ---|---           |       -----|----
   -> [6] | [4] <-       |     -> [6] | [4]
      [7] | [5]          |        [7] | [5] <-
      [8] | [9]          |        [8] | [9]
```

We can then use in every step the comparison information to join the combinations. For example, if we have an equal comparison and an equal join, we know that we emit the cross product between the rows from the left and right and run.

## Semi implementation

We need to check that we find for the current left run a right one that has the same value. So we are only interested in the equal case and we can directly emit the left run. If we have the predicate condition `equals` with multiple predicates we need to check for every row inside the left run that there is at least one row inside the right run where all predicates are satisfied. For example, if a should be equal to c and b < d:

```
	left run      |   right run
--------------------------------	
row | a | b       row | c | d
1.  | 2 | 10      1.  | 2 | 8
2.  | 2 | 20      2.  | 2 | 12
3.  | 2 | 30      3.  | 2 | 22
```

Then row 3 will inside the left run will find no match in the right run.

## Anti Null as False implementation: 

We only want to emit the rows from the left table where we know that there is **no match** inside the right table. So the Equal expression is always false for the row. For that, we can use the run comparison information from the SMJ.

If the comparison is less it means that there can not be a match. The reason is the following:
```
| (1) -> [5] | [5] <- |
|        [6] | [7]    |
| ------------------- |
| (2)    [5] | [5]    |
|     -> [6] | [7] <- |
```
If the comparison result was equal in the step before it means that the right and the left pointer were
moved. Since the runs are sorted we can be sure that there is no equal value on the right side for the
left value.
```
| (1) -> [4] | [7] <- |
|        [6] |        |
| ------------------- |
| (2)   [4] | [7] <-  |
|    -> [6] |         |
```
If the comparison value in the last run was less, that means that only the left pointer was moved. If the
the left value is still smaller we can be sure that there can not be an equal value on the right side.
```
| (1) -> [6] | [3] <- |
|            | [7]    |
| ------------------- |
| (2) -> [6] | [3]    |
|            | [7] <- |
```
If the comparison result was greater, it means that only the right pointer was moved. If the result is less
that means that there can not be an equal value, since then the right pointer would be pointing to it.
We know that if we find **no** equal math for a row the join condition with the multiple predicates will be
in any case false:  *False AND ...* . So we do not need to check in this case if the multiple predicates are
satisfied or not.
If we find an equal match of runs and we have multiple predicates we need to check that there is no case where all predicates are satisfied for a row in the left run. In this case, the logical expression *True (equal) AND PREDICATES* needs to be in all cases False. 

### Edge cases

* If the right table is empty we just return the left one.
* We emit all rows where the values are null. 

## Anti Null as True implementation:

Is handled the same as in Anti null as false. The case of multiple predicates is also handled the same with the addition that we also need to check that there is no null row (comparison value is null) where the multiple predicates are satisfied. 

### Edge cases

If we have only one predicate we check if one value in the column of the right table is null. If so we can return an empty left table. If the join mode is anti null as true with secondary predicates, we need to check for every left null row that:
* There is no match with one of the right null rows where all predicates are satisfied.
* There is no match with one of the left rows (not null) where all predicates are satisfied.
We need to check the two cases because if  NULL AND X AND Y ... is for every combination False, we need to emit the row.


Bouncner (commented):
More will follow. Just making sure your pipeline does not run empty.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <memory>
 #include <optional>
+#include <set>
```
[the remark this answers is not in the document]
Alexander-Dubrawski: Yes, but only once. https://github.com/hyrise/hyrise/blob/4b315062696cee8f628e731fac7db1e99cb5314d/src/lib/operators/join_sort_merge.cpp#L662

src/lib/operators/join_sort_merge.cpp
```diff
@@ -28,11 +29,46 @@ namespace opossum {
...
+    case JoinMode::AntiNullAsFalse:
+      if (config.predicate_condition == PredicateCondition::Equals) {
+        return true;
+      }
+      if (config.predicate_condition == PredicateCondition::NotEquals && !config.secondary_predicates) {
+        return true;
+      }
+      return false;
+    default:
+      if (config.predicate_condition != PredicateCondition::NotEquals) {
```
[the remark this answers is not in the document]
Alexander-Dubrawski: How about:

```
  // The sort-merge join supports inner joins for all predicate conditions and multiple predicates. Cross joins are not
  // supported. Semi and anti joins are supported for multiple predicates if the predicate condition is equal. If
  // the predicate condition is not equals, only single predicates are supported. All other joins are supported if the
  // predicate condition is not equals.
```

src/lib/operators/join_sort_merge.cpp
```diff
@@ -845,14 +1176,72 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+    for (ChunkID cluster_id{0}; cluster_id < chunk_count; ++cluster_id) {
+      const auto chunk = table.get_chunk(cluster_id);
+      const auto segment = chunk->get_segment(_sort_merge_join._primary_predicate.column_ids.second);
+      segment_iterate<T>(*segment, [&](const auto& position) {
+        if (position.is_null()) {
+          has_null = true;
+          return;
+        }
+      });
+      if (has_null) return true;
```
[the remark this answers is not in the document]
Alexander-Dubrawski: We are iterating over every segment and are stopping at the first null value that we found.  If we would return `return has_null` we would stop after the first segment and could miss null values in the following ones.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -845,14 +1176,72 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+      }
+      if (_secondary_join_predicates.empty()) {
+        auto right_side_has_null = _check_if_table_contains_null(*(_sort_merge_join.right_input_table()));
+        if (right_side_has_null) {
+          return Table::create_dummy_table(_sort_merge_join.left_input_table()->column_definitions());
+        }
+      }
+    }
+
+    auto sort = true;
```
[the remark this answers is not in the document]
Alexander-Dubrawski: How about `sort_tables` or `_sort_clusters` ?

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Bouncner: 👍

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Bouncner: What about changing it to `RowHashSet`?
Now that we already use `std::unordered_set` for hashing anyways.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -65,6 +101,8 @@ std::shared_ptr<const Table> JoinSortMerge::_on_execute() {
                    !_secondary_predicates.empty(), left_input_table()->type(), right_input_table()->type()}),
          "JoinSortMerge doesn't support these parameters");
 
+  Assert(_mode != JoinMode::Cross, "Sort merge join does not support cross joins.");
```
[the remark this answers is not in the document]
Alexander-Dubrawski: Not really. I removed it.

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Bouncner: Stupid me.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -226,6 +264,206 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+      // from the right side with an equal value to the ones from the left.
+      if (!match) {
+        right_range.for_every_row_id(_sorted_right_table, [&](RowID right_row_id) {
+          if (multi_predicate_join_evaluator.satisfies_all_predicates(left_row_id, right_row_id)) {
+            match = true;
+            return;
+          }
+        });
+      }
+      if (!match) {
```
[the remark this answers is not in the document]
Alexander-Dubrawski: Please have a look if you are happy with the merge 🙃 .

src/lib/operators/join_sort_merge.cpp
```diff
@@ -592,8 +900,10 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
 
   // Determines the largest value in a sorted materialized table.
   const T& _table_max_value(const MaterializedSegmentList<T>& sorted_table) {
-    DebugAssert(_primary_predicate_condition != PredicateCondition::Equals,
-                "The table needs to be sorted for _table_max_value() which is only the case for non-equi joins");
+    DebugAssert(
+        !((_mode == JoinMode::AntiNullAsFalse || _mode == JoinMode::AntiNullAsTrue) &&
```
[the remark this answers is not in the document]
Alexander-Dubrawski: I think it is an artifact from a solution in the early stage of this branch. I removed it since we are not needing this assertion anymore.

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Bouncner: `sort_clusters` sounds good to me.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -845,14 +1160,72 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+      }
+    }
+
+    auto sort = true;
+    if (_mode == JoinMode::AntiNullAsFalse || _mode == JoinMode::AntiNullAsTrue) {
+      // Since the not equal anti join does not support multiple predicates we are only interested in equal values. For
+      // that, we do not need to sort the tables.
+      sort = !(_primary_predicate_condition == PredicateCondition::NotEquals);
+    } else {
+      sort = !(_primary_predicate_condition == PredicateCondition::Equals);
```
Bouncner, line 1217: `sort = _primary_predicate_condition != PredicateCondition::Equals;`

src/lib/operators/join_sort_merge.cpp
```diff
@@ -1354,7 +1335,8 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
           }
         }
 
-        for (size_t row = 0; row < left_null_rows_matches.size(); row++) {
+        const auto left_null_rows_match_count = left_null_rows_matches.size();
+        for (size_t row = 0; row < left_null_rows_match_count; ++row) {
```
Bouncner, line 1342: auto row

src/lib/operators/join_sort_merge/radix_cluster_sort.hpp
```diff
@@ -56,7 +56,7 @@ class RadixClusterSort {
         _right_input_table{right},
         _left_column_id{column_ids.first},
         _right_column_id{column_ids.second},
-        _sort{sort},
+        _sort_result_clusters{sort},
```
Bouncner, line 59: I'd say either `_sort_resulting_clusters` or `_sort_clusters` (my favorite).

src/lib/operators/join_sort_merge.cpp
```diff
@@ -226,6 +260,204 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+      right_range.for_every_row_id(_sorted_right_table, [&](RowID right_row_id) {
+        if (multi_predicate_join_evaluator.satisfies_all_predicates(left_row_id, right_row_id)) {
+          match = true;
+          return;
+        }
+      });
+      if (match && _mode == JoinMode::Semi) {
+        _emit_left_only(output_cluster, left_row_id);
+      }
+      if (!match && _mode == JoinMode::AntiNullAsFalse) {
```
Bouncner, line 280:
```
if ((match &&  _mode == JoinMode::Semi) || (!match && _mode == JoinMode::AntiNullAsFalse)) {
  _emit_left_only(output_cluster, left_row_id);
}```

File: src/lib/operators/join_sort_merge.cpp (hunk 9 above)
Comment (Alexander-Dubrawski) [reply src/lib/operators/join_sort_merge.cpp:1342]: If I use auto I get the error `comparison of integers of different signs: 'int' and 'const unsigned long'`. Do you have a hint on how we could still use auto?

File: src/lib/operators/join_sort_merge/radix_cluster_sort.hpp (hunk 10 above)
Comment (Alexander-Dubrawski) [reply src/lib/operators/join_sort_merge/radix_cluster_sort.hpp:59]: I used `_sort_resulting_clusters` because we have already a function with the name  `_sort_clusters `.

File: src/lib/operators/join_sort_merge.cpp (hunk 1 above)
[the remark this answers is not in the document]
Comment (Alexander-Dubrawski) [reply src/lib/operators/join_sort_merge.cpp:6]: True, I replaced the map with the HashSet.

File: src/lib/operators/join_sort_merge.cpp [hunk 12]
```diff
@@ -893,6 +1293,79 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+        const auto null_rows_left_size = _null_rows_left.size();
+        std::vector<bool> left_null_rows_matches(null_rows_left_size, false);
+        null_output_left.reserve(null_rows_left_size);
+
+        auto jobs = std::vector<std::shared_ptr<AbstractTask>>{};
+        const auto right_table_row_count = _right_input_table->row_count();
+
+        // We do not want to span to many threads. The upper limit is the number of CPU cores.
+        const auto num_cpus = Hyrise::get().topology.num_cpus();
+        size_t bunch_size = 1;
```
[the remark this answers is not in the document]
Alexander-Dubrawski: If the join mode is anti null as true with secondary predicates, we need to check for every left null row that:
1. There is no match with one of the right null rows where all predicates are satisfied
2. There is no match with one of the left rows (not null) where all predicates are satisfied

We could do this by iterating over every single left row with a null value. My idea was that if we have a lot of rows to iterate over we could do this in parallel. For that, I divide the rows into bunches and these bunches are handled in separate threads. The size of a bunch is calculated with the number of CPUs.

This approach has some drawbacks:
1. If we have a lot of rows and for example 20 bunches and only the first row is a match, 19 bunches would do unneeded work.
2. I am not sure if there is even a case where we have so many null values where we would benefit from the parallel execution here.

Do you think we should keep it simple and just iterate in one thread over all null rows?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -845,14 +1160,72 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+
  public:
   std::shared_ptr<const Table> _on_execute() override {
-    const auto include_null_left = (_mode == JoinMode::Left || _mode == JoinMode::FullOuter);
-    const auto include_null_right = (_mode == JoinMode::Right || _mode == JoinMode::FullOuter);
-    auto radix_clusterer = RadixClusterSort<T>(
-        _sort_merge_join.left_input_table(), _sort_merge_join.right_input_table(),
-        _sort_merge_join._primary_predicate.column_ids, _primary_predicate_condition == PredicateCondition::Equals,
-        include_null_left, include_null_right, _cluster_count, _performance);
+    if (_mode == JoinMode::Semi) {
```
Bouncner, line 1184: Can you add a comment that you're now going to check for early exit cases?

Further, I think we should remove the code for non-equal semi joins. It bloats the code is not going to occur in Hyrise any time soon.

Besides that, I don't like this current code too much. Too many lines of code for too simple checks. Can we somehow reduce that?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -845,14 +1160,72 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+          return Table::create_dummy_table(_sort_merge_join.left_input_table()->column_definitions());
+        }
+      }
+    }
+
+    auto sort = true;
+    if (_mode == JoinMode::AntiNullAsFalse || _mode == JoinMode::AntiNullAsTrue) {
+      // Since the not equal anti join does not support multiple predicates we are only interested in equal values. For
+      // that, we do not need to sort the tables.
+      sort = !(_primary_predicate_condition == PredicateCondition::NotEquals);
```
Bouncner: Similar to the semi joins: I would remove the non-equal cases for simplicity and add a comment to the supports function, that this PR contains code for it but we discarded it for simplicity.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -845,14 +1160,72 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+        auto right_side_has_null = _check_if_table_contains_null(*(_sort_merge_join.right_input_table()));
+        if (right_side_has_null) {
+          return Table::create_dummy_table(_sort_merge_join.left_input_table()->column_definitions());
+        }
+      }
+    }
+
+    auto sort = true;
+    if (_mode == JoinMode::AntiNullAsFalse || _mode == JoinMode::AntiNullAsTrue) {
+      // Since the not equal anti join does not support multiple predicates we are only interested in equal values. For
```
Bouncner, line 1210: I don't understand the comment to be honest.

Bouncner (commented):
Next batch.

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Bouncner: Is it GitHub or aren't there any changes? I still see `if (!match)` twice.

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Bouncner: Yes, please remove all threading code.

Actually, we don't allow operators to spawn threads on their own. That's what we have tasks for. But I don't see a good reason to have tasks here for this edge case. Let's tackle that in case it will ever become a bottleneck.

src/lib/operators/join_sort_merge.cpp
Bouncner: `size_t row = size_t{0};`

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Alexander-Dubrawski: It is Github. The new Code is: 

```cpp
      if (!match) {
        right_range.for_every_row_id(_sorted_right_table, [&](RowID right_row_id) {
          if (multi_predicate_join_evaluator.satisfies_all_predicates(left_row_id, right_row_id)) {
            match = true;
            return;
          }
        });
        if (match) return;
        _emit_left_only(output_cluster, left_row_id);
      }
```

src/lib/operators/join_sort_merge.cpp
```diff
@@ -226,6 +260,204 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+    });
+  }
+
+  void _emit_multipredicate_anti_null_as_true(size_t output_cluster, TableRange left_range, TableRange right_range,
+                                              MultiPredicateJoinEvaluator& multi_predicate_join_evaluator) {
+    left_range.for_every_row_id(_sorted_left_table, [&](RowID left_row_id) {
+      auto match = false;
+      // We need to make sure that there is not one combination between the left rows from the run and the null rows
+      // from the right where all predicates are satisfied.
+      for (const auto& right_row_id : _null_rows_right) {
```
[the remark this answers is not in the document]
Alexander-Dubrawski: Yes, because NULL is always true in this case. Probably it makes more sense to check if there is a match with the right run and then check the null rows (If there was no match). What do you think?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -226,6 +260,204 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+            return;
+          }
+        });
+        if (match) return;
+        _emit_left_only(output_cluster, left_row_id);
+      }
+    });
+  }
+
+  void _emit_multipredicate_anti_null_as_true_left_range_only(
```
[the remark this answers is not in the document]
Alexander-Dubrawski: Will do. So you don't like `if (match) return;`?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -226,6 +260,204 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+        // the same.
+        if (compare_result != CompareResult::Equal) {
+          _emit_left_range_only(cluster_id, left_run);
+        }
+        break;
+      case PredicateCondition::GreaterThan:
+      case PredicateCondition::GreaterThanEquals:
+      case PredicateCondition::LessThan:
+      case PredicateCondition::LessThanEquals:
+        break;
```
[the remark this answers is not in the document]
Alexander-Dubrawski: How about we use `throw std::logic_error("Unsupported PredicateCondition");` instead of the `break;` ?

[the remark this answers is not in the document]
Bouncner: Yes, but rather "Unsupported predicate condition"?

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Bouncner: It's a guard clause at the very end of the block. I does not increase understanding (I think) and the alternative is even shorter codeline-wise.

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Bouncner: Hmm, assuming there are most often less NULLs than values, I think the current approach is fine. It's still baffling me somehow, but multi-predicate joins simply aren't that simple I guess.

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Bouncner: As said below, I don't see a good reason not to use:
`if (!match)  _emit_left_only(output_cluster, left_row_id);`

src/lib/operators/join_sort_merge.cpp
```diff
@@ -568,10 +839,41 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+          if (multi_predicate_join_evaluator) {
+            _emit_multipredicate_anti_null_as_true_left_range_only(cluster_id, left_rest,
+                                                                   *multi_predicate_join_evaluator);
+          } else {
+            _emit_left_range_only(cluster_id, left_rest);
+          }
+        }
+      } else if (_mode == JoinMode::Semi) {
+        if (_primary_predicate_condition == PredicateCondition::NotEquals) {
+          // We only get here if the right table has only one value. Since we know that there is no equal match
```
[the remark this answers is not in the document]
Alexander-Dubrawski: I removed the `not equal` case for the Semi mode.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -813,7 +1117,19 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
                                                  *_sort_merge_join.right_input()->get_output(), _mode,
                                                  _secondary_join_predicates);
         }
-
+        if (_mode == JoinMode::Semi && _primary_predicate_condition == PredicateCondition::NotEquals) {
+          // We need to check if there are at lest two values. As a result all left rows will always find at least one
```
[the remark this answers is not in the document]
Alexander-Dubrawski: I removed the not equal case for the Semi mode.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -517,6 +758,22 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
 
   // Performs the join on a single cluster. Runs of entries with the same value are identified and handled together.
   // This constitutes the merge phase of the join. The output combinations of row ids are determined by _join_runs.
+  // The main logic of the algorithm is the following:
+  // We have for the left and right side a list of runs. There are also pointers that are pointing to the current run
```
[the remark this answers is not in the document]
Alexander-Dubrawski: I improved the comment. I hope it is now easier to understand:

```
  // Performs the join on a single cluster. Runs of entries with the same value are identified and handled together.
  // This constitutes the merge phase of the join. The output combinations of row ids are determined by _join_runs.
  // The main logic of the algorithm is the following:
  // The values of the left and right join columns are grouped into runs (entries with the same value). The values of
  // the runs are compared and then we determine if we advanced the right run, left run or both runs. If the comparison
  // result is equal, we advance both runs, if it is less we advance the left one if it is greater we advance the right
  // one (the arrows are representing the runs (entries with the same value) we are currently looking at):
  //  Compare result equal:   | After advancing run:
  //         a  | b           |         a  | b
  //       -----|----         |       -----|----
  //     -> [2] | [2] <-      |        [2] | [2]
  //        [3] | [4]         |     -> [3] | [4] <-
  //        [6] | [7]         |        [6] | [7]
  //
  // Compare result less:     | After advancing run:
  //        a  | b            |         a  | b
  //        ---|---           |       -----|----
  //    -> [3] | [4] <-       |        [3] | [4] <-
  //       [6] | [7]          |     -> [6] | [7]
  //       [7] | [9]          |        [7] | [9]
  //
  // compare result greater:  | After advancing run:
  //        a  | b            |         a  | b
  //        ---|---           |       -----|----
  //    -> [6] | [4] <-       |     -> [6] | [3]
  //       [7] | [7]          |        [7] | [7] <-
  //       8] | [9]          |         [8] | [9]
  // We can than use in every step the comparison information to join the combinations. For example if we have the
  // equal comparison and an equal join, we now that we emit the cross product between the rows from the left and right
  // and run.
```

src/lib/operators/join_sort_merge.cpp
Alexander-Dubrawski: I removed the `non-equal ` semi-join. Since GitHub is confusing me quite a bit right now, can you tell me again which part of the code you do not like at the moment?

src/lib/operators/join_sort_merge.cpp
Alexander-Dubrawski: The comment is not relevant anymore, since I removed the `not equal` case.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -517,6 +732,34 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
 
   // Performs the join on a single cluster. Runs of entries with the same value are identified and handled together.
   // This constitutes the merge phase of the join. The output combinations of row ids are determined by _join_runs.
+  // The main logic of the algorithm is the following:
+  // The values of the left and right join columns are grouped into runs (entries with the same value). The values of
```
Bouncner, line 736: "into runs (rows of a run have the same value)"?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -517,6 +732,34 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+  // the runs are compared and then we determine if we advanced the right run, left run or both runs. If the comparison
```
Bouncner, line 737: I guess that should be rather "then we determine if we advance the right run, left runCOMMA, or both runs." (comma and advance in present).

src/lib/operators/join_sort_merge.cpp
```diff
@@ -517,6 +732,34 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+  // result is equal, we advance both runs, if it is less we advance the left one if it is greater we advance the right
```
Bouncner, line 738: "equal, we advance both runs. If it is less, we advance the left one. If it is greater, we ...".

src/lib/operators/join_sort_merge.cpp
```diff
@@ -517,6 +732,34 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+  //       [6] | [5]          |     -> [6] | [5]
+  //       [7] | [9]          |        [7] | [9]
+  //
+  // compare result greater:  | After advancing right run:
+  //        a  | b            |         a  | b
+  //        ---|---           |       -----|----
+  //    -> [6] | [4] <-       |     -> [6] | [4]
+  //       [7] | [5]          |        [7] | [5] <-
+  //       [8] | [9]          |        [8] | [9]
+  // We can than use in every step the comparison information to join the combinations. For example if we have the
```
Bouncner, line 760: than > then

src/lib/operators/join_sort_merge.cpp
```diff
@@ -568,10 +825,28 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
     // Join the rest of the unfinished side, which is relevant for outer joins and non-equi joins
     const auto right_rest = TableRange(cluster_id, right_run_start, right_size);
     const auto left_rest = TableRange(cluster_id, left_run_start, left_size);
+
     if (left_run_start < left_size) {
-      _join_runs(left_rest, right_rest, CompareResult::Less, multi_predicate_join_evaluator, cluster_id);
+      if (_mode == JoinMode::AntiNullAsFalse) {
+        // We know that there are no matches for the rest of  the left runs. Through the radix clustering we know that
```
Bouncner, line 831: double space

Bouncner: "Because of the radix clustering, " ?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -845,14 +1120,65 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+    if (_mode == JoinMode::Semi) {
+      if (_sort_merge_join.right_input_table()->empty()) {
+        return Table::create_dummy_table(_sort_merge_join.left_input_table()->column_definitions());
+      }
+    } else if (_mode == JoinMode::AntiNullAsFalse && _sort_merge_join.right_input_table()->empty()) {
+      return _sort_merge_join.left_input_table();
+    } else if (_mode == JoinMode::AntiNullAsTrue) {
+      if (_sort_merge_join.right_input_table()->empty()) {
+        return _sort_merge_join.left_input_table();
+      } else if (_secondary_join_predicates.empty()) {
```
Bouncner, line 1158: guard clause.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -845,14 +1120,65 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
  public:
   std::shared_ptr<const Table> _on_execute() override {
-    const auto include_null_left = (_mode == JoinMode::Left || _mode == JoinMode::FullOuter);
-    const auto include_null_right = (_mode == JoinMode::Right || _mode == JoinMode::FullOuter);
-    auto radix_clusterer = RadixClusterSort<T>(
-        _sort_merge_join.left_input_table(), _sort_merge_join.right_input_table(),
-        _sort_merge_join._primary_predicate.column_ids, _primary_predicate_condition == PredicateCondition::Equals,
-        include_null_left, include_null_right, _cluster_count, _performance);
+    if (_mode == JoinMode::Semi) {
+      if (_sort_merge_join.right_input_table()->empty()) {
```
Bouncner, line 1150: You can merge both if's here as well as done in the following else if.

src/lib/operators/join_sort_merge.cpp
Bouncner: Hmm, never mind. A comment stating that we are going to check for simple early exists should be sufficient.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -866,7 +1192,11 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
 
     _perform_join();
 
-    if (include_null_left || include_null_right) {
+    if (_mode == JoinMode::AntiNullAsFalse) {
```
Bouncner, line 1195: That looks like it's already handled in row 1176. Bool can be made const then.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -893,6 +1223,49 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
       }
     }
+
+    if (_mode == JoinMode::AntiNullAsTrue) {
+      // If the join mode is anti null as true with secondary predicates, we need to check for every left null row that:
+      // 1. There is no match with one of the right null rows where all predicates are satisfied.
+      // 2. There is no match with one of the left rows (not null) where all predicates are satisfied.
+      // We need to check the two cases because if  NULL AND X AND Y ... is for every combination False, we need
+      // to emit the row.
+      if (!_secondary_join_predicates.empty()) {
```
Bouncner, line 1233: Merge the IFs?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -893,6 +1223,49 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+        auto null_output_left = RowIDPosList();
+
```
Bouncner, line 1235: remove empty line

src/lib/operators/join_sort_merge.cpp
```diff
@@ -893,6 +1223,49 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+        null_output_left.reserve(_null_rows_left.size());
+
+        MultiPredicateJoinEvaluator multi_predicate_join_evaluator(*_sort_merge_join._left_input->get_output(),
+                                                                   *_sort_merge_join.right_input()->get_output(), _mode,
+                                                                   _secondary_join_predicates);
+        auto end_of_right_table = _end_of_table(_sorted_right_table);
```
Bouncner, line 1241: const?

Bouncner: Right range also const and maybe move `end_of_right_table` into `.to()`.

src/lib/operators/join_sort_merge.cpp
```diff
@@ -893,6 +1223,49 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
         _output_pos_lists_right.push_back(std::move(null_output_right));
       }
     }
+
+    if (_mode == JoinMode::AntiNullAsTrue) {
```
Bouncner, line 1227: Can you add a sentence about the rows that are not NULLs?

src/lib/operators/join_sort_merge.cpp
```diff
@@ -893,6 +1223,49 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+      // If the join mode is anti null as true with secondary predicates, we need to check for every left null row that:
+      // 1. There is no match with one of the right null rows where all predicates are satisfied.
+      // 2. There is no match with one of the left rows (not null) where all predicates are satisfied.
```
Bouncner, line 1230: I don't understand. We compare a left NULL value with another left non-NULL value?

Bouncner (commented):
Many small typo things. Still need to fully digest the actual join code with the "pointer" moving.

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Bouncner: 👍

src/lib/operators/join_sort_merge.cpp
```diff
@@ -34,29 +34,28 @@ bool JoinSortMerge::supports(const JoinConfiguration config) {
   }
 
   // The sort-merge join supports inner joins for all predicate conditions and multiple predicates. Cross joins are not
-  // supported. Semi and anti joins are supported for multiple predicates if the predicate condition is equal. If
-  // the predicate condition is not equals, only single predicates are supported. All other joins are supported if the
-  // predicate condition is not equals.
+  // supported. Semi and anti joins are supported for multiple predicates if the predicate condition is equal. Anti and
+  // semi joins where the predicate condition is not equals, are not supported. The reason is that we do not have a
+  // query at the moment where the predicate is not equals and the implementation would increase the complexity. You
```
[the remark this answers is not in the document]
Alexander-Dubrawski: yes :)

src/lib/operators/join_sort_merge.cpp
```diff
@@ -226,6 +259,179 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+        if (check_right_run) {
+          right_range.for_every_row_id(_sorted_right_table, [&](RowID right_row_id) {
+            if (multi_predicate_join_evaluator.satisfies_all_predicates(left_row_id, right_row_id)) {
+              match = true;
+              return;
+            }
+          });
+        }
+      }
+      if (!match) _emit_left_only(output_cluster, left_row_id);
```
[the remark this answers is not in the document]
Alexander-Dubrawski: In the first if-case the status of `match` can still change. Before we emit the left row we need to check that match. I am not really sure how to merge that.

src/lib/operators/join_sort_merge.cpp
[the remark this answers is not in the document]
Alexander-Dubrawski: It was still there, so I removed it.

src/lib/operators/join_sort_merge.cpp
Alexander-Dubrawski: I added a comment:

```
    // In some certain semi and anti join cases, we do not need to execute the sort-merge join algorithm. These can be
    // determined before the merge.
    // The case for the semi-join is:
    // - The right table is empty. We can return an empty left table.
    // The case for the anti null as false join is:
    // - The right table is empty. We can return the left input table.
    // The cases for the anti null as false join are:
    // - The right table is empty. We can return the left input table.
    // - If there is only one join predicate and the right table contains a null value in the join column, we can
    //   return an empty left table.
```

src/lib/operators/join_sort_merge.cpp
```diff
@@ -226,6 +259,179 @@ class JoinSortMerge::JoinSortMergeImpl : public AbstractReadOnlyOperatorImpl {
...
+          return;
+        }
+      });
+      if ((match && _mode == JoinMode::Semi) || (!match && _mode == JoinMode::AntiNullAsFalse)) {
+        _emit_left_only(output_cluster, left_row_id);
+      }
+    });
+  }
+
+  void _emit_multipredicate_anti_null_as_true(size_t output_cluster, TableRange left_range, TableRange right_range,
```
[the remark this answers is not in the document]
Alexander-Dubrawski: How about the following comment:

```
  // In the case of the anti null as true join, we need to make sure that there is no match between the rows of the
  // left range and the NULL rows from the right table where all predicates are satisfied. In some cases, we also need
  // to check that there is no match with the rows from the right range. That is for example the case if the right and
  // left range have the same value because than there still can be a match for the current row where all predicates
  // are satisfied. If that is the case check_right_range should be set.
```

src/lib/operators/join_sort_merge.cpp
Alexander-Dubrawski: It's a typo. It should be right non-NULL.

src/lib/operators/join_sort_merge.cpp
Alexander-Dubrawski: How exactly do you mean that?

src/lib/operators/join_sort_merge.cpp
Alexander-Dubrawski: Does it make now sence?

Bouncner:
Closing for now. I am still on it, but just don't have the time right now. :(

PR #2371, opened 2021-06-03
Fix null handling in CSV import

Alexander-Dubrawski:
This PR fixes the import with the `null string as null` handling.  Before this PR it was possible to define a column as not nullable and the parser would not recognize a `NULL` value and throw an exception instead it would import it as a string. This was only happen in the case that the config was set to `"config": {"null_handling": "null_string_as_null"}`.


src/lib/import_export/csv/csv_converter.hpp
```diff
@@ -59,6 +59,11 @@ class CsvConverter : public BaseCsvConverter {
       Assert(_config.null_handling != NullHandling::RejectNullStrings,
              "Unquoted null found in CSV file. Quote it for string literal \"null\", leave field empty for null "
              "value, or set 'null_handling' to the appropriate strategy in parse config.");
+      if (_config.null_handling == NullHandling::NullStringAsNull) {
+        Assert(_is_nullable,
+               "Unquoted null found in CSV file, while the associated column is not nullable. Quote it for string "
+               "literal \"null\".");
```
Bensk1, line 65: The suggestion could work. But if NULL occurs in a non-nullable column the user might also consider making the column nullable?

src/lib/import_export/csv/csv_converter.hpp
```diff
@@ -59,6 +59,11 @@ class CsvConverter : public BaseCsvConverter {
...
+      }
```
Bensk1, line 66: Could we merge that with below's `if (_is_nullable && _config.null_handling == NullHandling::NullStringAsNull) {`? Isn't `if (_config.null_handling == NullHandling::NullStringAsNull) {` and `Assert(_is_nullable` effectively the same? If that is correct, it might even communicate the intent clearer:
- if we encountered a NULL_STRING
- and if we should treat a NullStringAsNull
- then we assert that the column really `_is_nullable`
- if all of that is the case, we can set `_null_values` to `true` at `position`

or am I wrong?

Bensk1:
Merging the current master might fix the CI's problem. fc2a056 fixes the JOB's test table definitions.

src/lib/import_export/csv/csv_converter.hpp
Alexander-Dubrawski: Hm, that's true. How about: ` Quote it for string literal \"null\" or make the column nullable.`?

src/lib/import_export/csv/csv_converter.hpp
Alexander-Dubrawski: Good point 🙃 . I merged the two cases into one.

src/lib/import_export/csv/csv_converter.hpp
Bensk1: Works for me.

Bensk1: I would still keep the first sentence

Bensk1 (changes requested):
See comment regarding assert comment

Bensk1 (approved):

PR #2377, opened 2021-06-25
Update Google Benchmark

Bouncner:
Used version of Google benchmark does not build with GCC 11 (see here: https://github.com/google/benchmark/commit/3d1c2677686718d906f28c1d4da001c42666e6d2).

This PR simply updates google bench to the last master commit.

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -27,6 +27,7 @@ add_library(
     benchmark/src/counter.cc
     benchmark/src/csv_reporter.cc
     benchmark/src/json_reporter.cc
+    benchmark/src/perf_counters.cc
     benchmark/src/reporter.cc
     benchmark/src/sleep.cc
     benchmark/src/statistics.cc
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -8,46 +8,6 @@ endif()
 # unity builds for this folder, even if the user manually activated it.
 set(CMAKE_UNITY_BUILD "OFF")
 
-# Build google benchmark
-add_library(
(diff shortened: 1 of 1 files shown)


Bouncner:
Hmm, cannot reproduce the linking errors on Mac or linux. Any help is welcome.

Bensk1:
I can reproduce that. 309307e9b should do the trick, not sure whether further files are missing from `third_party/CMakeLists.txt`.

Bensk1:
If we don't have a comment that updates to `third_party/CMakeLists.txt` might be necessary when upgrading submodules, we should add one.

Bouncner:
> If we don't have a comment that updates to third_party/CMakeLists.txt might be necessary when upgrading submodules, we should add one.

Good point, but I am not sure where to place such a comment. The file does not have a "header" like most hpp files where I would assume everybody is reading it.
And I hope most people aren't that stupid to not include a required file. :pensive:

Bouncner:
@Bensk1 : any idea where to place the comment?

Bensk1:
> @Bensk1 : any idea where to place the comment?

No, I thought about it but couldn't come up with any sensible location for such a comment.

Bensk1 (approved):

Bouncner:
Now using `add_subdirectory`. Thanks for bringing it up.

One thing about it is that now, the CMake output of the benchmark are written to stdout as well. Does anybody know how to get rid of the non-error output?

Bensk1:
> Does anybody know how to get rid of the non-error output?

No, and if stackoverflow does not suggest anything better than overwriting `message()`, it is likely that nothing else exist :)

PR #2380, opened 2021-06-29
lint.sh: Work even if only a single file is present

mrks:
I was just using this as a base for something else and realized that our custom lint scripts expect at least two cpp files. Adding `-H` makes grep print the file name even if there is just a single file.

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -15,35 +15,35 @@ let "exitcode |= $?"
 #             Finds all .cpp and .hpp files, separated by \0
 
 # All disabled tests should have an issue number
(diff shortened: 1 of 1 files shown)


Bouncner:
> I was just using this as a base for something else [..]

Stop doing that. Work on Hyrise.

Bouncner:
I wanted to check what exactly happens without the `H` but could not really reproduce the problem on macOS or linux.
I created a new directory with only a single cpp file. The output remains the same. Did I misunderstand your problem?

mrks:
```
mkdir src
echo '#include "foo.cpp"' > src/bad.cpp
./lint.sh
```

On nemea, this gives me

```
1:#include "foo.cpp" 1:#include "foo.cpp"
```

Bouncner (approved):

PR #2382, opened 2021-07-01
Fix: isnan is undefined

Bouncner:
Building Hyrise with GCC 10 fails as `isnan` is not defined. It's either std::isnan of `<cmath>` or it's `isnan`, if we include `<math.h>` (which we do not want).

I do not know why this hasn't been catched before.


Bensk1:
Gcc-10 should be running on Nemea for a while already. On which platform did this occur?

Bouncner:
Good question. It was an Debian with an AMD CPU. But that shouldn't change anything I guess.

I'll have another look at it.

Bouncner:
I can reproduce it on the Debian system and only with make. Not with ninja. Wtf.

Should still be a bug though.

Bensk1 (approved):

PR #2394, opened 2021-08-01
[DYOD] Shared Dictionaries Plugin

hig-dev:
This PR adds a new plugin called **SharedDictionariesPlugin** to Hyrise. 

## Motivation
The intention of this plugin is to save memory by using "dictionary sharing" while trying to not decrease the performance of Hyrise.

Reasons:
- Per default every dictionary encoded segment in Hyrise has its own dictionary
- Real-world data contains patterns and repetitions
- With one dictionary per segment, many dictionaries might overlap to a certain degree
 
## Overview
This plugin compares the dictionaries within a column for their similarity using the jaccard-index. 
 If the jaccard-index is equal or higher than the specified threshold and additionally if the merging does not increase the width of the attribute vector, this plugin creates a shared dictionary.  
Then, the plugin replaces the dictionary segment with a new dictionary segment that has a shared dictionary.

With the start of the plugin, the dictionary sharing compressor is automatically started for every table in the database.

## Considerations
- The default _jaccard_index_threshold_ is set to 0.4, an other default value may be better
- The algorithm prevents per default the sharing of dictionaries if it would increase the attribute vector width. Benchmark results have shown that this check is important to prevent performance penalties.
- The algorithm only compares dictionaries column-wise, it may be beneficial to also compare across columns and tables.
- The used similarity metric is the Jaccard-Index, an other metric may be more suitable 
- Certain operators (e.g. "LIKE") are performing worse due to pre evalutations of the dictionary segment. For instance, it is more likely for a chunk to not be skipped if it contains a shared dictionary.
- The PR adds a new property _uses_dictionary_sharing_ to the _DictionarySegment_ class. It would be better if the plugin does not change code in /lib. However, this change was made to support multiple executions of the plugin even if the plugin is unloaded and loaded again.
- The memory estimation in DictionarySegments that have a shared dictionary is misleading 

## Impact on the memory usage
The following charts show the memory savings in the IMDB database which is used in the Join-Order Benchmark.

### Absolute
image
### Relative
image

## Impact on the performance

The following chart shows the performance impact on the Join-Order Benchmark which uses the IMDB database.
Full data of the benchmark: https://github.com/hyrise/hyrise/files/6943601/results_nemea.zip

image

- There is a positive impact on the performance with Jaccard-Index=0.5 which is negligible
- There is no negative impact on the performance for Jaccard-Index > 0.3

## Flow chart of the algorithm
image


hig-dev:
We now have the benchmark results that were produced by the VM.

Zip: [benchmark-results-vmappleton.zip](https://github.com/hyrise/hyrise/files/6935230/benchmark-results-vmappleton.zip)


@Bensk1 
Can you run the benchmarks on a dedicated machine so that we can add reliable benchmark results?

Therefore, we created an extra branch, in which you only have to execute the benchmark script:
https://github.com/hig-dev/hyrise/blob/exp/bench_shared_dictionaries_plugin/start_shared_dictionaries_bench.sh

Bensk1:
> @Bensk1
> Can you run the benchmarks on a dedicated machine so that we can add reliable benchmark results?

Benchmarks are running. I will send you the results as soon as the benchmarks finish. I added warm up time `-w 5` to the benchmark executions.

Bensk1:
Here are the results from one of our servers: [results_nemea.zip](https://github.com/hyrise/hyrise/files/6943601/results_nemea.zip)

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin_settings.hpp
```diff
@@ -0,0 +1,32 @@
...
+ */
+class JaccardIndexThresholdSetting : public AbstractSetting {
+ public:
+  JaccardIndexThresholdSetting() : AbstractSetting("Plugin::SharedDictionaries.jaccard_index_threshold") {}
+  const std::string& description() const;
+  const std::string& display_name() const;
+  const std::string& get();
+  void set(const std::string& value);
+
+  std::string _value = "0.1";
```
PeterTsayun, line 28: In the PR description, the default value of the jaccard index is set to 0.4. Is the default value of 0.1 intentional?

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin_settings.cpp
```diff
@@ -0,0 +1,13 @@
...
+
+namespace opossum {
+
+const std::string& JaccardIndexThresholdSetting::description() const {
+  static const auto description = std::string{"Threshold for the similarity metric between dictionaries for merging"};
+  return description;
+}
+const std::string& JaccardIndexThresholdSetting::display_name() const { return _display_name; }
+const std::string& JaccardIndexThresholdSetting::get() { return _value; }
+void JaccardIndexThresholdSetting::set(const std::string& value) { _value = value; }
```
PeterTsayun, line 11: Maybe, the setter could check whether the new value falls into the pre-defined range of [0.0, 1.0].

src/test/plugins/shared_dictionaries_plugin_test.cpp
```diff
@@ -0,0 +1,265 @@
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "base_test.hpp"
+#include "lib/utils/plugin_test_utils.hpp"
+
+#include "../../plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp"
+#include "../../plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.hpp"
```
PeterTsayun, line 9: Are relative imports indeed needed or would 
```
#include "plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp"
#include "plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.hpp"
```
already suffice?

src/test/plugins/shared_dictionaries_plugin_test.cpp
```diff
@@ -0,0 +1,265 @@
...
+  auto& pm = Hyrise::get().plugin_manager;
+
+  pm.load_plugin(build_dylib_path("libhyriseSharedDictionariesPlugin"));
+  _validate_all_test_tables();
+  pm.unload_plugin("hyriseSharedDictionariesPlugin");
+
+  pm.load_plugin(build_dylib_path("libhyriseSharedDictionariesPlugin"));
+  _validate_all_test_tables();
+  pm.unload_plugin("hyriseSharedDictionariesPlugin");
+}
```
PeterTsayun, line 136: Since `LoadUnloadPlugin` is already a part of the `ReloadPlugin` test, consider removing the`LoadUnloadPlugin` test

PeterTsayun (commented):
Great work, well done :+1:

Below are some minor suggestions. This is an initial review, it may be updated until August 15th.

P. S. Don't forget to mark your PR with the `FullCI` label

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,86 @@
...
+   * Merge plan for a shared dictionary containing the segments to merge and additional information
+   */
+  struct MergePlan {
+   public:
+    std::shared_ptr<const pmr_vector<T>> shared_dictionary;
+    std::vector<SegmentChunkPair<T>> segment_chunk_pairs_to_merge;
+    bool contains_non_merged_segment;
+    bool contains_already_merged_segment;
+    uint64_t non_merged_total_bytes = 0ul;
+    uint64_t non_merged_dictionary_bytes = 0ul;
```
manner, line 41: Specifying `ul` when the data type is clear seems redundant.

```suggestion
    uint64_t non_merged_total_bytes = 0;
    uint64_t non_merged_dictionary_bytes = 0;
```

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.hpp
```diff
@@ -0,0 +1,70 @@
...
+ * started for every table in the database.
+ */
+class SharedDictionariesPlugin : public AbstractPlugin {
+ public:
+  inline static const std::string LOG_NAME = "SharedDictionariesPlugin";
+
+  struct SharedDictionariesStats {
+    uint64_t total_bytes_saved = 0ul;
+    uint64_t total_previous_bytes = 0ul;
+    uint64_t modified_previous_bytes = 0u;
```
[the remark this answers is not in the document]
manner: I would argue all of those specifiers (L.37-43) are redundant, as the data type is clearly defined.

src/test/plugins/shared_dictionaries_plugin_test.cpp
```diff
@@ -0,0 +1,265 @@
...
+                            const std::vector<SegmentChunkPair<pmr_string>>& shared_segment_chunk_pairs) {
+    return column_processor._should_merge(jaccard_index, current_dictionary_size, shared_dictionary_size,
+                                          shared_segment_chunk_pairs);
+  }
+
+  static void _add_segment_chunk_pair(SharedDictionariesColumnProcessor<pmr_string>::MergePlan& merge_plan,
+                                      const SegmentChunkPair<pmr_string>& segment_chunk_pair, bool is_already_merged) {
+    SharedDictionariesColumnProcessor<pmr_string>::_add_segment_chunk_pair(merge_plan, segment_chunk_pair,
+                                                                           is_already_merged);
+  }
```
manner, line 108: What's the reasoning behind creating those four functions? 
Instead of directly calling `column_processor._initialize_merge_plans(merge_plans)` in the test, you now call `_initialize_merge_plans(column_processor, merge_plans)`
Does this give you any benefits?

Bensk1:
> * The PR adds a new property _uses_dictionary_sharing_ to the _DictionarySegment_ class. It would be better if the plugin does not change code in /lib. However, this change was made to support multiple executions of the plugin even if the plugin is unloaded and loaded again.

As discussed before, if the plugin is unloaded and loaded again, it does not have to deal with such cases. 

> * The memory estimation in DictionarySegments that have a shared dictionary is misleading

Could you elaborate on that?


Did you use the new benchmarks I provided to update the figures?

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
```diff
@@ -0,0 +1,291 @@
...
+                            return _increases_attribute_vector_width(shared_dictionary_size,
+                                                                     segment_chunk_pair.first->dictionary()->size());
+                          });
+    }
+  }
+  return false;
+}
+
+template <typename T>
+void SharedDictionariesColumnProcessor<T>::_add_segment_chunk_pair(MergePlan& merge_plan,
```
Bensk1, line 253: Am I wrong or does that seem like a method that could be implemented  (or is even better suited) as part of the `MergePlan` class itself.

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,86 @@
+#pragma once
+
+#include "hyrise.hpp"
+#include "shared_dictionaries_plugin.hpp"
+#include "storage/table.hpp"
+#include "utils/size_estimation_utils.hpp"
+
+namespace opossum {
+
+/**
```
Bensk1, line 10: I could imagine some higher level documentation of the process to be helpful here.

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
```diff
@@ -0,0 +1,291 @@
...
+  auto merge_plans = std::vector<std::shared_ptr<MergePlan>>{};
+  _initialize_merge_plans(merge_plans);
+
+  std::optional<SegmentChunkPair<T>> previous_segment_chunk_pair_opt = std::nullopt;
+
+  const auto chunk_count = table->chunk_count();
+  for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+    const auto chunk = table->get_chunk(chunk_id);
+    const auto segment = chunk->get_segment(column_id);
+    stats->total_previous_bytes += segment->memory_usage(MemoryUsageCalculationMode::Full);
```
Bensk1, line 29: Have you checked the difference in terms of accuracy and runtime when changing the mode from `Full` to `Sampled`?

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
```diff
@@ -0,0 +1,291 @@
...
+    merge_plans.emplace_back(merge_plan);
+  }
+}
+
+template <typename T>
+std::pair<int32_t, std::shared_ptr<const pmr_vector<T>>>
+SharedDictionariesColumnProcessor<T>::_compare_with_existing_merge_plans(
+    const std::shared_ptr<const pmr_vector<T>> current_dictionary,
+    const std::vector<std::shared_ptr<MergePlan>>& merge_plans, const PolymorphicAllocator<T>& allocator) {
+  auto best_merge_plan_index = -1;
```
Bensk1, line 101: Could we communicate the intent clearer by using and optional here instead of passing `-1` to encode no best merge plan?

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
```diff
@@ -0,0 +1,291 @@
...
+        best_shared_dictionary = union_result;
+      }
+    }
+  }
+
+  return std::make_pair(best_merge_plan_index, best_shared_dictionary);
+}
+
+template <typename T>
+std::shared_ptr<const pmr_vector<T>> SharedDictionariesColumnProcessor<T>::_compare_with_previous_dictionary(
```
Bensk1, line 130: That does not only compare with the previous dictionary but also actually do the merge, doesn't it?

Bensk1 (changes requested):
Just some first comments. I think a high level overview of your algorithm (not super high level stating that we are using jaccard indexes) would be really helpful.

hig-dev:
> As discussed before, if the plugin is unloaded and loaded again, it does not have to deal with such cases.

One of the requirements resulting from the last meeting, was that our plugin should be able to reprocess the database. We thought about including a state to the plugin, so that the plugin can remember if a dictionary segment uses a shared dictionary.  We came to the conclusion that a state inside the plugin is problematic as it increases the complexity and memory usage. Furthermore, over time the internal state could be outdated. That's why we modified the dictionary segment class.

Shall we change this back and remove the requirement for multiple executions?

> 
> > * The memory estimation in DictionarySegments that have a shared dictionary is misleading
> 
> Could you elaborate on that?
> 

Since a shared dictionary is used by multiple segments it will also be taken into account for each individual memory estimation of the affected segments (thus the memory estimation of multiple segments which share a dictionary together will take the dictionary memory usage into account multiple times)


> Did you use the new benchmarks I provided to update the figures?

Yes, the figures use the latest benchmarks from you (with warmup).

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
hig-dev: In our tests we had an assertion error because of this check, if the mode was set to `Sampled`:
```
Assert(new_dictionary_memory_usage < previous_dictionary_memory_usage,
             "New dictionary memory usage should be lower than previous");
```

That's why we switched to use `Full`

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
```diff
@@ -0,0 +1,291 @@
...
+}
+
+template <typename T>
+void SharedDictionariesColumnProcessor<T>::_add_segment_chunk_pair(MergePlan& merge_plan,
+                                                                   const SegmentChunkPair<T>& segment_chunk_pair,
+                                                                   bool is_already_merged) {
+  if (is_already_merged) {
+    merge_plan.contains_already_merged_segment = true;
+  }
+  if (!is_already_merged) {
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,86 @@
...
+  const std::shared_ptr<Table> table;
+  const std::string& table_name;
+  const ColumnID column_id;
+  const std::string& column_name;
+  const double jaccard_index_threshold;
+  std::shared_ptr<SharedDictionariesPlugin::SharedDictionariesStats> stats;
+  SharedDictionariesColumnProcessor(
+      const std::shared_ptr<Table>& init_table, const std::string& init_table_name, const ColumnID init_column_id,
+      const std::string& init_column_name, const double init_jaccard_index_threshold,
+      const std::shared_ptr<SharedDictionariesPlugin::SharedDictionariesStats>& init_stats);
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,86 @@
...
+    bool contains_already_merged_segment;
+    uint64_t non_merged_total_bytes = 0ul;
+    uint64_t non_merged_dictionary_bytes = 0ul;
+
+    explicit MergePlan(const std::shared_ptr<const pmr_vector<T>>& init_shared_dictionary)
+        : shared_dictionary(init_shared_dictionary) {
+      segment_chunk_pairs_to_merge = {};
+      contains_non_merged_segment = false;
+      contains_already_merged_segment = false;
+    }
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,86 @@
...
+  static size_t _calc_dictionary_memory_usage(const std::shared_ptr<const pmr_vector<T>> dictionary);
+
+  bool _should_merge(const double jaccard_index, const size_t current_dictionary_size,
+                     const size_t shared_dictionary_size,
+                     const std::vector<SegmentChunkPair<T>>& shared_segment_chunk_pairs);
+
+  static void _add_segment_chunk_pair(MergePlan& merge_plan, const SegmentChunkPair<T>& segment_chunk_pair,
+                                      bool is_already_merged);
+
+  static double _calc_jaccard_index(size_t union_size, size_t intersection_size);
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,86 @@
...
+      const std::shared_ptr<Table>& init_table, const std::string& init_table_name, const ColumnID init_column_id,
+      const std::string& init_column_name, const double init_jaccard_index_threshold,
+      const std::shared_ptr<SharedDictionariesPlugin::SharedDictionariesStats>& init_stats);
+
+  void process();
+
+  /**
+   * Merge plan for a shared dictionary containing the segments to merge and additional information
+   */
+  struct MergePlan {
```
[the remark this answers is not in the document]
hig-dev: We decided to keep `MergePlan` there and we improved the constructor of it.

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.hpp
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.hpp
```diff
@@ -0,0 +1,70 @@
...
+    uint32_t num_merged_dictionaries = 0u;
+    uint32_t num_shared_dictionaries = 0u;
+    uint32_t num_existing_merged_dictionaries = 0u;
+    uint32_t num_existing_shared_dictionaries = 0u;
+  };
+
+  SharedDictionariesPlugin() : storage_manager(Hyrise::get().storage_manager), log_manager(Hyrise::get().log_manager) {
+    stats = std::make_shared<SharedDictionariesStats>();
+  }
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.hpp
```diff
@@ -0,0 +1,70 @@
...
+  std::string description() const final;
+
+  void start() final;
+
+  void stop() final;
+
+  StorageManager& storage_manager;
+  LogManager& log_manager;
+
+  std::shared_ptr<SharedDictionariesStats> stats;
```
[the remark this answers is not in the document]
hig-dev: We made `storage_manager` and `log_manager` private and kept `stats` public

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+ private:
+  std::shared_ptr<JaccardIndexThresholdSetting> _jaccard_index_threshold_setting;
+
+  void _process_for_every_column();
+
+  void _log_plugin_configuration();
+  void _log_processing_result();
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "shared_dictionaries_plugin_settings.hpp"
+
+namespace opossum {
+
+std::string SharedDictionariesPlugin::description() const { return "Shared dictionaries plugin"; }
+
+void SharedDictionariesPlugin::start() {
+  _jaccard_index_threshold_setting = std::make_shared<JaccardIndexThresholdSetting>();
+  _jaccard_index_threshold_setting->register_at_settings_manager();
+  stats = std::make_shared<SharedDictionariesStats>();
```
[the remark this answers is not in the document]
hig-dev: We moved the initialization of `_jaccard_index_threshold_setting` to the constructor.
However, the problem still exists as the start method will be called automatically after the plugin loads.

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+  _log_plugin_configuration();
+  _process_for_every_column();
+  _log_processing_result();
+}
+
+void SharedDictionariesPlugin::stop() { _jaccard_index_threshold_setting->unregister_at_settings_manager(); }
+
+void SharedDictionariesPlugin::_process_for_every_column() {
+  log_manager.add_message(LOG_NAME, "Starting database compression", LogLevel::Info);
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.cpp
```diff
@@ -0,0 +1,85 @@
...
+  const auto jaccard_index_threshold = std::stod(_jaccard_index_threshold_setting->get());
+
+  auto table_names = storage_manager.table_names();
+  std::sort(table_names.begin(), table_names.end());
+  for (const auto& table_name : table_names) {
+    {
+      const auto log_message = "> compressing table: " + table_name;
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.cpp
```diff
@@ -0,0 +1,85 @@
...
+      log_manager.add_message(LOG_NAME, log_message, LogLevel::Debug);
+    }
+    const auto table = storage_manager.get_table(table_name);
+    const auto column_count = table->column_count();
+
+    for (auto column_id = ColumnID{0}; column_id < column_count; ++column_id) {
+      const auto column_data_type = table->column_definitions()[column_id].data_type;
+      const auto column_name = table->column_definitions()[column_id].name;
+      {
+        const auto log_message = "  - compressing column: " + column_name;
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.cpp
```diff
@@ -0,0 +1,85 @@
...
+      resolve_data_type(column_data_type, [&](const auto type) {
+        using ColumnDataType = typename decltype(type)::type;
+        auto column_processor = SharedDictionariesColumnProcessor<ColumnDataType>{
+            table, table_name, column_id, column_name, jaccard_index_threshold, stats};
+        column_processor.process();
+      });
+    }
+  }
+
+  log_manager.add_message(LOG_NAME, "Completed database compression", LogLevel::Info);
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin_settings.hpp
```diff
@@ -0,0 +1,32 @@
...
+ *   - option to select all tables / include tables / exclude tables
+ *     The plugin currently tries to merge dictionary segments of all available tables.
+ */
+
+/**
+ * Threshold for the similarity metric between dictionaries for merging
+ */
+class JaccardIndexThresholdSetting : public AbstractSetting {
+ public:
+  JaccardIndexThresholdSetting() : AbstractSetting("Plugin::SharedDictionaries.jaccard_index_threshold") {}
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin_settings.hpp
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin_settings.cpp
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/test/plugins/shared_dictionaries_plugin_test.cpp
hig-dev: We removed the unused import, however the import does not work without the relative import.

src/test/plugins/shared_dictionaries_plugin_test.cpp
hig-dev: We decided to keep both tests as it is possible for one test to fail and the other not.
So if `LoadUnloadPlugin` is successful and `ReloadPlugin` is not, the user sees at a glance that the error is on the reload.

src/lib/storage/dictionary_segment.hpp
```diff
@@ -19,8 +19,14 @@ class BaseCompressedVector;
 template <typename T>
 class DictionarySegment : public BaseDictionarySegment {
  public:
+  /// Creates a dictionary encoded segment
+  /// \param dictionary The dictionary
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,86 @@
...
+   public:
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
```diff
@@ -0,0 +1,291 @@
...
+                        merge_plan->segment_chunk_pairs_to_merge)) {
+        best_merge_plan_index = static_cast<int32_t>(merge_plan_index);
+        best_jaccard_index = jaccard_index;
+        union_result->shrink_to_fit();
+        best_shared_dictionary = union_result;
+      }
+    }
+  }
+
+  return std::make_pair(best_merge_plan_index, best_shared_dictionary);
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "hyrise.hpp"
+#include "shared_dictionaries_plugin.hpp"
+#include "storage/table.hpp"
+#include "utils/size_estimation_utils.hpp"
+
+namespace opossum {
+
+/**
+ * Processes one column for the SharedDictionariesPlugin
```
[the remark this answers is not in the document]
hig-dev: We added further documentation. Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
hig-dev: We changed the name to `_union_with... `

Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/7f88317d0867f9bfd92688f5061daea73a39be35

src/test/plugins/shared_dictionaries_plugin_test.cpp
hig-dev: They are private functions and in order to test them, we have to use a `friend class`.
But the friend class does not work inside the TEST_F test fixtures. That's why we have the methods inside the test class.

Bensk1:
Thank you for your improvements, I will have another look as soon as possible!

Bensk1:
> Shall we change this back and remove the requirement for multiple executions?

No, it's fine like that for now.

> Since a shared dictionary is used by multiple segments it will also be taken into account for each individual memory estimation of the affected segments (thus the memory estimation of multiple segments which share a dictionary together will take the dictionary memory usage into account multiple times)

That's true. One (imperfect) way to handle that would be to divide the size of the dictionary by the number of segments that share it. Are you correcting for that inaccuracy in your size calculations?

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+  /*
+   * The following settings are currently hard coded for the plugin but could be made configurable:
+   *   - option to disable auto start of compression
+   *     The plugin currently starts the merging of dictionaries automatically once it is loaded.
+   *   - option to disable tracking of compression statistics
+   *     The plugin currently tracks memory usage statistics for merged data.
+   *   - option to select all tables / include tables / exclude tables
+   *     The plugin currently tries to merge dictionary segments of all available tables.
+   */
```
Bensk1, line 68: I think we can remove that. Also, disabling the automatic start of compression should be difficult because it would need to happen before the Plugin is loaded when the settings do not exist yet, shouldn't it?

ShirleyNekoDev:
> > Since a shared dictionary is used by multiple segments it will also be taken into account for each individual memory estimation of the affected segments (thus the memory estimation of multiple segments which share a dictionary together will take the dictionary memory usage into account multiple times)
> 
> That's true. One (imperfect) way to handle that would be to divide the size of the dictionary by the number of segments that share it. Are you correcting for that inaccuracy in your size calculations?

Yes, we consider the memory usage of shared dictionaries while scanning - see:
https://github.com/hyrise/hyrise/blob/56ff295e85e15ba3df3ce532611f69441443503f/src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp#L161-L187

Your proposal to divide by the number of shares of a dictionary could be used for size estimation, but would require to know this count. We could implement this by e.g. replacing the `bool  _uses_dictionary_sharing` in `dictionary_segment.hpp` with a `byte _dictionary_shareholder_count` (default = 0, instead of _uses_dictionary_sharing = false), while keeping the same object size with the only drawback of only allowing up to 255 shares (or more when using uint16/uint32). What is your opinion on that?

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -8,54 +8,56 @@
 namespace opossum {
 
 /**
- * Processes one column for the SharedDictionariesPlugin
+ * Finds and merges dictionaries with shared data for the given column of a table.
+ * Each dictionary segment is compared based on its similarity with all other existing shared dictionaries.
+ * If the similarity (e.g. jaccard-index) is above the threshold it merges the dictionaries.
+ * If none of the existing shared dictionaries meet the threshold it compares the current dictionary with the
+ * previous dictionary and eventually merges them to create a new shared dictionary.
```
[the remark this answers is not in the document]
hig-dev: 😊

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.cpp
```diff
@@ -0,0 +1,291 @@
...
+  auto best_merge_plan_index = -1;
+  auto best_jaccard_index = -1.0;
+  std::shared_ptr<const pmr_vector<T>> best_shared_dictionary = nullptr;
+  const auto merge_plans_size = merge_plans.size();
+  for (auto merge_plan_index = 0ul; merge_plan_index < merge_plans_size; ++merge_plan_index) {
+    const auto merge_plan = merge_plans[merge_plan_index];
+    const auto shared_dictionary = merge_plan->shared_dictionary;
+    auto union_result = std::make_shared<pmr_vector<T>>(allocator);
+    union_result->reserve(std::max(current_dictionary->size(), shared_dictionary->size()));
+    std::set_union(current_dictionary->cbegin(), current_dictionary->cend(), shared_dictionary->cbegin(),
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/8ab006a666ed496b67ec0c88e833a1c96a269a3a

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,85 @@
...
+ */
+template <typename T>
+class SharedDictionariesColumnProcessor {
+  friend class SharedDictionariesPluginTest;
+
+ public:
+  /**
+ * Merge plan for a shared dictionary containing the segments to merge and additional information
+ */
+  struct MergePlan {
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/8ab006a666ed496b67ec0c88e833a1c96a269a3a

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,85 @@
...
+        : shared_dictionary(init_shared_dictionary) {}
+
+    void add_segment_chunk_pair(const SegmentChunkPair<T>& segment_chunk_pair, bool is_already_merged);
+  };
+
+  const std::shared_ptr<Table> table;
+  const std::string table_name;
+  const ColumnID column_id;
+  const std::string column_name;
+  const double jaccard_index_threshold;
```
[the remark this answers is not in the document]
hig-dev: We changed it to float. Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/8ab006a666ed496b67ec0c88e833a1c96a269a3a

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.hpp
```diff
@@ -0,0 +1,78 @@
...
+#include "utils/abstract_plugin.hpp"
+#include "utils/singleton.hpp"
+
+namespace opossum {
+
+template <typename T>
+using SegmentChunkPair = std::pair<std::shared_ptr<DictionarySegment<T>>, std::shared_ptr<Chunk>>;
+
+/**
+ * The intention of this plugin is to save memory by using "dictionary sharing"
```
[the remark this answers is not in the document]
hig-dev: Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/8ab006a666ed496b67ec0c88e833a1c96a269a3a

src/plugins/shared_dictionaries_plugin/shared_dictionaries_plugin.hpp
hig-dev: We removed it. Fixed in https://github.com/hyrise/hyrise/pull/2394/commits/8ab006a666ed496b67ec0c88e833a1c96a269a3a

Bensk1:
Is there a specific reason for keeping the object size? I think, its fine to spend 2 or even 4 bytes on that data.

src/test/plugins/shared_dictionaries_plugin_test.cpp
```diff
@@ -0,0 +1,258 @@
...
+      _get_customer_table(), "customer", _market_column_id, _market_column_name, 0.1f, stats};
+  column_processor.process();
+
+  EXPECT_GT(stats.total_bytes_saved, 0);
+  EXPECT_GT(stats.total_previous_bytes, 0);
+  EXPECT_GT(stats.modified_previous_bytes, 0);
+  EXPECT_GT(stats.num_merged_dictionaries, 0);
+  EXPECT_GT(stats.num_shared_dictionaries, 0);
+  EXPECT_EQ(stats.num_existing_shared_dictionaries, 0);
+  EXPECT_EQ(stats.num_existing_merged_dictionaries, 0);
```
Bensk1, line 144: Should/Could we be more precise here than `> 0`?

src/plugins/shared_dictionaries_plugin/shared_dictionaries_column_processor.hpp
```diff
@@ -0,0 +1,89 @@
...
+  const std::string table_name;
+  const ColumnID column_id;
+  const std::string column_name;
+  const float jaccard_index_threshold;
+  SharedDictionariesPlugin::SharedDictionariesStats& stats;
+
+  SharedDictionariesColumnProcessor(const std::shared_ptr<Table>& init_table, const std::string& init_table_name,
+                                    const ColumnID init_column_id, const std::string& init_column_name,
+                                    const float init_jaccard_index_threshold,
+                                    SharedDictionariesPlugin::SharedDictionariesStats& init_stats);
```
Bensk1, line 50: If the Plugin is unloaded while the processor is working, the accesses to stats could access memory that was already freed, right?

Bensk1:
The only failing tests are some size mismatches under macOS probably caused by the fact that we use libc++ on macOS.

Bensk1:
Currently, we do not have the resources to work on the remaining open issues. Thus, I am going to close this PR for now.

PR #2397, opened 2021-08-02
[DYOD] Cardinality estimation: cardinality analysis tool

leowe:
This PR is part of the DYOD seminar and aims to improve the cardinality estimation for processing SQL queries in hyrise and delivers a tool to do more precise analyses of such estimations.

This was achieved by

**(1) Implementing a new cardinality estimation class - a Top K histogram modeled as a generic histogram (`TopKAsGenericHistogram`) 
(2) Fixing a bug in the selectivity calculation negatively impacting the cardinality estimation 
(3) Implementing a cardinality estimation analyses tool**

This is one of 3 PRs that each implement one described part.

## (3) Cardinality Estimation Tool
To further improve the future analysis of cardinality estimations, we implemented a tool to calculate the differences between estimations and actual cardinalities per operation in a query plan.

For this we implemented:

1. a `cardinality_writer` that hooks into the query plan visualization workflow and produces a `benchmark_cardinality_estimation.csv` with cardinality information for each operation in each item of a benchmark.  As the evaluation depends on the assumption that lqp and pqp will usually have the same order of nodes, additional debug information is printed to identify and manually correct cases where this is not the case.
2. a shell script to run all benchmarks with a cardinality estimation analysis and store the created files in a fitting folder structure for later analysis

Usage:

1. Printing cardinality estimation is currently automatically enabled when setting the `--visualize` flag on a benchmark run. (The tool is intended to be stored on an extra branch and accessed when needed.) It produces the following output:

* A `benchmark_cardinality_estimation.csv` with the following structure:
  `Benchmark Item ID ; LQP Node ; PQP Node ; LQP Row Count ; PQP Row Count`
  (The csv uses semicolons as seperators to be able to work with database values including commas.)

image

The csv can then be used for further analysis e.g. analysing the distribution of differences between estimation and actual cardinality.
image

image

* a `benchmark_cardinality_estimation_debug_info.csv` with the following structure:
  `Benchmark Item ID ; Number of LQP Nodes; Number of PQP Nodes`

1. The `cardinality_estimations_all.sh` shell script can be run from the root folder via:
   `./scripts/cardinality_estimations_all.sh  ./path/to/build_folder ./path/to/cardinality_information_folder`
   It runs the following benchmarks: `hyriseBenchmarkTPCH hyriseBenchmarkTPCDS hyriseBenchmarkTPCC hyriseBenchmarkJoinOrder`
   The produced folder output structure is:

```shell
├── <cardinality_information_folder>
│   ├──  hyriseBenchmarkJoinOrder                    # Test files (alternatively `spec` or `tests`)
│   │   ├── hyriseBenchmarkJoinOrder_benchmark_cardinality_estimation.csv
│   │   ├── hyriseBenchmarkJoinOrder_benchmark_cardinality_estimation_debug_info.csv
│   │   └── visualized_plans
│   ├── hyriseBenchmarkTPCC
...
```

The script currently works on the assumption that when we traverse the LQP and PQP the same way, we will get the same nodes. All nodes that don't match up are ignored. This works well for the majority of tested LQPs and PQPs in the JoinOrderBenchmark. Future work might be to find ways to automatically work with LQPs and PQPs that don't have the same number of nodes, e.g. by pattern matching the description of nodes or programmatically building a reference table between LQP and PQP nodes, when transforming a LQP to a PQP.


regnujAx (commented):
Code looks fine for us, we have only minor suggestions.

src/lib/visualization/cardinality_writer.cpp
```diff
@@ -0,0 +1,218 @@
...
+                 << ";"
+                 << "LQP Number of nodes"
+                 << ";"
+                 << "PQP Number of nodes"
+                 << "\n";
+  }
+  csv_file.close();
+
+  for (auto node = 0u; node < number_nodes; node++) {
+    // Get lqp and pqp node descriptions (for first two columns)
```
Bensk1, line 85: Do we need that comment? Also I don't understand the "for first two columns" part.

Bensk1: Also, if the description is only used once, we don't necessarily have to store it.

src/lib/visualization/cardinality_writer.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+    auto pqp_row_count = NAN;
+    const auto& performance_data = *_pqp_nodes[node]->performance_data;
+
+    if (_pqp_nodes[node]->executed() && performance_data.has_output) {
+      pqp_row_count = static_cast<float>(performance_data.output_row_count);
+    }
+
+    // Write benchmark id, node descriptions and cardinalities in CSV
+    // Use semicolon as seperator instead of comma to be able to work with database values containing commas
```
Bensk1, line 100: Repetition

Bensk1: Maybe define a separator constant in the header and add a single comment there?

src/lib/visualization/cardinality_writer.cpp
```diff
@@ -0,0 +1,218 @@
...
+    output_estimations << benchmark_item_id << ";" << lqp_description << ";" << pqp_description << ";" << lqp_row_count
+                       << ";" << pqp_row_count << "\n";
+  }
+
+  // Write debug information about possible different numbers of lqp and pqp nodes
+  output_debug << benchmark_item_id << ";" << lqp_nodes_size << ";" << pqp_nodes_size << "\n";
+
+  // Close file handles
```
Bensk1, line 108: unnecessary

src/lib/visualization/cardinality_writer.cpp
```diff
@@ -0,0 +1,218 @@
...
+  }
+}
+
+void CardinalityWriter::_build_lqp_subtree(const std::shared_ptr<AbstractLQPNode>& node,
+                                           std::unordered_set<std::shared_ptr<const AbstractLQPNode>>& visited_nodes,
+                                           ExpressionUnorderedSet& visited_sub_queries) {
+  if (visited_nodes.find(node) != visited_nodes.end()) return;
+  visited_nodes.insert(node);
+
+  // Add current node to our container
```
Bensk1, line 128: unnecessary

src/lib/visualization/cardinality_writer.cpp
```diff
@@ -0,0 +1,218 @@
...
+      if (!subquery_expression) return ExpressionVisitation::VisitArguments;
+
+      if (!visited_sub_queries.emplace(subquery_expression).second) return ExpressionVisitation::VisitArguments;
+
+      _build_lqp_subtree(subquery_expression->lqp, visited_nodes, visited_sub_queries);
+
+      return ExpressionVisitation::VisitArguments;
+    });
+  }
+}
```
Bensk1, line 154: Is that part obtained from another file, i.e., the same functionality used multiple times?

src/lib/visualization/cardinality_writer.cpp
```diff
@@ -0,0 +1,218 @@
...
+    _build_pqp_subtree(plan, visited_ops);
+  }
+}
+
+void CardinalityWriter::_build_pqp_subtree(const std::shared_ptr<const AbstractOperator>& op,
+                                           std::unordered_set<std::shared_ptr<const AbstractOperator>>& visited_ops) {
+  if (visited_ops.find(op) != visited_ops.end()) return;
+  visited_ops.insert(op);
+
+  // Add current node to our container
```
Bensk1, line 169: see above

Bensk1 (changes requested):
That will be really useful! A first bunch of comments.

src/benchmarklib/benchmark_sql_executor.cpp
```diff
@@ -42,6 +43,9 @@ std::pair<SQLPipelineStatus, std::shared_ptr<const Table>> BenchmarkSQLExecutor:
 
   if (_visualize_prefix) {
     _visualize(pipeline);
+
+    // Print cardinality estimations from lqp and actual cardinalities from pqp to csv for debugging purposes
+    _write_cardinalities(pipeline);
```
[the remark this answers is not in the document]
phoeinx: We need the `_visualize_prefix`, which for most benchmarks holds the current item / query_id. We use this to write out the estimations connected to each query. Reusing the `_visualize_prefix` appeared to be the easiest way of implementation with the least amount of changes in other files. This seemed reasonable as the cardinality estimation tool appeared more like an add-on to store on an extra branch and use when needed (as you suggested). 
We could however pass down a dedicated `cardinality_estimation_prefix`, which would be set similarly to the `_visualize_prefix`.

src/lib/visualization/cardinality_writer.cpp
```diff
@@ -0,0 +1,218 @@
...
+  const auto lqp_nodes_size = _lqp_nodes.size();
+  const auto pqp_nodes_size = _pqp_nodes.size();
+
+  // We assume that when we traverse LQP and PQP the same way, we will get the same nodes.
+  // All nodes that don't match up are ignored. This works well for the majority of tested LQPs and PQPs.
+  const auto number_nodes = std::min(_lqp_nodes.size(), _pqp_nodes.size());
+
+  // Open estimation CSV file
+  // APPEND mode is necessary, since the method will be called multiple times during benchmark execution
+  // and the previous results shouldn't be erased.
```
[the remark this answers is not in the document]
phoeinx: Yes, the method is called multiple times if `--runs` is larger than `1`, setting `--runs` to `1` or postprocessing the data is currently in the responsibility of the user. But additionally, this method is also called for each benchmark item in one benchmarking run and so the append mode is currently necessary. We could write the results for each benchmark item in a separate csv, but that would make analysing the results together harder.

src/lib/visualization/cardinality_writer.cpp
```diff
@@ -0,0 +1,218 @@
...
+  // Open debug CSV file for PQP and LQP node number information, which can be used to identify possible
+  // benchmark items with mismatched nodes and for manual correction of the matching.
+  std::ofstream output_debug;
+  output_debug.open("./benchmark_cardinality_estimation_debug_info.csv", std::ios_base::app);
+
+  // If files are empty, write CSV headers.
+  // Use semicolon as seperator instead of comma to be able to work with database values containing commas
+  std::ifstream csv_file;
+  csv_file.open("./benchmark_cardinality_estimation.csv");
+  if (csv_file.peek() == std::ifstream::traits_type::eof()) {
```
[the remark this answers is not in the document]
phoeinx: No, since we create a new `CardinalityWriter` every time we call the method (see `src/benchmarklib/benchmark_sql_executor.cpp` line 174). We think that with the current architecture of the BenchmarkRunner this would be quite cumbersome to improve, since it would have to be created and held by the topmost execution threat and passed along for each new benchmark item.

src/lib/visualization/cardinality_writer.cpp
phoeinx: Almost. The functionality is similar, but different in a few key places. We obtained it from the `lqp_visualizer.cpp` and `pqp_visualizer.cpp` respectively, but changed e.g. the traversion order for the `pqp` and removed a lot of actions that are unnecessary for only printing out cardinality estimations. The common part / the part we reused is the `lqp` and `pqp`-specific traversion of the lqp and `pqp` trees. To refactor it we probably would have to define a new class and then pass around a lot of lambdas to change the behavior during the traversion, which would perhaps make it harder to understand the code.

src/benchmarklib/benchmark_sql_executor.cpp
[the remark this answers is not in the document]
Bensk1: It's fine for now, thanks for clarification.

src/lib/visualization/cardinality_writer.cpp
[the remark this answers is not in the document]
Bensk1: Yes, we definitely want to write to the same CSV file.

src/lib/visualization/cardinality_writer.cpp
Bensk1: That's fine, you don't have to do that.

Bensk1:
I restarted the CI.

Bensk1:
This is now part of the main repository at https://github.com/hyrise/hyrise/tree/cardinality_estimation_analysis

PR #2399, opened 2021-08-20
Remove unnecessary cast from ExpressionReductionRule

julianmenzler:


src/test/lib/optimizer/strategy/expression_reduction_rule_test.cpp
```diff
@@ -254,10 +252,10 @@ TEST_F(ExpressionReductionRuleTest, RemoveDuplicateAggregate) {
                               stored_table_node));
 
     const auto expected_aliases = std::vector<std::string>{"SUM(a)", "COUNT(a) + 1", "AVG(a) + 2"};
-    const auto expected_lqp = AliasNode::make(expression_vector(sum_(col_a), add_(count_(col_a), 1), add_(div_(cast_(sum_(col_a), DataType::Double), count_(col_a)), 2)), expected_aliases,  // NOLINT
-                                ProjectionNode::make(expression_vector(sum_(col_a), add_(count_(col_a), 1), add_(div_(cast_(sum_(col_a), DataType::Double), count_(col_a)), 2)),             // NOLINT
-                                  ProjectionNode::make(expression_vector(sum_(col_a), count_(col_a), div_(cast_(sum_(col_a), DataType::Double), count_(col_a))),                             // NOLINT
```
Bouncner, line 259: Can you elaborate a bit? For me first casting it to double makes sense when a is an integer. Do we have a test that tests this case?

What do you mean by Q1 gets simplified? The LQP or even the PQP?

julianmenzler: Thanks! Now I think I'm wrong in this case, too. 

I read 

> // The cast will become unnecessary once #1799 is fixed

and saw #1799 is closed. Then I ran TPC-H with `-q 1 --verify` successfully.

---

Digging deeper, however, I find:

https://github.com/hyrise/hyrise/blob/d50c5db23f1f53dfa5745706ea95cbc3b4f7a0a3/src/lib/expression/expression_utils.hpp#L125-L129

Long / Long → Long, so we should lose precision when calculating an average in a projection without a cast. But how did full-ci pass?

I think it is because in TPC-H Q1, `l_quantity` is `float`. With 

https://github.com/hyrise/hyrise/blob/d50c5db23f1f53dfa5745706ea95cbc3b4f7a0a3/src/lib/operators/aggregate/aggregate_traits.hpp#L58-L66

So, `SUM(l_quantity)` returns a double, making `SUM(l_quantity) / COUNT(*)` a double as well. No loss in precision.

PR #2400, opened 2021-08-22
Prettier Visualizer plans with consistent label wrapping & updated operator descriptions

julianmenzler:
The visualizer uses a `_wrap_label` function for shortening vertex labels. However, the implementation is messy and does not respect newline characters in string arguments, leading to line-width-inconsistencies and sometimes ugly plans.

This PR refactors `_wrap_label` and some operator descriptions to make plans look more nicely. For example: 

> image
> image
> (new plans on the right)

### DIFF
--- a/src/lib/operators/abstract_aggregate_operator.cpp
+++ b/src/lib/operators/abstract_aggregate_operator.cpp
@@ -40,14 +40,12 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
       desc << "," << separator;
     }
     if (lqp_node) {
-      desc << lqp_node->left_input()->output_expressions()[groupby_column_idx]->as_column_name();
+      desc << " " << lqp_node->left_input()->output_expressions()[groupby_column_idx]->as_column_name();
     } else {
       desc << "Column #" + std::to_string(_groupby_column_ids[groupby_column_idx]);
     }
   }
-  desc << "}" << separator;
-  // Add additional newline for clear separation of group-by and aggregate expressions in query plans
-  if (group_by_count > 1 && description_mode == DescriptionMode::MultiLine) desc << separator;
+  desc << " }" << separator;
   auto aggregate_count = _aggregates.size();
   for (auto aggregate_idx = size_t{0}; aggregate_idx < aggregate_count; ++aggregate_idx) {
     if (aggregate_idx > 0) {
--- a/src/lib/operators/abstract_aggregate_operator.cpp
+++ b/src/lib/operators/abstract_aggregate_operator.cpp
@@ -32,7 +32,8 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
   const char separator = (description_mode == DescriptionMode::SingleLine ? ' ' : '\n');
 
   std::stringstream desc;
-  desc << AbstractOperator::description(description_mode) << separator << "GroupBy ColumnIDs: ";
+  desc << AbstractOperator::description(description_mode) << separator;
+  desc << "GroupBy ColumnIDs: ";
   for (size_t groupby_column_idx = 0; groupby_column_idx < _groupby_column_ids.size(); ++groupby_column_idx) {
     desc << _groupby_column_ids[groupby_column_idx];
 
@@ -40,8 +41,8 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
       desc << ", ";
     }
   }
-
-  desc << " Aggregates: ";
+  desc << separator;
+  desc << "Aggregates: ";
   for (size_t expression_idx = 0; expression_idx < _aggregates.size(); ++expression_idx) {
     const auto& aggregate = _aggregates[expression_idx];
     desc << aggregate->as_column_name();
--- a/src/lib/operators/abstract_aggregate_operator.cpp
+++ b/src/lib/operators/abstract_aggregate_operator.cpp
@@ -35,6 +35,7 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
   desc << AbstractOperator::description(description_mode) << separator;
   desc << "GroupBy {";
   auto group_by_count = _groupby_column_ids.size();
+  if (group_by_count > 1 && description_mode == DescriptionMode::MultiLine) desc << separator;
   for (auto groupby_column_idx = size_t{0}; groupby_column_idx < group_by_count; ++groupby_column_idx) {
     if (groupby_column_idx > 0) {
       desc << "," << separator;
@@ -45,6 +46,7 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
       desc << "Column #" + std::to_string(_groupby_column_ids[groupby_column_idx]);
     }
   }
+  if (group_by_count > 1 && description_mode == DescriptionMode::MultiLine) desc << separator;
   desc << "}" << separator;
   auto aggregate_count = _aggregates.size();
   for (auto aggregate_idx = size_t{0}; aggregate_idx < aggregate_count; ++aggregate_idx) {
--- a/src/lib/operators/abstract_aggregate_operator.cpp
+++ b/src/lib/operators/abstract_aggregate_operator.cpp
@@ -41,7 +41,7 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
     }
     const size_t group_by_column_id = _groupby_column_ids[groupby_column_idx];
     if (lqp_node) {
-      desc << lqp_node->left_input()->output_expressions()[group_by_column_id];
+      desc << lqp_node->left_input()->output_expressions()[group_by_column_id]->as_column_name();
     } else {
       desc << "Column #" + std::to_string(group_by_column_id);
     }
--- a/src/lib/operators/abstract_aggregate_operator.cpp
+++ b/src/lib/operators/abstract_aggregate_operator.cpp
@@ -39,10 +39,11 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
     if (groupby_column_idx > 0) {
       desc << "," << separator;
     }
+    const size_t group_by_column_id = _groupby_column_ids[groupby_column_idx];
     if (lqp_node) {
-      desc << lqp_node->left_input()->output_expressions()[groupby_column_idx]->as_column_name();
-    } else {
-      desc << "Column #" + std::to_string(_groupby_column_ids[groupby_column_idx]);
+      desc << lqp_node->left_input()->output_expressions()[group_by_column_id];
+    } else { 
+      desc << "Column #" + std::to_string(group_by_column_id);
     }
   }
   desc << "}" << separator;
--- a/src/lib/operators/abstract_aggregate_operator.cpp
+++ b/src/lib/operators/abstract_aggregate_operator.cpp
@@ -29,7 +29,7 @@ const std::vector<std::shared_ptr<AggregateExpression>>& AbstractAggregateOperat
 const std::vector<ColumnID>& AbstractAggregateOperator::groupby_column_ids() const { return _groupby_column_ids; }
 
 std::string AbstractAggregateOperator::description(DescriptionMode description_mode) const {
-  const auto* const separator = description_mode == DescriptionMode::SingleLine ? " " : "\n";
+  const char separator = (description_mode == DescriptionMode::SingleLine ? ' ' : '\n');
 
   std::stringstream desc;
   desc << AbstractOperator::description(description_mode) << separator << "GroupBy ColumnIDs: ";
--- a/src/lib/operators/abstract_aggregate_operator.cpp
+++ b/src/lib/operators/abstract_aggregate_operator.cpp
@@ -42,7 +42,6 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
     }
   }
   desc << "}" << separator;
-  desc << "Aggregates:" << separator;
   for (size_t expression_idx = 0; expression_idx < _aggregates.size(); ++expression_idx) {
     const auto& aggregate = _aggregates[expression_idx];
     desc << aggregate->as_column_name();
--- a/src/lib/operators/abstract_aggregate_operator.cpp
+++ b/src/lib/operators/abstract_aggregate_operator.cpp
@@ -42,7 +42,7 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
     const size_t group_by_column_id = _groupby_column_ids[groupby_column_idx];
     if (lqp_node) {
       desc << lqp_node->left_input()->output_expressions()[group_by_column_id];
-    } else { 
+    } else {
       desc << "Column #" + std::to_string(group_by_column_id);
     }
   }
--- a/src/lib/operators/abstract_aggregate_operator.cpp
+++ b/src/lib/operators/abstract_aggregate_operator.cpp
@@ -33,13 +33,17 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
 
   std::stringstream desc;
   desc << AbstractOperator::description(description_mode) << separator;
-  desc << "GroupBy ColumnIDs: {";
+  desc << "GroupBy {";
   auto group_by_count = _groupby_column_ids.size();
   for (auto groupby_column_idx = size_t{0}; groupby_column_idx < group_by_count; ++groupby_column_idx) {
     if (groupby_column_idx > 0) {
-      desc << ", ";
+      desc << "," << separator;
     }
-    desc << _groupby_column_ids[groupby_column_idx];
+    if (lqp_node) {
+      desc << lqp_node->left_input()->output_expressions()[groupby_column_idx]->as_column_name();
+    } else {
+      desc << "Column #" + std::to_string(_groupby_column_ids[groupby_column_idx]);
+    }
   }
   desc << "}" << separator;
   auto aggregate_count = _aggregates.size();
--- a/src/lib/operators/abstract_aggregate_operator.cpp
+++ b/src/lib/operators/abstract_aggregate_operator.cpp
@@ -33,7 +33,7 @@ std::string AbstractAggregateOperator::description(DescriptionMode description_m
 
   std::stringstream desc;
   desc << AbstractOperator::description(description_mode) << separator;
-  desc << "GroupBy ColumnIDs: ";
+  desc << "GroupBy ColumnIDs: {";
   for (size_t groupby_column_idx = 0; groupby_column_idx < _groupby_column_ids.size(); ++groupby_column_idx) {
     desc << _groupby_column_ids[groupby_column_idx];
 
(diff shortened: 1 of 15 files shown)


src/lib/operators/abstract_aggregate_operator.cpp
```diff
@@ -29,24 +29,24 @@ const std::vector<std::shared_ptr<AggregateExpression>>& AbstractAggregateOperat
 const std::vector<ColumnID>& AbstractAggregateOperator::groupby_column_ids() const { return _groupby_column_ids; }
 
 std::string AbstractAggregateOperator::description(DescriptionMode description_mode) const {
-  const auto* const separator = description_mode == DescriptionMode::SingleLine ? " " : "\n";
+  const char separator = (description_mode == DescriptionMode::SingleLine ? ' ' : '\n');
 
   std::stringstream desc;
-  desc << AbstractOperator::description(description_mode) << separator << "GroupBy ColumnIDs: ";
+  desc << AbstractOperator::description(description_mode) << separator;
+  desc << "GroupBy ColumnIDs: {";
```
Bensk1, line 36: Should we use `[` over `{`? I associate `{` with sets and `[` with lists.

src/lib/operators/abstract_join_operator.cpp
```diff
@@ -57,18 +57,20 @@ std::string AbstractJoinOperator::description(DescriptionMode description_mode)
...
+  stream << _primary_predicate.predicate_condition << " ";
+  stream << column_name(false, _primary_predicate.column_ids.second);
 
-  // add information about secondary join predicates
+  // Add information about secondary join predicates
   for (const auto& secondary_predicate : _secondary_predicates) {
-    stream << " AND " << column_name(true, secondary_predicate.column_ids.first) << " "
-           << secondary_predicate.predicate_condition << " "
-           << column_name(false, secondary_predicate.column_ids.second);
+    stream << separator << "AND ";
```
Bensk1, line 70: Is that change for MultiLineMode intended? From always ` AND `, we now have `\nAND ` for MultiLineMode which is fine for me.

src/lib/visualization/abstract_visualizer.hpp
```diff
@@ -244,27 +244,43 @@ class AbstractVisualizer {
...
-        current_line_length = 0u;
+    std::stringstream label_stream;
+
+    // 1. Split label into lines
+    std::vector<std::string> lines;
+    boost::split(lines, label, boost::is_any_of("\n"));
+
+    for (size_t line_idx = 0; line_idx < lines.size(); ++line_idx) {
+      if (line_idx > 0) label_stream << '\n';
+      const auto& line = lines.at(line_idx);
```
Bensk1, line 255: It should be safe to use `[]` instead of `.at()` to avoid bounds checking. However, since visualization is not performance critical, it should not be an issue.

src/lib/visualization/abstract_visualizer.hpp
```diff
@@ -244,27 +244,43 @@ class AbstractVisualizer {
...
-      if (current_line_length + word_length > MAX_LABEL_WIDTH) {
-        wrapped_label << "\\n";
-        current_line_length = 0u;
+    std::stringstream label_stream;
+
+    // 1. Split label into lines
+    std::vector<std::string> lines;
+    boost::split(lines, label, boost::is_any_of("\n"));
+
+    for (size_t line_idx = 0; line_idx < lines.size(); ++line_idx) {
```
Bensk1, line 253: Performance is not an issue here, but for consistency, the `size()` should be retrieved only once because the CONTRIBUTING.md says so :)

src/lib/operators/abstract_join_operator.cpp
Bensk1: Could you post an example output plan?

Bensk1 (changes requested):

Bensk1:
Thanks for bringing that up! I agree that the visualizations look less cluttered now.

src/lib/operators/abstract_aggregate_operator.cpp
julianmenzler: Good point 👍

src/lib/operators/abstract_join_operator.cpp
julianmenzler: > Is that change for MultiLineMode intended? 

Yes. For example:

image

julianmenzler: We could capitalize `where` as well?

src/lib/operators/abstract_aggregate_operator.cpp
julianmenzler: Although GroupBy-IDs should be a set?

src/lib/operators/abstract_join_operator.cpp
julianmenzler: Maybe even better:

```
JoinHash
(<join_mode>)
WHERE <predicate>
AND <predicate>
```

julianmenzler: To consider: 
image

Bensk1: Personally, I think I would consider

```
JoinHash (<join_mode>)
WHERE <predicate1>
AND <predicate2>
```

to be the best option.

julianmenzler: image

Bensk1: Looks good to me, @Bouncner any opinion on that?

Bouncner: I like the formatting, but I don't like the WHERE here. WHERE is SQL and filters tuples out, that do not match.
A join predicate can mean something very different (e.g., for outer joins which tuples to keep even though there is no match or anti-joins where is means the opposite).

src/lib/visualization/abstract_visualizer.hpp
Bouncner: And `for (auto line_idx = size_t{0};` ...

Bouncner:
Already love it for the removed line break in the pruning information: "1/3 \n chunk(s)".

Nice.

src/lib/operators/abstract_join_operator.cpp
julianmenzler: Then `ON` or omit at all?

Bouncner: Good question. I would say omit as `ON` is also pretty SQL'y and there is no way one can use `ON` to generate semi joins.

I tend to say omit, but I don't have a strong opinion here.

Bensk1: Same for me, no strong opinion but I would omit it.

julianmenzler: Alright, I removed WHERE.

src/lib/operators/abstract_aggregate_operator.cpp
```diff
@@ -29,24 +29,24 @@ const std::vector<std::shared_ptr<AggregateExpression>>& AbstractAggregateOperat
 const std::vector<ColumnID>& AbstractAggregateOperator::groupby_column_ids() const { return _groupby_column_ids; }
 
 std::string AbstractAggregateOperator::description(DescriptionMode description_mode) const {
-  const auto* const separator = description_mode == DescriptionMode::SingleLine ? " " : "\n";
+  const char separator = (description_mode == DescriptionMode::SingleLine ? ' ' : '\n');
```
Bouncner, line 32: Is the `char` necessary?

src/lib/operators/abstract_aggregate_operator.cpp
```diff
@@ -29,24 +29,24 @@ const std::vector<std::shared_ptr<AggregateExpression>>& AbstractAggregateOperat
...
 
   std::stringstream desc;
-  desc << AbstractOperator::description(description_mode) << separator << "GroupBy ColumnIDs: ";
+  desc << AbstractOperator::description(description_mode) << separator;
+  desc << "GroupBy ColumnIDs: {";
   for (size_t groupby_column_idx = 0; groupby_column_idx < _groupby_column_ids.size(); ++groupby_column_idx) {
     desc << _groupby_column_ids[groupby_column_idx];
 
     if (groupby_column_idx + 1 < _groupby_column_ids.size()) {
```
Bouncner, line 40: What about moving the `if` to the front with `if (groupby_column_idx > 0)` and prepend the comma?

src/lib/operators/abstract_aggregate_operator.cpp
```diff
@@ -29,24 +29,24 @@ const std::vector<std::shared_ptr<AggregateExpression>>& AbstractAggregateOperat
...
       desc << ", ";
     }
   }
-
-  desc << " Aggregates: ";
+  desc << "}" << separator;
   for (size_t expression_idx = 0; expression_idx < _aggregates.size(); ++expression_idx) {
```
Bouncner, line 45: auto and size()

src/lib/operators/abstract_aggregate_operator.cpp
```diff
@@ -29,24 +29,24 @@ const std::vector<std::shared_ptr<AggregateExpression>>& AbstractAggregateOperat
...
     const auto& aggregate = _aggregates[expression_idx];
     desc << aggregate->as_column_name();
 
-    if (expression_idx + 1 < _aggregates.size()) desc << ", ";
+    if (expression_idx + 1 < _aggregates.size()) desc << ", " << separator;
```
Bouncner, line 49: Save as above.

src/lib/operators/abstract_join_operator.cpp
```diff
@@ -57,22 +57,22 @@ std::string AbstractJoinOperator::description(DescriptionMode description_mode)
     return "Column #"s + std::to_string(column_id);
   };
 
-  const auto* const separator = description_mode == DescriptionMode::MultiLine ? "\n" : " ";
-
+  const char separator = (description_mode == DescriptionMode::SingleLine ? ' ' : '\n');
```
Bouncner, line 60: See question above.

src/lib/operators/alias_operator.cpp
```diff
@@ -22,10 +22,11 @@ const std::string& AliasOperator::name() const {
 }
 
 std::string AliasOperator::description(DescriptionMode description_mode) const {
-  const auto* const separator = description_mode == DescriptionMode::SingleLine ? " " : "\n";
+  const char separator = (description_mode == DescriptionMode::SingleLine ? ' ' : '\n');
   std::stringstream stream;
 
-  stream << AbstractOperator::description(description_mode) << separator << "[";
+  stream << AbstractOperator::description(description_mode) << separator;
+  stream << "[";
```
Bouncner, line 29: That added line looks unnecessary. Is it added by the linter?

src/lib/operators/abstract_join_operator.cpp
```diff
@@ -57,22 +57,22 @@ std::string AbstractJoinOperator::description(DescriptionMode description_mode)
...
   std::stringstream stream;
-  stream << AbstractOperator::description(description_mode) << separator << "(" << _mode << " Join where "
-         << column_name(true, _primary_predicate.column_ids.first) << " " << _primary_predicate.predicate_condition
-         << " " << column_name(false, _primary_predicate.column_ids.second);
+  stream << AbstractOperator::description(description_mode);
+  stream << " (" << _mode << ")" << separator;
```
Bouncner, line 63: This does not use the full 120 chars, right? Formatter?

src/lib/visualization/abstract_visualizer.hpp
```diff
@@ -244,27 +244,43 @@ class AbstractVisualizer {
...
+      if (line_idx > 0) label_stream << '\n';
+      const auto& line = lines[line_idx];
+      if (line.length() <= MAX_LABEL_WIDTH) {
+        label_stream << line;
+        continue;
+      }
+      // 2. Split line into words, so we don't break a line in the middle of a word
+      std::vector<std::string> line_words;
+      boost::split(line_words, line, boost::is_any_of(" "));
+      size_t line_length = 0;
```
Bouncner, line 263: auto

src/lib/visualization/abstract_visualizer.hpp
```diff
@@ -244,27 +244,43 @@ class AbstractVisualizer {
...
-    for (const auto& word : label_words) {
-      auto word_length = word.length() + 1;  // include whitespace
-
-      if (current_line_length + word_length > MAX_LABEL_WIDTH) {
-        wrapped_label << "\\n";
-        current_line_length = 0u;
+    std::stringstream label_stream;
+
+    // 1. Split label into lines
+    std::vector<std::string> lines;
```
Bouncner, line 250: auto

src/lib/visualization/abstract_visualizer.hpp
```diff
@@ -244,27 +244,43 @@ class AbstractVisualizer {
...
+    const auto line_count = lines.size();
+    for (auto line_idx = size_t{0}; line_idx < line_count; ++line_idx) {
+      if (line_idx > 0) label_stream << '\n';
+      const auto& line = lines[line_idx];
+      if (line.length() <= MAX_LABEL_WIDTH) {
+        label_stream << line;
+        continue;
+      }
+      // 2. Split line into words, so we don't break a line in the middle of a word
+      std::vector<std::string> line_words;
```
Bouncner, line 261: auto

src/lib/visualization/abstract_visualizer.hpp
```diff
@@ -244,27 +244,43 @@ class AbstractVisualizer {
...
+      size_t word_idx = 0;
+      while (true) {
+        label_stream << line_words.at(word_idx);
+        line_length += line_words.at(word_idx).length();
+
+        // Exit on last word
+        if (word_idx == line_words.size() - 1) break;
+
+        line_length++;  // include whitespace
+        size_t next_line_length = line_length + line_words.at(++word_idx).length();
```
Bouncner, line 273: auto

Bouncner: I am not a big fan ob incrementing a variable within a command.

src/lib/visualization/abstract_visualizer.hpp
```diff
@@ -244,27 +244,43 @@ class AbstractVisualizer {
...
+        if (next_line_length < MAX_LABEL_WIDTH) {
```
Bouncner, line 274: `<=`?

src/lib/operators/abstract_aggregate_operator.cpp
```diff
@@ -29,24 +29,24 @@ const std::vector<std::shared_ptr<AggregateExpression>>& AbstractAggregateOperat
...
 
 std::string AbstractAggregateOperator::description(DescriptionMode description_mode) const {
-  const auto* const separator = description_mode == DescriptionMode::SingleLine ? " " : "\n";
+  const char separator = (description_mode == DescriptionMode::SingleLine ? ' ' : '\n');
 
   std::stringstream desc;
-  desc << AbstractOperator::description(description_mode) << separator << "GroupBy ColumnIDs: ";
+  desc << AbstractOperator::description(description_mode) << separator;
+  desc << "GroupBy ColumnIDs: {";
   for (size_t groupby_column_idx = 0; groupby_column_idx < _groupby_column_ids.size(); ++groupby_column_idx) {
```
Bouncner, line 37: `size_t groupby_column_idx = 0; `. `.size()` as a variable.

src/lib/operators/join_hash.cpp
```diff
@@ -46,9 +46,10 @@ const std::string& JoinHash::name() const {
 }
 
 std::string JoinHash::description(DescriptionMode description_mode) const {
+  const char separator = (description_mode == DescriptionMode::SingleLine ? ' ' : '\n');
   std::ostringstream stream;
-  stream << AbstractJoinOperator::description(description_mode);
-  stream << " Radix bits: " << (_radix_bits ? std::to_string(*_radix_bits) : "Unspecified");
+  stream << AbstractJoinOperator::description(description_mode) << separator;
+  stream << "Radix bits: " << (_radix_bits ? std::to_string(*_radix_bits) : "Unspecified");
```
Bouncner, line 52: Radix bits became part of the tooltip a while ago. Since I am deeply invested in the join, I like the radix information. But it's probably irrelevant for most people.

@julianmenzler @Bensk1 : does it have a value to you listing the radix bits in the PQP? Or is the tooltip sufficient?

src/lib/operators/abstract_aggregate_operator.cpp
```diff
@@ -29,24 +29,24 @@ const std::vector<std::shared_ptr<AggregateExpression>>& AbstractAggregateOperat
...
     desc << _groupby_column_ids[groupby_column_idx];
```
Bouncner, line 38: Sorry to bring that up in your PR. But why on earth don't we print the names of the group by columns? That has been bothering me for quite a long time.

We could extract the `column_name()` method of `abstract_join_operator.cpp:42` and use it here as well.

Bouncner (commented):
Mostly auto, auto, auto, auto ... :)

src/lib/operators/abstract_aggregate_operator.cpp
julianmenzler: No, but it may be more space-efficient?

src/lib/operators/alias_operator.cpp
julianmenzler: No, added for readability reasons.
Maybe this fits better:

```cpp
stream << "[" << boost::algorithm::join(_aliases, ", ") << "]";
```

src/lib/operators/join_hash.cpp
julianmenzler: For me, the tooltip would be sufficient.

src/lib/operators/abstract_aggregate_operator.cpp
Bensk1: I think, the question is not about `char` vs `char*` but more whether `const auto separator` would work.

src/lib/operators/join_hash.cpp
Bensk1: Same for me.

src/lib/visualization/abstract_visualizer.hpp
Bouncner: Let's say you have a limit of 30. You have 19 chars already in place, add a space now. The next word is 10 chars long. You would currently no append the word, even though 19 + 1 + 10 is not over the max line limit.

You would immediately add a linebreak and start the next line.

Or am I wrong here?

src/lib/operators/alias_operator.cpp
Bouncner: Hmm, looking at it again I can see your point.

Do whatever pleases your eye. ;)

src/lib/operators/join_hash.cpp
Bouncner: Then feel free to remove the radix information from the node.

src/lib/operators/abstract_aggregate_operator.cpp
julianmenzler: Got it, thanks! 😄

src/lib/visualization/abstract_visualizer.hpp
julianmenzler: Oh. I did not notice that I already incremented

```cpp
line_length++;  // include whitespace
```

before the if. So you are right, thanks!

src/lib/operators/abstract_aggregate_operator.cpp
julianmenzler: Changed it:

image


image

 
image

Bouncner: 👍

Bouncner (approved):
Looks good to me.

Bensk1 (approved):

PR #2405, opened 2021-09-01
Adjust benchmark_all.sh to allow faster benchmark cycles

Bensk1:
Fixes #2381.

I updated the `benchmark_all.sh` according to #2381 and I want to get some feedback on the changes and whether they reflect what all of you had in mind when we decided to change the script. Currently, this PR is only meant to be used for discussions. Once we agree on the script, I will run some benchmarks and we can validate whether we still trust the script.

- [ ] Tackle the _green for increased throughput issue_

PR #2406, opened 2021-09-08
TPC-DS benchmark: Use SF 10 as default scale factor

mweisgut:
This PR changes the default scale factor of the TPC-DS benchmark to 10.


Bensk1:
That's also included in #2405

PR #2409, opened 2021-09-15
GitHub's "Cite this repository"

Bouncner:
PR to add a bib file (`CITATION.bib`) that refers to the 2019 EDBT paper.

I hope I read this correctly here: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files#other-citation-files

The default syntax is quite strange and annying. Directly using a bibtex file means that GitHub does only provide this file and no other "styles" (e.g., APA style). But I doubt anybody from CS is going to use something else that Bibtex anyways.

### DIFF
--- a/CITATION.bib
+++ b/CITATION.bib
@@ -1,0 +1,17 @@
+@inproceedings{HyriseEDBT2019,
+  author    = {Markus Dreseler and
+               Jan Kossmann and
+               Martin Boissier and
+               Stefan Klauck and
+               Matthias Uflacker and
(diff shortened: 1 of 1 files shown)


Bensk1:
Should we then remove the cite part in the README? Or link from the README to the GitHub citation function?

Bensk1:
If I click on _Cite this repository_ [here](https://github.com/hyrise/hyrise/tree/martin/github_hyrise_citation), it does not work.

PR #2410, opened 2021-09-15
Update Jenkinsfile to new GitHub authentication method

Bensk1:


Bensk1:
The old version still works.

PR #2417, opened 2021-11-03
Improve runtime of histogram creation

Bouncner:
This small PR improves the runtime performance of the histogram creation a little bit, which is currently the main bottleneck for large scale factors by using `boost::pdqsort` over `std::sort`.

Further, we add a (hopefully temporary) note about the current issues with TBB and LLVM 13.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ The install script was tested under macOS Big Sur (10.16) and Ubuntu 20.10 (apt-
 See [dependencies](DEPENDENCIES.md) for a detailed list of dependencies to use with `brew install` or `apt-get install`, depending on your platform. As compilers, we generally use the most recent version of clang and gcc (Linux only). Please make sure that the system compiler points to the most recent version or use cmake (see below) accordingly.
 Older versions may work, but are neither tested nor supported.
 
-**Node about LLVM 13 and TBB 2021:** Hyrise can currently not be built with LLVM 13. We hope to get LLVM 13 running soon. For TBB, please use a `2020*` version until https://github.com/oneapi-src/oneTBB/issues/378 is resolved. On MacOS with brew, LLVM 12 and TBB 2020 can be installed as follows: `brew install tbb@2020 && brew install llvm@12`
+**Note about LLVM 13 and TBB 2021:** Hyrise can currently not be built with LLVM 13. We hope to get LLVM 13 running soon. For TBB, please use a `2020*` version until https://github.com/oneapi-src/oneTBB/issues/378 is resolved. On MacOS with brew, LLVM 12 and TBB 2020 can be installed as follows: `brew install tbb@2020 && brew install llvm@12`
 
 ## Setup using Docker
 If you want to create a Docker-based development environment using CLion, head over to our [dedicated tutorial](https://github.com/hyrise/hyrise/wiki/Use-Docker-with-CLion). 
--- a/README.md
+++ b/README.md
@@ -56,9 +56,9 @@ We support a number of benchmarks out of the box. This makes it easy to generate
 | ---------- | ------------------------------------------------------------------------------------------------------------------------ |
 | TPC-DS     | [Query Plans](https://hyrise-ci.epic-hpi.de/job/hyrise/job/hyrise/job/master/lastStableBuild/artifact/query_plans/tpcds) |
 | TPC-H      | [Query Plans](https://hyrise-ci.epic-hpi.de/job/hyrise/job/hyrise/job/master/lastStableBuild/artifact/query_plans/tpch)  |
+| Join Order | [Query Plans](https://hyrise-ci.epic-hpi.de/job/hyrise/job/hyrise/job/master/lastStableBuild/artifact/query_plans/job)   |
 | JCC-H      | Call the hyriseBenchmarkTPCH binary with the -j flag.                                                                    | 
 | TPC-C      | In development, no proper optimization done yet                                                                          |
-| Join Order |                                                                                                                          |
 
 # Getting started
 
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ The install script was tested under macOS Big Sur (10.16) and Ubuntu 20.10 (apt-
 See [dependencies](DEPENDENCIES.md) for a detailed list of dependencies to use with `brew install` or `apt-get install`, depending on your platform. As compilers, we generally use the most recent version of clang and gcc (Linux only). Please make sure that the system compiler points to the most recent version or use cmake (see below) accordingly.
 Older versions may work, but are neither tested nor supported.
 
+**Node about LLVM 13 and TBB 2:** Hyrise does currently not build using LLVM 13. We hope to get LLVM 13 running soon. For TBB, please use a `2020*` version until https://github.com/oneapi-src/oneTBB/issues/378 is resolved. On MacOS with brew, LLVM 12 and TBB 2020 can be installed as follows: `brew install tbb@2020 && brew install llvm@12`
+
 ## Setup using Docker
 If you want to create a Docker-based development environment using CLion, head over to our [dedicated tutorial](https://github.com/hyrise/hyrise/wiki/Use-Docker-with-CLion). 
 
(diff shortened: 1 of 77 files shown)


src/lib/statistics/statistics_objects/equal_distinct_count_histogram.cpp
```diff
@@ -49,22 +51,55 @@ template <typename T>
...
+    }
+  } 
+
+  if (!pool) {
+    // In case we are not able to estimate the cardinality, we use a memory pool of 1 MiB.
+    constexpr auto 1_MiB = 1'048'576;
+    pool = std::make_unique<std::pmr::monotonic_buffer_resource>(1'048'576);
+  }
+
+  ValueDistributionMap<T> value_distribution_map{estimated_distinct_value_count, &*pool};
```
mrks, line 85: pool.get()

Bouncner:
I'll address obvious typos once the FullCI gives a green light.

README.md
```diff
@@ -75,6 +75,8 @@ The install script was tested under macOS Big Sur (10.16) and Ubuntu 20.10 (apt-
 See [dependencies](DEPENDENCIES.md) for a detailed list of dependencies to use with `brew install` or `apt-get install`, depending on your platform. As compilers, we generally use the most recent version of clang and gcc (Linux only). Please make sure that the system compiler points to the most recent version or use cmake (see below) accordingly.
 Older versions may work, but are neither tested nor supported.
 
+**Node about LLVM 13 and TBB 2021:** Hyrise can currently not be built with LLVM 13. We hope to get LLVM 13 running soon. For TBB, please use a `2020*` version until https://github.com/oneapi-src/oneTBB/issues/378 is resolved. On MacOS with brew, LLVM 12 and TBB 2020 can be installed as follows: `brew install tbb@2020 && brew install llvm@12`
```
Bensk1, line 78: Note

src/lib/statistics/statistics_objects/equal_distinct_count_histogram.cpp
```diff
@@ -49,11 +50,36 @@ template <typename T>
 std::vector<std::pair<T, HistogramCountType>> value_distribution_from_column(const Table& table,
                                                                              const ColumnID column_id,
                                                                              const HistogramDomain<T>& domain) {
-  // TODO(anybody) If you want to look into performance, this would probably benefit greatly from monotonic buffer
-  //               resources.
-  ValueDistributionMap<T> value_distribution_map;
-
+  // Since this function can be the main bottleneck for loading data in the benchmarks, we put some effort in
```
Bensk1, line 53: Minor: `in the benchmark`, do we need this? Isn't you statement true for loading data in general?

src/lib/statistics/statistics_objects/equal_distinct_count_histogram.cpp
```diff
@@ -49,11 +50,36 @@ template <typename T>
...
+  // optimizing it. For once, we try to get the distinct count in case we have dictionary-encoded data in place. This
```
Bensk1, line 54: Is `for once` as in `ausnahmsweise` what you meant to say here?

Bensk1: _allows us to allocate the value_distribution_map with the correct size_?

src/lib/statistics/statistics_objects/equal_distinct_count_histogram.cpp
```diff
@@ -49,11 +50,36 @@ template <typename T>
...
+  // allows us to pre-size the robin_map which helps a lot with performance (no resizes for large maps) and memory
```
Bensk1, line 55: _no resizing for large maps while the histograms are built_?

src/lib/statistics/statistics_objects/equal_distinct_count_histogram.cpp
```diff
@@ -49,11 +50,36 @@ template <typename T>
...
+  // consumption (as we can avoid pre-sizes all maps with the table size, which is often by far too large).
```
Bensk1, line 56: _pre-allocating all maps_? Also, I don't know if we need this second memory consumption reason because it targets another optimization, allocating the maps with the table size, which is not the default case of using the map without any pre-allocation.

src/lib/statistics/statistics_objects/equal_distinct_count_histogram.cpp
```diff
@@ -49,11 +50,36 @@ template <typename T>
...
+
+  // Use a defensive default in case we cannot get a better estimate (i.e., no dictionary-encoded data).
+  auto estimated_distinct_value_count = std::min(table.row_count(), uint64_t{1'024});
+  auto max_load_factor = std::optional<float>{};
   const auto chunk_count = table.chunk_count();
+
+  // Meta tables do not have chunks, we thus need to check for having a chunk count larger zero.
+  if (chunk_count > 0 && table.get_chunk(ChunkID{0})) {
```
Bensk1, line 64: Why do we need both checks? Wouldn't the first one be sufficient for ruling out a meta table?

src/lib/statistics/statistics_objects/equal_distinct_count_histogram.cpp
```diff
@@ -49,11 +50,36 @@ template <typename T>
...
+    // Check if first chunk is dictionary-encoded.
+    if (const auto* dictionary_segment =
+            dynamic_cast<const BaseDictionarySegment*>(&*table.get_chunk(ChunkID{0})->get_segment(column_id))) {
+      estimated_distinct_value_count = std::max(
+          table.row_count(),
+          static_cast<uint64_t>(1.1 * static_cast<double>(chunk_count * dictionary_segment->unique_values_count())));
+      // For rather accurate distinct count estimates, the max_load_factor is upped as we are rather sure than no
+      // resize will happen.
+      max_load_factor = 0.9f;
```
Bensk1, line 73: upped as in increased? In my opinion it needs to have a value to be increased but the optional should not have a value here? Also, I don't understand the argument made in the comment, is the above procedure called `rather accurate`?

Bensk1 (changes requested):

Bensk1:
Have you tried that for large SFs for the TPC-DS?

src/lib/statistics/statistics_objects/equal_distinct_count_histogram.cpp
mweisgut: A previously existing chunk could have been removed (`Table::remove_chunk`), i.e, the pointer has been set to `nullptr`.

src/lib/statistics/statistics_objects/equal_distinct_count_histogram.cpp
mweisgut: than -> that

mweisgut: Using the number of only the first segment's distinct values and multiplying it by the number of chunks assumes a uniform value distribution and that all chunks are existing, doesn't it?

What is the rationale for the factor `1.1`?

Using `MAXIMUM_MAX_LOAD_FACTOR` instead of 0.9f might be more understandable for the reader? To me, 0.9f looks like a magic number.

Bouncner:
Sorry for the unnecessar review work: @Bensk1's remark was point on. For skewed data, the simple heuristic can be even slower. Overall,  it's not a nice estimation and does not bring a significant benefit. Removed it.

Running benchmarks.

Bouncner:
Runtimes:

 * TPC-DS
  * single-threaded: 135s > 134,8s
  * multi-threaded (64 core AMD): 13,3 > 14,3s
 * TPC-H
  * single-threaded: 799s > 746s
  * multi-threaded (64 core AMD): 388s > 352s

The slower TPC-DS runtimes with multi-threading **could** be noise. The MT numbers are rather flaky.

Bensk1 (approved):

Bensk1:
> But it's a single line of code and we are using boost::sort all over anyways. 

I agree with that.

mweisgut (approved):

PR #2420, opened 2021-11-18
Small CMake fixes

Bouncner:
Small fixes for CMake that blocked building Hyrise on a non-root machine.
Boost cannot be build with the rather strict build levels of Hyrise. As we install it via apt, it's included as a system library. If one passes boost to CMake, Hyrise can no longer be build as now boost is no longer included as a system library.

Further, we look for SQLite (`FindSQLite3.cmake`) but do not use the result. Since we usually find in the in the system libs (apt install...) that hasn't been a problem.

~One issue that is not tackled in this PR (yet): we need `libpq` for building Hyrise, but I have not found a way to pass it to CMake. Help is welcome here.~

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,6 +4,7 @@ else()
     set(HYRISE_DEBUG 0)
 endif()
 
+
 add_definitions(-DHYRISE_DEBUG=${HYRISE_DEBUG})
 
 # Provide ENABLE_NUMA_SUPPORT option and automatically disable NUMA if libNUMA was not found
@@ -78,6 +79,7 @@ include_directories(
 include_directories(
     SYSTEM
     ${TBB_INCLUDE_DIR}
+    ${SQLITE3_INCLUDE_DIR}
     ${PROJECT_SOURCE_DIR}/third_party/benchmark/include
     ${PROJECT_SOURCE_DIR}/third_party/cpp-btree/include/btree
     ${PROJECT_SOURCE_DIR}/third_party/cxxopts/include
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -71,13 +71,12 @@ include_directories(
     ${PROJECT_SOURCE_DIR}/src/benchmarklib/
     ${PROJECT_SOURCE_DIR}/src/lib/
     ${PROJECT_SOURCE_DIR}/src/plugins/
-
-    ${Boost_INCLUDE_DIRS}
 )
 
 # Include these libraries as system libraries to silence some warnings caused by -Weverything
 include_directories(
     SYSTEM
+    ${Boost_INCLUDE_DIRS}
     ${TBB_INCLUDE_DIR}
     ${SQLITE3_INCLUDE_DIR}
     ${PROJECT_SOURCE_DIR}/third_party/benchmark/include
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,7 +4,6 @@ else()
     set(HYRISE_DEBUG 0)
 endif()
 
-
 add_definitions(-DHYRISE_DEBUG=${HYRISE_DEBUG})
 
 # Provide ENABLE_NUMA_SUPPORT option and automatically disable NUMA if libNUMA was not found
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,11 @@ if(${CMAKE_VERSION} VERSION_LESS "3.16.0")
     endif()
 endif()
 
+# On Linux, use the lld linker, which is twice as fast for incremental debug builds
+if (NOT APPLE)
+    set(CMAKE_SHARED_LINKER_FLAGS "-fuse-ld=lld")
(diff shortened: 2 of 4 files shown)


Bouncner:
For documentation, this is my current CMake call:
```
cmake .. -DBOOST_INCLUDEDIR="${HOME}/boost/include" -DPostgreSQL_INCLUDE_DIRS="${HOME}/opt/postgres/include/" -DPostgreSQL_LIBRARY_DIR="${HOME}/opt/postgres/lib/" -DHAVE_POSIX_REGEX=0 -DCMAKE_BUILD_TYPE=Release -DREADLINE_INCLUDE_DIR="${HOME}" -DREADLINE_LIBRARY="${HOME}/readline/libreadline.a" -DTBB_INCLUDE_DIR="${HOME}/tbb/include" -DTBB_LIBRARY="${HOME}/tbb/lib/intel64/gcc4.8/libtbb.so" -DSQLITE3_INCLUDE_DIR="${HOME}/opt/sqlite/include" -DSQLITE3_LIBRARY="${HOME}/opt/sqlite/lib/libsqlite3.so" -DCMAKE_CXX_FLAGS="-I ${HOME}/systemtap-4.5/includes -L${HOME}/opt/sqlite/lib/"
```

src/CMakeLists.txt
```diff
@@ -70,14 +70,14 @@ include_directories(
...
-
-    ${Boost_INCLUDE_DIRS}
 )
 
 # Include these libraries as system libraries to silence some warnings caused by -Weverything
 include_directories(
     SYSTEM
+    ${Boost_INCLUDE_DIRS}
     ${TBB_INCLUDE_DIR}
+    ${SQLITE3_INCLUDE_DIR}
```
mweisgut, line 80: With my limited knowledge about CMake (mostly based on https://cliutils.gitlab.io/modern-cmake/), I think we should avoid adding include directories of third-party libraries with global functions such as `include_directories`. #2032 also mentions this. Instead, I suggest creating a library target with the include directories and linking the library target to another target. If the library target only contains header files, which would be true for the boost library here, this target can be an INTERFACE target.

I'm cleaning up some CMake files in #2421. In that PR, I also created a tbb library target, which can be linked to another library/executable then. This is a way we could also do it for boost and sqlite.

Bouncner: I don't have a clue what you're talking about. :D 
But do you think #2421 will make it into the master soon? Merging these two lines would help me right now while working within the DE lab. If it's in the end about rewriting this include lines, this PR here wouldn't interfere with yours right?

mweisgut: #2421 won't make it into the master soon. One CI stage fails and I have to debug it...

This PR would not interfere with #2421, but it also does not add include directories of third-party libraries properly IMO. I defined library targets for tbb, sqlite3, and boost in #2422 and linked them as system libraries. Would that also solve your issue? However, the CI run is still running, so we will see whether it succeeds.

mweisgut: Never mind, #2422 also fails, and it will probably take some time to find the root cause. I'm okay with applying these changes for now and finding a cleaner later on.

mweisgut (approved):

Bensk1 (changes requested):
See question in discussion

Bensk1:
> For documentation, this is my current CMake call:
>  [...]


It this detailed CMake call necessary on a non-root machine or even on root machines now? Is it worth it adding it to README?

src/CMakeLists.txt
Bouncner: I would usually say "I'll have a look and help you", but I have absolutely  no idea about that whole cmake thing. :-/

Bensk1 (approved):

Bouncner:
When building on NUMA servers, don't forget to link `libnuma` (thanks @Bensk1 ).

```
martin.boissier@node-19:~/hyrise/rel_gcc$ rm -rf * && cmake .. -DBOOST_INCLUDEDIR="${HOME}/boost/include" -DPostgreSQL_INCLUDE_DIRS="${HOME}/opt/postgres/include/" -DHAVE_POSIX_REGEX=0 -DCMAKE_BUILD_TYPE=Release -DREADLINE_INCLUDE_DIR="${HOME}" -DREADLINE_LIBRARY="${HOME}/readline/libreadline.a" -DTBB_INCLUDE_DIR="${HOME}/tbb/include" -DTBB_LIBRARY="${HOME}/tbb/lib/intel64/gcc4.8/libtbb.so" -DSQLITE3_INCLUDE_DIR="${HOME}/opt/sqlite/include" -DSQLITE3_LIBRARY="${HOME}/opt/sqlite/lib/libsqlite3.so" -DCMAKE_CXX_FLAGS="-I ${HOME}/systemtap-4.5/includes -L${HOME}/opt/sqlite/lib/ -L${HOME}/opt/postgres/lib/ -lpq" -DNUMA_INCLUDE_DIR="${HOME}/opt/numactl/include" -DNUMA_LIBRARY="${HOME}/opt/numactl/lib/libnuma.so"
```

PR #2431, opened 2022-01-14
Upgrade SQL-parser version

mweisgut:
This PR upgrades the SQL parser to the latest version, which requires some changes in the SQL translator.

### DIFF
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -1355,7 +1355,7 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
           column_definition.data_type = DataType::String;
           break;
         case hsql::DataType::TIME:
-          std::cout << "WARNING: Parsing DATETIME to string since date and time data types are not supported yet\n";
+          std::cout << "WARNING: Parsing TIME to string since date and time data types are not supported yet\n";
           column_definition.data_type = DataType::String;
           break;
         case hsql::DataType::UNKNOWN:
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -1263,6 +1263,8 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create(const hsql::Cr
       return _translate_create_table(create_statement);
     case hsql::CreateType::kCreateTableFromTbl:
       FailInput("CREATE TABLE FROM is not yet supported");
+    case hsql::CreateType::kCreateIndex:
+      FailInput("CREATE INDEX is not yet supported");
   }
   Fail("Invalid enum value");
 }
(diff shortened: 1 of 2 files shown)


mweisgut:
The SQL parser contains a bug causing the CI crash. Patch: https://github.com/hyrise/sql-parser/pull/201

src/lib/sql/sql_translator.cpp
```diff
@@ -1309,27 +1311,53 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+          column_definition.data_type = DataType::Float;
+          break;
         case hsql::DataType::FLOAT:
           column_definition.data_type = DataType::Float;
           break;
         case hsql::DataType::DOUBLE:
           column_definition.data_type = DataType::Double;
           break;
         case hsql::DataType::CHAR:
+          std::cout << "WARNING: Ignoring the length of CHAR as Hyrise has no way of working with it\n";
```
Bouncner, line 1338: "Ignoring the length of CHAR. Hyrise's strings are not length-limited." ?

Bouncner: I don't think the "no way of working with it" is not clear enough as a user information. Sounds like we don't know what to do with the text.

Bouncner (commented):
Just two small issues (see review).

And is it "not yet supported" (CREATE INDEX) or is it "not supported yet" (date types)?

Besides, are there any tests we need with the latest SQL parser changes? I don't have an overview what parser changes get included with this update.

src/lib/sql/sql_translator.cpp
```diff
@@ -1335,27 +1335,27 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+          std::cout << "WARNING: Ignoring the length of CHAR. Hyrise's strings are not length-limited\n";
           column_definition.data_type = DataType::String;
           break;
         case hsql::DataType::VARCHAR:
-          std::cout << "WARNING: Ignoring the length of VARCHAR as Hyrise has no way of working with it\n";
+          std::cout << "WARNING: Ignoring the length of VARCHAR. Hyrise's strings are not length-limited\n";
           column_definition.data_type = DataType::String;
           break;
         case hsql::DataType::TEXT:
-          std::cout << "WARNING: Ignoring the length of TEXT as Hyrise has no way of working with it\n";
```
Bouncner: Wait, as far as I remember, TEXT does not have an optional size. At least in the major DBMSs. So actually, TEXT is pretty much what we do in Hyrise as the default.

mweisgut:
> Besides, are there any tests we need with the latest SQL parser changes? I don't have an overview what parser changes get included with this update.

Additional features and changes of the SQL parser should be tested in its test suite accordingly.

For Hyrise, besides running the test suite successfully, I think we should check if the benchmarks (TPC-H, TPC-DS, JOB, JCC-H, and TPC-C) can be executed successfully. The TPC-H, TPC-DS, and JOB is checked in the CI already.

Bouncner (approved):

mweisgut:
Check if the following benchmarks still run successfully with this PR
- [x] TPC-H
- [x] TPC-DS
- [x] TPC-C
- [x] JOB
- [ ] JCC-H (not checked yet)

Bensk1:
> For Hyrise, besides running the test suite successfully, I think we should check if the benchmarks (TPC-H, TPC-DS, JOB, JCC-H, and TPC-C) can be executed successfully. The TPC-H, TPC-DS, and JOB are checked in the CI already.

FullCI should execute the TPCC (clang-release stage) and JCCH (system-tests stage) with verification.

PR #2435, opened 2022-02-15
Update Jenkinsfile for upgraded macOS CIs

Bensk1:


src/test/lib/utils/meta_tables/meta_log_table_test.cpp
```diff
@@ -39,9 +39,10 @@ TEST_F(MetaLogTest, TableGeneration) {
 
   const auto values = meta_table->get_row(0);
 
-  // Log entry should be created less than a minute ago but before the meta table was generated
+  // Log entry should be created less than a minute ago but before the meta table was generated.
   EXPECT_GT(boost::get<int64_t>(values[0]), timestamp_ns - static_cast<int64_t>(60e9));
-  EXPECT_LT(boost::get<int64_t>(values[0]), timestamp_ns);
+  // Since the clock is not necessarily precise enough, we use "less or equals" instead of "less than" here.
+  EXPECT_LE(boost::get<int64_t>(values[0]), timestamp_ns);
```
Bensk1, line 45: I don't really know why this is necessary. Without this change the test fails sometimes on ARM because `timestamp_ns` is equal to `values[0]`. I have never seen this happening before.

The macOS and package upgrades might have changed something about the clock resolution or speed. Still, with the additional comment and minor code change, it should be fine.

Bouncner (approved):

Issue #2437, opened 2022-02-25
Discussion: `PRAGMA` commands


Bouncner:
There is a growing number of plugins or projects on Hyrise that would benefit from "some way" of allowing the user to send some from of commands or to call a plugin-provided function.

Looking at the Postgres-Client, I have seen a suitable way to implement them with our parser. Another way could be to support `PRAGMA` queries in Hyrise (similar to SQLite or DuckDB; e.g., as a `simply query` in the request_handler).
We would extend to the Plugin Manager to accept hooks, that a plugin is responsible for certain `PRAGMAs`.

One example here is the decision that certain segments should be reenencoded or tiered. Enable us to instrument Hyrise from a simply Python script using PRAGMAs would ease a lot of such scenarios.

Any comments on that?

Bensk1:
Good point.

We could also (mis)use the MetaTables and/or plugin settings from that. I don't like that idea from an interface and user experience perspective but it probably easy to implement.

Bouncner:
Given that SQLite and co. do the same, I think it's one of better approaches concerning user experience.

The problem with the meta tables is that there is no clean way to articulate "an action" via SQL, as far as I know. I currently use a hacky plugin that creates a meta table `commands`. First, I insert the "command" into the table (inserting does currently not trigger an action) and then I need to use a `select * from meta_commands` which calls `_on_generate` of the meta table and here I can react.
We could extend the interfact to allow actions on insert (or I might have missed it?), but nothing "SQL'ish" looks better to me than `PRAGMA`.

Bensk1:
Discussion moved to draft implementation in #2438.

mrks:
> The problem with the meta tables is that there is no clean way to articulate "an action" via SQL, as far as I know.

`CALL` maybe?

Bensk1:
The demanded functionality was implemented via #2438 even though it does neither use `PRAGMA` nor `CALL` for now but `INSERT INTO meta_exec...`.

PR #2438, opened 2022-03-02
Add user-callable functions to plugins

Bensk1:
As highlighted by @Bouncner in #2437 there are many scenarios where the user (_Anwender_) wants to trigger a plugin function, e.g., to manually start reoptimization. 

There are two challenges to solve for this to work.

1. How should the user interaction look like? 
1. How can the PluginManager call plugin functions that are not known in advance without modifying the PluginManager. 

A possible solution for challenge 1 could be to use a new meta table. Inserting a row containing the name of the plugin and action to execute into this table could trigger the action.

This is the first proposal for a solution for challenge 2. The plugin's developer can just implement user-callable functions as a _normal_ plugin function. Afterward, the developer only has to use the macro `EXPORT_USER_CALLABLE_FUNCTION` that the AbstractPlugin provides. 

Alternatively, @Bouncner also had the idea that plugins can subscribe keywords and function calls to the PluginManager upon the instantiation of the plugins. The PluginManager maintains a list of such keywords. If a keyword is sent, the corresponding function is called. This might be a cleaner solution but also require more changes and I don't find the above's solution particularly unclean.
~~_Edit: I don't know whether this solution would work at all. The PluginManager would need to keep a map of keywords and function pointers. However, function pointers to member functions also contain the class type. For different plugins, the class type varies so we could not create such a map at compile time. Using AbstractPlugin as type would also not work because then we would need to define all user-callable functions for AbstractPlugin, too. Am I missing something or would this really not work?_~~ << It [does work,](https://gist.github.com/Bouncner/6d31c7ebe55aeb9d9fb8d0a5efb654fb) as @Bouncner pointed out.

The interface via MetaTables would be mostly identical for both solutions. Also, I believe both solutions can only offer non-parameterized and void-returning functions as the types must be known during the PluginManager's compile time.

### DIFF
--- a/src/lib/utils/abstract_plugin.hpp
+++ b/src/lib/utils/abstract_plugin.hpp
@@ -12,6 +12,9 @@ namespace opossum {
 #define EXPORT_PLUGIN(PluginName) \
   extern "C" AbstractPlugin* factory() { return new PluginName(); }
 
+using PluginFunctionName = std::string;
+using PluginFunctionPointer = std::function<void(void)>;
+
 // AbstractPlugin is the abstract super class for all plugins. An example implementation can be found
 // under test/utils/test_plugin.cpp. Usually plugins are implemented as singletons because there
 // shouldn't be multiple instances of them as they would compete against each other.
@@ -25,7 +28,7 @@ class AbstractPlugin {
 
   virtual void stop() = 0;
 
-  virtual std::vector<std::pair<std::string, std::function<void(void)>>> keywords_functions() const {
+  virtual std::vector<std::pair<PluginFunctionName, PluginFunctionPointer>> get_user_executable_functions() const {
     return {};
   }
 };
--- a/src/lib/utils/abstract_plugin.hpp
+++ b/src/lib/utils/abstract_plugin.hpp
@@ -28,7 +28,7 @@ class AbstractPlugin {
 
   virtual void stop() = 0;
 
-  virtual std::vector<std::pair<PluginFunctionName, PluginFunctionPointer>> get_user_executable_functions() const {
+  virtual std::vector<std::pair<PluginFunctionName, PluginFunctionPointer>> provided_user_executable_functions() const {
     return {};
   }
 };
--- a/src/lib/utils/abstract_plugin.hpp
+++ b/src/lib/utils/abstract_plugin.hpp
@@ -24,6 +24,10 @@ class AbstractPlugin {
   virtual void start() = 0;
 
   virtual void stop() = 0;
+
+  virtual std::vector<std::pair<std::string, std::function<void(void)>>> keywords_functions() const {
+    return {};
+  }
 };
 
 }  // namespace opossum
--- a/src/lib/utils/abstract_plugin.hpp
+++ b/src/lib/utils/abstract_plugin.hpp
@@ -28,6 +28,12 @@ class AbstractPlugin {
 
   virtual void stop() = 0;
 
+  // This method provides an interface for making plugin functions executable by plugin users.
+  // Please see the test_plugin.cpp and the meta_exec_table_test.cpp for examples on how to use it.
+  // IMPORTANT: The execution of such user-executable functions is blocking, see the PluginManager's
+  // exec_user_function method. If you are writing a plugin and provide user-exectuable functions it
+  // is YOUR responsibility to keep these function calls as short and efficient as possible, e.g.,
+  // by spinning up a thread inside the plugin to execute the actual functionality.
   virtual std::vector<std::pair<PluginFunctionName, PluginFunctionPointer>> provided_user_executable_functions() const {
     return {};
   }
--- a/src/lib/utils/plugin_manager.cpp
+++ b/src/lib/utils/plugin_manager.cpp
@@ -35,6 +35,11 @@ std::vector<PluginName> PluginManager::loaded_plugins() const {
   return plugin_names;
 }
 
+std::unordered_map<std::pair<PluginName, PluginFunctionName>, PluginFunctionPointer, plugin_name_function_name_hash>
+PluginManager::user_executable_functions() const {
+  return _user_executable_functions;
+}
+
 void PluginManager::load_plugin(const std::filesystem::path& path) {
   const auto plugin_name = plugin_name_from_path(path);
 
--- a/src/lib/utils/plugin_manager.cpp
+++ b/src/lib/utils/plugin_manager.cpp
@@ -70,7 +70,8 @@ void PluginManager::load_plugin(const std::filesystem::path& path) {
 }
 
 void PluginManager::exec_user_function(const PluginName& plugin_name, const PluginFunctionName& function_name) {
-  Assert(_user_executable_functions.count({plugin_name, function_name}) > 0, "There is no " + function_name + " defined for plugin " + plugin_name + ".");
+  Assert(_user_executable_functions.count({plugin_name, function_name}) > 0,
+         "There is no " + function_name + " defined for plugin " + plugin_name + ".");
 
   const auto user_executable_function = _user_executable_functions[{plugin_name, function_name}];
   user_executable_function();
@@ -80,7 +81,8 @@ void PluginManager::exec_user_function(const PluginName& plugin_name, const Plug
 
 void PluginManager::unload_plugin(const PluginName& plugin_name) {
   auto plugin_iter = _plugins.find(plugin_name);
-  Assert(plugin_iter != _plugins.cend(), "Unloading plugin failed: A plugin with name " + plugin_name + " does not exist.");
+  Assert(plugin_iter != _plugins.cend(),
+         "Unloading plugin failed: A plugin with name " + plugin_name + " does not exist.");
 
   _unload_and_erase_plugin(plugin_iter);
 }
--- a/src/lib/utils/plugin_manager.cpp
+++ b/src/lib/utils/plugin_manager.cpp
@@ -76,7 +76,7 @@ void PluginManager::load_plugin(const std::filesystem::path& path) {
 }
 
 void PluginManager::exec_user_function(const PluginName& plugin_name, const PluginFunctionName& function_name) {
-  Assert(_user_executable_functions.count({plugin_name, function_name}) > 0,
+  Assert(_user_executable_functions.count({plugin_name, function_name}) == 1,
          "There is no " + function_name + " defined for plugin " + plugin_name + ".");
 
   const auto user_executable_function = _user_executable_functions[{plugin_name, function_name}];
--- a/src/lib/utils/plugin_manager.cpp
+++ b/src/lib/utils/plugin_manager.cpp
@@ -62,6 +62,7 @@ void PluginManager::load_plugin(const std::filesystem::path& path) {
   plugin_handle_wrapper.plugin->start();
   _plugins[plugin_name] = std::move(plugin_handle_wrapper);
 
+  // Add the newly loaded plugin's user executable functions to our map of functions.
   const auto user_executable_functions = _plugins[plugin_name].plugin->get_user_executable_functions();
   for (const auto& [function_name, function_pointer] : user_executable_functions) {
     _user_executable_functions[{plugin_name, function_name}] = function_pointer;
@@ -89,7 +90,7 @@ std::unordered_map<PluginName, PluginHandleWrapper>::iterator PluginManager::_un
   const PluginName plugin_name = plugin_iter->first;
   const auto& plugin_handle_wrapper = plugin_iter->second;
 
-  // Delete keywords and functions for plugin to be unloaded.
+  // Delete user exectuable functions of the plugin to be unloaded from the map of functions.
   std::erase_if(_user_executable_functions, [&](const auto& item) {
     const auto& [item_plugin_name, _] = item.first;
     return item_plugin_name == plugin_name;
--- a/src/lib/utils/plugin_manager.cpp
+++ b/src/lib/utils/plugin_manager.cpp
@@ -43,7 +43,8 @@ PluginManager::user_executable_functions() const {
 void PluginManager::load_plugin(const std::filesystem::path& path) {
   const auto plugin_name = plugin_name_from_path(path);
 
-  Assert(!_plugins.count(plugin_name), "Loading plugin failed: A plugin with name " + plugin_name + " already exists.");
+  Assert(!_plugins.contains(plugin_name),
+         "Loading plugin failed: A plugin with name " + plugin_name + " already exists.");
 
   PluginHandle plugin_handle = dlopen(path.c_str(), static_cast<uint8_t>(RTLD_NOW) | static_cast<uint8_t>(RTLD_LOCAL));
   Assert(plugin_handle, std::string{"Loading plugin failed: "} + dlerror());
--- a/src/lib/utils/plugin_manager.cpp
+++ b/src/lib/utils/plugin_manager.cpp
@@ -69,7 +69,7 @@ void PluginManager::load_plugin(const std::filesystem::path& path) {
   _plugins[plugin_name] = std::move(plugin_handle_wrapper);
 
   // Add the newly loaded plugin's user executable functions to our map of functions.
-  const auto user_executable_functions = _plugins[plugin_name].plugin->get_user_executable_functions();
+  const auto user_executable_functions = _plugins[plugin_name].plugin->provided_user_executable_functions();
   for (const auto& [function_name, function_pointer] : user_executable_functions) {
     _user_executable_functions[{plugin_name, function_name}] = function_pointer;
   }
--- a/src/lib/utils/plugin_manager.cpp
+++ b/src/lib/utils/plugin_manager.cpp
@@ -35,9 +35,9 @@ std::vector<PluginName> PluginManager::loaded_plugins() const {
 }
 
 void PluginManager::load_plugin(const std::filesystem::path& path) {
-  const auto name = plugin_name_from_path(path);
+  const auto plugin_name = plugin_name_from_path(path);
 
-  Assert(!_plugins.count(name), "Loading plugin failed: A plugin with name " + name + " already exists.");
+  Assert(!_plugins.count(plugin_name), "Loading plugin failed: A plugin with name " + plugin_name + " already exists.");
 
   PluginHandle plugin_handle = dlopen(path.c_str(), static_cast<uint8_t>(RTLD_NOW) | static_cast<uint8_t>(RTLD_LOCAL));
   Assert(plugin_handle, std::string{"Loading plugin failed: "} + dlerror());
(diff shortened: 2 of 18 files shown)


src/lib/utils/abstract_plugin.hpp
```diff
@@ -12,6 +12,21 @@ namespace opossum {
...
+// This macro allows the function `FunctionName` of the plugin called `PluginName` to be
+// externally callable. This is necessary to enable user-callable functions. As above,
+// it leads to plain C linkage. Use EXPORT in the implementation file of your plugin.
+#define EXPORT_USER_CALLABLE_FUNCTION(PluginName, FunctionName) \
+  /* Careful! We are getting a raw pointer from a smart (unique) pointer here. */ \
+  /* This could conflict with the intention of unique_ptr. It should not be    */ \
+  /* a problem, as long as we handle it responsible and let it go out of scope */ \
+  /* at the end of this function.                                              */ \
+  extern "C" void FunctionName ## _extern(const std::unique_ptr<AbstractPlugin>& plugin) { \
+    const auto plugin_raw = plugin.get(); \
```
Bensk1, line 24: Unfortunately, I think there is no way around that.

Bouncner:
> The interface via MetaTables would be mostly identical for both solutions. Also, I believe both solutions can only offer non-parameterized and void-returning functions as the types must be known during the PluginManager's compile time.

Yes, I don't see a nice way around non-parameterized void functions. But that shouldn't be a big issue. Return data can be written to the meta log and parameters can be obtained via the settings.

Bensk1:
> Yes, I don't see a nice way around non-parameterized void functions. But that shouldn't be a big issue. Return data can be written to the meta log and parameters can be obtained via the settings.

Exactly.

Bensk1:
[This is](https://github.com/hyrise/hyrise/commit/430679a2021d2b8455b90158bf0d57dcc04b6427) how the alternative implementation would look like.

src/lib/utils/abstract_plugin.hpp
mweisgut: Do we have to use an `std::unique_ptr` here?

mweisgut:
From a user experience perspective, I personally would prefer executing the SQL command `PRAGMA test_user_callable_function` instead of `INSERT INTO meta_exec (plugin_name, function_name) VALUES ('hyriseTestPlugin', 'test_user_callable_function_extern')`.

For this, a pragma operator could be added, which calls the plugins' function calls and could also allow enabling/loading and disabling/unloading plugins.
Furthermore, in [this approach](https://github.com/hyrise/hyrise/commit/430679a2021d2b8455b90158bf0d57dcc04b6427), another way is to map a keyword instead of a {plugin name, keyword} pair to an std::function. This, however, requires that the keywords are unique across all plugins, which can be ensured by naming guidelines and integration tests.

Nevertheless, the `IMPORT` approach probably requires less effort since it does not require an additional operator.

>Yes, I don't see a nice way around non-parameterized void functions.

I think it would not be assessed as `nice way around`, but string-only functions might be a potential way to provide parameterized functions. In the function body, the parameters can then be converted to the required data type.

Bensk1:
> From a user experience perspective, I personally would prefer executing the SQL command `PRAGMA test_user_callable_function` instead of `INSERT INTO meta_exec (plugin_name, function_name) VALUES ('hyriseTestPlugin', 'test_user_callable_function_extern')`.

I agree, it might be a bit cleaner from a user perspective. On the other hand, `PRAGMA` is not part of the SQL standard, is it? Also we would need a new operator and add non-standard keywords to the parser just for a feature that is rarely used. Lastly, in industry it is apparently quite common to use special tables (similar to what we do with meta tables) for such tasks. I am not against the PRAGMA but I believe the other approach is more PRAGMAtic (haha).

> For this, a pragma operator could be added, which calls the plugins' function calls and could also allow enabling/loading and disabling/unloading plugins. Furthermore, in [this approach](https://github.com/hyrise/hyrise/commit/430679a2021d2b8455b90158bf0d57dcc04b6427), another way is to map a keyword instead of a {plugin name, keyword} pair to an std::function. This, however, requires that the keywords are unique across all plugins, which can be ensured by naming guidelines and integration tests.

So what's the advantage of just using a keyword instead of the pair? I just see additional limitations: only unique keywords. Also, since the development plugins is on purpose decoupled from the core, I don't know if we could easily build such integration tests. 

> Nevertheless, the `IMPORT` approach probably requires less effort since it does not require an additional operator.

Just to be sure, you meant `INSERT` here, right?

> I think it would not be assessed as `nice way around`, but string-only functions might be a potential way to provide parameterized functions. In a corresponding function body, the parameters can then be converted to the required data types.

Good point, that could be an option. For now, I find the current implementation sufficient as we can use the settings table for parameters as @Bouncner suggested.

src/lib/utils/abstract_plugin.hpp
Bensk1: The PluginManager stores plugins as unique_ptrs. I think in the beginning we thought that the PluginManager should have exclusive ownership and that there is no reason for shared_ptrs. We could change that but might go with the alternative implementation in the end anyway.

mweisgut:
> So what's the advantage of just using a keyword instead of the pair?

I think convenience on the user side by simplified/shorter SQL commands is the only benefit.

> Just to be sure, you meant INSERT here, right?

Right.

src/lib/utils/abstract_plugin.hpp
mweisgut: With shared pointers, we can avoid a raw pointer object in this code location, a raw pointer cast, and the comments explaining why we use a raw pointer here.
The minimal overhead for reference counting with shared pointers should also be negligible. Am I missing some drawbacks of changing to shared pointers here?

Bouncner:
Hmm, seeing this unique_ptr “issue”, I’d vote for the functor.
Especially with the `register_keyword()` function, the interface is much better, no extern usage, and same functionality.

mweisgut:
> Hmm, seeing this unique_ptr “issue”, I’d vote for the functor map solution. Especially with the `register_keyword()` function, the interface is much better, no extern usage, and same functionality.

+1

src/lib/utils/abstract_plugin.hpp
Bensk1: Technically, we could change to a `shared_ptr` but we chose the `unique_ptr` on purpose to declare exclusive ownership. The idea was that the PluginManager is not sharing plugin ownership because otherwise things could get nasty when unloading the shared library.

Bensk1:
> Hmm, seeing this unique_ptr “issue”, I’d vote for the functor map solution.
> Especially with the `register_keyword()` function, the interface is much better, no extern usage, and same functionality.

Just for clarification, the `extern` only enables C linkage to avoid name mangling. We use to be able to call `test_user_callable_function` instead of `_ZN9Opossum7TestPlugin6test_user_callable_functionEv`.

What I find the biggest argument for the alternative map solution: it enables us to allow providing a list of all available execution functions such that the plugin user does not have to consult the plugin's documentation for that.

Bouncner:
I read a little bit about extern and yes, it's apparently no magic. But it's still kind of a work around from a pure C++ perspective and I'd say we need a very good reason to use workarounds.

So the advantages I see are (i) pure C++ with easy debugging (e.g., looping prints over the map) and (ii) the explicit `register_keyword()` API. But both are not strong points. Unfortunately, I am not sufficiently familiar with this whole topic to estimate potential problems that might occur later on (anything to be aware of when dynamic loading magic is used? I don't know!)

dey4ss:
I like the idea of calling plugin code from outside. For instance, it would be a nice functionality for my master's thesis plugin. Regarding `PRAGMA` vs. meta table for triggering execution, I see one advantage for the first. If we want to allow parameters for a call, we can easily use a vector of strings that defaults to be empty in the parser. Thus, the user can (i) omit providing a parameter if there is none, and (ii) provide multiple parameters. The downside, besides lacking conformance to the standard, is that the user has to know the plugin implementation. I would also note that we have other keywords that are not part of the standard, such as `IMPORT`, or `COPY` only being part of the Postgres dialect.
From my POV, having to know the plugin code is the most severe argument against a `PRAGMA` command and for the meta table. Though parameters could be provided via a string (similar to settings), the user would have to always provide an empty string if there is no parameter. Again, this requires knowledge of the plugin implementation, does not nicely adapt to functions with multiple parameters, and is kind of ugly when no parameters are required.

tl;dr: I'd say using a meta table and settings for parameters (default values, uniform query structure) maybe is the best solution for now. Just a minor note: do we want to let the meta table keep track of the call stack? This would make the history accessible even if plugins do not log their execution itself and stores execution and execution history at one place.

Bouncner:
> From my POV, having to know the plugin code is the most severe argument against a `PRAGMA` command and for the meta table.

And do you have an opinion on the `std::map` vs. `EXPORT_USER_CALLABLE_FUNCTION` discussion?

> tl;dr: I'd say using a meta table and settings for parameters (default values, uniform query structure) maybe is the best solution for now. Just a minor note: do we want to let the meta table keep track of the call stack? This would make the history accessible even if plugins do not log their execution itself and stores execution and execution history at one place.

Sound reasonable and helpful. Should be basically for free, I guess. Would you store the call stack as a string in a `meta_exec_log` table with a timestamp or something different or more complex? Any ideas how to obtain the call stack? I have found `boost::stacktrace` and a bunch of platform-dependent stuff.

Bensk1:
> Just a minor note: do we want to let the meta table keep track of the call stack? This would make the history accessible even if plugins do not log their execution itself and stores execution and execution history at one place.

Do you mean an actual call stack or just the history of user-triggered plugin function calls? These could be simply stored in the table itself. 

In general, how about moving the whole discussion to today's fortnightly? I think this could make the whole discussion more hyrisey,i.e., most efficient.

dey4ss:
>Do you mean an actual call stack or just the history of user-triggered plugin function calls? These could be simply stored in the table itself.

Exactly, I thought about something like `std::vector<std::pair<std::string, std::string>>` within the meta table. If we return this list, it also preserves the semantics of inserting something into this table.

>In general, how about moving the whole discussion to today's fortnightly? I think this could make the whole discussion more hyrisey,i.e., most efficient.

+1

Bensk1:
In today's Fortnightly we decided:

- to go for the alternative approach because it is more expressive and allows us to provide a list of user-callable functions to the user
- to use the `INSERT INTO` over the `PRAGMA` approach because it does not require parser adaption and is in general easier to implement. At the same time it does not prevent any future more powerful implementations
- to actually insert into the meta table which function was called to have a history of called functions.

src/test/lib/utils/meta_tables/meta_exec_table_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+  EXPECT_TRUE(sm.has_table("TableOfTestPlugin"));
+
+  SQLPipelineBuilder{
+      "INSERT INTO meta_exec (plugin_name, function_name) VALUES ('hyriseSecondTestPlugin', "
+      "'OurFreelyChoosableFunctionName')"}
+      .create_pipeline()
+      .get_result_table();
+  // The second test plugin creates the below table when the called function is executed
+  EXPECT_TRUE(sm.has_table("TableOfSecondTestPlugin"));
+}
```
Bensk1, line 63: These and the following tests kind of mirror some of the PluginManagerTests. It's pretty much the except that the SQLPipeline and everything that comes with it is in between. However, it's still nice to see everything working from end to end but we don't have to keep it.

src/lib/utils/plugin_manager.cpp
```diff
@@ -60,21 +66,48 @@ void PluginManager::load_plugin(const std::filesystem::path& path) {
...
 
-void PluginManager::unload_plugin(const PluginName& name) {
-  auto plugin_iter = _plugins.find(name);
-  Assert(plugin_iter != _plugins.cend(), "Unloading plugin failed: A plugin with name " + name + " does not exist.");
+void PluginManager::exec_user_function(const PluginName& plugin_name, const PluginFunctionName& function_name) {
+  Assert(_user_executable_functions.count({plugin_name, function_name}) > 0,
+         "There is no " + function_name + " defined for plugin " + plugin_name + ".");
+
+  const auto user_executable_function = _user_executable_functions[{plugin_name, function_name}];
+  user_executable_function();
```
Bensk1, line 83: This call is blocking. The called plugin could do some long running fancy optimization here which would be bad. Currently, it should be the plugin's responsibility to keep these functions short and, e.g., start a thread. We could also execute the function in a thread to avoid problems here. Also, what happens if `user_executable_function()` crashes?

src/test/lib/utils/lossless_predicate_cast_test.cpp
```diff
@@ -18,7 +18,7 @@ TEST_F(LosslessPredicateCastTest, NextFloatTowards) {
   // Edge cases:
   EXPECT_EQ(next_float_towards(3.1, 3.1), std::nullopt);
 
-  // // Maximum double that can losslessly represented as a float and the next higher double
+  // Maximum double that can losslessly represented as a float and the next higher double
```
Bensk1, line 21: This has nothing to do with plugins or this PR but I found the double comment while searching for that in the code.

benrobby: 👍

src/test/lib/utils/meta_tables/meta_exec_table_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+TEST_F(MetaExecTest, CallNotCallableUserExecutableFunctions) {
+  auto& pm = Hyrise::get().plugin_manager;
+
+  // We have to manually rollback the transaction contexts below because otherwise their destruction would cause an
+  // exeception to be thrown. This is due to an assert in the TransactionContext's destructor checking for failed
+  // operators. See ~TransactionContext for details.
+
+  // Call non-existing function
```
benrobby, line 72: maybe `// Call non-existing plugin (with non-existing function)`?

benrobby (commented):
@Bouncner invited me to also take a look since I'm also writing a plugin for my master thesis right now. When this is merged we could maybe add the functionality to the example plugin (https://github.com/hyrise/example_plugin).

src/lib/utils/plugin_manager.cpp
```diff
@@ -60,21 +66,48 @@ void PluginManager::load_plugin(const std::filesystem::path& path) {
...
+
+  auto& log_manager = Hyrise::get().log_manager;
+  log_manager.add_message(
+      "PluginManager",
+      "Called user executable function `" + function_name + "` provided by plugin `" + plugin_name + "`.",
```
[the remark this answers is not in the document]
Bensk1: First to the general discussion on how to handle long running functions. I talked to @Bouncner about that. We would leave it to the plugin to take care of long running functions and document that, e.g., in the AbstractPlugin.

Making users aware is a good idea. However, users might get a distorted picture if the blocking call of the PluginManager terminates quickly but the actual functionality executed by the plugin itself in a separate thread runs forever. I don't have a strong opinion here but I would not add wall clock information. Maybe we can get a third opinion @Bouncner?

Bensk1:
Thanks for the review, I addressed the points you mentioned and added a comment to AbstractPlugin.

src/lib/utils/plugin_manager.cpp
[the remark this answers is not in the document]
Bouncner: Hmm, I think extending the message by "(function returned after 17 ms)" can't hurt and might help in certain situations.
In case somebody thinks starting a thread in the function is a good idea, this person should be aware that one cannot rely on the runtime in the log message. What users are we talking about, when we talk about "users might get a distorted pictures"? I'd assume we'll never have the situation where the "user" executing the plugin has not been (at partially) involved in developing the plugin. I wouldn't see a problem in this case.

Or did I get you wrong?

src/lib/utils/plugin_manager.cpp
```diff
@@ -60,21 +66,48 @@ void PluginManager::load_plugin(const std::filesystem::path& path) {
...
+  for (const auto& [function_name, function_pointer] : user_executable_functions) {
+    _user_executable_functions[{plugin_name, function_name}] = function_pointer;
+  }
 }
 
-void PluginManager::unload_plugin(const PluginName& name) {
-  auto plugin_iter = _plugins.find(name);
-  Assert(plugin_iter != _plugins.cend(), "Unloading plugin failed: A plugin with name " + name + " does not exist.");
+void PluginManager::exec_user_function(const PluginName& plugin_name, const PluginFunctionName& function_name) {
+  Assert(_user_executable_functions.count({plugin_name, function_name}) == 1,
```
Bouncner, line 79: Minor: since that's the key there cannot be more than 1 and thus I'd favor contains().

src/lib/utils/plugin_manager.cpp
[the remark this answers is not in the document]
Bensk1: I agree that it won't hurt. On the other hand, this is a user-triggered blocking function. Hence, if the function takes forever, the user will notice because the function will not return quickly. If the function takes 17ms the call returns immediately. On the console or via a client, the response typically even contains the execution time.

No, I think you understood me correctly. Plugin users will probably be involved in plugin development. However, plugins might be released as part of research papers where the user was not involved in the development. 

All in all, it's easy to add the timing information and it does not come with much overhead. However, I don't see much benefit of it. In particular because we want the PluginManager's blocking call to always finish really quickly, because the user is usually involved in the development, and we want the heavy lifting to be done inside the plugin itself.

Bouncner:
Can't comment above. I don't know why.

> I agree that it won't hurt. On the other hand, this is a user-triggered blocking function. Hence, if the function takes forever, the user will notice because the function will not return quickly. If the function takes 17ms the call returns immediately. On the console or via a client, the response typically even contains the execution time.

Yeah, makes totally sense. I am fine without timing information. What about you @benrobby ?

benrobby:
> Yeah, makes totally sense. I am fine without timing information. What about you @benrobby ?

Same for me 👍

src/plugins/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ function(add_plugin)
 endfunction(add_plugin)
 
 add_plugin(NAME hyriseMvccDeletePlugin SRCS mvcc_delete_plugin.cpp mvcc_delete_plugin.hpp DEPS sqlparser magic_enum gtest)
+add_plugin(NAME hyriseSecondTestPlugin SRCS second_test_plugin.cpp second_test_plugin.hpp DEPS sqlparser)
```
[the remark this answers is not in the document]
Bensk1: Second one so that we can have better tests: testing name clashes of user callable functions originating from different plugins. 

Second before first because it is ordered alphabetically.

src/test/lib/utils/plugin_manager_test.cpp
```diff
@@ -69,6 +69,83 @@ TEST_F(PluginManagerTest, LoadPluginAutomaticUnload) {
...
+  EXPECT_EQ(plugins.size(), 0u);
+  pm.load_plugin(build_dylib_path("libhyriseTestPlugin"));
+  pm.load_plugin(build_dylib_path("libhyriseSecondTestPlugin"));
+  EXPECT_EQ(plugins.size(), 2u);
+
+  {
+    auto user_executable_functions = pm.user_executable_functions();
+
+    EXPECT_EQ(user_executable_functions.size(), 3);
+    EXPECT_EQ(user_executable_functions.count({"hyriseSecondTestPlugin", "OurFreelyChoosableFunctionName"}), 1);
```
Bouncner, line 85: 3* `.contains()`? clang-tidy (versions 13+) might complain.

Bouncner (approved):
I tiny thing. Otherwise, I thinks it's good to go.

src/test/lib/utils/plugin_manager_test.cpp
Bensk1: Makes sense. I think some time ago I always wanted to use contains but it was unsupported. Now it is supported and I don't use it :)

Bensk1: I fixed some other occurrences along the way.

PR #2402, opened 2021-08-23
Update CI and clang-tidy adaptions

Bouncner:
Changing a few things to get the CI and our Docker image up to date:
 - use Ubuntu 22.04 (for the Docker CI image)
   - with that we now use the following compilers in the CI
     - clang11 and clang 14 (clang11 only a single debug build)
     - GCC 9 and GCC 11 (GCC 9 only a single debug build)
 - many small adaptions to make clang-tidy14 happy (+ exclude several new tidy rules)
 - we use clang11 for the coverage calculation as there is an unresolved problem with clang14 (https://github.com/llvm/llvm-project/issues/54907)

Adresses #1901.

### Todos:
- [x] Reevaluated necessity for `.load()` in some std::atomic cases (see #2439 and `chunk.cpp`/`abstract_task.cpp`)
- [x] Pull out format changes and create upfront PR (#2470)
- [x] Determine where failing ThreadSanitizer test comes from (~the sanitizer does not complain, but we run into timeouts~ probably a sanitizer issues with the timeouts, significantly larger tests in release build work)
- [x] ~Test GCC 12 (should be available in Jammy)~ (see #2473)
- [x] Fix Valgrind issues (we need to use GCC for Valgrind runs, as there is an open bug with Valgrind and Clang 14)

### Review parts:

- [x]  @dey4ss 
```
./src/lib/import_export 5 files changed, 40 insertions(+), 35 deletions(-)
./src/lib/utils 13 files changed, 99 insertions(+), 84 deletions(-)
./src/lib/operators 38 files changed, 204 insertions(+), 176 deletions(-)
```
- [x]  @dey4ss 
```
./src/lib/sql 4 files changed, 80 insertions(+), 61 deletions(-)
./src/lib/memory 1 file changed, 14 insertions(+), 6 deletions(-)
./src/lib/storage 30 files changed, 292 insertions(+), 260 deletions(-)
```
- [x]  @mweisgut 
```
./src/lib/expression 7 files changed, 142 insertions(+), 127 deletions(-)
./src/lib/scheduler 6 files changed, 15 insertions(+), 15 deletions(-)
./src/lib/logical_query_plan 7 files changed, 61 insertions(+), 38 deletions(-)
./src/lib/statistics 7 files changed, 67 insertions(+), 58 deletions(-)
```
- [x]  @mweisgut 
```
./src/lib/server 7 files changed, 39 insertions(+), 35 deletions(-)
[...]

PR #2439, opened 2022-03-09
Make TransactionID and CommitID strong typedefs

Bouncner:
Draft for now to check if Full CI runs.

### DIFF
--- a/src/lib/storage/chunk.cpp
+++ b/src/lib/storage/chunk.cpp
@@ -254,7 +254,7 @@ void Chunk::set_individually_sorted_by(const std::vector<SortColumnDefinition>& 
 }
 
 std::optional<CommitID> Chunk::get_cleanup_commit_id() const {
-  if (_cleanup_commit_id == 0) {
+  if (_cleanup_commit_id == CommitID{0}) {
     // Cleanup-Commit-ID is not yet set
     return std::nullopt;
   }
--- a/src/lib/storage/chunk.hpp
+++ b/src/lib/storage/chunk.hpp
@@ -201,8 +201,8 @@ class Chunk : private Noncopyable {
   mutable std::atomic<ChunkOffset> _invalid_row_count{0};
 
   // Default value of zero means "not set"
-  std::atomic<CommitID> _cleanup_commit_id{0};
-  static_assert(std::is_same<uint32_t, CommitID>::value, "Type of _cleanup_commit_id does not match type of CommitID.");
+  std::atomic<CommitID> _cleanup_commit_id{CommitID{0}};
+  static_assert(std::is_same<uint32_t, CommitID::base_type>::value, "Type of _cleanup_commit_id does not match type of CommitID.");
 };
 
 }  // namespace opossum
--- a/src/lib/types.hpp
+++ b/src/lib/types.hpp
@@ -49,6 +49,11 @@ STRONG_TYPEDEF(uint32_t, ValueID);  // Cannot be larger than ChunkOffset
 STRONG_TYPEDEF(uint32_t, NodeID);
 STRONG_TYPEDEF(uint32_t, CpuID);
 
+// When changing these to 64-bit types, reading and writing to them might not be atomic anymore.
+// Among others, the validate operator might break when another operator is simultaneously writing begin or end CIDs.
+STRONG_TYPEDEF(uint32_t, CommitID);
+STRONG_TYPEDEF(uint32_t, TransactionID);
+
 // Used to identify a Parameter within a subquery. This can be either a parameter of a Prepared SELECT statement
 // `SELECT * FROM t WHERE a > ?` or a correlated parameter in a subquery.
 STRONG_TYPEDEF(uint16_t, ParameterID);
@@ -126,11 +131,6 @@ struct RowID {
 
 using WorkerID = uint32_t;
 using TaskID = uint32_t;
-
-// When changing these to 64-bit types, reading and writing to them might not be atomic anymore.
-// Among others, the validate operator might break when another operator is simultaneously writing begin or end CIDs.
-using CommitID = uint32_t;
-using TransactionID = uint32_t;
 
 using CompressedVectorTypeID = uint8_t;
 
--- a/src/lib/concurrency/transaction_manager.cpp
+++ b/src/lib/concurrency/transaction_manager.cpp
@@ -28,8 +28,8 @@ TransactionManager& TransactionManager::operator=(TransactionManager&& transacti
 CommitID TransactionManager::last_commit_id() const { return _last_commit_id; }
 
 std::shared_ptr<TransactionContext> TransactionManager::new_transaction_context(const AutoCommit auto_commit) {
-  const TransactionID snapshot_commit_id = _last_commit_id;
-  return std::make_shared<TransactionContext>(_next_transaction_id++, snapshot_commit_id, auto_commit);
+  const CommitID snapshot_commit_id = _last_commit_id;
+  return std::make_shared<TransactionContext>(TransactionID{_next_transaction_id++}, snapshot_commit_id, auto_commit);
 }
 
 void TransactionManager::_register_transaction(const CommitID snapshot_commit_id) {
@@ -86,7 +86,7 @@ std::shared_ptr<CommitContext> TransactionManager::_new_commit_context() {
       current_context = std::atomic_load(&_last_commit_context);
     }
 
-    next_context = std::make_shared<CommitContext>(current_context->commit_id() + 1u);
+    next_context = std::make_shared<CommitContext>(CommitID{current_context->commit_id() + 1});
 
     success = current_context->try_set_next(next_context);
 
@@ -108,7 +108,7 @@ void TransactionManager::_try_increment_last_commit_id(const std::shared_ptr<Com
   auto current_context = context;
 
   while (current_context->is_pending()) {
-    auto expected_last_commit_id = current_context->commit_id() - 1;
+    auto expected_last_commit_id = CommitID{current_context->commit_id() - 1};
 
     if (!_last_commit_id.compare_exchange_strong(expected_last_commit_id, current_context->commit_id())) return;
 
--- a/src/lib/concurrency/transaction_manager.hpp
+++ b/src/lib/concurrency/transaction_manager.hpp
@@ -87,7 +87,7 @@ class TransactionManager : public Noncopyable {
   void _register_transaction(CommitID snapshot_commit_id);
   void _deregister_transaction(CommitID snapshot_commit_id);
 
-  std::atomic<TransactionID> _next_transaction_id;
+  std::atomic<TransactionID::base_type> _next_transaction_id;
 
   std::atomic<CommitID> _last_commit_id;
   // We use commit_id=0 for rows that were inserted and then rolled back. Also, this can be used for rows that have
--- a/src/lib/operators/delete.cpp
+++ b/src/lib/operators/delete.cpp
@@ -29,7 +29,8 @@ std::shared_ptr<const Table> Delete::_on_execute(std::shared_ptr<TransactionCont
 
   _transaction_id = context->transaction_id();
 
-  for (ChunkID chunk_id{0}; chunk_id < _referencing_table->chunk_count(); ++chunk_id) {
+  const auto chunk_count = _referencing_table->chunk_count();
+  for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
     const auto chunk = _referencing_table->get_chunk(chunk_id);
 
     DebugAssert(chunk->references_exactly_one_table(),
@@ -64,7 +65,7 @@ std::shared_ptr<const Table> Delete::_on_execute(std::shared_ptr<TransactionCont
             "Trying to delete a row that is not visible to the current transaction. Has the input been validated?");
 
         // Actual row "lock" for delete happens here, making sure that no other transaction can delete this row
-        auto expected = 0u;
+        const auto expected = TransactionID{0};
         const auto success = mvcc_data->compare_exchange_tid(row_id.chunk_offset, expected, _transaction_id);
 
         if (!success) {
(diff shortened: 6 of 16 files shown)


src/lib/types.hpp
```diff
@@ -29,17 +29,13 @@
...
- * WorkerID, TaskID, CommitID, and TransactionID are used in std::atomics and
- * therefore need to be trivially copyable. That's currently not possible with
- * the strong typedef (as far as I know).
- *
- * TODO(anyone): Also, strongly typing ChunkOffset causes a lot of errors in
- * the group key and adaptive radix tree implementations. Unfortunately, I
- * wasn’t able to properly resolve these issues because I am not familiar with
- * the code there
+ * We prefer strong typedefs, whenever they are applicable and make sense. However, there are cases where we cannot
+ * directly use them. For example in std::atomics as they need to be trivially copyable. Therefore, we use the
```
Bouncner, line 37: I have to think about that again. I am not sure the copyable part is what bothers us here, but that we want the specialization for integral types (e.g. `++atomic_var`).

src/lib/operators/limit.cpp
```diff
@@ -62,17 +62,18 @@ std::shared_ptr<const Table> Limit::_on_execute() {
    */
   auto output_chunks = std::vector<std::shared_ptr<Chunk>>{};
 
-  ChunkID chunk_id{0};
+  auto chunk_id = ChunkID{0};
   const auto chunk_count = input_table->chunk_count();
-  for (size_t i = 0; i < num_rows && chunk_id < chunk_count; chunk_id++) {
+  for (size_t i = 0; i < num_rows && chunk_id < chunk_count; ++chunk_id) {
```
mweisgut, line 67: Since you touched the line, could you also change the name of `i`?

src/lib/operators/table_scan/column_between_table_scan_impl.cpp
```diff
@@ -131,10 +131,12 @@ void ColumnBetweenTableScanImpl::_scan_dictionary_segment(
       {}  // clang-format off
       #pragma omp simd
       // clang-format on
-      for (auto chunk_offset = ChunkOffset{0}; chunk_offset < static_cast<ChunkOffset>(output_size); ++chunk_offset) {
+      // OMP pragma does not work with strong type defs.
```
mweisgut, line 134: OMP?

src/lib/operators/table_scan/column_vs_value_table_scan_impl.cpp
```diff
@@ -132,10 +132,12 @@ void ColumnVsValueTableScanImpl::_scan_dictionary_segment(
       {}  // clang-format off
       #pragma omp simd
       // clang-format on
-      for (auto chunk_offset = ChunkOffset{0}; chunk_offset < static_cast<ChunkOffset>(output_size); ++chunk_offset) {
+      // OMP pragma does not work with strong type defs.
```
mweisgut, line 135: Highlighting this line since there is this `OMP` again.

src/lib/storage/chunk.hpp
```diff
@@ -161,7 +161,7 @@ class Chunk : private Noncopyable {
    * However, `size() - invalid_row_count()` does not necessarily tell you how many rows are visible for
    * the current transaction.
    */
-  ChunkOffset invalid_row_count() const { return _invalid_row_count.load(); }
+  ChunkOffset invalid_row_count() const { return ChunkOffset{_invalid_row_count}; }
```
mweisgut, line 164: Since `load` was added at three previous locations, shouldn't we be consistent here and also call `load`?

src/lib/storage/index/group_key/variable_length_key_store.hpp
```diff
@@ -137,7 +137,7 @@ class VariableLengthKeyStore {
 
     void increment() { _data += _key_alignment; }
     void decrement() { _data -= _key_alignment; }
-    void advance(ChunkOffset n) { _data += n * _key_alignment; }
+    void advance(size_t n) { _data += n * _key_alignment; }
```
mweisgut, line 140: Change `n`?

src/lib/storage/run_length_segment.cpp
```diff
@@ -45,8 +45,8 @@ AllTypeVariant RunLengthSegment<T>::operator[](const ChunkOffset chunk_offset) c
 
 template <typename T>
 ChunkOffset RunLengthSegment<T>::size() const {
-  if (_end_positions->empty()) return 0u;
-  return _end_positions->back() + 1u;
+  if (_end_positions->empty()) return ChunkOffset{0};
```
mweisgut, line 48: Prefer braced control statements

mweisgut (changes requested):
Reviewed up to and including `table.hpp`. Next: `types.hpp`.

src/lib/storage/index/group_key/variable_length_key_store.hpp
Bouncner: I think it is one of the cases where we it's super obvious what `n` is about and we are actually extending the boost facade, which uses `n`. I would keep it like that.

src/lib/storage/chunk.hpp
Bouncner: I removed `load()` whenever possible as the direct access is guaranteed to do the same. However, some functions (e.g., `DTRACE` or `to_string`) do not know how to handle atomics. We thus explicitly obtain the underlying integral type.

src/lib/storage/chunk.cpp
```diff
@@ -254,7 +254,8 @@ void Chunk::set_individually_sorted_by(const std::vector<SortColumnDefinition>&
 }
 
 std::optional<CommitID> Chunk::get_cleanup_commit_id() const {
-  if (_cleanup_commit_id == CommitID{0}) {
+  // TODO(Martin): check with PR #2402 whether GCC still needs the `.load()` (some used in abstract_task.cpp twice).
```
mweisgut, line 257: Highlighting it so that this open point is not forgotten in this PR's large diff.

Bouncner: I added a todo in the PR description of #2402.

src/test/lib/storage/reference_segment_test.cpp
```diff
@@ -36,8 +36,11 @@ class ReferenceSegmentTest : public BaseTest {
     TableColumnDefinitions column_definitions2;
     column_definitions2.emplace_back("a", DataType::Int, false);
     column_definitions2.emplace_back("b", DataType::Int, false);
-    _test_table_dict = std::make_shared<opossum::Table>(column_definitions2, TableType::Data, 5, UseMvcc::Yes);
-    for (int i = 0; i <= 24; i += 2) _test_table_dict->append({i, 100 + i});
+    _test_table_dict =
+        std::make_shared<opossum::Table>(column_definitions2, TableType::Data, ChunkOffset{5}, UseMvcc::Yes);
+    for (int i = 0; i <= 24; i += 2) {
```
mweisgut, line 41: Since you touch this line:
- rename `i`
- be explicit with the data type: `uint8_t`

mweisgut (changes requested):

src/test/lib/storage/reference_segment_test.cpp
mweisgut: Right - I should have checked the loop body in more detail :) But we can still be specific and use `int32_t`.

Bouncner: Ah yes, right.

src/test/lib/operators/join_hash/join_hash_steps_test.cpp
```diff
@@ -80,9 +80,9 @@ TEST_F(JoinHashStepsTest, SmallHashTableAllPositions) {
 
 TEST_F(JoinHashStepsTest, LargeHashTableExistenceOnly) {
   auto table = PosHashTable<int>{JoinHashBuildMode::ExistenceOnly, 100};
-  for (auto i = 0; i < 100; ++i) {
-    table.emplace(i, RowID{ChunkID{ChunkID::base_type{100} + i}, ChunkOffset{200} + i});
-    table.emplace(i, RowID{ChunkID{ChunkID::base_type{100} + i}, ChunkOffset{200} + i + 1});
+  for (auto index = uint16_t{0}; index < 100; ++index) {
```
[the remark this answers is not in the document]
Bouncner: Now that you bring that up, I would even vote for `uint32_t`.
First, `uint8_t` or `uint16_t` do not bring any advantage in memory consumption or performance (I think they are even slower on most CPUs here). I do not remember why I used `uint16_t`. :-/

So, I'd vote for `uint32_t`. Your take on this?

src/lib/types.hpp
```diff
@@ -48,6 +45,14 @@ STRONG_TYPEDEF(opossum::ColumnID::base_type, ColumnCount);
 STRONG_TYPEDEF(uint32_t, ValueID);  // Cannot be larger than ChunkOffset
 STRONG_TYPEDEF(uint32_t, NodeID);
 STRONG_TYPEDEF(uint32_t, CpuID);
+STRONG_TYPEDEF(uint32_t, WorkerID);
+STRONG_TYPEDEF(uint32_t, TaskID);
+STRONG_TYPEDEF(uint32_t, ChunkOffset);
+
+// When changing these to 64-bit types, reading and writing to them might not be atomic anymore.
```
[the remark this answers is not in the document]
Bouncner: I don't know if we should keep this comment. Every ARM starting with ARMv8-1a should support the necessary 64bit atomics. The M1 is covered, it appears to use at least the ARM v8.4-A instruction set.
BUT even the most recent Raspberry Pi versions do "only" have a ARM v8.0 instruction set. So here we would actually not run into problems as the atomics cover proper accesses, but 64-bit values wouldn't be as efficient as 32-bit ones here.

Please note that my hardware understanding is quite limited.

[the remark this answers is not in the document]
Bouncner: What do you think about a comment like:
```
When changing these to 64-bit types, please be aware that not all platforms that Hyrise runs on support atomic 64-bit  instructions. Any Intel and AMD CPU of the last ten years should work fine. But 64-bit atomics on ARM CPUs might lead to less efficient code when using `std::atomic` on instructions sets earlier than ARMv8-1A (e.g., Raspberry Pi v3 or Amazon's Graviton1). Apple's M1 and Graviton2 support the required LSE instructions.
More information can be found here: https://community.arm.com/arm-community-blogs/b/tools-software-ides-blog/posts/making-the-most-of-the-arm-architecture-in-gcc-10
```

src/test/lib/operators/join_hash/join_hash_steps_test.cpp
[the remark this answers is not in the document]
mweisgut: > I think they are even slower on most CPUs here

Why?

In the following two lines, `index` is implicitly converted into `uint32_t`, which might be another argument to go for `uint32_t`.

src/lib/types.hpp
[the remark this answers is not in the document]
mweisgut: I would shorten the comment a bit by removing the part `(e.g.,
Raspberry Pi v3 or Amazon's Graviton1). Apple's M1 and Graviton2 support the required LSE instructions`.

src/test/lib/operators/join_hash/join_hash_steps_test.cpp
```diff
@@ -52,18 +52,18 @@ class JoinHashStepsTest : public BaseTest {
...
-  inline static size_t _chunk_size_zero_one = 10;
+  inline static auto _table_size_zero_one = ChunkOffset{1'000};
+  inline static auto _chunk_size_zero_one = ChunkOffset{10};
   inline static std::shared_ptr<Table> _table_zero_one;
   inline static std::shared_ptr<TableWrapper> _table_int_with_nulls, _table_with_nulls_and_zeros;
   inline static std::shared_ptr<TableScan> _table_with_nulls_and_zeros_scanned;
 };
 
 TEST_F(JoinHashStepsTest, SmallHashTableAllPositions) {
   auto table = PosHashTable<int>{JoinHashBuildMode::AllPositions, 50};
```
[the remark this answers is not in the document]
Bouncner: Changed to iter64_t, which makes it easy to put `uint32_t`'s in it.

src/test/lib/operators/join_hash/join_hash_steps_test.cpp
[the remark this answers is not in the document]
Bouncner: > Why?

Because most CPUs work the most efficient when operating on their native word size, which is 64bit for x64 (but I think it's just as fast for 32-bit).

src/lib/types.hpp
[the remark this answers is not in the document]
Bouncner: Hmm, took me quite some time to understand the different ARM specs and which CPU supports which instruction set. Why would you remove it?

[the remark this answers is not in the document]
mweisgut: Only for reference:
From my current understanding, x86 CPUs support atomic 64-bit operations via compare and exchange (CMPXCHG) instructions (https://en.wikipedia.org/wiki/Compare-and-swap#Implementations).

>Hmm, took me quite some time to understand the different ARM specs and which CPU supports which instruction set. Why would you remove it?

I think the key information regarding ARM CPUs is that on instruction sets earlier than ARMv8.1-A, atomic operations might be less efficient.
The Raspberry Pi v3 CPU and Graviton1 are then "only" CPU examples with earlier instruction sets, whereas M1 and Graviton2 are CPU examples with later instruction sets.

- Graviton1: https://en.wikichip.org/wiki/annapurna_labs/alpine/al73400
- Raspberry Pi v3 CPU: https://de.wikipedia.org/wiki/Arm_Cortex-A#Arm_Cortex-A53
- M1: https://en.wikichip.org/wiki/apple/mx/m1
- Graviton2: https://en.wikipedia.org/wiki/ARM_Cortex-A76#Neoverse_N1

While the instruction set version is an important note, the CPUs are only examples that may not be relevant in a few years and the comment may need to be updated. If a Hyrise user/developer wants to check for a target ARM chip to see if there are any inefficiencies in atomic operations, they can look up the instruction set for that chip IMO.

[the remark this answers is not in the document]
mweisgut: `ARMv8-1A` --> `ARMv8.1-A`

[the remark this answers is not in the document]
Bouncner: Convinced.

What do you think about:
```
When changing these to 64-bit types, please be aware that not all platforms that Hyrise runs on support atomic
64-bit instructions. Any Intel and AMD CPU since 2010 should work fine. For 64-bit atomics on ARM
CPUs, the instruction set should be at least ARMv8.1-A. Earlier instruction sets also work, but might yield less efficient code.
More information can be found here:
https://community.arm.com/arm-community-blogs/b/tools-software-ides-blog/posts/making-the-most-of-the-arm-architecture-in-gcc-10
```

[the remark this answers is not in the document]
mweisgut: Looks good to me

mweisgut (approved):

PR #2443, opened 2022-03-18
Add maintainer

mweisgut:


Bensk1 (approved):

README.md
```diff
@@ -149,6 +149,7 @@ When trying to optimize the time spent building the project, it is often helpful
 ## Maintainers
 - Jan Kossmann
 - Martin Boissier
+- Marcel Weisgut
```
Bouncner, line 152: Do we have an order here? Looked like firstname. In this case you should be second.

Bouncner (approved):

PR #2449, opened 2022-04-14
Jenkinsfile: Add root repository and submodule repositories as safe directories

mweisgut:
After the security vulnerability [CVE-2022-24765](https://github.blog/2022-04-12-git-security-vulnerability-announced/), the latest git version available in ubuntu 20.04 was [patched](https://launchpad.net/ubuntu/+source/git/1:2.25.1-1ubuntu3.3).

With this patch, git checks the owner of the top-level directory and does not perform a git command if the owner is not the current user. This check can be disabled for a repository by adding it as a safe directory, i.e., `git config --global --add safe.directory <dir>`.

In a current CI run, the user is `root`, but the files and directories in the cloned Hyrise project directory have the owner `113`.

This PR adds the directory of the root repository and the directories of the submodule repositories as safe directories to avoid CI errors of the following kind:

```
fatal: unsafe repository ('/var/lib/jenkins/workspace/hyrise_hyrise_PR-2448' is owned by someone else)
```

References
- https://github.com/actions/checkout/issues/760
- https://github.com/gaurav-nelson/github-action-markdown-link-check/issues/132
- https://stackoverflow.com/questions/71849415/cannot-add-parent-directory-to-safe-directory-on-git
- https://github.blog/2022-04-12-git-security-vulnerability-announced/
- git repo, commit adding the safe.directory check: https://github.com/git/git/commit/0f85c4a30b072a26d74af8bbf63cc8f6a5dfc1b8
- git repo, commit enabling the safe.directory check: https://github.com/git/git/commit/bb50ec3cc300eeff3aba7a2bea145aabdb477d31

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -66,12 +66,9 @@ try {
           stage("Setup") {
             checkout scm
 
-            sh "echo $WORKSPACE"
-            sh "echo \$(pwd)"
-            // pwd = sh "pwd"
-            // sh "git config --global --add safe.directory $pwd"
-            // // Get the paths of the submodules; for each path, add it as a git safe.directory
-            // sh "grep path .gitmodules | sed 's/.*=//' | xargs -n 1 -I '{}' git config --global --add safe.directory $pwd/'{}'"
+            sh "git config --global --add safe.directory $WORKSPACE"
+            // Get the paths of the submodules; for each path, add it as a git safe.directory
+            sh "grep path .gitmodules | sed 's/.*=//' | xargs -n 1 -I '{}' git config --global --add safe.directory $WORKSPACE/'{}'"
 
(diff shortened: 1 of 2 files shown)


mweisgut:
The Jenkinsfile might be the better place for the fix.

Bouncner (approved):
So we cannot do anything against having a different user than the dir owner here?

Have to trust you here. 🙂

mweisgut:
> So we cannot do anything against having a different user than the dir owner here?
> 
> Have to trust you here. 🙂

I'm sure there is something we can do against having a different user than the dir owner. However, so far, I do not know how to achieve it. Since this problem is currently blocking CI runs, I would use this workaround now to be able to perform CI runs again. The user-owner discrepancy is captured in #2450.

PR #2456, opened 2022-04-25
Refactor EnableMakeForLQPNode

julianmenzler:
Improve code readability & Doc.

### DIFF
--- a/src/lib/logical_query_plan/enable_make_for_lqp_node.hpp
+++ b/src/lib/logical_query_plan/enable_make_for_lqp_node.hpp
@@ -25,53 +25,54 @@ class AbstractLQPNode;
 template <typename DerivedNode>
 class EnableMakeForLQPNode {
  public:
-  template <int N, typename... Ts>
-  using NthTypeOf = std::tuple_element_t<N, std::tuple<Ts...>>;
+  // The following declaration is a variadic function template taking any number of arguments of arbitrary types.
+  template <typename... ArgumentTypes>
+  static std::shared_ptr<DerivedNode> Make(ArgumentTypes&&... arguments) {
+    if constexpr (sizeof...(ArgumentTypes) > 0) {
+      auto arguments_tuple = std::forward_as_tuple(arguments...);
 
-  template <typename... Args>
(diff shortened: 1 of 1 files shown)


src/lib/logical_query_plan/enable_make_for_lqp_node.hpp
```diff
@@ -10,68 +10,75 @@ namespace opossum {
...
 
  private:
-  template <class Tuple, size_t... I>
-  static std::shared_ptr<DerivedNode> make_impl(const Tuple& constructor_arguments,
-                                                std::index_sequence<I...> num_constructor_args) {
-    return std::make_shared<DerivedNode>(std::get<I>(constructor_arguments)...);
+  template <class ArgumentsTupleType, size_t... ConstructorIndices>
+  static std::shared_ptr<DerivedNode> create_lqp_node(const ArgumentsTupleType& arguments_tuple,
+                                                      std::index_sequence<ConstructorIndices...> /* indices */) {
+    return std::make_shared<DerivedNode>(std::get<ConstructorIndices>(arguments_tuple)...);
```
Bouncner, line 76: I don't get it. I haven't found a std::get version that accepts a sequence. How does that exactly work here?

Bouncner (approved):
Nice work. Even though I still have a hard time following the code, it's much better to follow now in my opinion.

src/lib/logical_query_plan/enable_make_for_lqp_node.hpp
julianmenzler: True, this compiler magic is not very intuitive. My understanding:

* `ConstructorIndices` represents a non-type template parameter pack. For example the values `0, 1, 2`
* The `...` in `std::get<ConstructorIndices>(arguments_tuple)...` triggers parameter pack expansion, which is described as follows: 

  > A pattern followed by an ellipsis, in which the name of at least one parameter pack appears at least once, is expanded into zero or more comma-separated instantiations of the pattern, where the name of the parameter pack is replaced by each of the elements from the pack, in order

PR #2457, opened 2022-04-26
Remove ChunkCompressionTask

Bouncner:
This PR removes the `ChunkCompressionTask` and the tests of it.

The task has been dead code and wasn't used. It lacked support for encoding with anything else than dictionary encoding.
According to the comments, the task was supposed to facilitate concurrent encoding, but it neither solved the problem it describes and even states that the current implementation is UB.

The issue that concurrent modifications to segments/chunks can be problematic is still true. But this task did not help.

There is one thing, that I am not sure about. It's about zeroing `begin_cid` on rollback. I am not sure if we still need to do it (see removed comment in insert.cpp). @julianmenzler: as you have worked on MVCC topics, can you comment on that?


src/lib/operators/insert.cpp
```diff
@@ -251,8 +251,6 @@ void Insert::_on_rollback_records() {
      * foreign tid - which is what a visible row that is being deleted by a different transaction looks like. Thus,
      * the other transaction would consider the row (that is in the process of being rolled back and should have never
      * been visible) as visible.
-     *
-     * We need to set `begin_cid = 0` so that the ChunkCompressionTask can identify "completed" Chunks.
      */
 
     for (auto chunk_offset = target_chunk_range.begin_chunk_offset; chunk_offset < target_chunk_range.end_chunk_offset;
```
julianmenzler, line 256: We should be able to merge the following two for-loops into one.

julianmenzler (commented):
> There is one thing, that I am not sure about. It's about zeroing begin_cid on rollback. I am not sure if we still need to do it (see removed comment in insert.cpp). @julianmenzler: as you have worked on MVCC topics, can you comment on that?

Having had a look, I do not think so. Keeping `begin_cid = MAX_COMMIT_ID` after a rollback should be fine.

src/lib/operators/insert.cpp
Bouncner: I lack the understanding to judge on that. I guess we have tests in place that would recognize if we cannot merge the loops?

Bouncner: Hmm, not sure about these fences though. :-/

julianmenzler: After removing `begin_cid = 0` and merging both for-loops, the first fence should become irrelevant because `begin_cid = MAX_COMMIT_ID` is always true.

From my understanding, this leaves the possibility that other threads can see rows as follows:

* `end_cid = 0` & `tid = 0` (desired result after rollback)
* `end_cid = MAX_COMMIT_ID` & `tid = 0`
* `end_cid = 0` & `tid = CurrentTransactionID`
* `end_cid = MAX_COMMIT_ID` & `tid = CurrentTransactionID`


These options do not look critical to me since other transactions should not see these rows anyway (thanks to `begin_cid = MAX_COMMIT_ID`).

I'm not sure if we need the last memory fence though.

julianmenzler: CI passes. Do you see any issues @mrks?

mrks: From the first look, this should be fine. I'll have to find more than just a free minute to think about it though.

Bouncner:
Closing for now until we have a clear way to approach issue #2463. Until then, this task can be seen as a kind of documentation.

PR #2453, opened 2022-04-22
Replace obsolete `tbb:zero_allocator` by `ZeroAllocator`

mweisgut:
As discussed in #2419, oneTBB (former `tbb`) does not contain the `zero_allocator` anymore, which we use in Hyrise's `Table` class to allocate zero-initialized chunk pointers.

This PR
- adds a simple `ZeroAllocator` class implementation, which is derived from `tbb::cache_aligned_allocator` (required for `tbb::concurrent_vector`)
- removes the `tbb::zero_allocator` as template parameter for the `tbb::concurrent_vector` (fixes #2419) and uses the newly implemented `ZeroAllocator` instead

TODOs
- [x] create an issue about stress tests and data races (see #2454)

PR #2458, opened 2022-04-26
Support LLVM 13

Bouncner:
Since #2402 might take a while, this PR adds support for LLVM 13.

Together with #2453, this should finally clean up the brew setup for DYOD students using MacOS.

Changes:
- use compact_vector as a system library (Clang complains about deprecated code)
- disable unity builds for GCC (they trigger a GoogleTest problem with GCC)
- use Clang 13 in Mac CI runners
- disable `reserved-identifier` warning which Clang 13 is super picky about and Clang 14 no longer (the warning is only deactivated for clang 13)
- Update GoogleTest module and update Valgrind suppressions

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -26,7 +26,7 @@ option(ENABLE_COVERAGE "Set to ON to build Hyrise with enabled coverage checking
 if (${ENABLE_COVERAGE})
     add_compile_options(-O0 -fno-inline)
 
-    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
         add_compile_options(-fprofile-arcs -ftest-coverage)
         set(CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} --coverage")
     elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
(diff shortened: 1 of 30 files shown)


Bouncner:
Ready to be reviewed. The Valgrind issue took me way to long. :-/

dey4ss:
Can we remove the [note in the readme](https://github.com/hyrise/hyrise/blob/38527047ba5d11906fa933aac6350bf659425dcd/README.md?plain=1#L78) with this PR?

> **Note about LLVM 13:** Hyrise can currently not be built with LLVM 13. We hope to get LLVM 13 running soon. On MacOS with brew, LLVM 12 can be installed as follows: `brew install llvm@12`. Keep in mind that this package version is an alternate version and, thus, not symlinked into `/usr/local`. `brew link llvm@12` symlinks the package.

src/CMakeLists.txt
```diff
@@ -60,6 +60,13 @@ else()
     message(WARNING "Hyrise will be built with most compiler warnings deactivated. This is fine if you want to test Hyrise but will become an issue when you want to contribute code.")
 endif()
 
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+     if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 13 AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS_EQUAL 13.99)
+    message(WARNING "Disabling 'reserved-identifier' warning for clang 13 (version is ${CMAKE_CXX_COMPILER_VERSION}).")
```
mweisgut, line 65: This line should be more indented than the previous line and on the same indention level as the following line.

mweisgut (changes requested):
Minor linter errors :)

src/CMakeLists.txt
Bouncner: Wow, that looked terrible. No idea how that happened.

Fixed.

mweisgut (approved):

PR #2460, opened 2022-05-06
Support BIGINT Numeric Type

dey4ss:
This PR adds `BIGINT` as an alias for `LONG`. `BIGINT` column types in `CREATE TABLE` statements or `CAST`statements are translated to `DataType::Long`, as the two types are equal according to Postgres [1] (8 byte integer). 

[1] https://www.postgresql.org/docs/14/datatype-numeric.html

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -47,12 +47,13 @@
 [submodule "third_party/robin-map"]
 	path = third_party/robin-map
 	url = https://github.com/Tessil/robin-map.git
-[submodule "third_party/sql-parser"]
-	path = third_party/sql-parser
-	url = https://github.com/hyrise/sql-parser.git
 [submodule "third_party/tpcds-kit"]
 	path = third_party/tpcds-kit
 	url = https://github.com/hyrise-mp/tpcds-kit.git
 [submodule "third_party/zstd"]
 	path = third_party/zstd
 	url = https://github.com/facebook/zstd.git
+[submodule "third_party/sql-parser"]
+	path = third_party/sql-parser
+	url = https://github.com/hyrise/sql-parser.git
+	branch = dey4ss/bigint
--- a/.gitmodules
+++ b/.gitmodules
@@ -47,13 +47,12 @@
 [submodule "third_party/robin-map"]
 	path = third_party/robin-map
 	url = https://github.com/Tessil/robin-map.git
+[submodule "third_party/sql-parser"]
+	path = third_party/sql-parser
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -47,12 +47,13 @@
...
-[submodule "third_party/sql-parser"]
-	path = third_party/sql-parser
-	url = https://github.com/hyrise/sql-parser.git
 [submodule "third_party/tpcds-kit"]
 	path = third_party/tpcds-kit
 	url = https://github.com/hyrise-mp/tpcds-kit.git
 [submodule "third_party/zstd"]
 	path = third_party/zstd
 	url = https://github.com/facebook/zstd.git
+[submodule "third_party/sql-parser"]
```
Bouncner, line 56: Why the change? I think the module have been sorted alphabetically before.

dey4ss: I only changed the submodules using GIT CLI to include the parser's feature branch with the `BIGINT` type, the submodules will be reverted once this feature branch is merged into the parser's master.

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -3125,6 +3127,17 @@ TEST_F(SQLTranslatorTest, CastStatement) {
...
     EXPECT_LQP_EQ(actual_lqp, expected_lqp);
   }
+  {
+    // BIGINT is an alias for LONG
+    const auto cast_expression = expression_vector(cast_(value_(int32_t{1}), DataType::Long));
+    // clang-format off
+    const auto expected_lqp =
+      ProjectionNode::make(cast_expression,
+        DummyTableNode::make());
+    // clang-format on
```
julianmenzler, line 3137:
```suggestion
    const auto expected_lqp =
    ProjectionNode::make(cast_expression,
      DummyTableNode::make());
    // clang-format on
```

julianmenzler (approved):

src/test/lib/sql/sql_translator_test.cpp
dey4ss: Hm, we are inconsistently regarding the spacing in this file. 
`\s{2}const auto expected_lqp.+\n[ ]{4}` --> 30 matches
`\s{2}const auto expected_lqp.+\n[ ]{2}` --> 143 matches

However, as mostly the unindented version is used, I'm going to change all occurrences according to your suggested change.

julianmenzler (approved):
Great!

PR #2312, opened 2021-01-27
Semi Join Reduction Removal

mrks:
Quite WIP. Turns out that while semi join reductions are useful, we do not use all of them. The TPC-H and TPC-DS results are quite promising, but in a handful of cases, we murder the JOB performance.

- [ ] Verify benchmark results. Some of these numbers look *too* good.

**System**
<details>
<summary>pella - click to expand</summary>

| property | value |
| -- | -- |
| Hostname | pella |
| CPU | Intel(R) Xeon(R) CPU E7- 8870  @ 2.40GHz |
| Memory | 1.4TB |
| numactl | nodebind: 2  |
| numactl | membind: 2  |
</details>

**Commit Info and Build Time**
| commit | date | message | build time |
| -- | -- | -- | -- |
| 8cc653b85 | 11.06.2021 11:51 | Add key constraints for the JoinOrder benchmark (#2372) | real 755.44 user 4404.08 sys 181.59|
| fbdd00884 | 11.06.2021 14:53 | Merge remote-tracking branch 'origin/master' into mrks/semi_join_removal | real 748.24 user 4285.88 sys 176.30|

**hyriseBenchmarkTPCH - single-threaded, SF 10.0**
<details>
<summary>
Sum of avg. item runtimes: -3%
 || 
Geometric mean of throughput changes: +4%
</summary>

<details>
<summary>Configuration Overview - click to expand</summary>

```diff
 +Configuration Overview----+--------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------+
 | Parameter                | /home/Julian.Menzler/repo_hyrise2/cmake_release/benchmark_all_results/hyriseBenchmarkTPCH_8cc653b85f66b7f10c0dcb829f02e91e84412aa5_st.json | /home/Julian.Menzler/repo_hyrise2/cmake_release/benchmark_all_results/hyriseBenchmarkTPCH_fbdd00884d7ca013fc0125c856161be717b8db09_st.json |
 +--------------------------+--------------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------------------------------+
 |  GIT-HASH                | 8cc653b85f66b7f10c0dcb829f02e91e84412aa5-dirty                                                                                             | fbdd00884d7ca013fc0125c856161be717b8db09-dirty                                                                                             |
 |  benchmark_mode          | Ordered                                                                                                                                    | Ordered                                                                                                                                    |
[...]

PR #2465, opened 2022-05-12
JoinNode: Add semi reduction flag & reduced join reference as properties

julianmenzler:
This PR prepares for

* #2312
* #2467 

It adds the following functions/properties to `JoinNode`, which become relevant in case of `JoinMode::Semi`:

* `::is_semi_reduction`
* `::mark_as_semi_reduction`
* `::get_or_find_reduced_join_node`

### DIFF
--- a/src/lib/logical_query_plan/join_node.cpp
+++ b/src/lib/logical_query_plan/join_node.cpp
@@ -230,35 +230,31 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
 
 const std::vector<std::shared_ptr<AbstractExpression>>& JoinNode::join_predicates() const { return node_expressions; }
 
-void JoinNode::mark_as_semi_reduction_for(const std::shared_ptr<JoinNode>& corresponding_join_node) {
-  Assert(corresponding_join_node, "Corresponding JoinNode must be provided.");
+void JoinNode::mark_as_semi_reduction(const std::shared_ptr<JoinNode>& reduced_join_node) {
   Assert(!_is_semi_reduction, "The semi reducer status should be set once only.");
+  Assert(reduced_join_node, "Reduced JoinNode must be provided.");
   Assert(join_mode == JoinMode::Semi, "Semi join reductions require JoinMode::Semi.");
   DebugAssert(join_predicates().size() == 1,
               "Currently, semi join reductions are expected to have a single join predicate.");
-  DebugAssert(
-      std::any_of(corresponding_join_node->join_predicates().begin(), corresponding_join_node->join_predicates().end(),
-                  [&](const auto predicate) { return *predicate == *join_predicates()[0]; }),
-      "Did not find matching join predicate in given corresponding JoinNode.");
+  DebugAssert(std::any_of(reduced_join_node->join_predicates().begin(), reduced_join_node->join_predicates().end(),
+                          [&](const auto predicate) { return *predicate == *join_predicates()[0]; }),
+              "Both semi join reduction node and the reduced join should have a common join predicate.");
   _is_semi_reduction = true;
-  _semi_reduction_corresponding_join_node = std::weak_ptr<JoinNode>(corresponding_join_node);
+  _reduced_join_node = std::weak_ptr<JoinNode>(reduced_join_node);
 }
 
 bool JoinNode::is_semi_reduction() const { return _is_semi_reduction; }
 
-std::shared_ptr<JoinNode> JoinNode::get_or_find_semi_reduction_corresponding_join_node() const {
+std::shared_ptr<JoinNode> JoinNode::get_or_find_reduced_join_node() const {
   if (!_is_semi_reduction) {
     return nullptr;
   }
 
-  if (_semi_reduction_corresponding_join_node.expired()) {
-    // The weak pointer is expired in deep copies of the LQP, for example. In such cases, find the corresponding join by
-    // traversing the LQP upwards.
+  if (_reduced_join_node.expired()) {
+    // In deep copies of the LQP, the weak pointer to the reduced join is expired (lazy discovery). In such cases,
+    // find the reduced join by traversing the LQP upwards.
     visit_lqp_upwards(std::const_pointer_cast<AbstractLQPNode>(shared_from_this()), [&](const auto& current_node) {
-      if (current_node.get() == this) {
-        return LQPUpwardVisitation::VisitOutputs;
-      }
-      if (current_node->type != LQPNodeType::Join) {
+      if (current_node->type != LQPNodeType::Join || current_node.get() == this) {
         return LQPUpwardVisitation::VisitOutputs;
       }
       const auto join_node = std::static_pointer_cast<JoinNode>(current_node);
@@ -267,15 +263,14 @@ std::shared_ptr<JoinNode> JoinNode::get_or_find_semi_reduction_corresponding_joi
         return LQPUpwardVisitation::VisitOutputs;
       }
 
-      _semi_reduction_corresponding_join_node = std::weak_ptr<JoinNode>(join_node);
+      _reduced_join_node = std::weak_ptr<JoinNode>(join_node);
       return LQPUpwardVisitation::DoNotVisitOutputs;
     });
 
-    Assert(!_semi_reduction_corresponding_join_node.expired(),
-           "Could not find corresponding join node for this semi join reduction.");
+    Assert(!_reduced_join_node.expired(), "Could not find JoinNode that gets reduced by this semi join reduction.");
   }
 
-  return _semi_reduction_corresponding_join_node.lock();
+  return _reduced_join_node.lock();
 }
 
 size_t JoinNode::_on_shallow_hash() const { return boost::hash_value(join_mode); }
@@ -293,8 +288,8 @@ std::shared_ptr<AbstractLQPNode> JoinNode::_on_shallow_copy(LQPNodeMapping& node
 
 bool JoinNode::_on_shallow_equals(const AbstractLQPNode& rhs, const LQPNodeMapping& node_mapping) const {
   const auto& join_node = static_cast<const JoinNode&>(rhs);
-  // We do not consider the _is_semi_reduction attribute in this comparison, because we want the LQPTranslator to translate
-  // identical semi joins into a single operator, regardless of the `is_semi_reduction` property.
+  // We do not consider the _is_semi_reduction attribute in this comparison, because we want the LQPTranslator to
+  // translate identical semi joins into a single operator, regardless of the `is_semi_reduction` property.
   if (join_mode != join_node.join_mode) {
     return false;
   }
--- a/src/lib/logical_query_plan/join_node.cpp
+++ b/src/lib/logical_query_plan/join_node.cpp
@@ -228,14 +228,55 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
 
 const std::vector<std::shared_ptr<AbstractExpression>>& JoinNode::join_predicates() const { return node_expressions; }
 
+void JoinNode::mark_as_reducer_of(std::shared_ptr<JoinNode> corresponding_join_node) {
+  Assert(corresponding_join_node, "Corresponding JoinNode must be provided.");
+  Assert(!_is_reducer, "Function is meant to be called once only.");
+  DebugAssert(join_mode == JoinMode::Semi, "Semi join reductions require JoinMode::Semi.");
+  DebugAssert(join_predicates().size() == 1, "Semi join reductions are expected to have one join predicate only.");
+  DebugAssert(
+      std::any_of(corresponding_join_node->join_predicates().begin(), corresponding_join_node->join_predicates().end(),
+                  [&](const auto predicate) { return *predicate == *join_predicates()[0]; }),
+      "Did not find matching join predicate in given corresponding JoinNode.");
+  _is_reducer = true;
+  comment = "Semi Reduction";
+  _corresponding_join_node = std::weak_ptr<JoinNode>(corresponding_join_node);
+}
+
+bool JoinNode::is_reducer() const { return _is_reducer; }
+
+std::shared_ptr<JoinNode> JoinNode::get_or_find_corresponding_join_node() const {
+  if (!_is_reducer) return nullptr;
+
+  if (_corresponding_join_node.expired()) {
+    visit_lqp_upwards(const_cast<JoinNode*>(this)->shared_from_this(), [&](const auto& current_node) {
+      if (current_node.get() == this) return LQPUpwardVisitation::VisitOutputs;
+      if (current_node->type != LQPNodeType::Join) return LQPUpwardVisitation::VisitOutputs;
+      const auto join_node = std::static_pointer_cast<JoinNode>(current_node);
+      // Currently, semi reductions are supported for single predicate joins only.
+      if (join_node->join_predicates().size() != 1) return LQPUpwardVisitation::VisitOutputs;
+      if (*join_predicates()[0] != *join_node->join_predicates()[0]) return LQPUpwardVisitation::VisitOutputs;
+
+      _corresponding_join_node = std::weak_ptr<JoinNode>(join_node);
+      return LQPUpwardVisitation::DoNotVisitOutputs;
+    });
+    // TODO make if?
+    Assert(!_corresponding_join_node.expired(), "Could not find corresponding join node.");
+  }
+
+  return _corresponding_join_node.lock();
+}
+
 size_t JoinNode::_on_shallow_hash() const { return boost::hash_value(join_mode); }
 
 std::shared_ptr<AbstractLQPNode> JoinNode::_on_shallow_copy(LQPNodeMapping& node_mapping) const {
-  if (!join_predicates().empty()) {
-    return JoinNode::make(join_mode, expressions_copy_and_adapt_to_different_lqp(join_predicates(), node_mapping));
-  } else {
+  if (join_predicates().empty()) {
+    Assert(join_mode == JoinMode::Cross, "Expected cross join.");
     return JoinNode::make(join_mode);
   }
+  const auto copied_join_node =
+      JoinNode::make(join_mode, expressions_copy_and_adapt_to_different_lqp(join_predicates(), node_mapping));
+  copied_join_node->_is_reducer = _is_reducer;
+  return copied_join_node;
 }
 
 bool JoinNode::_on_shallow_equals(const AbstractLQPNode& rhs, const LQPNodeMapping& node_mapping) const {
--- a/src/lib/logical_query_plan/join_node.cpp
+++ b/src/lib/logical_query_plan/join_node.cpp
@@ -252,7 +252,8 @@ std::shared_ptr<JoinNode> JoinNode::get_or_find_semi_reduction_corresponding_joi
   }
 
   if (_semi_reduction_corresponding_join_node.expired()) {
-    // Find corresponding join by traversing upwards
+    // The weak pointer is expired in deep copies of the LQP, for example. In such cases, find the corresponding join by
+    // traversing the LQP upwards.
     visit_lqp_upwards(std::const_pointer_cast<AbstractLQPNode>(shared_from_this()), [&](const auto& current_node) {
       if (current_node.get() == this) {
         return LQPUpwardVisitation::VisitOutputs;
--- a/src/lib/logical_query_plan/join_node.cpp
+++ b/src/lib/logical_query_plan/join_node.cpp
@@ -231,12 +231,12 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
 const std::vector<std::shared_ptr<AbstractExpression>>& JoinNode::join_predicates() const { return node_expressions; }
 
 void JoinNode::mark_as_semi_reduction(const std::shared_ptr<JoinNode>& reduced_join_node) {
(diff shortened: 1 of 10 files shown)


src/lib/logical_query_plan/join_node.hpp
```diff
@@ -52,9 +52,33 @@ class JoinNode : public EnableMakeForLQPNode<JoinNode>, public AbstractLQPNode {
...
+   * @param corresponding_join_node having the same join predicate.
+   * Note: This function is meant to be called by the SemiJoinReductionRule, which adds semi join reductions to LQPs.
+   */
+  void mark_as_reducer_of(const std::shared_ptr<JoinNode>& corresponding_join_node);
+
   JoinMode join_mode;
 
  protected:
+  bool _is_reducer = false;
+  mutable std::weak_ptr<JoinNode> _corresponding_join_node;
```
Bouncner, line 80: Can we find a better that notes that states that we have a "normal" join node  and a pre-filtering semi join node which we artificially add? Maybe `_corresponding_semi_join_reducer_node`. That might make it clearer that this properties might not be used for "normal" join nodes.

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -228,18 +228,62 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
+                  [&](const auto predicate) { return *predicate == *join_predicates()[0]; }),
+      "Did not find matching join predicate in given corresponding JoinNode.");
+  _is_reducer = true;
+  _corresponding_join_node = std::weak_ptr<JoinNode>(corresponding_join_node);
+}
+
+bool JoinNode::is_reducer() const { return _is_reducer; }
+
+std::shared_ptr<JoinNode> JoinNode::get_or_find_corresponding_join_node() const {
+  if (!_is_reducer) return nullptr;
```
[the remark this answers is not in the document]
Bouncner: Same below.

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -228,18 +228,62 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
+
+  if (_corresponding_join_node.expired()) {
```
Bouncner, line 250: Comment, when it might be expired.

src/lib/logical_query_plan/join_node.hpp
```diff
@@ -52,9 +52,33 @@ class JoinNode : public EnableMakeForLQPNode<JoinNode>, public AbstractLQPNode {
 
   const std::vector<std::shared_ptr<AbstractExpression>>& join_predicates() const;
 
+  /**
+   * @returns true if the JoinNode has as a semi reduction role in the LQP. If so, it was most likely created by the
```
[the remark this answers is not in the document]
julianmenzler: Currently, there are no others. Shall I change the Doc to the following?

```
@returns true if the JoinNode has a semi reduction role in the LQP, and was added by the SemiJoinReductionRule.
```

src/lib/operators/abstract_join_operator.cpp
```diff
@@ -72,6 +73,14 @@ std::string AbstractJoinOperator::description(DescriptionMode description_mode)
     stream << column_name(false, secondary_predicate.column_ids.second);
   }
 
+  // Add comment about semi join reduction if known.
+  if (_mode == JoinMode::Semi && lqp_node) {
+    const auto semi_join_node = std::dynamic_pointer_cast<const JoinNode>(lqp_node);
+    if (semi_join_node->is_reducer()) {
+      stream << separator << "Semi Reduction";
```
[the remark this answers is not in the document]
julianmenzler: We could also remove this logic, and just set the comment field like before.

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -585,4 +594,30 @@ TEST_F(JoinNodeTest, UniqueConstraintsCrossJoin) {
...
+    _mock_node_b);
+  // clang-format on
+  const auto join_node = std::static_pointer_cast<JoinNode>(lqp);
+  semi_join_reduction_node->mark_as_reducer_of(join_node);
+
+  // The semi join reduction node should use the stored weak pointer to create and return a shared pointer to the
+  // corresponding join.
+  EXPECT_EQ(semi_join_reduction_node->get_or_find_corresponding_join_node(), join_node);
+
+  // After a deep copy, the semi join reduction node's weak pointer to the corresponding join node should be unset.
```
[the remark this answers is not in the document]
julianmenzler: When deep-copying the LQP, both the corresponding join, and the semi reduction are deep-copied. We do not want trigger the upwards traversal in the course of the deep copy. Hence, we trigger the traversal in a lazy fashion, whenever needed.

src/lib/logical_query_plan/join_node.hpp
dey4ss: The corresponding node is the reduced join, right (not the reducer)? So `_reduced_join_node`?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -228,18 +228,62 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
 const std::vector<std::shared_ptr<AbstractExpression>>& JoinNode::join_predicates() const { return node_expressions; }
 
+void JoinNode::mark_as_reducer_of(const std::shared_ptr<JoinNode>& corresponding_join_node) {
+  Assert(corresponding_join_node, "Corresponding JoinNode must be provided.");
+  Assert(!_is_reducer, "The semi reducer status should be set once only.");
+  Assert(join_mode == JoinMode::Semi, "Semi join reductions require JoinMode::Semi.");
+  DebugAssert(join_predicates().size() == 1,
+              "Currently, semi join reductions are expected to have a single join predicate.");
+  DebugAssert(
+      std::any_of(corresponding_join_node->join_predicates().begin(), corresponding_join_node->join_predicates().end(),
```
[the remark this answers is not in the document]
julianmenzler: Yes.

https://github.com/hyrise/hyrise/blob/2b220de44332d6367e9d6528fad7fdc4acacf729/src/lib/optimizer/strategy/semi_join_reduction_rule.cpp#L32-L45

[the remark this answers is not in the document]
julianmenzler: have to add a test for this though...

src/lib/logical_query_plan/join_node.hpp
julianmenzler: `_reduced_join_node` is a good suggestion, thanks!

src/lib/logical_query_plan/join_node.cpp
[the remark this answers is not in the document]
julianmenzler: Done.

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -228,18 +228,62 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
+    // Find corresponding join by traversing upwards
+    visit_lqp_upwards(std::const_pointer_cast<AbstractLQPNode>(shared_from_this()), [&](const auto& current_node) {
+      if (current_node.get() == this) return LQPUpwardVisitation::VisitOutputs;
+      if (current_node->type != LQPNodeType::Join) return LQPUpwardVisitation::VisitOutputs;
+      const auto join_node = std::static_pointer_cast<JoinNode>(current_node);
+      // Currently, semi reductions are supported for single predicate joins only.
+      if (join_node->join_predicates().size() != 1) return LQPUpwardVisitation::VisitOutputs;
+      if (*join_predicates()[0] != *join_node->join_predicates()[0]) return LQPUpwardVisitation::VisitOutputs;
+
+      _corresponding_join_node = std::weak_ptr<JoinNode>(join_node);
```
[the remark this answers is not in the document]
julianmenzler: I'm not seeing a possibility for this.

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -228,18 +228,62 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
+  const auto copied_join_node =
+      JoinNode::make(join_mode, expressions_copy_and_adapt_to_different_lqp(join_predicates(), node_mapping));
+  copied_join_node->_is_reducer = _is_reducer;
+  return copied_join_node;
 }
 
 bool JoinNode::_on_shallow_equals(const AbstractLQPNode& rhs, const LQPNodeMapping& node_mapping) const {
   const auto& join_node = static_cast<const JoinNode&>(rhs);
+  // We do not consider the _is_reducer attribute in this comparison, because we want the LQPTranslator to translate
+  // identical semi joins into a single operator, regardless of the `is_reducer` property.
```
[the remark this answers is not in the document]
julianmenzler: If we have two identical semi joins in the LQP, the `LQPTranslator` would de-duplicate them by creating a single operator for the semi join. 
If one semi join, however, is a marked as a semi reducer while the other semi join is not, the deduplication would no longer take place (if we consider the reducer status in shallow_equals).

src/lib/operators/abstract_join_operator.cpp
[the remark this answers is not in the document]
julianmenzler: I changed it to the most compact form:

image

Stick with it?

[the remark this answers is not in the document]
julianmenzler: LQP nodes right now:

image

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -228,19 +230,72 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
+  _reduced_join_node = std::weak_ptr<JoinNode>(reduced_join_node);
+}
+
+bool JoinNode::is_semi_reduction() const {
+  DebugAssert(!_is_semi_reduction || join_mode == JoinMode::Semi, "Non-semi join is marked as a semi reduction.");
+  return _is_semi_reduction;
+}
+
+std::shared_ptr<JoinNode> JoinNode::get_or_find_reduced_join_node() const {
+  if (!_is_semi_reduction) {
```
dey4ss, line 252: What is the semantic of returning a `nullptr` when the node is not a reducer? I'd expect an error (aka `Assert`) in that case

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -228,19 +230,72 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
+    return nullptr;
+  }
+
+  if (_reduced_join_node.expired()) {
+    // In deep copies of the LQP, the weak pointer to the reduced join is expired (lazy discovery). In such cases,
+    // find the reduced join by traversing the LQP upwards.
+    visit_lqp_upwards(std::const_pointer_cast<AbstractLQPNode>(shared_from_this()), [&](const auto& current_node) {
+      if (current_node->type != LQPNodeType::Join || current_node.get() == this) {
```
dey4ss, line 260: We can't use `current_node == /*current_node ==` because we do not take care of the reducer status in the `_on_shallow_equals` dunction, right? However, is there a nicer way to compare that the pointers poinmt to the same address?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -228,19 +230,72 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
+      return LQPUpwardVisitation::DoNotVisitOutputs;
+    });
+
+    Assert(!_reduced_join_node.expired(), "Could not find JoinNode that gets reduced by this semi join reduction.");
+  }
+
+  return _reduced_join_node.lock();
+}
+
 size_t JoinNode::_on_shallow_hash() const { return boost::hash_value(join_mode); }
```
dey4ss, line 279: What implicatons arise from ignoring the semi join flag in the hash and equality check for hash-based containers? Are we sure that there are not false matches?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -228,19 +230,72 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
+  return _is_semi_reduction;
+}
+
+std::shared_ptr<JoinNode> JoinNode::get_or_find_reduced_join_node() const {
+  if (!_is_semi_reduction) {
+    return nullptr;
+  }
+
+  if (_reduced_join_node.expired()) {
+    // In deep copies of the LQP, the weak pointer to the reduced join is expired (lazy discovery). In such cases,
```
dey4ss, line 257: Is it really expired or just not set by `_on_shallow_copy`? Also, does this make a difference when copying the LPQs?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -228,19 +230,72 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
+    // find the reduced join by traversing the LQP upwards.
+    visit_lqp_upwards(std::const_pointer_cast<AbstractLQPNode>(shared_from_this()), [&](const auto& current_node) {
+      if (current_node->type != LQPNodeType::Join || current_node.get() == this) {
+        return LQPUpwardVisitation::VisitOutputs;
+      }
+      const auto join_node = std::static_pointer_cast<JoinNode>(current_node);
+      if (std::none_of(join_node->join_predicates().begin(), join_node->join_predicates().end(),
+                       [&](const auto& predicate) { return *predicate == *join_predicates()[0]; })) {
```
dey4ss, line 265: Get `*join_predicates()[0]` once and not for every join predicate of every join node later in the LQP

dey4ss (commented):
I have some questions, apart from these this looks good to me

src/lib/logical_query_plan/join_node.cpp
julianmenzler: Yes, you are right!

src/lib/logical_query_plan/join_node.cpp
julianmenzler: It is not set for deep copies – Also makes no difference for additional deep copies.

src/lib/logical_query_plan/join_node.cpp
julianmenzler: Not really. It is just a (cheap) check to skip the first node in the upwards traversal.

julianmenzler: > However, is there a nicer way to compare that the pointers poinmt to the same address?

I guess not, but happy to hear, if so. 😄

src/lib/logical_query_plan/join_node.cpp
julianmenzler: Hmm. As a consequence, we will not be able to add two different semi join node instances to a hash-based container, if their only difference is the semi reduction flag property.

dey4ss: Having the same semi-join (intendedly) twice in the LQP with and without reduction flag should be pretty rare, too. So I'd also accept that they replace each other in a hash container if we document it somewhere. Maybe some different opinions from @Bouncner or @mweisgut?

mweisgut: But in this scenario, the visualization of the resulting physical query plan would show incorrect information: Since we translate in a bottom-up way, we would translate the reducer node first. When the reduced node is translated, we would find a match in `_operator_by_lqp_node` and use the stored join operator which points to the reducer JoinNode. If we call `description` for the visualization, we would evaluate the JoiNode as semi reducer and return `(Semi Reduction)` - still preferring `(Semi Reducer)`. Thus, both the reducer and reduced nodes would both show as reducer nodes, wouldn't they?

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -585,4 +594,62 @@ TEST_F(JoinNodeTest, UniqueConstraintsCrossJoin) {
   EXPECT_TRUE(_cross_join_node->unique_constraints()->empty());
 }
 
+TEST_F(JoinNodeTest, GetOrFindReducedJoinNode) {
+  const auto join_predicate = equals_(_t_a_a, _t_b_x);
+  auto semi_reduction_node = JoinNode::make(JoinMode::Semi, join_predicate, _mock_node_a, _mock_node_b);
```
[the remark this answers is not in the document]
julianmenzler: This is bad naming, thanks! "reducer node" refers to the node that the semi reduction uses for filtering a given table. Compare `SemiJoinReductionRule`

src/lib/logical_query_plan/join_node.cpp
julianmenzler: The reduced JoinNode should have at least one different input node because it follows the semi reduction node.

Anyway, I add the attributes to shallow hash and equals.

dey4ss: You mean sth like
```c++
const auto reducer_node = JoinNode::make(JoinMode::Semi, equals(_b_a, _a_a)),
  _node_a,
  _node_b));

const auto inner_join = JoinNode::make(JoinMode::Inner, redcuer_node, _node_b);
reducer_node->mark_as_semi_reduction(inner_join);
```

vs a dedicated constructor? How often do we have this problem in the tests? I wouldn't say we need a new join constructor for just a few test cases

mweisgut: I also vote for not adding a new constructor for a few test cases.

What about sticking to the `expected_lqp` version, removing const, and calling `mark_as_semi_reduction(expected_lqp)` for the expected lqp's left input node?

dey4ss: even shorter, I think that's fine

julianmenzler: > What about sticking to the expected_lqp version, removing const, and calling mark_as_semi_reduction(expected_lqp) for the expected lqp's left input node?

> then we need to cast expected_lqp->left_input() as join node to call the marker function, right?

Exactly. It works but adds some boilerplate code to the tests. In this PR, it only affects 4 tests. Some more will come with the SemiJoinReductionRemovalRule.

src/lib/logical_query_plan/join_node.cpp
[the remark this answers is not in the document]
julianmenzler: Outdated.

src/lib/operators/abstract_join_operator.cpp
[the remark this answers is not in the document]
Bouncner: Looks good to me.

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -229,19 +229,76 @@ bool JoinNode::is_column_nullable(const ColumnID column_id) const {
...
 
 bool JoinNode::_on_shallow_equals(const AbstractLQPNode& rhs, const LQPNodeMapping& node_mapping) const {
   const auto& join_node = static_cast<const JoinNode&>(rhs);
-  if (join_mode != join_node.join_mode) return false;
+  if (join_mode != join_node.join_mode) {
+    return false;
+  }
+  if (_is_semi_reduction != join_node._is_semi_reduction) {
+    return false;
+  }
```
dey4ss, line 301:
```suggestion
  if (join_mode != join_node.join_mode || _is_semi_reduction != join_node._is_semi_reduction) {
    return false;
  }
```

src/test/lib/optimizer/strategy/semi_join_reduction_rule_test.cpp
```diff
@@ -45,35 +45,53 @@ TEST_F(SemiJoinReductionRuleTest, CreateSimpleReduction) {
...
     _node_b);
   // clang-format on
+  expected_reduction->mark_as_semi_reduction(expected_lqp);
 
   auto actual_lqp = StrategyBaseTest::apply_rule(_rule, input_lqp);
   EXPECT_LQP_EQ(actual_lqp, expected_lqp);
+
+  // Check whether the added semi join was also marked as a semi reduction.
+  auto join_node = std::static_pointer_cast<JoinNode>(actual_lqp->left_input());
+  ASSERT_TRUE(join_node->is_semi_reduction());
```
dey4ss, line 70: I'm aware that we do not use `ASSERT/EXPECT` consistently, but I'd argue that there is no semantic difference in the checks here. Thus, we should decide for one option.

dey4ss (changes requested):
Minor comments, apart from these this LGTM.

src/test/lib/optimizer/strategy/semi_join_reduction_rule_test.cpp
julianmenzler: `ASSERT_*` aborts the test if it fails. But yeah, I changed it to `EXPECT_*`.

dey4ss: That's why I'd use `ASSERT` only when it's preventing the following cases from segfaulting (e.g, `ASSERT` that variable is not `nullptr`, `EXPECT` its members to be equal value `x`). But I guess we do not have a guideline for that.

dey4ss (approved):
Feel free to give it a go once you've casted the join node in the test.

src/test/lib/optimizer/strategy/semi_join_reduction_rule_test.cpp
julianmenzler: Makes sense 👍

PR #2466, opened 2022-05-12
Deep-Copy LQP node comments

julianmenzler:
`AbstractLQPNode::_shallow_copy` does not cover the `comment` field. As a result, previously set comments get lost in deep-copied LQPs. With this PR, comments also get copied.


dey4ss (approved):
CI currently crashes because the Intel Mac's disk is full

PR #2393, opened 2021-08-01
[DYOD] DDL

simonstadlinger:
# [DYOD] DDL Project Pull Request

## What changes?

In this pull request we introduce support for DDL statements for the following topics: 
- create/drop index
- metatables for indexes
- tablekey and column constraints
- support for unsupported datatypes
- alter table drop column

In the next sections the changes and additions will be explained in detail.

This pull request should be reviewed and merged in conjunction with [this pull request](https://github.com/hyrise/sql-parser/pull/177) on the sql parser repository. 

### Create/drop index

For create and drop index we edited each part of the pipeline. Starting with the sql-parser we introduced rules for parsing create index and drop index statemens. We added translation methods in the sql-translator and introduced new lqp nodes for both operations. We added two new operatos which are created in newly added methods in the lqp-translator. The create index operator decides which index type to use; currently we support `GroupKeyIndex`, if a single column are given within the statement and `CompositeGroupKeyIndex`, if multiple columns are provided with the input.

The create index DDL statement for now supports:
- the `IF NOT EXISTS` flag
- multiple columns

The drop index DDL statement for now supports:
- the `IF EXISTS` flag

### Metatables for indexes

Hyrise has several meta tables (see meta_table_manager.hpp/cpp). For example, it is possible to get meta information about the stored tables with `SELECT * FROM meta_tables` or meta information about the columns with `SELECT * FROM meta_columns`. Similarly, you can now use `SELECT * FROM meta_indexes` to view meta information about the created index structures. In meta_indexes table one sees for each index the index type, the name, the table and the columns concerned.

### Tablekey and column constraints

For the table key constraints we edited the create table statement in all parts of the pipeline. Starting with an optional non terminal in the sql-parser, passing down information about the type of the constraint (PRIMARY KEY and UNIQUE), the name of the chosen table, and the specified columns, until we then, in the create table operator, use the already implemented add_soft_key_constraints method to add the constraints to the table.

For the column constraints, we modified the column_definition to now contain a variable of an enum class, storing information if and what constraint type is defined for the column. The constraints can again be set in the postgresql syntax in the console in the create table statement and then be parsed and stored in the dummy table that is send through the pipeline and used to create the actual table in the operator at the end. 

### Unsupported Datatypes

[...]

PR #2468, opened 2022-05-16
Change `CREATE TABLE` Columns' Default Nullability and Track Constraints

dey4ss:
This PR makes the columns of `CREATE TABLE` statements nullable by default to conform to the SQL standard (see hyrise/sql-parser#209 for details). Furthermore, it stores `UNIQUE` and `PRIMARY KEY` constraints of `CREATE TABLE` statements as soft table key constraints of the created table. Since Hyrise does not enforce these constraints, a warning is printed. The insertion order of such soft key constraints does not have an impact.


src/lib/sql/sql_translator.cpp
```diff
@@ -1363,9 +1364,61 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+                  << std::string{magic_enum::enum_name(constraint_type)} + " constraint for column " +
+                         column_definition.name + " will not be enforced\n";
+      }
+    }
+
+    // Translate table constraints
+    const auto column_count = column_definitions.size();
+    for (const auto& table_constraint : *create_statement.tableConstraints) {
+      std::unordered_set<ColumnID> column_ids;
+      const bool is_primary_key_constraint = table_constraint->type == hsql::ConstraintType::PrimaryKey;
```
Bouncner, line 1388: auto

src/lib/sql/sql_translator.cpp
```diff
@@ -1363,9 +1364,61 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+
+      // Resolve column IDs
+      for (const auto column_name : *table_constraint->columnNames) {
```
Bouncner, line 1391: reference? Looks like a string.

src/lib/sql/sql_translator.cpp
```diff
@@ -1363,9 +1364,61 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+        for (auto column_id = ColumnID{0}; column_id < column_count; ++column_id) {
+          auto& column_definition = column_definitions[column_id];
+          if (column_definition.name == column_name) {
```
Bouncner, line 1394: Guard clause?

src/lib/storage/table_key_constraint.cpp
```diff
@@ -7,6 +7,14 @@ TableKeyConstraint::TableKeyConstraint(std::unordered_set<ColumnID> init_columns
 
 KeyConstraintType TableKeyConstraint::key_type() const { return _key_type; }
 
+size_t TableKeyConstraint::hash() const {
+  auto hash = boost::hash_value(_key_type);
+  for (const auto column : columns()) {
```
Bouncner, line 12: Reference?

src/lib/sql/sql_translator.cpp
```diff
@@ -1363,9 +1364,61 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+      // Translate column constraints
+      for (const auto& column_constraint : *parser_column_definition->column_constraints) {
+        if (column_constraint != hsql::ConstraintType::Unique &&
+            column_constraint != hsql::ConstraintType::PrimaryKey) {
+          continue;
+        }
+        const auto constraint_type = column_constraint == hsql::ConstraintType::PrimaryKey
+                                         ? KeyConstraintType::PRIMARY_KEY
+                                         : KeyConstraintType::UNIQUE;
+        table_key_constraints.emplace_back(std::unordered_set<ColumnID>{column_id}, constraint_type);
```
Bouncner, line 1377: As I saw the line `const auto unique_constraint = TableKeyConstraint{{ColumnID{0}}, KeyConstraintType::UNIQUE};` down below, maybe we can use `emplace_back({ColumnID{0}}, constraint_type)`?

src/lib/sql/sql_translator.cpp
```diff
@@ -1363,9 +1364,61 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+        std::cout << "WARNING: "
+                  << std::string{magic_enum::enum_name(constraint_type)} + " constraint for column " +
```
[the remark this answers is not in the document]
dey4ss: AFAIK it's a string view

src/lib/sql/sql_translator.cpp
dey4ss: Unfortunately, this does not work (`column_id` is of type `ColumnID`)
```
/Users/deyass/Documents/phd/hyrise/src/lib/sql/sql_translator.cpp:1377:31: error: no matching member function for call to 'emplace_back'
        table_key_constraints.emplace_back({column_id}, constraint_type);
        ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
/opt/homebrew/bin/../include/c++/v1/vector:721:19: note: candidate template ignored: substitution failure: deduced incomplete pack <(no value), const opossum::KeyConstraintType &> for template parameter '_Args'
        reference emplace_back(_Args&&... __args);
```

src/lib/sql/sql_translator.cpp
dey4ss: Currently, the parser ensures that `columnNames` is not a `nullptr`, however, a (`DebugAssert`) guard does not hurt. `column_count` is set to `column_definitions.size()` and we don't need a guard there.

dey4ss:
Currently, the comparison of two `StaticTableNode`s would fail if their tables' key constraints are inserted in a different order. If we do not consider the insert order as a feature, the (imo) most elegant solution would be to use sets instead of vectors to store a table's constraints. Any comments on that?

dey4ss:
Just used sets for TableKeyConstraints, too. Guess this can be reviewed now.

dey4ss:
Since hyrise/sql-parser#214 was merged before hyrise/sql-parser#210, I suggest to schedule this PR after #2471 to separate the corresponding functionality from this PR.

src/lib/sql/sql_translator.cpp
Bouncner: That means what? :)

dey4ss: Furthermore, I added a check that all column names defined in a table constraint refer to actual columns. However, it is still possible to request multiple primary keys (--> we crash) or overlapping constraints.

src/lib/storage/abstract_table_constraint.cpp
```diff
@@ -30,3 +29,12 @@ bool AbstractTableConstraint::operator!=(const AbstractTableConstraint& rhs) con
 }
 
 }  // namespace opossum
+
+namespace std {
+
+size_t hash<opossum::AbstractTableConstraint>::operator()(
+    const opossum::AbstractTableConstraint& table_constraint) const {
+  return table_constraint.hash();
+}
```
mweisgut, line 38: What is this required for?

mweisgut (commented):
Reviewing the test cases will follow within the next few days.

Could you please merge the master branch into this branch? I could not build this branch locally with llvm clang 14 due to reserved identifier issues caused by dtrace.

Referencing #2393, which has some functional overlap with this code.

src/lib/storage/abstract_table_constraint.cpp
dey4ss: We might want to use hash-based containers, but I can also remove it

src/lib/operators/maintenance/create_table.cpp
```diff
@@ -46,6 +46,28 @@ std::string CreateTable::description(DescriptionMode description_mode) const {
...
+        if (std::next(column_it) != columns.cend()) {
+          stream << ", ";
+        }
+      }
+      stream << ")";
+      if (std::next(constraint_it) != table_key_constraints.cend()) {
+        stream << separator;
+      }
+    }
+  }
```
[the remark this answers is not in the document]
dey4ss: Do you prefer this utility function in `string_utils.hpp` or as a a method of a `Table` object (we need tp pass the table in any case to get the column names)?

src/lib/sql/sql_translator.cpp
```diff
@@ -1398,9 +1399,65 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
       column_definition.name = parser_column_definition->name;
       column_definition.nullable = parser_column_definition->nullable;
+
+      // Translate column constraints
+      DebugAssert(parser_column_definition->column_constraints,
+                  "Column " + column_definition.name + " is missing constraint information");
+      for (const auto& column_constraint : *parser_column_definition->column_constraints) {
+        if (column_constraint != hsql::ConstraintType::Unique &&
+            column_constraint != hsql::ConstraintType::PrimaryKey) {
+          continue;
```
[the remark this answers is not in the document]
dey4ss: These are all constraint types from the parser:
```c++
enum struct ConstraintType { None, NotNull, Null, PrimaryKey, Unique };
```
We do not care about the nullability constraints when adding key constraints.

src/lib/sql/sql_translator.cpp
```diff
@@ -1398,9 +1399,65 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+        }
+        const auto constraint_type = column_constraint == hsql::ConstraintType::PrimaryKey
+                                         ? KeyConstraintType::PRIMARY_KEY
+                                         : KeyConstraintType::UNIQUE;
```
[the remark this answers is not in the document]
dey4ss: Indeed, we have mor constraints, but we only reach this line when the constraint is either unique/primary key. We checked this directly before this line

src/lib/sql/sql_translator.cpp
```diff
@@ -1398,9 +1399,65 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+        std::cout << "WARNING: " << magic_enum::enum_name(constraint_type) << " constraint for column "
+                  << column_definition.name << " will not be enforced\n";
+      }
+    }
+
+    // Translate table constraints
+    const auto column_count = column_definitions.size();
+    for (const auto& table_constraint : *create_statement.tableConstraints) {
+      auto column_ids = std::set<ColumnID>{};
+      const auto constraint_type = table_constraint->type == hsql::ConstraintType::PrimaryKey
```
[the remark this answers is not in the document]
dey4ss: In this case, I'm fine to add a switch, as nullability constraints should not be table constraints but column constraints.

src/lib/storage/abstract_table_constraint.cpp
mweisgut: I don't have a strong opinion on this, but I'm tending to remove it since we don't need it now and add it in the future if it's needed at some point.

src/lib/sql/sql_translator.cpp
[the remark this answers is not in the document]
mweisgut: I'd be fine with an assertion.

src/test/lib/operators/maintenance/create_table_test.cpp
```diff
@@ -66,6 +85,25 @@ TEST_F(CreateTableTest, Execute) {
...
 }
 
+TEST_F(CreateTableTest, SoftKeyConstraints) {
+  const auto context = Hyrise::get().transaction_manager.new_transaction_context(AutoCommit::No);
+  const auto input_table = const_pointer_cast<Table>(dummy_table_wrapper->table);
+  const auto unique_constraint = TableKeyConstraint{{ColumnID{0}}, KeyConstraintType::UNIQUE};
+  input_table->add_soft_key_constraint(unique_constraint);
+
+  create_table->set_transaction_context(context);
+
```
mweisgut, line 95: In `CreateTableTest, NameAndDescriptionWithConstraints`, setting up the table happens first before setting up and performing the transaction. It's only a minor concern, but perhaps, we could do it in a similar way here. Further, `context` is only used in line 94 so it can be moved close to the code location where it's actually used, i.e., above line 94.

src/test/lib/utils/string_utils_test.cpp
```diff
@@ -38,4 +39,17 @@ TEST_F(StringUtilsTest, trim_source_file_path) {
...
 
+TEST_F(StringUtilsTest, table_key_constraints_to_stream) {
+  const auto table = Table::create_dummy_table(
+      {{"a", DataType::Int, false}, {"b", DataType::Int, false}, {"c", DataType::Int, false}});
+  table->add_soft_key_constraint({{ColumnID{0}, ColumnID{1}}, KeyConstraintType::UNIQUE});
+  table->add_soft_key_constraint({{ColumnID{2}}, KeyConstraintType::PRIMARY_KEY});
+  const auto separator = std::string{" | "};
+  auto stream = std::ostringstream{};
+
+  table_key_constraints_to_stream(stream, table, separator);
```
mweisgut, line 50: Can you add another check for `table_key_constraints_to_stream` where `separator` is not given?

mweisgut (commented):
Nice work! Mainly suggestions for improving the readability/comprehensibility.

src/lib/sql/sql_translator.cpp
```diff
@@ -1398,9 +1399,68 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+        const auto constraint_type = column_constraint == hsql::ConstraintType::PrimaryKey
+                                         ? KeyConstraintType::PRIMARY_KEY
+                                         : KeyConstraintType::UNIQUE;
+        table_key_constraints.emplace(std::set<ColumnID>{column_id}, constraint_type);
+        std::cout << "WARNING: " << magic_enum::enum_name(constraint_type) << " constraint for column "
+                  << column_definition.name << " will not be enforced\n";
+      }
+    }
+
+    // Translate table constraints
```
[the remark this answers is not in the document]
dey4ss: Well, in this case I'm just making a difference on what level a constraint was requested in the statement:
```SQL
CREATE TABLE foo (
    a int,
    b UNIQUE NOT NULL,  -- constraints are part of the ColumnDefinition
    PRIMARY KEY (a)  -- constraint is attached to the CreateTableStatement
);
```
I'm gonna add your comment and and also add a note that explains this difference.

src/lib/utils/string_utils.cpp
```diff
@@ -52,4 +55,25 @@ std::string replace_addresses(const std::string& input) {
   return std::regex_replace(input, std::regex{"0x[0-9A-Fa-f]{4,}"}, "0x00000000");
 }
 
+void table_key_constraints_to_stream(std::ostream& stream, const std::shared_ptr<const Table>& table,
+                                     const std::string& separator) {
+  const auto& table_key_constraints = table->soft_key_constraints();
+  for (auto constraint_it = table_key_constraints.cbegin(); constraint_it != table_key_constraints.cend();
```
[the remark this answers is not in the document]
dey4ss: I guess your version is fine, too. I'd just add `Assert`s that neither the constraints nor each constraint's columns are empty to prevent segfaults when dereferencing the decreased iterators.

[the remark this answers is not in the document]
dey4ss: Anyways, is this method well-placed in `string_utils` or do yo prefer it to be a member of `Table`?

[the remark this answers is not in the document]
mweisgut: As discussed in person: creating `print_utils.hpp/cpp` files for this function might be an option. Then, we might also move other print-supportive code to the print_utils files (not as part of this PR).

src/lib/utils/print_utils.hpp
```diff
@@ -0,0 +1,79 @@
...
+using NodePrintFn = std::function<void(const std::shared_ptr<Node>&, std::ostream& stream)>;
+
+namespace detail {
+
+/**
+ *
+ * @param indentation   its size determines the indentation of a node, a true means a vertical line "|",
+ *                      a false means, a space " " should be used to increase the indentation
+ * @param id_by_node    used to determine whether a node was already printed and which id it had
+ * @param id_counter    used to generate ids for nodes
```
mweisgut: Since we referenced the contribution guidelines several times today: `Maintain correct orthography and grammar. Comments should start with a capital letter, sentences should be finished with a full stop.`

src/lib/utils/print_utils.hpp
```diff
@@ -0,0 +1,79 @@
...
+ *  |     \_Recurring Node --> [3]
+ *  |     \_Recurring Node --> [5]
+ *  \_[7] [Cross Join]
+ *     \_Recurring Node --> [3]
+ *     \_Recurring Node --> [5]
+ *
+ * @param node              The root node originating from which the graph should be printed
+ * @param get_children_fn   Callback that returns a nodes children
+ * @param print_node_fn     Callback that prints the information about a node, e.g. "[Predicate] a = 42" in the example
+ *                              above
```
mweisgut, line 39: Align `above` with `Callback`?

src/test/lib/utils/print_utils_test.cpp
```diff
@@ -1,13 +1,65 @@
...
+   * [0] node_a
+   *  \_[1] node_b
+   *  |  \_[2] recurring_node
+   *  \_[3] node_c
+   *     \_Recurring Node --> [2]
+   *     \_Recurring Node --> [2]
+   */
+  const auto expected_string = std::string{
+      "[0] node_a\n \\_[1] node_b\n |  \\_[2] recurring_node\n \\_[3] node_c\n    \\_Recurring Node --> [2]\n    "
+      "\\_Recurring Node --> [2]\n"};
```
mweisgut, line 56: Instead of adding the pretty formatted string representation above as a comment and the expected string in a compact format, how about removing the comment and formatting the expected string similar to the current comment? Then, we would not have the string twice and one could argue that this would also increase the expected string's readablity.

dey4ss: I initially dit it this way, but escaping the backslashes (adds one character offset at some places) and adding newline characters in the string kind of did not made this look nice IMO:
```c++
const auto expected_string = std::string{
  "[0] node_a\n"
  " \\_[1] node_b\n"
  " |  \\_[2] recurring_node\n"
  " \\_[3] node_c\n"
  "    \\_Recurring Node --> [2]\n"
  "    \\_Recurring Node --> [2]\n"};
```

mweisgut: Even with the escaping backslashes and the `\n`s, I think the structure of the lqp is quite clear with that formatting.
Pinging @Bouncner and @tjjordan for more opinions on that.

mweisgut (approved):

src/test/lib/utils/print_utils_test.cpp
tjjordan: Hm, in my view both options are totally fine.
Wouldn't it be an option to keep both the nice formatted comment and the nice formatted ´expected_string`?
If we want to get rid of the duplicate I'd prefer the nice formatted ´expected_string` because it makes more clear how the string should be build.

dey4ss: Current version is:
```c++
  auto stream = std::ostringstream{};
  const auto expected_string = std::string{
      R"([0] node_a
 \_[1] node_b
 |  \_[2] recurring_node
 \_[3] node_c
    \_Recurring Node --> [2]
    \_Recurring Node --> [2]
)"};

  print_directed_acyclic_graph<const AbstractLQPNode>(lqp, get_inputs_fn, node_print_fn, stream);
  EXPECT_EQ(stream.str(), expected_string);
```

PR #2471, opened 2022-06-10
Support `BOOLEAN` and `TIMESTAMP` Columns

dey4ss:
The SQL parser introduced `BOOLEAN` and `TIMESTAMP` column types with hyrise/sql-parser#214. This PR adds the required functionality to Hyrise. Furthermore, `CAST` statements of string literals to `TIMESTAMP`/`DATETIME` are checked for semantical correctness.


src/lib/sql/sql_translator.cpp
```diff
@@ -1370,6 +1370,10 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
           std::cout << "WARNING: Parsing TIME to string since date and time data types are not yet supported\n";
           column_definition.data_type = DataType::String;
           break;
+        case hsql::DataType::BOOLEAN:
+          std::cout << "WARNING: Parsing BOOLEAN to string\n";
```
Bouncner, line 1374: Add information how this looks like? "TRUE", "true", "1"?

src/lib/sql/sql_translator.cpp
```diff
@@ -1799,16 +1803,33 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_translate_hsql_expr(
...
-        const auto date_string = boost::get<pmr_string>(static_cast<ValueExpression&>(*left).value);
-        const auto date = string_to_date(std::string{date_string});
-        if (date) return std::const_pointer_cast<AbstractExpression>(left);
-        FailInput("'" + std::string{date_string} + "' is not a valid date");
+      const auto target_hsql_data_type = expr.columnType.data_type;
+
+      if (target_hsql_data_type == hsql::DataType::DATE || target_hsql_data_type == hsql::DataType::DATETIME) {
+        AssertInput(source_data_type == DataType::String,
+                    "Cannot cast " + left->as_column_name() + " as " +
+                        std::string{magic_enum::enum_name(target_hsql_data_type)});
```
Bouncner, line 1811: Isn't enum_name already returning an std::string?

src/lib/sql/sql_translator.cpp
```diff
@@ -1799,16 +1803,33 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_translate_hsql_expr(
...
+        // We do not know if an expression to be casted other than a ValueExpression actually contains date time
+        // values, and we cannot check this later due to the lack of proper data types.
+        AssertInput(left->type == ExpressionType::Value, "Only ValueExpressions can be casted as " +
+                                                             std::string{magic_enum::enum_name(target_hsql_data_type)});
+        const auto input_string = std::string{boost::get<pmr_string>(static_cast<ValueExpression&>(*left).value)};
+
+        if (target_hsql_data_type == hsql::DataType::DATE) {
+          const auto date = string_to_date(input_string);
+          AssertInput(date, "'" + input_string + "' is not a valid date");
```
Bouncner, line 1820: "Cannot be casted"? Not that is matters much, but there a sooo many date formats (imagine these wild people from the west which put the month in front) that it might fail because we cannot cast it rather than the string not being a valid date.

src/lib/sql/sql_translator.cpp
```diff
@@ -1799,16 +1803,33 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_translate_hsql_expr(
...
+          return std::const_pointer_cast<AbstractExpression>(left);
```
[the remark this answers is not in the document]
Bouncner: Besides that I am kind of confused that we don't to any casting here. We throw away `date` and that's it?

src/lib/utils/date_time_utils.cpp
```diff
@@ -17,6 +18,21 @@ std::optional<boost::gregorian::date> string_to_date(const std::string& date_str
   return std::nullopt;
 }
 
+std::optional<boost::posix_time::ptime> string_to_date_time(const std::string& date_time_string) {
+  try {
+    const auto date_time = boost::posix_time::time_from_string(date_time_string);
+    if (!date_time.is_not_a_date_time()) {
+      return date_time;
+    }
+  } catch (const boost::wrapexcept<boost::gregorian::bad_day_of_month>&) {
```
Bouncner, line 27: Add a comment why we ignore these issues and other not?

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -3097,8 +3099,10 @@ TEST_F(SQLTranslatorTest, CastStatement) {
   EXPECT_THROW(sql_to_lqp_helper("SELECT CAST(a AS DATE) FROM int_string"), InvalidInputException);
   // We do not have a real Date data type, so we cannot pass this cast anywhere else.
   EXPECT_THROW(sql_to_lqp_helper("SELECT CAST(b AS DATE) FROM int_string"), InvalidInputException);
-  // This is valid SQL, but we do not have a DateTime data type
-  EXPECT_THROW(sql_to_lqp_helper("SELECT CAST('2000-01-01 00:00:00' AS DATETIME)"), InvalidInputException);
+  EXPECT_THROW(sql_to_lqp_helper("SELECT CAST('2000-01-01' AS DATETIME)"), InvalidInputException);
```
Bouncner, line 3102: Test for too short integers? Such as `10-01-01 00:00:00` or `'2020-01-01 0:0:17`?

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -3132,6 +3136,22 @@ TEST_F(SQLTranslatorTest, CastStatement) {
     const auto [actual_lqp, translation_info] = sql_to_lqp_helper("SELECT CAST('2000-01-01' as DATE);");
     EXPECT_LQP_EQ(actual_lqp, expected_lqp);
   }
+  {
+    // Date times need to be resolved, as we do not have a DateTime DataType
```
Bouncner, line 3140: What does "resolved" mean here?

src/test/lib/utils/date_time_utils_test.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+TEST_F(DateTimeUtilsTest, StringToDateTime) {
+  EXPECT_EQ(string_to_date_time("2000-01-45 00:00:00"), std::nullopt);
+  EXPECT_EQ(string_to_date_time("2000-13-01 00:00:00"), std::nullopt);
+  EXPECT_EQ(string_to_date_time("2000-04-31 00:00:00"), std::nullopt);
+  EXPECT_EQ(string_to_date_time("2001-02-29 00:00:00"), std::nullopt);
+  EXPECT_EQ(string_to_date_time("-1-02-29 00:00:00"), std::nullopt);
+  EXPECT_EQ(string_to_date_time("2000-01-01"), std::nullopt);
+  EXPECT_EQ(string_to_date_time("2000-01-01 00:00:x"), std::nullopt);
+  EXPECT_EQ(string_to_date_time("2000-01-01 00:x:00"), std::nullopt);
```
Bouncner, line 38: Add a test with the middle `T`?

src/lib/sql/sql_translator.cpp
dey4ss: This depends on what the user adds to that table. We do not parse the input values, so it could be any string the user inserts. Postgres [supports a variety of options](https://www.postgresql.org/docs/14/datatype-boolean.html), but we do not enforce a date format (unless if it's a date literal or cast as date) either.

src/lib/sql/sql_translator.cpp
dey4ss: It's a [string view](https://github.com/Neargye/magic_enum/blob/master/doc/reference.md#enum_name) and iirc we have to create a string from it to use the `+` operator.

src/lib/sql/sql_translator.cpp
```diff
@@ -1799,16 +1803,33 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_translate_hsql_expr(
...
+        }
+
+        if (target_hsql_data_type == hsql::DataType::DATETIME) {
+          const auto date_time = string_to_date_time(input_string);
+          AssertInput(date_time, "'" + input_string + "' is not a valid date time");
+          // Overflowing time components are added to the subsequent unit while parsing, and we want to get a
```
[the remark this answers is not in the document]
dey4ss: Hm, found it confusing myself. Boost accepts sth like `2000-01-01 00:00:61` and parses it to a timestamp equivalent to `2000-01-01 00:01:01`. So "overflowing" time components are accepted and added to the next component. This also applies to minute (`2000-01-01 00:61:00` --> `2000-01-01 01:01:00`) and hour values (`2000-01-01 25:00:00` --> `2000-01-02 01:00:00`). 

I'm aware this is not quite intuitive, and I'm open to restrict such timestamps. However, I did not have a nice idea on how to do this other than parsing the time components manually.

src/lib/sql/sql_translator.cpp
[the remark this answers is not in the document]
dey4ss: Absolutely correct, the `cast` is not necessary.

Well, we don't have a `Date` data type, so we just check if the given string is a correct date and return the string's value expression.

src/lib/utils/date_time_utils.cpp
dey4ss: Is `// We catch parsing errors since we return a std::nullopt if the input string is not a valid date/timestamp` sufficient?

src/test/lib/sql/sql_translator_test.cpp
dey4ss: Currently, the latter should be acceptable (at least boost accepts it, we can change this if you like). Gonny add a test for the first case.

src/test/lib/sql/sql_translator_test.cpp
dey4ss: `resolved/translated to a value expression`?

src/lib/sql/sql_translator.cpp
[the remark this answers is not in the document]
dey4ss: Added a comment here

src/lib/sql/sql_translator.cpp
Bouncner: Sounds good.

src/lib/sql/sql_translator.cpp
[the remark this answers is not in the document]
Bouncner: Maybe keep it simple and add a link to the boost docs?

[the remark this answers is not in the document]
Bouncner: Somewhere else you had an example. We could also refer to that.

src/lib/sql/sql_translator.cpp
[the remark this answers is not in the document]
dey4ss: > `left` is a n abstract expression, right? Why the cast?
Argh, the cast was there for a reason...
```
/var/lib/jenkins/workspace/hyrise_hyrise_PR-2471/src/lib/sql/sql_translator.cpp:1822:18: error: constness of 'left' prevents automatic move [performance-no-automatic-move,-warnings-as-errors]

          return left;
```

So I *could* remove the `const` qualifier from the variable declaration, do the cast, or return `left->deep_copy()`. Opinions?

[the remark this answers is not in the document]
Bouncner: Haven't found the code I was commenting on. Can we resolve? I am confused. :D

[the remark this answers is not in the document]
dey4ss: Still there somehow, the line is:
https://github.com/hyrise/hyrise/blob/dey4ss/datetime/src/lib/sql/sql_translator.cpp#L1850

I removed the `const` qualifier or the `left` variable, which is one possible solution. The `CAST` expression also returns the left input in L1869. I dod not remove the `const_pointer_cast` there since I was not sure what was your opinion on how we should return the left input.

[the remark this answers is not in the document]
Bouncner: Feel free to remove the other const cast as well.

Bouncner (approved):

Issue #2472, opened 2022-06-20
BinaryWriterTest: Superfluous file removal during `SetUp`


mweisgut:
```
class BinaryWriterTest : public BaseTest {
 protected:
  void SetUp() override {
    std::remove(filename.c_str());
  }

  void TearDown() override {
    std::remove(filename.c_str());
  }
```

Removing the file in `TearDown` should be sufficient.

PR #2474, opened 2022-06-27
Removed superfluous file removal during `SetUp `

tjjordan:
Minor change to the BinaryWriterTest with regard to Issue #2472:
- Removed unnecessary file removal in test setup

The tests still run successful and no files exist after finishing the tests.

### DIFF
--- a/src/test/lib/import_export/binary/binary_writer_test.cpp
+++ b/src/test/lib/import_export/binary/binary_writer_test.cpp
@@ -19,7 +19,9 @@ namespace opossum {
 
 class BinaryWriterTest : public BaseTest {
  protected:
-  void TearDown() override { std::remove(filename.c_str()); }
+  void TearDown() override {
+    std::remove(filename.c_str());
+  }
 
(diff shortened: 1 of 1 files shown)


mweisgut:
I was also having this in mind and yes, I think an error could occur in that case. And sure, with the file removal in the setup function, we could avoid such an error. However, if a segfault happens during the test suite execution, the fault is probably at a different location which should be fixed in that case.

dey4ss:
My question was just out of curiosity, and I don't know if we also cover this for the csv writer tests. I'm fine with removing the initial test file removal (what a sentence 😅): if a segfault happens during auch a test, it should be a problem with the binary writer and the developer should know best what happend where.

Bouncner:
@tjjordan : I see the manual format run (#c4ce917). Do I get that right, that the format script of the CI did not complain but your "local" clang-format did the "right thing"?

tjjordan:
@Bouncner Yep, this is what happened.

PR #2478, opened 2022-07-11
Use Clang 14 on macOS CI runners

Bouncner:
This PR updates the llvm version on the macOS CI machines to llvm 14.


mweisgut:
>but still complains about double underscores, which are generated by dtrace. [...]
>As a consequence, we [...] additionally disable them for Clang 14 on Mac OS.

After removing DTrace (#2480), disabling it for clang-13 should be sufficient then.

Bouncner:
I'll have a look at your PR. Did not expect that to happen that soon. :)
Will happily close this PR here then.

Bouncner:
The M1 mac sucessfully run with llvm 14, the x86 lost its connection at linking. I assume it's working.
I will issue another run, once the x86 is back up again.

Once approved, I'll change the macs and the Jenkinsfile to use the default llvm installation, which will be 14.

Bouncner:
@mweisgut , @dey4ss : now it's failing with another error, stemming from `readline`. I think I've seen this issue before. It's caused by `readline` being "locally" installed via brew, while it's considered a system library under Linux (via `apt`).

We could either use readline as a system  library, or silence the warning for clang 14. Since we check for proper identifiers within the CI, I would personally prefer silencing this warning under macOS. Your opinions?

mweisgut:
> @mweisgut , @dey4ss : now it's failing with another error, stemming from `readline`. I think I've seen this issue before. It's caused by `readline` being "locally" installed via brew, while it's considered a system library under Linux (via `apt`).
> 
> We could either use readline as a system library, or silence the warning for clang 14. Since we check for proper identifiers within the CI, I would personally prefer silencing this warning under macOS. Your opinions?

Even though we check for proper identifiers in a CI run, I would prefer to keep the reserved identifier check also for the local development in non-`readline` related files.

mweisgut (approved):

PR #2482, opened 2022-07-13
Ensure correct results of `EXTRACT` expressions

### LINKED ISSUE #2476: `ExtractExpression` Has the Wrong Data Type
`ExtractExpressions` handle the extraction of components from dates/timestamps or intervals.  In detail, they model the user's request to get the year, month, day in the month, hour, minute, or second from a date/timestamp. Currently, the result of such an expression is a string:
```c++
DataType ExtractExpression::data_type() const {
  // Dates are Strings, DateComponents COULD be Ints, but lets leave it at String for now
  return DataType::String;
}
```

The SQL-92 standard [1] defines:
>If `<extract expression>` is specified, then
>Case:
>
>* a) If `<datetime field>` does not specify SECOND, then the data type of the result is exact numeric with implementation-defined precision and scale 0.
>
>* b) Otherwise, the data type of the result is exact numeric with implementation-defined precision and scale. The implementation-defined scale shall not be less than the specified or implied `<time fractional seconds precision>` or `<interval fractional seconds precision>`, as appropriate, of the SECOND `<datetime field>` of the `<extract source>`.

Thus, extracting year, month, day, hour, and minute should result in an int/long, and extracting second may be an int/long (when no fractional seconds are given) or a float/double. Postgres [2] implements this requirement accordingly, but only for timestamp and interval literals.

My suggestion is to change the result type of an `ExtractExpression` to `DataType::Int` if anything but seconds should be extracted and to `DataType::Float` or `DataType::Double` when seconds should be extracted, regardless of whether fractional seconds are given or not. Furthermore, this requires an adaption of the `ExpressionEvaluator`. Together with this change, I also suggest supporting the extraction of hour, minute, and second if the input contains the corresponding fields.

For now, we simply access the underlying strings at special positions. Since we already have utilities for dates (and utilities for timestamps may be available soon), correctly parsing the given strings and returning the members of the parsed date or timestamp objects can be done. This allows for adequate error handling, input verification, and ensuring correct results. 

[1] https://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt page 129
[2] https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT

dey4ss:
This PR returns numerical results for extract expressions. Furthermore, it adds checks tht the unterlying strings contain correct dates or timestamps. For details, see #2476.

closes #2476


src/lib/utils/date_time_utils.cpp
```diff
@@ -36,20 +36,19 @@ std::optional<boost::posix_time::ptime> string_to_date_time(const std::string& d
...
     }
     case DatetimeComponent::Month: {
-      const auto interval = boost::date_time::month_functor<boost::gregorian::date>(offset);
-      return start_date + interval.get_offset(start_date);
+      return *(++boost::gregorian::month_iterator(start_date, offset));
     }
     case DatetimeComponent::Day: {
-      const auto interval = boost::date_time::day_functor<boost::gregorian::date>(offset);
-      return start_date + interval.get_offset(start_date);
+      return *(++boost::gregorian::day_iterator(start_date, offset));
```
mweisgut, line 45: Why do we need to increment here?

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -801,52 +801,77 @@ std::shared_ptr<ExpressionResult<Result>> ExpressionEvaluator::_evaluate_functio
...
+        Assert(value.size() >= 10u, "Invalid ISO 8601 extended timestamp '" + value + "'");
+        if (value.size() > 10u) {
+          const auto& timestamp = string_to_date_time(value);
+          Assert(timestamp, "Invalid ISO 8601 extended timestamp '" + value + "'");
+          values[chunk_offset] = extract_component(*timestamp);
+
+        } else {
+          const auto& date = string_to_date(value);
+          Assert(date, "Invalid ISO 8601 extended date '" + value + "'");
+          values[chunk_offset] = extract_component(boost::posix_time::ptime{*date, {}});
```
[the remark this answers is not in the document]
dey4ss: Alright, I'm gonna adapt the code. Do you prefer `date_interval()` and `date_to_string` to accept a `ptime` object or a `date`object then?

src/lib/utils/date_time_utils.cpp
dey4ss: I added a comment:
```c++
// We obtain an iterator that adds offset years when incremented. Thus, we have to actually increment and
// dereference it. The same applies to the other cases.
```

src/lib/sql/sql_translator.cpp
```diff
@@ -1740,13 +1742,13 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_translate_hsql_expr(
                       "Interval can only be applied to ValueExpression with String value");
           const auto start_date_string =
               std::string{boost::get<pmr_string>(static_cast<ValueExpression&>(*left).value)};
-          const auto start_date = string_to_date(start_date_string);
+          const auto start_date = string_to_timestamp(start_date_string);
```
mweisgut, line 1745: start_timestamp?

mweisgut: or add `->date()` since we don't use the timestamp anyway?

src/lib/utils/date_time_utils.cpp
```diff
@@ -4,26 +4,20 @@
...
-    if (!date_time.is_not_a_date_time()) {
-      return date_time;
+    if (timestamp_string.size() == 10) {
+      // This is a date without time information.
+      const auto date = boost::gregorian::from_string(timestamp_string);
+      if (!date.is_not_a_date()) {
+        return boost::posix_time::ptime{date};
+      }
+    } else {
+      const auto date_time = boost::posix_time::time_from_string(timestamp_string);
```
mweisgut, line 17: Since the signature was changed to `timestamp`, do we keep the naming consistent and use `timestamp` here as well?

src/lib/utils/date_time_utils.cpp
```diff
@@ -4,26 +4,20 @@
...
-std::optional<boost::posix_time::ptime> string_to_date_time(const std::string& date_time_string) {
+std::optional<boost::posix_time::ptime> string_to_timestamp(const std::string& timestamp_string) {
   // We catch parsing exceptions since we return a std::nullopt if the input string is not a valid timestamp.
   try {
-    const auto date_time = boost::posix_time::time_from_string(date_time_string);
-    if (!date_time.is_not_a_date_time()) {
-      return date_time;
+    if (timestamp_string.size() == 10) {
+      // This is a date without time information.
+      const auto date = boost::gregorian::from_string(timestamp_string);
```
mweisgut, line 12: `from_string` seems to be deprecated. Can we use `from_simple_string` here?
https://github.com/boostorg/date_time/blob/0aed920bfb25d779ee6ffe36e0ef2d5f9f557448/include/boost/date_time/gregorian/parsers.hpp#L47

src/lib/utils/date_time_utils.hpp
```diff
@@ -8,19 +8,14 @@
 namespace opossum {
 
 /**
- * Takes delimited date string with order year-month-day (ISO 8601 extended format), e.g., "2001-01-01".
+ * Takes delimited timestampt string with order year-month-day [hour:minute:second<.second fraction>] (ISO 8601 extended
```
mweisgut, line 11: timestamp**~t~**

src/lib/utils/date_time_utils.hpp
```diff
@@ -8,19 +8,14 @@
...
+ * format), e.g., "2001-01-01", "2000-01-01 13:01:02" or "2000-01-01 13:01:02.5000". Boost allows out-of-bounds values
+ * for timestamps, e.g., '2000-01-01 25:61:61' is a valid date time and the overflow is added to the subsequent time
```
mweisgut, line 13: Do we want to differentiate here between timestamp and data time? How about `is a valid timestamp`? or `is valid`?

src/lib/utils/date_time_utils.hpp
```diff
@@ -8,19 +8,14 @@
...
+ * unit. In this example, the resulting date time is '2000-01-02 02:02:01'. This behavior is enabled by Boost's time
```
mweisgut, line 14: `resulting timestamp`?

src/lib/sql/sql_translator.cpp
dey4ss: I'll go for the first suggestion. The second one will segfault if the given date is invalid.

mweisgut (approved):

PR #953, opened 2018-07-04
coverage with clang

nilsthamm:
#905 
Enables clang coverage on Linux (to test JIT files). Coverage diff will still be executed with gcc (and only on linux), since llvm does not have a usable export file type (https://llvm.org/docs/CommandGuide/llvm-cov.html#llvm-cov-report)

PR #2483, opened 2022-07-13
Small readme correction

tjjordan:
Since PR #953, the coverage script only supports clang on both Linux and MacOS.
Now this is also reflected in the readme.


README.md
```diff
@@ -127,7 +127,7 @@ Subsets of all available tests can be selected via `--gtest_filter=`.
 ### Coverage
 `./scripts/coverage.sh` will print a summary to the command line and create detailed html reports at ./coverage/index.html
 
-*Supports only clang on MacOS and only gcc on linux*
+*Supports only clang on both MacOS and Linux.*
```
Bouncner, line 130: Does the coverage script "support" clang? Sounds somewhat confusing to me.

"Requires clang on macOS and Linux"?   (macOS is the official name now)

Bouncner (approved):

PR #2485, opened 2022-07-19
[DYOD] UCC Discovery Plugin

LukasBudach:
This PR contains one of two parts of the Data Dependencies Project conducted during the DYOD seminar.

The Unique Column Combination Discovery Plugin is standalone. It provides a user callable function to be invoked. This choice was made to avoid slowing down the system if it is already under high load, which could've been the case if the plugin were to be run periodically.

The Plugin checks the current LQP cache for all its queries and determines candidate columns for uniqueness validation. Candidates are chosen if a unique constraint on them may have a positive impact on query optimization.

A missing component is attaching the information of when a UCC was validated to the constraint that is being set. This is required to ensure that Uniqueness Constraints are not assumed to be true after commits have changed values in the column, making it non-unique.


src/lib/storage/table_key_constraint.cpp
```diff
@@ -2,13 +2,17 @@
 
 namespace opossum {
 
-TableKeyConstraint::TableKeyConstraint(std::unordered_set<ColumnID> init_columns, KeyConstraintType init_key_type)
-    : AbstractTableConstraint(std::move(init_columns)), _key_type(init_key_type) {}
+TableKeyConstraint::TableKeyConstraint(std::unordered_set<ColumnID> init_columns, KeyConstraintType init_key_type, std::optional<CommitID> latest_validated_commit/* = std::nullopt*/)
```
Bouncner, line 5: Oh like that for default arguments. Nice.

tjjordan (commented):
Overall, really nice work!
Your code is really gut readable and well commented.
I have not noticed any major issues in your implementation.
Anyway, here are some smaller annotations, mostly regarding the style guide or minor performance improvements.

src/lib/storage/pos_lists/entire_chunk_pos_list.hpp
```diff
@@ -21,7 +21,7 @@ class EntireChunkPosList : public AbstractPosList {
 
   // Implemented in hpp for performance reasons (to allow inlining)
   RowID operator[](const size_t index) const final {
-    return RowID{_common_chunk_id, static_cast<ChunkOffset>(index)};
+    return RowID{_common_chunk_id, ChunkOffset{static_cast<ChunkOffset::base_type>(index)}};
```
twollnik, line 24: Why is this needed?

src/lib/storage/table_key_constraint.cpp
```diff
@@ -2,13 +2,20 @@
 
 namespace opossum {
 
-TableKeyConstraint::TableKeyConstraint(std::unordered_set<ColumnID> init_columns, KeyConstraintType init_key_type)
-    : AbstractTableConstraint(std::move(init_columns)), _key_type(init_key_type) {}
+TableKeyConstraint::TableKeyConstraint(std::unordered_set<ColumnID> init_columns, KeyConstraintType init_key_type,
+                                       std::optional<CommitID> latest_validated_commit /* = std::nullopt*/)
```
twollnik, line 6: Is this a debugging artifact? 
```suggestion
                                       std::optional<CommitID> latest_validated_commit)
```

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
+/*
+ *  In this plugin, we implement one way for UCC dependency discovery in table columns.
+ *  In general, sort and adjacent_find are used for duplicate detection. The sort is optimized for sorted sub-vectors which are merged as a whole.
+ *  Early outs are exploited for dictionary segments.
+ */
```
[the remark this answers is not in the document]
twollnik: Please be either (1) more specific here and explain why it is important to use sort and adjacend_find or (2) mention these details more closely to the code that implements them. This comment applies to all sentences except for the first one.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+namespace opossum {
+
+std::string UccDiscoveryPlugin::description() const {
+  return "Unary Unique Column Combination Discovery Plugin";
+}
+
+void UccDiscoveryPlugin::start() {}
+
+void UccDiscoveryPlugin::discover_uccs() const {
```
twollnik, line 33: This function is ~100 lines long and deeply nested. The deep nesting concerns me more than the length of the function. Please simplify to improve readability.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
...
+      }
+
+      return LQPVisitation::VisitInputs;
+    });
+  }
+
+  return candidates;
+}
+
+UCCCandidates UccDiscoveryPlugin::identify_ucc_candidates() const {
```
twollnik, line 202: This function is well over 100 lines long and also deeply nested. Please simplify.

src/plugins/ucc_discovery_plugin.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include "utils/pausable_loop_thread.hpp"
+
+namespace opossum {
+
+class UCCCandidate {
+ public:
+  UCCCandidate(std::string table_name, ColumnID column_id) : _table_name(table_name), _column_id(column_id) {}
+
+  const std::string table_name() const {
+    return this->_table_name;
```
twollnik, line 15: Please use `this` as little as possible as specified in the hyrise style guide
```suggestion
    return _table_name;
```

src/plugins/ucc_discovery_plugin.hpp
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  const ColumnID column_id() const {
+    return this->_column_id;
```
twollnik, line 19: Please use `this` as little as possible as specified in the hyrise style guide

```suggestion
    return _column_id;
```

src/plugins/ucc_discovery_plugin.hpp
```diff
@@ -0,0 +1,64 @@
...
+  std::vector<std::pair<PluginFunctionName, PluginFunctionPointer>> provided_user_executable_functions() const final;
+
+ protected:
+  friend class UccDiscoveryPluginTest;
+
+  UCCCandidates identify_ucc_candidates() const;
+  std::shared_ptr<std::vector<UCCCandidate>> generate_valid_candidates(
+      std::shared_ptr<AbstractLQPNode> root_node, std::shared_ptr<LQPColumnExpression> column_candidate) const;
+
+  void discover_uccs() const;
```
twollnik, line 50: why should this function be protected?

src/plugins/ucc_discovery_plugin.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+  void stop() final;
+
+  std::vector<std::pair<PluginFunctionName, PluginFunctionPointer>> provided_user_executable_functions() const final;
+
+ protected:
+  friend class UccDiscoveryPluginTest;
+
+  UCCCandidates identify_ucc_candidates() const;
+  std::shared_ptr<std::vector<UCCCandidate>> generate_valid_candidates(
```
twollnik, line 47: Should these internal functions be prefixed with an underscore?
```suggestion
  UCCCandidates _identify_ucc_candidates() const;
  std::shared_ptr<std::vector<UCCCandidate>> _generate_valid_candidates(
```

src/plugins/ucc_discovery_plugin.hpp
```diff
@@ -0,0 +1,64 @@
...
+};
+
+}  // namespace opossum
+
+template <>
+struct std::hash<opossum::UCCCandidate> {
+  std::size_t operator()(opossum::UCCCandidate const& s) const noexcept {
+    std::size_t h1 = std::hash<std::string>{}(s.table_name());
+    std::size_t h2 = std::hash<opossum::ColumnID>{}(s.column_id());
+    return h1 ^ (h2 << 1);
```
twollnik, line 62: Can you please add a short comment to clarify what is happening in this function?

twollnik (commented):
Thanks for the PR!

Some overall comments:
- Please provide some documentation on how to invoke the user-callable function. E.g. you could add a comment to `ucc_discovery_plugin.hpp` and provide a usage example there. This will be especially helpful for new users who are not already familiar with user-callable functions.
- As far as I understand, the UCC discovery is triggered manually by the user and looks at the LQP cache to determine candidate columns. What about UCC discovery for "cold" databases or for previously unseen workloads?

src/lib/storage/pos_lists/entire_chunk_pos_list.hpp
dey4ss: Recent clang versions require this, they're just the first ones to build Hyrise with clang-14.

src/lib/storage/table_key_constraint.cpp
dey4ss: I somehow like the idea of knowing default arguments in the .cpp files, maybe this annotation is something we can discuss internally in the fortnightly.

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -0,0 +1,191 @@
...
+};
+
+TEST_F(UccDiscoveryPluginTest, LoadUnloadPlugin) {
+  auto& pm = Hyrise::get().plugin_manager;
+  pm.load_plugin(build_dylib_path("libhyriseUccDiscoveryPlugin"));
+  pm.unload_plugin("hyriseUccDiscoveryPlugin");
+}
+
+TEST_F(UccDiscoveryPluginTest, CorrectCandidatesGeneratedForJoin) {
+  Hyrise::get().default_lqp_cache = std::make_shared<SQLLogicalPlanCache>();
```
dey4ss, line 91: You set the chaches in many test cases. Add to `SetUp()`?

src/plugins/ucc_discovery_plugin.cpp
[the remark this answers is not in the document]
dey4ss: Do you really use sorting for duplicate detection rather than sets?

dey4ss (changes requested):
I don't see any behavioral problems with your code, but I have some suggestions in addition to the previous reviews that aim to make your code better understandable.

src/lib/storage/table_key_constraint.cpp
Bouncner: Same for me. Find it annoying to look those values up in the header file.

src/lib/storage/table_key_constraint.cpp
```diff
@@ -2,13 +2,20 @@
...
+    : AbstractTableConstraint(std::move(init_columns)),
+      _key_type(init_key_type),
```
Bouncner, line 8: Not yours, but we actually only use the `init_` prefix in case there is a naming conflict as the member is public. Since key type is private and thus prefixed with `_`, the is no reason to use the `init_` prefix here.

Otherwise it's also not consistent with `latest_validated_commit`.

src/plugins/ucc_discovery_plugin.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+  const std::string table_name() const {
+    return this->_table_name;
+  }
+
+  const ColumnID column_id() const {
+    return this->_column_id;
+  }
+
+  friend bool operator==(const UCCCandidate& lhs, const UCCCandidate& rhs) {
```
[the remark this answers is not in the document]
Bouncner: Good catch, but rhs/lhs is among the few exceptions. We use it quite often in the code base.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
...
+      const auto& join_mode = join_node.join_mode;
+      // get join predicate with equals condition, that's the only one we would want to work on
+      std::shared_ptr<BinaryPredicateExpression> join_predicate = nullptr;
+      for (auto predicate : join_node.join_predicates()) {
+        join_predicate = std::dynamic_pointer_cast<BinaryPredicateExpression>(predicate);
+        if (!join_predicate) {
+          continue;
+        }
+        if (join_predicate->predicate_condition != PredicateCondition::Equals) {
+          join_predicate = nullptr;
```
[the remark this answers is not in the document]
LukasBudach: Looks like we do need that to me -
``join_predicate`` is not a nullprt, but a pointer to a ``BinaryPredicateExpression`` in this case. However, one that does not have an equals condition.
Since we don't want to consider other join conditions than the equality condition (discussion in its own thread), we need to reset the variable to a nullptr in case this was the last iteration, as we will only continue iterating over the LQP if no ``join_predicate`` was found.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
...
+      auto& join_node = static_cast<JoinNode&>(*node);
+      const auto& join_mode = join_node.join_mode;
+      // get join predicate with equals condition, that's the only one we would want to work on
+      std::shared_ptr<BinaryPredicateExpression> join_predicate = nullptr;
+      for (auto predicate : join_node.join_predicates()) {
+        join_predicate = std::dynamic_pointer_cast<BinaryPredicateExpression>(predicate);
+        if (!join_predicate) {
+          continue;
+        }
+        if (join_predicate->predicate_condition != PredicateCondition::Equals) {
```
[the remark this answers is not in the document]
LukasBudach: It should be possible to do that, however, we didn't work out the details of those rewrites. It may be a future expansion on the plugin and rewrite rule though.

It should be fine to remove this condition for the UCC discovery. However, considering the time left as well as the fact that we wouldn't utilize the UCCs discovered in these new cases, I'd keep it this way to avoid some nasty side-effects we didn't think of when dealing with predicates that do not have the equals condition.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
...
+          auto candidates = generate_valid_candidates(subtree_root, column_candidate);
+          if (candidates) {
+            for (const auto& candidate : *candidates) {
+              ucc_candidates.insert(candidate);
+            }
+          }
+          break;
+        }
+
+        default: {
```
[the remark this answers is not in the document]
LukasBudach: Not really needed, as this is what happens in line 335 anyways, without any more code being executed in between.
I do see though, that it may add some layer of safety, if someone were to write code in between the end of the switch case and the return in the future.

Is the (currently unnecessary) code duplication worth that security in the future in your eyes?

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+      if (type == LQPNodeType::Aggregate) {
+        // in case of aggregate, extract all predicates used in groupby operations
+        auto& aggregate_node = static_cast<AggregateNode&>(*node);
+        auto column_candidates = std::vector<std::shared_ptr<AbstractExpression>>{
+            aggregate_node.node_expressions.begin(),
+            aggregate_node.node_expressions.begin() + aggregate_node.aggregate_expressions_begin_idx};
+
+        for (const auto& column_candidate : column_candidates) {
+          const auto casted_candidate = std::dynamic_pointer_cast<LQPColumnExpression>(column_candidate);
```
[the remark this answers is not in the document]
LukasBudach: Which approach would you recommend between dynamic cast and the typecheck + static cast?
We've pretty consistently used dynamic casts in our code I believe.

src/plugins/ucc_discovery_plugin.cpp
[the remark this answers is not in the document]
dey4ss: Can you add a comment or mark this as TODO? Totally fine that this is out of scope when you document it.

src/plugins/ucc_discovery_plugin.cpp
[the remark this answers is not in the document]
dey4ss: This was related to my previous comment, keeping it this way and adding a comment is fine

src/plugins/ucc_discovery_plugin.cpp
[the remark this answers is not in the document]
dey4ss: I think it's better not to leave a case empty. If I'm not missing something on a quick look, you can also think of replacing your `break`s by `return`s and be a little more explicit there (so one directly sees what happens next without scrolling).

src/plugins/ucc_discovery_plugin.cpp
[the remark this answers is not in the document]
dey4ss: We're not consistent in the codebase tbh. So I guess use what you prefer. However letting something crash when you expect something to have a certain type is more explicit.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
...
+      const auto expression_table =
+          std::static_pointer_cast<const StoredTableNode>(column_expression->original_node.lock());
+      const auto candidate_table =
+          std::static_pointer_cast<const StoredTableNode>(column_candidate->original_node.lock());
+
+      if (expression_table == candidate_table) {
+        // both columns same table -> if both are UCC we could still convert join -> both are UCC candidate
+        candidates->push_back(UCCCandidate{expression_table->table_name, column_expression->original_column_id});
+        candidates->push_back(UCCCandidate{candidate_table->table_name, column_candidate->original_column_id});
+      }
```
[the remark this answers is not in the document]
LukasBudach: ``column_expression`` and ``column_candidate`` are no longer compared.

``column_candidate`` is added to the UCC candidates anyways, as it is the join column for the table that could be removed, so that piece of code just inserted it again.

While we still insert the join column twice (as the table comparison still returns true in case the join column is also used in a predicate), we do deduplication of the UCC candidates anyways, so that isn't too bad.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
...
+        column_expression = std::dynamic_pointer_cast<LQPColumnExpression>(predicate->right_operand());
+        value_expression = std::dynamic_pointer_cast<ValueExpression>(predicate->left_operand());
+      }
+
+      if ((!column_expression) || (!value_expression)) {
+        // predicate needs to look like column = value or value = column; if not, move on
+        return LQPVisitation::VisitInputs;
+      }
+
+      if (column_expression == column_candidate) {
```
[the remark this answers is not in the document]
LukasBudach: You are right - this particular comparison is not needed, as comparing the tables does the same thing, but also includes columns that are not equal to the join column (``column_candidate``).

However, the stored table is only resolved in the case of an Inner Join, and the join column is also only added in the case of an inner join, as only for those the semi-join rewrite can be performed, and for the predicate rewrite we need not only the join column but also a matching predicate.
Therefore, I'd stick with passing the node and resolving the table in this function if needed.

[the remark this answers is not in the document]
dey4ss: alright

src/plugins/ucc_discovery_plugin.hpp
LukasBudach: The plugin provides a user-callable function interface.
It is not supposed tp be called from the code, so making its functions protected/private seems fair.

The two protedted functions (_identify_ucc_candidates and _validate_ucc_candidates, the prior discover_uccs) are protected and not private, as the test needs to be able to call those.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
...
+        return LQPVisitation::VisitInputs;
+      }
+
+      // get the column expression, should be left, but also check the right operand if the left one is not column
+      auto column_expression = std::dynamic_pointer_cast<LQPColumnExpression>(predicate->left_operand());
+      auto value_expression = std::dynamic_pointer_cast<ValueExpression>(predicate->right_operand());
+      if (!column_expression) {
+        column_expression = std::dynamic_pointer_cast<LQPColumnExpression>(predicate->right_operand());
+        value_expression = std::dynamic_pointer_cast<ValueExpression>(predicate->left_operand());
+      }
```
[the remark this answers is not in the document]
LukasBudach: So, the code below works, however, despite using the ``visit_expression`` function it is longer, requires more casts and is, in my opinion, not much easier to understand.

The issue is, that we already casted the ``predicate`` to a ``BinaryPredicateExpression``, as we need to know whether the predicate is using the == condition. However, the ``visit_expression`` function requires a shared pointer to an ``AbstractExpression``, so it throws compilation errors when simply plugging in the previously casted ``predicate``.

The function itself is then fine, but as I said - I don't really see the advantage over the swapping which side of the operand to cast.

```
std::shared_ptr<LQPColumnExpression> column_expression = nullptr;
std::shared_ptr<ValueExpression> value_expression = nullptr;

const auto abstreact_expression_predicate = std::dynamic_pointer_cast<AbstractExpression>(predicate);
visit_expression(abstreact_expression_predicate, [&](auto& sub_expression) {
  if (sub_expression->type == ExpressionType::LQPColumn) {
    column_expression = std::static_pointer_cast<LQPColumnExpression>(sub_expression);
  } else if (sub_expression->type == ExpressionType::Value) {
    value_expression = std::static_pointer_cast<ValueExpression>(sub_expression);
  }
  return ExpressionVisitation::VisitArguments;
});
```

As for the loop - I am not quite sure whether that would improve the code either, since we already know that there will be a left and a right operand and that they have to be one column and one value expression.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,345 @@
...
+          all_values->insert(begin(*dict), end(*dict));
+          if (all_values->size() == all_values_size + dict->size()) {
+            all_values_size += dict->size();
+          } else {
+            // If not all elements have been inserted, there be a duplicate, so the UCC constraint is violated
+            message << " [rejected in " << format_duration(candidate_time.lap()) << "]";
+            Hyrise::get().log_manager.add_message("UccDiscoveryPlugin", message.str(), LogLevel::Info);
+            return;
+          }
+        } else {
```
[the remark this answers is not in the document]
LukasBudach: This code should work:

```
bool duplicates_encountered = false;
segment_iterate<ColumnDataType>(*source_segment, [&](const auto& position) {
  // no special handling for null values, as they are not excluded from the check for other segment types
  all_values.insert(position.value());
  if (all_values.size() == all_values_size + 1) {
    all_values_size += 1;
  } else {
    // the element was already in the all_values set, therefore is a duplicate - UCC violated
    duplicated_encountered = true;
  }
});
if (duplicates_encountered) {
  return false;
}
```

However, I get the following compiler error, and this is just no longer the time to deal with that sadly.

```
Building CXX object src/plugins/CMakeFiles/hyriseUccDiscoveryPlugin.dir/ucc_discovery_plugin.cpp.o
FAILED: src/plugins/CMakeFiles/hyriseUccDiscoveryPlugin.dir/ucc_discovery_plugin.cpp.o 
/usr/bin/c++ -DBOOST_THREAD_VERSION=5 -DHYRISE_DEBUG=1 -DHYRISE_NUMA_SUPPORT=1 -DTEST_PLUGIN_DIR=\"/home/Lukas.Budach/hyrise/cmake-build-release/lib/\" -DhyriseUccDiscoveryPlugin_EXPORTS -I/home/Lukas.Budach/hyrise/src/benchmarklib -I/home/Lukas.Budach/hyrise/src/lib -I/home/Lukas.Budach/hyrise/src/plugins -I/home/Lukas.Budach/hyrise/third_party/sql-parser/src -I/home/Lukas.Budach/hyrise/third_party/magic_enum/include -g -fopenmp-simd -O0 -fPIC -DBOOST_ASIO_DISABLE_CONCEPTS -pthread -Wno-unknown-warning-option -Wall -Wextra -pedantic -Werror -Wno-unused-parameter -Wno-dollar-in-identifier-extension -Wno-unknown-pragmas -Wno-subobject-linkage -Wno-deprecated-dynamic-exception-spec -std=gnu++20 -MD -MT src/plugins/CMakeFiles/hyriseUccDiscoveryPlugin.dir/ucc_discovery_plugin.cpp.o -MF src/plugins/CMakeFiles/hyriseUccDiscoveryPlugin.dir/ucc_discovery_plugin.cpp.o.d -o src/plugins/CMakeFiles/hyriseUccDiscoveryPlugin.dir/ucc_discovery_plugin.cpp.o -c /home/Lukas.Budach/hyrise/src/plugins/ucc_discovery_plugin.cpp
In file included from /home/Lukas.Budach/hyrise/src/lib/storage/vector_compression/bitpacking/bitpacking_decompressor.hpp:3,
                 from /home/Lukas.Budach/hyrise/src/lib/storage/vector_compression/bitpacking/bitpacking_vector.hpp:3,
                 from /home/Lukas.Budach/hyrise/src/lib/storage/vector_compression/resolve_compressed_vector_type.hpp:7,
                 from /home/Lukas.Budach/hyrise/src/lib/storage/dictionary_segment/dictionary_segment_iterable.hpp:9,
                 from /home/Lukas.Budach/hyrise/src/lib/storage/create_iterable_from_segment.ipp:3,
                 from /home/Lukas.Budach/hyrise/src/lib/storage/create_iterable_from_segment.hpp:82,
                 from /home/Lukas.Budach/hyrise/src/lib/storage/segment_iterate.hpp:6,
                 from /home/Lukas.Budach/hyrise/src/plugins/ucc_discovery_plugin.cpp:15:
/home/Lukas.Budach/hyrise/src/lib/storage/vector_compression/bitpacking/bitpacking_vector_type.hpp:3:10: fatal error: compact_vector.hpp: No such file or directory
    3 | #include "compact_vector.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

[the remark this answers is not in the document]
dey4ss: I think you just need to add `compact_vector` as a dependency in the CMakelists, similarly to sqlparser and magic_enum

src/plugins/ucc_discovery_plugin.hpp
```diff
@@ -0,0 +1,124 @@
...
+ */
+struct UCCCandidate {
+  UCCCandidate(const std::string& init_table_name, const ColumnID init_column_id);
+
+  bool operator==(const UCCCandidate& other) const;
+  bool operator!=(const UCCCandidate& other) const;
+  size_t hash() const;
+
+  const std::string table_name;
+  const ColumnID column_id;
```
Bouncner, line 25: Initialize or disable default constr?

src/plugins/ucc_discovery_plugin.hpp
```diff
@@ -0,0 +1,124 @@
...
+}  // namespace hyrise
+
+namespace std {
+
+/**
+ * Hash function required to manage UCCCandidates in an unordered set for deduplication.
+ */
+template <>
+struct hash<hyrise::UCCCandidate> {
+  size_t operator()(const hyrise::UCCCandidate& uc) const;
```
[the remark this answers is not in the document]
Bouncner: ?

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+      // We save UCC constraints directly inside the table so they can be forwarded to nodes in a query plan.
+      message << " [confirmed in " << format_duration(candidate_time.lap()) << "]";
+      Hyrise::get().log_manager.add_message("UccDiscoveryPlugin", message.str(), LogLevel::Info);
+      table->add_soft_key_constraint(TableKeyConstraint({col_id}, KeyConstraintType::UNIQUE));
+    });
+  }
+  Hyrise::get().log_manager.add_message("UccDiscoveryPlugin", "Clearing LQP and PQP cache...", LogLevel::Debug);
+
+  Hyrise::get().default_lqp_cache->clear();
```
Bouncner, line 129: Why clean? I am curious as the plan cache is the single source of information for multiple "autonomous components".

dey4ss: Without cleaning or replacing the cache, optimized versions using UCCs will not be applied when a query is executed before and after discovery.

Bouncner: Makes sense. But I don't like the answer.  :D

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,369 @@
...
+}
+
+void UccDiscoveryPlugin::_ucc_candidates_from_join_node(std::shared_ptr<AbstractLQPNode> node,
+                                                        UccCandidates& ucc_candidates) {
+  const auto& join_node = static_cast<JoinNode&>(*node);
+  // Fetch the join predicate to extract the UCC candidates from. Right now, limited to only equals predicates.
+  const auto& join_predicates = join_node.join_predicates();
+  const auto binary_join_predicate = (join_predicates.size() == 1)
+                                         ? std::dynamic_pointer_cast<BinaryPredicateExpression>(join_predicates[0])
+                                         : std::shared_ptr<BinaryPredicateExpression>{};
```
mweisgut, line 250:
```suggestion
                                         : nullptr;
```

mweisgut (changes requested):
I haven't reviewed the tests yet.

mweisgut (commented):
The existing tests look good to me. I haven't checked the line coverage yet. It would be insightful to add plugin code coverage to the llvm coverage report.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -0,0 +1,369 @@
...
+    return;
+  }
+
+  const auto column_expression_for_join_input = [&](const auto& input_node) {
+    // The join redicate may be swapped, so get proper operand.
+    auto column_candidate = std::dynamic_pointer_cast<LQPColumnExpression>(binary_join_predicate->left_operand());
+    if (!column_candidate || !expression_evaluable_on_lqp(column_candidate, *input_node)) {
+      column_candidate = std::dynamic_pointer_cast<LQPColumnExpression>(binary_join_predicate->right_operand());
+    }
+    Assert(!column_candidate || expression_evaluable_on_lqp(column_candidate, *input_node),
```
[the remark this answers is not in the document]
dey4ss: This is no bug, indeed. Imagine the following queries:
```SQL
SELECT ... foo, bar
WHERE foo.x + 1 = foo.y + 2
```
or from the TPC-H data:
```SQL
SELECT ... from orders, lineitem
WHERE EXTRACT(YEAR FROM o_orderdate) = EXTRACT(YEAR FROM l_shipdate)
```
Neither of the join expressions is an LQPColumnExpression. However, I'll add a comment and a test for this.

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -0,0 +1,257 @@
...
+class UccDiscoveryPluginMultiEncodingTest : public UccDiscoveryPluginTest,
+                                            public ::testing::WithParamInterface<SegmentEncodingSpec> {};
+
+TEST_F(UccDiscoveryPluginTest, LoadUnloadPlugin) {
+  auto& plugin_manager = Hyrise::get().plugin_manager;
+  plugin_manager.load_plugin(build_dylib_path("libhyriseUccDiscoveryPlugin"));
+  plugin_manager.unload_plugin("hyriseUccDiscoveryPlugin");
+}
+
+TEST_F(UccDiscoveryPluginTest, UserCallableFunction) {
```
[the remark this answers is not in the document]
dey4ss: Added a test for both, also a test for `MvccDeletePlugin::description()`. However, I could not convince llvm-cov to show coverage information for the plugins.

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -0,0 +1,257 @@
...
+  std::shared_ptr<LQPColumnExpression> _predicate_column_A{};
+  std::shared_ptr<LQPColumnExpression> _predicate_column_B{};
+};
+
+class UccDiscoveryPluginMultiEncodingTest : public UccDiscoveryPluginTest,
+                                            public ::testing::WithParamInterface<SegmentEncodingSpec> {};
+
+TEST_F(UccDiscoveryPluginTest, LoadUnloadPlugin) {
+  auto& plugin_manager = Hyrise::get().plugin_manager;
+  plugin_manager.load_plugin(build_dylib_path("libhyriseUccDiscoveryPlugin"));
```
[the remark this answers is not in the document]
dey4ss: Did the same for the corresponding `MvccDeletePlugin` test.

src/plugins/ucc_discovery_plugin.cpp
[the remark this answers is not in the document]
mweisgut: Thanks for the clarification.

dey4ss:
Measurements with mined UCCs can be found here: https://github.com/hyrise/hyrise/pull/2486#issuecomment-1308708414

mweisgut (dismissed):

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -243,12 +243,12 @@ void UccDiscoveryPlugin::_ucc_candidates_from_aggregate_node(std::shared_ptr<Abs
...
+  // Fetch the join predicate to extract the UCC candidates from. Right now, limited to single-predicate equi joins.
   const auto& join_predicates = join_node.join_predicates();
-  const auto binary_join_predicate = (join_predicates.size() == 1)
-                                         ? std::dynamic_pointer_cast<BinaryPredicateExpression>(join_predicates[0])
-                                         : std::shared_ptr<BinaryPredicateExpression>{};
-
+  if (join_predicates.size() != 1) {
+    return;
+  }
+  const auto binary_join_predicate = std::dynamic_pointer_cast<BinaryPredicateExpression>(join_predicates[0]);
```
mweisgut: `join_predicates.front()`

dey4ss: Was done by merging your suggestions in [251c0a8](https://github.com/hyrise/hyrise/pull/2485/commits/251c0a87e1c8bc67cc88b9e66d06c3b1a2d21d27)

mweisgut (approved):

dey4ss (approved):

PR #2497, opened 2022-08-11
Refactor SortedSegmentSearch and add information to performance data 

dey4ss:
#2283 tracked SortedSegmentSearch's shortcuts (each/none value matches) for ColumnBetweenScans. This PR implements and tracks them for ColumnVsValueScans. Furthermore, the SortedSegmentSearch is refactored: (i) strict order of private/public methods, (ii) simplification of the public interface, and (iii) some readability improvements.

Besides, subsequently flipping OperatorJoinPredicates leads to alternating the flag indicating they were flipped. IndexJoins add a flag indicating the side of the index to their performance data.


dey4ss:
Note that a quick search for wrongly ordered access specifiers (e.g., `public` after `protected` or `private`) gave quite some results. Once #2402 is merged, I'm gonna have look into these cases and refactor if necessary. For instance, my search did not account for the decalration of multiple (nested) classes/structs in a single header file.

src/test/lib/operators/table_scan_sorted_segment_search_test.cpp
```diff
@@ -78,57 +92,99 @@ INSTANTIATE_TEST_SUITE_P(
...
-        ASSERT_EQ(output_begin->value(), _expected[index]) << "row " << index << " is invalid";
-      }
-    });
+        _second_search_value
+            ? SortedSegmentSearch(input_begin, input_end, _sorted_by, _nullable, _predicate_condition, _search_value,
+                                  *_second_search_value)
+            : SortedSegmentSearch(input_begin, input_end, _sorted_by, _nullable, _predicate_condition, _search_value);
+    auto matches = RowIDPosList{};
+    const auto initial_chunk_id = ChunkID{0};
+    sorted_segment_search.scan_sorted_segment(initial_chunk_id, matches, nullptr);
```
Bouncner, line 174: Do we test the early out for "all NULLs"?

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -78,53 +103,82 @@ class SortedSegmentSearch {
...
-          return;
-        case PredicateCondition::GreaterThanEquals:
-          _end = _get_last_bound(_first_search_value, _begin, _end);
-          return;
-        case PredicateCondition::GreaterThan:
-          _end = _get_first_bound(_first_search_value, _begin, _end);
-          return;
-        default: Fail("Unsupported predicate condition encountered");
-      }
+    auto first_value = _begin->value();
```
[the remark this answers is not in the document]
dey4ss: We swap first and last value and flip the pr3edicate when not sorted ascending. Doing so, we don't have to handle descending sorted segments. No const though.

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -155,23 +209,35 @@ class SortedSegmentSearch {
...
+        upper_exclusive ? last_value < *_second_search_value : last_value <= *_second_search_value;
+    if (first_value_matches && last_value_matches) {
       all_rows_matching = true;
       return;
     }
 
-    // early out nothing matches
-    if (first_value > *_second_search_value || last_value < _first_search_value) {
+    // Early out if nothing matches.
+    const auto first_value_matches_not =
```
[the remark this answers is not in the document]
dey4ss: Hm, the first and second search vales are the upper and lower bound of the between range. So `lower_bound_larger_than_segment_max`, `upper_boubd_smaller_than_segment_min`?

src/test/lib/operators/table_scan_sorted_segment_search_test.cpp
dey4ss: Good catch, gonna check that `no_rows_matching` is set.

src/lib/operators/table_scan/sorted_segment_search.hpp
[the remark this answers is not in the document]
Bouncner: I just typed a random string that tells more. Maybe `lower_bound_larger_than_segment` is already clear enough?

src/lib/operators/table_scan/sorted_segment_search.hpp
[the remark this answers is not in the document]
Bouncner: No, if they are not const I am fine with keeping it that way. The current version is nice to read, no need to change it.

Still would rename `first_value` etc.

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -40,6 +40,31 @@ class SortedSegmentSearch {
...
+    if (_second_search_value) {
+      _set_begin_and_end_positions_for_between_scan();
+    } else {
+      _set_begin_and_end_positions_for_vs_value_scan();
+    }
+    _write_rows_to_matches(_begin, _end, chunk_id, matches, position_filter);
+  }
+
+  // Flags to indicate whether a shortcut was taken to skip scanning.
+  bool no_rows_matching{false};
```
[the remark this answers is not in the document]
dey4ss: clang really does not like this

src/test/lib/operators/table_scan_sorted_segment_search_test.cpp
dey4ss: I added checks that `all_rows_matching` is set when `matches.size() == _expected.size()` and `no_rows_matching` is set when `matches.empty()`

src/lib/operators/table_scan/sorted_segment_search.hpp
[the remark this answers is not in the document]
Bouncner: I did not realize that it's a member def here. Sorry.

src/test/lib/operators/table_scan_sorted_segment_search_test.cpp
Bouncner: Nice.

And is there a test case that explicitely tests for all values being NULL? I haven't found this case when skipping over this test file here.

dey4ss: I changed the testing parameter that switches the segment's nullability from a bool to an enum with three values. One of these enum values implies that all values should be NULL and non-null values are not appended to the segment. When this flag is set, we do not check whether the scan result is equal to the expected test data but check if the scan returns an empty PosList.

dey4ss:
To ensure nothing gets worse with this PR, I started multiple benchmarks:
1. `benchmark_all.sh` PR vs master. Sanity check, nothing gets marked as sorted and SortedSgmentSearch does not execute.
2. TPC-H, TPC-DS, and JOB PR vs master with sorted columns (sort orders in table generators). To determine the columns, I counted value and between scans and picked the columns with the most scans. These are:
    ```
    ====================
    TPCH
    ====================
    Value
        8  part.p_brand
    
    Between
        6  lineitem.l_shipdate
        4  orders.o_orderdate
    
    ====================
    TPCDS
    ====================
    Value
        54  customer_address.ca_state
        29  household_demographics.hd_dep_count
        14  date_dim.d_moy
        9  time_dim.t_hour
        9  store.s_store_name
        4  item.i_manager_id
    
    Between
        9  store_sales.ss_sales_price
        9  date_dim.d_moy
        6  web_sales.ws_net_profit
    
    ====================
    JOB
    ====================
    Value
        191  movie_info.info
        96  info_type.info
        64  company_name.country_code
        56  title.production_year
        47  movie_companies.note
        47  keyword.keyword
        42  comp_cast_type.kind
        40  cast_info.note
        38  kind_type.kind
        29  name.gender
        25  company_type.kind
        21  movie_info_idx.info
        19  role_type.role
        17  link_type.link
        16  char_name.name
    
    Between
        24  title.production_year
    
    ```
3. PR with these sorted columns and `--verify`

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -78,64 +110,85 @@ class SortedSegmentSearch {
...
+        return;
+      default:
+        Fail("Unsupported predicate condition encountered");
     }
-    // clang-format on
   }
 
   // This function sets the offset(s) that delimit the result set based on the predicate condition and the sort order
   void _set_begin_and_end_positions_for_between_scan() {
     DebugAssert(_second_search_value, "Second Search Value must be set for between scan");
```
tjjordan, line 183: Maybe this should be asserted in release builds, too.

src/test/lib/operators/operator_join_predicate_test.cpp
```diff
@@ -49,4 +49,21 @@ TEST_F(OperatorJoinPredicateTest, FromExpressionImpossible) {
...
 
+TEST_F(OperatorJoinPredicateTest, Flip) {
+  auto predicate = OperatorJoinPredicate{{ColumnID{0}, ColumnID{1}}, PredicateCondition::LessThanEquals};
+  EXPECT_FALSE(predicate.is_flipped());
+
+  predicate.flip();
+  EXPECT_TRUE(predicate.is_flipped());
+  EXPECT_EQ(predicate.column_ids.first, ColumnID{1});
+  EXPECT_EQ(predicate.column_ids.second, ColumnID{0});
+  EXPECT_EQ(predicate.predicate_condition, PredicateCondition::GreaterThanEquals);
```
tjjordan, line 60: Maybe the flip should be tested for some more PredicateConditions.

src/lib/operators/join_index.hpp
```diff
@@ -44,6 +44,9 @@ class JoinIndex : public AbstractJoinOperator {
   struct PerformanceData : public OperatorPerformanceData<OperatorSteps> {
     void output_to_stream(std::ostream& stream, DescriptionMode description_mode) const override;
 
+    // Per default, the right input is the index side and the left side is the probe side.
+    bool right_input_is_index_side{true};
```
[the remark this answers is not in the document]
dey4ss: Just consistency with `JoinHash::PerformanceData` (esp. `JoinHash::PerformanceData.left_input_is_build_side`) and `TableScan::PerformanceData` that stick to primitives and initialize them accordingly.

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -40,6 +40,38 @@ class SortedSegmentSearch {
         _nullable{nullable},
         _is_ascending{sorted_by == SortMode::Ascending} {}
 
+  void scan_sorted_segment(const ChunkID chunk_id, RowIDPosList& matches,
+                           const std::shared_ptr<const AbstractPosList>& position_filter) {
```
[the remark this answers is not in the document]
dey4ss: Some methods (constructors and `_write_rows_to_matches`) cannot be moved to a .cpp without explicitly instantating them for the different segment itearators, which is not nice and has to be updated whenever we decide to add something there. However, the rest can be moved.

src/lib/operators/table_scan/sorted_segment_search.hpp
dey4ss: We call `_set_begin_and_end_positions...` internally depending on the existence of `_second_search_value`. As this method cannot be called externally without that parameter, I don't see a reason for an `Assert` tbh,

src/test/lib/operators/operator_join_predicate_test.cpp
dey4ss: Not sure if that fits here. Flipping the predicate conditions is implemented in `types.hpp/.cpp` and maybe we could tests for the methods defined there. However, the tests would be pretty much 1:1 reimplementation, as we do simple mappings there.

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -196,74 +263,64 @@ class SortedSegmentSearch {
 
   /*
    * NotEquals may result in two matching ranges (one below and one above the search_value) and needs special handling.
-   * The function contains four early outs. These are all only for performance reasons and, if removed, would not
+   * The function contains five early outs. These are all only for performance reasons and, if removed, would not
```
Bouncner, line 266: Counting 6x `return;` here. How do you count the five?

src/lib/operators/table_scan/sorted_segment_search.hpp
```diff
@@ -303,13 +359,8 @@ class SortedSegmentSearch {
     }
   }
 
-  // Flags to indicate whether a shortcut was taken to skip scanning.
-  bool no_rows_matching{false};
```
Bouncner, line 307: Lines above: we typically avoid increments as part of other statements (here: `matches[output_idx++]`). As you are working on the file: could you move a line below (or in the loop header, I am fine with that as well). The compiler will optimize it ÷anyways.

Bouncner (approved):
Just two tiny issues.

Apart from that the PR looks good to me. It also looks as if the the points raised by @tjjordan have all been addressed, right?

dey4ss:
> It also looks as if the the points raised by @tjjordan have all been addressed, right?

I answered all of them and got 👍, so I guess it should be fine.

PR #2499, opened 2022-08-17
Link profiling libraries for coverage builds

### LINKED ISSUE #2426: `coverage.sh` fails on macOS
Running `./scripts/coverage.sh` on macOS BigSur fails with llvm 12.0.1
```
dyld: Symbol not found: ___llvm_profile_runtime
  Referenced from: [...]/hyrise/build-coverage/lib/libhyriseMvccDeletePlugin.dylib
  Expected in: flat namespace
 in [...]/hyrise/build-coverage/lib/libhyriseMvccDeletePlugin.dylib
./scripts/coverage.sh: line 51:  3419 Abort trap: 6           ./build-coverage/hyriseTest build-coverage --gtest_filter=-SQLiteTestRunnerInstances/*
```

dey4ss:
fixes #2426


dey4ss:
Interestingly, the linker flags are aleardy set in `src/CMakeLists.txt`. Gonna have a more detailed look at it.

Bouncner:
Any chance to check if it runs on macOS?

dey4ss:
Both in the coverage script and the cmake lists we can check the OS type, if that's what you mean.

Bouncner:
No, I was just curious of you have run the script on macOS since the addressed issue by @mweisgut was about macOS.

If it runs, I don't see a reason to differentiate between the OSs since the linking flag makes sense imo.

dey4ss:
> Still, i get the warning warning: 1 functions have mismatched data and the coverage totals are not equal to the CI's version (clang 14.06/macOS 12.4).

Linking is successful on macOS, but the script runs with a warning.

dey4ss:
I ran llvm-cov with `-dump` and I got the following:
```
warning: 1 functions have mismatched data
hash-mismatch: No profile record found for '_ZNK6hyrise14AbstractPlugin34provided_user_executable_functionsEv' with hash = 0x0
```
After some research, it seems that this warning can stem from functions that are never instantiated. Thus, I added a cpp for `AbstractPlugin` and instantiated the function. The warning did not came up then.

Bouncner:
So everything is fine now?

Bouncner (approved):

PR #2501, opened 2022-08-23
Rename misnamed BaseCompressedVector and Corresponding Files and maybe more

tjjordan:


dey4ss:
This is all about `BaseCompressedVector` &rarr; `AbstractCompressedVector` (and corresponding `*Decompressor`), right? Maybe use a more specific PR name.

dey4ss (approved):
Pedantic, but you *rename* the classes rather than *refactoring* their implementations. LGTM. However, I think it's fair to merge #2402 first.

dey4ss:
Just came across `BaseSegmentEncoder`/`SegmentEncoder` (see `storage/base_segment_encoder.hpp`). Would this class fit into this PR as well?

tjjordan:
Good catch!
But `BaseSegmentEncoder` is only an non-templated interface for `SegmentEncoder`, so Base should be the correct prefix.

dey4ss:
Maybe we should discuss our definitions of `Base*`/`Abstract*` in a fortnightly. Currently we have:
 | Name      | Meaning                                                                                                                                                                                                                                                |
| --------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| Abstract* | An incomplete class used for *dynamic* polymorphism (i.e., OOP with run-time resolved virtual methods). It has pure virtual methods that need to be implemented by subclasses.                                                                         |
| Base*     | An incomplete class used for *static* polymorphism (i.e., compile-time resolved templates). It implements methods shared by all implementations. Mostly, it is used so that we have a non-templated superclass, which can be stored in, e.g., a vector. |

`BaseSegmentencoder`: &#x2611; static polymorphism, &#x2612; methods shared by implementations.

mweisgut:
Let's discuss this soon again, @Bouncner @tjjordan @dey4ss @benrobby @mweisgut.

mweisgut:
> Maybe we should discuss our definitions of `Base*`/`Abstract*` in a fortnightly. Currently we have:
> 
> Name	Meaning
> Abstract*	An incomplete class used for _dynamic_ polymorphism (i.e., OOP with run-time resolved virtual methods). It has pure virtual methods that need to be implemented by subclasses.
> Base*	An incomplete class used for _static_ polymorphism (i.e., compile-time resolved templates). It implements methods shared by all implementations. Mostly, it is used so that we have a non-templated superclass, which can be stored in, e.g., a vector.
> `BaseSegmentencoder`: ☑ non-templated, ☒ methods shared by implementations.

I suggest moving the discussion to a new issue and renaming `BaseCompressedVector` with this PR.

mweisgut:
Closing it for now. Feel free to reopen if it becomes a priority.

PR #2505, opened 2022-09-05
Fix SegmentEncoder issue wrongly initializing the _null_values optional of ValueSegments

tjjordan:
This PR fixes bug in the ´SegmentEncoder´.
I'm not sure if there is a corresponding issue.

The error was that when a not nullable encoded segment was passed to the segment encoder and `Unencoded` was selected as the target encoding, the new `ValueSegment` is created with an empty, but initialized `null_values` vector.
Because of that, the `_null_values` optional is set and `is_nullable` returned true, which resulted in errors.

A way to recreate this error on the master branch is to lead encoded tables in an unencoded benchmark:
First run `./hyriseBenchmarkTPCH -s 0.01 -r 1`, after that run `./hyriseBenchmarkTPCH -s 0.01 -r 1 -e Unencoded`.

### DIFF
--- a/src/lib/storage/chunk_encoder.cpp
+++ b/src/lib/storage/chunk_encoder.cpp
@@ -54,6 +54,7 @@ std::shared_ptr<AbstractSegment> ChunkEncoder::encode_segment(const std::shared_
     if (encoding_spec.encoding_type == EncodingType::Unencoded) {
       pmr_vector<ColumnDataType> values;
       pmr_vector<bool> null_values;
+      auto nullable = bool{false};
 
       auto iterable = create_any_segment_iterable<ColumnDataType>(*segment);
       iterable.with_iterators([&](auto it, const auto end) {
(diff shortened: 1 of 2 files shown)


src/lib/storage/chunk_encoder.cpp
```diff
@@ -54,6 +54,7 @@ std::shared_ptr<AbstractSegment> ChunkEncoder::encode_segment(const std::shared_
     if (encoding_spec.encoding_type == EncodingType::Unencoded) {
       pmr_vector<ColumnDataType> values;
       pmr_vector<bool> null_values;
+      auto nullable = bool{false};
```
Bouncner, line 57: I think the term `nullable` is confusing here. Maybe `contains_nulls`?

A segment without NULLs can still be nullable (let's ignore MVCC/append-only). At least we usually use these terms this way.

Bouncner:
And please change the PR title to something more related to the fix.

PR #2507, opened 2022-09-20
Integrate ShellCheck in linting script

tjjordan:


scripts/lint.sh
```diff
@@ -94,4 +94,16 @@ if [ ! -z "$output" ]; then
 	exitcode=1
 fi
 
+# ShellCheck linting
+for file in `find . -name "*.sh"`
+do
+	if ! [[ $file == ./third_party/* ]] && ! [[ $file == ./cmake-* ]] && ! [[ $file == ./clang-* ]];
```
mweisgut, line 100: Is `! [[ $file == ./clang-* ]]` used to exclude clang build directories? Build directories can be named arbitrarily so they wouldn't be generally excluded in any cases.

How about including certain directories instead? For Hyrise, scripts are expected to be located in the scripts directory. In addition, `install_dependencies.sh` is located in the root directory.

How about only considering sh-files located in the project's root or `scripts` directory?

scripts/lint.sh
```diff
@@ -94,4 +94,16 @@ if [ ! -z "$output" ]; then
 	exitcode=1
 fi
 
+# ShellCheck linting
+for file in `find . -name "*.sh"`
```
mweisgut, line 98: For consistency reasons, how about using the `$(<bash command>)` syntax as used above, see line 68.

Bouncner: As far as I know, backticks are considered "old style" and `$(..)` should be preferred.

tjjordan: Thanks for pointing this out! :)

scripts/lint.sh
tjjordan: This is the better way, done!

Bouncner: When you're only accessing sh-files in the scripts directory, you can remove all the checks as Marcel hinted.

mweisgut: In addition, I would include `install_dependencies.sh` (or .sh files located in the root directory in general).

Bouncner:
What is the status here?

tjjordan:
It has not been forgotten.
The SpellCheck errors in the scripts need to be fixed be corrected otherwise the CI will fail.
I'm planning to do this after the PartialIndex-PR or when I've some spare time.

dey4ss:
Still did not get if it's S<ins>p</ins>ellCheck or S<ins>h</ins>ellcheck (and at that point, I'm too afraid to ask).

mweisgut:
Closing it for now. Feel free to reopen if it becomes a priority.

PR #2509, opened 2022-10-14
Disable encrypted connections in server tests

dey4ss:
On sidon (Ubuntu 20.04, psql 12.7) I got the following error when running the `hyriseServerTest.py` script/connecting to a `hyriseServer` with psql:
```
psql: error: could not connect to server: Connection refused
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 41963?
received invalid response to GSSAPI negotiation: R
```

When applying this suggestion [1], everything worked fine. Basically, we disable encrypted connections to the server for the test [2]. Since we do not perform user authentication/encryption by any means, this should be okay. Though I could not reproduce the error on macOS/psql 14.05, I adapted the test file. 

[1] https://github.com/psycopg/psycopg2/issues/1084#issuecomment-656778107
[2] https://www.postgresql.org/docs/13/libpq-connect.html#LIBPQ-CONNECT-GSSENCMODE

### DIFF
--- a/scripts/test/hyriseServer_test.py
+++ b/scripts/test/hyriseServer_test.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import os
 import pexpect
 import re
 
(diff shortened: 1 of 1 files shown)


scripts/test/hyriseServer_test.py
```diff
@@ -9,27 +10,31 @@
...
+    server.expect(search_regex, timeout=120)
+
+    server_port = int(re.search(search_regex, str(server.after)).group(1))
+    # Recent Postgres/psql versions changed the authentication behavior, resulting in connection errors on some setups.
+    # Disabling encrypted connections solves the issue. Since hyriseServer does not implement authentication at all,
+    # this is no problem.
+    # See https://github.com/psycopg/psycopg2/issues/1084#issuecomment-656778107 and
+    # https://www.postgresql.org/docs/13/libpq-connect.html#LIBPQ-CONNECT-GSSENCMODE
+    environment_variables = os.environ.copy()
+    environment_variables.update({"PGGSSENCMODE": "disable"})
```
dey4ss, line 27: Comment only: not using
```python
environment_variables = os.environ | {"PGGSSENCMODE": "disable"}
```
as it is Python 3.9+ syntax and we require Python 3.6+.

Bouncner: Just a side note: if you see any reason to update to 3.9, feel free. Python 3.6 is six years old and reached its end of life.

Bouncner (approved):
LGTM.

scripts/test/hyriseServer_test.py
dey4ss: Could also be
```python
environment_variables["PGGSSENCMODE"] = "disable"
```
so guess it's just a matter of taste. However, updating some of our (python) dependencies could be done, right. But if we do so, we should maybe also refactor our scripts with new python functionality and more pythonic approaches (code in main function rather than globally, argparse, etc) in general.

Bouncner: I doubt somebody finds the time for such "pythonications" if we cannot automate them.

dey4ss: Don't think so either. Thus, I guess updating Python is not necessary now.

PR #2511, opened 2022-10-21
Clang 15 support

tjjordan:
Removed unnecessary variable `output_chunk_id`, which was written to but never read from, causing a compiler warning.

Also, a DebugAssert in the function `string_heap_size()` from `size_estimation_utils.hpp` was failing, caused by the test case `SingleString` in `SizeEstimationUtilsTest`.
If have no clue why it was failing, but just changing the structure made it work again.
Is somebody aware of the the idea behind this DebugAssert?


src/lib/utils/size_estimation_utils.hpp
```diff
@@ -17,7 +17,8 @@ size_t
 string_heap_size(const T& string) {
   // Get the default pre-allocated capacity of SSO strings. Note that the empty string has an unspecified capacity, so
   // we use a really short one here.
-  const auto sso_string_capacity = T{"."}.capacity();
+  auto small_string = T{"."};
```
Bouncner, line 20: const

src/lib/utils/size_estimation_utils.hpp
```diff
@@ -17,7 +17,8 @@ size_t
 string_heap_size(const T& string) {
   // Get the default pre-allocated capacity of SSO strings. Note that the empty string has an unspecified capacity, so
   // we use a really short one here.
-  const auto sso_string_capacity = T{"."}.capacity();
```
Bouncner, line 20: Just to get you right. The previous code here `T{"."}.capacity();` compiles just fine but one assert does fail in that case?

tjjordan: Yes!
Do you have an idea how this could be?

dey4ss: Tried the following for `std::string`:
```c++
const auto sso_string_capacity = std::string{"."}.capacity();
const auto sso_string = std::string{"."};
const auto sso_string_capacity_2 = sso_string.capacity();
std::cout << sso_string_capacity << " " << sso_string_capacity_2 << std::endl;
```
and it outputs (as expected)
```
22 22
```
for both clang-14 and clang-15 on macOS. Seems to be a quite weird error.

tjjordan: How does the (unedited) test case behave on your machine?

dey4ss: Tried it and it's just getting weirder. I added the following print:
```c++
const auto sso_string_capacity = T{"."}.capacity();
const auto small_string = T{"."};
const auto sso_string_capacity_2 = small_string.capacity();
std::cout << string.capacity() << " " << sso_string_capacity << " " << sso_string_capacity_2 << " "
          << T{"."}.capacity() << " " << small_string.capacity() << std::endl;
```
and it outputs
```
22 23 22 22 22
```

I guess adding a comment here would be helpful. Furthermore, I wonder if we can capture the SSO capacity as a constant to avoid the evaluation for each execution of the utility function.

klauck: From my understanding `std::string.capacity()` should always return the same value for current implementations using sso for short strings: I found something like the following implementation for short strings: `return static_cast<size_type>(__min_cap)) - 1;`
My guess, the weird 23 comes from some compiler magic/optimization, because the capacity could be determined at compile time. I am not sure whether the current fix handles future compiler optimizations well (better than the previous version).

klauck: This is why, removing the assert and leaving a comment is maybe a better fix.
Just as note: besides evaluating the string's capacity to assume sso, one could also evaluate the address of data(). (see, for example, https://stackoverflow.com/questions/65736613/c-sso-how-to-programatically-find-if-a-stdwstring-is-allocated-with-short)

dey4ss:
Was there an issue when using the templated version of `string_heap_size()`? `pmr_string` is basically just a std::string with `PolymorphicAllocator` instead of `std::allocator` (and the SOO capacity _should_ be the same)

tjjordan:
I was thinking about different string types having a differnt SSO capacity.
But if it is the same the templatization should be fine.
How do you like the current solution?

dey4ss (approved):

PR #2512, opened 2022-10-21
Fix binary export of reference segments and enable coverage generation on macOS

dey4ss:
This PR fixes the binary export of reference segments. Though we materialize these segments, the current implementation (i) did not meet the output format of value segments and (ii) did not handle `NULL` values.

To ensure correct export results, the test for exporting reference segments compares the generated file with a binary dump of value segments.

Furthermore, it enables to run the coverage script on macOS.


src/lib/import_export/binary/binary_writer.cpp
```diff
@@ -6,15 +6,14 @@
 #include <string>
 #include <vector>
 
+#include "constant_mappings.hpp"
```
Bouncner, line 9: Needed?

Bouncner (approved):
Looks good (just a tiny question concerning an include).

PR #2513, opened 2022-10-27
Remove unused server test and extend existing one

Bouncner:
Short PR that removes a test file I have never seen, never used, and haven't found a usage in Hyrise.
It uses an unsupported `load` command (only supported in the console as far as i know). So it did not even work.

I added the loading (done via `copy from`) to the existing server test which is part of the CI.


Bouncner:
> Just wanted to point that we also test certain server capabilities, such as import/export, in [server_test_runner.cpp](../blob/master/src/test/lib/server/server_test_runner.cpp). Will accept once CI passes.

Yes, but for some reason it satisfies me to see that I can local files via Python.

dey4ss (approved):

PR #2514, opened 2022-11-08
Export meta tables

dey4ss:
Besides storing meta tables to files from the SQL interface, this PR removes the unused `table_name` attribute from `ExportNode` (table to be exported is passed as input node) and consistently replaces `EXPECT_EQ(..., true)`/`EXPECT_EQ(..., false)` with `EXPECT_TRUE(...)` or `EXPECT_FALSE(...)`.

The first functionality enables us to easily export, e.g., logs from benchmark runs if functionality uses logging:
```c++
SQLPipelineBuilder{"COPY meta_log TO 'output.csv'"}
  .create_pipeline()
  .get_result_table();
```

If we enhanced the parser/SQL translator functionality further to allow for selections on exported tables (maybe even encoding for binary exports), we could also do the following in the future (also provided by Postgres [1]):
```SQL
COPY (SELECT *
        FROM meta_log
       WHERE log_level IN ('Info', 'Warning'))
  TO 'output_file'
WITH FORMAT BINARY,
     ENCODING 'LZ4'
```
 
A step further, we can also add encoding and filtering to the import (matches Postgres [1]):
```SQL
 COPY a_table
 FROM 'table_file'
 WITH FORMAT CSV,
      ENCODING 'Dictionary'
WHERE foo = "bar"
```

[1] https://www.postgresql.org/docs/15/sql-copy.html


src/lib/logical_query_plan/export_node.cpp
```diff
@@ -6,34 +6,26 @@
...
-      file_name(init_file_name),
-      file_type(init_file_type) {}
+ExportNode::ExportNode(const std::string& init_file_name, const FileType init_file_type)
+    : AbstractNonQueryNode(LQPNodeType::Export), file_name(init_file_name), file_type(init_file_type) {}
 
 std::string ExportNode::description(const DescriptionMode mode) const {
-  std::ostringstream stream;
-  stream << "[Export] Name: '" << table_name << "'";
-  return stream.str();
+  return "[Export]";
```
dey4ss, line 13: Comment only: I could add the export path here if desired an return sth like:
`[Export] to 'given/export/path.csv'`.
```suggestion
  return "[Export] to '" + file_name + '";
```
The same could be done for the operator description.

Bouncner: Sounds good to me.

dey4ss: Added the info. For the export operator, the description also prints the file type (binary/csv; `FileType::Auto` is only translated to an actual type in the operator).

scripts/test/hyriseServer_test.py
```diff
@@ -42,7 +42,7 @@ def main():
 
     client.sendline("COPY loaded_table_from_tbl FROM 'resources/test_data/tbl/int.tbl';")
     client.expect_exact("SELECT 0")
-    client.sendline("SELECT COUNT(*) AS \"row_count\" FROM loaded_table_from_tbl;")
+    client.sendline('SELECT COUNT(*) AS "row_count" FROM loaded_table_from_tbl;')
```
tjjordan, line 45: It could be an idea to change the other statements to single-quotes, too.
But I think it's a matter of taste.

src/lib/logical_query_plan/export_node.hpp
```diff
@@ -14,11 +14,10 @@ namespace hyrise {
...
 class ExportNode : public EnableMakeForLQPNode<ExportNode>, public AbstractNonQueryNode {
  public:
-  ExportNode(const std::string& init_table_name, const std::string& init_file_name, const FileType init_file_type);
+  ExportNode(const std::string& init_file_name, const FileType init_file_type);
 
   std::string description(const DescriptionMode mode = DescriptionMode::Short) const override;
 
-  const std::string table_name;
   const std::string file_name;
   const FileType file_type;
```
tjjordan, line 22: Not in scope of this PR but couldn't they be protected?

src/lib/operators/export.cpp
```diff
@@ -22,6 +23,18 @@ const std::string& Export::name() const {
   return name;
 }
 
+std::string Export::description(DescriptionMode description_mode) const {
+  const auto separator = (description_mode == DescriptionMode::SingleLine ? ' ' : '\n');
+
+  auto stream = std::stringstream{};
```
tjjordan, line 29: What was the reason for using a stringstream instead of string manipulation?

src/test/lib/operators/export_test.cpp
```diff
@@ -99,6 +93,16 @@ TEST_P(OperatorsExportMultiFileTypeTest, ExportWithoutFileType) {
   EXPECT_TRUE(compare_files(reference_filename, filename));
 }
 
+TEST_P(OperatorsExportMultiFileTypeTest, NameAndDescription) {
+  const auto exporter = std::make_shared<Export>(nullptr, "a/file/path", GetParam());
+  const auto file_type_string = GetParam() == FileType::Binary ? "binary" : "csv";
```
tjjordan, line 98: Should the type of a variable be in the variable`s name?

tjjordan (approved):
Looks good and clean to me.
Not really something to add, just some questions out of curiosity.

scripts/test/hyriseServer_test.py
tjjordan: Nevermind, I just saw the format script is doing this.

dey4ss: Black prefers double-quote strings (and changes strings to use them). However, black 22.10 on my machine changed it to this version.

src/lib/operators/export.cpp
dey4ss: No specific reason, basically just copied it from elsewhere 😅

src/test/lib/operators/export_test.cpp
dey4ss: Also do not really like it. However, we have a `FileType` enum and this is a string represenation of the enum value (which we have to create for getting a lowercase string)... Do you have any better suggestion?

src/lib/logical_query_plan/export_node.hpp
dey4ss: We often do not protect members of LQP nodes when they do not change and are accessed for creating the corresponding operator. IMO I think it's okay to have these const public members rather than private/protected member + public accessor for these intermediate representations

src/lib/utils/meta_tables/meta_log_table.cpp
```diff
@@ -28,14 +28,13 @@ std::shared_ptr<Table> MetaLogTable::_on_generate() const {
     // We need this to format the timestamp in a thread-safe way.
     // https://stackoverflow.com/questions/25618702/
     //   why-is-there-no-c11-threadsafe-alternative-to-stdlocaltime-and-stdgmtime
-    std::ostringstream timestamp;
-    auto time = std::chrono::system_clock::to_time_t(entry.timestamp);
+    auto timestamp_stream = std::ostringstream{};
+    auto timestamp = std::chrono::system_clock::to_time_t(entry.timestamp);
+    auto buffer = tm{};
```
Bouncner, line 33: What's tm?

Bouncner (approved):
Short question, nothing else.

src/lib/utils/meta_tables/meta_log_table.cpp
dey4ss: > Structure holding a calendar date and time broken down into its components.

https://en.cppreference.com/w/c/chrono/tm

Bouncner: IFF you're going to touch this PR, a short comment would be nice.

Besides that, Knüller zünden?

dey4ss: Can touch this when allowing queries on exported tables / add a switch to export logs on benchmarks. Ready for Zündung if this is fine for you.

Bouncner: Fine with me. Zünd it!

PR #2515, opened 2022-11-08
Implemented in memory benchmark and imrpoved read benchmarks

SanJSp:
This pr
- introduces in memory read and write benchmarks
- adds sanity checks to read and pread benchmark

The sanity check is currently to simply add up all read numbers and compare them to a sum that was calculated beforehand.
The in memory benchmarks simply read and write contents from a vector. The are quite similar as reading from a vector mainly also means to write these values somewhere.


SanJSp:
Accidentaly created it here, should be in our repo :D sorry

PR #2448, opened 2022-04-08
Integrate Partial Indexes

mweisgut:
This PR implements a partial hash index (`PartialHashIndex`), i.e., a single-column index structure that can be created for one or more segments of the corresponding column without the need to index all segments. This PR is based on the work started by @bengelhaupt, @Jasperhino, and @vxrahn in #2386.

This `PartialHashIndex` index differs from the already implemented indexes because it is created over multiple chunks of a table instead of being on a per-chunk basis. It can be constructed for a set of chunks and can later be modified by adding additional chunks or removing already indexed chunks.
The already existing index implementations also share the property of having the methods `lower_bound` and `upper_bound` used for range queries. The `PartialHashIndex` is based on an unsorted hash map and, therefore, cannot provide such functionality efficiently. Instead, it provides the functions `equals` and `not_equals`, which allow lookups in constant time.
This is achieved by using a hash map that maps a given value to a vector of positions (i.e., `RowID`s) where this value is located. The `FlatMapIterator` allows iterating over these lists of `RowID`s.

We made the following design decisions:
- The `PartialHashIndex` does not inherit from the `AbstractIndex` or does share a common superclass with, e.g., the `GroupKeyIndex`, because of an iterator inheritance problem. All indexes that currently inherit from `AbstractIndex` coincidentally share its iterator type `std::vector<ChunkOffset>::const_iterator`, making inheritance work. The iterator of the `PartialHashIndex` does not match that, nor is it covariant. To resolve this, the iterator type of the `AbstractIndex` would need to be made polymorph, further requiring that pointer or references of it need to be passed around. Since no other partial index or table index is implemented that shares the same interface, the `PartialHashIndex` is not derived from a certain abstract index class.
- The templated class `PartialHashIndexImpl` has been introduced to allow for typesafe usage of the map keys (which are of the segment/column datatype). Such an object is held by the actual `PartialHashIndex` class, which forwards respective calls (using the pImpl idiom). 
- `PartialHashIndex` uses the pointer to implementation (pImpl) idiom to allow for typesafe usage of the map keys (which are of the segment/column datatype). I.e., the non-templated `PartialHashIndex` instance
holds an opaque pointer to its templated implementation `PartialHashIndexImpl`. The template parameter is supposed to
be the data type of the indexed column. Since the positions (i.e., RowIDs) stored in the index are accessed via
iterators (more precisely, via its access methods that also work with iterators), a non-templated iterator is
also needed.
[...]

PR #2522, opened 2022-11-29
Add Pre-/Post-Benchmark Hooks for Plug-ins

dey4ss:
First, this PR adds optional hooks to the plug-in interface that are called before and after benchmark executions. For instance, these hooks can be used for applying physical designs (tiering, indexes, compression, data dependencies) before the items are executed or for exporting workload information after benchmark runs.

The `BenchmarkRunner`'s benchmark item runner is passed as an argument to the pre-benchmark hooks. Thus, plug-ins can execute the items, run advisors, etc. The post-benchmark hook gets the JSON report as argument to add or modify metrics.

Second, this PR gets rid of `constant_mappings` and the extern bimaps to map enum types to strings. We use `magic_enum::enum_name()` wherever possible. If string representations do not match the enum names, this PR moves the bimaps to the corresponding header (but *not* as `extern`). The reason is that we link other executables againt the test target and `extern` definitions of said bimaps caused segfaults in the CI coverage build (presumably multiple deallocations).


src/benchmarklib/benchmark_runner.cpp
```diff
@@ -7,19 +7,18 @@
 #include <boost/accumulators/accumulators.hpp>
 #include <boost/accumulators/statistics/mean.hpp>
 #include <boost/accumulators/statistics/stats.hpp>
-
 #include <boost/algorithm/string/join.hpp>
 #include <boost/range/adaptors.hpp>
-#include "cxxopts.hpp"
+#include <cxxopts.hpp>
```
Bouncner, line 12: Actually, I think boost's usage of `<>` is rather wrong. For `cxxopts` it is most certainly wrong as it is not a system library.

We might be better of pulling boost as a third party library (different boost version on Ubuntu/MacOS brew can be a pain). Then the entire discussio would be rather simple.

src/benchmarklib/encoding_config.cpp
```diff
@@ -1,6 +1,6 @@
 #include "encoding_config.hpp"
 
-#include "constant_mappings.hpp"
+#include <magic_enum.hpp>
```
Bouncner, line 3: See above.

src/lib/storage/vector_compression/compressed_vector_type.cpp
```diff
@@ -0,0 +1,29 @@
+#include "compressed_vector_type.hpp"
+
+namespace hyrise {
+
+std::ostream& operator<<(std::ostream& stream, const CompressedVectorType compressed_vector_type) {
+  switch (compressed_vector_type) {
+    case CompressedVectorType::FixedWidthInteger4Byte: {
```
Bouncner, line 7: Can't we use `magic_enum` here as well?

src/lib/storage/vector_compression/compressed_vector_type.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstdint>
+#include <optional>
 
 #include <boost/version.hpp>
 #if BOOST_VERSION < 107100                 // TODO(anyone): remove this block once Ubuntu ships boost 1.71
```
Bouncner, line 7: Just checked: Ubuntu 20 ships boost 1.71 (https://launchpad.net/ubuntu/focal/+package/libboost-dev). Do we want to support older Ubuntu versions? I'd say no.

src/lib/utils/meta_tables/meta_log_table.cpp
```diff
@@ -1,6 +1,7 @@
 #include "meta_log_table.hpp"
 
-#include "constant_mappings.hpp"
+#include <magic_enum.hpp>
```
Bouncner, line 3: See above.

src/lib/utils/meta_tables/segment_meta_data.cpp
```diff
@@ -1,5 +1,7 @@
 #include "segment_meta_data.hpp"
 
+#include <magic_enum.hpp>
```
Bouncner, line 3: See above.

src/lib/utils/plugin_manager.cpp
```diff
@@ -74,16 +74,25 @@ void PluginManager::load_plugin(const std::filesystem::path& path) {
...
+  const auto& post_benchmark_hook = plugin_ref.post_benchmark_hook();
+  if (post_benchmark_hook) {
+    _post_benchmark_hooks[plugin_name] = *post_benchmark_hook;
+  }
 }
 
 void PluginManager::exec_user_function(const PluginName& plugin_name, const PluginFunctionName& function_name) {
   Assert(_user_executable_functions.contains({plugin_name, function_name}),
-         "There is no " + function_name + " defined for plugin " + plugin_name + ".");
+         "There is no " + function_name + " defined for plugin `" + plugin_name + ".");
```
Bouncner, line 95: Either remove or change to straight quotes with an end quote as well.

src/lib/utils/plugin_manager.cpp
```diff
@@ -95,6 +104,39 @@ void PluginManager::exec_user_function(const PluginName& plugin_name, const Plug
       LogLevel::Info);
 }
 
+void PluginManager::exec_pre_benchmark_hook(const PluginName& plugin_name,
+                                            AbstractBenchmarkItemRunner& benchmark_item_runner) {
+  Assert(_pre_benchmark_hooks.contains({plugin_name}),
+         "There is no pre-benchmark hook defined for plugin `" + plugin_name + "`.");
```
Bouncner, line 110: Wrong quotes. Multiple times.

src/lib/utils/plugin_manager.cpp
```diff
@@ -95,6 +104,39 @@ void PluginManager::exec_user_function(const PluginName& plugin_name, const Plug
...
+  auto& log_manager = Hyrise::get().log_manager;
+  log_manager.add_message("PluginManager", "Called pre-benchmark hook provided by plugin `" + plugin_name + "`.",
+                          LogLevel::Info);
+}
+
+void PluginManager::exec_post_benchmark_hook(const PluginName& plugin_name) {
+  Assert(_post_benchmark_hooks.contains({plugin_name}),
+         "There is no post-benchmark hook defined for plugin `" + plugin_name + "`.");
+
+  const auto post_benchmark_hook = _post_benchmark_hooks[plugin_name];
```
Bouncner, line 124: ref?

src/lib/utils/plugin_manager.cpp
```diff
@@ -95,6 +104,39 @@ void PluginManager::exec_user_function(const PluginName& plugin_name, const Plug
       LogLevel::Info);
 }
 
+void PluginManager::exec_pre_benchmark_hook(const PluginName& plugin_name,
+                                            AbstractBenchmarkItemRunner& benchmark_item_runner) {
+  Assert(_pre_benchmark_hooks.contains({plugin_name}),
+         "There is no pre-benchmark hook defined for plugin `" + plugin_name + "`.");
+
+  const auto pre_benchmark_hook = _pre_benchmark_hooks[plugin_name];
```
Bouncner, line 112: ref?

src/test/CMakeLists.txt
```diff
@@ -268,6 +268,7 @@ set(
     # Added plugin targets so that we can test member methods without going through dlsym
     hyriseMvccDeletePlugin
     hyriseUccDiscoveryPlugin
+    # hyriseBenchmarkLib
```
Bouncner, line 271: ?

src/test/lib/operators/sort_test.cpp
```diff
@@ -116,7 +116,7 @@ inline std::string sort_test_formatter(const testing::TestParamInfo<SortTestPara
   }
   stream << (param.input_is_reference ? "Reference" : "Data") << "Input";
   for (const auto& sort_column : param.sort_columns) {
-    stream << "Col" << sort_column.column << sort_mode_to_string.left.at(sort_column.sort_mode);
```
Bouncner, line 119: Not yours, but that looks ugly. How does not look when printed? `coll_orderkeyAscending`?

src/test/lib/utils/plugin_manager_test.cpp
```diff
@@ -1,5 +1,6 @@
 #include "base_test.hpp"
 
+// #include "../benchmarklib/tpch/tpch_benchmark_item_runner.hpp"
```
Bouncner, line 3: ?

Bouncner (commented):
Nothing serious.

Nice clean ups!

src/benchmarklib/benchmark_runner.cpp
dey4ss: Just grepped a few variations, seems we usually import magic_enum and boost with `<>` and only in a few cases with `""`. Totally fine to change this (altogether with the Boost <1.71 workaround). However, I'd move this to a new issue/PR if that's okay for you.

src/lib/storage/vector_compression/compressed_vector_type.cpp
dey4ss: Sure we can, just did not notice it.

src/lib/storage/vector_compression/compressed_vector_type.hpp
dey4ss: see https://github.com/hyrise/hyrise/pull/2522#discussion_r1091640241

src/lib/utils/plugin_manager.cpp
dey4ss: Argh, copied it from the here (and obviously not in the correct way.
https://github.com/hyrise/hyrise/blob/54770d02e097b783bde1c1924f45cca43502a4f3/src/lib/utils/plugin_manager.cpp#L94
I'd go for `'` if you agree - just like emphasizing the user-provided names in the messages.

src/test/CMakeLists.txt
dey4ss: The outcommented cmake-stuff and the outcommented includes are an artifact from playing around to fix the CI stage, gonna remove.

dey4ss: Just uncommented everything again (required for testing the benchmark hooks). Let's see if it also passes CI. 🤞

src/benchmarklib/benchmark_runner.cpp
dey4ss: Taken from the [Boost wiki](https://github.com/boostorg/wiki/wiki/Getting-Started%3A-Overview#installing-boost):
> The initial cloning will take at least 45 minutes on a 3.0 mbps Internet connection and will consume roughly 1.5 GB of disk space. Cloning is only done once, so this lengthy initial time is not a long-term concern.

Are we sure we want this to be a third party library (CI builds, etc.)?

dey4ss:
@Bouncner @mweisgut Still open to your opinion on passing `AbstractBenchmarkItemuRunner` to the `pre_benchmark_hook` (to, e.g., execute the workload in advance for workload-driven optimization plugins) and passing nothing to the `post_benchmark_hook`. Guess globally reachable stuff (like exporting the log) can be done with that.

src/benchmarklib/benchmark_runner.cpp
Bouncner: Well, that might be a good reason to keep it as a sys library. :D

src/test/lib/operators/sort_test.cpp
Bouncner: Hmm, no. But then again: it's just autogenerated test names. Feel free to resolve.
Sorry.

Bouncner:
> @Bouncner @mweisgut Still open to your opinion on passing `AbstractBenchmarkItemuRunner` to the `pre_benchmark_hook` (to, e.g., execute the workload in advance for workload-driven optimization plugins) and passing nothing to the `post_benchmark_hook`. Guess globally reachable stuff (like exporting the log) can be done with that.

I think I need to discuss that in person. Not sure I can follow.

dey4ss:
> I think I need to discuss that in person. Not sure I can follow.

Let's discuss in the fortnightly

dey4ss:
As discussed in the fortnightly, pre-benchmark hooks now have access to the benchmark report JSON. Thus, I had to split the `BenchmarkRunner::write_report_to_file()` into report creation and actually writing to always run the benchmark hooks (they might not touch the JSON at all). However, we do not actually create the JSON if we do not write it and it is the plugin's reponsibility to cope with that or to crash.

src/benchmarklib/benchmark_runner.cpp
dey4ss: At least the <1.71 workaround is now fixed 🤷‍♀️

Bouncner (approved):
Just a potential miss of a const.

src/benchmarklib/benchmark_runner.cpp
```diff
@@ -151,26 +152,29 @@ void BenchmarkRunner::run() {
     }
   }
 
+  // Create report.
+  const auto write_report = _config.output_file_path && (!_config.verify && !_config.enable_visualization);
+  auto report = write_report ? _create_report() : nlohmann::json{};
```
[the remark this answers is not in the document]
dey4ss: Cannot be `const` when the plugin shall be allowed to modify / add metrics to the report, right? We *could* return a ptr to the report which then can be `const`.

src/plugins/CMakeLists.txt
```diff
@@ -31,11 +31,11 @@ function(add_plugin)
...
 endfunction(add_plugin)
 
-add_plugin(NAME hyriseMvccDeletePlugin SRCS mvcc_delete_plugin.cpp mvcc_delete_plugin.hpp DEPS sqlparser magic_enum gtest)
-add_plugin(NAME hyriseSecondTestPlugin SRCS second_test_plugin.cpp second_test_plugin.hpp DEPS sqlparser)
-add_plugin(NAME hyriseTestPlugin SRCS test_plugin.cpp test_plugin.hpp DEPS sqlparser)
-add_plugin(NAME hyriseTestNonInstantiablePlugin SRCS non_instantiable_plugin.cpp)
-add_plugin(NAME hyriseUccDiscoveryPlugin SRCS ucc_discovery_plugin.cpp ucc_discovery_plugin.hpp DEPS sqlparser magic_enum compact_vector)
+add_plugin(NAME hyriseMvccDeletePlugin SRCS mvcc_delete_plugin.cpp mvcc_delete_plugin.hpp DEPS sqlparser magic_enum gtest hyriseBenchmarkLib)
+add_plugin(NAME hyriseSecondTestPlugin SRCS second_test_plugin.cpp second_test_plugin.hpp DEPS sqlparser hyriseBenchmarkLib)
+add_plugin(NAME hyriseTestPlugin SRCS test_plugin.cpp test_plugin.hpp DEPS sqlparser hyriseBenchmarkLib hyriseBenchmarkLib)
```
mweisgut, line 36: Redundant `hyriseBenchmarkLib`. Will be fixed in #2448.

PR #2524, opened 2022-11-30
Update MacOS to LLVM 15

Bouncner:
This PR updates the Jenkinsfile to use LLVM 15 on both CI macs.

Besides that, some small changes:
 - clang tidy does no longer check `bugprone-suspicious-include`, which blocks Unity builds (see below)
 - LLVM 15 complained about a set-but-unsed variable. The CSV parser only checked the variable in debug mode. This check is now no longer a `DebugAssert`, but an `Assert`. I don't see a good reason to complain about malformed CSVs depending on whether it's a debug or release build. Should be the same in both cases and being strict here helps to find CSV problems.
 - some code style changes in csv_parser.cpp

Btw: both Macs are now running Ventura.


Bouncner:
@dey4ss & @mweisgut : what is your opinion on the clang-tidy change? From my point of view, I don't think we need to flag `#include *.cpp` as I am pretty sure we would catch that anyways. So this check is not that important for us.

On the other hand, removing this check allows to run unity builds for clang-tidy which brings clang-tidy from 1h15-1h45 down to ~45 minutes. As the other checks mostly take less than 50 mins, we sped up the full CI run quite a bit.

dey4ss:
cpplint/`lint.sh` complains anyways: `Do not include .cpp files from other packages  [build/include] [4]`. So fine from my side.

dey4ss (dismissed):

.clang-tidy
```diff
@@ -103,6 +104,8 @@ WarningsAsErrors: '*'
 #   - bugprone-easily-swappable-parameters                      Ignore issues with swapple parameters as we found them to be of no help.
 #   - readability-function-cognitive-complexity                 When appropriate, long functions with a sequential data flow (which is sometimes easier to read) are fine. In many
 #                                                                places, the rule catches functions where the code could be improved, but will likely not be due to a lack of time.
+#   -bugprone-suspicious-include                                Unity builds use #include's for cpp files. However, unity build significantly improve the runtime of clang-tidy,
+#                                                                which is otherwise the slowest step of the CI pipeline by a factor of two.
```
dey4ss, line 108:
```suggestion
#   -bugprone-suspicious-include                                Unity builds use `#include`s for cpp files. However, unity builds significantly improve the runtime of clang-tidy,
#                                                                which is otherwise the slowest step of the CI pipeline by a factor of two. Furthermore, the linter checks for 
#                                                                these includes as well.
```

mweisgut:
> cpplint/`lint.sh` complains anyways: `Do not include .cpp files from other packages [build/include] [4]`. So fine from my side. We might add this info to the comment.

Agreed. Especially with the linter catching that case.

mweisgut (changes requested):
It seems the PR is in an experimental state. Feel free to ping me again when the Jenkins file is ready to be reviewed.

Bouncner:
Pinging @mweisgut and @dey4ss .

Jenkinsfile
```diff
@@ -207,7 +207,7 @@ try {
               if (env.BRANCH_NAME == 'master' || full_ci) {
                 // We do not run tidy checks on the src/test folder, so there is no point in running the expensive clang-tidy for those files
                 // As clang-tidy is the slowest step, we allow it to use more parallel jobs.
-                sh "cd clang-debug-tidy && make hyrise_impl hyriseBenchmarkFileBased hyriseBenchmarkTPCH hyriseBenchmarkTPCDS hyriseBenchmarkJoinOrder hyriseConsole hyriseServer -k -j \$(( \$(nproc) / 5))"
+                sh "cd clang-debug-tidy && make hyrise_impl hyriseBenchmarkFileBased hyriseBenchmarkTPCH hyriseBenchmarkTPCDS hyriseBenchmarkJoinOrder hyriseConsole hyriseServer -k -j \$(( \$(nproc) / 10))"
```
mweisgut, line 210: Why?

Bouncner: clang-tidy was the exception before with `ncores/5` cores instead of `ncores/10` which we used for all the other steps (we have 11 steps of which 10 take quite long, thus the `/10`).

Now with unity builds, clang-tidy is no longer the slowest step. Thus, it is no longer "preferred".

PR #2528, opened 2022-12-20
Refactor UCCs and FDs

dey4ss:
This PR is the first one of a few to introduce new data dependencies in Hyrise. It prepares the interfaces for these new dependencies.

Notable changes in this PR are:
- `LQPUniqueConstraint` is renamed to `UniqueColumnCombination`.
- `AbstractTableConstraint` does not store a set, but a vector of columns since order and inclusion dependencies have a list semantic for their columns (i.e., column order matters). To ensure different orders are not a problem for `TableKeyConstraint`, we sort and de-duplicate the columns in the constructor.
- LQP nodes collect their dependencies in sets rather than in lists. This is more intuitive and ensures that dependencies cannot be duplicated for the same node. Unique column combinations are passed as `std::unordered_set<UniqueColumnCombination>`, not as `std::unordered_set<UniqueColumnCombination>`.

Most of the actual changes are in `logical_query_plan/data_dependencies` and `storage/constraints`. The rest of the lib changes are mainly the adoption of the interface changes. Same goes for the tests. Furthermore, it gets rid of many (duplicated) formatter definitions in tests, which are now mainly part of `base_test.hpp`.


mweisgut (commented):
Partially reviewed, I'll continue with `src/lib/logical_query_plan/dummy_table_node.cpp`. 

The codebase still has multiple occurrences of `unique constraints`. I assume some of them should be changed to `unique column combination` or `UCC` for consistency.

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -94,73 +94,74 @@ bool AggregateNode::is_column_nullable(const ColumnID column_id) const {
...
   if (group_by_columns_count > 0) {
-    ExpressionUnorderedSet group_by_columns(group_by_columns_count);
+    auto group_by_columns = ExpressionUnorderedSet{group_by_columns_count};
     std::copy_n(node_expressions.begin(), group_by_columns_count,
                 std::inserter(group_by_columns, group_by_columns.begin()));
 
-    // Make sure, we do not add an already existing or a superset unique constraint.
-    if (unique_constraints->empty() || !contains_matching_unique_constraint(unique_constraints, group_by_columns)) {
-      unique_constraints->emplace_back(group_by_columns);
+    // Make sure that we do not add an already existing or a superset unique constraint.
```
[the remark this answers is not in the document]
dey4ss: Went through all matches of ` constraint` in the code base and think I renamed them properly.

src/lib/logical_query_plan/data_dependencies/functional_dependency.hpp
```diff
@@ -5,27 +5,27 @@
...
- * {Left} => {Right}
+ * Models a functional dependency (FD), which consists out of two sets of expressions. The left set of expressions
+ * (determinants) unambigiously identifies the right set (dependents):
+ *   {Left} => {Right}
  *
  * Example A:
- * Think of a table with three columns: (Semester, CourseID, Lecturer)
- * The primary key is defined across the first two columns, which leads to the following FD:
- * {Semester, CourseID} => {Lecturer}
+ * Think of a table with three columns: Course(Semester, CourseID, Lecturer). The primary key is defined across the
```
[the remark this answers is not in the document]
dey4ss: `CourseAssignment`?

src/lib/logical_query_plan/mock_node.cpp
```diff
@@ -121,6 +97,32 @@ std::string MockNode::description(const DescriptionMode mode) const {
...
 
+std::shared_ptr<UniqueColumnCombinations> MockNode::unique_column_combinations() const {
+  auto unique_column_combinations = std::make_shared<UniqueColumnCombinations>();
+
+  for (const auto& table_key_constraint : _table_key_constraints) {
+    // Discard key constraints that involve pruned column id(s).
+    const auto& key_constraint_column_ids = table_key_constraint.columns();
+    if (std::any_of(_pruned_column_ids.cbegin(), _pruned_column_ids.cend(),
+                    [&key_constraint_column_ids](const auto& pruned_column_id) {
+                      return std::any_of(key_constraint_column_ids.cbegin(), key_constraint_column_ids.cend(),
```
mweisgut, line 108: Same as above: `find` or `binary_search`?

src/lib/logical_query_plan/static_table_node.hpp
```diff
@@ -21,8 +21,9 @@ class StaticTableNode : public EnableMakeForLQPNode<StaticTableNode>, public Abs
   std::vector<std::shared_ptr<AbstractExpression>> output_expressions() const override;
   bool is_column_nullable(const ColumnID column_id) const override;
 
-  // Generates unique constraints from table's key constraints.
-  std::shared_ptr<LQPUniqueConstraints> unique_constraints() const override;
+  // Generates UCCs from table's key constraints. We use these UCCs for storing table key constraints with CREATE TABLE
+  // statements.
```
mweisgut, line 25: The new comment is confusing to me: These UCCs, generated by this function, are used for storing table key constraints with CREATE TABLE statement? Aren't they stored as soft key constraints?

```suggestion
  // Generates UCCs from table's soft key constraints. We use these soft key constraints for storing table key constraints with CREATE TABLE
  // statements.
```

mweisgut (commented):
Partially reviewed. I'll continue with `src/lib/logical_query_plan/stored_table_node.cpp`.

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -106,89 +113,92 @@ std::shared_ptr<LQPUniqueConstraints> JoinNode::_output_unique_constraints(
...
-    auto unique_constraints =
-        std::make_shared<LQPUniqueConstraints>(left_unique_constraints->begin(), left_unique_constraints->end());
-    std::copy(right_unique_constraints->begin(), right_unique_constraints->end(),
-              std::back_inserter(*unique_constraints));
-    return unique_constraints;
+    // Due to the one-to-one relationship, the UCCs of both sides remain valid.
+    auto unique_column_combinations = std::make_shared<UniqueColumnCombinations>(
+        left_unique_column_combinations->begin(), left_unique_column_combinations->end());
+    std::copy(right_unique_column_combinations->begin(), right_unique_column_combinations->end(),
+              std::inserter(*unique_column_combinations, std::next(unique_column_combinations->begin())));
```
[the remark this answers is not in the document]
dey4ss: UCCs are stored in sets now. Thus, `inserter` rather than `back_inserter`. The iterator position was indeed a copy mistake, changed to `unique_column_combinations->end()`. However, this position is irrelevant for sets.

src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -435,7 +436,13 @@ ExpressionUnorderedSet find_column_expressions(const AbstractLQPNode& lqp_node,
 
   for (const auto& output_expression : output_expressions) {
     const auto column_expression = dynamic_pointer_cast<LQPColumnExpression>(output_expression);
-    if (column_expression && column_ids.contains(column_expression->original_column_id) &&
+    if (!column_expression) {
+      continue;
+    }
+
+    const auto original_column_id = column_expression->original_column_id;
+    if (std::any_of(column_ids.cbegin(), column_ids.cend(),
```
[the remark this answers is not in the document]
dey4ss: You're correct, they should be ordered (the function is only called by `MockNode`, `StaticTableNode`, and `StoredTableNode`. So `binary_search` + `DebugAssert` for sortedness (`Assert` would not make sense with binary search)?

src/lib/logical_query_plan/mock_node.cpp
dey4ss: both possible, and I do't think there will be a performance impact on binary search vs find (also, `MockNode` should not be in queries at all).
How about the following (more readable than `find/binary_search(...) != whatever.end()` IMHO)?
```c++
const auto pruned_column_ids = std::unordered_set<ColumnID>{_pruned_column_ids.cbegin(), 
                                                            _pruned_column_ids.cend()};

for (const auto& table_key_constraint : _table_key_constraints) {
  // Discard key constraints that involve pruned column id(s).
  const auto& key_constraint_column_ids = table_key_constraint.columns();
  if (std::any_of(key_constraint_column_ids.cbegin(), key_constraint_column_ids.cend(),
      [&](const auto column_id){ return pruned_column_ids.contains(column_id); })) {
    continue;
  }
  ...
}
```

src/lib/logical_query_plan/static_table_node.hpp
dey4ss: You're right

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -487,121 +487,197 @@ TEST_F(JoinNodeTest, FunctionalDependenciesUnify) {
...
   _mock_node_b->set_key_constraints({*_key_constraint_x});
 
-  for (const auto join_mode : {JoinMode::Semi, JoinMode::AntiNullAsTrue, JoinMode::AntiNullAsFalse}) {
+  const auto& join_unique_column_combinations = join_node->unique_column_combinations();
+  if (is_semi_or_anti_join(join_mode)) {
+    // Semi- and Anti-Joins act as filters for the left input and preserve its unique column combinations.
+    EXPECT_EQ(join_unique_column_combinations->size(), 1);
+    EXPECT_TRUE(find_ucc_by_key_constraint(*_key_constraint_b_c, join_unique_column_combinations));
+  } else {
+    // Not unique join columns might lead to row-/value-duplication in the opposite relation. No UCCs remain valid.
```
mweisgut, line 521:
```suggestion
    // Not unique join columns might lead to row/value duplication in the opposite relation. No UCCs remain valid.
```

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -487,121 +487,197 @@ TEST_F(JoinNodeTest, FunctionalDependenciesUnify) {
...
+    // clang-format off
+    join_node =
+    JoinNode::make(join_mode, equals_(_t_a_a, _t_b_y),
+      _mock_node_a,
+      _mock_node_b);
+    // clang-format on
+  }
+
+  // Case 1 – LEFT  table's join column (a) uniqueness : No
+  //          RIGHT table's join column (y) uniqueness : No
```
mweisgut, line 511: Since the different cases are not in a single test case anymore, adding some comments above the test cases to make it clear that they relate to each other might be helpful. This could be, for example, a combination table for the four different cases with a reference to the test cases.

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -487,121 +487,197 @@ TEST_F(JoinNodeTest, FunctionalDependenciesUnify) {
...
+    case JoinMode::AntiNullAsFalse:
+      // Semi- and Anti-Joins act as filters for the left input and preserve its unique column combinations.
+      EXPECT_EQ(join_unique_column_combinations->size(), 2);
+      EXPECT_EQ(*join_unique_column_combinations, *_mock_node_a->unique_column_combinations());
+      break;
+    case JoinMode::Inner:
+    case JoinMode::Left:
+    case JoinMode::Right:
+    case JoinMode::FullOuter:
+      // UCCs of the right input should be forwarded. Since our current UCCimplementation is compatible with NULL
```
mweisgut, line 562:
```suggestion
      // UCCs of the right input should be forwarded. Since our current UCC implementation is compatible with NULL
```

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -487,121 +487,197 @@ TEST_F(JoinNodeTest, FunctionalDependenciesUnify) {
...
+  //          RIGHT table's join column (y) uniqueness : Yes
+  _mock_node_a->set_key_constraints({*_key_constraint_a, *_key_constraint_b_c});
+  _mock_node_b->set_key_constraints({*_key_constraint_x, *_key_constraint_y});
+
+  const auto& join_unique_column_combinations = join_node->unique_column_combinations();
+  if (join_mode == JoinMode::Cross) {
+    EXPECT_TRUE(join_unique_column_combinations->empty());
+    return;
+
+    // For all other joins, UCCs of the left input should be forwarded.
```
mweisgut, line 624:
```suggestion
    // For all other join modes, UCCs of the left input should be forwarded.
```

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -487,121 +487,197 @@ TEST_F(JoinNodeTest, FunctionalDependenciesUnify) {
...
+  // Case 4 – LEFT  table's join column (a) uniqueness : Yes
+  //          RIGHT table's join column (y) uniqueness : Yes
+  _mock_node_a->set_key_constraints({*_key_constraint_a, *_key_constraint_b_c});
+  _mock_node_b->set_key_constraints({*_key_constraint_x, *_key_constraint_y});
+
+  const auto& join_unique_column_combinations = join_node->unique_column_combinations();
+  if (join_mode == JoinMode::Cross) {
+    EXPECT_TRUE(join_unique_column_combinations->empty());
+    return;
+
```
mweisgut, line 623: `}` missing

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -487,121 +487,197 @@ TEST_F(JoinNodeTest, FunctionalDependenciesUnify) {
...
-    // In-depth checks
-    EXPECT_TRUE(find_unique_constraint_by_key_constraint(*_key_constraint_a, join_unique_constraints));
-    EXPECT_TRUE(find_unique_constraint_by_key_constraint(*_key_constraint_b_c, join_unique_constraints));
-    EXPECT_TRUE(find_unique_constraint_by_key_constraint(*_key_constraint_x, join_unique_constraints));
-    EXPECT_TRUE(find_unique_constraint_by_key_constraint(*_key_constraint_y, join_unique_constraints));
+    // Inner- and outer joins should also forward the right input's UCCs.
+    EXPECT_EQ(join_unique_column_combinations->size(), 4);
+    EXPECT_TRUE(find_ucc_by_key_constraint(*_key_constraint_x, join_unique_column_combinations));
+    EXPECT_TRUE(find_ucc_by_key_constraint(*_key_constraint_y, join_unique_column_combinations));
   }
```
mweisgut, line 637: See above: this brace `}` should probably be in line 623

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -487,121 +487,197 @@ TEST_F(JoinNodeTest, FunctionalDependenciesUnify) {
...
 }
 
-TEST_F(JoinNodeTest, UniqueConstraintsNonEquiJoin) {
-  // Currently, we do not support unique constraint forwarding for Non-Equi- or Theta-Joins
+TEST_P(JoinNodeMultiJoinModeTest, UniqueColumnCombinationsNonEquiJoin) {
+  // Currently, we do not support UCC forwarding for Non-Equi- or Theta-Joins. Semi-, cross-, and Anti-Joins only
```
mweisgut, line 641:
```suggestion
  // Currently, we do not support UCC forwarding for Non-Equi- or Theta-Joins. Semi-, Cross-, and Anti-Joins only
```

mweisgut (commented):
Partially reviewed. Next: `src/test/lib/logical_query_plan/mock_node_test.cpp` (Not before Monday afternoon).

src/lib/storage/constraints/abstract_table_constraint.hpp
```diff
@@ -22,7 +20,10 @@ class AbstractTableConstraint {
 
   virtual ~AbstractTableConstraint() = default;
 
-  const std::set<ColumnID>& columns() const;
+  // Some data dependencies have set semantics for columns, others have list semantics. A list is generally applicable,
```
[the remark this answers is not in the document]
dey4ss: Do you prefer this to be done in this PR (where we only have key constraints with set semantics) or in the PR introducing order or foreign key constraints (whatever comes first) that actually have list semantics?

src/test/lib/logical_query_plan/join_node_test.cpp
dey4ss: Added such a table: https://github.com/hyrise/hyrise/blob/7d18d5669ed497ee31e989e72d1a5614ec1e90ad/src/test/lib/logical_query_plan/join_node_test.cpp#L497

src/lib/logical_query_plan/mock_node.cpp
mweisgut: How about using a convenience lambda function to avoid copies? Something like the following:
```
auto contains = [](auto container, auto value) {
  return std::find(container.begin(), container.end(), value) != container.end();
};
```

src/lib/storage/constraints/abstract_table_constraint.hpp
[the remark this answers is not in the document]
mweisgut: I'm fine with both. If examples are added in a subsequent PR, we might relax the current documentation to `data dependencies might have either set or list semantics`?

src/test/lib/storage/constraints/table_key_constraint_test.cpp
```diff
@@ -145,6 +142,9 @@ TEST_F(TableKeyConstraintTest, Less) {
   EXPECT_FALSE(key_constraint_a < key_constraint_a_reordered);
 
   EXPECT_TRUE(primary_key_constraint_a < key_constraint_a);
+  EXPECT_TRUE(primary_key_constraint_a < key_constraint_a_reordered);
+  EXPECT_TRUE(primary_key_constraint_a < key_constraint_b);
+  EXPECT_TRUE(primary_key_constraint_a < key_constraint_c);
```
mweisgut, line 147: I think we should also add a check for two constraints with the following properties:
`TableKeyConstraint{{<column c1>}, <contraint type t1>}` and `TableKeyConstraint{{<column c1>, <column c2>}, <contraint type t1>}`

dey4ss: This is exactly what L151 does:
```c++
const auto key_constraint_a = TableKeyConstraint{{ColumnID{0}, ColumnID{2}}, KeyConstraintType::UNIQUE};
...
const auto key_constraint_c = TableKeyConstraint{{ColumnID{0}}, KeyConstraintType::UNIQUE};
...
EXPECT_TRUE(key_constraint_c < key_constraint_a);
```

Do you want the same for `KeyConstraintType::PRIMARY_KEY`?

mweisgut: The line was hidden in the diff, i.e., not expanded. No, I'm fine with it, thank you.

mweisgut (approved):

dey4ss:
Waiting for merge until @Bouncner has the chance to look at this.

src/lib/logical_query_plan/data_dependencies/unique_column_combination.cpp
```diff
@@ -1,25 +1,25 @@
...
 }
 
-bool LQPUniqueConstraint::operator!=(const LQPUniqueConstraint& rhs) const {
+bool UniqueColumnCombination::operator!=(const UniqueColumnCombination& rhs) const {
   return !(rhs == *this);
 }
 
-size_t LQPUniqueConstraint::hash() const {
+size_t UniqueColumnCombination::hash() const {
   size_t hash = 0;
```
Bouncner, line 23: auto

src/lib/logical_query_plan/intersect_node.cpp
```diff
@@ -28,18 +28,18 @@ bool IntersectNode::is_column_nullable(const ColumnID column_id) const {
...
-              "Merging of unique constraints should be implemented.");
-  return _forward_left_unique_constraints();
+  DebugAssert(left_input()->unique_column_combinations() == right_input()->unique_column_combinations(),
+              "Merging of unique column combinations should be implemented.");
+  return _forward_left_unique_column_combinations();
 }
 
-std::vector<FunctionalDependency> IntersectNode::non_trivial_functional_dependencies() const {
+FunctionalDependencies IntersectNode::non_trivial_functional_dependencies() const {
   Fail("Merging of FDs should be implemented.");
```
Bouncner, line 43: "is not implemented" would be better here. Otherwise is more a TODO.

src/lib/logical_query_plan/lqp_utils.hpp
```diff
@@ -239,27 +240,30 @@ std::vector<std::shared_ptr<AbstractLQPNode>> lqp_find_leaves(const std::shared_
...
  *         it only works on data source nodes. Currently, these are StoredTableNodes, StaticTableNodes and MockNodes.
  */
-ExpressionUnorderedSet find_column_expressions(const AbstractLQPNode& lqp_node, const std::set<ColumnID>& column_ids);
+ExpressionUnorderedSet find_column_expressions(const AbstractLQPNode& lqp_node,
+                                               const std::vector<ColumnID>& column_ids);
 
 /**
- * @return True, if there is unique constraint in the given set of @param unique_constraints matching the given
- *         set of expressions. A unique constraint matches if it covers a subset of @param expressions.
+ * @return True, if there is unique column combination in the given set of @param unique_column_combinations matching
```
Bouncner, line 247: "a unique column combination"?

src/lib/logical_query_plan/stored_table_node.hpp
```diff
@@ -43,8 +43,8 @@ class StoredTableNode : public EnableMakeForLQPNode<StoredTableNode>, public Abs
   std::vector<std::shared_ptr<AbstractExpression>> output_expressions() const override;
   bool is_column_nullable(const ColumnID column_id) const override;
 
-  // Generates unique constraints from table's key constraints and pays respect to pruned columns.
-  std::shared_ptr<LQPUniqueConstraints> unique_constraints() const override;
+  // Generates unique column combinations from table's key constraints and pays respect to pruned columns.
```
Bouncner, line 46: "a table's"?

Bouncner: Can we reformulate "pays respect"?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -82,44 +82,72 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
...
+      if (left_determinant_size != right_determinant_size) {
+        return left_determinant_size < right_determinant_size;
+      }
+
+      // The FDs are expected to be equally useful for the rewrite. However, we have to decide on semantics here to make
+      // the order independent of the position of the FDs in the original set (which might differ due to standard
+      // library implementation details). Thus, we compare the ColumnIDs of the determinants.
+      const auto& left_column_ids = get_column_ids(fd_left.determinants);
+      const auto& right_column_ids = get_column_ids(fd_right.determinants);
+      return std::lexicographical_compare(left_column_ids.cbegin(), left_column_ids.cend(), right_column_ids.cbegin(),
```
Bouncner, line 134: Can you add a comment why you use `std::lexicographical_compare`?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -82,44 +82,72 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
...
                 std::inserter(group_by_columns, group_by_columns.end()));
       return group_by_columns;
     };
     auto group_by_columns = fetch_group_by_columns();
 
+    // Get a sorted list of ColumnIDs from an FD's set of determinants.
+    const auto get_column_ids = [&](const auto& determinants) {
+      auto column_ids = std::vector<ColumnID>{};
+      column_ids.reserve(determinants.size());
+      std::transform(determinants.cbegin(), determinants.cend(), std::back_inserter(column_ids),
```
Bouncner, line 110: I don't see transform here being better to read than a range-based for loop. I am not asking you to change that, but is there a particular reason you are using transform?

src/lib/types.cpp
```diff
@@ -137,6 +137,10 @@ PredicateCondition conditions_to_between(const PredicateCondition lower, const P
   Fail("Unexpected PredicateCondition");
 }
 
+bool is_semi_or_anti_join(const JoinMode join_mode) {
```
Bouncner, line 140: Good one!

src/lib/visualization/lqp_visualizer.cpp
```diff
@@ -151,35 +151,37 @@ void LQPVisualizer::_build_dataflow(const std::shared_ptr<AbstractLQPNode>& sour
...
                    << "Functional Dependencies (trivial): \n";
     if (trivial_fds.empty()) {
       tooltip_stream << " <none>\n";
     }
 
-    const auto trivial_fd_count = trivial_fds.size();
-    for (auto fd_idx = size_t{0}; fd_idx < trivial_fd_count; ++fd_idx) {
-      tooltip_stream << " (" << fd_idx + 1 << ") ";
-      tooltip_stream << trivial_fds.at(fd_idx) << "\n";
+    auto trivial_fd_idx = 0;
```
Bouncner, line 180: Doesn't matter but just curious why you initialize with 0 just to immediately increment it?

Bouncner (commented):
I skipped the test and a lot of code that looked like renamings while skipping over it. The only "main" issue if have is the difference of the dependency types (shared_ptr vs. none) we pass around. Is there a good reason for it or just legacy?

src/lib/logical_query_plan/abstract_non_query_node.hpp
```diff
@@ -14,8 +14,11 @@ class AbstractNonQueryNode : public AbstractLQPNode {
   using AbstractLQPNode::AbstractLQPNode;
 
   std::vector<std::shared_ptr<AbstractExpression>> output_expressions() const override;
-  std::shared_ptr<LQPUniqueConstraints> unique_constraints() const override;
-  std::vector<FunctionalDependency> non_trivial_functional_dependencies() const override;
+
+  std::shared_ptr<UniqueColumnCombinations> unique_column_combinations() const override;
```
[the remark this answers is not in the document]
dey4ss: UCCs are often just passed around (they do not change in most cases and are forwarded), whereas the "non-trivial FDs" are usually genuine and not passed around that much _but_ are merged, inflated, whatever (making copies necessary in most cases). So having a pointer for UCCs makes sense IMO to ensure nothing is copied too often. In contrast, we have to make copies of the FD structures anyway. But I'm open if you can convince me to unify the interfaces.

src/lib/logical_query_plan/lqp_utils.hpp
dey4ss: changed to "a UCC" to achieve a nicer formatting / brevity.

src/lib/logical_query_plan/stored_table_node.hpp
dey4ss: `Generates unique column combinations from a table's key constraints. Drops UCCs that include pruned columns.`?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
dey4ss: No particular reason. Think a for loop does not look that weird for some ppl, gonna change that.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
dey4ss: Your're right, should make no difference to just use `<`.

src/lib/logical_query_plan/stored_table_node.hpp
Bouncner: Sound good.

src/lib/logical_query_plan/abstract_non_query_node.hpp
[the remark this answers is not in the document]
Bouncner: Are they passed around or really shared?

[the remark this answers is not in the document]
dey4ss: Passed around, so it would be okay to return the sets when always using references insted of copy assignments. It just makes it harder for anyone to create copies by accident.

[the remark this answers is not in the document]
Bouncner: If it is not shared, why not a unique_ptr? That should avoid any overheads.

Trying to avoid copies makes sense, but (i) clang-tidy finds a lot of these issues and (ii) we already ref-to-vector-passing a lot in Hyrise.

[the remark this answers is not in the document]
dey4ss: So do you have preferences for either unique pointers or ref-to-set? I don't have a strong opinion tbh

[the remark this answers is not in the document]
Bouncner: As we use it widely in the code base, I'd prefer references. But if you think there will be cases where someone will cause copies (usage in plugins, unclear interfaces, ...), I am fine with enforcing it via `unique_ptrs`.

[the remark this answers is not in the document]
dey4ss: When sets are the common interface, I think it should be clear then. For plugins, I guess it's the plugin programmer's responsibility not to do stupid things and they should not interfere with query optimization directly at all. So references it shall be. Gonna ping you when that's done.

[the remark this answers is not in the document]
dey4ss: @Bouncner now only `UniqueColumnCombinations` (a.k.a `std::unordered_set<UniqueColumnCombination>`) is passed around.

Bouncner (approved):

PR #2529, opened 2023-01-06
Read benchmarks now are timed on execution not on creation

SanJSp:
We always measured the time of the thread initialization with our benchmarks. This is a proposed solution to start the threads synchronously using a flag and avoid measuring the initialization of the threads. 

Stefan mentioned this here: https://hyrise-db.slack.com/archives/C046JFXJFFS/p1672787912252779?thread_ts=1672739258.520879&cid=C046JFXJFFS

I dont know if this is the best solution, but I think it suffices for the cause

Feel free to add any comments :)


SanJSp:
Sorry wrong repo

PR #2396, opened 2021-08-02
[DYOD] Cardinality estimation: selectivity calculation

leowe:
This PR is part of the DYOD seminar and aims to improve the cardinality estimation for processing SQL queries in hyrise and delivers a tool to do more precise analyses of such estimations.

This was achieved by

**(1) Implementing a new cardinality estimation class - a Top K histogram modeled as a generic histogram (`TopKAsGenericHistogram`) 
(2) Fixing a bug in the selectivity calculation negatively impacting the cardinality estimation 
(3) Implementing a cardinality estimation analyses tool**

This is one of 3 PRs that each implement one described part.

## (2) Selectivity Calculation Bug Fix
We also added a small bugfix that caused the selectivity calculation of an operator on a table to go wrong:

To estimate the cardinality of a result table after a Scan Predicate Operator comparing a Column vs. a Value was applied, the following steps happen:

1. If the predicate_condition is not one of `{IS (NOT) NULL, (NOT) LIKE, BETWEEN ... AND .. }` (these are handled in a separate case), the previously calculated histogram of values in the Column is sliced according to the predicate. The sliced histogram then only holds the values of the column that fulfill the predicate.
2. As a next step, the selectivity of the predicate on the column is calculated as:

```c++
751  selectivity = sliced_histogram->total_count() / scan_statistics_object->total_count();
```

Here `scan_statistics_object` holds the previous, not-sliced histogram. This means that the selectivity is calculated based on the total_count of values in the histogram.

1. Finally, the other columns `AttributeStatistics` are scaled according to the calculated selectivity, and the new row_count estimation for the result table is calculated as:

```c++
774 const auto row_count = Cardinality{input_table_statistics->row_count * selectivity};
```

So the selectivity is applied on the row_count of the input_table.

The estimation error is caused by the fact that histograms do not hold `NULL` values. Therefore the `row_count` of the `input_table_statistics` object can be different from the `total_count` of the histograms for the table's columns.
As the selectivity is applied on the input_table, it should be calculated based on the row_count of the input_table and not on the differing values count in the histogram.

In estimations for other Operator Types (e.g. `inner_equi_join` and `semi_join`), the selectivity is actually already being calculated like this.

```c++
912 const auto left_selectivity = Selectivity{
        left_input_table_statistics.row_count 0 ? cardinality / left_input_table_statistics.row_count : 0.0f};
```

We introduced the same selectivity calculation in the calculation of Scan Predicate Operator comparing a Column vs. a Value estimation and updated the tests accordingly.

PR #2531, opened 2023-01-09
Avoid coverage badge caching and change info text

Bouncner:
This tiny PR does two things:
 - change a text during benchmark loading
 - adds contributors

**Badge**:
The coverage badge URL remains the same (Jenkin's badget) but we internally rewrite the header caching information with the HPI to ensure that badges are not cached. This has been the reason for the previous solution on heroku.

**Benchmark information**:
Change the text for table encoding and stats generation from `Encoding 'lineitem' - no encoding necessary (12 s 248 ms)` to `Processing 'lineitem' - no encoding necessary (12 s 248 ms)`. Since there are stats generated, I am always curious what a lot cores can do for 12 seconds, when no re-encoding is necessary.


dey4ss:
Is there any chance we can host this under an hpi address? So it would be independent of any of us personally

Bouncner:
> Is there any chance we can host this under an hpi address? So it would be independent of any of us personally

This is rather a quick fix for now. But honestly, I don't see a good change running it within the HPI. I haven't found a way to instrument Jenkins to provide the correct headers and running a service within the HPI that is accessible from the outside has always been quite a pain.

I'll ask Matthias if things are different now.

Bouncner:
We might have found a better solution. Will reopen later.

Bouncner:
We found a way to overwrite Jenkin's header information which is also picked up by Github. Thus, badges are no longer cached.

Reopening this PR for this little info text change and to to add missing contributors of Hyrise. The students contributed #2396 as part of the DYOD seminar but missed to add themselves to the contributors list.

dey4ss (approved):
Sure Markus will find the time to review these serious and comprehensive changes?

Bouncner:
He will regret it if he doesn't find the time!

No, I already told him that I did mausgerutscht.

PR #2533, opened 2023-01-17
Rename column `p_retailsize` in TPC-H's `part` table to `p_retailprice`

dey4ss:
According to the TPC-H schema [1, p. 15], the column's name is `p_retailprice` (which also makes more sense). The mistake did not show since the column is not queried. This PR replaces the incorrect column name with the correct one.

[1] https://www.tpc.org/tpc_documents_current_versions/pdf/tpc-h_v3.0.1.pdf

### DIFF
--- a/resources/benchmark/jcch/part.csv.json
+++ b/resources/benchmark/jcch/part.csv.json
@@ -36,7 +36,7 @@
             "type": "string"
         },
         {
-            "name": "p_retailsize",
+            "name": "p_retailprice",
             "nullable": false,
             "type": "float"
         },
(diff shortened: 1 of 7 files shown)


Bouncner (approved):
Hahaha, nice catch. I assume you already grepped the entire code base and found all occurrences.

dey4ss:
> Hahaha, nice catch. 

Was just by chance, I needed the `CREATE TABLE` statements for TPC-H somewhere else, printed them and noticed it when changing `TEXT` to according `(VAR)CHAR` from the spec.

> I assume you already grepped the entire code base and found all occurrences.

Exactly, it's just not use anywhere else than is these few schema definitions.

PR #2535, opened 2023-01-26
Remove dummy table output from maintenance operators

dey4ss:
Just stumbled across table creation and noticed that some maintenance operators return a dummy table. Since others don't (`Insert/Delete`, `Import/Export`, ...) and this is also nothing SQL standardish, this PR removes this behavior.

### DIFF
--- a/src/lib/operators/abstract_operator.cpp
+++ b/src/lib/operators/abstract_operator.cpp
@@ -127,12 +127,9 @@ void AbstractOperator::execute() {
       const auto& lqp_expressions = lqp_node->output_expressions();
       if (!_output) {
         Assert(lqp_expressions.empty(), "Operator did not produce a result, but the LQP expects it to");
-      } else if (std::dynamic_pointer_cast<const AbstractNonQueryNode>(lqp_node) ||
-                 std::dynamic_pointer_cast<const DummyTableNode>(lqp_node)) {
-        // AbstractNonQueryNodes do not have any consumable output_expressions, but the corresponding operators return
-        // 'OK' for better compatibility with the console and the server. We do not assert anything here.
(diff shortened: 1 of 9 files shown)


rrcomtech (approved):

rrcomtech:
Never mind - just tried, if I can actually approve. Did not expect, that I can.

dey4ss:
> Did not expect, that I can.

You can't from now on, I guess 😅

src/lib/operators/abstract_operator.cpp
```diff
@@ -127,12 +127,9 @@ void AbstractOperator::execute() {
       const auto& lqp_expressions = lqp_node->output_expressions();
       if (!_output) {
         Assert(lqp_expressions.empty(), "Operator did not produce a result, but the LQP expects it to");
-      } else if (std::dynamic_pointer_cast<const AbstractNonQueryNode>(lqp_node) ||
```
tjjordan, line 130: Why is the  `AbstractNonQueryNode` kicked out?

tjjordan (approved):
Looks fine!
Just one (probably irrelevant) question.

src/lib/operators/abstract_operator.cpp
dey4ss: Operators implementing behavior of an 'AbstractNonQueryNode' (i.e., `Insert/Update/Delete`, `Create/Drop[Table|View]`, `CreatePreparedPlan`, `Import/Export`, `ChangeMetaTable`) should not have an output (and `AbstractNonQueryNode::output_expressions()` returns an empty vector). Thus, they match the first `if` condition we do not have the exceptional case fo them anymore.

PR #2538, opened 2023-02-07
Remove Boost overrides

dey4ss:
This PR removes overrides for `boost::bimap` and `boost::hana::curry` that were added to backport fixes patched in Boost 1.71 (`boost::bimap`) and Boost 1.74 (`boost::hana::curry`). Ubuntu 20.04 ships Boost 1.71 and we do not aim to support older versions (also, our last server running it was recently updated). Since the override for `boost::hana::curry` was never used, I removed it, too. Successfully built this PR on sidon (Ubuntu 20.04) and executed `hyriseTest`.


tjjordan (approved):
Successfully built this PR on pella (Ubuntu 22.10) and executed hyriseTest.

PR #2539, opened 2023-02-08
Remove Chunk Indexes

tjjordan:
This PR removes the Chunk Indexes because they can not compete with the (hopefully soon finalized) Table Indexes.
It is based on the PartialHashIndex branch so it may not be merged after the PartialHashIndex PR is merged.
Also there is still some (or much) work to to, for instance the IndexScan could be adjusted to support Table Indexes.


dey4ss:
Do you want to convert this to a draft PR until the partial indexes are merged?

tjjordan:
Closed for now because we want smaller PRs.

PR #2542, opened 2023-02-09
Export result of SELECT statements

dey4ss:
This PR allows to export any table that is the result of a SELECT statement, e.g.:
```SQL
COPY (SELECT firstname, lastname FROM students WHERE grade < 1.3) TO 'excellent_students.csv';
```


src/lib/sql/sql_translator.cpp
```diff
@@ -1530,6 +1530,7 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_execute(const hsql::E
 
 // NOLINTNEXTLINE - while this particular method could be made static, others cannot.
 std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_import(const hsql::ImportStatement& import_statement) {
+  AssertInput(!import_statement.whereClause, "Predicates on imported files are not yet supported.");
```
Bouncner, line 1533: Not yet? Do you plan to support them?

dey4ss: Ah, the philosopher :D
No, I do not plan to do so. We would need to read the CSV meta file or the headers if tbl and binary files to know about columns and data types to resolve the columns or completely load the files in the SQL translator to make everything work. Gonna add this as a comment and remove the "yet".

Bouncner: I think the assertion is fine. But removing the "yet" would please the philosopher.

Bouncner (approved):

PR #2545, opened 2023-02-15
Add ForeignKeyConstraint and TableOrderConstraint

dey4ss:
This PR adds constraints that represent inclusion dependencies (INDs) and order dependencies (ODs) on a table level. Translation into these dependencies and management in the LQPs will be part of subsequent PRs.

Furthermore, LQP node tests for UCCs are adapted or added were missing.


Bouncner:
> Translation into these dependencies and management in the LQPs will be part of subsequent PRs.

What do you mean by "Translation into these dependencies [...]"?

src/lib/storage/table.cpp
```diff
@@ -353,36 +353,99 @@ const TableKeyConstraints& Table::soft_key_constraints() const {
...
+
+  // Check validity of key columns from other table.
+  const auto referenced_table_column_count = referenced_table->column_count();
+  for (const auto& column_id : foreign_key_constraint.columns()) {
+    Assert(column_id < referenced_table_column_count, "ColumnID out of range.");
+  }
+
+  {
+    const auto scoped_lock = acquire_append_mutex();
+    Assert(!_foreign_key_constraints.contains(foreign_key_constraint), "ForeignKeyConstraint is already set.");
```
Bouncner, line 409: This will not be measurable, but: you could first insert and then check the result for the bool value. This way, we have only one lookup into the set.

Otherwise, over the years the caused CO2 by this inefficiency can quickly sum up to a flight between Berlin and San Fransisco.

src/lib/storage/table.hpp
```diff
@@ -207,6 +209,15 @@ class Table : private Noncopyable {
   void add_soft_key_constraint(const TableKeyConstraint& table_key_constraint);
   const TableKeyConstraints& soft_key_constraints() const;
 
+  // Adds foreign key constraint s.t. it can be retrieved by soft_foreign_key_constraints() of this table and by
```
Bouncner, line 212: s.t.? Not really needed here, isn't it?

src/lib/storage/table.hpp
```diff
@@ -243,6 +254,11 @@ class Table : private Noncopyable {
   tbb::concurrent_vector<std::shared_ptr<Chunk>, ZeroAllocator<std::shared_ptr<Chunk>>> _chunks;
 
   TableKeyConstraints _table_key_constraints;
+  TableOrderConstraints _table_order_constraints;
+  ForeignKeyConstraints _foreign_key_constraints;
+  // Stores the ForeignKeyCostraints of another table that reference this table. Required for maintenance of inclusion
```
Bouncner, line 259: Ahh ... now I somewhat understand why the code above talked about INDs. Can you elaborate a bit about it? Otherwise it's pretty confusing.

Bouncner (commented):
Haven't looked at the tests yet.

There are a few smaller things. The most important issue for me is the missing explanation of the connection between Foreign Keys and inclusion dependencies.

src/lib/storage/table.hpp
dey4ss: Might also add something to `AbstractTableConstraint` that our constraints are structures on table level that are used to represent / are translated to data dependencies on LQP level.

src/lib/storage/table.cpp
dey4ss: You're absolutely right. We could as well do nothing. Inserting the exact same constraint into a set is not a problem and just a leftover from when we used vectors to store constraints. However, if I think about it, you maybe should not be able to add FK constraints that have a subset/superset of columns (e.g., when having [a, b] in [c, d], you should not add [a] in [c] or [b] in [d] and vice versa) - only the "largest" constraint should be added then, right?

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -0,0 +1,106 @@
...
+std::vector<size_t> sort_permutation(const std::vector<ColumnID>& column_ids) {
+  auto permutation = std::vector<size_t>(column_ids.size());
+  // Fill permutation with [0, 1, ..., n - 1] and order the permutation by sorting column_ids.
+  std::iota(permutation.begin(), permutation.end(), 0);
+  std::sort(permutation.begin(), permutation.end(),
+            [&](auto lhs, auto rhs) { return column_ids[lhs] < column_ids[rhs]; });
+  return permutation;
+}
+
+void apply_permutation_in_place(std::vector<ColumnID>& column_ids, const std::vector<size_t>& permutation) {
```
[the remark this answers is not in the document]
dey4ss: Your're right, I copied from somewhere else where I did this with larger vectors. In practice (with constraints being added by dependency discovery), these vectors will in almost all cases have the size of 1.

src/lib/storage/table.cpp
```diff
@@ -353,36 +353,99 @@ const TableKeyConstraints& Table::soft_key_constraints() const {
...
   }
 }
 
+void Table::add_soft_foreign_key_constraint(const ForeignKeyConstraint& foreign_key_constraint) {
+  Assert(_type == TableType::Data, "ForeignKeyConstraints are not tracked for reference tables across the PQP.");
+
+  Assert(&*foreign_key_constraint.foreign_key_table() == &*this, "ForeignKeyConstraint is added to the wrong table.");
+
+  // Check validity of specified columns.
+  const auto column_count = this->column_count();
```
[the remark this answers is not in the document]
dey4ss: clang requires it:
```
.../table.cpp:393:29: error: variable 'column_count' declared with deduced type 'const auto' cannot appear in its own initializer
  const auto column_count = column_count();
                            ^
1 error generated.
```

src/lib/storage/table.cpp
```diff
@@ -353,36 +353,99 @@ const TableKeyConstraints& Table::soft_key_constraints() const {
...
+
+const ForeignKeyConstraints& Table::referenced_foreign_key_constraints() const {
+  return _referenced_foreign_key_constraints;
+}
+
+void Table::add_soft_order_constraint(const TableOrderConstraint& table_order_constraint) {
+  Assert(_type == TableType::Data, "TableOrderConstraint are not tracked for reference tables across the PQP.");
+
+  // Check validity of columns.
+  const auto column_count = this->column_count();
```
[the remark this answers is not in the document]
dey4ss: (see above)

src/lib/storage/table.cpp
Bouncner: I have no idea what you are talking about.

src/lib/storage/table.cpp
```diff
@@ -353,36 +353,99 @@ const TableKeyConstraints& Table::soft_key_constraints() const {
...
+  const auto referenced_table = foreign_key_constraint.table();
+  Assert(referenced_table, "ForeignKeyConstraint must reference an existing table.");
+
+  // Check validity of key columns from other table.
+  const auto referenced_table_column_count = referenced_table->column_count();
+  for (const auto& column_id : foreign_key_constraint.columns()) {
+    Assert(column_id < referenced_table_column_count, "ColumnID out of range.");
+  }
+
+  {
```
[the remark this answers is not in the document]
dey4ss: nope

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -8,7 +8,9 @@ namespace {
 
 using namespace hyrise;  // NOLINT(build/namespaces)
 
-// Taken from https://stackoverflow.com/a/17074810
+// ForeignKeyConstraints with swapped columns are equivalent. To ensure they are represented the same way, we determine
+// the permutation that sorts `columns` and apply it to `columns` and `foreign_key_columns`. Idea taken from from
```
Bouncner, line 12: from from from from

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -18,25 +20,12 @@ std::vector<size_t> sort_permutation(const std::vector<ColumnID>& column_ids) {
...
-      std::swap(column_ids[original_position], column_ids[target_position]);
-      done[target_position] = true;
-      original_position = target_position;
-      target_position = permutation[target_position];
-    }
-  }
+std::vector<ColumnID> apply_permutation(std::vector<ColumnID>& column_ids, const std::vector<size_t>& permutation){
+  auto sorted_column_ids = std::vector<ColumnID>(column_ids.size());
+
+    std::transform(permutation.begin(), permutation.end(), sorted_column_ids.begin(),
```
Bouncner, line 26: The indent looks wrong.

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -58,9 +47,12 @@ ForeignKeyConstraint::ForeignKeyConstraint(const std::vector<ColumnID>& columns,
...
   // dependencies, we order the columns and apply the same permutation to the included columns. Doing so, we obtain the
   // same costraint for [A.a, A.b] in [B.x, B.y] and [A.b, A.a] in [B.y, B.x].
-  const auto& permutation = sort_permutation(_columns);
-  apply_permutation_in_place(_columns, permutation);
-  apply_permutation_in_place(_foreign_key_columns, permutation);
+  if (columns.size() > 1) {
+    const auto& permutation = sort_permutation(_columns);
+    _columns = apply_permutation(_columns, permutation);
+    _foreign_key_columns = apply_permutation(_foreign_key_columns, permutation);
+
```
Bouncner, line 54: Remove empty line.

src/lib/storage/table.cpp
```diff
@@ -460,11 +459,9 @@ void Table::add_soft_order_constraint(const TableOrderConstraint& table_order_co
     Assert(column_id < column_count, "ColumnID out of range.");
   }
 
-  {
-    const auto scoped_lock = acquire_append_mutex();
+    const auto append_lock = acquire_append_mutex();
```
Bouncner, line 462: Indent ... I guess you are lazy and let format do the entire work here? Tzzz ....

dey4ss: Of course I am. Yes, did not format yet since this was still WIP.
Life's too short to indent manually, I guess.

src/lib/storage/table.cpp
```diff
@@ -22,6 +22,34 @@
 #include "utils/assert.hpp"
 #include "value_segment.hpp"
 
+namespace {
+
+using namespace hyrise;  // NOLINT(build/namespaces)
+
+// Checks if the two vectors have common elements, e.g., lhs = [0, 2] and rhs = [2, 1, 3].
```
Bouncner, line 29: What are expected cardinalities here?

dey4ss: Very few. In most cases just 1. Also, we don't add constraints over and over but maybe 2 or 3 ODs per selected dimension table, one or two UCCs, and one IND.

Bouncner: Yeah, the overhead isn't worth if for such small values. Okay.

src/lib/storage/table.hpp
Bouncner: Might, will, or have? :)

dey4ss: As for now, I already have:

https://github.com/hyrise/hyrise/blob/9235278e8b85861c78e11a7337a1abbdf87f1603/src/lib/storage/constraints/abstract_table_constraint.hpp#L8-L15

src/lib/storage/constraints/table_order_constraint.cpp
```diff
@@ -0,0 +1,40 @@
+#include "table_order_constraint.hpp"
+
+#include <boost/container_hash/hash.hpp>
+
+namespace hyrise {
+
+TableOrderConstraint::TableOrderConstraint(const std::vector<ColumnID>& ordering_columns,
+                                           const std::vector<ColumnID>& ordered_columns)
+    : _ordering_columns(ordering_columns), _ordered_columns(ordered_columns) {}
```
Bouncner, line 9: Just curious, there is nothing we need to check here? Right order etc.?

src/lib/storage/table.cpp
```diff
@@ -22,6 +22,36 @@
...
+bool columns_intersect(const std::vector<ColumnID>& lhs, const std::vector<ColumnID>& rhs) {
+  for (const auto column_id : lhs) {
+    if (std::find(rhs.cbegin(), rhs.cend(), column_id) != rhs.cend()) {
+      return true;
+    }
+  }
+  return false;
+}
+
+// Checks if @param lhs is a sublist of @param rhs, e.g., lhs = [0, 2, 1] and rhs = [0, 2, 1, 3, 4], or vice versa.
```
Bouncner, line 41: Note which function assumes both inputs to be ordered acordingly?? `intersect` does apparently not, `overlap` does.

Besides that,  I would have expected `[0,1,2]` to overlap with `[1,2]`.

src/lib/storage/table.cpp
```diff
@@ -373,39 +403,114 @@ void Table::create_chunk_index(const std::vector<ColumnID>& column_ids, const st
...
-             "Another primary key already exists for this table.");
+  for (const auto& existing_constraint : _table_key_constraints) {
+    // Ensure that no other PRIMARY KEY is defined.
+    Assert(existing_constraint.key_type() == KeyConstraintType::UNIQUE ||
+               table_key_constraint.key_type() == KeyConstraintType::UNIQUE,
+           "Another primary TableKeyConstraint exists for this table.");
 
-      // Ensure there is only one key constraint per column set.
-      Assert(table_key_constraint.columns() != existing_constraint.columns(),
-             "Another key constraint for the same column set has already been defined.");
```
Bouncner, line 404: This would accept Unique constraints like `[0]` and [0,1]`, right? Don't we want to enforce "maximal constraints"?

src/lib/storage/table.cpp
```diff
@@ -373,39 +403,114 @@ void Table::create_chunk_index(const std::vector<ColumnID>& column_ids, const st
...
+  for (const auto& column_id : foreign_key_constraint.primary_key_columns()) {
+    Assert(column_id < referenced_table_column_count, "ColumnID out of range.");
+  }
+
+  const auto append_lock = acquire_append_mutex();
+  for (const auto& existing_constraint : _foreign_key_constraints) {
+    // Do not allow overlapping foreign key constraints. Though a table may have unlimited inclusion dependencies for
+    // the same columns (and especially the existence of [a] in [x] and [b] in [y] does not mean [a, b] in [x, y]
+    // holds), it is reasonable to assume disjoint (soft) foreign keys for now.
+    Assert(!columns_intersect(existing_constraint.foreign_key_columns(), foreign_key_constraint.foreign_key_columns()),
```
Bouncner, line 465: You say "overlapping" and use intersect, while there is a function called `columns_overlap`.

src/lib/storage/table.cpp
```diff
@@ -373,39 +403,114 @@ void Table::create_chunk_index(const std::vector<ColumnID>& column_ids, const st
...
+  const auto referenced_table = foreign_key_constraint.primary_key_table();
+  Assert(referenced_table && &*referenced_table != this, "ForeignKeyConstraint must reference another existing table.");
+
+  // Check validity of key columns from other table.
+  const auto referenced_table_column_count = referenced_table->column_count();
+  for (const auto& column_id : foreign_key_constraint.primary_key_columns()) {
+    Assert(column_id < referenced_table_column_count, "ColumnID out of range.");
+  }
+
+  const auto append_lock = acquire_append_mutex();
```
Bouncner, line 460: Lock after the loop?

Bouncner: Well, no. I get your point. But we probably fail hard once two users concurrently modify constraints, anyways. We don't lock inserts etc. Also most loops might lose references when there are concurrent appends.

src/lib/storage/table.cpp
```diff
@@ -373,39 +403,114 @@ void Table::create_chunk_index(const std::vector<ColumnID>& column_ids, const st
...
+  for (const auto& column_id : table_order_constraint.ordering_columns()) {
+    Assert(column_id < column_count, "ColumnID out of range.");
   }
+
+  // Check validity of ordered columns.
+  for (const auto& column_id : table_order_constraint.ordered_columns()) {
+    Assert(column_id < column_count, "ColumnID out of range.");
+  }
+
+  const auto append_lock = acquire_append_mutex();
```
Bouncner, line 495: Discussion above.

src/lib/storage/table.cpp
```diff
@@ -373,39 +403,114 @@ void Table::create_chunk_index(const std::vector<ColumnID>& column_ids, const st
...
+  for (const auto& existing_constraint : _table_order_constraints) {
+    // Do not allow order constraints that can be represented by existing constraints (and vice versa). For instance,
+    // we do not allow constraints that (i) order the same ordered columns but require fewer or more ordering columns
+    // or (ii) have the same ordering columns but order fewer or more ordered columns.
+    const auto ordering_columns_invalid =
+        columns_overlap(existing_constraint.ordering_columns(), table_order_constraint.ordering_columns()) &&
+        existing_constraint.ordered_columns() == table_order_constraint.ordered_columns();
+    const auto ordered_columns_invalid =
+        columns_overlap(existing_constraint.ordered_columns(), table_order_constraint.ordered_columns()) &&
```
Bouncner, line 504: See question above. Adding `[1,2]` would not fail in `[0,1,2]` is already in place, right?

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -110,7 +110,7 @@ void UccDiscoveryPlugin::_validate_ucc_candidates(const UccCandidates& ucc_candi
     if (std::any_of(soft_key_constraints.cbegin(), soft_key_constraints.cend(),
                     [&column_id](const auto& key_constraint) {
                       const auto& columns = key_constraint.columns();
-                      return columns.size() == 1 && columns.front() == column_id;
+                      return columns.size() == 1 && *columns.cbegin() == column_id;
```
Bouncner, line 113: why?

src/test/lib/logical_query_plan/change_meta_table_node_test.cpp
```diff
@@ -44,4 +44,8 @@ TEST_F(ChangeMetaTableNodeTest, Copy) {
   EXPECT_EQ(*_change_meta_table_node, *_change_meta_table_node->deep_copy());
 }
 
+TEST_F(ChangeMetaTableNodeTest, NoUniqueColumnCombinations) {
+  EXPECT_THROW(_change_meta_table_node->unique_column_combinations(), std::logic_error);
```
Bouncner, line 48: Comment why that fails? Looks like a valid call.

src/test/lib/logical_query_plan/dummy_table_node_test.cpp
```diff
@@ -40,4 +40,8 @@ TEST_F(DummyTableNodeTest, NodeExpressions) {
   ASSERT_EQ(_dummy_table_node->node_expressions.size(), 0u);
 }
 
+TEST_F(DummyTableNodeTest, NoUniqueColumnCombinations) {
+  EXPECT_TRUE(_dummy_table_node->unique_column_combinations().empty());
```
Bouncner, line 44: Comment?

src/test/lib/logical_query_plan/except_node_test.cpp
```diff
@@ -27,9 +27,7 @@ class ExceptNodeTest : public BaseTest {
 
   std::shared_ptr<MockNode> _mock_node1, _mock_node2, _mock_node3;
   std::shared_ptr<ExceptNode> _except_node;
-  std::shared_ptr<LQPColumnExpression> _a;
-  std::shared_ptr<LQPColumnExpression> _b;
-  std::shared_ptr<LQPColumnExpression> _c;
+  std::shared_ptr<LQPColumnExpression> _a, _b, _c;
```
Bouncner, line 30: We usually avoid that.

Bouncner (commented):
Will review the tests later.

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <boost/container_hash/hash.hpp>
+
+namespace {
+
+using namespace hyrise;  // NOLINT(build/namespaces)
+
+// ForeignKeyConstraints with swapped columns are equivalent. To ensure they are represented the same way, we determine
+// the permutation that sorts `foreign_key_columns` and apply it to `foreign_key_columns` and `primary_key_columns`.
+// Idea taken from https://stackoverflow.com/a/17074810
+std::vector<size_t> sort_permutation(const std::vector<ColumnID>& column_ids) {
```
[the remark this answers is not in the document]
dey4ss: Well, it's actually the indexes of a perutation of the input that happens to be sorted. `sort_indexes` sounds okayish

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -0,0 +1,104 @@
...
+      _primary_key_table{primary_key_table} {
+  Assert(_foreign_key_columns.size() == _primary_key_columns.size(),
+         "Invalid number of columns for ForeignKeyConstraint.");
+  // MockNodes can hold ForeignKeyConstraints that do not reference a primary key table (since they mock the node
+  // representing this table). When added to a table, Table::add_soft_foreign_key_constraint(...) checks that the
+  // pointer is set and valid.
+  Assert(foreign_key_table, "ForeignKeyConstraint must reference a table.");
+
+  // For foreign key constraints / INDs with the same columns, the order of the expressions is relevant. For instance,
+  // [A.a, A.b] in [B.x, B.y] is equal to [A.b, A.a] in [B.y, B.x], but not equal to [A.a, A.b] in [B.y, B.x]. To
```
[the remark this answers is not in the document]
dey4ss: Imagine these Tables:

```
   A           B
 a | b       x | y
-------     -------
 1 | u       1 | u
 2 | u       2 | u
```
In this case, `[A.a, A.b] in [B.x, B.y]`/`[A.b, A.a] in [B.y, B.x]`holds: We find the tuples `(1,u)` and `(2,u)` in `A` and in `B`. If we project *both* tables s.t. the columns are flipped (so `A` has columns `a,b` and `B` has columns `x,y`), we preserver this property: `(u,1)` and `(u,2)` are part of both (projected) relations. However, it is a difference when we flip table `A` but not table `B`: Now, the tuples from a look like `(u,1)` and `(u,2)` and the tuples from `B` are `(1,u)` and `(2,u)`. Obviously, they are not the same tuples anymore. Hence, the order matters.

src/lib/storage/constraints/table_key_constraint.cpp
```diff
@@ -4,16 +4,20 @@
 
 namespace hyrise {
 
-TableKeyConstraint::TableKeyConstraint(const std::set<ColumnID>& columns, KeyConstraintType init_key_type)
-    : AbstractTableConstraint{{columns.cbegin(), columns.cend()}}, _key_type{init_key_type} {}
+TableKeyConstraint::TableKeyConstraint(const std::set<ColumnID>& columns, KeyConstraintType key_type)
```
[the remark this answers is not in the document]
dey4ss: Sure, missed this one.

src/lib/storage/constraints/table_order_constraint.cpp
dey4ss: We (could) have a `(Debug)Assert` checking that the columns do not intersect (which we could also check for key constraints). Besides that, no restrictions.

src/lib/storage/table.cpp
dey4ss: Yeah, the point of `overlap` is to match from the beginning/have the same sequence from the start on. `columns_share_start_sequence`?

src/lib/storage/table.cpp
dey4ss: In most cases, we want minimal constraints (if a column is unique, also the projection on this column plus another columns is unique). Anyways, this is only a whitespace change. So you're somehow right, we should for now assume the columns do not intersect.

src/lib/storage/table.cpp
dey4ss: I would argue that one should not add constraints over and over and over concurrently. However, we should ensure that we do not allow a table to have malformed constraints and thus need to ensure we don't add weird stuff just because of concurrent writes. This way, we are just sure to catch this in any given scenario.

src/lib/storage/table.cpp
dey4ss: You're right and I really had to think this through. What basically happens if both your OD examples are valid is that there is a transitive OD: if `[a,b] |-> [x]` and `[b] |-> [x]`, also `[a] |-> [b]` must be valid, and hence, also `[a] |-> [c]`. So we should not look from the beginning here but ensure there are no intersections (might be to pessimistic, but should work for now). I think this would at least make some noise if we ever run into an intricate case

src/plugins/ucc_discovery_plugin.cpp
dey4ss: `.front()` vs. `*.cbegin()`? We discussed to remove `columns` as a member of the `AbstractTableConstraint`. Since a key constraint has set semantics, it now holds a set of columns – and a set has no `.front()`

src/test/lib/logical_query_plan/change_meta_table_node_test.cpp
dey4ss: Gonna add that. Non-query nodes should not be asked for dependencies.

src/test/lib/logical_query_plan/except_node_test.cpp
dey4ss: I know we dow, but we also use this heavily in tests to not have 20 lines of expression declarations. So I guess we need to come up with a rule for that. TBH, I prefer https://github.com/hyrise/hyrise/blob/f7ad570cc30208c71593c0d7f609e34c0f60decb/src/test/lib/logical_query_plan/aggregate_node_test.cpp#L38 over https://github.com/hyrise/hyrise/blob/f7ad570cc30208c71593c0d7f609e34c0f60decb/src/test/lib/logical_query_plan/join_node_test.cpp#L46-L60 since we initialize everything in `SetUp()` anyways. Not gonna fight for it though.

src/plugins/ucc_discovery_plugin.cpp
Bouncner: Got it. Wasn't aware a sorted set does not have `front()`.

src/lib/storage/constraints/foreign_key_constraint.cpp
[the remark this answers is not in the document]
dey4ss: I added this example here and removed the IND part.

src/lib/storage/table.cpp
dey4ss: Simplified to fail if columns intersect.

src/lib/storage/table.cpp
dey4ss: (removed it entirely)

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                           const std::shared_ptr<Table>& table,
+                                           const std::shared_ptr<Table>& foreign_key_table)
+    : AbstractTableConstraint{columns},
+      _foreign_key_columns{foreign_key_columns},
+      _table{table},
+      _foreign_key_table{foreign_key_table} {
+  Assert(_columns.size() == _foreign_key_columns.size(), "Invalid number of columns for ForeignKeyConstraint.");
+  Assert(table && foreign_key_table, "ForeignKeyConstraint must reference two tables.");
+
+  // For INDs with the same columns, the order of the expressions is relevant. For instance, [A.a, A.b] in [B.x, B.y] is
```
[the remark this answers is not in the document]
Bouncner: That's still open, I guess.

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -0,0 +1,106 @@
...
+  // equals to [A.b, A.a] in [B.y, B.x], but not equals to [A.a, A.b] in [B.y, B.x]. To guarantee unambiguous
+  // dependencies, we order the columns and apply the same permutation to the included columns. Doing so, we obtain the
```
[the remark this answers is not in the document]
Bouncner: Still open?

src/lib/storage/constraints/foreign_key_constraint.hpp
```diff
@@ -0,0 +1,64 @@
...
+                            n_name int NOT NULL,
+                            n_regionkey NOT NULL,
+                            PRIMARY KEY (n_nationkey),
+                            FOREIGN KEY (n_regionkey) REFERENCES region(r_regionkey));
+ *
+ * Given this example, @param foreign_key_columns and @param foreign_key_table reference the table with the foreign key
+ * (e.g., nation.n_regionkey) whereas @param primary_key_columns and @param primary_key_table reference the table with
+ * the primary key (e.g., region.r_regionkey).
+ *
+ * Foreign key constraints are translated to inclusion dependencies (INDs) in the LQP.
```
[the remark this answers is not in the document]
Bouncner: But doesn't that contradict your bloody `(u,2)` example? The definition doesn't require any orders.

src/test/lib/logical_query_plan/except_node_test.cpp
Bouncner: Your call.

I I think I'd prefer the long form, but also won't fight for it.
I am afraid you bring U2 to the fight.

src/lib/storage/constraints/foreign_key_constraint.hpp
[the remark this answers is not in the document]
dey4ss: Of course it does. The projection is sensitive to the order, thus, X<sub>1</sub> with [a, b, c] ≠ X<sub>2</sub> as [a, c, b].

src/test/lib/logical_query_plan/except_node_test.cpp
dey4ss: Maybe we just do a poll in the next roundtable?

Nah, for fighting I use the hurting stuff. Maybe I'll punish you with Schlager, Eurodance, or whoever German rapper is currently in the charts. 🤷‍♀️

src/test/lib/logical_query_plan/union_node_test.cpp
```diff
@@ -245,14 +234,33 @@ TEST_F(UnionNodeTest, UniqueColumnCombinationsUnionPositionsInvalidInput) {
   const auto key_constraint_b = TableKeyConstraint{{ColumnID{2}}, KeyConstraintType::UNIQUE};
   _mock_node1->set_key_constraints(TableKeyConstraints{key_constraint_a_b, key_constraint_b});
 
-  const auto mock_node1_changed = static_pointer_cast<MockNode>(_mock_node1->deep_copy());
-  mock_node1_changed->set_key_constraints({key_constraint_a_b});
+  // Input are not allowed to have differing output expressions.
```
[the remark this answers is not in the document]
Bouncner: Or "Input nodes" as below.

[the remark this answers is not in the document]
Bouncner: And isn't that a characteristic of the union node itself? Why does it fail when asking for unique column combinations?

src/test/lib/storage/constraints/foreign_key_constraint_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+  const auto large_foreign_key_constraint =
+      ForeignKeyConstraint{reverse_column_ids, _table_a, reverse_column_ids, _table_b};
+  EXPECT_EQ(large_foreign_key_constraint.foreign_key_columns(), column_ids);
+  EXPECT_EQ(large_foreign_key_constraint.primary_key_columns(), column_ids);
+}
+
+TEST_F(ForeignKeyConstraintTest, AddForeignKeyConstraints) {
+  EXPECT_EQ(_table_a->soft_foreign_key_constraints().size(), 0);
+
+  const auto foreign_key_constraint_1 = ForeignKeyConstraint{{ColumnID{0}}, _table_a, {ColumnID{0}}, _table_b};
```
Bouncner, line 55: These repetitions for each FK don't look nice. Can we get rid of them?

src/test/lib/storage/constraints/foreign_key_constraint_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+  // Invalid because the column id of the primary key table is out of range.
+  EXPECT_THROW(_table_a->add_soft_foreign_key_constraint({{ColumnID{1}}, _table_a, {ColumnID{5}}, _table_b}),
+               std::logic_error);
+
+  // Invalid because inclusion constraint for the given column sets and primary key table already exists.
+  EXPECT_THROW(_table_a->add_soft_foreign_key_constraint({{ColumnID{0}}, _table_a, {ColumnID{1}}, _table_b}),
+               std::logic_error);
+
+  // Invalid,because the primary key table does not exist.
```
Bouncner, line 101: "," > " " ?

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include <numeric>
+
+#include <boost/container_hash/hash.hpp>
+
+namespace {
+
+using namespace hyrise;  // NOLINT(build/namespaces)
+
+// ForeignKeyConstraints with swapped columns are equivalent. To ensure they are represented the same way, we determine
+// the permutation that sorts `foreign_key_columns` and apply it to `foreign_key_columns` and `primary_key_columns`.
```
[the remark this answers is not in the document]
dey4ss: To transform `[b, a] in [y ,x]` to `[a, b] in [x, y]` and not to `[b, a] in [x, y]` (which would now mean every value of `b` is also in `x` instead of `y`).

src/test/lib/logical_query_plan/limit_node_test.cpp
```diff
@@ -13,17 +13,24 @@ using namespace expression_functional;  // NOLINT(build/namespaces)
...
+  std::shared_ptr<LQPColumnExpression> _a, _b;
 };
 
 TEST_F(LimitNodeTest, Description) {
   EXPECT_EQ(_limit_node->description(), "[Limit] 10");
 }
 
 TEST_F(LimitNodeTest, HashingAndEqualityCheck) {
+  _limit_node->set_left_input(nullptr);
   EXPECT_EQ(*_limit_node, *_limit_node);
```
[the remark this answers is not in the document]
dey4ss: Well, it should definitely not be false 😅 
Not very useful though. Remove?

src/test/lib/storage/constraints/foreign_key_constraint_test.cpp
dey4ss:
```c++
const auto foreign_key_constraint_1 = ForeignKeyConstraint{{ColumnID{0}}, _table_a, {ColumnID{0}}, _table_b};
_table_a->add_soft_foreign_key_constraint(foreign_key_constraint);
```
?

Also like this better, but adapted to the style used in `table_key_constraint_test.hpp`

src/test/lib/logical_query_plan/union_node_test.cpp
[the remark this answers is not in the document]
dey4ss: It is a characteristic of a union node with `SetOperationMode::Positions`, yes. However, we cannot check it during construction as the inputs are set later. So remove the check here and leave it to `output_expressions()` to fail in this case?

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -0,0 +1,121 @@
...
+                                           const std::vector<ColumnID>& primary_key_columns,
+                                           const std::shared_ptr<Table>& primary_key_table)
+    : _foreign_key_columns{foreign_key_columns},
+      _foreign_key_table{foreign_key_table},
+      _primary_key_columns{primary_key_columns},
+      _primary_key_table{primary_key_table} {
+  Assert(_foreign_key_columns.size() == _primary_key_columns.size(),
+         "Invalid number of columns for ForeignKeyConstraint.");
+  // In general, ForeignKeyConstraints should always reference the foreign key table and the primary key table. However,
+  // MockNodes can hold ForeignKeyConstraints that do not reference a primary key table (since they mock the node
```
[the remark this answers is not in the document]
dey4ss: reformulated the comment

src/lib/storage/constraints/foreign_key_constraint.cpp
```diff
@@ -0,0 +1,121 @@
...
+  // representing this table). This is fine and we allow such constraints. When we actually add a ForeignKeyConstraint
+  // to a table, we ensure that the constraint references a primary key table:
+  // Table::add_soft_foreign_key_constraint(...) checks that the pointer is set and valid (i.e., the primary key table
```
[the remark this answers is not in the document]
dey4ss: (see above)

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -21,13 +21,19 @@ class JoinNodeTest : public BaseTest {
     _mock_node_a = MockNode::make(
         MockNode::ColumnDefinitions{{DataType::Int, "a"}, {DataType::Int, "b"}, {DataType::Int, "c"}}, "t_a");
     _mock_node_b = MockNode::make(MockNode::ColumnDefinitions{{DataType::Int, "x"}, {DataType::Float, "y"}}, "t_b");
+    _mock_node_c = MockNode::make(
```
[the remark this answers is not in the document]
dey4ss: It's a surprise tool that will help us later (when testing the join node's IND handling. Removed for now.

src/test/lib/logical_query_plan/union_node_test.cpp
```diff
@@ -245,14 +234,33 @@ TEST_F(UnionNodeTest, UniqueColumnCombinationsUnionPositionsInvalidInput) {
...
+  _union_node->set_right_input(_mock_node2);
+  EXPECT_THROW(_union_node->unique_column_combinations(), std::logic_error);
 
   // Input nodes are not allowed to have differing UCCs.
-  EXPECT_EQ(_mock_node1->unique_column_combinations().size(), 2);
-  EXPECT_EQ(mock_node1_changed->unique_column_combinations().size(), 1);
-  _union_node->set_right_input(mock_node1_changed);
+  // clang-format off
+  const auto projection_node =
+  ProjectionNode::make(expression_vector(_a, _b, _c),
```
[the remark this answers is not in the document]
dey4ss: (added a comment)

src/test/lib/logical_query_plan/limit_node_test.cpp
```diff
@@ -13,17 +13,24 @@ using namespace expression_functional;  // NOLINT(build/namespaces)
...
+  std::shared_ptr<MockNode> _mock_node;
+  std::shared_ptr<LQPColumnExpression> _a, _b;
 };
 
 TEST_F(LimitNodeTest, Description) {
   EXPECT_EQ(_limit_node->description(), "[Limit] 10");
 }
 
 TEST_F(LimitNodeTest, HashingAndEqualityCheck) {
+  _limit_node->set_left_input(nullptr);
```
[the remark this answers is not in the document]
Bouncner: Add comment?

Bouncner (approved):
Only a tiny note.

Bouncner (approved):

PR #2547, opened 2023-02-17
Rename `BetweenExpression` and `InExpression`s' `value()` to `operand()`

dey4ss:
The arguments reference expressions to which the predicate should be applied. Thus, these are operands, not values (triggered my inner Monk, also see the other predicate expressions: BinaryPredicateExpression and IsNullExpression). Furthermore, as I grepped the codebase for `value()`, I removed some superfluous calls to `value()` and `has_value()` when dealing with `std::optional`s.

### DIFF
--- a/src/test/lib/optimizer/strategy/column_pruning_rule_test.cpp
+++ b/src/test/lib/optimizer/strategy/column_pruning_rule_test.cpp
@@ -440,8 +440,9 @@ TEST_F(ColumnPruningRuleTest, AnnotatePrunableJoinInput) {
 
       apply_rule(rule, lqp);
 
-      EXPECT_TRUE(join_node->prunable_input_side().has_value()) << "With JoinMode::" << join_mode;
-      EXPECT_EQ(*join_node->prunable_input_side(), prunable_input_side) << "With JoinMode::" << join_mode;
+      SCOPED_TRACE("With JoinMode::" + std::string{magic_enum::enum_name(join_mode)});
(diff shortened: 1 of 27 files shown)


src/test/lib/optimizer/strategy/column_pruning_rule_test.cpp
```diff
@@ -440,8 +440,9 @@ TEST_F(ColumnPruningRuleTest, AnnotatePrunableJoinInput) {
 
       apply_rule(rule, lqp);
 
-      EXPECT_TRUE(join_node->prunable_input_side().has_value()) << "With JoinMode::" << join_mode;
-      EXPECT_EQ(*join_node->prunable_input_side(), prunable_input_side) << "With JoinMode::" << join_mode;
+      SCOPED_TRACE("With JoinMode::" + std::string{magic_enum::enum_name(join_mode)});
```
Bouncner, line 443: On purpose? If so, what are you using it for here?

Bouncner (approved):
Just a single tiny question.

src/test/lib/optimizer/strategy/column_pruning_rule_test.cpp
dey4ss: On purpose to have the `With JoinMode::` stuff once. I think it's okay soince we look over multiple join modes and if we mess up sth in the rule that might break the behavior for one of them. If so, we directly know in which case (we don't see it from the `EXPECT`).

PR #2552, opened 2023-02-22
Fix CI after merge

SanJSp:
Yesterday in our frustration we merged without waiting for the CI to run through. This was bad.

Theresa and Sandro are bad elves!

Here is the fix for it. Sorry if it caused any problems.

But this should now close the topic around our XXL PR. Now we are free (of it ( for now))


SanJSp:
Wrong Repository...

PR #2561, opened 2023-04-12
Remove superfluous DISTINCT

dey4ss:
Thanks to @Bouncner, I read the following note in [a paper](https://www.vldb.org/pvldb/vol16/p1208-liu.pdf):
> It took more than two years for PostgreSQL developers to implement a pattern that
removes the DISTINCT clause if the result is unique by definition, and this feature is
yet to be merged [26].

That's true and it does not look like they will even merge it soon: https://commitfest.postgresql.org/35/2433/

Thus, I was triggered to implement it in Hyrise and prove we can do this in less than half an hour. It does not change anything in benchmarks, but now, we have it.
(To be fair: The Postgres contributors had to implement passing UCCs in the LQP and tried multiple versions. Also, I needed more than half an hour with all tests and an edge case I forgot to cover.)

### DIFF
--- a/src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
+++ b/src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
@@ -96,13 +96,13 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
     if (group_by_columns.size() == node->node_expressions.size() &&
         node->left_input()->has_matching_ucc(group_by_columns)) {
       const auto& output_expressions = aggregate_node.output_expressions();
-      // Remove the AggregateNode if does not change the output expressions.
+      // Remove the AggregateNode if it does not limit or reorder the output expressions.
       if (expressions_equal(output_expressions, node->left_input()->output_expressions())) {
         lqp_remove_node(node);
         return LQPVisitation::VisitInputs;
       }
 
-      // Else, add a ProjectionNode.
+      // Else, replace it with a ProjectionNode.
       const auto projection_node = ProjectionNode::make(output_expressions);
       lqp_replace_node(node, projection_node);
       return LQPVisitation::VisitInputs;
--- a/src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
+++ b/src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
@@ -2,7 +2,6 @@
 
 #include <unordered_map>
 
-#include "expression/abstract_expression.hpp"
 #include "expression/expression_functional.hpp"
 #include "expression/expression_utils.hpp"
 #include "hyrise.hpp"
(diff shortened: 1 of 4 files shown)


Bouncner:
> Thus, I was triggered to implement it in Hyrise and prove we can do this in less than half an hour.

Angeber.

Bouncner (commented):
Only minor things.

src/lib/sql/sql_translator.cpp
```diff
@@ -1186,11 +1186,11 @@ void SQLTranslator::_translate_select_groupby_having(const hsql::SelectStatement
   //
   // This might create unnecessary aggregate nodes when we already have an aggregation that creates unique results:
   // `SELECT DISTINCT a, MIN(b) FROM t GROUP BY a` would have one aggregate that groups by a and calculates MIN(b), and
-  // one that groups by both a and MIN(b) without calculating anything. Fixing this should be done by an optimizer rule
-  // that checks for each GROUP BY whether it guarantees the results to be unique or not. Doable, but no priority.
+  // one that groups by both a and MIN(b) without calculating anything. Fixing this is done by an optimizer rule
```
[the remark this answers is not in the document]
dey4ss: I do. We did two aggregates before, now, we do one (the table is unique for the group keys after the first aggregate).

src/test/lib/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -141,8 +144,8 @@ TEST_F(DependentGroupByReductionRuleTest, IncompleteKey) {
     stored_table_node_b);
   // clang-format on
 
-  const auto actual_lqp = apply_rule(rule, lqp);
   const auto expected_lqp = lqp->deep_copy();
+  const auto actual_lqp = apply_rule(rule, lqp);
```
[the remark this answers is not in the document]
dey4ss: As discussed offline, we copied the optimized version before, which is of course equal, but that's not what is intended.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -102,7 +106,9 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
         return LQPVisitation::VisitInputs;
       }
 
-      // Else, replace it with a ProjectionNode.
+      // Else, replace it with a ProjectionNode. For instance, SELECT DISTINCT n_name, n_regionkey FROM nation (the
+      // column order is different than in the original table) turns from [ Aggregate GROUP BY n_name, n_regionkey ] to
```
Bouncner, line 110: "turns from to"? Sounds strange.

Bouncner (approved):
Just a minor note on a strange sounding sentencing. Otherwise ready to zünd.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
dey4ss: `, which we represent by [ Aggregate GROUP BY n_name, n_regionkey ], is rewritten to ...`?

Bouncner: Jop

Bouncner: Hmm ... maybe I was drunk? It doesn't sound problematic to me today. Feel free to ignore. I might have misread the sentence. :-/

dey4ss: 🤣 
🧨 then?

PR #2562, opened 2023-04-18
Manage Order Dependencies in the LQP

dey4ss:
This PR adds order dependencies (ODs) on an LQP level and forwards them for all LQP nodes.

Do not be afraid when looking at the diff: In most cases, the nodes simply forward the ODs (so not much code and very repetetive, but *Kleinvieh macht auch Mist* for all nodes' cpps, hpps, and tests). Also, I removed some superfluous includes along the way.


src/lib/logical_query_plan/lqp_utils.hpp
```diff
@@ -250,6 +252,13 @@ bool contains_matching_unique_column_combination(const UniqueColumnCombinations&
 FunctionalDependencies fds_from_unique_column_combinations(const std::shared_ptr<const AbstractLQPNode>& lqp,
                                                            const UniqueColumnCombinations& unique_column_combinations);
 
+/**
+ * @return A set of FDs, derived from the given @param unique_column_combinations and based on the output expressions of
```
Bouncner, line 256: update comment.

src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -626,4 +675,38 @@ std::shared_ptr<AbstractLQPNode> find_diamond_origin_node(const std::shared_ptr<
   return nullptr;
 }
 
+bool contains_matching_order_dependency(const OrderDependencies& order_dependencies,
+                                        const std::vector<std::shared_ptr<AbstractExpression>>& ordering_expressions,
+                                        const std::vector<std::shared_ptr<AbstractExpression>>& ordered_expressions) {
+  DebugAssert(!order_dependencies.empty(), "Invalid input: Set of INDs should not be empty.");
```
Bouncner, line 681: INDs?

src/lib/logical_query_plan/mock_node.cpp
```diff
@@ -80,11 +90,11 @@ const std::vector<ColumnID>& MockNode::pruned_column_ids() const {
 
 std::string MockNode::description(const DescriptionMode /*mode*/) const {
   std::ostringstream stream;
```
Bouncner, line 92: auto

src/lib/logical_query_plan/mock_node.cpp
```diff
@@ -1,15 +1,25 @@
 #include "mock_node.hpp"
 
-#include <memory>
-#include <string>
```
Bouncner, line 4: + `"types.hpp"` and `<algorithm>` (at least).

src/lib/logical_query_plan/mock_node.cpp
```diff
@@ -25,7 +35,7 @@ std::shared_ptr<LQPColumnExpression> MockNode::get_column(const std::string& col
     }
   }
 
-  Fail("Couldn't find column named '"s + column_name + "' in MockNode");
+  Fail("Couldn't find column named '" + column_name + "' in MockNode");
```
Bouncner, line 38: Could not

Bouncner: "MockNode."

src/lib/logical_query_plan/update_node.hpp
```diff
@@ -1,9 +1,5 @@
 #pragma once
 
-#include <memory>
-#include <string>
```
Bouncner, line 4: Readd?

src/lib/logical_query_plan/update_node.cpp
```diff
@@ -1,9 +1,6 @@
 #include "update_node.hpp"
 
-#include <memory>
```
Bouncner, line 3: readd?

src/lib/logical_query_plan/union_node.hpp
```diff
@@ -1,11 +1,6 @@
 #pragma once
 
-#include <memory>
-#include <string>
```
Bouncner, line 4: readd.

src/lib/logical_query_plan/mock_node.cpp
```diff
@@ -123,6 +129,26 @@ UniqueColumnCombinations MockNode::unique_column_combinations() const {
...
+  _order_constraints = order_constraints;
+}
+
+OrderDependencies MockNode::order_dependencies() const {
+  auto order_dependencies = OrderDependencies{};
+  for (const auto& order_constraint : _order_constraints) {
+    const auto& ordering_columns = order_constraint.ordering_columns();
+    const auto& ordered_columns = order_constraint.ordered_columns();
+    if (contains_any_column(_pruned_column_ids, ordering_columns) ||
+        contains_any_column(_pruned_column_ids, ordered_columns)) {
```
Bouncner, line 142: Wouldn't  `[a,b] -> [c,d,e]` still work if `e` is pruned?

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -2,7 +2,6 @@
 
 #include "expression/expression_functional.hpp"
```
Bouncner, line 3: Add `<vector>` and all the others?

src/lib/logical_query_plan/union_node.cpp
```diff
@@ -1,10 +1,5 @@
 #include "union_node.hpp"
 
-#include <memory>
-#include <numeric>
-#include <string>
```
Bouncner, line 5: Readd?

src/lib/logical_query_plan/union_node.cpp
```diff
@@ -64,6 +59,24 @@ UniqueColumnCombinations UnionNode::unique_column_combinations() const {
...
+      // Currently, Hyrise does not allow unions of different tables.
+      const auto& left_order_dependencies = _forward_left_order_dependencies();
+      Assert(left_order_dependencies == right_input()->order_dependencies(),
+             "Input tables should have the same order depedencies.");
+      return left_order_dependencies;
+    }
+    case SetOperationMode::Unique:
+      Fail("ToDo, see discussion https://github.com/hyrise/hyrise/pull/2156#discussion_r452803825");
+  }
+  Fail("Unhandled UnionMode");
```
Bouncner, line 77: Periods.

Bouncner (commented):
First batch. Tests will follow.

src/test/lib/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -438,16 +436,121 @@ TEST_F(StoredTableNodeTest, HasMatchingUniqueColumnCombination) {
...
+}
+
+TEST_F(StoredTableNodeTest, OrderDependenciesPrunedColumns) {
+  // Discard ODs that involve pruned columns.
+  _table_a->add_soft_order_constraint({{ColumnID{0}}, {ColumnID{1}}});
+  _table_a->add_soft_order_constraint({{ColumnID{0}}, {ColumnID{2}}});
+  EXPECT_EQ(_table_a->soft_order_constraints().size(), 2);
+
+  _stored_table_node->set_pruned_column_ids({ColumnID{0}, ColumnID{2}});
+  const auto& order_dependencies = _stored_table_node->order_dependencies();
```
Bouncner, line 483: (If I remember correctly, we get a vector her)

Not that it matters but we are copying the vector first and use a reference to the temporarily materialized object. Why not `EXPECT_TRUE(_stored_table_node->order_dependencies().empty());`

src/test/lib/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -438,16 +436,121 @@ TEST_F(StoredTableNodeTest, HasMatchingUniqueColumnCombination) {
...
+  EXPECT_TRUE(order_dependencies.empty());
+}
+
+TEST_F(StoredTableNodeTest, OrderDependenciesTransitive) {
+  // Construct transitive ODs, but do not run into cycles. Furthermore, do not add ODs where LHS and RHS include the
```
Bouncner, line 488: The description sounds strange. Rather "construct ... and check that no cycles are added.". "Check that no ODs where left and right ..." ??

Bouncner: Would what the effect of the second sentence look like? I don't see how that is tested.

src/test/lib/logical_query_plan/union_node_test.cpp
```diff
@@ -259,4 +256,79 @@ TEST_F(UnionNodeTest, UniqueColumnCombinationsUnionAll) {
...
+      _mock_node2);
+  const auto join_node_2 =
+    JoinNode::make(JoinMode::Inner, equals_(_a, _u),
+      _mock_node1,
+      _mock_node2);
+  // clang-format on
+
+  _union_node->set_left_input(join_node_1);
+  _union_node->set_right_input(join_node_2);
+  EXPECT_THROW(_union_node->order_dependencies(), std::logic_error);
```
Bouncner, line 295: I wonder how these would be used in the optimizer. Especially since a simple check for order dependencies can throw (I would expect it returns nullopt or something).

But maybe I just don't have a good idea how to use order dependencies.

Bouncner:
The merge skrewed up GitHub. This appears to be a known issue (https://github.com/orgs/community/discussions/61091).
Changing the base branch does not change anything.

Some say creating a new PR helps but that would delete the discussions. Maybe do that after we resolve all comments?

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -161,6 +154,23 @@ UniqueColumnCombinations AggregateNode::unique_column_combinations() const {
   return unique_column_combinations;
 }
 
+OrderDependencies AggregateNode::order_dependencies() const {
+  auto order_dependencies = OrderDependencies{};
+
+  // Similarly to UCCs, forward ODs if all expressions are part of the GROUP-BY expressions.
+  const auto& input_order_dependencies = left_input()->order_dependencies();
+  const auto& output_expressions = this->output_expressions();
```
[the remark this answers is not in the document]
dey4ss: Aggregated columns look like `avg(a)`, where the input OD contains `a`. Thus, they are filtered out.

src/test/lib/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -192,7 +192,7 @@ TEST_F(DependentGroupByReductionRuleTest, JoinSingleKeyPrimaryKey) {
 
   const auto expected_lqp =
   ProjectionNode::make(expression_vector(add_(column_a_0, 5), add_(column_a_1, 5), sum_(column_b_2)),
-    AggregateNode::make(expression_vector(column_a_0, column_b_0, column_b_2), expression_vector(sum_(column_a_0), sum_(column_a_1), sum_(column_b_2), any_(column_a_1)),  // NOLINT(whitespace/line_length)
+    AggregateNode::make(expression_vector(column_a_0, column_b_2), expression_vector(sum_(column_a_0), sum_(column_a_1), sum_(column_b_2), any_(column_a_1), any_(column_b_0)),  // NOLINT(whitespace/line_length)
```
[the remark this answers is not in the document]
dey4ss: I'll have a look at it, but will anyway only effect TPD-H Q 10.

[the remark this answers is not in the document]
dey4ss: nope, does not seem like that changes anything

src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -447,8 +443,7 @@ ExpressionUnorderedSet find_column_expressions(const AbstractLQPNode& lqp_node,
     }
 
     const auto original_column_id = column_expression->original_column_id;
-    if (std::find(column_ids.cbegin(), column_ids.cend(), original_column_id) != column_ids.cend() &&
-        *column_expression->original_node.lock() == lqp_node) {
+    if (column_ids.contains(original_column_id) && *column_expression->original_node.lock() == lqp_node) {
```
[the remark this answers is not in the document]
dey4ss: We insert into `column_expressions` if we find the ColumnID in `column_ids`. So i guess the emplace part was overseeing that these are different sets.

Second part: Sure. Maybe we can turn the check for `LQPColumnExpression` and the same node to (debug) asserts?

src/test/lib/logical_query_plan/data_dependencies/order_dependency_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+ protected:
+  std::shared_ptr<MockNode> _mock_node_a, _mock_node_b;
+  std::shared_ptr<LQPColumnExpression> _a_a, _a_b, _a_c, _b_x, _b_y;
+};
+
+TEST_F(OrderDependencyTest, InvalidDependencies) {
+  EXPECT_THROW(OrderDependency({}, {_a_a}), std::logic_error);
+  EXPECT_THROW(OrderDependency({_a_a}, {}), std::logic_error);
```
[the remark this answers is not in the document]
dey4ss: Hm, from pure OD side, this is perfectly valid (although not useful for us).

src/test/lib/logical_query_plan/data_dependencies/order_dependency_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+  const auto od_a_to_b = OrderDependency{{_a_a}, {_a_b}};
+  const auto od_a_to_b_c = OrderDependency{{_a_a}, {_a_b, _a_c}};
+  const auto od_a_b_to_c = OrderDependency{{_a_a, _a_b}, {_a_c}};
+
+  // Equal Hash
+  EXPECT_EQ(od_a_to_b.hash(), OrderDependency({_a_a}, {_a_b}).hash());
+  EXPECT_EQ(od_a_to_b_c.hash(), OrderDependency({_a_a}, {_a_b, _a_c}).hash());
+  EXPECT_EQ(od_a_b_to_c.hash(), OrderDependency({_a_a, _a_b}, {_a_c}).hash());
+
+  // Not Equal Hash
```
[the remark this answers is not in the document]
dey4ss: I think we also do this at other places. I agree that it's not a strong assumption, but it could point to cases where we produce hash collisions too easily. Can as well remove if you like.

src/test/lib/logical_query_plan/data_dependencies/order_dependency_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+  EXPECT_TRUE(order_dependencies.contains(od_a_to_b));
+
+  order_dependencies.emplace(od_a_to_b);
+  EXPECT_EQ(order_dependencies.size(), 1);
+
+  order_dependencies.emplace(od_a_to_b_c);
+  EXPECT_EQ(order_dependencies.size(), 2);
+  EXPECT_TRUE(order_dependencies.contains(od_a_to_b_c));
+
+  order_dependencies.emplace(od_a_b_to_c);
```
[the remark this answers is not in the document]
dey4ss: Agree, can as well remove the test.

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -678,6 +688,177 @@ TEST_P(JoinNodeMultiJoinModeTest, UniqueColumnCombinationsMultiPredicateJoin) {
...
+  const auto od_a_to_c = OrderDependency{{_t_a_a}, {_t_a_c}};
+  const auto od_v_to_u = OrderDependency{{_t_c_v}, {_t_c_u}};
+  const auto order_constraint_a_to_c = TableOrderConstraint{{ColumnID{0}}, {ColumnID{2}}};
+  const auto order_constraint_v_to_u = TableOrderConstraint{{ColumnID{1}}, {ColumnID{0}}};
+  _mock_node_a->set_order_constraints({order_constraint_a_to_c});
+  _mock_node_c->set_order_constraints({order_constraint_v_to_u});
+  EXPECT_EQ(_mock_node_a->order_dependencies().size(), 1);
+  EXPECT_EQ(_mock_node_c->order_dependencies().size(), 1);
+
+  // Inner joins forward ODs from left and right input. If the join predicates are equals predicates and the join
```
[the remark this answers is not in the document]
dey4ss: hm, the statement is true and we test multiple cases of ot below. You are referring to case (iii).

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -678,6 +688,177 @@ TEST_P(JoinNodeMultiJoinModeTest, UniqueColumnCombinationsMultiPredicateJoin) {
...
+    EXPECT_EQ(order_dependencies.size(), 2);
+    EXPECT_TRUE(order_dependencies.contains(od_a_to_c));
+    EXPECT_TRUE(order_dependencies.contains(od_v_to_u));
+  }
+
+  // Case (iii): Multiple join predicates.
+  {
+    // clang-format off
+    const auto inner_join_node =
+    JoinNode::make(JoinMode::Inner, expression_vector(equals_(_t_a_a, _t_c_u), equals_(_t_a_b, _t_c_v)),
```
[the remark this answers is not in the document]
dey4ss: I added a comment.

src/test/lib/logical_query_plan/join_node_test.cpp
```diff
@@ -678,6 +688,177 @@ TEST_P(JoinNodeMultiJoinModeTest, UniqueColumnCombinationsMultiPredicateJoin) {
...
+    EXPECT_TRUE(order_dependencies.contains(OrderDependency{{_t_a_a}, {_t_b_y}}));
+    EXPECT_TRUE(order_dependencies.contains(OrderDependency{{_t_b_y}, {_t_a_a}}));
+
+    EXPECT_TRUE(order_dependencies.contains(OrderDependency{{_t_b_x}, {_t_a_a}}));
+    EXPECT_TRUE(order_dependencies.contains(OrderDependency{{_t_b_x}, {_t_a_b}}));
+    EXPECT_TRUE(order_dependencies.contains(OrderDependency{{_t_b_y}, {_t_a_b}}));
+  }
+
+  // Case (vi): Self join. No new ODs should be built. Inner equi join on a = c and ODs [a] |-> [b] and [a] |-> [c]
+  // would lead to [c] |-> [a] and [c] |-> [b] normally.
```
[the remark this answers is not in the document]
dey4ss:
```c++
// Case (vi): Self join. No new ODs should be built. The inner equi join on a = c with input ODs [a] |-> [b] and
//            [a] |-> [c] would lead to [c] |-> [a] and [c] |-> [b], normally. However, we cannot distinguish WHICH
//            version of c orders b (only the one that equals a in the output).
```

src/test/lib/logical_query_plan/union_node_test.cpp
dey4ss: https://github.com/hyrise/hyrise/blob/01629332003bae75fff620bfece98bdb1f962a85/src/test/lib/logical_query_plan/union_node_test.cpp#L281

Just like when we try to union tables with different columns, this currently means that we deliberately broke the query plan. Throwing should be fine.

src/test/lib/logical_query_plan/stored_table_node_test.cpp
dey4ss: So we have ODs `0 |-> 1`, `1 |-> 2` `2 |-> 0`. The `StoredTableNode` should generate all transitive ODs out of the input ODs. Thus, in a naïve implementation, the node would run circles creating these transitive ODs (starting with, e.g., `0 |-> 0`, and never ever stop. This is exactly what is tested here.

src/lib/expression/expression_utils.hpp
```diff
@@ -189,4 +186,11 @@ template <typename ExpressionContainer>
 bool contains_all_expressions(const ExpressionContainer& search_expressions,
                               const std::vector<std::shared_ptr<AbstractExpression>>& expression_vector);
 
+/**
+ * Checks that @param lhs_expressions is the head of @param rhs_expressions, i.e., all expressions of the left-hand
+ * side are the first expressions of the right-hand side in the same order.
```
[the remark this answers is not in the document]
dey4ss: Any suggestions?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -293,6 +288,24 @@ bool AbstractLQPNode::has_matching_ucc(const ExpressionUnorderedSet& expressions
...
+    const std::vector<std::shared_ptr<AbstractExpression>>& ordering_expressions,
+    const std::vector<std::shared_ptr<AbstractExpression>>& ordered_expressions) const {
+  Assert(!ordering_expressions.empty(), "Invalid input. List of ordering expressions should not be empty.");
+  DebugAssert(has_output_expressions({ordering_expressions.cbegin(), ordering_expressions.cend()}),
+              "The given ordering expressions are not a subset of the LQP's output expressions.");
+  Assert(!ordered_expressions.empty(), "Invalid input. List of ordered expressions should not be empty.");
+  DebugAssert(has_output_expressions({ordered_expressions.cbegin(), ordered_expressions.cend()}),
+              "The given ordered expressions are not a subset of the LQP's output expressions.");
+
+  const auto& order_dependencies = this->order_dependencies();
```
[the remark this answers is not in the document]
dey4ss: I mean, the checks at the beginning are sanity checks for the input. I do not see why we should only do them in some cases.

src/lib/logical_query_plan/data_dependencies/order_dependency.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include "expression/lqp_column_expression.hpp"
+#include "logical_query_plan/abstract_lqp_node.hpp"
+
+namespace hyrise {
+
+using namespace expression_functional;  // NOLINT(build/namespaces)
+
+OrderDependency::OrderDependency(const std::vector<std::shared_ptr<AbstractExpression>>& init_ordering_expressions,
+                                 const std::vector<std::shared_ptr<AbstractExpression>>& init_ordered_expessions)
+    : ordering_expressions{init_ordering_expressions}, ordered_expressions{init_ordered_expessions} {
```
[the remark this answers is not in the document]
dey4ss: In #2611, you changed the other way around for `WindowExpression`. So what do we want to make a guideline here?

src/lib/logical_query_plan/mock_node.cpp
dey4ss: We would have to transform to `[a, b] |-> [c, d]`. However, it only really works for this exact case and we ignore that for now.

src/test/lib/logical_query_plan/aggregate_node_test.cpp
```diff
@@ -279,4 +278,44 @@ TEST_F(AggregateNodeTest, FunctionalDependenciesAdd) {
...
+  EXPECT_TRUE(_mock_node->order_dependencies().empty());
+  EXPECT_TRUE(_aggregate_node->order_dependencies().empty());
+
+  const auto od_a_to_b = OrderDependency{{_a}, {_b}};
+  const auto od_a_to_b_c = OrderDependency{{_a}, {_b, _c}};
+  const auto od_a_b_to_c = OrderDependency{{_a, _b}, {_c}};
+  const auto order_constraint_a_to_b = TableOrderConstraint{{ColumnID{0}}, {ColumnID{1}}};
+  const auto order_constraint_a_to_b_c = TableOrderConstraint{{ColumnID{0}}, {ColumnID{1}, ColumnID{2}}};
+  const auto order_constraint_a_b_to_c = TableOrderConstraint{{ColumnID{0}, ColumnID{1}}, {ColumnID{2}}};
+  _mock_node->set_order_constraints({order_constraint_a_to_b, order_constraint_a_to_b_c, order_constraint_a_b_to_c});
```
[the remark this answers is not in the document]
dey4ss: We also do this for UCCs.

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -157,6 +157,23 @@ UniqueColumnCombinations AggregateNode::unique_column_combinations() const {
...
+  for (const auto& input_order_dependency : input_order_dependencies) {
+    if (!(contains_all_expressions(input_order_dependency.ordering_expressions, output_expressions) &&
+          contains_all_expressions(input_order_dependency.ordered_expressions, output_expressions))) {
+      continue;
+    }
+    order_dependencies.emplace(input_order_dependency);
+  }
+
+  return order_dependencies;
```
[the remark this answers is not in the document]
dey4ss: Think we discussed this offline and we basically had originally sets, the pointers, then sets again (https://github.com/hyrise/hyrise/pull/2528#discussion_r1091075056). I do not think this makes a difference now, since we do not request dependencies often.

src/lib/expression/expression_utils.hpp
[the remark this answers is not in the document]
Bouncner: Not any good ones. It it would be strings, something like "is_prefix" would make sense. Not sure if that's obvious for lists though.

src/lib/logical_query_plan/data_dependencies/order_dependency.cpp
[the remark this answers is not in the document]
Bouncner: For WindowExpression it was tidy complaining that we only access the vector to obtain its size (if I remember correctly). Here, we are copying the entire vector.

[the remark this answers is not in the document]
dey4ss: Aah yes, got it, gonna change.

src/lib/expression/expression_utils.hpp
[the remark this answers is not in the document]
dey4ss: `expressions_are_prefix`, `expression_list_is_prefix`?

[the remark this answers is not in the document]
Bouncner: I think `expression_list_is_prefix` would be my favorite, although I still don't like it too much. :D

src/lib/logical_query_plan/aggregate_node.cpp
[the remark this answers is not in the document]
Bouncner: I cannot follow you. :(

src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -457,10 +452,33 @@ ExpressionUnorderedSet find_column_expressions(const AbstractLQPNode& lqp_node,
...
-                                                        const std::vector<ColumnID>& column_ids);
+std::vector<std::shared_ptr<AbstractExpression>> find_column_expressions(const AbstractLQPNode& lqp_node,
+                                                                         const std::vector<ColumnID>& column_ids) {
+  DebugAssert(lqp_node.type == LQPNodeType::StoredTable || lqp_node.type == LQPNodeType::StaticTable ||
+                  lqp_node.type == LQPNodeType::Mock,
+              "Did not expect other node types than StoredTableNode, StaticTableNode and MockNode.");
+  DebugAssert(!lqp_node.left_input(), "Only valid for data source nodes");
+
+  const auto& output_expressions = lqp_node.output_expressions();
+  auto column_expressions = std::vector<std::shared_ptr<AbstractExpression>>(column_ids.size());
```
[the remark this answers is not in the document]
Bouncner: Yes, that would have helped me. Or `get_expressions_for_column_ids`?

src/lib/logical_query_plan/lqp_utils.cpp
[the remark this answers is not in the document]
Bouncner: >  So i guess the emplace part was overseeing that these are different sets.

Yes, indeed. Got it.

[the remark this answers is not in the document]
Bouncner: Resolving since you have already change it to a debug assert.

Bouncner (approved):
Apart from the tiny little remaining issues, everything looks good to me.

src/lib/logical_query_plan/aggregate_node.cpp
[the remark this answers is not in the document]
dey4ss: Alright: Assume we have an input OD `[a, b] |-> [c]`
The aggregate says `GROUP BY a, b` and the aggregation is `SUM(c)`. As there is an FD `{a} => {b}`, the group-by attribute `b` was turned to an `ANY(b)` aggregate. `output_expressions()` resolves the `ANY` aggregates back to the input expression, so we get the list `[a, b, SUM(c)]`. When we check which dependencies hold after the AggregateNode, we check that all expressions of the input OD are part of the output expressions. That is not the case for `c`, as we have `SUM(c)` in the output. Thus, we drop the OD.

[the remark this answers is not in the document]
Bouncner: Got it!

Can you elaborate current comment just a tiny bit? I wouldn't get it from the comment.

src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
```diff
@@ -92,7 +81,8 @@ FunctionalDependencies inflate_fds(const FunctionalDependencies& fds) {
       inflated_fds.insert(fd);
     } else {
       for (const auto& dependent : fd.dependents) {
-        inflated_fds.emplace(fd.determinants, ExpressionUnorderedSet{dependent});
+        auto determinants = fd.determinants;
```
Bouncner, line 84: Do you do that to have a movable copy?

src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
```diff
@@ -137,8 +127,8 @@ FunctionalDependencies deflate_fds(const FunctionalDependencies& fds) {
   }
 
   auto deflated_fds = FunctionalDependencies(fds.size());
-  for (const auto& [determinants, dependents] : existing_fds) {
-    deflated_fds.emplace(determinants, dependents);
+  for (auto [determinants, dependents] : existing_fds) {
+    deflated_fds.emplace(std::move(determinants), std::move(dependents));
```
Bouncner, line 131: You probably have a better  feeling of what is how often called. Is this the the dominating case? If so, should we rather copy a const ref to vector in the constructor?

src/lib/utils/print_utils.cpp
```diff
@@ -126,14 +140,42 @@ void print_table_key_constraints(const std::shared_ptr<const Table>& table, std:
...
-template void print_directed_acyclic_graph<const AbstractOperator>(
-    const std::shared_ptr<const AbstractOperator>& node,
-    const NodeGetChildrenFn<const AbstractOperator>& get_children_fn,
-    const NodePrintFn<const AbstractOperator>& print_node_fn, std::ostream& stream);
+void print_expressions(const ExpressionUnorderedSet& expressions, std::ostream& stream, const std::string& separator) {
+  auto expressions_vector = std::vector<std::shared_ptr<AbstractExpression>>{expressions.begin(), expressions.end()};
+
+  // Obtain first ColumnID found in expression.
+  const auto original_column_id = [](const auto& expression) {
+    // Initialize with maximum value. Thus, we print expressions where we cannot get the original ColumnID last.
```
Bouncner, line 148: Don't get this comment.

src/lib/utils/print_utils.cpp
```diff
@@ -126,14 +140,42 @@ void print_table_key_constraints(const std::shared_ptr<const Table>& table, std:
...
+  std::sort(expressions_vector.begin(), expressions_vector.end(), [&](const auto& lhs, const auto& rhs) {
+    return original_column_id(lhs) < original_column_id(rhs);
+  });
+
+  print_expressions(expressions_vector, stream, separator);
+}
+
+void print_expressions(const std::vector<std::shared_ptr<AbstractExpression>>& expressions, std::ostream& stream,
+                       const std::string& separator) {
+  DebugAssert(!expressions.empty(), "Did not expect empty expression vector.");
```
Bouncner, line 172: So it's okay to print unsorted vectors but sets must the sorted first?
Or do you assume that vectors are always sorted? Then maybe check via `is_sorted`?

Just from an interface point-of-view, it is surprising to have both versions that order differently.

Bouncner (commented):
Small things. Wondering about the decision about when to sort in printing.

src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
dey4ss: Exactly.

src/lib/utils/print_utils.cpp
dey4ss: We only sort the sets to have a deterministic output (which might be slightly fancier with sorted colimns). However, the list-based versions imply that the order matters (e.g., when we print the FK `[a, b] in [d, c]`, we should not swap anything around here.

src/lib/utils/print_utils.cpp
dey4ss: `value_(1)` does not have an original ColumnID from an `LQPColumnExpression`. Thus, the ColumnID here stays in the initialized state and the expression would end up at the end.

src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
dey4ss: I'd say we do not query for dependencies too often, compared to, e.g., output expressions. Both would make a copy at the one or the other point, so I do not think it makes any difference.

src/lib/utils/print_utils.cpp
Bouncner: Ah, got it.

Maybe add a comment stating "Sort for deterministic output and set order not relevant." and below "not sorting as vector interface assume list-order is relevant" or so?

src/lib/utils/print_utils.cpp
Bouncner: `    // Initialize with maximum value to print expressions where we cannot get the original ColumnID last.` ?

src/lib/utils/print_utils.cpp
```diff
@@ -126,18 +140,46 @@ void print_table_key_constraints(const std::shared_ptr<const Table>& table, std:
...
-    const NodePrintFn<const AbstractLQPNode>& print_node_fn, std::ostream& stream);
-template void print_directed_acyclic_graph<const AbstractOperator>(
-    const std::shared_ptr<const AbstractOperator>& node,
-    const NodeGetChildrenFn<const AbstractOperator>& get_children_fn,
-    const NodePrintFn<const AbstractOperator>& print_node_fn, std::ostream& stream);
+void print_expressions(const ExpressionUnorderedSet& expressions, std::ostream& stream, const std::string& separator) {
+  auto expressions_vector = std::vector<std::shared_ptr<AbstractExpression>>{expressions.begin(), expressions.end()};
+
+  // Obtain minimal ColumnID found in expression.
+  const auto original_column_id = [](const auto& expression) {
```
Bouncner, line 147: minimal_column_id instead of original?

PR #2563, opened 2023-04-18
Add guideline for declaring member variables of test classes

dey4ss:
Guideline as discussed at the roundtable. [Click for preview](https://github.com/hyrise/hyrise/blob/dey4ss/test_variables_guideline/CONTRIBUTING.md#formatting-and-naming).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,7 +90,7 @@
       `void function(const size_t /* count */) { ... }`). If the context does not provide a good name (such as
       `/* count */`), use `/* unused */`.
 
-* In general, we declare (member) variables of the same type in an own line each ([isolated declaration](https://clang.llvm.org/extra/clang-tidy/checks/readability/isolate-declaration.html)):
+* Prefer declaring (member) variables of the same type in a new line each ([isolated declaration](https://clang.llvm.org/extra/clang-tidy/checks/readability/isolate-declaration.html)):
   ```c++
   class Foo {
     ...
@@ -98,7 +98,7 @@
     int b;
   };
   ```
-  However, certain test classes have many member variables of the same type, e.g., `std::shared_ptr<LQPColumnExpression>` for LQP node tests. Thus, we permit violations of isolated declaration in such test cases if the number of variables is high (see below). Please use common sense to decide on which format you use.
+  However, certain test classes have many member variables of the same type, e.g., `std::shared_ptr<LQPColumnExpression>` for LQP node tests. Declarations with more than one variables can be used in such test classes if the number of variables is high (see below). Please use common sense to decide on which format you use.
   ```c++
   class FooTest : public BaseTest {
     ...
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -90,6 +90,22 @@
       `void function(const size_t /* count */) { ... }`). If the context does not provide a good name (such as
       `/* count */`), use `/* unused */`.
 
+* In general, we declare (member) variables of the same type in an own line each ([isolated declaration](https://clang.llvm.org/extra/clang-tidy/checks/readability/isolate-declaration.html)):
```
mweisgut, line 93: The other guidelines are rather requests addressed to the reader:
> prefix it with `init`
> comment the parameter name out
> choose clear and concise names

So, for consistency, I would prefer formulating the new guideline in a similar way:
> Prefer declaring (member) variables of the same type in a new line each

CONTRIBUTING.md
```diff
@@ -90,6 +90,22 @@
...
+  ```c++
+  class Foo {
+    ...
+    int a;
+    int b;
+  };
+  ```
+  However, certain test classes have many member variables of the same type, e.g., `std::shared_ptr<LQPColumnExpression>` for LQP node tests. Thus, we permit violations of isolated declaration in such test cases if the number of variables is high (see below). Please use common sense to decide on which format you use.
```
mweisgut, line 101: > Thus, we permit violations of isolated declaration in such test cases if the number of variables is high (see below).

I would change this legislative tone. Addressing the reader, something like the following might work:
```
Declarations with more than one variables can be used in such test cases if [...]
```

CONTRIBUTING.md
```diff
@@ -90,6 +90,22 @@
...
 
+* Prefer declaring (member) variables of the same type in a new line each ([isolated declaration](https://clang.llvm.org/extra/clang-tidy/checks/readability/isolate-declaration.html)):
+  ```c++
+  class Foo {
+    ...
+    int a;
+    int b;
+  };
+  ```
+  However, certain test classes have many member variables of the same type, e.g., `std::shared_ptr<LQPColumnExpression>` for LQP node tests. Declarations with more than one variables can be used in such test classes if the number of variables is high (see below). Please use common sense to decide on which format you use.
```
mweisgut, line 101:
```suggestion
  However, certain test classes have many member variables of the same type, e.g., `std::shared_ptr<LQPColumnExpression>` for LQP node tests. Declarations with more than one variable can be used in such test classes if the number of variables is high (see below). Please use common sense to decide on which format you use.
```

Bouncner:
Do we enforce separate lines via format?

dey4ss:
> Do we enforce separate lines via format?

You mean that 
```c++
int a; int b;
```

becomes
```c++
int a;
int b;
```
?

Just tried it, `format.sh` adds the line break.

Bouncner (approved):
LGTM. @mweisgut : anything to add from your side?

mweisgut (approved):
lgtm

PR #2565, opened 2023-04-28
DYOD preparation: Fix style of style guide

Bouncner:
Unacceptable


dey4ss (approved):
Much code to look at, but after extensive review I agree with you.

PR #2566, opened 2023-05-04
Add Star Schema Benchmark

dey4ss:
This PR integrates the Star Schema Benchmark (SSB) as a Hyrise benchmark executable.


src/benchmarklib/ssb/ssb_table_generator.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+std::unordered_map<std::string, BenchmarkTableInfo> SSBTableGenerator::generate() {
+  generate_csv_tables_with_external_dbgen(_dbgen_path, ssb_table_names, "resources/benchmark/ssb/schema", _path,
+                                          _scale_factor, "-T a");
+
+  // Having generated the .csv files, call the FileBasedTableGenerator just as if those files were user-provided.
+  const auto& generated_tables = FileBasedTableGenerator::generate();
+
+  // FileBasedTableGenerator automatically stores a binary file. Remove the CSV data to save some space.
+  remove_csv_tables(_path);
```
Bouncner, line 29: If there is not a third call that's difference later on: execute the CSV removal in the `generate()` function?

src/bin/console/console.cpp
```diff
@@ -536,6 +543,42 @@ int Console::_generate_tpcds(const std::string& args) {
...
+
+  if (arguments.empty() || arguments.size() > 2) {
+    out("Usage: ");
+    out("  generate_ssb SCALE_FACTOR [CHUNK_SIZE]   Generate SSB tables with the specified scale factor. \n");
+    out("                                           Chunk size is " + std::to_string(Chunk::DEFAULT_SIZE) +
+        " by default. \n");
+    return ReturnCode::Error;
+  }
+
+  const auto scale_factor = std::stof(arguments.at(0));
```
Bouncner, line 557: Why parse with with stof here but later with boost lexical cast?

src/benchmarklib/external_dbgen_utils.cpp
```diff
@@ -0,0 +1,82 @@
...
+void generate_csv_tables_with_external_dbgen(const std::string& dbgen_path, const std::vector<std::string>& table_names,
+                                             const std::string& csv_meta_path, const std::string& tables_path,
+                                             const float scale_factor, const std::string& additional_cli_args) {
+  // NOLINTBEGIN(concurrency-mt-unsafe)
+  // clang-tidy complains that system() is not thread-safe. We ignore this warning as we expect that users will not call
+  // table generator executables in parallel.
+
+  // Check if table data has already been generated (and converted to .bin by the FileBasedTableGenerator).
+  if (!std::filesystem::exists(tables_path + "customer.bin")) {
+    Timer timer;
```
Bouncner, line 20: auto timer

Bouncner (commented):
Looks pretty good to me.

src/benchmarklib/ssb/ssb_table_generator.cpp
dey4ss: you mean in `FileBasedTableGenerator::generate()`? I would not do so. If `FileBasedTableGenerator` removed all CSVs, it removes its source data. In the worst case, this would mean we would have to, e.g., download IMDB data multiple times if we change sth with the binary parser, or we could only remove the files if binary data is cached, ...

src/bin/console/console.cpp
dey4ss: Should not be a difference here, according to [Stack Overflow](https://stackoverflow.com/questions/23582089/is-boostlexical-cast-redundant-with-c11-stoi-stof-and-family). `boost::lexical_cast` seems to be way more powerful, but that's nothing we need here.

Bouncner: They also parse 1e17 etc. So leave it?

dey4ss: I tries some things with `lexical_cast` and STL:
```
(release)> generate_ssb 0.01 1000
SCALE_FACTOR
  std::stof 0.01
  boost::lexical_cast 0.01
CHUNK_SIZE
  std::stoul 1000
  boost::lexical_cast 1000

(release)> generate_ssb 0,01 1,000
SCALE_FACTOR
  std::stof 0
libc++abi: terminating with uncaught exception of type boost::wrapexcept<boost::bad_lexical_cast>: bad lexical cast: source type value could not be interpreted as target
CHUNK_SIZE
 std::stoul 1
libc++abi: terminating with uncaught exception of type boost::wrapexcept<boost::bad_lexical_cast>: bad lexical cast: source type value could not be interpreted as target

(release)> generate_ssb 10'000 100'000
SCALE_FACTOR
 std::stof 10
libc++abi: terminating with uncaught exception of type boost::wrapexcept<boost::bad_lexical_cast>: bad lexical cast: source type value could not be interpreted as target
CHUNK_SIZE
 std::stoul 100
libc++abi: terminating with uncaught exception of type boost::wrapexcept<boost::bad_lexical_cast>: bad lexical cast: source type value could not be interpreted as target

(release)> generate_ssb 1e1 1e2
SCALE_FACTOR
  std::stof 10
  boost::lexical_cast 10
CHUNK_SIZE
  std::stoul 1
libc++abi: terminating with uncaught exception of type boost::wrapexcept<boost::bad_lexical_cast>: bad lexical cast: source type value could not be interpreted as target

```

So actually boost is more explicit by throwing errors while the STL yields garbage when in doubt, which is by definition (`std::toul`):
> Discards any whitespace characters (as identified by calling [std::isspace](https://en.cppreference.com/w/cpp/string/byte/isspace)) until the first non-whitespace character is found, then takes as many characters as possible to form a valid base-n (where n=base) unsigned integer number representation and converts them to an integer value.

Thus, I changed all casts to `boost::lexical` cast for all benchmarks.

src/benchmarklib/external_dbgen_utils.cpp
```diff
@@ -28,8 +28,8 @@ void generate_csv_tables_with_external_dbgen(const std::string& dbgen_path, cons
       auto cmd = std::stringstream{};
       // `2>` in a string seems to break Sublime Text's formatter, so it's split into two strings.
       cmd << "cd " << tables_path << " && " << dbgen_path << "/dbgen -f -s " << scale_factor << " "
-          << additional_cli_args << " "
-          << " -b " << dbgen_path << "/dists.dss";
+          << additional_cli_args << "  -b " << dbgen_path << "/dists.dss >/dev/null 2"
+          << ">/dev/null";
```
Bouncner, line 32: linebreak looks strange. You are redirecting both stdout and stderr?

Bouncner (approved):

src/benchmarklib/external_dbgen_utils.cpp
dey4ss: I did not change anything here from before. Since I did not encounter the problem in the comment, I gust changed it to `<< "/dists.dss >/dev/null 2>&1"`.

src/bin/console/console.cpp
```diff
@@ -536,6 +543,44 @@ int Console::_generate_tpcds(const std::string& args) {
   return ReturnCode::Ok;
 }
 
+int Console::_generate_ssb(const std::string& args) {
```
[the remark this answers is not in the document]
dey4ss: Added a test for SF 0.01

PR #2573, opened 2023-06-08
Limit scale factors for TPC-DS and SSB verification

dey4ss:
TPC-DS verification uses fixed result sets for some queries that are not supported by SQLite, which are only valid for SF 1. Some SSB queries `SUM` integers, which lead to `int` overfloaws in SQLite (and thus, wrong results) for SF > 0.1. This PR ensures that the two benchmark binaries crash when invoked with unfitting SFs and `--verify`.

### DIFF
--- a/src/benchmark/star_schema_benchmark.cpp
+++ b/src/benchmark/star_schema_benchmark.cpp
@@ -63,11 +63,11 @@ int main(int argc, char* argv[]) {
   context.emplace("scale_factor", scale_factor);
   // We cannot verify the results for larger scale factors (SFs) since SQLite overflows integers for aggregation
   // results. We could use dedicated result sets in these cases similar to TPC-DS. However, we need to generate these
-  // result sets using a trustworthy DBMS, such as Postgres. Since we do not consider this worth the effort for now and
-  // in particular, it does not solve the issue of requiring special SFs (the result set is only valid fot the SF it
-  //  was generated with), we simply limit verification to rather small SFs.
-  We empirically tested that errors occur for SF > 0.1 (0.11 due to float
-  // comparison).
+  // result sets using a trustworthy DBMS, such as Postgres. We decided against this approach for two reasons. First,
+  // we do not consider this worth the effort for now. Second, it does not solve the issue of requiring specific SFs:
+  // The result set would only be valid for the SF it was generated on. Thus, we simply limit verification to rather
+  // small SFs.
+  // We empirically figured out that errors do not occur for SF <= 0.1 (0.11 to account for float comparison).
   Assert(!config->verify || scale_factor < 0.11,
          "SSB result verification is only supported fo scale factors <= 0.1 (--scale 0.1).");
 
--- a/src/benchmark/star_schema_benchmark.cpp
+++ b/src/benchmark/star_schema_benchmark.cpp
@@ -61,9 +61,12 @@ int main(int argc, char* argv[]) {
 
   std::cout << "- SSB scale factor is " << scale_factor << std::endl;
   context.emplace("scale_factor", scale_factor);
-  // We cannot verify the results for larger SFs since SQLite overflows integers for aggregation results. We could use
-  // dedicated result sets in these cases similar to TPC-DS. However, we need to generate these result sets using a
-  // trustworthy DBMS, such as Postgres. We empirically tested that errors occur for SF > 0.1 (0.11 due to float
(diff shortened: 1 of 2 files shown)


src/benchmark/star_schema_benchmark.cpp
```diff
@@ -61,6 +61,12 @@ int main(int argc, char* argv[]) {
 
   std::cout << "- SSB scale factor is " << scale_factor << std::endl;
   context.emplace("scale_factor", scale_factor);
+  // We cannot verify the results for larger SFs since SQLite overflows integers for aggregation results. We could use
```
Bouncner, line 64: Is the abbrev. SF known here?

src/benchmark/star_schema_benchmark.cpp
```diff
@@ -61,6 +61,12 @@ int main(int argc, char* argv[]) {
...
+  // dedicated result sets in these cases similar to TPC-DS. However, we need to generate these result sets using a
```
Bouncner, line 65: Mention that Postgres results would be a solution but are currently not worth the effort?
Otherwise I would read that and ask: so why dont you do Postgres?

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -57,13 +60,15 @@ int main(int argc, char* argv[]) {
...
 
-  std::string query_path = "resources/benchmark/tpcds/tpcds-result-reproduction/query_qualification";
+  const auto query_path = std::string{"resources/benchmark/tpcds/tpcds-result-reproduction/query_qualification"};
 
   Assert(std::filesystem::is_directory(query_path), "Query path (" + query_path + ") has to be a directory.");
   Assert(std::filesystem::exists(std::filesystem::path{query_path + "/01.sql"}), "Queries have to be available.");
 
   auto query_generator = std::make_unique<FileBasedBenchmarkItemRunner>(config, query_path, filename_blacklist());
   if (config->verify) {
+    // We can only verify the results for SF 1 since the dedicated result sets were obtained on this SF.
```
Bouncner, line 70: Is SF known?

Bouncner (commented):
Just some comments clear up

src/benchmark/star_schema_benchmark.cpp
dey4ss: It's not only a matter of effort. If we have to use dedicated result sets, this means we can only perform verification on the same SF the result sets were obtained. This does not solve the SF problem at all (though we could probably notice chunk pruning issues when the SF leads to more chunks per table). Gonna add a comment.

src/benchmark/tpcds_benchmark.cpp
```diff
@@ -32,6 +34,7 @@ const std::unordered_set<std::string> filename_blacklist() {
   }
   return filename_blacklist;
 }
+
```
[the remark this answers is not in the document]
dey4ss: I just prefer
```c++
namespace {

void some_function(...) {
 ...
}

}  // namespace
```
(blank line around functions) over 
```c++
namespace {
void some_function(...) {
 ...
}
}  // namespace
```

Guess that's a matter of taste.

src/benchmark/star_schema_benchmark.cpp
```diff
@@ -61,6 +61,15 @@ int main(int argc, char* argv[]) {
 
   std::cout << "- SSB scale factor is " << scale_factor << std::endl;
   context.emplace("scale_factor", scale_factor);
+  // We cannot verify the results for larger scale factors (SFs) since SQLite overflows integers for aggregation
+  // results. We could use dedicated result sets in these cases similar to TPC-DS. However, we need to generate these
+  // result sets using a trustworthy DBMS, such as Postgres. Since we do not consider this worth the effort for now and
+  // in particular, it does not solve the issue of requiring special SFs (the result set is only valid fot the SF it
```
Bouncner, line 67: `fot the SF`

src/benchmark/star_schema_benchmark.cpp
```diff
@@ -61,6 +61,15 @@ int main(int argc, char* argv[]) {
...
+  //  was generated with), we simply limit verification to rather small SFs.
+  We empirically tested that errors occur for SF > 0.1 (0.11 due to float
```
Bouncner, line 69: Blub.

Bouncner (approved):

PR #2575, opened 2023-06-12
[DYOD] Team 1 - Sprint 4

niklasmohrin:


niklasmohrin:
@dey4ss Do you have an idea why the full CI is failing? We only modified the test binaries, so `hyriseConsole` should behave just like before, right? Should we just trigger a re-run?

dey4ss:
We've recently been encountering flaky behavior of said test stage and are investigating the root cause. I restarted the CI pipeline.

src/test/lib/operators/aggregate_test.cpp
```diff
@@ -25,6 +25,14 @@
 
 namespace hyrise {
 
+class OperatorsAggregateHashTest : public BaseTest {};
+
+TEST_F(OperatorsAggregateHashTest, EmptyHash) {
+  const auto expected_empty_hash = 0;
```
dey4ss, line 31: why declare and not `EXPECT_EQ(..., 0)`?

src/test/lib/logical_query_plan/except_node_test.cpp
```diff
@@ -23,17 +24,43 @@ class ExceptNodeTest : public BaseTest {
...
+    EXPECT_EQ(_except_node->is_column_nullable(column_id), false);
+  }
+
+  // both left_input and right_input need to be set
+  EXPECT_THROW(_except_node_without_right_input->is_column_nullable(ColumnID{0}), std::logic_error);
+}
+
+TEST_F(ExceptNodeTest, NonTrivialFunctionalDependencies) {
+  auto fd_a = FunctionalDependency({_a}, {_b, _c});
+  auto fd_b_two_dependents = FunctionalDependency({_b}, {_a, _c});
```
dey4ss, line 54: const?

src/test/lib/operators/limit_test.cpp
```diff
@@ -147,4 +147,14 @@ TEST_F(OperatorsLimitTest, ForwardSortedByFlag) {
   }
 }
 
+TEST_F(OperatorsLimitTest, OnlyIntegralTypes) {
+  auto limit = std::make_shared<Limit>(_table_wrapper, to_expression(AllTypeVariant{"test_string"}));
```
dey4ss, line 151: const? `value_("test_string")` rather than `to_expression(...)`?

dey4ss: also: The name says only integral, but I only see a string value in the test. Add another case with a floating-point?

src/test/lib/logical_query_plan/except_node_test.cpp
```diff
@@ -23,17 +24,43 @@ class ExceptNodeTest : public BaseTest {
...
     _except_node->set_left_input(_mock_node1);
     _except_node->set_right_input(_mock_node1);
+
+    _except_node_without_right_input = ExceptNode::make(SetOperationMode::Positions);
+    _except_node_without_right_input->set_left_input(_mock_node1);
   }
 
   std::shared_ptr<MockNode> _mock_node1, _mock_node2, _mock_node3;
-  std::shared_ptr<ExceptNode> _except_node;
+  std::shared_ptr<ExceptNode> _except_node, _except_node_without_right_input;
```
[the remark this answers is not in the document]
Finn-HPI: Should it be:
```suggestion
  std::shared_ptr<ExceptNode> _except_node;
  std::shared_ptr<ExceptNode> _except_node_without_right_input;
```?
I just chose to continue with this style because it was already heavily used in the surrounding lines (e.g. see _mock_nde1, ...) .

File: src/test/lib/operators/aggregate_test.cpp (hunk 1 above)
Comment (niklasmohrin) [reply src/test/lib/operators/aggregate_test.cpp:31]: I wanted to emphasize that both hashes below equal the same value and are not just coincidentally 0 - would you still prefer the inlined 0?

File: src/test/lib/operators/aggregate_test.cpp [hunk 5]
```diff
@@ -25,6 +25,14 @@
 
 namespace hyrise {
 
+class OperatorsAggregateHashTest : public BaseTest {};
```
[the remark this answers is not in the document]
niklasmohrin: Do you mean `OperatorsAggregateTest`? Because that one is not specific to hashing, but instantiated for both sorting and hashing (which is why I didn't want to add the new test to that test suite)

src/test/lib/logical_query_plan/except_node_test.cpp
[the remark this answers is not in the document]
dey4ss: Exactly. We introduced this guideline recently but did not check every test file for it. So if you encounter stuff that should be changed, please change it.

src/test/lib/operators/aggregate_test.cpp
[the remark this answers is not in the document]
dey4ss: I did not realize that the rest is typed for both operators, so just ignore it

src/test/lib/operators/aggregate_test.cpp
dey4ss: If you ask me like this, I would prefer the inlined `0` with a comment why both values are not just coincidentally `0`:D

niklasmohrin: See the comment I added - I though it made sense, but if you say it truly is a coincidence, I can also remove it again :D

Bouncner (approved):

src/test/lib/logical_query_plan/except_node_test.cpp
```diff
@@ -23,17 +24,48 @@ class ExceptNodeTest : public BaseTest {
...
+
+TEST_F(ExceptNodeTest, NonTrivialFunctionalDependencies) {
+  const auto fd_a = FunctionalDependency({_a}, {_b, _c});
+  const auto fd_b_two_dependents = FunctionalDependency({_b}, {_a, _c});
+
+  const auto non_trivial_dependencies = FunctionalDependencies{fd_a, fd_b_two_dependents};
+  _mock_node1->set_non_trivial_functional_dependencies(non_trivial_dependencies);
+
+  EXPECT_EQ(_except_node->non_trivial_functional_dependencies(), non_trivial_dependencies);
+  // Also works if right_input is not set
```
dey4ss, line 65: Period at the end.

src/test/lib/operators/limit_test.cpp
```diff
@@ -147,4 +147,17 @@ TEST_F(OperatorsLimitTest, ForwardSortedByFlag) {
...
 
+TEST_F(OperatorsLimitTest, OnlyIntegralTypes) {
+  const auto limit_string = std::make_shared<Limit>(_table_wrapper, value_("test_string"));
+  EXPECT_THROW(limit_string->execute(), std::logic_error);
+
+  const auto limit_float = std::make_shared<Limit>(_table_wrapper, value_(2.0f));
+  EXPECT_THROW(limit_float->execute(), std::logic_error);
+}
+
+TEST_F(OperatorsLimitTest, GetName) {
```
dey4ss, line 158: Maybe only `Name` instead of `GetName`?

dey4ss (approved):

src/test/lib/logical_query_plan/except_node_test.cpp
dey4ss, line 65:
```suggestion
  // Also works if right_input is not set.
```

PR #2576, opened 2023-06-13
[DYOD] Team 4 - Sprint 4

rrcomtech:


src/test/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index_test.cpp
```diff
@@ -285,4 +285,26 @@ TEST_F(AdaptiveRadixTreeIndexTest, DenseVectorOfInts) {
   _search_elements(values);
 }
 
+TEST_F(AdaptiveRadixTreeIndexTest, UpperLowerBoundForNonExistingElements) {
+  for (int32_t test_size : {4, 16, 48, 256}) {
+    std::vector<std::optional<int32_t>> ints(test_size);
+    for (auto index = 0; index < test_size; ++index) {
```
dey4ss, line 291: `index` as a loop variable in an index test... I'm not convinced.

src/test/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index_test.cpp
```diff
@@ -285,4 +285,26 @@ TEST_F(AdaptiveRadixTreeIndexTest, DenseVectorOfInts) {
...
+      // leave out the second and second to last element (e.g. 1 & 2, 1 & 14, ...)
+      if (index != 1 && index != test_size - 2) {
+        ints[index] = index;
+      }
+    }
+    auto segment = create_dict_segment_by_type<int32_t>(DataType::Int, ints);
+    auto index =
+        std::make_shared<AdaptiveRadixTreeIndex>(std::vector<std::shared_ptr<const AbstractSegment>>({segment}));
```
dey4ss, line 299: const?

src/test/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index_test.cpp
```diff
@@ -285,4 +285,26 @@ TEST_F(AdaptiveRadixTreeIndexTest, DenseVectorOfInts) {
...
+
+    if (test_size == 4) {
+      EXPECT_EQ(*index->lower_bound({1}), 3);
+    } else {
+      EXPECT_EQ(*index->lower_bound({1}), 2);
+    }
+    EXPECT_EQ(*index->upper_bound({test_size - 2}), test_size - 1);
```
dey4ss, line 306: Can you add some comments here about the expected results?

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -61,4 +61,37 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
   // TODO(XPERIANER): Maybe add a better check than just size, cause the returned iterators should also handle
   // this case, which we right now don't check.
 }
+
+TEST_F(EntireChunkPosListTest, CheckIterators) {
+  auto table = Table::create_dummy_table({{"a", DataType::Int, false}});
```
dey4ss, line 66: const

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -61,4 +61,37 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
...
+  table->append({int32_t{1}});
+  table->append({int32_t{2}});
+  table->append({int32_t{3}});
+
+  EXPECT_EQ(table->chunk_count(), 1);
+  EXPECT_EQ(table->row_count(), 3);
+
+  const auto chunk_id = ChunkID{0};
+  const auto chunk_size = ChunkOffset{3};
```
dey4ss, line 75: Why assign to own variables? I would be happier if I saw an explanation of the values in some comments.

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -61,4 +61,37 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
...
+  // Newly added rows are not in the position list
+  EXPECT_EQ(entire_chunk_pos_list->begin().dereference().chunk_offset, 0);
+  EXPECT_EQ(entire_chunk_pos_list->size(), 3);
+  EXPECT_EQ(entire_chunk_pos_list->begin().distance_to(entire_chunk_pos_list->end()), 3);
+  EXPECT_EQ(entire_chunk_pos_list->cbegin().distance_to(entire_chunk_pos_list->cend()), 3);
+}
+
+TEST_F(EntireChunkPosListTest, MemoryUsage) {
+  const auto chunk_id = ChunkID{0};
+  const auto chunk_size = ChunkOffset{10};
```
dey4ss, line 91: see above. Add a comment on this case, please.

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+  EXPECT_TRUE(directory_exists);
+
+  const auto files = std::vector<std::string>{TEST_DIRECTORY + "/one.test", TEST_DIRECTORY + "/two.test",
+                                              TEST_DIRECTORY + "/three.test", TEST_DIRECTORY + "/four.test"};
+
+  for (const auto& file : files) {
+    std::ofstream{file};
+  }
+  std::filesystem::create_directory(TEST_DIRECTORY + "/test-subdirectory");
```
dey4ss, line 31: 1. This should be in `SetUp()`, with a `_directory_path` member. Even though you only have a single test case, it separates more clearly what you actually test.
2. Please also look at `base_test_hpp`, where we have a `test_data_path` (in which your directory should be created.
3. Please have a look at https://en.cppreference.com/w/cpp/filesystem/create_directory about what `create_directory` does and remove everything you do with `stat` here.

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+  const auto entries = list_directory(TEST_DIRECTORY);
+
+  EXPECT_EQ(entries.size(), files.size());
+  for (const auto& entry : entries) {
+    const auto position = std::find(files.begin(), files.end(), entry);
+    EXPECT_TRUE(position != files.end());
+  }
+
+  // Clean up
+  std::filesystem::remove_all(TEST_DIRECTORY);
```
dey4ss, line 42: This should be in `TearDown()`

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+    std::ofstream{file};
+  }
+  std::filesystem::create_directory(TEST_DIRECTORY + "/test-subdirectory");
+
+  const auto entries = list_directory(TEST_DIRECTORY);
+
+  EXPECT_EQ(entries.size(), files.size());
+  for (const auto& entry : entries) {
+    const auto position = std::find(files.begin(), files.end(), entry);
+    EXPECT_TRUE(position != files.end());
```
dey4ss, line 38: Merge into a single line? Or maybe create a `std::unordered_set` and lookup with `.contains()`?

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -61,4 +61,37 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
   // TODO(XPERIANER): Maybe add a better check than just size, cause the returned iterators should also handle
   // this case, which we right now don't check.
 }
+
+TEST_F(EntireChunkPosListTest, CheckIterators) {
```
dey4ss, line 65: Can you think of a more descriptive test name?

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <filesystem>
+#include <fstream>
+
+#include "base_test.hpp"
+
+namespace hyrise {
+
+class ListDirectoryTest : public BaseTest {};
+
+TEST_F(ListDirectoryTest, ListBasicDirectory) {
```
dey4ss, line 13: Hm, maybe split this into two cases for directory with and without subdir?

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -61,4 +61,35 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
   // TODO(XPERIANER): Maybe add a better check than just size, cause the returned iterators should also handle
```
Bouncner, line 61: Not your's, but ...

Since XPERIANER hasn't done something on Hyrise for years, could you please change it to `TODO(anyone):`?

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -58,7 +58,41 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
...
+  table->append({int32_t{2}});
+  table->append({int32_t{3}});
+
+  EXPECT_EQ(table->chunk_count(), 1);
+  EXPECT_EQ(table->row_count(), 3);
+
+  const auto entire_chunk_pos_list = std::make_shared<const EntireChunkPosList>(ChunkID{0}, ChunkOffset{3});
+
+  table->append({int32_t{4}});
+  // Extra Lines have been added to the table:
```
dey4ss, line 80:
```suggestion
  // One row has been added to the table.
```

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -58,7 +58,41 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
   // Newly added rows are not in the position list
   EXPECT_EQ(entire_chunk_pos_list->size(), 3);
 
-  // TODO(XPERIANER): Maybe add a better check than just size, cause the returned iterators should also handle
+  // TODO(anyone): Maybe add a better check than just size, cause the returned iterators should also handle
```
dey4ss, line 61:
```suggestion
  // TODO(anyone): Maybe add a better check than just size, cause the returned iterators should also handle.
```

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -58,7 +58,41 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
...
+  EXPECT_EQ(table->chunk_count(), 1);
+  EXPECT_EQ(table->row_count(), 3);
+
+  const auto entire_chunk_pos_list = std::make_shared<const EntireChunkPosList>(ChunkID{0}, ChunkOffset{3});
+
+  table->append({int32_t{4}});
+  // Extra Lines have been added to the table:
+  EXPECT_EQ(table->chunk_count(), 1);
+  EXPECT_EQ(table->row_count(), 4);
+  // Newly added rows are not in the position list
```
dey4ss, line 83:
```suggestion
  // Newly added rows are not in the position list.
```

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -58,7 +58,41 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
...
   EXPECT_EQ(entire_chunk_pos_list->size(), 3);
 
-  // TODO(XPERIANER): Maybe add a better check than just size, cause the returned iterators should also handle
+  // TODO(anyone): Maybe add a better check than just size, cause the returned iterators should also handle
   // this case, which we right now don't check.
 }
+
+TEST_F(EntireChunkPosListTest, AfterAddCheckIterators) {
+  // This checks if the EntireChunkPosList correctly handles Iterators, after rows are added
+  // to the table. These added rows should not be contained in the PosList.
```
dey4ss, line 67:
```suggestion
  // This checks that the EntireChunkPosList does not change its iterators after rows were added to the table. These
  // added rows should not be contained in the PosList.
```

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,71 @@
+#include "utils/list_directory.hpp"
+
+#include <sys/stat.h>
+#include <filesystem>
+#include <fstream>
+#include <unordered_set>
+
+#include "base_test.hpp"
```
dey4ss, line 8:
```suggestion
#include <filesystem>
#include <fstream>
#include <unordered_set>

#include "base_test.hpp"

#include "utils/list_directory.hpp"
```

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -58,7 +58,41 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
   // Newly added rows are not in the position list
   EXPECT_EQ(entire_chunk_pos_list->size(), 3);
 
-  // TODO(XPERIANER): Maybe add a better check than just size, cause the returned iterators should also handle
+  // TODO(anyone): Maybe add a better check than just size, cause the returned iterators should also handle
   // this case, which we right now don't check.
 }
+
+TEST_F(EntireChunkPosListTest, AfterAddCheckIterators) {
```
dey4ss, line 65:
```suggestion
TEST_F(EntireChunkPosListTest, InsertDoesNotAffectIterators) {
```

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -58,7 +58,41 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
...
+  EXPECT_EQ(entire_chunk_pos_list->begin().distance_to(entire_chunk_pos_list->end()), 3);
+  EXPECT_EQ(entire_chunk_pos_list->cbegin().distance_to(entire_chunk_pos_list->cend()), 3);
+}
+
+TEST_F(EntireChunkPosListTest, MemoryUsage) {
+  // Memory_usage should return only the sizeof the object, as we do not store or manage any other objects.
+  // Create dummy EntireChunkPosList.
+  const auto entire_chunk_pos_list = std::make_shared<const EntireChunkPosList>(ChunkID{0}, ChunkOffset{10});
+
+  EXPECT_EQ(entire_chunk_pos_list->memory_usage(MemoryUsageCalculationMode::Full), sizeof(EntireChunkPosList));
```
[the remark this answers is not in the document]
Tratori: Would you prefer iterating over the member sizes? Or just leave this test out? The only purpose on my part was to increase coverage...

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -58,7 +58,41 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
...
+  EXPECT_EQ(table->row_count(), 3);
+
+  const auto entire_chunk_pos_list = std::make_shared<const EntireChunkPosList>(ChunkID{0}, ChunkOffset{3});
+
+  table->append({int32_t{4}});
+  // One row has been added to the table.
+  EXPECT_EQ(table->chunk_count(), 1);
+  EXPECT_EQ(table->row_count(), 4);
+  // Newly added rows are not in the position list.
+  EXPECT_EQ(entire_chunk_pos_list->begin().dereference().chunk_offset, 0);
```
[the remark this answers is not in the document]
Tratori: Because we look at the first element of the chunk? The last element of the PLI would be equal to 3, I guess. (Which is kind when checking the distance further down).

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -58,7 +58,41 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
...
+
+  const auto table = Table::create_dummy_table({{"a", DataType::Int, false}});
+  table->append({int32_t{1}});
+  table->append({int32_t{2}});
+  table->append({int32_t{3}});
+
+  EXPECT_EQ(table->chunk_count(), 1);
+  EXPECT_EQ(table->row_count(), 3);
+
+  const auto entire_chunk_pos_list = std::make_shared<const EntireChunkPosList>(ChunkID{0}, ChunkOffset{3});
```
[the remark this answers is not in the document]
Tratori: The list's iterators shouldn't change when we change the table's content. That's also kind of what I intended to test here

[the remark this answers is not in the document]
Bouncner: There is no connection between these two data structures. So why should something change?

I am not super happy with the test, but we discussed it yesterday and it might be of help, once the list is changed and actually somewhat aware of the table.

Hence, can you add a comment like `// Currently, the EjtirePosList is not linked to the table. We thus don't expect changes when table change. Nonetheless, this test ensures the behaviour remains the same when link is connected/aware of table.` or so?

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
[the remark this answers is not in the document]
Bouncner: We could do something like this test here: https://github.com/hyrise/hyrise/blob/59fc081345511932ef91f55a6764a2d469b152d1/src/test/lib/storage/fixed_string_dictionary_segment_test.cpp#L142

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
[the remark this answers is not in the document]
rrcomtech: See 84b2372

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,70 @@
...
+ protected:
+  void SetUp() override {
+    std::filesystem::create_directory(_directory_path);
+  }
+
+  void TearDown() override {
+    std::filesystem::remove_all(_directory_path);
+  }
+
+  const std::string _directory_path = test_data_path + "/test-directory";
```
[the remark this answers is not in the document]
Bouncner: Sorry, I was inprecise. The issue is less with concurrently running tests within a test instance (we don't do that), but with concurrently running hyriseTest instances (in the CI, the tests are run for GCC/clang; release/debug; etc.).

A mutex does not help in this case.

BUT: the methods Daniel mentioned here (https://github.com/hyrise/hyrise/pull/2576#discussion_r1231114584) already ensure that you probably do not need to do anything. 
So, sorry for bringing that up.

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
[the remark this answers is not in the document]
Tratori: Fair. After reconsideration with the team, we would also prefer to just leave it out, as there is no real benefit here...

src/test/lib/utils/list_directory_test.cpp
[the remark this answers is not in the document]
rrcomtech: I see. I fear, that would mess up a lot. Therefore, I now added a mechanism, that always adds a somewhat random ID to the folder path, which should avoid parallel accesses to the same folder.

src/test/lib/storage/pos_lists/entire_chunk_pos_list_test.cpp
```diff
@@ -58,7 +58,35 @@ TEST_F(EntireChunkPosListTest, AddAfterMatchedAllTest) {
...
+  // This checks that the EntireChunkPosList does not change its iterators after rows were added to the table. These
+  // added rows should not be contained in the PosList.
+
+  const auto table = Table::create_dummy_table({{"a", DataType::Int, false}});
+  table->append({int32_t{1}});
+  table->append({int32_t{2}});
+  table->append({int32_t{3}});
+
+  // Currently, the EntireChunkPosList is not linked to the table. Therefore, we do not expect changes when
+  // the table changes. Nevertheless, this test ensures the behavior remains the same.
```
Bouncner, line 75:
```suggestion
  // Currently, the EntireChunkPosList is not linked to the table. Therefore, we do not expect changes to the PosList
  // when new tuples are appended to the table. Nevertheless, we added this test to ensure this assumption is still
  // true even when the EntireChunkPosList might later be aware of the referenced table.
```

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,81 @@
+#include <filesystem>
+#include <fstream>
+#include <unordered_set>
+#include <random>
```
Bouncner, line 4:
```suggestion
```

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+ protected:
+  ListDirectoryTest() {
+    // You can do set-up work for each test here.
+    auto rand_dev = std::random_device();
+    auto generator = std::mt19937();
+    auto distribution = std::uniform_int_distribution<uint32_t>(0, INT_MAX);
+
+    _directory_path.append(std::to_string(distribution(generator)));
+  }
+
```
Bouncner, line 23:
```suggestion
```

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+  void TearDown() override {
+    std::filesystem::remove_all(_directory_path);
+  }
+
+  std::string _directory_path = test_data_path + "/test-directory";
+};
+
+TEST_F(ListDirectoryTest, ListDirectoryWithFilesAndSubdirectory) {
+  const auto files = std::unordered_set<std::string>{_directory_path + "/one.test", _directory_path + "/two.test",
+                                                     _directory_path + "/three.test", _directory_path + "/four.test"};
```
Bouncner, line 37:
```suggestion
  const auto files = std::vector<std::string>{_directory_path + "/one.test", _directory_path + "/two.test",
                                              _directory_path + "/three.test", _directory_path + "/four.test"};
```

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+TEST_F(ListDirectoryTest, ListEmptyDirectory) {
+  EXPECT_EQ(list_directory(_directory_path).size(), 0);
+}
+
+TEST_F(ListDirectoryTest, ListDirectoryWithFilesOnly) {
+  const auto files = std::unordered_set<std::string>{_directory_path + "/one.test", _directory_path + "/two.test",
+                                                     _directory_path + "/three.test", _directory_path + "/four.test"};
```
Bouncner, line 59:
```suggestion
  const auto files = std::vector<std::string>{_directory_path + "/one.test", _directory_path + "/two.test",
                                              _directory_path + "/three.test", _directory_path + "/four.test"};
```

src/test/lib/utils/list_directory_test.cpp
```diff
@@ -0,0 +1,81 @@
+#include <filesystem>
+#include <fstream>
+#include <unordered_set>
```
Bouncner, line 3:
```suggestion
#include <vector>
```

src/test/lib/utils/list_directory_test.cpp
rrcomtech: This was changed to an unordered set following a suggestion [here](https://github.com/hyrise/hyrise/pull/2576#discussion_r1231117267) by Daniel. I also think it is a more elegant implementation, since we later can use `contains(..)` to look up the values. 

I was first a little worried, since `contains(..)` is introduced in C++20, but it seems the pipeline likes it anyway.

Bouncner (approved):

PR #2580, opened 2023-06-15
Recursive Scheduler Test and Scheduler Shutdown Fix

Bouncner:
Planned work on Hyrise might recursively add tasks to the scheduler. This PR adds a test doing exactly that (a simple merge sort).

This test triggered a minor issue when shutting down the scheduler. This issue should be fixed with the PR.

Minor:
- changes to Jenkins are solely to avoid starting multiple Hyrise instances using all the threads of the system (the tests are using small scale factors, but we still spawns (partially spinning) worker threads on each core).
- changes in the benchmarklib are solely minor stuff as auto-to-stick

### DIFF
--- a/src/lib/scheduler/node_queue_scheduler.cpp
+++ b/src/lib/scheduler/node_queue_scheduler.cpp
@@ -69,13 +69,13 @@ void NodeQueueScheduler::wait_for_all_tasks() {
     std::this_thread::sleep_for(std::chrono::milliseconds(10));
   }
 
-  auto queue_check_runs = size_t{0};
   for (auto& queue : _queues) {
+    auto queue_check_runs = size_t{0};
     while (!queue->empty()) {
       // The following assert checks that we are not looping forever. The empty() check can be inaccurate for
       // concurrent queues when many tiny tasks have been scheduled (see MergeSort scheduler test). When this assert is
       // triggered in other situations, there have probably been new tasks added after wait_for_all_tasks() was called.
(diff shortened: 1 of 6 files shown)


src/lib/scheduler/node_queue_scheduler.cpp
```diff
@@ -68,6 +68,18 @@ void NodeQueueScheduler::wait_for_all_tasks() {
...
+  auto queue_check_runs = size_t{0};
+  for (auto& queue : _queues) {
+    // The following assert checks that we are not looping forever. The empty() check can be inaccurate for concurrent
+    // queues when many tiny tasks have been scheduled (see MergeSort scheduler test). When this assert is triggered in
+    // other situations, there have probably been new tasks added after wait_for_all_tasks() was called.
+    Assert(queue_check_runs < 1'000, "Queues are not empty but all registered tasks have already been processed.");
+    while (!queue->empty()) {
+      std::this_thread::sleep_for(std::chrono::milliseconds(1));
+    }
+    ++queue_check_runs;
```
dey4ss, line 81: Do I get this right and you increment `queue_check_runs` once whenever a queue is not empty rather than once for each `empty()` call/sleep. How will this variable ever be large/larger than the number of queues?

Bouncner: Argh, rewrote the loop from a `while (true)` loop. Thanks. Will rewrite the check.

Bouncner: Should be addressed.

dey4ss (approved):

PR #2581, opened 2023-06-20
Use `std::stringstream` for default-type encoding message

ClFeSc:
This PR uses a `std::stringstream` to encapsulate writing the default-type encoding message to `std::cout` to prevent messages from multiple threads being mangled.

Example output line without this change:

```
orders.o_totalprice' of type float cannot be encoded as FixedStringDictionary and is - Column 'left Unencoded.lineitem
```

### DIFF
--- a/src/benchmarklib/benchmark_table_encoder.cpp
+++ b/src/benchmarklib/benchmark_table_encoder.cpp
@@ -1,7 +1,7 @@
 #include "benchmark_table_encoder.hpp"
 
-#include <syncstream>
 #include <atomic>
+#include <syncstream>
 
 #include "encoding_config.hpp"
 #include "hyrise.hpp"
--- a/src/benchmarklib/benchmark_table_encoder.cpp
+++ b/src/benchmarklib/benchmark_table_encoder.cpp
@@ -1,7 +1,7 @@
 #include "benchmark_table_encoder.hpp"
 
 #include <atomic>
-#include <syncstream>
+#include <mutex>
 
 #include "encoding_config.hpp"
 #include "hyrise.hpp"
@@ -103,11 +103,11 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
       // orders.o_totalprice' of type float cannot be encoded as FixedStringDictionary and is  - Column 'left Unencoded.lineitem // NOLINT
       // Example with osyncstream:
       // - Column 'part.p_size' of type int cannot be encoded as FixedStringDictionary and is left Unencoded.
-      auto ostream = std::osyncstream(std::cout);
-      ostream << " - Column '" << table_name << "." << table->column_name(column_id) << "' of type ";
-      ostream << column_data_type << " cannot be encoded as ";
-      ostream << encoding_config.default_encoding_spec.encoding_type << " and is ";
-      ostream << "left Unencoded." << std::endl;
+      auto lock = std::lock_guard{output_stream_mutex};
+      std::cout << " - Column '" << table_name << "." << table->column_name(column_id) << "' of type ";
+      std::cout << column_data_type << " cannot be encoded as ";
+      std::cout << encoding_config.default_encoding_spec.encoding_type << " and is ";
+      std::cout << "left Unencoded." << std::endl;
       chunk_encoding_spec.emplace_back(EncodingType::Unencoded);
     }
   }
@@ -138,5 +138,7 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
 
   return encoding_performed;
 }
+
+std::mutex BenchmarkTableEncoder::output_stream_mutex{};
 
 }  // namespace hyrise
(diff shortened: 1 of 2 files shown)


src/benchmarklib/benchmark_table_encoder.cpp
```diff
@@ -97,10 +98,12 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
     if (encoding_supports_data_type(encoding_config.default_encoding_spec.encoding_type, column_data_type)) {
       chunk_encoding_spec.push_back(encoding_config.default_encoding_spec);
     } else {
-      std::cout << " - Column '" << table_name << "." << table->column_name(column_id) << "' of type ";
-      std::cout << column_data_type << " cannot be encoded as ";
-      std::cout << encoding_config.default_encoding_spec.encoding_type << " and is ";
-      std::cout << "left Unencoded." << std::endl;
+      // Use an osyncstream to prevent multiple threads writing at the same time to std::cout.
+      auto ostream = std::basic_osyncstream(std::cout);
```
Bouncner, line 102: The comment is okay, but I think using a syncstream makes the problem here already quite clear.

Further, please use `std::osyncstream` without the `basic_` prefix.

src/benchmarklib/benchmark_table_encoder.cpp
```diff
@@ -97,10 +98,12 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
...
+      ostream << " - Column '" << table_name << "." << table->column_name(column_id) << "' of type ";
+      ostream << column_data_type << " cannot be encoded as ";
+      ostream << encoding_config.default_encoding_spec.encoding_type << " and is ";
+      ostream << "left Unencoded." << std::endl;
```
Bouncner, line 106: Can you add a short example of what you are fixing with this PR and how the result looks?

src/benchmarklib/benchmark_table_encoder.cpp
```diff
@@ -1,5 +1,6 @@
 #include "benchmark_table_encoder.hpp"
 
+#include <syncstream>
```
Bouncner, line 3: Please sort alphabetically.

ClFeSc: But then the linter complains with `src/benchmarklib/benchmark_table_encoder.cpp:4:  Found C system header after C++ system header. Should be: benchmark_table_encoder.h, c system, c++ system, other.  [build/include_order] [4]`.

Bouncner:
Your force-pushing scares me.

src/benchmarklib/benchmark_table_encoder.cpp
Bouncner: Can you please push "this error" so that I can see the CI issue?

src/benchmarklib/benchmark_table_encoder.cpp
```diff
@@ -97,6 +98,12 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
     if (encoding_supports_data_type(encoding_config.default_encoding_spec.encoding_type, column_data_type)) {
       chunk_encoding_spec.push_back(encoding_config.default_encoding_spec);
     } else {
+      // Use an osyncstream to prevent multiple threads writing at the same time to std::cout.
+      // Example output without osyncstream:
+      // orders.o_totalprice' of type float cannot be encoded as FixedStringDictionary and is  - Column 'left Unencoded.lineitem // NOLINT
+      // Example with osyncstream:
+      // - Column 'part.p_size' of type int cannot be encoded as FixedStringDictionary and is left Unencoded.
+      auto lock = std::lock_guard{output_stream_mutex};
```
Bouncner, line 106: Please don't use locking, when we don't need it.
Writing to `stdout` via `std::cout` is thread-safe (within a single write, there will not by any problems).

phkeese: Understood, I was wondering why you recommended a string stream. Will push a fix.

src/benchmarklib/benchmark_table_encoder.cpp
```diff
@@ -97,10 +97,13 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
     if (encoding_supports_data_type(encoding_config.default_encoding_spec.encoding_type, column_data_type)) {
       chunk_encoding_spec.push_back(encoding_config.default_encoding_spec);
     } else {
-      std::cout << " - Column '" << table_name << "." << table->column_name(column_id) << "' of type ";
-      std::cout << column_data_type << " cannot be encoded as ";
-      std::cout << encoding_config.default_encoding_spec.encoding_type << " and is ";
-      std::cout << "left Unencoded." << std::endl;
+      // Use a stringstream here to bundle all writes into a single one and avoid locking.
+      std::ostringstream output{};
```
dey4ss, line 101: auto to stick :)

src/benchmarklib/benchmark_table_encoder.cpp
ClFeSc: Ah, I see. Updated it.

PR #2583, opened 2023-06-21
Support Clang-16

dey4ss:
This PR makes small changes to support builds with clang 16. Notably, we suppress the `-Wunsafe-buffer-usage` check since we access buffers via pointers / pointer arithmetic in some cases when we know what we do (e.g., uninitialized vector works like this). The rest are mainly changes to silence clang-tidy-16. However, currently encounter build errors for tidy builds (see #258), so we do not update the CI yet.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -54,7 +54,11 @@ if (NOT "${HYRISE_RELAXED_BUILD}")
     # -Wno-deprecated-dynamic-exception-spec is needed for jemalloc, at least for the older version that we are using
 
     if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-        add_compile_options(-Weverything -Wshadow-all -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-documentation -Wno-padded -Wno-global-constructors -Wno-sign-conversion -Wno-exit-time-destructors -Wno-switch-enum -Wno-weak-vtables -Wno-double-promotion -Wno-covered-switch-default -Wno-unused-macros -Wno-newline-eof -Wno-missing-variable-declarations -Wno-weak-template-vtables -Wno-missing-prototypes -Wno-float-equal -Wno-return-std-move-in-c++11 -Wno-unreachable-code-break -Wno-undefined-func-template -Wno-unknown-warning-option -Wno-pass-failed -Wno-ctad-maybe-unsupported -Wno-header-hygiene -Wno-poison-system-directories -Wno-zero-as-null-pointer-constant -Wno-unsafe-buffer-usage)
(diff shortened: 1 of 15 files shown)


src/CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ if (NOT "${HYRISE_RELAXED_BUILD}")
...
+        add_compile_options(-Weverything -Wshadow-all -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-documentation -Wno-padded -Wno-global-constructors -Wno-sign-conversion -Wno-exit-time-destructors -Wno-switch-enum -Wno-weak-vtables -Wno-double-promotion -Wno-covered-switch-default -Wno-unused-macros -Wno-newline-eof -Wno-missing-variable-declarations -Wno-weak-template-vtables -Wno-missing-prototypes -Wno-float-equal -Wno-return-std-move-in-c++11 -Wno-unreachable-code-break -Wno-undefined-func-template -Wno-unknown-warning-option -Wno-pass-failed -Wno-ctad-maybe-unsupported -Wno-header-hygiene -Wno-poison-system-directories -Wno-zero-as-null-pointer-constant -Wno-unsafe-buffer-usage)
```
Bouncner, line 57: I guess we can call `add_compile_options()` multiple times, right?
If so, we should start to use multiple calls and add a comment what required this exception.

src/lib/storage/lz4_segment.cpp
```diff
@@ -428,7 +428,8 @@ template <typename T>
 std::shared_ptr<AbstractSegment> LZ4Segment<T>::copy_using_allocator(const PolymorphicAllocator<size_t>& alloc) const {
   auto new_lz4_blocks = pmr_vector<pmr_vector<char>>{alloc};
   for (const auto& block : _lz4_blocks) {
-    new_lz4_blocks.emplace_back(pmr_vector<char>{block, alloc});
+    auto block_copy = pmr_vector<char>{block, alloc};
```
Bouncner, line 431: Just curious, why is that needed?

src/CMakeLists.txt
dey4ss: I think so, good point. No ideas for the previous in-/exclusions though

src/lib/storage/lz4_segment.cpp
dey4ss: Tidy complains about the unnecessary constructor call, but without it, the compiler does not now which constructor to use. Construct + move seemed to be the thing that comes close to what's intended here

src/CMakeLists.txt
dey4ss: Added a short comment

Bouncner (approved):

PR #2584, opened 2023-06-25
Allow Apple Clang

Bouncner:
This PR enables Apple's clang version as it has sufficiently caught up with LLVM. An additional LLVM installation on MacOS is quite large. If our students don't need it, we shouldn't require it. Benchmarking won't be done on macOS machines anyways.

Changes:
- adapt Jenkinsfile to no longer fail on Apple clang compilers but issue a warning mentioning "the recommended clang"
- adapt CI pipeline to also compile with the Apple compiler versions


dey4ss:
😱

Bouncner:
> 😱

I am an Apple fanboy. ¯\_(ツ)_/¯

Next step is getting rid of Linux.

Bouncner:
Requesting a review from you @dey4ss, even though the pipeline is not mergeable right now. When we agree on the PR, I will re-add the deleted stages.

dey4ss (approved):

PR #2589, opened 2023-07-01
Jenkins - Retry Mac x86 CI Runs

Bouncner:
The mac x86 CI node shows a pretty low network stability. As if it executed in the last CI group, we regularly fail jobs perfectly fine jobs after 3h or more.

This change simply tells Jenkins to repeat the step several times before failing. As long as we do not get the network under control, I am afraid that this is the best measure at the moment.


dey4ss:
❤️

dey4ss (approved):

PR #2591, opened 2023-07-11
[DYOD] Maintain UCCs in face of potentially changing data

j-hellenberg:
`UCCs` (Unique Column Combinations) indicate the uniqueness across all entries of a column (or a set of columns). UCCs can be given by the schema (a primary key guarantees uniqueness), but also happen "incidentally" in real-world data.

Because UCCs can be used for query optimization, we want to detect these "incidental" UCCs as well, because, for optimization purposes, we can use them just as we would use a primary key constraint.

Before this PR, hyrise was already capable to detect such UCCs. It did so by actually generating `UNIQUE` constraints on a table, which are translated into UCCs during query optimization. However, this discovery happened under the assumption that the data of the table never changes. This means, if the data of the table *were* to change and violate the previously discovered "incidental" UCC by creating a duplicate value in the targeted column, this using the UCC for query optimization would lead to incorrect query results.

Therefore, in this MR, we assign a validation `commit ID` to the TableKeyConstraint (UCC) such that it may be only used for optimization if the table has not seen any changes since this `commit ID` (note that every modifying transaction increments the global `commit ID`). We will then regularly revalidate the UCC and see if we can expand the constraint to larger commit IDs.

For optimization purposes, we also make use of the `MVCC` (Multi-Version Concurrency Control) data of chunks, which includes but is not limited to the `commit IDs` of when any data was last inserted to or deleted from the chunk (note that hyrise models modifications as delete+insert).

_Note that this PR currently targets the `dey4ss/set_cids` branch, because we depend on the changes Daniel made for us in there. Targeting this branch should make the review a bit easier since you won't have to care about his changes :) We will change the PR to target master before merging._

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,7 @@ tpch_cached_tables/
 imdb_data/
 jcch_data/
 ssb_data/
+
+# Python poetry
+poetry.lock
+pyproject.toml
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ Contact: firstname.lastname@hpi.de
 -   Lawrence  Benson
 -   Jasper    Blum
 -   Lukas     Budach
+-   Margarete Dippel
 -   Timo      Djürken
 -   Alexander Dubrawski
 -   Fabian    Dumke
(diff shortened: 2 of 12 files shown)


j-hellenberg:
Please note: This PR was replaced by #2599 (we needed to create the PR from a branch within the Hyrise repo so that we could create our two PRs on top of each other, but the content stayed practically unchanged)

PR #2594, opened 2023-07-17
Make CI runs more stable

dey4ss:
Recently, we encountered many CI issues with SSB tests that called the dbgen in in parallel. This PR moves execution of `hyriseBenchmarkStarSchema_test.py` to an own stage and removes SSB data generation from `hyriseConsole_test.py`. Furthermore, an issue with concurrent writes of `StoredtableNode`'s cached output expressions for cached PQPs has been resolved by @Bouncner.

### DIFF
--- a/scripts/test/hyriseConsole_test.py
+++ b/scripts/test/hyriseConsole_test.py
@@ -129,11 +129,11 @@ def main():
     console.sendline("select * from meta_tables")
     console.expect("0 rows total")
 
-    # Test SSB generation.
-    console.sendline("generate_ssb 0.01")
-    console.expect("Generating tables done")
+    # Test TPC-DS generation.
+    console.sendline("generate_tpcds 1")
+    console.expect("Generating tables done", timeout=300)
     console.sendline("select * from meta_tables")
-    console.expect("5 rows total")
+    console.expect("24 rows total")
 
     # Test meta table modification.
     console.sendline("insert into meta_settings values ('foo', 'bar', 'baz')")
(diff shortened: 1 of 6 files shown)


scripts/test/hyriseConsole_test.py
```diff
@@ -129,11 +129,11 @@ def main():
     console.sendline("select * from meta_tables")
     console.expect("0 rows total")
 
-    # Test SSB generation.
-    console.sendline("generate_ssb 0.01")
-    console.expect("Generating tables done")
+    # Test TPC-DS generation.
+    console.sendline("generate_tpcds 1")
+    console.expect("Generating tables done", timeout=300)
```
Bouncner, line 134: Do you have a rough estimate how long SF 1 generation takes?

Bouncner (approved):

scripts/test/hyriseConsole_test.py
dey4ss:
```
real 231.16
user 1443.67
sys 264.70
```

in docker on imres

Bouncner:
If there was no other load on the CI server, over 200 s for generation is likely to timeout under load.
Set to 600 s?

Bouncner:
Hmm, wondering if 200-300s is too much in general. What do you think?

dey4ss:
> Hmm, wondering if 200-300s is too much in general. What do you think?

As a timeout or for generation? Think a timeout of 5 min is a long time in general, but we don't set it for most commands, right?

Bouncner:
Zünd it.

PR #2598, opened 2023-07-26
Numa merged

rothaarlappen:


src/lib/utils/numa.cpp
```diff
@@ -0,0 +1,44 @@
...
+  return distance_matrix;
+}
+
+NodeMatrix sort_relative_node_ids(DistanceMatrix distance_matrix) {
+  int matrix_size = distance_matrix.size();
+  auto node_matrix = NodeMatrix(static_cast<size_t>(matrix_size), std::vector<NodeID>(matrix_size, NodeID{0}));
+
+  for (auto node_id = int{0}; node_id < matrix_size; ++node_id) {
+    std::iota(node_matrix[node_id].begin(), node_matrix[node_id].end(), 0);
+    std::sort(node_matrix[node_id].begin(), node_matrix[node_id].end(),
```
Bouncner, line 39: Short question: would all nodes on a second blade (let's say node 4,5,6,7) have the same order for the closest sockets? Might it make sense to "shuffle the order" each time to have different orders (only for equal distances obviously) for the nodes?

Tratori: Just added some example output for our 8 Node system.
I guess the local blade already looks somewhat shuffled,
the distant blade might benefit from shuffling the orders, would have to test both, though.

src/lib/storage/dictionary_segment.cpp
```diff
@@ -1,5 +1,6 @@
 #include "dictionary_segment.hpp"
 
+#include <numa.h>
```
Finn-HPI, line 3: This include looks unused, maybe remove it?

Finn-HPI (commented):
Nice project 😀, mainly just a few comments regarding code style. I noticed you currently only have two new test cases (in NumaMemoryResourceTest). Maybe adding some additional unit tests would be good.

src/lib/memory/numa_memory_resource.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+#include <jemalloc/jemalloc.h>
+#include <unordered_map>
```
niklasmohrin, line 3: Nit: I think that the standard format would be
```cpp
#pragma once

#include <unordered_map>

#include <jemalloc/jemalloc.h>
```

I think by using cpplint, this applies to us: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

src/lib/memory/numa_memory_resource.hpp
```diff
@@ -0,0 +1,34 @@
...
+ public:
+  static void* alloc(extent_hooks_t* extent_hooks, void* new_addr, size_t size, size_t alignment, bool* zero,
+                     bool* commit, unsigned arena_index);
+  static bool dalloc(extent_hooks_t* extent_hooks, void* addr, size_t size, bool committed, unsigned arena_ind);
+
+  static void store_node_id_for_arena(ArenaID, NodeID);
+
+ private:
+  static std::unordered_map<ArenaID, NodeID> node_id_for_arena_id;
+};
```
niklasmohrin, line 19: I would refactor this as follows: Replace the class with something like
```cpp
namespace numa_extent_hooks {

void* alloc_hook(extent_hooks_t* extent_hooks, void* new_addr, size_t size, size_t alignment, bool* zero, bool* commit, unsigned arena_index);
bool dalloc_hook(extent_hooks_t* extent_hooks, void* addr, size_t size, bool committed, unsigned arena_ind);

NodeID get_node_id_of_arena(ArenaID);
void store_node_id_for_arena(ArenaID, NodeID);

};
```

where the `get_` and `store_` functions access the global variable `node_id_for_arena_id` only after taking a lock (currently, I think you have a data race in the `NumaMemoryResource` constructor when updating the map). Note that I removed any mention of `node_id_for_arena_id` from the header, because it is actually not needed (noone links to it) and you can just have it in your cpp file (in an anonymous namespace). If you limit access to this global map with the lock, I think it is a fair way to handle it like this (the jemalloc hooks api seems to kind of force you to do so)

niklasmohrin: Note that moving the variable into an anonymous namespace doesn't really hurt debugability, as you can still inspect them in gdb:
```plain
(gdb) info variables niklas
All variables matching regular expression "niklas":

File /home/niklas/Dev/hyrise/src/lib/memory/numa_memory_resource.cpp:
12:     static std::unordered_map<unsigned int, hyrise::NodeID, std::hash<unsigned int>, std::equal_to<unsigned int>, std::allocator<std::pair<unsigned int const, hyrise::NodeID> > > hyrise::(anonymous namespace)::niklas_cool_map;

...
(gdb) p 'hyrise::(anonymous namespace)::niklas_cool_map'
$1 = std::unordered_map with 0 elements
```

src/lib/memory/numa_memory_resource.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+class NumaMemoryResource : public boost::container::pmr::memory_resource {
+ public:
+  explicit NumaMemoryResource(const NodeID node_id);
+  void* do_allocate(std::size_t bytes, std::size_t alignment) override;
+  void do_deallocate(void* p, std::size_t bytes, std::size_t alignment) override;
+  bool do_is_equal(const memory_resource& other) const noexcept override;
+
+ protected:
+  NodeID _node_id{0};
```
niklasmohrin, line 29: Do you really want a default value?
```suggestion
  NodeID _node_id;
```

src/lib/memory/numa_memory_resource.cpp
```diff
@@ -0,0 +1,87 @@
...
+NumaMemoryResource::NumaMemoryResource(const NodeID node_id) : _node_id(node_id) {
+  // Setup arena.
+  auto arena_id = uint32_t{0};
+  auto size = sizeof(arena_id);
+  Assert(mallctl("arenas.create", static_cast<void*>(&arena_id), &size, nullptr, 0) == 0, "mallctl failed");
+
+  // Add extent hooks to arena.
+  _hooks.alloc = NumaExtentHooks::alloc;
+
+  auto hooks_ptr = &_hooks;
```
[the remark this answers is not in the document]
niklasmohrin: I don't really get why this variable exists in the first place, we are giving a pointer to the stack (`&hooks_ptr`) where there is a pointer to our member (`hooks_ptr`) to `mallctl`, but never actually read what this function wrote to our `hooks_ptr` variable. I don't get it

src/lib/memory/numa_memory_resource.cpp
```diff
@@ -0,0 +1,87 @@
...
+  char command[64];
+  snprintf(command, sizeof(command), "arena.%u.extent_hooks", arena_id);
+  NumaExtentHooks::store_node_id_for_arena(arena_id, _node_id);
+  Assert(mallctl(command, nullptr, nullptr, static_cast<void*>(&hooks_ptr), sizeof(extent_hooks_t*)) == 0,
```
niklasmohrin, line 69: If I understood correctly, this only works if jemalloc is used, so this could maybe use the renaming and explicitly call `je_mallctl` instead? (also applies to other places that use jemalloc stuff)

niklasmohrin (commented):
Comments on memory resource:

src/lib/scheduler/node_queue_scheduler.cpp
```diff
@@ -172,14 +180,17 @@ NodeID NodeQueueScheduler::determine_queue_id(const NodeID preferred_node_id) co
...
+  auto numa_aware = true;
+  for (auto task : tasks) {
+    if (task->node_id() > _queue_count - 1) {
+      numa_aware = false;
+    }
+  }
+  return numa_aware;
+}
 
+void NodeQueueScheduler::_group_default(const std::vector<std::shared_ptr<AbstractTask>>& tasks) const {
```
niklasmohrin, line 193: Maybe this should be called `group_round_robin`?

src/lib/scheduler/node_queue_scheduler.cpp
```diff
@@ -188,7 +199,6 @@ void NodeQueueScheduler::_group_tasks(const std::vector<std::shared_ptr<Abstract
     if (!task->predecessors().empty() || !task->successors().empty()) {
       return;
```
niklasmohrin, line 200: Side note: Should this be `continue` instead of `return`?

src/lib/scheduler/node_queue_scheduler.cpp
```diff
@@ -209,4 +219,41 @@ void NodeQueueScheduler::_group_tasks(const std::vector<std::shared_ptr<Abstract
   }
 }
 
+// with 240 groups.................
+void NodeQueueScheduler::_group_numa_aware(const std::vector<std::shared_ptr<AbstractTask>>& tasks) const {
+  auto round_robin_counter = std::vector<int>(_queue_count, 0);
+
+  // std::vector<std::shared_ptr<AbstractTask>> grouped_tasks(_workers_per_node * _queue_count);
+  std::vector<std::shared_ptr<AbstractTask>> grouped_tasks(_group_number_per_node * _queue_count);
```
niklasmohrin, line 227: What is a "group number"? Additionally, it should probably not be a member variable, but instead a constant

src/lib/scheduler/node_queue_scheduler.cpp
```diff
@@ -209,4 +219,41 @@ void NodeQueueScheduler::_group_tasks(const std::vector<std::shared_ptr<Abstract
...
+
+  for (const auto& task : tasks) {
+    if (!task->predecessors().empty() || !task->successors().empty()) {
+      return;
+    }
+    auto num_node = task->node_id();
```
niklasmohrin, line 233: Not sure, but ?
```suggestion
    auto numa_node = task->node_id();
```
(otherwise, maybe stick to `node_id`?)

src/lib/scheduler/node_queue_scheduler.cpp
```diff
@@ -209,4 +219,41 @@ void NodeQueueScheduler::_group_tasks(const std::vector<std::shared_ptr<Abstract
...
+    const auto group_id =
+        (_group_number_per_node * num_node) + (round_robin_counter[num_node] % _group_number_per_node);
```
niklasmohrin, line 235: If you make `_group_number_per_node` a compile time constant, you could change the type of `grouped_tasks` to `std::vector<std::array<..., _group_number_per_node>>` and access it as `grouped_tasks[numa_node][round_robin_counter[numa_node]]`

niklasmohrin (commented):
scheduler comments:

src/lib/storage/reference_segment.hpp
```diff
@@ -35,6 +35,7 @@ class ReferenceSegment : public AbstractSegment {
 
   std::shared_ptr<AbstractSegment> copy_using_allocator(const PolymorphicAllocator<size_t>& /*alloc*/) const override;
 
+  NodeID get_numa_node_location() override;
```
niklasmohrin, line 38: Kind of odd that `get` is specialized, but `set` is kept as is. In general, I was wondering if you really need the virtual getter and setter. Since the getter and setter are public, the member variable could just be public itself. The `ReferenceSegment` just sets the field at construction (and figures it out the same way you currently figure it out in this override). What do you think?

niklasmohrin (commented):
storage comments:

src/lib/operators/join_hash.cpp
```diff
@@ -435,13 +437,17 @@ class JoinHash::JoinHashImpl : public AbstractReadOnlyOperatorImpl {
      *    We use the probe side's Bloom filter to exclude values from the hash table that will not be accessed in the
      *    probe step.
      */
+
+    // We want to retrieve the hash tables node placements.
```
niklasmohrin, line 441:
```suggestion
```

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -252,19 +255,21 @@ using BloomFilter = boost::dynamic_bitset<>;
...
+// @param column_id               Column within that table to materialize
+// @param histograms              Out: If radix_bits > 0, contains one histogram per chunk where each histogram
+//                                contains 1 << radix_bits slots
+// @param radix_bits              Number of radix_bits, needed only for histogram calculation
+// @param output_bloom_filter     Out: A filled BloomFilter where `value & BLOOM_FILTER_MASK == true` for each value
+//                                encountered in the input column
+// @param partition_node_location Out: Contains the NUMA node location at position j of the returned partition at
+//                                position j.
+// @param input_bloom_filter      Optional: Materialization is skipped for each value where the corresponding slot
+//                                in the Bloom filter is false
```
niklasmohrin, line 268: Ah, I see, there are a handful of out-parameters already. Hm, then I guess you adding another is somewhat fair.

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -290,11 +295,15 @@ RadixContainer<T> materialize_input(const std::shared_ptr<const Table>& in_table
   // Create histograms per chunk
   histograms.resize(chunk_count);
 
+  auto radix_container_node_positions = std::vector<std::optional<NodeID>>(chunk_count, std::nullopt);
```
niklasmohrin, line 298: It seems that you set this to `UNKNOWN_NODE_ID` quite often (always, except for lines 440-441), could it be a better default value than `std::nullopt`? Do you even need optionals here, or could `merge_node_placements` be changed to not take `std::optional` either, but check for `UNKNOWN_NODE_ID` instead?

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -412,27 +422,49 @@ RadixContainer<T> materialize_input(const std::shared_ptr<const Table>& in_table
         output_bloom_filter |= local_output_bloom_filter;
       }
     };
+    const auto& segment = chunk_in->get_segment(column_id);
     if (JoinHash::JOB_SPAWN_THRESHOLD > num_rows) {
       materialize();
+      // try to retrieve current numa node
+      auto worker = Worker::get_this_thread_worker();
+      if (worker) {
```
niklasmohrin, line 430:
```suggestion
      if (const auto worker = Worker::get_this_thread_worker()) {
```

src/lib/operators/join_hash/join_hash_steps.hpp
```diff
@@ -412,27 +422,49 @@ RadixContainer<T> materialize_input(const std::shared_ptr<const Table>& in_table
...
 
 template <typename BuildColumnType, typename HashedType>
 std::vector<std::optional<PosHashTable<HashedType>>> build(const RadixContainer<BuildColumnType>& radix_container,
                                                            const JoinHashBuildMode mode, const size_t radix_bits,
-                                                           const BloomFilter& input_bloom_filter) {
+                                                           const BloomFilter& input_bloom_filter,
+                                                           std::vector<NodeID>& hash_table_node_locations) {
   Assert(input_bloom_filter.size() == BLOOM_FILTER_SIZE, "invalid input_bloom_filter");
-
+  DebugAssert(hash_table_node_locations.size() == 0, "hash_table_node_locations should be empty.");
```
niklasmohrin, line 467:
```suggestion
  DebugAssert(hash_table_node_locations.empty(), "hash_table_node_locations should be empty.");
```

niklasmohrin (commented):
hash join comments:

cambryx (commented):
Looks good!
Didn't have a lot to add to what niklas and finn already wrote.

src/lib/scheduler/node_queue_scheduler.cpp
Tratori: I guess we assume here, that if for some task, the predecessors or successors are set, the given tasks already contain some kind of grouping and we can preemptively exit (to not mess anything up/create deadlocks). 

I am not 100% sure though.

src/lib/utils/numa_helper.cpp
```diff
@@ -0,0 +1,74 @@
...
+  node_placements.reserve(total_size);
+  uint32_t task_index = 0;
+  for (auto& job : jobs) {
+    while (non_scheduled_placements[task_index].has_value()) {
+      node_placements.emplace_back(non_scheduled_placements[task_index].value());
+      task_index++;
+    }
+    node_placements.emplace_back(job->node_id());
+    task_index++;
+  }
```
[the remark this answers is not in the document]
Tratori: That code looks much better, thank you. Added a comment and a test. For the intended usage, the errors should not occur. Given the new code, in the worst case, the reserve is inaccurate, which should be fine.

src/lib/memory/numa_memory_resource.cpp
rrcomtech: That is an interesting input! I will test it shortly.

src/lib/memory/numa_memory_resource.cpp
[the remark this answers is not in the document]
rrcomtech: You are right. That is not too sensible.

src/test/lib/memory/numa_memory_resource_test.cpp
```diff
@@ -0,0 +1,49 @@
...
+namespace hyrise {
+
+class NumaMemoryResourceTest : public BaseTest {};
+
+TEST_F(NumaMemoryResourceTest, AllocateDeallocate) {
+  const auto num_numa_nodes = static_cast<NodeID>(numa_num_configured_nodes());
+  for (auto target_node_idx = NodeID{0}; target_node_idx < num_numa_nodes; ++target_node_idx) {
+    auto resource = NumaMemoryResource(target_node_idx);
+    const auto allocation_size = sizeof(uint64_t);
+    auto buffer = resource.do_allocate(allocation_size, allocation_size);
```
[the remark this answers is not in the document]
Tratori: What exactly do you mean? resource.do_allocate takes an alignment as a second parameter, which is set to allocation_size (sizeof(uint64_t)). Where do you think we should use alignof additionally?

src/test/lib/memory/numa_memory_resource_test.cpp
```diff
@@ -0,0 +1,49 @@
...
+    auto value = new (buffer) uint64_t{42};
+    *value = 50;
+    ASSERT_EQ(buffer, value);
+    ASSERT_EQ(*reinterpret_cast<uint64_t*>(buffer), 50);
```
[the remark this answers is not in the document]
Tratori: ISO C++ does not allow indirection on operand of type 'void *' [-Werror,-Wvoid-ptr-dereference]

I don't think there is an easy way around this. Or do you have any other ideas?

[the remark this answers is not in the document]
niklasmohrin: Oh I think I was confusing `value` and `buffer`, I suppose your current version captures the meaning better than comparing `*value`

src/test/lib/memory/numa_memory_resource_test.cpp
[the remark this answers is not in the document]
niklasmohrin: It takes a size and an alignment, so I would expect a use of both sizeof and alignof not only sizeof. I think they match for 64bit integers on all architectures that are tested but it still feels off

src/lib/scheduler/node_queue_scheduler.cpp
niklasmohrin: Not sure either, the comment appears to me as "skip this task" instead of "oh, seems like we are done with this set of tasks". It also appears that we could be stopping in the middle of the vector - if it is a "this vector is completely done or completely not-done" check, it would be clearer to check the first element before the loop and add a comment explaining why this is fair to do.

In the end this is not your change anyways, so no need to sweat it (although you did copy the pattern to your grouping :D)

PR #2599, opened 2023-08-02
[DYOD] Maintain UCCs in face of potentially changing data

j-hellenberg:
`UCCs` (Unique Column Combinations) indicate the uniqueness across all entries of a column (or a set of columns). UCCs can be given by the schema (a primary key guarantees uniqueness), but also happen "incidentally" in real-world data.

Because UCCs can be used for query optimization, we want to detect these "incidental" UCCs as well, because, for optimization purposes, we can use them just as we would use a primary key constraint.

Before this PR, hyrise was already capable to detect such UCCs. It did so by actually generating `UNIQUE` constraints on a table, which are translated into UCCs during query optimization. However, this discovery happened under the assumption that the data of the table never changes. This means, if the data of the table *were* to change and violate the previously discovered "incidental" UCC by creating a duplicate value in the targeted column, this using the UCC for query optimization would lead to incorrect query results.

Therefore, in this MR, we assign a validation `commit ID` to the TableKeyConstraint (UCC) such that it may be only used for optimization if the table has not seen any changes since this `commit ID` (note that every modifying transaction increments the global `commit ID`). We will then regularly revalidate the UCC and see if we can expand the constraint to larger commit IDs.

For optimization purposes, we also make use of the `MVCC` (Multi-Version Concurrency Control) data of chunks, which includes but is not limited to the `commit IDs` of when any data was last inserted to or deleted from the chunk (note that hyrise models modifications as delete+insert).


src/lib/storage/table.hpp
```diff
@@ -203,8 +203,15 @@ class Table : private Noncopyable {
...
    * We call them "soft" constraints to draw attention to that.
    */
-  void add_soft_key_constraint(const TableKeyConstraint& table_key_constraint);
   const TableKeyConstraints& soft_key_constraints() const;
+  void add_soft_key_constraint(const TableKeyConstraint& table_key_constraint);
+  void delete_key_constraint(const TableKeyConstraint& constraint);
+  /**
+   * Check if MVCC data tells us that the existing UCC is guaranteed to be still valid.
+   * To do this, we can simply check if the table chunks have seen any inserts/deletions since the last validation
+   * of the UCC. This is information is contained in the MVCC data of the chunks.
```
SanJSp, line 212:
```suggestion
   * of the UCC. This information is contained in the MVCC data of the chunks.
```

src/lib/storage/table.cpp
```diff
@@ -432,8 +436,28 @@ void Table::add_soft_key_constraint(const TableKeyConstraint& table_key_constrai
...
+void Table::delete_key_constraint(const TableKeyConstraint& constraint) {
+  _table_key_constraints.erase(constraint);
+}
+
+bool Table::constraint_guaranteed_to_be_valid(const TableKeyConstraint& table_key_constraint) const {
+  if (!table_key_constraint.can_become_invalid()) {
+    return true;
+  }
+
+  const auto chunk_count = this->chunk_count();
```
SanJSp, line 448: Can `this->` be removed?

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -106,38 +112,62 @@ void UccDiscoveryPlugin::_validate_ucc_candidates(const UccCandidates& ucc_candi
...
+      existing_ucc->revalidated_on(transaction_context->snapshot_commit_id());
       continue;
     }
 
+    // If no UCC already exists or the existing UCC is not guaranteed to be still valid, we have to now (re-)validate
+    // the UCC candidate.
     resolve_data_type(table->column_data_type(column_id), [&](const auto data_type_t) {
       using ColumnDataType = typename decltype(data_type_t)::type;
 
+      bool ucc_is_currently_valid;
```
SanJSp, line 138:
```suggestion
      auto ucc_is_currently_valid = false;
```

SanJSp: Makes it IMO more readable, but decide for yourselves :)

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -106,38 +112,62 @@ void UccDiscoveryPlugin::_validate_ucc_candidates(const UccCandidates& ucc_candi
...
       // Utilize efficient check for uniqueness inside each dictionary segment for a potential early out.
+      // If that does not allow us to reject the UCC right away, we have to run the more expensive
+      // cross-segment duplicate check (next clause of the if-condition).
```
SanJSp, line 141: Quick question regarding what is meant by `(next clause of the if-condition).`
Is it the next else-if clause or the next if block? Because if it's the latter I'd propose to keep this comment, that has been deleted `// If we reach here, we have to run the more expensive cross-segment duplicate check.`

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -180,57 +215,101 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+      if (distinct_values_across_segments.size() != expected_distinct_value_count) {
+        return false;
+      }
+
+      // If we managed to check this segment already, we don't need to do it again later using the validate operator.
+      unmodified_chunks.push_back(chunk_id);
     }
+  }
+
+  // Using the validate operator, get a view on the current content of the table, filtering out overwritten and deleted
```
SanJSp, line 272:
```suggestion
  // Using the validate operator, get a view of the current content of the table, filtering out overwritten and deleted
```

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -180,57 +215,101 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+  // values.
+  const auto logical_table = std::make_shared<GetTable>(table_name, unmodified_chunks, std::vector<ColumnID>());
+  logical_table->execute();
+  const auto validate_table_operator = std::make_shared<Validate>(logical_table);
+  validate_table_operator->set_transaction_context(transaction_context);
+  validate_table_operator->execute();
+  const auto& table_view = validate_table_operator->get_output();
+
+  // Check all chunks for duplicates that we haven't yet in the first loop above.
+  // Note that below loop will only contain these chunks, as we have excluded the others when executing the GetTable
```
SanJSp, line 282:
```suggestion
  // Note that the loop below will only contain these chunks, as we have excluded the others when executing the GetTable
```

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -73,6 +76,29 @@ class UccDiscoveryPluginTest : public BaseTest {
     ChunkEncoder::encode_all_chunks(table, chunk_encoding_spec);
   }
 
+  void _duplicate_table(std::string table_name) {
```
SanJSp, line 79:
```suggestion
  void _duplicate_table(const std::string table_name) {
```

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -73,6 +76,29 @@ class UccDiscoveryPluginTest : public BaseTest {
...
+    auto transaction_context = Hyrise::get().transaction_manager.new_transaction_context(AutoCommit::No);
+    const auto get_table = std::make_shared<GetTable>(table_name);
+    get_table->execute();
+    auto insert_op = std::make_shared<Insert>(table_name, get_table);
+    insert_op->set_transaction_context(transaction_context);
+    insert_op->execute();
+    transaction_context->commit();
+  }
+
+  void _delete_row(std::shared_ptr<Table> table, size_t rowId) {
```
SanJSp, line 89:
```suggestion
  void _delete_row(const std::shared_ptr<Table> table, const size_t rowId) {
```

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -73,6 +76,29 @@ class UccDiscoveryPluginTest : public BaseTest {
...
+    auto transaction_context = Hyrise::get().transaction_manager.new_transaction_context(AutoCommit::No);
+    const auto table_wrapper = std::make_shared<TableWrapper>(table);
+    table_wrapper->execute();
+    auto table_scan =
```
SanJSp, line 93: Can this also be `const`?

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -256,6 +282,65 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
 }
 
+TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidatesAfterDeletion) {
+  _encode_table(_table_A, GetParam());
+
+  // Delete row of _table_A that had a duplicate value regarding column 1 such that column 1 is unique afterwards.
+  _delete_row(_table_A, 3);
+
+  // We are only interested in column 1, since it was not unique before the deletion but should be now
+  auto ucc_candidates = UccCandidates{{"uniquenessTestTableA", ColumnID{1}}};
```
SanJSp, line 292:
```suggestion
  const auto ucc_candidates = UccCandidates{{"uniquenessTestTableA", ColumnID{1}}};
```
Is that possible?

SanJSp: If yes, same below

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -256,6 +282,65 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+
+TEST_P(UccDiscoveryPluginMultiEncodingTest, RevalidationUpdatesValidationTimestamp) {
+  _encode_table(_table_A, GetParam());
+
+  // We are only interested in column 1, since it was not unique before the deletion but should be now
+  auto ucc_candidates = UccCandidates{{"uniquenessTestTableA", ColumnID{0}}};
+
+  _validate_ucc_candidates(ucc_candidates);
+
+  // Perform transaction that does not affect table A but increments the global Commit ID
```
SanJSp, line 310:
```suggestion
  // Perform a transaction that does not affect table A but increments the global Commit ID
```

SanJSp (commented):
Hey all, I reviewed your code in the name of Team 3.
First of all: Good job with the implementation!


I checked mainly for hyrise coding standards, which were all pretty well followed. I already started to look inside your comments to find stuff I could pick on 😁  So I think this is a pretty good sign. Likewise, I've tried to add some comments regarding `const` in between, sometimes I was not too sure if it's applicable. Feel free to try it out and if it works implement it, if not, so be it 😊 

I will take a look at the other PR as well 😊

src/lib/storage/constraints/table_key_constraint.hpp
```diff
@@ -21,11 +21,22 @@ class TableKeyConstraint final : public AbstractTableConstraint {
    * voilating the set semantics of the constraint.
    */
   TableKeyConstraint(const std::set<ColumnID>& columns, const KeyConstraintType key_type);
+  TableKeyConstraint(const std::set<ColumnID>& columns, const KeyConstraintType key_type, CommitID last_validated_on);
```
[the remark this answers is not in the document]
kathrinthenhausen: Because of the datatype there is no reason for it to be const

src/lib/storage/constraints/table_key_constraint.cpp
```diff
@@ -5,14 +5,32 @@
...
-  return _columns;
+bool TableKeyConstraint::can_become_invalid() const {
+  return _last_validated_on != INVALID_COMMIT_ID;
+}
+
+CommitID TableKeyConstraint::last_validated_on() const {
+  return _last_validated_on;
+}
+
+void TableKeyConstraint::revalidated_on(CommitID revalidation_commit_id) const {
```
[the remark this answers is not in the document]
kathrinthenhausen: Same reason as above

src/lib/storage/table.cpp
Margarete01: No, if I try to remove `this->` I get the following error `Variable 'chunk_count' declared with deduced type 'const auto' cannot appear in its own initializer`.

src/plugins/ucc_discovery_plugin.cpp
Margarete01: The next else-if clause is meant. In the first clause of the if-condition, we try to cast each segment to a dictionary segment and check if the size of an unmutable dictionary segment is unequal to the unique values of the segment. If we haven't found a unmutable segment that fulfills this condition, there could still be mutable segments, that contain duplicates which we have to check in the next else-if clause.

src/lib/storage/constraints/table_key_constraint.hpp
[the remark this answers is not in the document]
Margarete01: I believe one could make last_validated_on `const, because we don't want to change the method argument value in the constructor (or setter) and const kind of just promises that we will not change the method argument value. The datatype shouldn't have an impact on making this variable const.

src/lib/storage/table.cpp
```diff
@@ -436,8 +440,28 @@ void Table::add_soft_key_constraint(const TableKeyConstraint& table_key_constrai
...
+  // Iterate through the chunks backwards as inserts are more likely to happen in later chunks, potentially enabling us
+  // to return faster.
+  // Subtract 1 from the chunk_id only inside the loop to avoid underflows.
```
dey4ss, line 455: Rather start with chunk_count - 1 then and loop to >= 0?

src/lib/storage/table.cpp
```diff
@@ -436,8 +440,28 @@ void Table::add_soft_key_constraint(const TableKeyConstraint& table_key_constrai
   _table_key_constraints.insert(table_key_constraint);
 }
 
-const TableKeyConstraints& Table::soft_key_constraints() const {
-  return _table_key_constraints;
+void Table::delete_key_constraint(const TableKeyConstraint& constraint) {
+  _table_key_constraints.erase(constraint);
+}
+
+bool Table::constraint_guaranteed_to_be_valid(const TableKeyConstraint& table_key_constraint) const {
```
dey4ss, line 447: I see why this is a member of `Table`, but I'd prefer this to be a helper function in a separate file taking the table and the constraint as arguments.

src/lib/storage/table.hpp
```diff
@@ -205,8 +205,15 @@ class Table : private Noncopyable {
...
-  void add_soft_key_constraint(const TableKeyConstraint& table_key_constraint);
   const TableKeyConstraints& soft_key_constraints() const;
+  void add_soft_key_constraint(const TableKeyConstraint& table_key_constraint);
+  void delete_key_constraint(const TableKeyConstraint& constraint);
+  /**
+   * Check if MVCC data tells us that the existing UCC is guaranteed to be still valid.
+   * To do this, we can simply check if the table chunks have seen any inserts/deletions since the last validation
+   * of the UCC. This information is contained in the MVCC data of the chunks.
+   */
+  bool constraint_guaranteed_to_be_valid(const TableKeyConstraint& table_key_constraint) const;
```
dey4ss, line 216: (see comment in cpp)

src/lib/storage/constraints/table_key_constraint.hpp
```diff
@@ -21,11 +21,22 @@ class TableKeyConstraint final : public AbstractTableConstraint {
    * voilating the set semantics of the constraint.
    */
   TableKeyConstraint(const std::set<ColumnID>& columns, const KeyConstraintType key_type);
+  TableKeyConstraint(const std::set<ColumnID>& columns, const KeyConstraintType key_type, const CommitID last_validated_on);
```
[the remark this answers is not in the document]
dey4ss: Also, is there a reason not to use `MAX_COMMIT_ID` (and maybe move it's definition to types.hpp?)

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -106,38 +112,60 @@ void UccDiscoveryPlugin::_validate_ucc_candidates(const UccCandidates& ucc_candi
...
 
-    // Skip already discovered UCCs.
-    if (std::any_of(soft_key_constraints.cbegin(), soft_key_constraints.cend(),
-                    [&column_id](const auto& key_constraint) {
-                      const auto& columns = key_constraint.columns();
-                      return columns.size() == 1 && *columns.cbegin() == column_id;
-                    })) {
-      message << " [skipped (already known) in " << candidate_timer.lap_formatted() << "]";
+    // Find UCC matching the candidate, if it already exists on the table.
+    const auto& existing_ucc = std::find_if(soft_key_constraints.cbegin(), soft_key_constraints.cend(),
```
dey4ss, line 116: not sure the reference buys you sth for the iterator you get (maybe hinders auto move)

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -106,38 +112,60 @@ void UccDiscoveryPlugin::_validate_ucc_candidates(const UccCandidates& ucc_candi
...
+                                              return columns.size() == 1 && *columns.cbegin() == column_id;
+                                            });
+
+    // Check if MVCC data tells us that the existing UCC is guaranteed to be still valid.
+    // If it is, we can skip the expensive revalidation of the UCC.
+    if (existing_ucc != soft_key_constraints.cend() && table->constraint_guaranteed_to_be_valid(*existing_ucc)) {
+      message << " [skipped (already known and guaranteed to be still valid) in " << candidate_timer.lap_formatted()
+              << "]";
       Hyrise::get().log_manager.add_message("UccDiscoveryPlugin", message.str(), LogLevel::Info);
+      existing_ucc->revalidated_on(transaction_context->snapshot_commit_id());
```
dey4ss, line 129: Wait a sec. You can end up setting the validation snapshot ID for a schema-provided constraint here. (i) This makes the constraint not schema-given anymore (the check for this is equality w/ special commit ID). (ii) Your `DebugAssert` in `TableKeyConstraint` (ensure new commit > last validated commit) should throw if you do this.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -162,6 +190,11 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
       continue;
     }
 
+    if (source_chunk->invalid_row_count() != 0 || source_chunk->is_mutable()) {
+      // The segment might have been modified. Because the modification might consist of deleting a duplicate value,
+      // we can't be sure that the segment still has duplicates using the heuristic employed below.
+      return false;
+    }
```
dey4ss, line 197: Mutable segments are always unencoded. Thus, the second condition is not necessary here (code will return false anyway).

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -180,57 +213,101 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+      }
+
+      // If we managed to check this segment already, we don't need to do it again later using the validate operator.
+      unmodified_chunks.push_back(chunk_id);
     }
+  }
+
+  // Using the validate operator, get a view of the current content of the table, filtering out overwritten and deleted
+  // values.
+  const auto logical_table = std::make_shared<GetTable>(table_name, unmodified_chunks, std::vector<ColumnID>());
```
dey4ss, line 272: get_table?

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -180,57 +213,101 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+  logical_table->execute();
+  const auto validate_table_operator = std::make_shared<Validate>(logical_table);
+  validate_table_operator->set_transaction_context(transaction_context);
+  validate_table_operator->execute();
+  const auto& table_view = validate_table_operator->get_output();
+
+  // Check all chunks for duplicates that we haven't yet in the first loop above.
+  // Note that the loop below will only contain these chunks, as we have excluded the others when executing the GetTable
+  // operator.
```
dey4ss, line 281: Bit weird formulation. Also, fill 120 chars.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -180,57 +213,101 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+  const auto validated_chunk_count = table_view->chunk_count();
+  for (auto chunk_id = ChunkID{0}; chunk_id < validated_chunk_count; ++chunk_id) {
+    // No need to do null checks here as we already did so above
```
dey4ss, line 284: Ah, you mean chunk nullptr and not NULL values. There's no need to check after the GetTable operator (it only forwards chunks with actual data).

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -73,6 +76,29 @@ class UccDiscoveryPluginTest : public BaseTest {
     ChunkEncoder::encode_all_chunks(table, chunk_encoding_spec);
   }
 
+  void _duplicate_table(const std::string table_name) {
```
dey4ss, line 79: take a reference? Not measurable in this test case though.

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -256,6 +282,65 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
   EXPECT_FALSE(constraints_B.contains({{ColumnID{1}}, KeyConstraintType::UNIQUE}));
 }
 
+TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidatesAfterDeletion) {
+  _encode_table(_table_A, GetParam());
+
+  // Delete row of _table_A that had a duplicate value regarding column 1 such that column 1 is unique afterwards.
+  _delete_row(_table_A, 3);
+
+  // We are only interested in column 1, since it was not unique before the deletion but should be now
```
dey4ss, line 291: pick here: end w/ full stop.

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -256,6 +282,65 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+  EXPECT_TRUE(constraints_A.contains({{ColumnID{0}}, KeyConstraintType::UNIQUE}));
+
+  // Insert table data into the table again -> creates duplicate for every row
+  _duplicate_table(_table_name_A);
+
+  _validate_ucc_candidates(ucc_candidates);
+  const auto& constraints_B = _table_A->soft_key_constraints();
+  EXPECT_FALSE(constraints_B.contains({{ColumnID{0}}, KeyConstraintType::UNIQUE}));
+}
+
```
dey4ss, line 343: Pick here: I'd love to see some more test cases, e.g., no CommitID ipdate for schema-given UCC, Updates (since it's delete + insert w/ and w/o duplicate), validation while transactions are active (insert/delete ops are executed, but not committed), ...

dey4ss (commented):
I have a few suggestions, but nothing deal-breaking. Nice :)

dey4ss:
Put this on hold until end of August. We plan to incorporate this feature, but there is currently no capacity for it. Will work on it when back from the internship.

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -1,3 +1,5 @@
+#include <gtest/gtest.h>
```
dey4ss, line 1: This is a false positive, third party heades should not be above STL headers. Usually, we don't care about IWYU in tests and do not include the gtest header.

@Bouncner what's your take here? Just remove the include or add it to the third-party libraries in the include order?

dey4ss (commented):
Have not looked at all tests yet. However, I already have some concers when it comes to concurrency.

dey4ss:
Just aborted the CI pipeline because `StressTest.AddRemoveModifyTableKeyConstraintsConcurrently` kept the Mac stages running for days.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -130,29 +133,54 @@ void UccDiscoveryPlugin::_validate_ucc_candidates(const UccCandidates& ucc_candi
...
+    const auto existing_ucc = soft_key_constraints.find(TableKeyConstraint{{column_id}, KeyConstraintType::UNIQUE});
+    if (existing_ucc != soft_key_constraints.end() && is_constraint_confidently_valid(table, *existing_ucc)) {
+      message << " [skipped (already known and guaranteed to be still VALID) in " << candidate_timer.lap_formatted()
+              << "]";
+      Hyrise::get().log_manager.add_message("UccDiscoveryPlugin", message.str(), LogLevel::Info);
+      continue;
+    }
+
+    if (existing_ucc != soft_key_constraints.end() && is_constraint_confidently_invalid(table, *existing_ucc)) {
+      message << " [skipped (already known and guaranteed to be INVALID) in " << candidate_timer.lap_formatted() << "]";
```
Rob2U, line 155: If a candidate:
- does not contain a primary key
- is neither known to be valid nor known to be invalid
We try to revalidate this candidate.

src/lib/logical_query_plan/static_table_node.cpp
```diff
@@ -100,8 +107,36 @@ std::shared_ptr<AbstractLQPNode> StaticTableNode::_on_shallow_copy(LQPNodeMappin
...
-  return table->column_definitions() == static_table_node.table->column_definitions() &&
-         table->soft_key_constraints() == static_table_node.table->soft_key_constraints();
+
+  if (table->column_definitions() != static_table_node.table->column_definitions()) {
+    return false;
+  }
+
+  // If the column definitions match, we check if the valid key constraints are also equal then we assume equality.
+  const auto& key_constraints = table->soft_key_constraints();
+  const auto& rhs_key_constraints = static_table_node.table->soft_key_constraints();
```
dey4ss: Can we use just `==` for the key constraints? `StaticTableNode`s are in essence used for meta tables and for create table statements. In both cases, there sould not be invalidated constraints.

src/lib/storage/constraints/table_key_constraint.hpp
```diff
@@ -51,6 +51,8 @@ class TableKeyConstraint final : public AbstractTableConstraint {
 
   /**
    * Returns whether or not the last validation and invalidation `CommitID`s tell us that this constraints is valid.
+   * Be aware that this does not mean that the constraint is valid in the current transaction, only that it was 
+   * validated last. To check if a constraint is guaranteed to be valid use `key_constraint_is_confidently_valid`.
```
dey4ss: If I read that now, the name `is_valid()` pretty much should return true if sth is safely valid. The meaning here is that it was not invalidated after it was validated. Do you have any idea for a name describing that better?

src/lib/storage/table.hpp
```diff
@@ -216,6 +216,10 @@ class Table : private Noncopyable {
    */
   void add_soft_constraint(const AbstractTableConstraint& table_constraint);
 
+  /**
+   * NOTE: All key constraints are currently stored. If a constraint invalidated it is not deleted. To check if a key
```
dey4ss: `If a constraint _is_ invalidated, it is not deleted.`

src/lib/utils/print_utils.cpp
```diff
@@ -115,8 +115,15 @@ template void print_directed_acyclic_graph<const AbstractOperator>(
...
-      std::set<TableKeyConstraint>{table->soft_key_constraints().cbegin(), table->soft_key_constraints().cend()};
+  auto table_key_constraints = std::set<TableKeyConstraint>{};
+  // We only print key constraints that are marked as valid. Note that this has a different meaning
+  // than "confidently valid". See TableKeyConstraint::is_valid() for details.
+  for (const auto& constraint : table->soft_key_constraints()) {
+    if (!constraint.is_valid()) {
+      continue;
+    }
+    table_key_constraints.emplace(constraint);
+  }
```
dey4ss: Could we still use the constructor with iterators and check for invalidity in the loop below?

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -133,24 +134,50 @@ void UccDiscoveryPlugin::_validate_ucc_candidates(const UccCandidates& ucc_candi
...
 
-    // Skip already discovered UCCs.
-    if (std::any_of(soft_key_constraints.cbegin(), soft_key_constraints.cend(),
-                    [&column_id](const auto& key_constraint) {
-                      const auto& columns = key_constraint.columns();
-                      return columns.size() == 1 && *columns.cbegin() == column_id;
-                    })) {
-      message << " [skipped (already known) in " << candidate_timer.lap_formatted() << "]";
+    // Check if a primary key constraint already exists for the column. If it does, we can skip the candidate.
+    const auto existing_pk = soft_key_constraints.find(TableKeyConstraint{{column_id}, KeyConstraintType::PRIMARY_KEY});
```
dey4ss: We do not expect a PK to be schema-given at the moment. Also, we miss on existing, minimal UCCs when searching for exact matches (e.g., UCC `a, b` can be skipped if `a` is already known). Thus, I guess it would be better to iterate the table constraints and look for containment of the column IDs.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -133,24 +134,50 @@ void UccDiscoveryPlugin::_validate_ucc_candidates(const UccCandidates& ucc_candi
...
       continue;
     }
 
+    // Check if MVCC data tells us that the existing UCC is guaranteed to be still valid. If it is, we can skip the
+    // expensive revalidation of the UCC. This also applies for permanent constraints.
+    const auto existing_ucc = soft_key_constraints.find(TableKeyConstraint{{column_id}, KeyConstraintType::UNIQUE});
+    if (existing_ucc != soft_key_constraints.end() && key_constraint_is_confidently_valid(table, *existing_ucc)) {
+      message << " [skipped (already known and guaranteed to be still VALID) in " << candidate_timer.lap_formatted()
+              << "]";
+      Hyrise::get().log_manager.add_message("UccDiscoveryPlugin", message.str(), LogLevel::Info);
```
dey4ss: Would we still want to update the validation CID?

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -133,24 +134,50 @@ void UccDiscoveryPlugin::_validate_ucc_candidates(const UccCandidates& ucc_candi
...
+      continue;
+    }
+
+    if (existing_ucc != soft_key_constraints.end() && key_constraint_is_confidently_invalid(table, *existing_ucc)) {
+      message << " [skipped (already known and guaranteed to be INVALID) in " << candidate_timer.lap_formatted() << "]";
+      Hyrise::get().log_manager.add_message("UccDiscoveryPlugin", message.str(), LogLevel::Info);
+      continue;
```
dey4ss: (same as above)

dey4ss (commented):
A few things I'd like to discuss before continuing with the tests

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -74,6 +86,50 @@ class UccDiscoveryPluginTest : public BaseTest {
     ChunkEncoder::encode_all_chunks(table, chunk_encoding_spec);
   }
 
+  std::shared_ptr<TransactionContext> _insert_row(const std::string& table_name,
```
dey4ss: We do not access members of the test class, do we? I guess we can declare this in anonymous namespace then (same for `_encode_table` and `delete_row`).

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -74,6 +86,50 @@ class UccDiscoveryPluginTest : public BaseTest {
...
+    const auto table_wrapper = std::make_shared<TableWrapper>(table);
+    table_wrapper->execute();
+
+    const auto table_scan =
+        create_table_scan(table_wrapper, ColumnID{0}, PredicateCondition::Equals, table->get_row(row_index).at(0));
+    table_scan->execute();
+
+    const auto delete_op = std::make_shared<Delete>(table_scan);
+    delete_op->set_transaction_context(transaction_context);
+    delete_op->execute();
```
dey4ss: If you Like to, you can also use `execute_all({table_wrapper, table_scan, delete_op});`

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +302,224 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+  EXPECT_TRUE(constraints_B.find(TableKeyConstraint{{ColumnID{0}}, KeyConstraintType::UNIQUE})->is_valid());
+
+  EXPECT_TRUE(constraints_B.contains(TableKeyConstraint{{ColumnID{1}}, KeyConstraintType::UNIQUE}));
+  EXPECT_FALSE(constraints_B.find(TableKeyConstraint{{ColumnID{1}}, KeyConstraintType::UNIQUE})->is_valid());
+}
+
+TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidatesAfterDeletion) {
+  _encode_table(_table_A, GetParam());
+
+  // Delete row of _table_A that had a duplicate value regarding column 1 such that column 1 is unique afterwards.
```
dey4ss:
```suggestion
  // Delete the row of `_table_A` that had a duplicate value in column 1. Column 1 is unique afterwards.
```
?

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +302,224 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+  _validate_ucc_candidates(ucc_candidates);
+
+  // Collect constraints known for the tables.
+  const auto& constraints_A = _table_A->soft_key_constraints();
+  EXPECT_EQ(constraints_A.size(), 1);
+  EXPECT_TRUE(constraints_A.contains({{ColumnID{1}}, KeyConstraintType::UNIQUE}));
+  EXPECT_TRUE(constraints_A.find(TableKeyConstraint{{ColumnID{1}}, KeyConstraintType::UNIQUE})->is_valid());
+
+  // Delete another row. This should have no effect on the previously discovered UCC.
+  auto transaction_context = _delete_row(_table_A, 2, false);
```
dey4ss: `const`?

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +302,224 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+
+  const auto& constraints_A_2 = _table_A->soft_key_constraints();
```
dey4ss: `soft_key_constraints()` returns a reference to the actual set and, thus, is being updated if something changes there. We could just obtain the constraints once, right? To make sure we don't just operate on a static snapshot, we could obtain them before deleting the row and assert that there are no constraints.

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +302,224 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+
+  // Delete row of _table_A that had a duplicate value regarding column 1 such that column 1 is unique afterwards.
+  _delete_row(_table_A, 3);
+
+  // We are only interested in column 1, since it was not unique before the deletion but should be now.
+  const auto ucc_candidates = UccCandidates{{"uniquenessTestTableA", ColumnID{1}}};
+
+  _validate_ucc_candidates(ucc_candidates);
+
+  // Collect constraints known for the tables.
```
dey4ss:
```suggestion
  // Collect constraints known for the table.
```

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +302,224 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+
+  transaction_context->commit();
+
+  const auto& constraints_A_3 = _table_A->soft_key_constraints();
+  EXPECT_EQ(constraints_A_3.size(), 1);
+  EXPECT_TRUE(constraints_A_3.contains({{ColumnID{1}}, KeyConstraintType::UNIQUE}));
+  EXPECT_TRUE(constraints_A_3.find(TableKeyConstraint{{ColumnID{1}}, KeyConstraintType::UNIQUE})->is_valid());
+}
+
+TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidatesAfterInsertion) {
```
dey4ss: The comments from above also apply here.

This test is pretty much the same code as before with the only difference that we insert instead of delete, right? Can we just merge the tests? E.g., put it in a loop `for (const auto insert : {true, false}) { ... }`?

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +302,224 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+  // Validate for a second time after the commit.
+  _validate_ucc_candidates(ucc_candidates);
+  const auto& constraints_A_3 = _table_A->soft_key_constraints();
+
+  EXPECT_EQ(constraints_A_3.size(), 1);
+  EXPECT_TRUE(constraints_A_3.contains({{ColumnID{1}}, KeyConstraintType::UNIQUE}));
+  EXPECT_TRUE(constraints_A_3.find(TableKeyConstraint{{ColumnID{1}}, KeyConstraintType::UNIQUE})->is_valid());
+}
+
+TEST_P(UccDiscoveryPluginMultiEncodingTest, InvalidateCandidatesAfterDuplicateInsertion) {
```
dey4ss, line 339: (see comments above regarding fetching of `soft_key_constraints`)

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +302,224 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+  // Validate for a second time. This time the UCC should be invalid.
+  _validate_ucc_candidates(ucc_candidates);
+  const auto& constraints_A_3 = _table_A->soft_key_constraints();
+
+  EXPECT_EQ(constraints_A_3.size(), 1);
+  EXPECT_TRUE(constraints_A_3.contains({{ColumnID{1}}, KeyConstraintType::UNIQUE}));
+  EXPECT_FALSE(constraints_A_3.find(TableKeyConstraint{{ColumnID{1}}, KeyConstraintType::UNIQUE})->is_valid());
+}
+
+TEST_P(UccDiscoveryPluginMultiEncodingTest, InvalidateCandidatesAfterUpdate) {
```
dey4ss, line 383: This is basically a superset of the first two tests, right? If so, we could as well just keep this test. Fetching of constraits as above.

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +302,224 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+  // Validate for a second time. This time the UCC should be invalid.
+  _validate_ucc_candidates(ucc_candidates);
+  const auto& constraints_A_3 = _table_A->soft_key_constraints();
+
+  EXPECT_EQ(constraints_A_3.size(), 1);
+  EXPECT_TRUE(constraints_A_3.contains({{ColumnID{1}}, KeyConstraintType::UNIQUE}));
+  EXPECT_FALSE(constraints_A_3.find(TableKeyConstraint{{ColumnID{1}}, KeyConstraintType::UNIQUE})->is_valid());
+}
+
+TEST_P(UccDiscoveryPluginMultiEncodingTest, RevalidationUpdatesValidationTimestamp) {
```
dey4ss, line 428: (fetching of constraints as above)

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +542,103 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
   const auto& constraints_B = _table_B->soft_key_constraints();
 
   EXPECT_EQ(constraints_A.size(), 2);
-  EXPECT_EQ(constraints_B.size(), 1);
+  EXPECT_EQ(constraints_B.size(), 2);
 
   EXPECT_TRUE(constraints_A.contains({{_join_columnA->original_column_id}, KeyConstraintType::UNIQUE}));
+  EXPECT_TRUE(constraints_A.find(TableKeyConstraint{{_join_columnA->original_column_id}, KeyConstraintType::UNIQUE})
+                  ->is_valid());
```
dey4ss: As this already gets quite long: Should we just call `find` once and store the iterator in a variable? Then, we can `ASSERT_NE(..., constraints.end());` and `EXPECT_TRUE(...->is_valid());` is a lot shorter. Note the use of `ASSERT_` because there is the risk of a segfault if the constraint is not contained.

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +542,103 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
...
   EXPECT_EQ(Hyrise::get().default_lqp_cache->size(), 0);
   EXPECT_EQ(Hyrise::get().default_pqp_cache->size(), 0);
 }
 
+TEST_F(UccDiscoveryPluginMultiEncodingTest, PluginIntegrationTestWithInvalidation) {
+  Hyrise::get().default_pqp_cache = std::make_shared<SQLPhysicalPlanCache>();
+  Hyrise::get().default_lqp_cache = std::make_shared<SQLLogicalPlanCache>();
+
+  const auto exec_groupby_statement = [&]() {
+    const std::string sql = "SELECT b,c FROM " + _table_name_A + " GROUP BY b,c;";
```
dey4ss: auto to stick

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +542,103 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
...
+    auto pipeline = SQLPipelineBuilder{sql}.create_pipeline();
+    const auto& [pipeline_status, _] = pipeline.get_result_table();
+    ASSERT_EQ(pipeline_status, SQLPipelineStatus::Success);
+  };
+
+  const auto exec_insert_statement = [&]() {
+    auto transaction_context = Hyrise::get().transaction_manager.new_transaction_context(AutoCommit::No);
```
dey4ss, line 531: const?

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +542,103 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
...
+    const std::string sql = "INSERT INTO " + _table_name_A + " (A_a, b, c) VALUES (14, 12, 'duplicate_b');";
```
dey4ss: auto

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +542,103 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
...
+  // Discover UCCs used in previous query.
+  auto& plugin_manager = Hyrise::get().plugin_manager;
+  plugin_manager.load_plugin(build_dylib_path("libhyriseUccDiscoveryPlugin"));
+  plugin_manager.exec_user_function("hyriseUccDiscoveryPlugin", "DiscoverUCCs");
+
+  // Check that key constraint was added to the table.
+  const auto constraint_A_c = TableKeyConstraint{{ColumnID{2}}, KeyConstraintType::UNIQUE};
+
+  ASSERT_EQ(_table_A->soft_key_constraints().size(), 2);
+  ASSERT_TRUE(_table_A->soft_key_constraints().find(constraint_A_c)->is_valid());
```
dey4ss: `ASSERT` that the constraint is set and `EXPECT` that it is valid?

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +542,103 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
...
+
+  exec_groupby_statement();
+  exec_insert_statement();
+  exec_insert_statement();
```
dey4ss: Can you add a comment what is done here so constraints are not valid anymore?

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +542,103 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
...
+
+  // Check that validity of constraint is not guaranteed anymore.
+  const auto constraint_A = _table_A->soft_key_constraints().find(constraint_A_c);
+  ASSERT_NE(constraint_A, _table_A->soft_key_constraints().end());
+  EXPECT_FALSE(key_constraint_is_confidently_valid(_table_A, *constraint_A));
```
dey4ss, line 567: That's what I meant with my comment above :)

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +542,103 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
...
+  // Flush and repopulate the cache.
+  Hyrise::get().default_pqp_cache->clear();
+  Hyrise::get().default_lqp_cache->clear();
+  exec_groupby_statement();
+
+  // Check that key constraint is properly invalidated.
+  plugin_manager.exec_user_function("hyriseUccDiscoveryPlugin", "DiscoverUCCs");
+
+  ASSERT_EQ(_table_A->soft_key_constraints().size(), 2);
+  ASSERT_FALSE(_table_A->soft_key_constraints().find(constraint_A_c)->is_valid());
```
dey4ss: please adapt here as well

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +542,103 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
...
+
+  // Remove duplicates.
+  {
+    const std::string sql = "DELETE FROM " + _table_name_A + " WHERE b > 10;";
```
dey4ss: auto?

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +542,103 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
...
+    auto pipeline = SQLPipelineBuilder{sql}.create_pipeline();
```
dey4ss: If you like to, you can also call
```c++
auto pipeline = SQLPipelineBuilder{"DELETE FROM " + _table_name_A + " WHERE b > 10;"}.create_pipeline();
```
directly. Or even combine to:
```c++
SQLPipelineBuilder{"DELETE FROM " + _table_name_A + " WHERE b > 10;"}.create_pipeline().get_result_table();
```

dey4ss (commented):
some comments on simplifying the test code

src/test/plugins/ucc_discovery_plugin_test.cpp
Rob2U: We need to call `commit()` which is not a const member function. So transaction_context cannot be const.

src/lib/utils/print_utils.cpp
```diff
@@ -115,14 +115,25 @@ template void print_directed_acyclic_graph<const AbstractOperator>(
 
 void print_table_key_constraints(const std::shared_ptr<const Table>& table, std::ostream& stream,
                                  const std::string& separator) {
-  const auto& table_key_constraints =
-      std::set<TableKeyConstraint>{table->soft_key_constraints().cbegin(), table->soft_key_constraints().cend()};
+  auto table_key_constraints =
+      std::set<TableKeyConstraint>{table->soft_key_constraints().begin(), table->soft_key_constraints().end()};
+  // We only print key constraints that are marked as valid. Note that this has a different meaning
+  // than "confidently valid". See TableKeyConstraint::was_valid_last() for details.
```
dey4ss: The method was renamed, right?

src/lib/utils/print_utils.cpp
```diff
@@ -115,14 +115,25 @@ template void print_directed_acyclic_graph<const AbstractOperator>(
...
   if (table_key_constraints.empty()) {
     return;
   }
 
-  const auto& last_constraint = *std::prev(table_key_constraints.cend());
+  bool first = true;
```
dey4ss: auto

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -205,57 +250,102 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+      unmodified_chunks.push_back(chunk_id);
       continue;
     }
     const auto source_segment = source_chunk->get_segment(column_id);
-    if (!source_segment) {
-      continue;
+    // Was the chunk modified since the last UCC validation (inserted or updated rows)? If not, than we do not need to
+    // probe its values for duplicates later, because duplicates can only be introduced by chunks that were modified
+    // after the last validation.
+    const bool not_modified_since_last_validation =
```
dey4ss: auto?

dey4ss: We do check here that (i) a UCC is already known and (ii) there were no insertions since its last validation (deletes are not checked). The name does not really tell that. I don't have any good idea for now.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -205,57 +250,102 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+        existing_ucc && existing_ucc->last_validation_result() == ValidationResultType::VALID &&
+        source_chunk->mvcc_data()->max_begin_cid.load() <= existing_ucc->last_validated_on();
+    if ((source_chunk->invalid_row_count() == 0 && !source_chunk->is_mutable()) || not_modified_since_last_validation) {
```
dey4ss: We iterate the segment with `segment_with_iterators` below. With `not_modified_since_last_validation` as a disjunction, we can happen to iterate a ValueSegment where values are inserted while we iterate it (`is_mutable()` is true). That feels dangerous. Usually, the Validate operator takes care about that by using a PosList that marks the final position we should be able to see:
https://github.com/hyrise/hyrise/blob/cb35cf306059ec4a498b1a0c2c2f40b4f1af4fa1/src/lib/operators/validate.cpp#L271-L274

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -205,57 +250,102 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
 }
 
 template <typename ColumnDataType>
-bool UccDiscoveryPlugin::_uniqueness_holds_across_segments(const std::shared_ptr<const Table>& table,
-                                                           const ColumnID column_id) {
+bool UccDiscoveryPlugin::_uniqueness_holds_across_segments(
+    const std::shared_ptr<const Table>& table, const std::string& table_name, const ColumnID column_id,
+    const std::shared_ptr<TransactionContext>& transaction_context,
+    const std::optional<TableKeyConstraint>& existing_ucc) {
```
dey4ss: The optional is never set, right?

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -205,57 +250,102 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+  // `distinct_values_across_segments` collects the segment values from all chunks.
+  auto distinct_values_across_segments = std::unordered_set<ColumnDataType>{};
+  auto unmodified_chunks = std::vector<ChunkID>();
```
dey4ss:
```suggestion
  auto unmodified_chunks = std::vector<ChunkID>{};
```

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -205,57 +250,102 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+        // Directly insert dictionary entries.
+        const auto& dictionary = dictionary_segment->dictionary();
+        distinct_values_across_segments.insert(dictionary->cbegin(), dictionary->cend());
+      } else {
+        segment_with_iterators<ColumnDataType>(*source_segment, [&](auto it, const auto end) {
+          while (it != end) {
+            distinct_values_across_segments.insert(it->value());
+            ++it;
+          }
+        });
```
dey4ss: If we do not break in the loop, we can also use `segment_iterate` here. With my comment above, it might be necessary to create a ReferenceSegment if the chunk is mutable.

dey4ss: `it->value()` could still throw if there is a NULL value, right? So we might want to check `it->is_null()` and break if that's true like below. That's pretty much the same code then. Can we pull this iteration + check into an own lambda / helper method?

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -18,15 +22,84 @@
...
+  table_wrapper->execute();
+
+  const auto table_scan =
+      create_table_scan(table_wrapper, ColumnID{0}, PredicateCondition::Equals, table->get_row(row_index).at(0));
+  table_scan->execute();
+
+  const auto delete_op = std::make_shared<Delete>(table_scan);
+  delete_op->set_transaction_context(transaction_context);
+  delete_op->execute();
+  // execute_all({table_wrapper, table_scan, delete_op});
```
dey4ss: ?

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -205,57 +250,102 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+      // If this chunk does not exist, we do not need to check it with the validate operator later either.
+      unmodified_chunks.push_back(chunk_id);
       continue;
     }
     const auto source_segment = source_chunk->get_segment(column_id);
-    if (!source_segment) {
-      continue;
+    // Was the chunk modified since the last UCC validation (inserted or updated rows)? If not, than we do not need to
+    // probe its values for duplicates later, because duplicates can only be introduced by chunks that were modified
+    // after the last validation.
```
dey4ss:
```suggestion
    // Were values inserted into the chunk since the last UCC validation (inserted or updated rows)? If not, we do not
    // need to probe its values for duplicates later: duplicates can only be introduced by insertions that happened
    // since the last validation.
```

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -205,57 +250,102 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+    const bool not_modified_since_last_validation =
+        existing_ucc && existing_ucc->last_validation_result() == ValidationResultType::VALID &&
+        source_chunk->mvcc_data()->max_begin_cid.load() <= existing_ucc->last_validated_on();
+    if ((source_chunk->invalid_row_count() == 0 && !source_chunk->is_mutable()) || not_modified_since_last_validation) {
+      // The set of distinct values across all segments should grow by the number of rows in the segment because,
+      // if it does not, it means some value must have been inserted twice -> duplicate detected.
+      // In this case, the UCC is violated.
```
dey4ss:
```suggestion
      // The set of distinct values across all segments should grow by the number of rows in the segment. Otherwise,
      // some value must have been inserted twice, and we detected a duplicate that violates the UCC.
```

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -132,25 +134,63 @@ void UccDiscoveryPlugin::_validate_ucc_candidates(const UccCandidates& ucc_candi
     message << "Checking candidate " << candidate.table_name << "." << table->column_name(column_id);
 
     const auto& soft_key_constraints = table->soft_key_constraints();
+    const auto candidate_columns = std::unordered_set<ColumnID>{column_id};
+    const auto existing_ucc =
+        std::find_if(soft_key_constraints.cbegin(), soft_key_constraints.cend(), [&](const auto& key_constraint) {
+          return std::includes(key_constraint.columns().cbegin(), key_constraint.columns().cend(),
+                               candidate_columns.cbegin(), candidate_columns.cend());
```
dey4ss:
```suggestion
          return std::ranges::includes(key_constraint.columns(), candidate_columns);
```

dey4ss: `TableKeyConstraint::columns()` returns a `std::set` reference, but `candidate_columns` is a `std::unordered_set`. If the candidate happens to have multiple columns in the future, we could run into false negatives if the order or columns differs, right? Should be fine if both are `std::set`s, or do I miss something?

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -205,57 +250,93 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
-          }
-          distinct_values.insert(it->value());
-          if (distinct_value_count + 1 != distinct_values.size()) {
-            break;
-          }
-          ++distinct_value_count;
-          ++it;
-        }
-      });
+      distinct_values_across_segments.insert(dictionary->cbegin(), dictionary->cend());
```
dey4ss: Given that we need to build the full set of values for _valid_ UCCs (and that filled-up chunks become immutable, but not encoded automatically yet), I think using the range insert for a non-nullable `ValueSegment`'s values does make sense performance-wise.
```c++
const auto dictionary_segment = ...;
const auto value_segment = ...;
if (dictionary_segment) {
  const auto& dictionary = dictionary_segment->dictionary();
  distinct_values_across_segments.insert(dictionary->cbegin(), dictionary->cend());
} else if (value_segment && !value_segment->is_nullable()) {
  const auto& values = value_segment->values();
  distinct_values_across_segments.insert(values.cbegin(), values.cend());
} else {
  continue;
}
...
```

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -18,15 +22,83 @@
...
+#include "storage/table_column_definition.hpp"
+#include "types.hpp"
 #include "utils/load_table.hpp"
 #include "utils/plugin_manager.hpp"
+#include "utils/template_type.hpp"
+
+namespace {
+using namespace hyrise;  // NOLINT(build/namespaces)
+
+void _encode_table(const std::shared_ptr<Table>& table, const SegmentEncodingSpec& encoding_spec) {
```
dey4ss: Pick here: These functions are not some protected/private members. Remove the underscore prefix.

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -277,19 +490,108 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, PluginFullRun) {
...
+  // Discover UCCs used in previous query.
+  auto& plugin_manager = Hyrise::get().plugin_manager;
+  plugin_manager.load_plugin(build_dylib_path("libhyriseUccDiscoveryPlugin"));
+  plugin_manager.exec_user_function("hyriseUccDiscoveryPlugin", "DiscoverUCCs");
+
+  // Check that key constraint was added to the table.
+  const auto constraint_A_c = TableKeyConstraint{{ColumnID{2}}, KeyConstraintType::UNIQUE};
+
+  EXPECT_EQ(_table_A->soft_key_constraints().size(), 2);
+  EXPECT_EQ(_table_A->soft_key_constraints().find(constraint_A_c)->last_validation_result(),
```
dey4ss: This can segfault if the constraint was not set, right? Please Split into two checks as you did below.

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -205,57 +250,93 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+    const std::shared_ptr<TransactionContext>& transaction_context) {
+  // `distinct_values_across_segments` collects the segment values from all chunks.
+  auto distinct_values_across_segments = std::unordered_set<ColumnDataType>{};
+  auto unmodified_chunks = std::vector<ChunkID>{};
+
   const auto chunk_count = table->chunk_count();
-  // `distinct_values` collects the segment values from all chunks.
-  auto distinct_values = std::unordered_set<ColumnDataType>{};
 
+  // For all segments we now want to verify if they contain duplicates. If a segment does not add the expected number
```
dey4ss:
```suggestion
  // For all segments, we now want to verify if they contain duplicates. If a segment does not add the expected number
```

src/plugins/ucc_discovery_plugin.cpp
```diff
@@ -205,57 +250,93 @@ bool UccDiscoveryPlugin::_dictionary_segments_contain_duplicates(const std::shar
...
+      // If this chunk does not exist, we do not need to check it with the validate operator later either.
+      unmodified_chunks.push_back(chunk_id);
       continue;
     }
     const auto source_segment = source_chunk->get_segment(column_id);
-    if (!source_segment) {
-      continue;
-    }
+    // Were values inserted into the chunk (inserted or updated rows)? If not and it is a dictionary segment, we can
+    // directly add all distinct values to the set of distinct values.
```
dey4ss: With the changes, we now really only check for deletions/updates, don't we?

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +307,167 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
-  EXPECT_FALSE(constraints_A.contains({{ColumnID{1}}, KeyConstraintType::UNIQUE}));
+  _validate_ucc_candidates(ucc_candidates);
+
+  EXPECT_EQ(constraints_A.size(), 1);
+  const auto& constraint_A_1 = constraints_A.find(TableKeyConstraint{{ColumnID{1}}, KeyConstraintType::UNIQUE});
+  ASSERT_NE(constraint_A_1, constraints_A.end());
+  EXPECT_EQ(constraint_A_1->last_validation_result(), ValidationResultType::VALID);
+
+  // Insert single row into table A that has a duplicate value regarding column 1. Do not commit to first test that UCC
+  // stays valid.
```
dey4ss:
```suggestion
  // Insert single row into table A that adds a duplicate value in column 1. Do not commit to first test that UCC
  // stays valid.
```

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +307,167 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+      constraints_A.find(TableKeyConstraint{{ColumnID{1}}, KeyConstraintType::UNIQUE});
+  ASSERT_NE(constraint_A_1_after_insert, constraints_A.end());
+  EXPECT_EQ(constraint_A_1_after_insert->last_validation_result(), ValidationResultType::VALID);
 
-  EXPECT_TRUE(constraints_B.contains({{ColumnID{0}}, KeyConstraintType::UNIQUE}));
-  EXPECT_FALSE(constraints_B.contains({{ColumnID{1}}, KeyConstraintType::UNIQUE}));
+  // Commit transaction and re-validate UCCs.
+  transaction_context->commit();
+
+  // Validate for a second time. This time the UCC should be invalid.
```
dey4ss:
```suggestion
  // Validate for a second time. This time, the UCC should be invalid.
```

src/test/plugins/ucc_discovery_plugin_test.cpp
```diff
@@ -246,15 +307,167 @@ TEST_P(UccDiscoveryPluginMultiEncodingTest, ValidateCandidates) {
...
+  ASSERT_NE(constraint_A_1_after_commit, constraints_A.end());
+  EXPECT_EQ(constraint_A_1_after_commit->last_validation_result(), ValidationResultType::INVALID);
+}
+
+TEST_P(UccDiscoveryPluginMultiEncodingTest, InvalidateCandidatesAfterUpdate) {
+  _encode_table(_table_A, GetParam());
+  const auto& constraints_A = _table_A->soft_key_constraints();
+  EXPECT_EQ(constraints_A.size(), 0);  // No constraints known for the table yet.
+
+  // Delete row of _table_A that had a duplicate value regarding column 1 such that column 1 is unique afterwards.
```
dey4ss:
```suggestion
  // Delete row of _table_A that had a duplicate value in column 1. Column 1 is unique afterwards.
```

dey4ss (approved):
Nice! Just two points, the rest is just minor stuff. Feel free to merge once you adressed them.

PR #2600, opened 2023-08-02
[DYOD] Only use valid UCCs in optimization and limit cacheability of plans optimized using UCCs

j-hellenberg:
Make sure to only use UCCs in optimization that are  `guaranteed_to_be_valid`.

Also, we want to ensure that Hyrise only caches logical and physical query plans that cannot become invalid, because executing an invalid query plan might produce incorrect results. Query plans could become invalid if they have been optimized using UCCs that are not permanent, e.g., incidental UCCs discovered by the `UccDiscoveryPlugin`. The same can happen if a Functional Dependency (FD) was used that was derived from a non-permanent UCC. We ensure correct caching by having each optimization rule return whether the resulting Logical Query Plan can be cached.

_Note that this PR currently targets [our first PR branch](https://github.com/hyrise/hyrise/pull/2599), in order to not include the changes from there again here :) Please review the other PR before this one. We will change the PR to target master before merging._

### DIFF
--- a/src/lib/logical_query_plan/abstract_lqp_node.cpp
+++ b/src/lib/logical_query_plan/abstract_lqp_node.cpp
@@ -274,17 +274,18 @@ bool AbstractLQPNode::is_column_nullable(const ColumnID column_id) const {
   return left_input()->is_column_nullable(column_id);
 }
 
-bool AbstractLQPNode::has_matching_ucc(const ExpressionUnorderedSet& expressions) const {
+std::optional<UniqueColumnCombination> AbstractLQPNode::get_matching_ucc(
+    const ExpressionUnorderedSet& expressions) const {
   DebugAssert(!expressions.empty(), "Invalid input. Set of expressions should not be empty.");
   DebugAssert(has_output_expressions(expressions),
               "The given expressions are not a subset of the LQP's output expressions.");
 
   const auto& unique_column_combinations = this->unique_column_combinations();
   if (unique_column_combinations.empty()) {
-    return false;
+    return std::nullopt;
   }
 
-  return contains_matching_unique_column_combination(unique_column_combinations, expressions);
+  return get_matching_unique_column_combination(unique_column_combinations, expressions);
 }
 
 FunctionalDependencies AbstractLQPNode::functional_dependencies() const {
--- a/src/lib/logical_query_plan/abstract_lqp_node.hpp
+++ b/src/lib/logical_query_plan/abstract_lqp_node.hpp
@@ -186,12 +186,12 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
   virtual UniqueColumnCombinations unique_column_combinations() const = 0;
 
   /**
-   * @return True if there is a unique column combination (UCC) matching the given subset of output expressions (i.e.,
-   *         the rows are guaranteed to be unique). This is preferred over calling
-   *         contains_matching_unique_column_combination(unique_column_combinations(), ...) as it performs additional
+   * @return Finds a unique column combination (UCC) matching the given subset of output expressions (i.e.,
+   *         the rows are guaranteed to be unique), if one exists. This is preferred over calling
+   *         get_matching_unique_column_combination(unique_column_combinations(), ...) as it performs additional
    *         sanity checks.
    */
-  bool has_matching_ucc(const ExpressionUnorderedSet& expressions) const;
+  std::optional<UniqueColumnCombination> get_matching_ucc(const ExpressionUnorderedSet& expressions) const;
 
   /**
    * @return The functional dependencies valid for this node. See functional_dependency.hpp for documentation.
--- a/src/lib/logical_query_plan/aggregate_node.cpp
+++ b/src/lib/logical_query_plan/aggregate_node.cpp
@@ -139,7 +139,7 @@ UniqueColumnCombinations AggregateNode::unique_column_combinations() const {
 
     // Make sure that we do not add an already existing or a superset UCC.
     if (unique_column_combinations.empty() ||
-        !contains_matching_unique_column_combination(unique_column_combinations, group_by_columns)) {
+        !get_matching_unique_column_combination(unique_column_combinations, group_by_columns).has_value()) {
       unique_column_combinations.emplace(group_by_columns);
     }
   }
--- a/src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
+++ b/src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
@@ -6,6 +6,8 @@ namespace hyrise {
 
 FunctionalDependency::FunctionalDependency(ExpressionUnorderedSet init_determinants,
                                            ExpressionUnorderedSet init_dependents)
+    // Most functional dependencies we use will be permanent because they are, for example, derived from a query
+    // directly. Therefore, use that as a default.
     : FunctionalDependency(std::move(init_determinants), std::move(init_dependents), true) {}
 
 FunctionalDependency::FunctionalDependency(hyrise::ExpressionUnorderedSet init_determinants,
--- a/src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
+++ b/src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
@@ -6,7 +6,11 @@ namespace hyrise {
 
 FunctionalDependency::FunctionalDependency(ExpressionUnorderedSet init_determinants,
                                            ExpressionUnorderedSet init_dependents)
-    : determinants(std::move(init_determinants)), dependents(std::move(init_dependents)) {
+    : FunctionalDependency(std::move(init_determinants), std::move(init_dependents), true) {}
+
+FunctionalDependency::FunctionalDependency(hyrise::ExpressionUnorderedSet init_determinants,
+                                           hyrise::ExpressionUnorderedSet init_dependents, bool permanent)
+    : determinants(std::move(init_determinants)), dependents(std::move(init_dependents)), permanent(permanent) {
   DebugAssert(!determinants.empty() && !dependents.empty(), "FunctionalDependency cannot be empty");
 }
 
@@ -50,6 +54,10 @@ size_t FunctionalDependency::hash() const {
   return boost::hash_value(hash - determinants.size());
 }
 
+bool FunctionalDependency::is_permanent() const {
+  return permanent;
+}
+
 std::ostream& operator<<(std::ostream& stream, const FunctionalDependency& expression) {
   stream << "{";
   auto determinants_vector =
@@ -83,7 +91,7 @@ FunctionalDependencies inflate_fds(const FunctionalDependencies& fds) {
       inflated_fds.insert(fd);
     } else {
       for (const auto& dependent : fd.dependents) {
-        inflated_fds.emplace(fd.determinants, ExpressionUnorderedSet{dependent});
+        inflated_fds.emplace(fd.determinants, ExpressionUnorderedSet{dependent}, fd.is_permanent());
       }
     }
   }
@@ -97,14 +105,14 @@ FunctionalDependencies deflate_fds(const FunctionalDependencies& fds) {
   }
 
   // We cannot use a set here as we want to add dependents to existing FDs and objects in sets are immutable.
-  auto existing_fds = std::vector<std::pair<ExpressionUnorderedSet, ExpressionUnorderedSet>>{};
+  auto existing_fds = std::vector<std::tuple<ExpressionUnorderedSet, ExpressionUnorderedSet, bool>>{};
   existing_fds.reserve(fds.size());
 
   for (const auto& fd_to_add : fds) {
     // Check if we have seen an FD with the same determinants.
     auto existing_fd = std::find_if(existing_fds.begin(), existing_fds.end(), [&](const auto& fd) {
       // Quick check for cardinality.
-      const auto& determinants = fd.first;
+      const auto& determinants = std::get<0>(fd);
       if (determinants.size() != fd_to_add.determinants.size()) {
         return false;
       }
@@ -121,15 +129,18 @@ FunctionalDependencies deflate_fds(const FunctionalDependencies& fds) {
 
     // If we have found an FD with same determinants, add the dependents. Otherwise, add a new FD.
     if (existing_fd != existing_fds.end()) {
-      existing_fd->second.insert(fd_to_add.dependents.cbegin(), fd_to_add.dependents.cend());
+      std::get<1>(*existing_fd).insert(fd_to_add.dependents.cbegin(), fd_to_add.dependents.cend());
+      if (!fd_to_add.is_permanent()) {
+        std::get<2>(*existing_fd) = false;
+      }
     } else {
-      existing_fds.emplace_back(fd_to_add.determinants, fd_to_add.dependents);
+      existing_fds.emplace_back(fd_to_add.determinants, fd_to_add.dependents, fd_to_add.is_permanent());
     }
   }
 
   auto deflated_fds = FunctionalDependencies(fds.size());
-  for (const auto& [determinants, dependents] : existing_fds) {
-    deflated_fds.emplace(determinants, dependents);
+  for (const auto& [determinants, dependents, is_permanent] : existing_fds) {
+    deflated_fds.emplace(determinants, dependents, is_permanent);
   }
 
   return deflated_fds;
--- a/src/lib/logical_query_plan/data_dependencies/functional_dependency.hpp
+++ b/src/lib/logical_query_plan/data_dependencies/functional_dependency.hpp
@@ -26,16 +26,25 @@ namespace hyrise {
  *
  * Currently, the determinant expressions are required to be non-nullable to be involved in FDs. Combining null values
  * and FDs is not trivial. For more reference, see https://arxiv.org/abs/1404.4963.
+ *
+ * If the FD may become invalid in the future (because it is not based on a schema constraint, but on the data
+ * incidentally fulfilling the constraint at the moment), the FD is marked as being not permanent.
  */
 struct FunctionalDependency {
   FunctionalDependency(ExpressionUnorderedSet init_determinants, ExpressionUnorderedSet init_dependents);
+  FunctionalDependency(ExpressionUnorderedSet init_determinants, ExpressionUnorderedSet init_dependents,
+                       bool permanent);
 
   bool operator==(const FunctionalDependency& other) const;
   bool operator!=(const FunctionalDependency& other) const;
   size_t hash() const;
 
+  bool is_permanent() const;
+
   ExpressionUnorderedSet determinants;
   ExpressionUnorderedSet dependents;
+
+  bool permanent;
 };
 
 std::ostream& operator<<(std::ostream& stream, const FunctionalDependency& expression);
--- a/src/lib/logical_query_plan/data_dependencies/functional_dependency.hpp
+++ b/src/lib/logical_query_plan/data_dependencies/functional_dependency.hpp
@@ -29,6 +29,8 @@ namespace hyrise {
  *
  * If the FD may become invalid in the future (because it is not based on a schema constraint, but on the data
  * incidentally fulfilling the constraint at the moment), the FD is marked as being not permanent.
+ * This information is important because query plans that were optimized using a non-permanent FD are probably
+ * not cacheable.
  */
 struct FunctionalDependency {
   FunctionalDependency(ExpressionUnorderedSet init_determinants, ExpressionUnorderedSet init_dependents);
--- a/src/lib/logical_query_plan/lqp_utils.cpp
+++ b/src/lib/logical_query_plan/lqp_utils.cpp
@@ -523,7 +523,7 @@ FunctionalDependencies fds_from_unique_column_combinations(const std::shared_ptr
                                return (fd.determinants == determinants) && (fd.dependents == dependents);
                              }) == fds.cend(),
                 "Creating duplicate functional dependencies is unexpected.");
-    fds.emplace(determinants, dependents);
+    fds.emplace(determinants, dependents, ucc.is_permanent());
   }
   return fds;
 }
--- a/src/lib/logical_query_plan/lqp_utils.cpp
+++ b/src/lib/logical_query_plan/lqp_utils.cpp
@@ -462,8 +462,8 @@ template ExpressionUnorderedSet find_column_expressions(const AbstractLQPNode& l
 template ExpressionUnorderedSet find_column_expressions(const AbstractLQPNode& lqp_node,
                                                         const std::vector<ColumnID>& column_ids);
 
-bool contains_matching_unique_column_combination(const UniqueColumnCombinations& unique_column_combinations,
-                                                 const ExpressionUnorderedSet& expressions) {
+std::optional<UniqueColumnCombination> get_matching_unique_column_combination(
+    const UniqueColumnCombinations& unique_column_combinations, const ExpressionUnorderedSet& expressions) {
   DebugAssert(!unique_column_combinations.empty(), "Invalid input: Set of UCCs should not be empty.");
   DebugAssert(!expressions.empty(), "Invalid input: Set of expressions should not be empty.");
 
@@ -473,11 +473,11 @@ bool contains_matching_unique_column_combination(const UniqueColumnCombinations&
         std::all_of(ucc.expressions.cbegin(), ucc.expressions.cend(),
                     [&](const auto& ucc_expression) { return expressions.contains(ucc_expression); })) {
       // Found a matching UCC.
-      return true;
+      return ucc;
     }
   }
   // Did not find a UCC for the given expressions.
-  return false;
+  return std::nullopt;
 }
 
 FunctionalDependencies fds_from_unique_column_combinations(const std::shared_ptr<const AbstractLQPNode>& lqp,
--- a/src/lib/logical_query_plan/stored_table_node.cpp
+++ b/src/lib/logical_query_plan/stored_table_node.cpp
@@ -117,7 +117,7 @@ UniqueColumnCombinations StoredTableNode::unique_column_combinations() const {
                 "Unexpected count of column expressions.");
 
     // Create UniqueColumnCombination
-    unique_column_combinations.emplace(column_expressions);
+    unique_column_combinations.emplace(column_expressions, !table_key_constraint.can_become_invalid());
   }
 
   return unique_column_combinations;
--- a/src/lib/logical_query_plan/stored_table_node.cpp
+++ b/src/lib/logical_query_plan/stored_table_node.cpp
@@ -105,6 +105,12 @@ UniqueColumnCombinations StoredTableNode::unique_column_combinations() const {
       continue;
     }
 
+    // We may only use the key constraints as UCCs for optimization purposes if they are certainly still valid,
+    // otherwise these optimizations could produce invalid query results.
+    if (!table->constraint_guaranteed_to_be_valid(table_key_constraint)) {
+      continue;
+    }
+
     // Search for expressions representing the key constraint's ColumnIDs.
     const auto& column_expressions = find_column_expressions(*this, table_key_constraint.columns());
     DebugAssert(column_expressions.size() == table_key_constraint.columns().size(),
--- a/src/lib/optimizer/optimizer.hpp
+++ b/src/lib/optimizer/optimizer.hpp
@@ -16,6 +16,15 @@ class AbstractRule;
 class AbstractLQPNode;
 
 /**
+ * Holds the result of optimizing an LQP, which consists of the optimized LQP and the information on whether it can be
+ * cached.
+ */
+struct OptimizedLogicalQueryPlan {
+  bool cacheable{true};
+  std::shared_ptr<AbstractLQPNode> logical_query_plan;
+};
+
+/**
  * Applies optimization rules to an LQP.
  * On each invocation of optimize(), these Batches are applied in the same order as they were added
  * to the Optimizer.
@@ -38,7 +47,7 @@ class Optimizer final {
    * Returns optimized version of @param input.
    * @param rule_durations may be set in order to retrieve runtime information for each applied rule.
    */
-  std::shared_ptr<AbstractLQPNode> optimize(
+  OptimizedLogicalQueryPlan optimize(
       std::shared_ptr<AbstractLQPNode> input,
       const std::shared_ptr<std::vector<OptimizerRuleMetrics>>& rule_durations = nullptr) const;
 
--- a/src/lib/optimizer/strategy/abstract_rule.cpp
+++ b/src/lib/optimizer/strategy/abstract_rule.cpp
@@ -9,9 +9,17 @@
 
 namespace hyrise {
 
-void AbstractRule::apply_to_plan(const std::shared_ptr<LogicalPlanRootNode>& lqp_root) const {
+IsCacheable operator&(IsCacheable lhs, IsCacheable rhs) {
+  return static_cast<IsCacheable>(static_cast<bool>(lhs) & static_cast<bool>(rhs));
+}
+
+IsCacheable& operator&=(IsCacheable& lhs, const IsCacheable rhs) {
+  return lhs = lhs & rhs;
+}
+
+IsCacheable AbstractRule::apply_to_plan(const std::shared_ptr<LogicalPlanRootNode>& lqp_root) const {
   // (1) Optimize root LQP
-  _apply_to_plan_without_subqueries(lqp_root);
+  auto cacheable = _apply_to_plan_without_subqueries(lqp_root);
 
   // (2) Optimize distinct subquery LQPs, one-by-one.
   auto subquery_expressions_by_lqp = collect_lqp_subquery_expressions_by_lqp(lqp_root);
@@ -23,7 +31,7 @@ void AbstractRule::apply_to_plan(const std::shared_ptr<LogicalPlanRootNode>& lqp
 
     // (2.1) Optimize subplan
     const auto local_lqp_root = LogicalPlanRootNode::make(lqp);
-    _apply_to_plan_without_subqueries(local_lqp_root);
+    cacheable &= _apply_to_plan_without_subqueries(local_lqp_root);
 
     // (2.2) Assign optimized subplan to all corresponding SubqueryExpressions
     for (const auto& subquery_expression : subquery_expressions) {
@@ -33,6 +41,8 @@ void AbstractRule::apply_to_plan(const std::shared_ptr<LogicalPlanRootNode>& lqp
     // (2.3) Untie the root node before it goes out of scope so that the outputs of the LQP remain correct.
     local_lqp_root->set_left_input(nullptr);
   }
+
+  return cacheable;
 }
 
 }  // namespace hyrise
--- a/src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
+++ b/src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
@@ -72,8 +72,10 @@ std::string DependentGroupByReductionRule::name() const {
   return name;
 }
 
-void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
+IsCacheable DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
     const std::shared_ptr<AbstractLQPNode>& lqp_root) const {
+  auto rule_was_applied_using_non_permanent_ucc = false;
+
   visit_lqp(lqp_root, [&](const auto& node) {
     if (node->type != LQPNodeType::Aggregate) {
       return LQPVisitation::VisitInputs;
@@ -187,6 +189,8 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
 
     return LQPVisitation::VisitInputs;
   });
+
+  return rule_was_applied_using_non_permanent_ucc ? IsCacheable::No : IsCacheable::Yes;
 }
 
 }  // namespace hyrise
--- a/src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
+++ b/src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
@@ -99,21 +99,25 @@ IsCacheable DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
     // Example: SELECT DISTINCT n_nationkey FROM nation is reflected as [ Aggregate GROUP BY n_nationkey ]. The
     // AggregateNode does not have more node expressions than the single GROUP BY column and the input (StoredTableNode
     // in this case) has a UCC for { n_nationkey }.
-    if (group_by_columns.size() == node->node_expressions.size() &&
-        node->left_input()->has_matching_ucc(group_by_columns)) {
-      const auto& output_expressions = aggregate_node.output_expressions();
-      // Remove the AggregateNode if it does not limit or reorder the output expressions.
-      if (expressions_equal(output_expressions, node->left_input()->output_expressions())) {
-        lqp_remove_node(node);
+    if (group_by_columns.size() == node->node_expressions.size()) {
+      auto matching_ucc = node->left_input()->get_matching_ucc(group_by_columns);
+      if (matching_ucc.has_value()) {
+        rule_was_applied_using_non_permanent_ucc = !matching_ucc->is_permanent();
+
+        const auto& output_expressions = aggregate_node.output_expressions();
+        // Remove the AggregateNode if it does not limit or reorder the output expressions.
+        if (expressions_equal(output_expressions, node->left_input()->output_expressions())) {
+          lqp_remove_node(node);
+          return LQPVisitation::VisitInputs;
+        }
+
+        // Else, replace it with a ProjectionNode. For instance, SELECT DISTINCT n_name, n_regionkey FROM nation (the
+        // column order is different than in the original table) turns from [ Aggregate GROUP BY n_name, n_regionkey ]
+        // to [ Projection n_name, n_regionkey ].
+        const auto projection_node = ProjectionNode::make(output_expressions);
+        lqp_replace_node(node, projection_node);
         return LQPVisitation::VisitInputs;
       }
-
-      // Else, replace it with a ProjectionNode. For instance, SELECT DISTINCT n_name, n_regionkey FROM nation (the
-      // column order is different than in the original table) turns from [ Aggregate GROUP BY n_name, n_regionkey ] to
-      // [ Projection n_name, n_regionkey ].
-      const auto projection_node = ProjectionNode::make(output_expressions);
-      lqp_replace_node(node, projection_node);
-      return LQPVisitation::VisitInputs;
     }
 
     // Early exit (ii): If there are no functional dependencies, we can skip this rule.
@@ -169,6 +173,10 @@ IsCacheable DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
 
       const auto success = remove_dependent_group_by_columns(fd, aggregate_node, group_by_columns);
       if (success) {
+        // Functional dependencies are derived from UCCs. In case we encounter a non-permanent FD,
+        // this means we encountered an underlying non-permanent UCC as well.
+        rule_was_applied_using_non_permanent_ucc |= !fd.is_permanent();
+
         // Refresh data structures correspondingly.
         group_by_list_changed = true;
         group_by_columns = fetch_group_by_columns();
--- a/src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
+++ b/src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
@@ -59,60 +59,60 @@ void gather_rewrite_info(
   // predicate that filters on a UCC, a maximum of one tuple remains in the result relation. Since at this point, we al-
   // ready know the candidate join is basically a semi join, we can further transform the join to a single predicate
   // node filtering the join column for the value of the remaining tuple's join attribute.
-  visit_lqp(
-      removable_subtree, [&removable_subtree, &rewrite_predicate, &non_permanent_ucc_was_used](auto& current_node) {
-        if (current_node->type == LQPNodeType::Union) {
-          return LQPVisitation::DoNotVisitInputs;
-        }
-        if (current_node->type != LQPNodeType::Predicate) {
-          return LQPVisitation::VisitInputs;
-        }
+  visit_lqp(removable_subtree, [&removable_subtree, &rewrite_predicate,
+                                &non_permanent_ucc_was_used](const auto& current_node) {
+    if (current_node->type == LQPNodeType::Union) {
+      return LQPVisitation::DoNotVisitInputs;
+    }
+    if (current_node->type != LQPNodeType::Predicate) {
+      return LQPVisitation::VisitInputs;
+    }
 
-        // We need to get a predicate node with a BinaryPredicateExpression.
-        const auto candidate = std::static_pointer_cast<PredicateNode>(current_node);
-        const auto candidate_expression = std::dynamic_pointer_cast<BinaryPredicateExpression>(candidate->predicate());
-        if (!candidate_expression) {
-          return LQPVisitation::VisitInputs;
-        }
+    // We need to get a predicate node with a BinaryPredicateExpression.
+    const auto candidate = std::static_pointer_cast<PredicateNode>(current_node);
+    const auto candidate_expression = std::dynamic_pointer_cast<BinaryPredicateExpression>(candidate->predicate());
+    if (!candidate_expression) {
+      return LQPVisitation::VisitInputs;
+    }
 
-        // Only predicates in the form `column = value` are useful to our optimization. These conditions have the
-        // potential (given filtered column is a UCC) to emit at most one result tuple.
-        if (candidate_expression->predicate_condition != PredicateCondition::Equals) {
-          return LQPVisitation::VisitInputs;
-        }
+    // Only predicates in the form `column = value` are useful to our optimization. These conditions have the
+    // potential (given filtered column is a UCC) to emit at most one result tuple.
+    if (candidate_expression->predicate_condition != PredicateCondition::Equals) {
+      return LQPVisitation::VisitInputs;
+    }
 
-        auto candidate_column_expression = std::shared_ptr<AbstractExpression>{};
-        auto candidate_value_expression = std::shared_ptr<AbstractExpression>{};
+    auto candidate_column_expression = std::shared_ptr<AbstractExpression>{};
+    auto candidate_value_expression = std::shared_ptr<AbstractExpression>{};
 
-        for (const auto& predicate_operand : candidate_expression->arguments) {
-          if (predicate_operand->type == ExpressionType::LQPColumn) {
-            candidate_column_expression = predicate_operand;
-          } else if (predicate_operand->type == ExpressionType::Value) {
-            candidate_value_expression = predicate_operand;
-          }
-        }
+    for (const auto& predicate_operand : candidate_expression->arguments) {
+      if (predicate_operand->type == ExpressionType::LQPColumn) {
+        candidate_column_expression = predicate_operand;
+      } else if (predicate_operand->type == ExpressionType::Value) {
+        candidate_value_expression = predicate_operand;
+      }
+    }
 
-        // There should not be a case where we run into no column, but there may be a case where we have no value but
-        // compare columns.
-        if (!candidate_value_expression || !candidate_column_expression) {
-          return LQPVisitation::VisitInputs;
-        }
+    // There should not be a case where we run into no column, but there may be a case where we have no value but
+    // compare columns.
+    if (!candidate_value_expression || !candidate_column_expression) {
+      return LQPVisitation::VisitInputs;
+    }
 
-        // Check whether the referenced column is available for the subtree root node and unique. Checking whether the
-        // column is unique on the current node is not sufficient. There could be unions or joins in between the subtree
-        // root and the current node, invalidating the unique column combination.
-        if (!expression_evaluable_on_lqp(candidate_column_expression, *removable_subtree)) {
-          return LQPVisitation::VisitInputs;
-        }
-        auto matching_ucc = removable_subtree->get_matching_ucc({candidate_column_expression});
-        if (!matching_ucc.has_value()) {
-          return LQPVisitation::VisitInputs;
-        }
+    // Check whether the referenced column is available for the subtree root node and unique. Checking whether the
+    // column is unique on the current node is not sufficient. There could be unions or joins between the subtree
+    // root and the current node, invalidating the unique column combination.
+    if (!expression_evaluable_on_lqp(candidate_column_expression, *removable_subtree)) {
+      return LQPVisitation::VisitInputs;
+    }
+    const auto matching_ucc = removable_subtree->get_matching_ucc({candidate_column_expression});
+    if (!matching_ucc.has_value()) {
+      return LQPVisitation::VisitInputs;
+    }
 
-        rewrite_predicate = candidate;
-        non_permanent_ucc_was_used |= !matching_ucc->is_permanent();
-        return LQPVisitation::DoNotVisitInputs;
-      });
+    rewrite_predicate = candidate;
+    non_permanent_ucc_was_used |= !matching_ucc->is_permanent();
+    return LQPVisitation::DoNotVisitInputs;
+  });
 
   if (rewrite_predicate) {
     rewritables.emplace_back(join_node, *prunable_side, rewrite_predicate);
--- a/src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
+++ b/src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
@@ -51,7 +51,7 @@ void gather_rewrite_info(
          "Neither column of the join predicate could be evaluated on the removable input.");
 
   // Check for uniqueness.
-  if (!removable_subtree->has_matching_ucc({exchangeable_column_expression})) {
+  if (!removable_subtree->get_matching_ucc({exchangeable_column_expression}).has_value()) {
     return;
   }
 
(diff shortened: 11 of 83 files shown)


src/lib/optimizer/strategy/abstract_rule.cpp
```diff
@@ -9,9 +9,17 @@
 
 namespace hyrise {
 
-void AbstractRule::apply_to_plan(const std::shared_ptr<LogicalPlanRootNode>& lqp_root) const {
+IsCacheable operator&(IsCacheable lhs, IsCacheable rhs) {
+  return static_cast<IsCacheable>(static_cast<bool>(lhs) & static_cast<bool>(rhs));
+}
+
+IsCacheable& operator&=(IsCacheable& lhs, const IsCacheable rhs) {
```
SanJSp, line 16: out of curiosity: why is only `IsCacheable` parameter passed as reference?

src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
```diff
@@ -50,62 +51,68 @@ void gather_rewrite_info(
...
-    if (!expression_evaluable_on_lqp(candidate_column_expression, *removable_subtree) ||
-        !removable_subtree->has_matching_ucc({candidate_column_expression})) {
-      return LQPVisitation::VisitInputs;
-    }
-
-    rewrite_predicate = candidate;
-    return LQPVisitation::DoNotVisitInputs;
-  });
+  visit_lqp(
+      removable_subtree, [&removable_subtree, &rewrite_predicate, &non_permanent_ucc_was_used](auto& current_node) {
```
SanJSp, line 63: Is this possible?
```suggestion
      removable_subtree, [&removable_subtree, &rewrite_predicate, &non_permanent_ucc_was_used](const auto& current_node) {
```

src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
```diff
@@ -50,62 +51,68 @@ void gather_rewrite_info(
...
+        }
+
+        // There should not be a case where we run into no column, but there may be a case where we have no value but
+        // compare columns.
+        if (!candidate_value_expression || !candidate_column_expression) {
+          return LQPVisitation::VisitInputs;
+        }
+
+        // Check whether the referenced column is available for the subtree root node and unique. Checking whether the
+        // column is unique on the current node is not sufficient. There could be unions or joins in between the subtree
```
SanJSp, line 102:
```suggestion
        // column is unique on the current node is not sufficient. There could be unions or joins between the subtree
```

src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
```diff
@@ -50,62 +51,68 @@ void gather_rewrite_info(
...
+        // root and the current node, invalidating the unique column combination.
+        if (!expression_evaluable_on_lqp(candidate_column_expression, *removable_subtree)) {
+          return LQPVisitation::VisitInputs;
+        }
+        auto matching_ucc = removable_subtree->get_matching_ucc({candidate_column_expression});
```
SanJSp, line 107: Is this possible?
```suggestion
        const auto matching_ucc = removable_subtree->get_matching_ucc({candidate_column_expression});
```

src/lib/optimizer/strategy/join_to_semi_join_rule.cpp
```diff
@@ -72,20 +74,30 @@ void JoinToSemiJoinRule::_apply_to_plan_without_subqueries(const std::shared_ptr
...
-          join_node->left_input()->has_matching_ucc(equals_predicate_expressions_left)) {
-        join_node->join_mode = JoinMode::Semi;
-        const auto temp = join_node->left_input();
-        join_node->set_left_input(join_node->right_input());
-        join_node->set_right_input(temp);
-      } else if (*join_node->prunable_input_side() == LQPInputSide::Right &&
-                 join_node->right_input()->has_matching_ucc(equals_predicate_expressions_right)) {
-        join_node->join_mode = JoinMode::Semi;
+      if (*join_node->prunable_input_side() == LQPInputSide::Left) {
+        auto matching_ucc = join_node->left_input()->get_matching_ucc(equals_predicate_expressions_left);
```
SanJSp, line 78: Again, not sure if this can be const
```suggestion
        const auto matching_ucc = join_node->left_input()->get_matching_ucc(equals_predicate_expressions_left);
```

SanJSp: (same below)

src/lib/optimizer/strategy/join_to_semi_join_rule.cpp
```diff
@@ -72,20 +74,30 @@ void JoinToSemiJoinRule::_apply_to_plan_without_subqueries(const std::shared_ptr
...
+
+          join_node->join_mode = JoinMode::Semi;
+        }
       }
     }
 
     return LQPVisitation::VisitInputs;
   });
+
+  return rule_was_applied_using_non_permanent_ucc ? IsCacheable::No : IsCacheable::Yes;
```
SanJSp, line 100: Like above, not sure if following coding conventions

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -123,7 +117,7 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_optimized_logi
       if (lqp_is_validated(plan) == (_use_mvcc == UseMvcc::Yes)) {
         // Copy the LQP for reuse as the LQPTranslator might modify mutable fields (e.g., cached output_expressions)
         // and concurrent translations might conflict.
-        _optimized_logical_plan = plan->deep_copy();
+        _optimized_logical_plan = {true, plan->deep_copy()};
```
SanJSp, line 120: Like in the other PR proposed, I like this style, as I am not sure what is actually being set to true here. Maybe you deem it a good idea as well :)
```suggestion
        _optimized_logical_plan = {/* bool variable changed here =*/ true, plan->deep_copy()};
```

src/lib/sql/sql_pipeline_statement.hpp
```diff
@@ -75,7 +75,7 @@ class SQLPipelineStatement : public Noncopyable {
   const SQLTranslationInfo& get_sql_translation_info();
 
   // Returns the optimized LQP for this statement.
-  const std::shared_ptr<AbstractLQPNode>& get_optimized_logical_plan();
+  OptimizedLogicalQueryPlan get_optimized_logical_plan();
```
SanJSp, line 78: Why can't this be `const` anymore? Maybe it's possible to const this

src/test/lib/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -332,27 +334,54 @@ TEST_F(StoredTableNodeTest, UniqueColumnCombinationsPrunedColumns) {
...
+  const auto& table_key_constraints = _table_a->soft_key_constraints();
+  EXPECT_EQ(table_key_constraints.size(), 2);
+  EXPECT_EQ(_stored_table_node->unique_column_combinations().size(), 2);
+
+  // Modify the table so that the UCC is no longer guaranteed to be valid
+  // (in fact, it is actually no longer valid because we duplicated all rows).
+  auto transaction_context = Hyrise::get().transaction_manager.new_transaction_context(AutoCommit::No);
+  const auto get_table = std::make_shared<GetTable>("t_a");
+  get_table->execute();
+  auto insert_op = std::make_shared<Insert>("t_a", get_table);
```
SanJSp, line 352: `const`?

src/test/lib/optimizer/optimizer_test.cpp
```diff
@@ -331,4 +337,43 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
   }
 }
 
+TEST_F(OptimizerTest, CheckTrueCacheablility) {
+  auto optimizer = Optimizer::create_default_optimizer();
+
+  auto lqp = ProjectionNode::make(expression_vector(add_(b, subquery_a)),
```
SanJSp, line 343: Can those be `const`?

src/test/lib/optimizer/optimizer_test.cpp
```diff
@@ -331,4 +337,43 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
...
+                                  PredicateNode::make(greater_than_(a, subquery_b), node_a));
+  auto optimize_results = optimizer->optimize(std::move(lqp));
```
SanJSp, line 345:
```suggestion
  const auto optimize_results = optimizer->optimize(std::move(lqp));
```

src/test/lib/optimizer/optimizer_test.cpp
```diff
@@ -331,4 +337,43 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
...
+  // clang-format off
+  auto lqp1 =
+    ProjectionNode::make(expression_vector(add_(a, 2)),
+      JoinNode::make(JoinMode::Inner, equals_(a, column0),
+        ProjectionNode::make(expression_vector(a),
+          node_a),
+        stored_table_node));
+  // clang-format on
+  static_cast<JoinNode&>(*lqp1->left_input()).mark_input_side_as_prunable(LQPInputSide::Right);
+  auto optimize_results1 = optimizer.optimize(std::move(lqp1));
```
SanJSp, line 375: `const` here and maybe in the lines above as well?

src/test/lib/optimizer/strategy/join_to_predicate_rewrite_rule_test.cpp
```diff
@@ -162,12 +175,15 @@ TEST_F(JoinToPredicateRewriteRuleTest, NoUnusedJoinSide) {
...
+  const auto expected_lqp = annotated_lqp->deep_copy();
+  const auto optimization_result = apply_rule_with_cacheability_check(rule, annotated_lqp);
+
+  EXPECT_TRUE(optimization_result.cacheable);  // Cacheable because rule was not applied
+  EXPECT_LQP_EQ(optimization_result.logical_query_plan, expected_lqp);
 }
 
 TEST_F(JoinToPredicateRewriteRuleTest, Union) {
-  // Do not rewrite if there is a union on table b.
+  // Do not rewrite if there is a union on table b that preserves the UCC but outputs more than one result tuple..
```
SanJSp, line 186:
```suggestion
  // Do not rewrite if there is a union on table b that preserves the UCC but outputs more than one result tuple.
```

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -344,6 +345,42 @@ TEST_F(SQLPipelineStatementTest, GetCachedOptimizedLQPNotValidated) {
...
+
+  auto validated_sql_pipeline =
+      SQLPipelineBuilder{_select_query_a}.with_lqp_cache(_lqp_cache).with_pqp_cache(_pqp_cache).create_pipeline();
+  auto& validated_statement = get_sql_pipeline_statements(validated_sql_pipeline).at(0);
+
+  const auto& validated_lqp = validated_statement->get_optimized_logical_plan().logical_query_plan;
+  EXPECT_TRUE(lqp_is_validated(validated_lqp));
+  // We need to call this method to generate and cache the PQP.
+  validated_statement->get_physical_plan();
+  // Expect cache to still not contain validated LQP as we ran a query that used a non-cacheable optimization
```
SanJSp, line 360:
```suggestion
  // Expect cache to still not contain validated LQP as we ran a query that used a non-cacheable optimization.
```

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -344,6 +345,42 @@ TEST_F(SQLPipelineStatementTest, GetCachedOptimizedLQPNotValidated) {
...
+  EXPECT_TRUE(_lqp_cache->has(_select_query_a));
+  EXPECT_TRUE(_pqp_cache->has(_select_query_a));
+}
+
+TEST_F(SQLPipelineStatementTest, OptimizedQPNotCachedWhenNotCacheableOptimizationUsed) {
+  // Expect cache to be empty
```
SanJSp, line 366:
```suggestion
  // Expect cache to be empty.
```

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -344,6 +345,42 @@ TEST_F(SQLPipelineStatementTest, GetCachedOptimizedLQPNotValidated) {
...
+  EXPECT_FALSE(_lqp_cache->has(_select_query_a));
+
+  auto validated_sql_pipeline =
+      SQLPipelineBuilder{_select_query_a}.with_lqp_cache(_lqp_cache).with_pqp_cache(_pqp_cache).create_pipeline();
+  auto& validated_statement = get_sql_pipeline_statements(validated_sql_pipeline).at(0);
+
+  const auto& validated_lqp = validated_statement->get_optimized_logical_plan().logical_query_plan;
+  EXPECT_TRUE(lqp_is_validated(validated_lqp));
+  // We need to call this method to generate and cache the PQP.
```
SanJSp, line 358: Either remove the period here or add below and above 😁

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -344,6 +345,42 @@ TEST_F(SQLPipelineStatementTest, GetCachedOptimizedLQPNotValidated) {
...
+
+TEST_F(SQLPipelineStatementTest, OptimizedQPNotCachedWhenNotCacheableOptimizationUsed) {
+  // Expect cache to be empty
+  EXPECT_FALSE(_lqp_cache->has(_select_query_using_join_to_semi_join_optimization_a));
+
+  auto validated_sql_pipeline = SQLPipelineBuilder{_select_query_using_join_to_semi_join_optimization_a}
+                                    .with_lqp_cache(_lqp_cache)
+                                    .with_pqp_cache(_pqp_cache)
+                                    .create_pipeline();
+  auto& validated_statement = get_sql_pipeline_statements(validated_sql_pipeline).at(0);
```
SanJSp, line 373: is it possible to `const` this?

SanJSp (commented):
A looooot of changes, but good job! 

Again only minor comments and all in all great changes! Feel free to consider my comments and if they're not relevant, skip them.

For the future: Maybe, when touching so many files, it makes sense to specify an order in which the reviewer can work through the files.

Happy coding!

src/lib/optimizer/optimizer.hpp
```diff
@@ -15,6 +15,15 @@ struct OptimizerRuleMetrics {
 class AbstractRule;
 class AbstractLQPNode;
 
+/**
+ * Holds the result of optimizing an LQP, which consists of the optimized LQP and the information on whether it can be
+ * cached.
+ */
+struct OptimizedLogicalQueryPlan {
+  bool cacheable{true};
```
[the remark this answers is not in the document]
Margarete01: No, if I try to change `bool` to `auto`, I get the following error `'auto' not allowed in non-static struct member`.

src/test/lib/optimizer/optimizer_test.cpp
Margarete01: No, they can't. 
If I try to make optimizer const, the build fails with the following message: `error: passing 'const hyrise::Optimizer' as 'this' argument discards qualifiers`.
If I try to make lqp const, the optimizer wouldn't have the only ownership over the query plan which it needs and the test fails.

src/lib/optimizer/strategy/abstract_rule.cpp
Margarete01: I'm not quite sure if I get the question right: `lhs` is passed by reference, because we actually want to modify this parameter with the operator. We don't want to modify `rhs`, so this parameter is passed by value.
But I guess, the return type of the function can be `IsCacheable` instead of `IsCacheable&`. Good catch!

src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
```diff
@@ -6,7 +6,11 @@ namespace hyrise {
 
 FunctionalDependency::FunctionalDependency(ExpressionUnorderedSet init_determinants,
                                            ExpressionUnorderedSet init_dependents)
-    : determinants(std::move(init_determinants)), dependents(std::move(init_dependents)) {
+    : FunctionalDependency(std::move(init_determinants), std::move(init_dependents), true) {}
```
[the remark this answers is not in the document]
j-hellenberg: > I've recently seen this in somebody else's code and I thought it improves readability. Maybe you find it useful as well?

While they improve readability, I personally dislike adding comments like this because they need to be maintained as well and can easily become out of sync. Also, in this case, the constructor that is called is just two lines below that, so one can easily look there to see how the parameter is named.

> I also wondered why this is being set to true. Maybe it makes sense to explain this in a comment?

I added a comment for that

src/lib/logical_query_plan/data_dependencies/functional_dependency.hpp
```diff
@@ -26,16 +26,25 @@ namespace hyrise {
  *
  * Currently, the determinant expressions are required to be non-nullable to be involved in FDs. Combining null values
  * and FDs is not trivial. For more reference, see https://arxiv.org/abs/1404.4963.
+ *
+ * If the FD may become invalid in the future (because it is not based on a schema constraint, but on the data
+ * incidentally fulfilling the constraint at the moment), the FD is marked as being not permanent.
```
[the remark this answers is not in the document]
j-hellenberg: I added a comment :)

src/lib/optimizer/strategy/abstract_rule.cpp
j-hellenberg: Actually, the explanation regarding the `lhs` is correct, but returning a reference is just the normal way to implement that operator: returning the result of the assignment again instead of making a copy.

src/lib/sql/sql_pipeline_statement.cpp
j-hellenberg: I added a comment on why it is set to true

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -97,21 +99,25 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
...
     // in this case) has a UCC for { n_nationkey }.
-    if (group_by_columns.size() == node->node_expressions.size() &&
-        node->left_input()->has_matching_ucc(group_by_columns)) {
-      const auto& output_expressions = aggregate_node.output_expressions();
-      // Remove the AggregateNode if it does not limit or reorder the output expressions.
-      if (expressions_equal(output_expressions, node->left_input()->output_expressions())) {
-        lqp_remove_node(node);
+    if (group_by_columns.size() == node->node_expressions.size()) {
+      auto matching_ucc = node->left_input()->get_matching_ucc(group_by_columns);
+      if (matching_ucc.has_value()) {
```
dey4ss, line 104: w/o has_value

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -72,8 +72,10 @@ std::string DependentGroupByReductionRule::name() const {
   return name;
 }
 
-void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
+IsCacheable DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
     const std::shared_ptr<AbstractLQPNode>& lqp_root) const {
+  auto rule_was_applied_using_non_permanent_ucc = false;
```
dey4ss, line 77: quite a long name, maybe just use cacheable and add a comment?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -97,21 +99,25 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
...
-    if (group_by_columns.size() == node->node_expressions.size() &&
-        node->left_input()->has_matching_ucc(group_by_columns)) {
-      const auto& output_expressions = aggregate_node.output_expressions();
-      // Remove the AggregateNode if it does not limit or reorder the output expressions.
-      if (expressions_equal(output_expressions, node->left_input()->output_expressions())) {
-        lqp_remove_node(node);
+    if (group_by_columns.size() == node->node_expressions.size()) {
+      auto matching_ucc = node->left_input()->get_matching_ucc(group_by_columns);
+      if (matching_ucc.has_value()) {
+        rule_was_applied_using_non_permanent_ucc = !matching_ucc->is_permanent();
```
dey4ss, line 105: you assign this member inside `visit_lqp`. So when you rewrite an AggregateNode with "non-permanent" FD first and another one with "permanent" FD later, you have a wrong result. You should use `|=`, right?

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -167,6 +173,10 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
 
       const auto success = remove_dependent_group_by_columns(fd, aggregate_node, group_by_columns);
       if (success) {
+        // Functional dependencies are derived from UCCs. In case we encounter a non-permanent FD,
+        // this means we encountered an underlying non-permanent UCC as well.
+        rule_was_applied_using_non_permanent_ucc |= !fd.is_permanent();
```
dey4ss, line 178: here, you use `|=`

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -187,6 +197,8 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
 
     return LQPVisitation::VisitInputs;
   });
+
+  return rule_was_applied_using_non_permanent_ucc ? IsCacheable::No : IsCacheable::Yes;
```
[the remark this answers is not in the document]
dey4ss: absolutely right, conditional/ternary operator is perfectly fine.

src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
```diff
@@ -50,15 +51,16 @@ void gather_rewrite_info(
          "Neither column of the join predicate could be evaluated on the removable input.");
 
   // Check for uniqueness.
-  if (!removable_subtree->has_matching_ucc({exchangeable_column_expression})) {
+  if (!removable_subtree->get_matching_ucc({exchangeable_column_expression}).has_value()) {
```
dey4ss, line 54: has_value needed?

src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
```diff
@@ -50,15 +51,16 @@ void gather_rewrite_info(
...
     return;
   }
 
   // Now, we look for a predicate that can potentially be used in a subquery to replace the join. If we find an equals
   // predicate that filters on a UCC, a maximum of one tuple remains in the result relation. Since at this point, we al-
   // ready know the candidate join is basically a semi join, we can further transform the join to a single predicate
   // node filtering the join column for the value of the remaining tuple's join attribute.
-  visit_lqp(removable_subtree, [&removable_subtree, &rewrite_predicate](auto& current_node) {
+  visit_lqp(removable_subtree, [&removable_subtree, &rewrite_predicate,
```
dey4ss, line 62: just capture all `[&]` then?

src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
```diff
@@ -73,10 +75,11 @@ void gather_rewrite_info(
       return LQPVisitation::VisitInputs;
     }
 
-    // Only predicates in the form `column = value` are useful to our optimization. These conditions have the potential
-    // (given filtered column is a UCC) to emit at most one result tuple.
-    if (candidate_expression->predicate_condition != PredicateCondition::Equals)
+    // Only predicates in the form `column = value` are useful to our optimization. These conditions have the
+    // potential (given filtered column is a UCC) to emit at most one result tuple.
+    if (candidate_expression->predicate_condition != PredicateCondition::Equals) {
```
dey4ss, line 80: nice catch

src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
```diff
@@ -96,14 +99,18 @@ void gather_rewrite_info(
...
-    // column is unique on the current node is not sufficient. There could be unions or joins in between the subtree
+    // column is unique on the current node is not sufficient. There could be unions or joins between the subtree
     // root and the current node, invalidating the unique column combination.
-    if (!expression_evaluable_on_lqp(candidate_column_expression, *removable_subtree) ||
-        !removable_subtree->has_matching_ucc({candidate_column_expression})) {
+    if (!expression_evaluable_on_lqp(candidate_column_expression, *removable_subtree)) {
+      return LQPVisitation::VisitInputs;
+    }
+    const auto matching_ucc = removable_subtree->get_matching_ucc({candidate_column_expression});
+    if (!matching_ucc.has_value()) {
```
dey4ss, line 108: w/o has_value

src/lib/optimizer/strategy/join_to_semi_join_rule.cpp
```diff
@@ -72,20 +74,30 @@ void JoinToSemiJoinRule::_apply_to_plan_without_subqueries(const std::shared_ptr
...
-        join_node->join_mode = JoinMode::Semi;
-        const auto temp = join_node->left_input();
-        join_node->set_left_input(join_node->right_input());
-        join_node->set_right_input(temp);
-      } else if (*join_node->prunable_input_side() == LQPInputSide::Right &&
-                 join_node->right_input()->has_matching_ucc(equals_predicate_expressions_right)) {
-        join_node->join_mode = JoinMode::Semi;
+      if (*join_node->prunable_input_side() == LQPInputSide::Left) {
+        const auto matching_ucc = join_node->left_input()->get_matching_ucc(equals_predicate_expressions_left);
+        if (matching_ucc.has_value()) {
```
dey4ss, line 79: you guess it...

src/lib/optimizer/strategy/join_to_semi_join_rule.cpp
```diff
@@ -72,20 +74,30 @@ void JoinToSemiJoinRule::_apply_to_plan_without_subqueries(const std::shared_ptr
...
+          rule_was_applied_using_non_permanent_ucc = !matching_ucc->is_permanent();
+
+          join_node->join_mode = JoinMode::Semi;
+          const auto temp = join_node->left_input();
+          join_node->set_left_input(join_node->right_input());
+          join_node->set_right_input(temp);
+        }
+      } else if (*join_node->prunable_input_side() == LQPInputSide::Right) {
+        const auto matching_ucc = join_node->right_input()->get_matching_ucc(equals_predicate_expressions_right);
+        if (matching_ucc.has_value()) {
```
dey4ss, line 89: ... again

src/lib/optimizer/strategy/join_to_semi_join_rule.cpp
```diff
@@ -72,20 +74,30 @@ void JoinToSemiJoinRule::_apply_to_plan_without_subqueries(const std::shared_ptr
...
+
+          join_node->join_mode = JoinMode::Semi;
+        }
       }
     }
 
     return LQPVisitation::VisitInputs;
   });
+
+  return rule_was_applied_using_non_permanent_ucc ? IsCacheable::No : IsCacheable::Yes;
```
dey4ss, line 100: pick here: if you implemented operator `|=` for `IsCacheable` taking a bool as second param, you could simplify the rules a bit. Not sure if maybe the data dependencies should also just have a member of type `IsCacheable` rather than the `permanent` bool.

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -183,10 +178,12 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
     }
   }
 
+  bool pqp_cacheable = true;
```
dey4ss, line 181: do you need this variable? `get_optimized_logical_plan()` sets the `_optimized_logical_plan` member, which contains the info you need later.

src/lib/sql/sql_pipeline_statement.hpp
```diff
@@ -75,7 +75,7 @@ class SQLPipelineStatement : public Noncopyable {
   const SQLTranslationInfo& get_sql_translation_info();
 
   // Returns the optimized LQP for this statement.
-  const std::shared_ptr<AbstractLQPNode>& get_optimized_logical_plan();
+  const OptimizedLogicalQueryPlan get_optimized_logical_plan();
```
dey4ss, line 78: same comment as in optimizer: pair of plan and cacheability?

src/lib/storage/constraints/table_key_constraint.hpp
```diff
@@ -21,7 +21,8 @@ class TableKeyConstraint final : public AbstractTableConstraint {
    * voilating the set semantics of the constraint.
    */
   TableKeyConstraint(const std::set<ColumnID>& columns, const KeyConstraintType key_type);
-  TableKeyConstraint(const std::set<ColumnID>& columns, const KeyConstraintType key_type, const CommitID last_validated_on);
+  TableKeyConstraint(const std::set<ColumnID>& columns, const KeyConstraintType key_type,
+                     const CommitID last_validated_on);
```
dey4ss, line 25: default argument?

src/test/lib/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -332,27 +334,54 @@ TEST_F(StoredTableNodeTest, UniqueColumnCombinationsPrunedColumns) {
   EXPECT_TRUE(find_ucc_by_key_constraint(key_constraint_c, unique_column_combinations));
 }
 
+TEST_F(StoredTableNodeTest, UniqueColumnCombinationsValidityNotGuaranteed) {
```
dey4ss, line 337: comment on what this test does?

src/test/lib/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -332,27 +334,54 @@ TEST_F(StoredTableNodeTest, UniqueColumnCombinationsPrunedColumns) {
...
+TEST_F(StoredTableNodeTest, GetMatchingUniqueColumnCombination) {
   const auto key_constraint_a = TableKeyConstraint{{_a->original_column_id}, KeyConstraintType::UNIQUE};
   _table_a->add_soft_key_constraint(key_constraint_a);
   EXPECT_EQ(_stored_table_node->unique_column_combinations().size(), 1);
 
   // Negative test.
-  EXPECT_FALSE(_stored_table_node->has_matching_ucc({_b}));
-  EXPECT_FALSE(_stored_table_node->has_matching_ucc({_c}));
-  EXPECT_FALSE(_stored_table_node->has_matching_ucc({_b, _c}));
+  EXPECT_FALSE(_stored_table_node->get_matching_ucc({_b}).has_value());
```
dey4ss, line 375: I guess has_value is not required for any of these checks

src/test/lib/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -332,27 +334,54 @@ TEST_F(StoredTableNodeTest, UniqueColumnCombinationsPrunedColumns) {
...
 
   // Test exact match.
-  EXPECT_TRUE(_stored_table_node->has_matching_ucc({_a}));
+  EXPECT_TRUE(_stored_table_node->get_matching_ucc({_a}).has_value());
 
   // Test superset of column ids.
-  EXPECT_TRUE(_stored_table_node->has_matching_ucc({_a, _b}));
-  EXPECT_TRUE(_stored_table_node->has_matching_ucc({_a, _c}));
+  EXPECT_TRUE(_stored_table_node->get_matching_ucc({_a, _b}).has_value());
+  EXPECT_TRUE(_stored_table_node->get_matching_ucc({_a, _c}).has_value());
```
dey4ss, line 384: since you return the UCCs now: check for the exact UCC and not just their existence?

src/test/lib/optimizer/optimizer_test.cpp
```diff
@@ -331,4 +337,43 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
   }
 }
 
+TEST_F(OptimizerTest, CheckTrueCacheablility) {
```
dey4ss, line 340: i think this test name (as well as the following) are a nit misleading. Also, add a comment on what they test.

src/test/lib/optimizer/optimizer_test.cpp
```diff
@@ -331,4 +337,43 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
...
+TEST_F(OptimizerTest, CheckFalseCacheabilityOfOptimization) {
+  auto optimizer = Optimizer{};
+  optimizer.add_rule(std::make_unique<JoinToSemiJoinRule>());
+
+  auto column_definitions = TableColumnDefinitions{};
+  column_definitions.emplace_back("column0", DataType::Int, false);
+  const auto table = std::make_shared<Table>(column_definitions, TableType::Data, ChunkOffset{2}, UseMvcc::Yes);
+
+  auto& sm = Hyrise::get().storage_manager;
+  sm.add_table("table", table);
```
dey4ss, line 358: I'll pay for the increased storage cost if you rename the variable. Alternatively, as you don't use the variable a second time, just make it a oneliner.

src/test/lib/optimizer/optimizer_test.cpp
```diff
@@ -331,4 +337,43 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
...
+  auto optimizer = Optimizer::create_default_optimizer();
+
+  auto lqp = ProjectionNode::make(expression_vector(add_(b, subquery_a)),
+                                  PredicateNode::make(greater_than_(a, subquery_b), node_a));
+  const auto optimize_results = optimizer->optimize(std::move(lqp));
+  EXPECT_EQ(optimize_results.cacheable, true);
+}
+
+TEST_F(OptimizerTest, CheckFalseCacheabilityOfOptimization) {
+  auto optimizer = Optimizer{};
```
dey4ss, line 350: add a comment why it's not cacheable?

src/test/lib/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -446,4 +446,11 @@ TEST_F(BetweenCompositionTest, HandleMultipleEqualExpressions) {
   EXPECT_LQP_EQ(result_lqp, expected_lqp);
 }
 
+TEST_F(BetweenCompositionTest, CheckCacheability) {
+  const auto input_lqp = PredicateNode::make(equals_(_a_a, 100), PredicateNode::make(equals_(_a_b, 100), _node_a));
+  const auto lqp_result = StrategyBaseTest::apply_rule_with_cacheability_check(_rule, input_lqp);
```
dey4ss, line 451: guess we don't need `StrategyBaseTest::` prefix here?

src/test/lib/optimizer/strategy/chunk_pruning_rule_test.cpp
```diff
@@ -480,4 +480,14 @@ TEST_F(ChunkPruningRuleTest, PredicateWithCorrelatedSubquery) {
   EXPECT_EQ(stored_table_node_1->pruned_chunk_ids(), expected_chunk_ids);
 }
 
+TEST_F(ChunkPruningRuleTest, CheckCacheability) {
+  const auto stored_table_node = StoredTableNode::make("fixed_string_compressed");
+
+  const auto predicate_node = PredicateNode::make(equals_(lqp_column_(stored_table_node, ColumnID{0}), "zzz"));
+  predicate_node->set_left_input(stored_table_node);
+
+  const auto lqp_result = StrategyBaseTest::apply_rule_with_cacheability_check(_rule, predicate_node);
```
dey4ss, line 489: (see above)

src/test/lib/optimizer/strategy/expression_reduction_rule_test.cpp
```diff
@@ -308,4 +308,20 @@ TEST_F(ExpressionReductionRuleTest, ApplyToLQP) {
...
+  const auto stored_table_node = StoredTableNode::make("agg_table");
+
+  const auto col_a = lqp_column_(stored_table_node, ColumnID{0});
+  const auto col_b = lqp_column_(stored_table_node, ColumnID{1});
+  const auto lqp = AggregateNode::make(
+      expression_vector(), expression_vector(sum_(col_b), count_star_(stored_table_node), avg_(col_b)),  // NOLINT
+      stored_table_node);
+  const auto lqp_result = apply_rule_with_cacheability_check(rule, lqp);
+  const auto cacheable = lqp_result.cacheable;
+  EXPECT_EQ(cacheable, true);
```
dey4ss, line 324: `EXPECT_TRUE`

src/test/lib/optimizer/strategy/in_expression_rewrite_rule_test.cpp
```diff
@@ -373,4 +373,13 @@ TEST_F(InExpressionRewriteRuleTest, AutoStrategy) {
   }
 }
 
+TEST_F(InExpressionRewriteRuleTest, CheckCacheability) {
+  auto rule = std::make_shared<InExpressionRewriteRule>();
+  rule->strategy = InExpressionRewriteRule::Strategy::ExpressionEvaluator;
+  const auto input_lqp = PredicateNode::make(single_element_in_expression, node);
+  const auto lqp_result = StrategyBaseTest::apply_rule_with_cacheability_check(rule, input_lqp);
+  const auto cacheable = lqp_result.cacheable;
+  EXPECT_EQ(cacheable, true);
```
dey4ss, line 382: (see above, skip this comment for the next few occurrences)

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -274,17 +274,18 @@ bool AbstractLQPNode::is_column_nullable(const ColumnID column_id) const {
   return left_input()->is_column_nullable(column_id);
 }
 
-bool AbstractLQPNode::has_matching_ucc(const ExpressionUnorderedSet& expressions) const {
+std::optional<UniqueColumnCombination> AbstractLQPNode::get_matching_ucc(
```
dey4ss, line 277: We never really access the UCCs, right? We only care about their cacheability later, don't we? So maybe return `std::pair<bool, IsCacheable>` directly?

dey4ss (commented):
I really like your code and most comments are just minor remarks or ideas on design choices (and some code guidelines though). Did not really find functionality issues or bugs, nice work.

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -105,13 +105,19 @@ UniqueColumnCombinations StoredTableNode::unique_column_combinations() const {
       continue;
     }
 
+    // We may only use the key constraints as UCCs for optimization purposes if they are certainly still valid,
+    // otherwise these optimizations could produce invalid query results.
+    if (!table->constraint_guaranteed_to_be_valid(table_key_constraint)) {
+      continue;
+    }
```
dey4ss, line 112: You discard a possibly valid UCC in an edge case here: Imagine UCC was valid for CommitID 5. Our transaction runs on a snapshot w/ CommitID 7. Another transaction commited with CommitID 8 and inserted tuples that could possibly invalidate the UCC. However, the UCC is still valid for the snapshot that out transcation sees. We cannot derive this knowledge only using the max begin CID, so we should add a comment here.

dey4ss:
Put this on hold until end of August. We plan to incorporate this feature, but there is currently no capacity for it. Will work on it when back from the internship.

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -152,7 +152,7 @@ UniqueColumnCombinations AggregateNode::unique_column_combinations() const {
 
     // Make sure that we do not add an already existing or a superset UCC.
     if (unique_column_combinations.empty() ||
-        !contains_matching_unique_column_combination(unique_column_combinations, group_by_columns)) {
+        !ucc_cacheability_if_exists(unique_column_combinations, group_by_columns)) {
       unique_column_combinations.emplace(std::move(group_by_columns));
```
Rob2U: There is an edge case here: If the group_by_columns exactly matches a `unique_column_combination` that is not schema-given, the UCC remains 'spurious', but we should modify it to be schema-given.

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -188,14 +190,14 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
    *         documentation.
    */
   virtual UniqueColumnCombinations unique_column_combinations() const = 0;
-
   /**
-   * @return True if there is a unique column combination (UCC) matching the given subset of output expressions (i.e.,
+   * @return Searchs for a unique column combination (UCC) matching the given subset of output expressions (i.e.,
```
dey4ss: `Search`, `Searches`, or leave as it was before.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -174,6 +187,12 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
 
       const auto success = remove_dependent_group_by_columns(fd, aggregate_node, group_by_columns);
       if (success) {
+        // Functional dependencies are derived from UCCs. In case we encounter a non-schema-given FD, this means we
+        // encountered an underlying non-schema-given UCC as well.
```
dey4ss: The connection to UCCs is not necessary IMO. The fact that the FD is not genuine (for whatever reason) is enough.

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -104,21 +106,27 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
...
-      if (expressions_equal(output_expressions, node->left_input()->output_expressions())) {
-        lqp_remove_node(node);
+    if (group_by_columns.size() == node->node_expressions.size()) {
+      const auto [has_matching_ucc, ucc_cacheable] = node->left_input()->has_matching_ucc(group_by_columns);
+      if (has_matching_ucc) {
+        if (!ucc_cacheable) {
+          optimization_context.set_not_cacheable();
+        }
+
+        const auto& output_expressions = aggregate_node.output_expressions();
```
dey4ss:
```suggestion
        const auto output_expressions = aggregate_node.output_expressions();
```

src/lib/optimizer/strategy/dependent_group_by_reduction_rule.cpp
```diff
@@ -159,6 +167,11 @@ void DependentGroupByReductionRule::_apply_to_plan_without_subqueries(
       // library implementation details). Thus, we compare the ColumnIDs of the determinants.
       const auto& left_column_ids = get_column_ids(fd_left.determinants);
       const auto& right_column_ids = get_column_ids(fd_right.determinants);
+
+      if (fd_left.is_schema_given() != fd_right.is_schema_given()) {
+        return fd_left.is_schema_given();
+      }
```
dey4ss: move below size comparison of determinants and add a comment (can as well be over the sort additional to the sentence describing whye sort by determinant size).

src/test/lib/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -424,12 +428,41 @@ TEST_F(StoredTableNodeTest, UniqueColumnCombinationsPrunedColumns) {
...
+  const auto key_constraint_a_b = TableKeyConstraint{{ColumnID{0}, ColumnID{1}}, KeyConstraintType::UNIQUE};
+  const auto key_constraint_c = TableKeyConstraint{{ColumnID{2}}, KeyConstraintType::UNIQUE, INITIAL_COMMIT_ID};
+  _table_a->add_soft_constraint(key_constraint_a_b);
+  _table_a->add_soft_constraint(key_constraint_c);
+  const auto& table_key_constraints = _table_a->soft_key_constraints();
+  EXPECT_EQ(table_key_constraints.size(), 2);
+  EXPECT_EQ(_stored_table_node->unique_column_combinations().size(), 2);
+
+  // Modify the table so that the UCC is no longer guaranteed to be valid
+  // (in fact, it is actually no longer valid because we duplicated all rows).
```
dey4ss: please fill 120c chars

src/test/lib/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -424,12 +428,41 @@ TEST_F(StoredTableNodeTest, UniqueColumnCombinationsPrunedColumns) {
   EXPECT_TRUE(find_ucc_by_key_constraint(key_constraint_c, unique_column_combinations));
 }
 
+TEST_F(StoredTableNodeTest, UniqueColumnCombinationsValidityNotGuaranteed) {
+  // Check that only the key constraints are used as UCCs for optimization purposes that are guaranteed to be valid.
```
dey4ss: "guaranteed to be valid for the current snapshot"? Otherwise, it sounds like we would only allow schema-defined UCCs.

src/test/lib/logical_query_plan/stored_table_node_test.cpp
```diff
@@ -451,11 +484,18 @@ TEST_F(StoredTableNodeTest, HasMatchingUniqueColumnCombination) {
 
   // Test exact match.
   _stored_table_node->set_pruned_column_ids({});
-  EXPECT_TRUE(_stored_table_node->has_matching_ucc({_a}));
+  const auto [has_matching_ucc_a, matching_ucc_cacheable_a] = _stored_table_node->has_matching_ucc({_a});
+  EXPECT_TRUE(has_matching_ucc_a);
+  EXPECT_EQ(matching_ucc_cacheable_a, true);
```
dey4ss: pick here (also below): `EXPECT_TRUE`

src/test/lib/optimizer/strategy/dependent_group_by_reduction_rule_test.cpp
```diff
@@ -120,7 +136,8 @@ TEST_F(DependentGroupByReductionRuleTest, SingleKeyReduction) {
 
     _apply_rule(rule, _lqp);
 
-    EXPECT_TRUE(_optimization_context.is_cacheable());
+    EXPECT_FALSE(
+        _optimization_context.is_cacheable());  // Not cacheable because FD derived from non-schema-given UCC was used.
```
dey4ss: pick here: move comment above code? Having it in the same line is not beneficial anymore if it foreces a linebreak

src/lib/logical_query_plan/lqp_utils.cpp
```diff
@@ -508,23 +509,18 @@ std::vector<std::shared_ptr<AbstractExpression>> get_expressions_for_column_ids(
...
-    }
-  }
-  // Did not find a UCC for the given expressions.
-  return false;
+  return std::find_if(unique_column_combinations.begin(), unique_column_combinations.end(), [&](auto& ucc) {
+    return ucc.expressions.size() <= expressions.size() &&
+           std::all_of(ucc.expressions.cbegin(), ucc.expressions.cend(), [&](const auto& ucc_expression) {
+             return expressions.contains(ucc_expression);
+           });
+  });
```
dey4ss: As discussed: Please prefer genuine over spurious dependencies

src/lib/logical_query_plan/data_dependencies/functional_dependency.hpp
```diff
@@ -28,16 +28,27 @@ namespace hyrise {
...
   bool operator!=(const FunctionalDependency& other) const;
   size_t hash() const;
 
   ExpressionUnorderedSet determinants;
   ExpressionUnorderedSet dependents;
+
+  bool is_schema_given() const;
+
+ private:
+  bool _is_schema_given;
```
dey4ss: as discussed: Please rename and make mutable

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -151,9 +152,19 @@ UniqueColumnCombinations AggregateNode::unique_column_combinations() const {
...
+    } else {
+      const auto existing_ucc = find_ucc(unique_column_combinations, group_by_columns);
+      if (existing_ucc == unique_column_combinations.end()) {
+        unique_column_combinations.emplace(std::move(group_by_columns), /*is_schema_given=*/true);
+      } else if (existing_ucc != unique_column_combinations.end() &&
+                 existing_ucc->expressions.size() == group_by_columns_count) {
+        // If we already have a UCC for the group-by columns that is an exact match, we do not need to add it again.
+        // However, we should still set it as schema-given, as the spurious UCC is not needed anymore.
+        existing_ucc->set_schema_given();
+      }
```
dey4ss:
```suggestion
      const auto existing_ucc = find_ucc(unique_column_combinations, group_by_columns);
      if (existing_ucc == unique_column_combinations.end() || !existing_ucc.is_schema_given()) {
        unique_column_combinations.emplace(std::move(group_by_columns), /*is_schema_given=*/true).first->set_schema_given();
      }
```

dey4ss: (please also comment with what we discussed)

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -188,14 +190,14 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
...
   virtual UniqueColumnCombinations unique_column_combinations() const = 0;
-
   /**
-   * @return True if there is a unique column combination (UCC) matching the given subset of output expressions (i.e.,
+   * @return Searches for a unique column combination (UCC) matching the given subset of output expressions (i.e.,
    *         the rows are guaranteed to be unique). This is preferred over calling
-   *         contains_matching_unique_column_combination(unique_column_combinations(), ...) as it performs additional
-   *         sanity checks.
+   *         `find_ucc(unique_column_combinations(), ...)` as it performs additional
+   *         sanity checks. Provides a pair<bool, bool> indicating whether a UCC was found and whether it is
```
Bouncner, line 197: First file I look at, so note for later: do we want to second bool to be an boolean enum?

src/lib/logical_query_plan/aggregate_node.cpp
```diff
@@ -144,16 +145,17 @@ UniqueColumnCombinations AggregateNode::unique_column_combinations() const {
...
-    if (unique_column_combinations.empty() ||
-        !contains_matching_unique_column_combination(unique_column_combinations, group_by_columns)) {
-      unique_column_combinations.emplace(std::move(group_by_columns));
+    auto group_by_columns =
+        ExpressionUnorderedSet{node_expressions.begin(), node_expressions.begin() + group_by_columns_count};
+
+    const auto [existing_ucc, inserted] =
+        unique_column_combinations.emplace(std::move(group_by_columns), /*is_genuine=*/true);
+
+    // If the UCC was already in the set but is not genuine, we set it to genuine.
```
Bouncner: Short comment why we know for sure that the UCC is genuine?

src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
```diff
@@ -95,40 +97,64 @@ FunctionalDependencies deflate_fds(const FunctionalDependencies& fds) {
...
   }
 
-  // We cannot use a set here as we want to add dependents to existing FDs and objects in sets are immutable.
-  auto existing_fds = std::vector<std::pair<ExpressionUnorderedSet, ExpressionUnorderedSet>>{};
-  existing_fds.reserve(fds.size());
+  const auto hash_pair = [](const auto& key) {
+    auto hash = size_t{0};
+    boost::hash_combine(hash, key.first.size());
+    for (const auto& expression : key.first) {
+      hash ^= expression->hash();
```
Bouncner, line 102: What are you doing here? Can you add a comment?

src/test/lib/optimizer/strategy/join_to_semi_join_rule_test.cpp
```diff
@@ -167,9 +169,8 @@ TEST_F(JoinToSemiJoinRuleTest, MultiPredicateInnerJoinToSemiJoinWithMultiEqui) {
 
   static_cast<JoinNode&>(*_lqp->left_input()).mark_input_side_as_prunable(LQPInputSide::Right);
   _apply_rule(rule, _lqp);
-  // Not cacheable because UCC used is not permanent. Non-cacheability is added in #2600. This is why it is still
-  // cacheable.
-  EXPECT_TRUE(_optimization_context.is_cacheable());
+  // Not cacheable because UCC used is not permanent.
```
Bouncner: perm > genuine?

src/test/lib/optimizer/strategy/join_to_semi_join_rule_test.cpp
```diff
@@ -142,7 +142,9 @@ TEST_F(JoinToSemiJoinRuleTest, MultiPredicateInnerJoinToSemiJoinWithMultiEqui) {
     auto& sm = Hyrise::get().storage_manager;
     sm.add_table("table", table);
 
-    table->add_soft_constraint(TableKeyConstraint{{ColumnID{0}, ColumnID{1}}, KeyConstraintType::UNIQUE});
+    // Non-genuine UCC
```
Bouncner: . ?

src/test/lib/optimizer/strategy/join_to_semi_join_rule_test.cpp
```diff
@@ -120,9 +121,8 @@ TEST_F(JoinToSemiJoinRuleTest, MultiPredicateInnerJoinToSemiJoinWithSingleEqui)
 
   static_cast<JoinNode&>(*_lqp->left_input()).mark_input_side_as_prunable(LQPInputSide::Right);
   _apply_rule(rule, _lqp);
-  // Not cacheable because UCC used is not permanent. Non-cacheability is added in #2600. This is why it is still
-  // cacheable.
-  EXPECT_TRUE(_optimization_context.is_cacheable());
+  // Not cacheable because UCC used is not permanent.
```
Bouncner: genuine?

src/test/lib/logical_query_plan/data_dependencies/functional_dependency_test.cpp
```diff
@@ -86,7 +89,38 @@ TEST_F(FunctionalDependencyTest, InflateFDs) {
...
 
-TEST_F(FunctionalDependencyTest, DeflateFDs) {
+TEST_F(FunctionalDependencyTest, InflateFDsGenuineAndSpurious) {
+  const auto fd_a_bc = FunctionalDependency({_a}, {_b, _c}, /*is_genuine=*/true);
+  const auto fd_a_bc2 = FunctionalDependency({_a}, {_b, _c, _x}, /*is_genuine=*/false);
+  const auto fd_a_b = FunctionalDependency({_a}, {_b}, /*is_genuine=*/true);
+  const auto fd_a_c = FunctionalDependency({_a}, {_c}, /*is_genuine=*/true);
+  const auto fd_a_x = FunctionalDependency({_a}, {_x}, /*is_genuine=*/false);
+
+  const auto fd_b_ac = FunctionalDependency({_b}, {_a, _c}, /*is_genuine=*/false);
```
Bouncner, line 99: I get the idea of having those comments, but wouldn't a boolean enum a better choice here?

Bouncner (commented):
Did a quick pass. Is there anything I should have a look at in particular?

And can you please search for "permanent" in the code base and check if "genuine" is applicable?

src/test/lib/optimizer/strategy/join_to_semi_join_rule_test.cpp
dey4ss: Changed to:
```c++
// Add a spurious UCC. Thus, the rewritten plan should not be cacheable.
```

src/lib/logical_query_plan/aggregate_node.cpp
dey4ss: added `[...] because columns in the GROUP BY clause must be unique.`

src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
dey4ss: Added
```c++
// Manually combine the hashes with XOR to make the result independent of the expressions' order.
```

src/lib/logical_query_plan/abstract_lqp_node.hpp
dey4ss: We could, but then we always have to do one of the following in the caller code:
```c++
const auto [has_matching_ucc, is_cacheable] = lqp_node->has_matching_ucc(...);

// Option 1 (literally testing bool == true).
if (is_cacheable == IsCacheable::Yes) {
  ...
}

// Option 2 (casting bool as bool).
if (static_cast<bool>(is_cacheable)) {
  ...
}
```

instead of just 
```c++
if (is_cacheable) {
  ...
}
```

Same goes for the `is_genuine` flag of dependencies. Though it's very explicit, it does not really ease readability when used in conditions IMO. In the end, we can substitute every bool with an enum. Not really sure where to start and where to end.

src/test/lib/logical_query_plan/data_dependencies/functional_dependency_test.cpp
dey4ss: (see above)

src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
Bouncner: Can you add that XOR is associative?

Bouncner:
Okay, anything else I should have a look at? Otherwise zünd?

src/lib/logical_query_plan/data_dependencies/functional_dependency.cpp
dey4ss: *commutative

PR #2548, opened 2023-02-18
Scheduler: use different queue and semaphores for improved worker sleeping

Bouncner:
This is the first of probably three PRs that adapt our scheduling, mostly to improve multi-threading (MT) performance.

**Changes:**
 - use of the [concurrentqueue](https://github.com/cameron314/concurrentqueue) library instead of TBBs queue
   - there are three reasons to switch the queue: (i) one part less relying on TBB (we last allocator issue was quite a pain and maintaining specific TBB versions is as well), (ii) the performance of the concurrentqueue is better than TBB's (see their website, althought our tasks usually have a granularity where the queue is never a bottleneck, (iii) the queue provides "lightweight semaphores"
 - usage of semaphores
   - we currently use mutexes and conditional variable to signal threads that new tasks arrived
   - mutexes are used to secure critical sections, but pulling tasks from a concurrent queue is not critical
   - semaphores are more lightweight and faster, as waking up tasks don't race for the mutex
   - (I guess the reason to use mutexes was (i) Hyrise wasn't MT-optimized back then (well, it's still not really) and nobody noticed the issue and (ii) C++20 just started to implement the first (still very limited) semaphores)
 - energy consumption
   - the current master implements workers that sleep for 300 microseconds and wake up afterwards. There have been several parts of our scheduler, where these wake ups have been necessary (waking up all workers can be a little bit of pain)
   - as a result, each worker woke up ~3000 times a second, on a 2x `AMD EPYC 7742 64-Core Processor` server, that leads to a steady CPU usage of over 400% for an **idle** server
   - this PR changes this behaviour and uses 0% CPU when idling
   - I leave the exercise of many hours of idling Hyrise servers accumulate to a flight from NY to SF to the interested readers
   - with this PR, Hyrise continues to further accelerate the HPIs commitment towards the 13th goal for **Clean IT** of the _UN Sustainable Development Goals_
 - I addressed a race in a DebugAssert in `abstract_operator` when the output expressions have been read concurrently (@dey4ss: we probably need to discuss this one)
   - First I thought it's an issue triggered by the (better?) scheduling with this PR, but other current PRs have also failed in the "TestParallelConnections" CI test


**Performance:**
Performance is improved for multi-threaded executions (single-threaded execution should be touched by this PR, perf. changes should be noise here). But since the queue is usually not the main bottleneck, changes are only significant for TPC-C with its small jobs. The remaining benchmarks improve a bit.

[...]

PR #2603, opened 2023-08-12
Revert "Scheduler: use different queue and semaphores for improved worker sleeping"

Bouncner:
Reverts hyrise/hyrise#2548.

Wasn't ready. New CI crashed four times, need to take a deeper look.


Bouncner:
Hmm, now that the "new CI" processes more PRs, the error can be also seen in other branches.

PR #2604, opened 2023-08-15
Align TPC-C's Stock-Level with the standard's example procedure

### LINKED ISSUE #2602: Stock Level transaction in  TPC-C
### Transaction processing
In the stock level transaction, we currently obtain a list of items inefficiently by first fetching the list and then putting this list into an `IN()` selection in another query.
Instead, we should use a simple join to fetch the list and use it for filtering as mentioned in the standard and pytpc (https://github.com/apavlo/py-tpcc/blob/master/pytpcc/drivers/sqlitedriver.py).

### ~Visualization~
~When visualizing queries of TPC-C, no stock level plans are visualized.~

dey4ss:
closes #2602


Bouncner (approved):
LGTM.

dey4ss:
@Bouncner still good?

Bouncner (approved):
Still good.

PR #2605, opened 2023-08-16
[DYOD] Add generic implementation of Parallel Inplace Merge Sort

niklasmohrin:
These changes are extracted into a separate PR from our project work on the window function operator, as discussed with supervisors. The main PR will be opened later :)

I broke down the changes into three clean commits, so that you can keep them if you want, but feel free to squash anyways if you feel like it.

Since this uses `span` (and other C++20 features) again, gcc-9 doesn't work. Should I also include the commit updating it here, or open another PR, or do something completely different?

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -4,13 +4,13 @@
 | ------------------------- | ---------------- | -------- | ------------------------------------- |
 | autoconf                  | >= 2.69          |    All   |                                    No |
 | boost                     | >= 1.70.0        |    All   |                                    No |
-| clang                     | >= 11.0          |    All   |                 Yes, if gcc installed |
-| clang-format              | >= 11.0          |    All   |                      Yes (formatting) |
-| clang-tidy                | >= 11.0          |    All   |                         Yes (linting) |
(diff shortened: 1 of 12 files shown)


Bouncner:
Given this list (https://en.cppreference.com/w/cpp/compiler_support), GCC 10 should at least support `std::span`. Do you know if we need more than GCC 10?

niklasmohrin:
In our intermediate review PR, we got it to get past the `span` error, but then it gcc-ed on us again with the "control reaches end of non-void function"; I can include the gcc-10 commit here and we can see what happens? (if it works, I can also open another PR and rebase this one)

niklasmohrin:
Okay, works with gcc-10, but somehow clang-11 gives the error `member reference base type 'std::span' is not a structure or union` when trying to use `std::span(something).subspan(anotherthing)`. Not sure what is going on, the same line seems to work on its own on https://godbolt.org/z/nxnzabxcr :thinking: 

The full file starts working with clang 13 and breaks for different reasons in clang 11 and 12 on godbolt: https://godbolt.org/z/qdvq61WWs

tjeyy:

PR #2607, opened 2023-08-17
[DYOD] Support window function queries

niklasmohrin:
For intermediate review, see #2597. Additionally, we factored out the implementation of the parallel inplace merge sort into #2605.

- [x] check for `nullptr` chunks
- [x] investigate slow sorting for query 51 (it seems that we are (de-)allocating too many vectors and strings)
- [x] implement and benchmark different fan-outs / -ins for `partition_and_sort` (#2605)
- [x] increase test coverage / cover everything except unreachable code (not sure where we stand currently) (missing: reverse(equal) and deep copy)
- [x] better operator description in query plans
- [x] more documentation
- [x] change constructor to not take the `WindowFunctionExpression`

Possible follow up work:
- [ ] benchmark against another implementation
- [ ] reduce allocations in general (make `HashPartitionedData` only have one vector and save offsets for hash buckets?)
- [ ] improve performance for string columns, for example by avoiding copies (something like `AllTypeVariantView`?)

Furthermore: Currently coverage says "warning: 100 functions have mismatched data"


tjeyy:

PR #2610, opened 2023-08-24
Fix scheduler issues on NUMA servers

Bouncner:
Due to the scheduler changes introduced with #2548, we triggered a situation where tasks have not been processed and Hyrise hang with 0% CPU usage. Turns out Hyrise creates TaskQueues for all NUMA nodes even though there might be no local workers to process these tasks (e.g., when using `numactl -n 3`).

This is a first step (more might be needed) to address such issues. With this PR, we only create queues for nodes that have workers and thus do no longer put tasks on empty queues when these nodes do not have any workers.

PR #2614, opened 2023-09-22
Add function prototypes in third-party DB generators

dey4ss:
Clang-16 complained about missing prototypes in the TPC-DS and TPC-H data generators (deprecated in C2x). This PR adds them if necessary. Furthermore, the `experimental/functional` header is not included anymore in the like matcher. The experimental version is not provided by clang-17 (libc++) anymore, leading to a compilation error (and all CI stages use compilers where the experimental header is not necessary).

Includes hyrise-mp/tpcds-kit#1.

### DIFF
--- a/src/lib/expression/evaluation/like_matcher.hpp
+++ b/src/lib/expression/evaluation/like_matcher.hpp
@@ -1,9 +1,13 @@
 #pragma once
 
-#include <experimental/functional>
+#include <functional>
 #include <regex>
 #include <utility>
 #include <variant>
+
+#ifndef __cpp_lib_boyer_moore_searcher
+#include <experimental/functional>
+#endif
 
 #include "types.hpp"
 
--- a/src/lib/expression/evaluation/like_matcher.hpp
+++ b/src/lib/expression/evaluation/like_matcher.hpp
@@ -5,10 +5,6 @@
 #include <utility>
 #include <variant>
 
-#ifndef __cpp_lib_boyer_moore_searcher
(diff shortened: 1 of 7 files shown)


Bouncner (approved):
Nice work!

src/lib/expression/evaluation/like_matcher.hpp
```diff
@@ -1,10 +1,14 @@
 #pragma once
 
-#include <experimental/functional>
+#include <functional>
 #include <regex>
 #include <utility>
 #include <variant>
```
Bouncner, line 7: Add a todo to re-check when we update GCC and Clang?
Do you know which compiler complains?

dey4ss: iirc GCC-9 did not have the matcher in the `functional` header yet. But I can double-check and everything might be fine if we update GCC anyways (so we don't need the `exerimental` include/feature check at all then). However, it was clang-17 on macOS (so libc++ 17) that does not have the `experimental` header anymore.

dey4ss: Nope, GCC9 works as well. Removed the whole thing.

dey4ss:
@Bouncner ported the pexpect server timeout from #2610 since that PR seems to still need a while. 🧨 then?

Bouncner:
> @Bouncner ported the pexpect server timeout from #2610 since that PR seems to still need a while. 🧨 then?

💣🕯️❗️

PR #2615, opened 2023-09-25
Use TPC-C in Console test

dey4ss:
We encountered several timeout issues with TPC-DS generation in the Console test. This PR replaces TPC-DS with TPC-C, hoping for faster generation times.

Furthermore, the SSB test is now moved to the system test stage, where it should not be executed in parallel.

Finally, this PR refactors range filters and their tests a bit, adding pruning with between exclusive predicates.

### DIFF
--- a/src/lib/statistics/statistics_objects/range_filter.cpp
+++ b/src/lib/statistics/statistics_objects/range_filter.cpp
@@ -78,7 +78,7 @@ std::shared_ptr<AbstractStatisticsObject> RangeFilter<T>::sliced(
       if (value <= iter->first) {
         sliced_ranges.emplace_back(*iter);
       } else {
-        sliced_ranges.emplace_back(std::pair<T, T>{value, iter->second});
+        sliced_ranges.emplace_back(value, iter->second);
       }
       ++iter;
 
@@ -149,10 +149,11 @@ std::unique_ptr<RangeFilter<T>> RangeFilter<T>::build_filter(const pmr_vector<T>
    * std::make_unsigned<T>::type would be possible to use for signed int types, but not for floating types.
    * Approach: take the min and max values and simply check if the distance between both might overflow.
    */
+  static_assert(std::is_signed_v<T>, "Expected a signed arithmetic type.");
   DebugAssert(std::is_sorted(dictionary.cbegin(), dictionary.cend()), "Dictionary must be sorted in ascending order.");
   const auto min = dictionary.front();
   const auto max = dictionary.back();
-  if ((min < 0) && (max > std::numeric_limits<T>::max() + min)) {
+  if ((min < 0) && (max >= std::numeric_limits<T>::max() + min)) {
     return std::make_unique<RangeFilter<T>>(std::vector<std::pair<T, T>>{{min, max}});
   }
 
--- a/src/lib/statistics/statistics_objects/range_filter.cpp
+++ b/src/lib/statistics/statistics_objects/range_filter.cpp
@@ -49,17 +49,15 @@ std::shared_ptr<AbstractStatisticsObject> RangeFilter<T>::sliced(
       return std::make_shared<MinMaxFilter<T>>(value, value);
     case PredicateCondition::LessThan:
     case PredicateCondition::LessThanEquals: {
-      auto end_iter =
+      const auto end_iter =
           std::lower_bound(ranges.cbegin(), ranges.cend(), value,
-                           [](const auto& range, const auto& search_value) { return range.second < search_value; });
+                           [](const auto& range, const auto search_value) { return range.second < search_value; });
 
       // Copy all the ranges before the value.
       auto iter = ranges.cbegin();
       for (; iter != end_iter; ++iter) {
         sliced_ranges.emplace_back(*iter);
       }
-
-      DebugAssert(iter != ranges.cend(), "does_not_contain() should have caught that.");
 
       // If value is not in a gap, limit the last range's upper bound to value.
       if (value >= iter->first) {
@@ -70,8 +68,7 @@ std::shared_ptr<AbstractStatisticsObject> RangeFilter<T>::sliced(
     case PredicateCondition::GreaterThanEquals: {
       auto iter =
           std::lower_bound(ranges.cbegin(), ranges.cend(), value,
-                           [](const auto& range, const auto& search_value) { return range.second < search_value; });
-
+                           [](const auto& range, const auto search_value) { return range.second < search_value; });
       DebugAssert(iter != ranges.cend(), "does_not_contain() should have caught that.");
 
       // If value is in a gap, use the next range, otherwise limit the next range's upper bound to value.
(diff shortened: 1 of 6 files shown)


src/lib/statistics/statistics_objects/range_filter.cpp
```diff
@@ -130,30 +130,30 @@ std::unique_ptr<RangeFilter<T>> RangeFilter<T>::build_filter(const pmr_vector<T>
...
-    return std::make_unique<RangeFilter<T>>(std::vector<std::pair<T, T>>{{*min_max.first, *min_max.second}});
+  /**
+   * The code to determine the range boundaries requires calculating the distances between values. We cannot express the
+   * distance between -DBL_MAX and DBL_MAX using any of the standard data types. For these cases, the RangeFilter
+   * effectively degrades to a MinMaxFilter (i.e., stores only a single range).
+   * While being rather unlikely for doubles, it's more likely to happen when Hyrise includes tinyint etc.
+   * std::make_unsigned<T>::type would be possible to use for signed int types, but not for floating types.
+   * Approach: take the min and max values and simply check if the distance between both might overflow.
+   */
+  DebugAssert(std::is_sorted(dictionary.cbegin(), dictionary.cend()), "Dictionary must be sorted in ascending order.");
```
Bouncner, line 152: I'd assume it's already checked before that the dict is sorted?

src/lib/statistics/statistics_objects/range_filter.cpp
```diff
@@ -130,30 +130,30 @@ std::unique_ptr<RangeFilter<T>> RangeFilter<T>::build_filter(const pmr_vector<T>
...
+  const auto min = dictionary.front();
+  const auto max = dictionary.back();
+  if ((min < 0) && (max > std::numeric_limits<T>::max() + min)) {
+    return std::make_unique<RangeFilter<T>>(std::vector<std::pair<T, T>>{{min, max}});
```
Bouncner, line 156: Too lazy to look for tests: is this tested?

src/lib/statistics/statistics_objects/range_filter.cpp
```diff
@@ -80,10 +80,10 @@ std::shared_ptr<AbstractStatisticsObject> RangeFilter<T>::sliced(
       } else {
         sliced_ranges.emplace_back(std::pair<T, T>{value, iter->second});
```
Bouncner, line 81: Can you try `emplace_back(value, iter->second)`?

src/lib/statistics/statistics_objects/range_filter.cpp
Bouncner, line 156: `static_assert(std::is_signed<t>)` ?

Bouncner (commented):
Few questions, looking good.

src/lib/statistics/statistics_objects/range_filter.cpp
dey4ss: Not necessarily here: Called from the following function:
https://github.com/hyrise/hyrise/blob/f96c18f3791ea82499360c11a44f442b34390147/src/lib/statistics/generate_pruning_statistics.cpp#L26-L38

The dictionary can either be the actual dictionary of a segment or is created on the fly:
https://github.com/hyrise/hyrise/blob/f96c18f3791ea82499360c11a44f442b34390147/src/lib/statistics/generate_pruning_statistics.cpp#L26-L38

So no assertion before.

src/lib/statistics/statistics_objects/range_filter.cpp
dey4ss: is this really required or not already captured by `min < 0`?

Bouncner: True, but the assert would be explicit.

Do what you like.

dey4ss: Is this condition right though (think you stumbled across this as well)?

```c++
template<typename T>
void print_condition() {
  const auto min = std::numeric_limits<T>::min();
  const auto max = std::numeric_limits<T>::max();
  std::cout << std::boolalpha << (max > std::numeric_limits<T>::max() + min) << std::endl;
}

int main() {
  std::cout << "int32_t ";
  print_condition<int32_t>();
  std::cout << "int64_t ";
  print_condition<int64_t>();
  std::cout << "  float ";
  print_condition<float>();
  std::cout << " double ";
  print_condition<double>();
  return 0;
}
```

gets me:
```
int32_t true
int64_t true
  float false
 double false
```
Thus, no early out for floating-point types, which seems incorrect.

src/lib/statistics/statistics_objects/range_filter.cpp
```diff
@@ -270,43 +254,55 @@ bool RangeFilter<T>::does_not_contain(const PredicateCondition predicate_conditi
...
-      const auto start_lower = std::lower_bound(ranges.cbegin(), ranges.cend(), value, range_comp);
-      const auto end_lower = std::lower_bound(ranges.cbegin(), ranges.cend(), value2, range_comp);
+      // Find the range that starts at an element >= value2.
+      const auto upper_bound_range =
+          std::lower_bound(ranges.cbegin(), ranges.cend(), value2,
+                           [](const auto& range, const auto compare_value) { return range.first < compare_value; });
+
+      // If we find such a range, check for inclusive case.
+      auto upper_bound_in_gap =
+          upper_bound_range == ranges.cend() || !is_upper_inclusive || upper_bound_range->first > value2;
```
Bouncner, line 294: Why does `is_upper_inclusive` being false lead to `upper_bound_in_gap` being true?

src/lib/statistics/statistics_objects/range_filter.cpp
```diff
@@ -17,7 +17,7 @@ namespace hyrise {
 template <typename T>
 RangeFilter<T>::RangeFilter(std::vector<std::pair<T, T>> init_ranges)
     : AbstractStatisticsObject(data_type_from_type<T>()), ranges(std::move(init_ranges)) {
-  DebugAssert(!ranges.empty(), "Cannot construct empty RangeFilter");
+  DebugAssert(!ranges.empty(), "Cannot construct empty RangeFilter.");
```
Bouncner, line 20: Can you include `utils/assert.hpp` and `types.hpp`? :)

src/lib/statistics/statistics_objects/range_filter.cpp
```diff
@@ -270,43 +254,55 @@ bool RangeFilter<T>::does_not_contain(const PredicateCondition predicate_conditi
...
+      //    (i) Both bounds are "outside" (not spanning) the segment's value range (i.e., either both are smaller than
+      //        the minimum or both are larger than the maximum).
+      //   (ii) Both bounds are within the same gap.
 
-      Assert(variant_value2, "Between operator needs two values.");
+      Assert(variant_value2, "Between predicate needs two values.");
       const auto value2 = boost::get<T>(*variant_value2);
 
-      // a BETWEEN 5 AND 4 will always be empty
+      // a BETWEEN 5 AND 4 will always be empty.
```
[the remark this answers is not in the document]
dey4ss: SQL92 says:
> 6) "X BETWEEN Y AND Z" is equivalent to "X>=Y AND X<=Z".

So no matches by definition.

src/lib/statistics/statistics_objects/range_filter.cpp
dey4ss: I'm gonna add some more comments (maybe a running example), but the idea is the following: With `lower_bound`, we find the first range $r_i$ s.t. the range min >= value2. If the predicate is not upper inclusive, value2 is always in the gap brtween $r_{i-1}$ and $r_i$. Otherwise (predicate is inclusive), we need to check if value2 > $\min r_i$ to see if it is in the gap.

dey4ss: Is the explanation with the example sufficient?

Bouncner: Yes, sounds good.

src/lib/statistics/statistics_objects/range_filter.cpp
```diff
@@ -270,43 +255,68 @@ bool RangeFilter<T>::does_not_contain(const PredicateCondition predicate_conditi
...
-      const auto start_lower = std::lower_bound(ranges.cbegin(), ranges.cend(), value, range_comp);
-      const auto end_lower = std::lower_bound(ranges.cbegin(), ranges.cend(), value2, range_comp);
-
-      const bool start_in_value_range =
-          (start_lower != ranges.cend()) && (*start_lower).first <= value && value <= (*start_lower).second;
-      const bool end_in_value_range =
-          (end_lower != ranges.cend()) && (*end_lower).first <= value2 && value2 <= (*end_lower).second;
+      // Case (iii): The predicate can match the segment's values. However, the predicate values can be exactly in a gap
+      //             between two ranges (e.g., a BETWEEN 5 AND 6).
+      //             We know that y >(=) 2. Thus, we find the range that starts at an element >= y. If we cannot find
```
Bouncner, line 295: I always stumble over "an element" here. Maybe "find the range whose minimum is >= y" ?

Bouncner (approved):
Just the note concerning "element". Otherwise good to go.

PR #2618, opened 2023-10-12
[Buffer Manager] Add PageID and Frame primitives for the buffer manager

nikriek:
As explained in our discussion https://github.com/hyrise/hyrise/discussions/2617, we add the basic primitives for the `PageID` and `Frame`. `PageID` is used for the addressing of pages in a buffer manager. `Frame` stores the state of a page and also helps with synchronization.


src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+                                                              Frame::StateVersionType new_state) {
+  constexpr auto SHIFT = NUM_BITS - STATE_SHIFT;
+  static_assert(SHIFT == 16, "Shift must be 8.");
+  return ((old_version_and_state << SHIFT) >> SHIFT) | (new_state << STATE_SHIFT);
+}
+
+Frame::StateVersionType Frame::update_state_with_increment_version(Frame::StateVersionType old_version_and_state,
+                                                                   Frame::StateVersionType new_state) {
+  constexpr auto SHIFT = NUM_BITS - STATE_SHIFT;
+  static_assert(SHIFT == 16, "Shift must be 8.");
```
mweisgut, line 130: When defining `NUM_BITS`, you use `sizeof(StateVersionType) * CHAR_BIT`, `STATE_MASK       = 0xFFFF000000000000`, and `STATE_SHIFT = std::countr_zero(STATE_MASK);`. Since the STATE_MASK is hardcoded and you ensure SHIFT being 16 here, do we have any benefit in not hard-coding `NUM_BITS`?

mweisgut (changes requested):
Just a partial review so far. I must look further into `frame.cpp` and the tests. Since this is not a trivial project, more documentation in the header files might help to understand the code.

src/lib/storage/buffer/page_id.hpp
```diff
@@ -0,0 +1,62 @@
...
+      : _valid(valid), _size_type(static_cast<PageIDType>(size_type)), index(index) {}
+};
+
+inline std::ostream& operator<<(std::ostream& os, const PageID& page_id) {
+  os << "PageID(valid = " << page_id.valid() << ", size_type = " << magic_enum::enum_name(page_id.size_type())
+     << ", index = " << page_id.index << ")";
+  return os;
+}
+
+static constexpr PageID INVALID_PAGE_ID = PageID{MIN_PAGE_SIZE_TYPE, 0, false};
```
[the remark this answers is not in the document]
nikriek: no, as soon as the valid flag is set, the other values like index and page type are ignored. However, they still need to be set by default.

src/lib/storage/buffer/page_id.hpp
```diff
@@ -0,0 +1,62 @@
+#pragma once
+#include <magic_enum.hpp>
+#include "boost/integer/static_log2.hpp"
+
+namespace hyrise {
+
+#ifdef __APPLE__
+constexpr size_t OS_PAGE_SIZE = 16384;
```
Bouncner, line 8: Short comment on that? I found this site (https://developer.apple.com/library/archive/documentation/Performance/Conceptual/ManagingMemory/Articles/AboutMemory.html) which mentions the changes of page sizes for MacOS. I was totally unaware, so a short comment might help.

src/lib/storage/buffer/page_id.hpp
```diff
@@ -0,0 +1,62 @@
...
+enum class PageSizeType { KiB16, KiB32, KiB64, KiB128, KiB256, KiB512, MiB1, MiB2 };
+#elif __linux__
+constexpr size_t OS_PAGE_SIZE = 4096;
+enum class PageSizeType { KiB4, KiB8, KiB16, KiB32, KiB64, KiB128, KiB256, KiB512, MiB1, MiB2 };
+#endif
+
+// Get the number of bytes for a given PageSizeType
```
Bouncner, line 15: Ending period.

src/lib/storage/buffer/page_id.hpp
```diff
@@ -0,0 +1,62 @@
...
+constexpr inline size_t bytes_for_size_type(const PageSizeType size) {
+  return OS_PAGE_SIZE << static_cast<size_t>(size);
```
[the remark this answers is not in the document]
Bouncner: Maybe better a static_assert? (which would mean having actual values here .... don't know)

src/lib/storage/buffer/page_id.hpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+constexpr size_t NUM_PAGE_SIZE_TYPES = magic_enum::enum_count<PageSizeType>();
+constexpr PageSizeType MIN_PAGE_SIZE_TYPE = magic_enum::enum_value<PageSizeType>(0);
+constexpr PageSizeType MAX_PAGE_SIZE_TYPE = magic_enum::enum_value<PageSizeType>(NUM_PAGE_SIZE_TYPES - 1);
+constexpr size_t PAGE_SIZE_TYPE_BITS = boost::static_log2<NUM_PAGE_SIZE_TYPES>::value + 1;
+
+/**
+ * PageIDs are used for addressing pages. They consist of a valid flag, a PageSizeType and an index. A Page ID can be unambiguously
```
Bouncner, line 26:
```suggestion
 * PageIDs are used for addressing pages. They consist of a valid flag, a PageSizeType and an index. A PageID can be unambiguously
```

src/lib/storage/buffer/page_id.hpp
```diff
@@ -0,0 +1,62 @@
...
+ * converted into a virtual address and back. If the valid flag is set, the page is stored in the buffer pool. Otherwise, the page
+ * does not exist or the virtual memory address is outside of the buffer pool.
+*/
+struct PageID {
+  using PageIDType = uint64_t;
+
+  PageIDType _valid : 1;
```
Bouncner, line 33: I have never seen bitfields before. Can you add a short column why you are doing this here?

Moreover, why are you doing it this way here and "manually" above?

src/lib/storage/buffer/page_id.hpp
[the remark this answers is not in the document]
Bouncner: Add a comment?

src/lib/storage/buffer/page_id.hpp
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  bool valid() const {
+    return _valid;
+  }
+
+  PageID() = default;
+
+  constexpr PageID(const PageSizeType size_type, const PageIDType index, bool valid = true)
+      : _valid(valid), _size_type(static_cast<PageIDType>(size_type)), index(index) {}
```
Bouncner, line 52: Different orders of parameters and members. Align them?

src/test/lib/storage/buffer/frame_test.cpp
```diff
@@ -0,0 +1,150 @@
+#include "base_test.hpp"
+#include "storage/buffer/frame.hpp"
+
+namespace hyrise {
+
+class FrameTest : public BaseTest {};
```
dey4ss, line 6: There are plenty of (Debug)Asserts in the Frame code. I think we should test them as well.

src/test/lib/storage/buffer/frame_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+  frame.reset_dirty();
+  EXPECT_FALSE(frame.is_dirty());
+
+  frame.unlock_exclusive();
+}
+
+TEST_F(FrameTest, TestStreamOperator) {
+  auto frame = Frame{};
+  {
+    std::stringstream out;
```
dey4ss, line 100: pick here: auto to stick

src/test/lib/storage/buffer/frame_test.cpp
dey4ss: Pick here, `TEST_F(XYTest, TestAbc)` is a bit too much for me. At least `TestAbc` could be just `Abc` I guess.

dey4ss: Furthermore, a lot of the Frame code focusses on parallel execution. I think some stress tests for these parts (`stress_test.cpp` are a good idea).

dey4ss (commented):
Not too much to add to what @Bouncner and @mweisgut already pointed out

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,110 @@
...
+  StateVersionType state_and_version() const;
+  static StateVersionType state(StateVersionType state_and_version);
+  static StateVersionType version(StateVersionType state_and_version);
+  static NodeID node_id(StateVersionType state_and_version);
+
+  void _debug_print();
+
+ private:
+  // clang-format off
+  static constexpr uint64_t NODE_ID_MASK   = 0x00000F0000000000;
```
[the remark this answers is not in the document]
nikriek: That was the plan. Somehow, clang still formatted it

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include "types.hpp"
+
+namespace hyrise {
+
+/**
+ * Frames are the metadata objects for each page. We are using a 64 bit atomic integer to store the latching state, NUMA node, dirty flag and the version of the frame.
+ * All operations are atomic. The basic idea and most of the code is based on the SIGMOD'23 paper "Virtual-Memory Assisted Buffer Management" by Leis et. al.
+ * 
+ * NodeID and the dirty flag can be set independently from the state of the frame. The state is encoded in the upper 16 bits of the 64 bit integer. 
+ * The state is encodoed as EVICTED, LOCKED, UNLOCKED, MARKED and LOCK_SHARED. Initially, the frame is in state EVICTED. Then, the page gets LOCKED for reading from disk.
```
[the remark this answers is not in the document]
nikriek: Fair!

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,110 @@
...
+class Frame final {
+ public:
+  using StateVersionType = uint64_t;
+
+  // State constants
+  static constexpr StateVersionType UNLOCKED = 0;
+  static constexpr StateVersionType LOCKED_SHARED = 0xFFFF - 3;  // 252 if 8 bits, 65532
+  static constexpr StateVersionType LOCKED = 0xFFFF - 2;         // 253 if 8 bits, 65533
+  static constexpr StateVersionType MARKED = 0xFFFF - 1;         // 254 if 8 bits, 65534
+  static constexpr StateVersionType EVICTED = 0xFFFF;            // 255 if 8 bits, 65535 for 16 bits
```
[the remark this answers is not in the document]
nikriek: Indeed. will fix

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+  Frame();
+
+  // Flags and other metadata
+  void set_node_id(const NodeID node_id);
+  void set_dirty(const bool new_dirty);
```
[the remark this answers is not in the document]
nikriek: Yes, I should use a function `mark_dirty` instead. There is not benefit in having a parameter.

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,110 @@
+#pragma once
+
+#include <bit>
+
+#include <atomic>
```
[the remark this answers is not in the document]
nikriek: Yes, that was an issue before. It complained about the order to C, C++ headers etc.

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,110 @@
...
+ *                                                                                    +------------------------------------------------------------+
+ * Graphic was created with https://github.com/ggerganov/dot-to-ascii
+*/
+
+class Frame final {
+ public:
+  using StateVersionType = uint64_t;
+
+  // State constants
+  static constexpr StateVersionType UNLOCKED = 0;
```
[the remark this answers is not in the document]
nikriek: Everything between 0 and LOCK_SHARED acts a a counter of shared readers. We could rename the variable to `MAX_LOCK_SHARED` to make it more clear.

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+  // Removes a shared locked and returns true if the frame is now unlocked
+  bool unlock_shared();
+
+  void unlock_exclusive();
+
+  bool is_unlocked() const;
+
+  // State and version helper
+  StateVersionType state_and_version() const;
```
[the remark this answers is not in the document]
nikriek: Its a compound type bases on a `uint64_t`.

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,110 @@
...
+  void _debug_print();
+
+ private:
+  // clang-format off
+  static constexpr uint64_t NODE_ID_MASK   = 0x00000F0000000000;
+  static constexpr uint64_t DIRTY_MASK       = 0x0000F00000000000;
+  static constexpr uint64_t STATE_MASK       = 0xFFFF000000000000;
+  static constexpr uint64_t VERSION_MASK     = 0x000000FFFFFFFFFF;
+  // clang-format on
+  static_assert((NODE_ID_MASK ^ DIRTY_MASK ^ STATE_MASK ^ VERSION_MASK) == std::numeric_limits<StateVersionType>::max(),
```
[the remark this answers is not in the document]
nikriek: Yes!

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include <bitset>
+
+namespace hyrise {
+
+Frame::Frame() {
+  _state_and_version.store(update_state_with_same_version(0, EVICTED), std::memory_order_release);
+}
+
+void Frame::set_node_id(const NodeID node_id) {
+  DebugAssert(node_id <= (NODE_ID_MASK >> NODE_ID_SHIFT), "Node ID must be smaller than 16.");
```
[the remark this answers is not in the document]
nikriek: This is a good question!

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,172 @@
...
+  DebugAssert(node_id != INVALID_NODE_ID, "Cannot set empty numa node");
+  DebugAssert(state(_state_and_version.load()) == LOCKED, "Frame must be locked to set memory node.");
+  auto old_state_and_version = _state_and_version.load();
+  while (true) {
+    auto new_state_and_version =
```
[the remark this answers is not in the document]
nikriek: no, because it is passed a reference to the CAS function.

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,172 @@
...
+        old_state_and_version ^
```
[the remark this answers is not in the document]
nikriek: ok!

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+      break;
+    }
+  }
+
+  DebugAssert(Frame::node_id(_state_and_version.load()) == node_id,
+              "Setting numa node didnt work: " + std::to_string(Frame::node_id(_state_and_version.load())));
+}
+
+void Frame::set_dirty(const bool new_dirty) {
+  DebugAssert(state(_state_and_version.load()) == LOCKED, "Frame must be locked to set dirty flag.");
```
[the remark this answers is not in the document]
nikriek: Yes, I prefer to have these debug asserts as safety checks during development.

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+  _state_and_version.fetch_or(DIRTY_MASK & (dirty << DIRTY_SHIFT));
+  DebugAssert(is_dirty() == new_dirty, "Setting dirty didnt work");
+}
+
+void Frame::reset_dirty() {
+  _state_and_version &= ~DIRTY_MASK;
+}
+
+void Frame::unlock_exclusive_and_set_evicted() {
+  Assert(state(_state_and_version.load()) == LOCKED, "Frame must be marked to set evicted flag.");
```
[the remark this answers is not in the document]
nikriek: Should be DebugAssert, too.

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+  _state_and_version.store(update_state_with_increment_version(_state_and_version.load(), EVICTED),
```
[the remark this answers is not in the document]
nikriek: There cannot be another current writer. Thereby it is safe to do it like this. Other than that, we should use a debug assert above.

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+                           std::memory_order_release);
+}
+
+bool Frame::try_mark(Frame::StateVersionType old_state_and_version) {
```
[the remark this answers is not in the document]
nikriek: const is better yes!

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+                                                    update_state_with_same_version(old_state_and_version, LOCKED));
+}
+
+bool Frame::unlock_shared() {
+  while (true) {
+    auto old_state_and_version = _state_and_version.load();
+    auto old_state = state(old_state_and_version);
+    Assert(old_state > 0 && old_state <= LOCKED_SHARED, "Frame must be locked shared to unlock shared.");
+    auto new_state = old_state - 1;
+    if (_state_and_version.compare_exchange_strong(old_state_and_version,
```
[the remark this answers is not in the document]
nikriek: No, this is not a spin-lock. This is is a classical [CAS loop](https://en.wikipedia.org/wiki/Compare-and-swap). We use the frame to implement a spin-lock using atomic updates.

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+  Assert(state(_state_and_version.load()) == LOCKED,
+         "Frame must be locked to unlock exclusive. " + std::to_string(state(_state_and_version.load())));
+  _state_and_version.store(update_state_with_increment_version(_state_and_version.load(), UNLOCKED),
+                           std::memory_order_release);
+}
+
+Frame::StateVersionType Frame::update_state_with_same_version(Frame::StateVersionType old_version_and_state,
+                                                              Frame::StateVersionType new_state) {
+  constexpr auto SHIFT = NUM_BITS - STATE_SHIFT;
+  static_assert(SHIFT == 16, "Shift must be 8.");
```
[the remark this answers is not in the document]
nikriek: upsi

src/lib/storage/buffer/page_id.hpp
nikriek: Good point! Generally, I am not entirely happy with this difference between OSX and Linux. Linux may also have larger pages..

src/lib/storage/buffer/page_id.hpp
nikriek: Atomic bitfields do not exist afaik. Bitfields for `uint64_t` are much easier to use to the `PageID`. I will add a comment to explain it.

src/lib/storage/buffer/page_id.hpp
```diff
@@ -0,0 +1,62 @@
...
+};
+
+inline std::ostream& operator<<(std::ostream& os, const PageID& page_id) {
```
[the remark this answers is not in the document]
nikriek: Fair, my idea was to use to use the top bit as the valid flag. This allows use to easily read the page id in binary mode. The valid flag has to be last value in order to set a default value.

src/lib/storage/buffer/page_id.hpp
nikriek: Having the valid flag as the top bit, allows us to better read the binary representation. Still, we need it as the last parameter to set a default value.

src/test/lib/storage/buffer/frame_test.cpp
nikriek: What's the point of testing DebugAsserts? they are basically tests themselves (or invariants).

nikriek: I have a stresstest ready for the whole buffer manager: https://github.com/nikriek/hyrise-fork/blob/3f847898c97925ca21e56e7c3594c8cd77b59a6a/src/test/lib/concurrency/buffer_manager_stress_test.cpp#L17

src/lib/storage/buffer/frame.hpp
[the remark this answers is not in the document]
nikriek: Underscore is fine! Also for private constants?

src/lib/storage/buffer/page_id.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+// Get the number of bytes for a given PageSizeType
+constexpr inline size_t bytes_for_size_type(const PageSizeType size) {
+  return OS_PAGE_SIZE << static_cast<size_t>(size);
+}
+
+constexpr size_t NUM_PAGE_SIZE_TYPES = magic_enum::enum_count<PageSizeType>();
+constexpr PageSizeType MIN_PAGE_SIZE_TYPE = magic_enum::enum_value<PageSizeType>(0);
+constexpr PageSizeType MAX_PAGE_SIZE_TYPE = magic_enum::enum_value<PageSizeType>(NUM_PAGE_SIZE_TYPES - 1);
+constexpr size_t PAGE_SIZE_TYPE_BITS = boost::static_log2<NUM_PAGE_SIZE_TYPES>::value + 1;
```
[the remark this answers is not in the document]
nikriek: Nice, much better!

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,172 @@
+#include "frame.hpp"
+#include <bitset>
+
+namespace hyrise {
+
+Frame::Frame() {
```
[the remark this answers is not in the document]
Bouncner: Hmm, why did I ask for a comment here?

src/lib/storage/buffer/frame.hpp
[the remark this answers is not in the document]
Bouncner: cpplint might not be aware of "new" headers. In this case, it assumes they are C header. You can try updating the cpplint module or simply use NOLINT (we also do it elsewhere).

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,110 @@
...
+ * Graphic was created with https://github.com/ggerganov/dot-to-ascii
+*/
+
+class Frame final {
+ public:
+  using StateVersionType = uint64_t;
+
+  // State constants
+  static constexpr StateVersionType UNLOCKED = 0;
+  static constexpr StateVersionType LOCKED_SHARED = 0xFFFF - 3;  // 252 if 8 bits, 65532
```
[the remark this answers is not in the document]
nikriek: Removed them.

src/lib/storage/buffer/frame.hpp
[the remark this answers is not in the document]
Bouncner: Also please add comments all over the place. This was absolutely not clear to me.

What happens if we have more readers?

src/lib/storage/buffer/frame.cpp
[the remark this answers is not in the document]
Bouncner: "Yes" to what? `Assert()` is too expensive?

src/lib/storage/buffer/page_id.hpp
Bouncner: Ahh ... default parameter. Got it.

src/test/lib/storage/buffer/frame_test.cpp
Bouncner: Proper stress tests also came to my mind when seeing the code.
I still have to dig into the BufferManager, but I would assume that buffering could hide some cases where might have concurrency issues because there are other overheads.

What I want to ask is: sure we don't want a stress test solely for modifying Frames?

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+std::ostream& operator<<(std::ostream& os, const Frame& frame) {
+  auto state = Frame::state(frame.state_and_version());
+  auto version = Frame::version(frame.state_and_version());
+  auto node_id = frame.node_id();
+  auto dirty = frame.is_dirty();
+
+  std::stringstream ss;
+  ss << "Frame { state = ";
+
+  switch (state) {
```
dey4ss, line 147: Seeing this code here: does it make sense to have the states in an enum with assigned values and cast or use the enum values in the code?

nikriek: I was doing this initially, too. However, you cannot represent values before `LOCKED_SHARED` properly.

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+  return state(_state_and_version.load()) == UNLOCKED;
+}
+
+std::ostream& operator<<(std::ostream& os, const Frame& frame) {
+  auto state = Frame::state(frame.state_and_version());
+  auto version = Frame::version(frame.state_and_version());
+  auto node_id = frame.node_id();
+  auto dirty = frame.is_dirty();
+
+  std::stringstream ss;
```
dey4ss, line 144: Besides the naming and auto to stick here, why not directly writing to the passed stream?

src/lib/storage/buffer/frame.cpp
dey4ss: I think sth like
```c++
enum class FrameState : uint64_t { Unlocked = UNLOCKED, Locked = LOCKED, Marked = MARKED, Evicted = EVICTED };
```

and then sth like:
```c++
const auto state = magic_enum::enum_cast<FrameState>(...);
if (!state) {
  stream << "LockedShared";
} else {
   stream << magic_enum::enum_name(*state);
}
```
would work here. For other occasions, casting to the underlying type if necessary and comparing just like you did should also work. However, not a blocker.

src/lib/storage/buffer/frame.cpp
nikriek: Multiple threads may write to the same stream. This would interleave the output.

nikriek: `osyncstream` might better though

src/lib/storage/buffer/frame.cpp
nikriek: Yeah, but what’s the point of an enum if it doesn’t capture all values? Also, what happens if the state is larger than EVICTED?

src/test/lib/storage/buffer/frame_test.cpp
nikriek: Ok noted.

src/lib/storage/buffer/frame.cpp
nikriek: To make things transparent basically.

src/lib/storage/buffer/frame.cpp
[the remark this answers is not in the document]
nikriek: I just added one

src/lib/storage/buffer/frame.cpp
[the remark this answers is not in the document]
nikriek: I updated it to 128 possible values (7 bits). We may never see much more than a few numa nodes

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+        ((old_state_and_version ^ static_cast<Frame::StateVersionType>(node_id) << NODE_ID_SHIFT) & NODE_ID_MASK);
+    if (_state_and_version.compare_exchange_strong(old_state_and_version, new_state_and_version)) {
+      DebugAssert((old_state_and_version & ~NODE_ID_MASK) == (new_state_and_version & ~NODE_ID_MASK),
+                  "Settings the numa node failed");
+      break;
+    }
+  }
+
+  DebugAssert(Frame::node_id(_state_and_version.load()) == node_id,
+              "Setting numa node didnt work: " + std::to_string(Frame::node_id(_state_and_version.load())));
```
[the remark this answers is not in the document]
nikriek: This is possible yes. But I feel that this line is just a sanity check anyways

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+}
+
+void Frame::set_dirty(const bool new_dirty) {
+  DebugAssert(state(_state_and_version.load()) == LOCKED, "Frame must be locked to set dirty flag.");
+  Frame::StateVersionType dirty = new_dirty;
+  _state_and_version.fetch_or(DIRTY_MASK & (dirty << DIRTY_SHIFT));
+  DebugAssert(is_dirty() == new_dirty, "Setting dirty didnt work");
```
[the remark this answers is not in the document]
nikriek: Unlikely, because the frame needs to be locked. I just added this for better debugging

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+}
+
+void Frame::unlock_exclusive() {
+  Assert(state(_state_and_version.load()) == LOCKED,
+         "Frame must be locked to unlock exclusive. " + std::to_string(state(_state_and_version.load())));
+  _state_and_version.store(update_state_with_increment_version(_state_and_version.load(), UNLOCKED),
+                           std::memory_order_release);
+}
+
+Frame::StateVersionType Frame::update_state_with_same_version(Frame::StateVersionType old_version_and_state,
```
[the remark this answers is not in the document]
nikriek: The version signifies concurrent updates of the pages. This is used for eviction and also helpful for optimistic latching (not implemented).

src/lib/storage/buffer/page_id.hpp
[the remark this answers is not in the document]
nikriek: Added a comment

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+  // Flags and other metadata
+  void set_node_id(const NodeID node_id);
+  void set_dirty(const bool new_dirty);
+  bool is_dirty() const;
+  void reset_dirty();
+  NodeID node_id() const;
+
+  // State transitions
+  void unlock_exclusive_and_set_evicted();
```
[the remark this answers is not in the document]
nikriek: Its on the left

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,172 @@
...
+namespace hyrise {
+
+Frame::Frame() {
+  _state_and_version.store(update_state_with_same_version(0, EVICTED), std::memory_order_release);
+}
+
+void Frame::set_node_id(const NodeID node_id) {
+  DebugAssert(node_id <= (NODE_ID_MASK >> NODE_ID_SHIFT), "Node ID must be smaller than 16.");
+  DebugAssert(node_id != INVALID_NODE_ID, "Cannot set empty numa node");
+  DebugAssert(state(_state_and_version.load()) == LOCKED, "Frame must be locked to set memory node.");
```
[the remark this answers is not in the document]
nikriek: I primarily added all these invariants for testing. I think, its fine to keep them in debug mode.

src/lib/storage/buffer/frame.cpp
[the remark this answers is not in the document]
nikriek: I would like to avoid any assert in frame code. Debug Assert are useful for testing.

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+  return node_id(_state_and_version.load());
+}
+
+NodeID Frame::node_id(Frame::StateVersionType state_and_version) {
+  return static_cast<NodeID>((state_and_version & NODE_ID_MASK) >> NODE_ID_SHIFT);
+}
+
+bool Frame::try_lock_shared(Frame::StateVersionType old_state_and_version) {
+  auto old_state = state(old_state_and_version);
+  if (old_state < LOCKED_SHARED) {
```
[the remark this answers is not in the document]
nikriek: I add a comment

src/lib/storage/buffer/frame.cpp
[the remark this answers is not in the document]
nikriek: done

src/lib/storage/buffer/frame.cpp
dey4ss: Isn't printing the info just meant for test and debugging purposes?
If we write many things concurrently/in a hot loop to stdout (which will be the main usage of printing anything), then we have other issues than mingled output. And if users plan to do so, it should be their call to ensure a nice output by using a secomnd stream I guess.

nikriek: Better idea! Thanks!

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,173 @@
...
+   */
+  bool try_lock_shared(const StateVersionType old_state_and_version);
+
+  // Try to latch in frame in exclusive node. Fails if the frame is not in state UNLOCKED or MARKED.
+  bool try_lock_exclusive(const StateVersionType old_state_and_version);
+
+  // Try to mark the frame. Fails if the frame is not in state UNLOCKED.
+  bool try_mark(const StateVersionType old_state_and_version);
+
+  // Decrement the shared latch. Reaches UNLOCKED if the state is 0.
```
mweisgut, line 104: Similar to the cpp, mentioning when this function returns `true` would clarify the description. `Reaches UNLOCKED` is not clear imo

src/test/lib/storage/buffer/frame_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+  frame.unlock_exclusive();
+  EXPECT_EQ(Frame::state(frame.state_and_version()), Frame::UNLOCKED);
+  EXPECT_EQ(Frame::version(frame.state_and_version()), 1);
+  EXPECT_TRUE(frame.is_unlocked());
+
+  // Lock 3 times in shared mode
+  old_state_and_version = frame.state_and_version();
+  EXPECT_TRUE(frame.try_lock_shared(old_state_and_version));
+  EXPECT_EQ(Frame::version(frame.state_and_version()), 1);
+  EXPECT_EQ(Frame::state(frame.state_and_version()), 1);
```
mweisgut, line 29: Replacing `1` with `SINGLE_LOCKED_SHARED` might demonstrate its usage and descriptiveness.

src/test/lib/storage/buffer/frame_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+  EXPECT_EQ(Frame::state(frame.state_and_version()), 3);
+
+  // Unlock 3 times in shared mode
+  EXPECT_FALSE(frame.unlock_shared());
+  EXPECT_EQ(Frame::version(frame.state_and_version()), 1);
+  EXPECT_EQ(Frame::state(frame.state_and_version()), 2);
+
+  EXPECT_FALSE(frame.unlock_shared());
+  EXPECT_EQ(Frame::version(frame.state_and_version()), 1);
+  EXPECT_EQ(Frame::state(frame.state_and_version()), 1);
```
mweisgut, line 48: Same as above: `SINGLE_LOCKED_SHARED`?

src/test/lib/storage/buffer/frame_test.cpp
```diff
@@ -0,0 +1,150 @@
...
+  old_state_and_version = frame.state_and_version();
+  EXPECT_TRUE(frame.try_mark(old_state_and_version));
+  EXPECT_EQ(Frame::version(frame.state_and_version()), 1);
+  EXPECT_EQ(Frame::state(frame.state_and_version()), Frame::MARKED);
+
+  // Lock shared again and unlock
+  old_state_and_version = frame.state_and_version();
+  EXPECT_TRUE(frame.try_lock_shared(old_state_and_version));
+  EXPECT_EQ(Frame::version(frame.state_and_version()), 1);
+  EXPECT_EQ(Frame::state(frame.state_and_version()), 1);
```
mweisgut, line 64: Same as above: `SINGLE_LOCKED_SHARED`?

src/test/lib/storage/buffer/page_id_test.cpp
```diff
@@ -0,0 +1,34 @@
...
+  EXPECT_EQ(larger_page.size_type(), PageSizeType::KiB256);
+  EXPECT_EQ(larger_page.byte_count(), 262144);
+  EXPECT_TRUE(larger_page.valid());
+  EXPECT_EQ(larger_page.index(), 59);
+}
+
+TEST_F(PageIDTest, TestInvalidPageID) {
+  EXPECT_FALSE(PageID(PageSizeType::KiB16, 29, false).valid());
+  EXPECT_FALSE(INVALID_PAGE_ID.valid());
+}
```
mweisgut, line 25: Could you please add some test cases testing the PageID's equality operator?
This might include (i) a valid page id, (ii) the comparison between an invalid page ID with a page size type > MIN_PAGE_SIZE and `INVALID_PAGE_ID`, and (ii) the comparison between an invalid page ID with an index > 0 and `INVALID_PAGE_ID`.

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  DebugAssert(Frame::node_id(_state_and_version.load()) == node_id,
+              "Setting NUMA node didnt work: " + std::to_string(Frame::node_id(_state_and_version.load())));
+}
+
+void Frame::mark_dirty() {
+  DebugAssert(state(_state_and_version.load()) == LOCKED, "Frame must be locked to set dirty flag.");
+  const auto new_dirty = StateVersionType{1};
+  _state_and_version.fetch_or(_DIRTY_MASK & (new_dirty << _DIRTY_SHIFT));
```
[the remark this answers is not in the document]
nikriek: yes, indeed. I simplified this function a bit and forgot to update with the mask!

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  DebugAssert(Frame::node_id(_state_and_version.load()) == node_id,
+              "Setting numa node didnt work: " + std::to_string(Frame::node_id(_state_and_version.load())));
+}
+
+void Frame::set_dirty(const bool new_dirty) {
+  DebugAssert(state(_state_and_version.load()) == LOCKED, "Frame must be locked to set dirty flag.");
+  Frame::StateVersionType dirty = new_dirty;
+  _state_and_version.fetch_or(DIRTY_MASK & (dirty << DIRTY_SHIFT));
```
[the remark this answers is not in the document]
nikriek: I removed it

src/lib/storage/buffer/frame.cpp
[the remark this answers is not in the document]
nikriek: It is not problematic since the frame is currently locked exclusively by the current thread.

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,169 @@
...
+  Assert(state(_state_and_version.load()) == LOCKED, "Frame must be marked to set evicted flag.");
+  _state_and_version.store(update_state_with_increment_version(_state_and_version.load(), EVICTED),
+                           std::memory_order_release);
+}
+
+bool Frame::try_mark(Frame::StateVersionType old_state_and_version) {
+  DebugAssert(state(_state_and_version.load()) == UNLOCKED,
+              "Frame must be unlocked to mark, instead: " + std::to_string(state(_state_and_version.load())));
+  return _state_and_version.compare_exchange_strong(old_state_and_version,
+                                                    update_state_with_same_version(old_state_and_version, MARKED));
```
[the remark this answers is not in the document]
nikriek: We might want to change the naming. But still, if we look at the function on its own, the name actually makes sense. We pass it the state_version and new state and return the updated value with the new state

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,180 @@
...
+  return _state_and_version.compare_exchange_strong(state_and_version,
+                                                    _update_state_with_same_version(old_state_and_version, LOCKED));
+}
+
+bool Frame::unlock_shared() {
+  while (true) {
+    auto old_state_and_version = _state_and_version.load();
+    const auto old_state = state(old_state_and_version);
+    DebugAssert(old_state > 0 && old_state <= MAX_LOCKED_SHARED, "Frame must be locked shared to unlock shared.");
+    // Decrement the state by 1 to remove a concurrent reader
```
Bouncner, line 116: That lowers the version, right? Say so?

src/lib/storage/buffer/frame.hpp
```diff
@@ -0,0 +1,177 @@
...
+#include "types.hpp"
+
+namespace hyrise {
+
+/**
+ * Frames are the metadata objects for each page. We are using a 64 bit atomic integer to store the (latching) state, NUMA node, dirty flag and the version of the frame.
+ * All operations are atomic. The basic idea and most of the code is based on the SIGMOD'23 paper "Virtual-Memory Assisted Buffer Management" by Leis et al.
+ * 
+ * The frame's upper 16 bits encode the (latching) state (see below). 1 bit is used for the dirty flag. 7 bits are used for the NUMA node. The lower 40 bits are used for the version. 
+ * The version is used to detect concurrent changes to the state of the frame. The version is incremented after unlocking the frame.
```
Bouncner, line 16: Decremented? (looking at `unlock_shared()`)

How exactly do you determine concurrent changes?

src/lib/storage/buffer/frame.cpp
nikriek: No, it des not. Only the state

src/lib/storage/buffer/frame.hpp
nikriek: The version is incremented for exclusive locks only. For shared locks, the state is decremented and the version stays as is. Concurrent changes are only allowed with exclusive locks

src/lib/storage/buffer/frame.hpp
```diff
@@ -35,73 +46,129 @@ namespace hyrise {
...
   static constexpr StateVersionType UNLOCKED = 0;
-  static constexpr StateVersionType LOCKED_SHARED = 0xFFFF - 3;  // 252 if 8 bits, 65532
-  static constexpr StateVersionType LOCKED = 0xFFFF - 2;         // 253 if 8 bits, 65533
-  static constexpr StateVersionType MARKED = 0xFFFF - 1;         // 254 if 8 bits, 65534
-  static constexpr StateVersionType EVICTED = 0xFFFF;            // 255 if 8 bits, 65535 for 16 bits
 
+  // SINGLE_LOCKED_SHARED signifies that the frame has a single reader. It has a value of 1. Used for convenience.
+  static constexpr StateVersionType SINGLE_LOCKED_SHARED = 1;
+
+  // LOCKED_SHARED signifies that the frame has the maxium amount of shared readers. It has a value of 65532.
```
mweisgut, line 63: `LOCK_SHARED` in the comment vs `MAX_LOCKED_SHARED` as the constant name might be misleading. In all the comments, I think the constant name does not need to be repeated. E.g., `The maximum number of shared readers. Value: 65532` is sufficient, in my opinion.

src/lib/storage/buffer/frame.hpp
```diff
@@ -35,73 +46,129 @@ namespace hyrise {
...
 
-  static constexpr uint64_t NUM_BITS = sizeof(StateVersionType) * CHAR_BIT;
-  static constexpr uint64_t NODE_ID_SHIFT = std::countr_zero(NODE_ID_MASK);
-  static constexpr uint64_t DIRTY_SHIFT = std::countr_zero(DIRTY_MASK);
-  static constexpr uint64_t STATE_SHIFT = std::countr_zero(STATE_MASK);
+  static_assert((_node_id_mask ^ _dirty_mask ^ _state_mask ^ _version_mask) ==
+                    std::numeric_limits<StateVersionType>::max(),
+                "The given masks either overlap or do not cover the whole StateVersionType.");
+
+  // The number of bits for the state and version. It should be 64.
```
mweisgut, line 153: > It should be 64.

Should we add a debug assertion for that somewhere?

src/lib/storage/buffer/page_id.hpp
```diff
@@ -1,62 +1,109 @@
...
-constexpr size_t OS_PAGE_SIZE = 16384;
+// OS pages on Mac OS are 16 KiB according to
+// https://developer.apple.com/library/archive/documentation/Performance/Conceptual/ManagingMemory/Articles/AboutMemory.html.
+constexpr uint64_t OS_PAGE_SIZE = 16384;
+#elif __linux__
+// OS pages on Linux are usually 4 KiB.
+constexpr uint64_t OS_PAGE_SIZE = 4096;
+#endif
+
+// Pages sizes are always a multiple of the OS page size and increase by powers of two.
```
mweisgut, line 18:
```suggestion
// Page sizes are always a multiple of the OS page size and increase by powers of two.
```

src/lib/storage/buffer/page_id.hpp
```diff
@@ -1,62 +1,109 @@
...
-  return OS_PAGE_SIZE << static_cast<size_t>(size);
+// Get the number of bytes for a given PageSizeType.
+constexpr inline uint64_t bytes_for_size_type(const PageSizeType size) {
+  // We assume that the OS page size is either 4 KiB on Linux or 16 KiB on Mac OS.
+  // The page size types increase by power of two.
+  return OS_PAGE_SIZE << static_cast<uint64_t>(size);
 }
 
-constexpr size_t NUM_PAGE_SIZE_TYPES = magic_enum::enum_count<PageSizeType>();
+// Get number of PageSizeTypes
```
mweisgut, line 33:
```suggestion
// The number of PageSizeTypes.
```

src/lib/storage/buffer/page_id.hpp
```diff
@@ -1,62 +1,109 @@
...
+constexpr uint64_t PAGE_SIZE_TYPES_COUNT = magic_enum::enum_count<PageSizeType>();
+
+// Get the minimum PageSizeType. KiB16 on Mac OS and KiB4 on Linux
 constexpr PageSizeType MIN_PAGE_SIZE_TYPE = magic_enum::enum_value<PageSizeType>(0);
-constexpr PageSizeType MAX_PAGE_SIZE_TYPE = magic_enum::enum_value<PageSizeType>(NUM_PAGE_SIZE_TYPES - 1);
-constexpr size_t PAGE_SIZE_TYPE_BITS = boost::static_log2<NUM_PAGE_SIZE_TYPES>::value + 1;
+
+// Get the maximum PageSizeType. MiB2 on Mac OS and Linux
```
mweisgut, line 39:
```suggestion
// The maximum PageSizeType. MiB2 on Mac OS and Linux.
```

src/lib/storage/buffer/page_id.hpp
```diff
@@ -1,62 +1,109 @@
...
+constexpr PageSizeType MAX_PAGE_SIZE_TYPE = magic_enum::enum_value<PageSizeType>(PAGE_SIZE_TYPES_COUNT - 1);
+
+// Get the number of bits required to store a PageSizeType
```
mweisgut, line 42:
```suggestion
// The number of bits required to store a PageSizeType
```

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,180 @@
...
+  auto old_state_and_version = _state_and_version.load();
+
+  // Execute a compare-and-swap loop to set the NUMA node
+  while (true) {
+    auto new_state_and_version =
+        old_state_and_version ^
+        ((old_state_and_version ^ static_cast<Frame::StateVersionType>(node_id) << _node_id_shift) & _node_id_mask);
+    if (_state_and_version.compare_exchange_strong(old_state_and_version, new_state_and_version)) {
+      DebugAssert((old_state_and_version & ~_node_id_mask) == (new_state_and_version & ~_node_id_mask),
+                  "Settings the NUMA node failed");
```
Bouncner, line 25: " failed." .. period.

src/lib/storage/buffer/page_id.hpp
```diff
@@ -1,62 +1,109 @@
...
+// TODO(nikriek): Replace with std::bit_width once we get rid of gcc9
```
mweisgut, line 43:
```suggestion
// TODO(nikriek): Replace with std::bit_width once we get rid of GCC 9.
```

src/lib/storage/buffer/frame.cpp
Bouncner: Ah ... misread. If we would decrement a `old_state_and_version`, that would decrement the version?

src/lib/storage/buffer/page_id.hpp
```diff
@@ -1,62 +1,109 @@
...
 struct PageID {
   using PageIDType = uint64_t;
 
-  PageIDType _valid : 1;
-  PageIDType _size_type : PAGE_SIZE_TYPE_BITS;
-  PageIDType index : (sizeof(PageIDType) * CHAR_BIT - PAGE_SIZE_TYPE_BITS - 1);
+  constexpr explicit PageID(const PageSizeType size_type, const uint64_t index, bool valid = true)
+      : _valid(valid), _size_type(static_cast<PageIDType>(size_type)), _index(index) {}
 
+  // Get the PageSizeType for the page
```
mweisgut, line 62:
```suggestion
  // Get the PageSizeType for the page.
```

src/lib/storage/buffer/page_id.hpp
```diff
@@ -1,62 +1,109 @@
...
   PageSizeType size_type() const {
-    return magic_enum::enum_value<PageSizeType>(this->_size_type);
+    return magic_enum::enum_value<PageSizeType>(_size_type);
   }
 
-  size_t num_bytes() const {
+  // Get the number of bytes for the page
```
mweisgut, line 67:
```suggestion
  // Get the number of bytes for the page.
```

src/lib/storage/buffer/frame.cpp
```diff
@@ -0,0 +1,180 @@
...
+  _state_and_version |= _dirty_mask;
+}
+
+void Frame::reset_dirty() {
+  _state_and_version &= ~_dirty_mask;
+}
+
+void Frame::unlock_exclusive_and_set_evicted() {
+  DebugAssert(state(_state_and_version.load()) == LOCKED, "Frame must be marked to set evicted flag.");
+  _state_and_version.store(_update_state_with_increment_version(_state_and_version.load(), EVICTED),
```
Bouncner, line 45: `_incrementED_version`?

src/lib/storage/buffer/page_id.hpp
mweisgut, line 67:
```suggestion
  // Get the number of bytes for the page.
```

src/lib/storage/buffer/page_id.hpp
```diff
@@ -1,62 +1,109 @@
...
+  uint64_t byte_count() const {
     return bytes_for_size_type(size_type());
   }
 
+  // Get the index of the page of a size type
```
mweisgut, line 72:
```suggestion
  // Get the index of the page of a size type.
```

src/lib/storage/buffer/page_id.hpp
```diff
@@ -1,62 +1,109 @@
...
+  uint64_t index() const {
+    return _index;
+  }
+
+  // Returns if the PageID is valid or not in the buffer pool
```
mweisgut, line 77:
```suggestion
  // Returns if the PageID is valid or not in the buffer pool.
```

src/lib/storage/buffer/frame.hpp
Bouncner: > Concurrent changes are only allowed with exclusive locks

You mean any changes are only allowed with exclusive locks?

Can you please adapt the comment to state that the we increment the version only when unlocking from excl. lock?

src/lib/storage/buffer/page_id.hpp
```diff
@@ -1,62 +1,109 @@
...
   bool valid() const {
     return _valid;
   }
 
-  PageID() = default;
+  bool operator==(const PageID& other) const {
+    return (_valid == other._valid) &&
+           (!_valid || (_valid && _size_type == other._size_type && _index == other._index));
```
mweisgut, line 84:
```suggestion
           (!_valid || (_size_type == other._size_type && _index == other._index));
```

We don't need to check `_valid` again since the the right operand of `||` is only evaluated if `_valid` is true.

mweisgut (commented):
Mainly minor documantation comments.

Feel free to add yourself to the list of [Contributors](https://github.com/hyrise/hyrise#contributors).

src/lib/storage/buffer/frame.hpp
nikriek: This is implicitly checked through the masks

src/lib/storage/buffer/page_id.hpp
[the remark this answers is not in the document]
nikriek: No preference. I removed it

mweisgut (approved):

Bouncner (approved):

src/lib/storage/buffer/page_id.hpp
```diff
@@ -0,0 +1,108 @@
...
+  PageSizeType size_type() const {
+    return magic_enum::enum_value<PageSizeType>(_size_type);
+  }
+
+  // Get the number of bytes for the page.
+  uint64_t byte_count() const {
+    return bytes_for_size_type(size_type());
+  }
+
+  // Get the index of the page of a size type.
```
Bouncner, line 72: Seeing the usage in the other PR, I am not sure what this index does. And the comment does not help me here. And the main comment above does not say much about it either. :-/

nikriek: I will add this as a follow-up. Things will also get more clear with the other components.

PR #2622, opened 2023-10-17
[Buffer Manager] Add buffer pool class to control eviction using the VolatileRegions and the PersistenceManager

nikriek:
This PR add the buffer pool class. It implements the Second-Change FIFO eviction queue when the defined virtual memory limit is reached.


tjeyy:

PR #2623, opened 2023-10-20
Add Google Test discovery with ctest

nikriek:
During development with VSCode I found it helpful to see and run directly tests from the UI. This PR adds the required test discovery mechanisms for our unit and system tests.


CMakeLists.txt
```diff
@@ -3,6 +3,8 @@ cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
 
 project(Hyrise)
 
+enable_testing()
```
Bouncner, line 6: When looking for documentation (which I haven't found so far), I stumpled over this post that states that including GoogleTest automatically calls `enable_testing()` (see: https://stackoverflow.com/a/50476125/1147726). Do we still need it here?

src/test/CMakeLists.txt
```diff
@@ -306,9 +309,11 @@ target_precompile_headers(hyriseTest PRIVATE
     [["storage/table.hpp"]]
     [["types.hpp"]]
 )
+gtest_discover_tests(hyriseTest WORKING_DIRECTORY ${PROJECT_SOURCE_DIR} DISCOVERY_TIMEOUT 60)
```
Bouncner, line 312: Quite a large timeout (default is 5). Have you run into issues here?

CMakeLists.txt
nikriek: The linked post mentions `include(CTest)` not `include(GoogleTest)`

src/test/CMakeLists.txt
nikriek: Yes, the hyrise binary was timing out frequently for me on a fairly fast Apple M2. 60 sec might be too much though

Bouncner: I am fine with 60s, but can you add a comment that the default timed out on a M2?

Bouncner: That means the CMake is now significantly slower even a recent M2 chip, yes?

nikriek: I forgot to mention: in order to discover the tests, cmake needs to run the test binary with the `—gtest_list_tests` option. This is what can timeout.

Bouncner: Can you add a comment?

Bouncner:
@mweisgut: Do we want to merge?

mweisgut (approved):

Bouncner:
Trying again to ping you, @nikriek.

Bouncner:
Trying again to ping you, @nikriek.

Bouncner:
Trying again to ping you, @nikriek.

Bouncner:
Trying again to ping you, @nikriek.

Bouncner:
@dey4ss : close or merge?

dey4ss:
Robert uses VSCode, I can ask him to test it and if it would benefit his workflow. I guess it's a nice feature, but if no one uses it, let's close it.

dey4ss:
> If Robert could test that, that would be super helpful. Might also help as a short description in our wiki or so. We have this CLion setup thing, but nothing for VSCode (which most people probably use nowadays).

I'll talk to him next week

Bouncner:
> > If Robert could test that, that would be super helpful. Might also help as a short description in our wiki or so. We have this CLion setup thing, but nothing for VSCode (which most people probably use nowadays).
> 
> I'll talk to him next week

Oh, we already discussed this. 🫣

Rob2U:
The VSCode test discovery also discovers the tests of third-party submodules. Therefore I added the `hyriseTest/` prefix to the discovered 'gtest'-tests. This allows filtering for only these tests via CTest arguments. Except this issue the changes work well for me.

Rob2U:
According to the documentation [here](https://cmake.org/cmake/help/latest/command/enable_testing.html): "This command should be in the source directory root because ctest expects to find a test file in the build directory root."

I also tried it now and the tests were (as expected) not discovered.

dey4ss:
@Rob2U can you also add a wiki page for this feature, maybe with some screenshots?

Rob2U:
@dey4ss I added a short wiki entry [here](https://github.com/hyrise/hyrise/wiki/Set-up-VSCode-for-Running-and-Debugging-Tests-from-the-UI). This entry also contains some general information of how to configure VSCode.

Bouncner:
@Greenscreen23 : can you take a quick look at the VS Code instructions? If I remember correctly, you are a VSCode user, right?

Greenscreen23:
The VS Code instructions are great, they worked (with some small changes for linux) out of the box. I tried running the tests, but got a deprecation warning during build: `atomic_load` is depracted, please use `atomic<shared_ptr<T>>`. This issue does not appear on master, so I think we need to rebase this branch (or merge master into it) if you want me to check whether the tests actually run.

Greenscreen23:
I can now start the tests from vscode. However, running all tests is really slow, and there is some issue with running them in parallel that might take some time to fix.

tjeyy:
Is it slower than running them using the binary?

Greenscreen23:
> Is it slower than running them using the binary?

Yes, it launches the binary for each test separately. If I enable parallel runs, then it runs the binary once on multiple cores. But I need to do some debugging to get that up and running. Will look into that tomorrow.

Bouncner:
Is that something that GTest provides or VSCode provides?

When I searched for it, I did not find anything "official". Calling the binary once for each test sounds like a huge overhead.

Bouncner:
We need to increase the timeouts. It just failed on the Intel mac: https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-2623/11/pipeline

Greenscreen23:
> Is that something that GTest provides or VSCode provides?
> 
> When I searched for it, I did not find anything "official". Calling the binary once for each test sounds like a huge overhead. 

This is something from the VSCode cmake extension. But since this PR seems to be about VSCode and tests, I believe it will fit in here.

Greenscreen23:
So it is possible to run the Tests in parallel with the extension. However, when I do so, the extension adds the names of all tests that should be executed to the ctest cli args, which is too much for the linux kernel (its like 900KiB, and the kernel limit is 128KiB afaik). Even after selection only a subset I get an error "RegularExpression::compile(): Expression too big." You can run single tests from the UI, but it does not seem to be trivial to just run all of them. I would just leave it as it is, and add a note to the wiki

Greenscreen23:
Edit: Even when enabling parallel, it just calls ctest with -j$(nproc), and ctest launches the test binary for each test, but launches $(nproc) many instances of it. So it is also very slow compared to just running the binary single threaded. I've added a note to the wiki.

Greenscreen23:
What timeout is it exactly? Is this the timeout that was specified in this PR (discovery timeout)?

Bouncner:
> What timeout is it exactly? Is this the timeout that was specified in this PR (discovery timeout)?

Yes, exactly. When you take  a look at the reason our CI complains, it is the MacMini with Intel we have running here at the HPI:
https://hyrise-ci.epic-hpi.de/blue/organizations/jenkins/hyrise%2Fhyrise/detail/PR-2623/11/pipeline/329

The error mentions `gtest_discover_tests_impl`
```
CMake Error at /usr/local/share/cmake/Modules/GoogleTestAddTests.cmake:132 (message):
  Error running test executable.

    Path: '/Users/hyrise-ci/.jenkins/workspace/hyrise_hyrise_PR-2623/clang-apple-debug/hyriseTest'
    Working directory: '/Users/hyrise-ci/.jenkins/workspace/hyrise_hyrise_PR-2623'
    Result: Process terminated due to timeout
    Output:
      

Call Stack (most recent call first):
  /usr/local/share/cmake/Modules/GoogleTestAddTests.cmake:275 (gtest_discover_tests_impl)
```

Greenscreen23:
I have increased it to 60 seconds. Let's see if the pipeline now works

Greenscreen23:
Seems to work. Is this fine to merge? @Bouncner

dey4ss (approved):

PR #2640, opened 2024-04-20
Evaluate `COALESCE`

dey4ss:
This PR adds SQL's `COALESCE` as a function and evaluates it in the `ExpressionEvaluator`. `COALESCE` is short-hand for a `CASE` expression selecting non-`NULL` values.


resources/benchmark/tpcds/query_blacklist.cfg
```diff
@@ -85,7 +85,7 @@
 #81.sql
 #82.sql
 #83.sql
-84.sql
+#84.sql
```
Bouncner, line 88: Looking at the excel file right now: What about q40 and q76?

src/lib/sql/sql_translator.cpp
```diff
@@ -2108,6 +2108,34 @@ std::shared_ptr<AbstractExpression> SQLTranslator::_translate_hsql_expr(
...
+      //    WHEN (b IS NOT NULL) THEN b
+      //    ELSE c
+      // END
+      // TODO(anyone): Though we have not seen this in benchmarks, there is no reason to add COALESCE list arguments if
+      // the preceding argument is not nullable.
+      if (name == "COALESCE") {
+        AssertInput(expr.exprList && !expr.exprList->empty(), "COALESCE list must not be empty.");
+        auto case_expression = std::shared_ptr<AbstractExpression>{};
+
+        // Build CASE expression bottom-up, i.e,. in reverse fashion.
```
Bouncner, line 2123: ,.

Bouncner (commented):
Can we add a few SQLite tests?

resources/benchmark/tpcds/query_blacklist.cfg
Bouncner: Okay, 40 is a date cast and 76 has a union. I updated the DS query list.

src/lib/expression/expression_utils.cpp
```diff
@@ -171,7 +171,7 @@ std::string expression_descriptions(const std::vector<std::shared_ptr<AbstractEx
 
 DataType expression_common_type(const DataType lhs, const DataType rhs) {
   Assert(lhs != DataType::Null || rhs != DataType::Null, "Cannot deduce common type if both sides are NULL.");
-  Assert((lhs == DataType::String) == (rhs == DataType::String), "Strings only compatible with strings.");
+  Assert((lhs == DataType::String) == (rhs == DataType::String), "Strings are only compatible with strings.");
```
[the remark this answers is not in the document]
dey4ss: `Arguments of binary expressions must both be either strings or numeric.`?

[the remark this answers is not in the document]
Bouncner: Ack

resources/test_data/sqlite_testrunner_queries.sql
```diff
@@ -419,6 +419,16 @@ SELECT CASE WHEN id + 3.4 < 50 THEN 'Hello' WHEN id < 70 THEN 'World' ELSE 'Ciao
 SELECT CASE id + 10 WHEN 15 THEN a WHEN 26 THEN 'World' ELSE d END AS case_column FROM mixed;
 SELECT a, CASE WHEN a IS NULL THEN 1 ELSE 2 END FROM mixed_null GROUP BY a
 
+-- COALESCE
+SELECT COALESCE(d, '') d_not_null FROM mixed_null;
+SELECT COALESCE(b, -1) AS b_not_null, COALESCE(c, 0.0) c_not_null FROM mixed_null;
+SELECT COALESCE(b, 1-1) b_not_null, COALESCE(c, 3.0/-4.0) c_not_null FROM mixed_null;
+SELECT COALESCE(c, CAST(COALESCE(b, 0) AS FLOAT)) c_not_null FROM mixed_null;
+SELECT COALESCE(b, CAST(c AS INT), 0) b_not_null FROM mixed_null;
```
Bouncner, line 427: What about `COALESCE(NULL, c, NULL)` on a nullable column? Does it make sense to test it?

Bouncner (approved):
Just a tiny question. Good to go.

resources/test_data/sqlite_testrunner_queries.sql
dey4ss: Can add it, we just get `NULL` then. From the standard:
> d) `COALESCE (V1, V2)` is equivalent to the following `<case specification>`:
>  ```SQL
>  CASE
>    WHEN NOT V1 IS NULL THEN V1 
>    ELSE V2
>  END
>  ```
> e) `COALESCE (V1, V2, ..., Vn)`, for n ≥ 3, is equivalent to the following `<case specification>`:
>  ```SQL
>  CASE
>    WHEN NOT V1 IS NULL THEN V1 
>    ELSE COALESCE (V2, ..., Vn)
>  END
> ```
[SQL:2023-2, p. 251]

Thus, your example is equivalent to:
```SQL
CASE
  WHEN NULL IS NOT NULL THEN NULL  -- okay, not gonna happen
  WHEN c IS NOT NULL THEN c        
  ELSE NULL                        -- so just c
```
or simply `c`. Gonna add that case and also the standard excerpt.

Bouncner: Ah, good to know. I wasn't sure how to read SQLServer's description "At least one of the null values must be a typed NULL" (https://learn.microsoft.com/en-us/sql/t-sql/language-elements/coalesce-transact-sql?view=sql-server-ver16). No idea what "typed" means here ...

PR #2641, opened 2024-04-20
Add `ABS` function

dey4ss:
This PR adds `ABS` (absolute value) as a function and evaluates it.


Bouncner (commented):
Any idea how to get around the `CONCAT` vs. `||` issue with SQLite?

dey4ss:
> Any idea how to get around the `CONCAT` vs. `||` issue with SQLite?

I looked into the parser and `||` is actually implemented as binary concatenate operator. It's just not implemented in Hyrise. However, it should be easy to translate (sequences of) such operator expressions to `CONCAT` function expressions.

src/lib/expression/evaluation/expression_evaluator.cpp
```diff
@@ -805,21 +805,21 @@ std::shared_ptr<ExpressionResult<Result>> ExpressionEvaluator::_evaluate_functio
...
         return _evaluate_concatenate(expression.arguments);
       }
-
-      Fail("Function can only be evaluated to a string.");
+      Fail("CONCAT() can only be evaluated to a string.");
+    case FunctionType::Substring:
+      if constexpr (std::is_same_v<Result, pmr_string>) {
+        return _evaluate_substring(expression.arguments);
+      }
+      Fail("SUBSTR() can only be evaluated to a string.");
```
Bouncner, line 816: "To a string"? Rather "on a string"?

Bouncner (approved):
Tiny thing.

src/lib/expression/evaluation/expression_evaluator.cpp
dey4ss: Also had that first. Technically, we determine the data type from the expression to be evalauted, not from the input expression(s). But I guess we can live with that small impreciseness in favor of a nicer error message. The expressions should already have complained if they were to be used on non-strings.

src/lib/utils/assert.hpp
```diff
@@ -15,7 +15,7 @@
  * --> Use DebugAssert() whenever a certain invariant must hold, as in:
  *
  * int divide(int numerator, int denominator) {
- *   DebugAssert(denominator == 0, "Divisions by zero are not allowed");
+ *   DebugAssert(denominator != 0, "Divisions by zero are not allowed.");
```
Bouncner, line 18: 😂

PR #2648, opened 2024-05-26
Turn on scheduler and benchmark data caching in Console

dey4ss:
This PR adds a config option to use binary cached benchmark tables in hyriseConsole, which is turned on by default. Furthermore, I noticed that according to the documentation, the scheduler should be turned on per default, as well. This is also fixed.

Backgroung for this was that I demoed Hyrise using the console and noticed that data generation was always triggered and running on only one core.

### DIFF
--- a/src/benchmark/tpch_data_micro_benchmark.cpp
+++ b/src/benchmark/tpch_data_micro_benchmark.cpp
@@ -27,21 +27,17 @@ using namespace expression_functional;  // NOLINT(build/namespaces)
 
 class TableWrapper;
 
-// Defining the base fixture class
+// Defining the base fixture class.
 class TPCHDataMicroBenchmarkFixture : public MicroBenchmarkBasicFixture {
  public:
   void SetUp(::benchmark::State& state) override {
     auto& sm = Hyrise::get().storage_manager;
     const auto scale_factor = 10.0f;
-    const auto default_encoding = EncodingType::Dictionary;
-
     const auto benchmark_config = std::make_shared<BenchmarkConfig>();
-    // TODO(anyone): setup benchmark_config with the given default_encoding
-    // benchmark_config.encoding_config = EncodingConfig{SegmentEncodingSpec{default_encoding}};
 
     if (!sm.has_table("lineitem")) {
-      std::cout << "Generating TPC-H data set with scale factor " << scale_factor << " and " << default_encoding
-                << " encoding:\n";
+      std::cout << "Generating TPC-H data set with scale factor " << scale_factor << " and "
+                << benchmark_config->encoding_config.default_encoding_spec << " encoding:\n";
       TPCHTableGenerator(scale_factor, ClusteringConfiguration::None, benchmark_config).generate_and_store();
     }
 
--- a/src/benchmark/tpch_data_micro_benchmark.cpp
+++ b/src/benchmark/tpch_data_micro_benchmark.cpp
@@ -35,16 +35,14 @@ class TPCHDataMicroBenchmarkFixture : public MicroBenchmarkBasicFixture {
     const auto scale_factor = 10.0f;
     const auto default_encoding = EncodingType::Dictionary;
 
-    auto benchmark_config = BenchmarkConfig::get_default_config();
+    const auto benchmark_config = std::make_shared<BenchmarkConfig>();
     // TODO(anyone): setup benchmark_config with the given default_encoding
     // benchmark_config.encoding_config = EncodingConfig{SegmentEncodingSpec{default_encoding}};
 
     if (!sm.has_table("lineitem")) {
       std::cout << "Generating TPC-H data set with scale factor " << scale_factor << " and " << default_encoding
                 << " encoding:\n";
-      TPCHTableGenerator(scale_factor, ClusteringConfiguration::None,
-                         std::make_shared<BenchmarkConfig>(benchmark_config))
-          .generate_and_store();
+      TPCHTableGenerator(scale_factor, ClusteringConfiguration::None, benchmark_config).generate_and_store();
     }
 
     _table_wrapper_map = create_table_wrappers(sm);
--- a/src/benchmark/tpch_table_generator_benchmark.cpp
+++ b/src/benchmark/tpch_table_generator_benchmark.cpp
@@ -7,11 +7,11 @@
 namespace hyrise {
 
 /**
- * This benchmark can only be use as a starting point for investigating TPCHTableGenerator performance, since secondary
+ * This benchmark can only be used as a starting point for investigating TPCHTableGenerator performance, since secondary
  * invocations of 'TPCHTableGenerator(scale_factor, 1000).generate();' will profit from cached data in tpch-dbgen
  * @param state
  */
-static void BM_TPCHTableGenerator(benchmark::State& state) {  // NOLINT
+static void BM_TPCHTableGenerator(benchmark::State& state) {
   for (auto _ : state) {
     TPCHTableGenerator(0.5f, ClusteringConfiguration::None, ChunkOffset{1000}).generate_and_store();
     Hyrise::reset();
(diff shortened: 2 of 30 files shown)


src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -348,10 +348,11 @@ void AbstractTableGenerator::generate_and_store() {
   Hyrise::get().set_scheduler(initial_scheduler);
 }
 
-std::shared_ptr<BenchmarkConfig> AbstractTableGenerator::create_benchmark_config_with_chunk_size(
-    ChunkOffset chunk_size) {
+std::shared_ptr<BenchmarkConfig> AbstractTableGenerator::create_benchmark_config_with_chunk_size(ChunkOffset chunk_size,
```
Bouncner, line 351: I don't like that change too much, given the name now only covers one of the two things that this function handles.
Also not a fan of default parameters.

But for  now, I don't have a better idea.

src/bin/console/console.cpp
```diff
@@ -896,6 +901,20 @@ int Console::_change_runtime_setting(const std::string& input) {
     return 0;
   }
 
+  if (property == "binary_caching") {
+    if (value == "on") {
+      _binary_caching = true;
+      out("Binary caching turned on\n");
```
Bouncner, line 907: Trailing period?

Bouncner (approved):
I am not big fan of the method name, but apart from that all good.

src/benchmarklib/abstract_table_generator.cpp
dey4ss: We could also not use the convencience constructors of the TableGenerators, but create the benchmark config in the console and call the regular constructors from there.

src/bin/console/console.cpp
Bouncner: Well, it's not a full sentence. Hmm ..

src/benchmarklib/abstract_table_generator.cpp
Bouncner: I guess I'd prefer that because this function is kind of strange anyways?!?

src/bin/console/console.cpp
dey4ss: So, what's your take? We are kinda inconsequent in the console.

src/benchmarklib/abstract_table_generator.cpp
dey4ss: I think that's because `BenchmarkConfig::get_default_config()` returns an object and not a shared ptr, which is almost always required, and that wrapper does that. Just searched the codebase and basically everywhere where `get_default_config()` is called, a shared ptr using the copy constructor is created afterwards.

src/bin/console/console.cpp
Bouncner: Without periods as those are not full sentences?

src/benchmarklib/abstract_table_generator.cpp
Bouncner: Change the interface then?

dey4ss: Also my first thought. But still, the following is not as easily possible w/o some utility wrapping the assignment of a custom chunk size:
https://github.com/hyrise/hyrise/blob/804f9ed10d7ee197868614287ec5403b3728c205/src/benchmarklib/tpch/tpch_table_generator.cpp#L129-L131

And apparently, we use at least TPCHTableGenerator's convenience constructor at multiple places. Removing the constructor could be one option. Then, a convenience creation method for a BenchmarkConfig could help to not write
```c++
const auto config = ...
config->chunk_size = chunk_size
...TableGenerator{..., config}.generate_and_store();
```
instead of
```c++
...TableGenerator{..., chunk_size}.generate_and_store();  // current way
```
or
```c++
...TableGenerator{..., BenchmarkConfig::with_chunk_size(chunk_size)}.generate_and_store();  // one option
```

Such a function could also be used by the table generators' convenience constructors if we do not want to modify their interfaces. And then we would have sth like:
```c++
// Returns shared_ptr instead of object.
std::shared_ptr<BenchmarkConfig> BenchmarkConfig::create_default_config() {
  return std::make_shared<BenchmarkConfig>(); 
}

// Naming tbd.
std::shared_ptr<BenchmarkConfig> BenchmarkConfig::with_chunk_size(const ChunkOffset chunk_size) {
  auto config = create_default_config();
  config->chunk_size = chunk_size;
  return config; 
}
```

Idk, what's your take on that?

Bouncner: I think the simplest solution would be my favorite. Having a few selected constructors for the benchmark config and use `make_shared` with these instead of having specialized functions in the abstract table generator.

But now that I see the result `...TableGenerator{..., std::make_shared<BenchmarkConfig>(chunk_size)}.generate_and_store();`, I guess I agree with your proposal of the convenience functions. But within the BenchmarkConfig itself and I would rather have overloads than named functions. What do you think?

dey4ss: Yeah, I guess overload constructors for `BenchmarkConfig` also sound viable.

dey4ss: I added overloaded constructors, and thus, got also rid of `BenchmarkConfig::get_default_config()`. @Bouncner do you wanna give it a look?

src/benchmark/tpch_data_micro_benchmark.cpp
```diff
@@ -35,16 +35,14 @@ class TPCHDataMicroBenchmarkFixture : public MicroBenchmarkBasicFixture {
     const auto scale_factor = 10.0f;
     const auto default_encoding = EncodingType::Dictionary;
 
-    auto benchmark_config = BenchmarkConfig::get_default_config();
+    const auto benchmark_config = std::make_shared<BenchmarkConfig>();
     // TODO(anyone): setup benchmark_config with the given default_encoding
     // benchmark_config.encoding_config = EncodingConfig{SegmentEncodingSpec{default_encoding}};
```
dey4ss, line 40: @Bouncner is this still relevant? AFAIK the default encoding spec is already used by the benchmark config.

Bouncner: Good question. Feel free to remove.

src/benchmark/tpch_table_generator_benchmark.cpp
```diff
@@ -7,11 +7,11 @@
 namespace hyrise {
 
 /**
- * This benchmark can only be use as a starting point for investigating TPCHTableGenerator performance, since secondary
+ * This benchmark can only be used as a starting point for investigating TPCHTableGenerator performance, since secondary
  * invocations of 'TPCHTableGenerator(scale_factor, 1000).generate();' will profit from cached data in tpch-dbgen
```
Bouncner, line 11: Trailing period?

src/bin/console/console.cpp
```diff
@@ -534,7 +541,9 @@ int Console::_generate_tpch(const std::string& args) {
   }
 
   out("Generating all TPCH tables (this might take a while) ...\n");
-  TPCHTableGenerator{scale_factor, ClusteringConfiguration::None, chunk_size}.generate_and_store();
+  const auto config = std::make_shared<BenchmarkConfig>(chunk_size);
```
Bouncner, line 544: What about a second constructor in BenchmarkConfig?

Bouncner: Or third ... taking chunk_size and bool for binary loading?

Bouncner (approved):
LGTM.

src/benchmark/tpch_data_micro_benchmark.cpp
dey4ss: Ah, it gets kind of clearer with the surrounding code:
https://github.com/hyrise/hyrise/blob/59e469fcdd16e941ebe241406ca9603bac8c1a07/src/benchmark/tpch_data_micro_benchmark.cpp#L36-L44

So it is basically just a helper here. We could either explicitly set the declared default spec via the member of `BenchmarkConfig` or just print the member of the default config. I'd just go for the second version, as someone who touches this and wants to try out a different encoding here should also easily find out how to do that.

Bouncner (approved):

PR #2650, opened 2024-06-12
Mini-PR: flush server start message

Bouncner:
This is a minimal PR that ensures that the server flushes the message "Server started at ..." when the server is starting.

We use this message in a couple of benchmarking scripts. The scripts wait for this message to be sure that the server has started. Without flushing, scripts sometimes wait forever.


dey4ss (approved):
If this solves waaiting issues, zünd. Do you have a concrete example?

We just changed using `std::endl` to `\n` recently due to clang-tidy suggestions that assumes the STLs' console outputs to be line-buffered. However, if that means we cannot rely on that, we might want to review that again.

PR #2658, opened 2024-08-29
Remove `AbstractCardinalityEstimator`

dey4ss:
As discussed, this PR removes the abstract super class of the `CardinalityEstimator`. The caching interface implemented in the abstract class mixes a lot of implementation details with the concrete implementation (e.g., that statistics are cached). As we will not have a different cardinality estimation implementation in the near future, we decided to remove the abstract interface.


src/lib/statistics/cardinality_estimator.hpp
```diff
@@ -31,23 +32,50 @@ class ValidateNode;
 class WindowNode;
 
 /**
- * Hyrise's default, statistics-based cardinality estimator.
+ * Hyrise's statistics-based cardinality estimator.
```
Bouncner, line 35: Since you gathered a lot of knowledge, is there something more we would to say here as part of the overview?

Something brief about caching, diamonds, "instances" etc.?

src/lib/statistics/cardinality_estimator.hpp
```diff
@@ -31,23 +32,50 @@ class ValidateNode;
...
+   * Promises to this CardinalityEstimator that it will only be used to estimate Cardinalities of plans that consist
+   * of the Vertices and Predicates in @param JoinGraph. This enables using the JoinGraphStatisticsCache during
+   * Cardinality estimation.
+   */
+  void guarantee_join_graph(const JoinGraph& join_graph);
+
+  /**
+   * For increased cardinality estimation performance:
+   * Promises to this CardinalityEstimator that it will only be used to estimate bottom-up
+   * constructed plans. That is, the Cost/Cardinality of a node, once constructed, never changes.
```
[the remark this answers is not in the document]
Bouncner: I think 1 example/counter example would be very helpful here. Or do we have that somewhere else?

Bouncner (approved):
Only tiny notes.

src/lib/optimizer/strategy/in_expression_rewrite_rule.hpp
```diff
@@ -44,9 +44,9 @@ class InExpressionRewriteRule : public AbstractRule {
  protected:
   void _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const override;
 
-  std::shared_ptr<AbstractCardinalityEstimator> _cardinality_estimator() const;
+  std::shared_ptr<CardinalityEstimator> _cardinality_estimator() const;
 
-  mutable std::shared_ptr<AbstractCardinalityEstimator> _cardinality_estimator_internal;
+  mutable std::shared_ptr<CardinalityEstimator> _cardinality_estimator_internal;
```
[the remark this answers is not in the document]
dey4ss: We usually get a fresh instance when we use the estimator to ensure nothing will be cached. Does not make sense to have a member variable here though. Gonna remove it.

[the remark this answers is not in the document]
dey4ss: Wait, we don't even enable caching in this rule, that makes no sense at all.

src/lib/statistics/cardinality_estimator.hpp
dey4ss: Added some comments here.

src/lib/statistics/join_graph_statistics_cache.cpp
```diff
@@ -19,13 +19,13 @@
...
-  for (auto vertex_idx = size_t{0}; vertex_idx < join_graph.vertices.size(); ++vertex_idx) {
+  const auto vertex_count = join_graph.vertices.size();
+  auto vertex_indices = VertexIndexMap{vertex_count};
+  for (auto vertex_idx = size_t{0}; vertex_idx < vertex_count; ++vertex_idx) {
     vertex_indices.emplace(join_graph.vertices[vertex_idx], vertex_idx);
   }
 
-  PredicateIndexMap predicate_indices;
-
+  auto predicate_indices = PredicateIndexMap{};
```
Bouncner, line 28: Can' we reserve here as well?

src/lib/statistics/cardinality_estimator.hpp
```diff
@@ -61,17 +71,42 @@ class CardinalityEstimator {
...
+   *              |
+   *        PredicateNode A
+   *              |
+   *              | 100'000
+   *              |
+   *       StoredTableNode R
+   *
+   * First, we go top-down. As we already estimated nodes further up in the LQP, the statistics/cardinalities of the
+   * PredicateNodes have been cached. PredicateNode B yields a smaller cardinality even though it has a worse
+   * selectivity, and we would falsely move it below PredicateNode A.
```
Bouncner, line 106: I don't get it. :D
How does a wrong ordering relate to the cardinality estimation here? It's not clear how caching has an impact here.

Bouncner (commented):
Mini issue and one comment isn't clear to me yet.

src/lib/statistics/cardinality_estimator.hpp
```diff
@@ -32,7 +32,17 @@ class ValidateNode;
...
+ * Hyrise's statistics-based cardinality estimator. It mostly relies on histograms that are taken from the base tables
+ * and adapted (sliced, scaled) according to join and scan predicates along the query plan. The estimation process
+ * happens recursively.
+ *
+ * To increase estimation performance, we cache already estimated statistics for (sub-)plans that are guaranteed not to
+ * change anymore. This is beneficial if we call the CardinalityEstimator multiple times for the same LQP (e.g., during
+ * join ordering, predicate reordering, etc.).
+ * Even if we cannot cache statistics across multiple estimation invocations, we maintain a cache for a single call.
+ * This helps if we have diamonds in the query plan (e.g., after predicate splitup or with semi-join reductions).
+ * When allowing the CardinalityEstimator to use caching, it is best practice to get a fresh instance. Thus, the filled
```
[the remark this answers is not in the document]
dey4ss: True, the comment is older thasn the assert :D

src/lib/statistics/cardinality_estimator.hpp
dey4ss: To get estimate the statistics of a node, we must know the statistics of input nodes, so we recursively estimate them. With caching, we store all estimated statistics. That's it already. When going top-down, ALL statistics will be saved once we estimate the root node. When we want to reorder the predicates, we always get the statistics that were estimated in the original order (and not, as we would like to, directly on the node below the predicate chain).

src/lib/statistics/join_graph_statistics_cache.cpp
dey4ss: I thought the same, but I was not sure about the value to use. The index map is an `ExpressionUnorderedMap` that only depends on the predicates. `join_graph.edges.size()`, however, could be a rough estimation, so let's use that.

src/lib/statistics/join_graph_statistics_cache.cpp
```diff
@@ -19,13 +19,13 @@
 namespace hyrise {
 
 JoinGraphStatisticsCache JoinGraphStatisticsCache::from_join_graph(const JoinGraph& join_graph) {
-  VertexIndexMap vertex_indices;
-  for (auto vertex_idx = size_t{0}; vertex_idx < join_graph.vertices.size(); ++vertex_idx) {
+  const auto vertex_count = join_graph.vertices.size();
+  auto vertex_indices = VertexIndexMap{vertex_count};
```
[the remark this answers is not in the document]
dey4ss: I guess both are in essence a guess when it comes to hash collisions. My intuition would be that if we have as many buckets as we have elements, with an ideal hash function each value should be in its own bucket. With `reserve()` (`reserves space for at least the specified number of elements`), I would assume that this takes the fill factor for buckets into account and generates fewer buckets.

src/lib/statistics/cardinality_estimator.hpp
dey4ss:
```c++
   * First, we go top-down. As we already estimated nodes further up in the LQP, the statistics/cardinalities of the
   * PredicateNodes have been cached and will stay unchanged whenever we try to re-estimate them. When placed directly
   * above the StoredTableNode R, PredicateNode B yields a smaller cardinality than PredicateNode A even though it has a
   * worse selectivity, and we would falsely move it below PredicateNode A.
   *
   * Second, we go bottom-up. No statistics have been cached yet, we correctly estimate the cardinality based on the
   * selectivity, and we preserve the predicate order before continuing to the nodes further up. Note that during
   * predicate reordering, we do not cache the statistics of the PredicateNodes when deciding on their placement, but
   * only when we estimate nodes above.
```
better?

Bouncner: Let's talk in person.

PR #2664, opened 2024-11-25
Parallelize chunk pruning statistics generation for full table

j-hellenberg:
Parallelize generation of chunk pruning statistics when doing so for a full table.

Currently, statistics-generation is only fast when generating benchmark data from scratch (not when loading the cached binary files from disk).

This is because we immediately invoke the statistics generation when encoding a chunk (which is done in a parallel fashion). However, when chunks are already encoded, we fall back to the `generate_chunk_pruning_statistics(const std::shared_ptr<Table>& table)` for the whole table, which did not use any parallelization at all.

This step reduces the time required for the `encoding/statistics generation` step on loaded data from about 14.5s to 2.5s for SF 20.

### DIFF
--- a/src/benchmarklib/benchmark_table_encoder.cpp
+++ b/src/benchmarklib/benchmark_table_encoder.cpp
@@ -136,10 +136,12 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
   }
   Hyrise::get().scheduler()->schedule_and_wait_for_tasks(jobs);
 
-  // Note: Pruning statistics might have already been generated during chunk encoding, in which case this call
+  // Note: Pruning statistics might have already been generated during chunk encoding above, in which case this call
   // is (almost) a NoOp.
-  // However, this does not apply when loading benchmark data that we have cached, because that data is encoded already.
-  // We still need to generate the statistics in that case, though, because we don't cache them.
+  // However, encoding might not happen when loading benchmark data that we have stored in binary, because that data is
+  // encoded already. Re-encoding will only be necessary in case the benchmark encoding config has changed, and might
+  // only affect certain chunks then.
+  // If encoding did not happen for loaded data, we still need to generate the statistics, because we do not cache those.
   generate_chunk_pruning_statistics(table);
 
   return encoding_performed;
--- a/src/benchmarklib/benchmark_table_encoder.cpp
+++ b/src/benchmarklib/benchmark_table_encoder.cpp
@@ -136,6 +136,10 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
   }
   Hyrise::get().scheduler()->schedule_and_wait_for_tasks(jobs);
 
+  // Note: Pruning statistics might have already been generated during chunk encoding, in which case this call
+  // is (almost) a NoOp.
+  // However, this does not apply when loading benchmark data that we have cached, because that data is encoded already.
+  // We still need to generate the statistics in that case, though, because we don't cache them.
   generate_chunk_pruning_statistics(table);
 
   return encoding_performed;
--- a/src/benchmarklib/benchmark_table_encoder.cpp
+++ b/src/benchmarklib/benchmark_table_encoder.cpp
@@ -136,12 +136,11 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
   }
   Hyrise::get().scheduler()->schedule_and_wait_for_tasks(jobs);
 
-  // Note: Pruning statistics might have already been generated during chunk encoding above, in which case this call
-  // is (almost) a NoOp.
-  // However, encoding might not happen when loading benchmark data that we have stored in binary, because that data is
-  // encoded already. Re-encoding will only be necessary in case the benchmark encoding config has changed, and might
-  // only affect certain chunks then.
-  // If encoding did not happen for loaded data, we still need to generate the statistics, because we do not cache those.
+  // Note: Chunk pruning statistics might have already been generated during chunk encoding above, in which case this
+  // call is (almost) a NoOp. However, encoding might not happen when loading binary table data, because that data is
+  // written after encoding, thus skipping statistic generation. Re-encoding will only be necessary in case different
+  // encoding schemes are requested by the user and might only affect certain chunks then. If no re-encoding happened,
+  // we still need to generate chunk pruning statistics, because those are not cached as part of the binary table data.
   generate_chunk_pruning_statistics(table);
 
   return encoding_performed;
--- a/src/lib/statistics/generate_pruning_statistics.cpp
+++ b/src/lib/statistics/generate_pruning_statistics.cpp
@@ -6,17 +6,16 @@
 #include <vector>
 
 #include "boost/sort/sort.hpp"
-
 #include "hyrise.hpp"
 #include "resolve_type.hpp"
+#include "scheduler/abstract_task.hpp"
+#include "scheduler/job_task.hpp"
 #include "statistics/attribute_statistics.hpp"
 #include "statistics/statistics_objects/distinct_value_count.hpp"
 #include "statistics/statistics_objects/equal_distinct_count_histogram.hpp"
 #include "statistics/statistics_objects/min_max_filter.hpp"
 #include "statistics/statistics_objects/range_filter.hpp"
 #include "statistics/table_statistics.hpp"
-#include "scheduler/abstract_task.hpp"
-#include "scheduler/job_task.hpp"
 #include "storage/chunk.hpp"
 #include "storage/create_iterable_from_segment.hpp"
 #include "storage/table.hpp"
@@ -44,13 +43,16 @@ void create_pruning_statistics_for_segment(AttributeStatistics<T>& segment_stati
 }  // namespace
 
 namespace hyrise {
+bool pruning_statistics_should_be_generated_for_chunk(const std::shared_ptr<Chunk>& chunk) {
+  // We do not generate statistics for chunks as long as they are mutable.
+  // Also, pruning statistics should be stable no matter what encoding or sort order is used.
+  // Hence, when they are present they are up to date, and we can skip the recreation.
+  return chunk && !chunk->is_mutable() && !chunk->pruning_statistics();
+}
 
 void generate_chunk_pruning_statistics(const std::shared_ptr<Chunk>& chunk) {
-  if (chunk->pruning_statistics()) {
-    // Pruning statistics should be stable no matter what encoding or sort order is used. Hence, when they are present
-    // they are up to date and we can skip the recreation.
-    return;
-  }
+  DebugAssert(pruning_statistics_should_be_generated_for_chunk(chunk),
+              "Method should only be called for chunks that need it.");
 
   auto chunk_statistics = ChunkPruningStatistics{chunk->column_count()};
 
@@ -97,7 +99,7 @@ void generate_chunk_pruning_statistics(const std::shared_ptr<Table>& table) {
   for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
     const auto chunk = table->get_chunk(chunk_id);
 
-    if (!chunk || chunk->is_mutable() || chunk->pruning_statistics()) {
+    if (!pruning_statistics_should_be_generated_for_chunk(chunk)) {
       continue;
     }
 
--- a/src/lib/statistics/generate_pruning_statistics.cpp
+++ b/src/lib/statistics/generate_pruning_statistics.cpp
@@ -6,6 +6,8 @@
 #include <unordered_set>
 #include <vector>
 
+#include "boost/sort/sort.hpp"
+
 #include "hyrise.hpp"
 #include "resolve_type.hpp"
 #include "statistics/attribute_statistics.hpp"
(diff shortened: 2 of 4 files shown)


src/lib/statistics/generate_pruning_statistics.cpp
```diff
@@ -86,15 +90,22 @@ void generate_chunk_pruning_statistics(const std::shared_ptr<Chunk>& chunk) {
 
 void generate_chunk_pruning_statistics(const std::shared_ptr<Table>& table) {
```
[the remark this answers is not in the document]
Bouncner: While currently not being the case, the chunk encoder should also be used when mutable chunks are reaching their limit and can be encoded (thus it is not part of the benchmark lib).

I’d like to get the numbers separated, but I do not think it is worth the effort right now.

src/lib/statistics/generate_pruning_statistics.cpp
```diff
@@ -65,14 +69,14 @@ void generate_chunk_pruning_statistics(const std::shared_ptr<Chunk>& chunk) {
...
-        std::unordered_set<ColumnDataType> values;
+        auto values = std::unordered_set<ColumnDataType>{};
         iterable.for_each([&](const auto& value) {
           // we are only interested in non-null values
           if (!value.is_null()) {
             values.insert(value.value());
           }
         });
-        pmr_vector<ColumnDataType> dictionary{values.cbegin(), values.cend()};
+        auto dictionary = pmr_vector<ColumnDataType>{values.cbegin(), values.cend()};
```
Bouncner, line 79: Can we use pdq sort here?

src/lib/statistics/generate_pruning_statistics.cpp
```diff
@@ -86,15 +90,22 @@ void generate_chunk_pruning_statistics(const std::shared_ptr<Chunk>& chunk) {
...
-      continue;
-    }
+      if (!chunk || chunk->is_mutable()) {
+        return;
+      }
 
-    generate_chunk_pruning_statistics(chunk);
+      generate_chunk_pruning_statistics(chunk);
+    };
+    jobs.emplace_back(std::make_shared<JobTask>(generate_statistics));
```
Bouncner, line 106: Will not be measurable but we could put that all into a single line. Make shared on existing data adds another allocation (data block and control block not adjacent) and then we need to copy with emplace instead of creating in place

src/lib/statistics/generate_pruning_statistics.cpp
```diff
@@ -86,15 +92,21 @@ void generate_chunk_pruning_statistics(const std::shared_ptr<Chunk>& chunk) {
...
+  auto jobs = std::vector<std::shared_ptr<AbstractTask>>{};
+  jobs.reserve(chunk_count);
+
   for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
-    const auto chunk = table->get_chunk(chunk_id);
+    const auto chunk = table->get_chunk(ChunkID{chunk_id});
 
     if (!chunk || chunk->is_mutable()) {
-      continue;
+      return;
```
dey4ss, line 102: I think this should stay `continue`. Otherwise, we don't do anything if the table's final chunk is mutable or there is a physically deleted chunk. Remember that this is also called when adding a table to the storage manager, not only during initial loading.

j-hellenberg: You're absolutely right, stupid me forgot to change it back when moving it out of the lambda. Oops :see_no_evil:

src/benchmarklib/benchmark_table_encoder.cpp
```diff
@@ -136,6 +136,10 @@ bool BenchmarkTableEncoder::encode(const std::string& table_name, const std::sha
   }
   Hyrise::get().scheduler()->schedule_and_wait_for_tasks(jobs);
 
+  // Note: Pruning statistics might have already been generated during chunk encoding, in which case this call
+  // is (almost) a NoOp.
+  // However, this does not apply when loading benchmark data that we have cached, because that data is encoded already.
+  // We still need to generate the statistics in that case, though, because we don't cache them.
```
Bouncner, line 142: Can you use "store binary tables" instead of caching? It is is some kind of caching, but written to disk. I think being more explicit helps here. Also, binary tables MIGHT be encoded correctly already, they might also be encoded with the "wrong" encoding.

And ... "do not" instead of "don't". :)

Bouncner (commented):
Just a note on the comment

j-hellenberg:
Btw: Is there a way to abort outdated CI pipelines when adding multiple commits in quick succession?

src/lib/statistics/generate_pruning_statistics.cpp
```diff
@@ -86,15 +92,21 @@ void generate_chunk_pruning_statistics(const std::shared_ptr<Chunk>& chunk) {
 
 void generate_chunk_pruning_statistics(const std::shared_ptr<Table>& table) {
   const auto chunk_count = table->chunk_count();
+  auto jobs = std::vector<std::shared_ptr<AbstractTask>>{};
+  jobs.reserve(chunk_count);
+
   for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
     const auto chunk = table->get_chunk(chunk_id);
 
     if (!chunk || chunk->is_mutable()) {
```
dey4ss, line 101: I think it's worth to add the condition from L50 also here. Otherwise, we schedule a task that does a no-op for each chunk when adding the table to the storage manager after encoding.

Bouncner:
> Btw: Is there a way to abort outdated CI pipelines when adding multiple commits in quick succession?

It might take a few seconds but previous commits should be automatically aborted.

src/lib/statistics/generate_pruning_statistics.cpp
j-hellenberg: How to deal with the fact that L50 has a comment? I'm always a bit reluctant of copying comments, because they get out of sync so easily

dey4ss: I agree. Do we ever call the version for a single chunk directly? And if so, one could argue that the caller could (should) do the check. So I'm fine with moving both the check and the comment to here.

Bouncner: The single-chunk call is the one done in the ChunkEncoder (when we "correctly parallelize").

dey4ss: O sure, we call this e.g. in the chunk encoder. I guess then it's fine to add the check there as well.

Bouncner:
> 14.5s to 2.5s

What machine was that? This is probably still limited by the grouping of size 10 in the master, right?

Bouncner:
I measured this PR on a 64 core server. I was curious why I don't see the expected improvements.

First, the results.

Master:
```
-  Processing 'lineitem' - no encoding necessary (1 min 26 s)
- Encoding tables and generating pruning statistic done (1 min 26 s)
```

This PR:
```
-  Processing 'lineitem' - no encoding necessary (9 s 380 ms)
- Encoding tables and generating pruning statistic done (9 s 381 ms)
```

This PR with disabled node grouping:
```
-  Processing 'lineitem' - no encoding necessary (2 s 257 ms)
- Encoding tables and generating pruning statistic done (2 s 258 ms)
```

So first, we see that we "only" improve by a factor of 8, even though we could use all the cores. The reason is the current node grouping. So no table uses more than 6 cores concurrently. Since we have on a few "heavy" tables, the  number of cores used is rather limited.
When disabling the cores, we see a much better result, roughly 35x improvement. Not 64x, but keep in mind that we previously encoded the tables in parallel (for the first third of the time, we encode orders and lineitem in parallel).


Here is the core usage over time:
image

src/lib/statistics/generate_pruning_statistics.cpp
j-hellenberg: So I extracted the check into a separate method. Thoughts?

dey4ss (approved):

Bouncner (changes requested):
Need to have another look

src/lib/statistics/generate_pruning_statistics.cpp
```diff
@@ -65,15 +73,15 @@ void generate_chunk_pruning_statistics(const std::shared_ptr<Chunk>& chunk) {
       } else {
         // if we have a generic segment we create the dictionary ourselves
         auto iterable = create_iterable_from_segment<ColumnDataType>(typed_segment);
-        std::unordered_set<ColumnDataType> values;
+        auto values = std::unordered_set<ColumnDataType>{};
         iterable.for_each([&](const auto& value) {
           // we are only interested in non-null values
```
Bouncner, line 78:
```suggestion
          // We are only interested in non-null values.
```

src/lib/statistics/generate_pruning_statistics.cpp
```diff
@@ -65,15 +73,15 @@ void generate_chunk_pruning_statistics(const std::shared_ptr<Chunk>& chunk) {
       } else {
         // if we have a generic segment we create the dictionary ourselves
```
Bouncner, line 74:
```suggestion
        // If we have a generic segment, we create the dictionary ourselves.
```

src/lib/statistics/generate_pruning_statistics.cpp
```diff
@@ -65,15 +73,15 @@ void generate_chunk_pruning_statistics(const std::shared_ptr<Chunk>& chunk) {
       } else {
```
Bouncner, line 73: I am no fan of `// TODO(...)` comments, but can you please add one before line 69?
`// TODO(anyone): use dictionary-optimized path for FixedStringDictionarySegments as well.`

Maybe I can implement that tomorrow, but I am not sure.

src/lib/statistics/generate_pruning_statistics.cpp
```diff
@@ -39,13 +44,16 @@ void create_pruning_statistics_for_segment(AttributeStatistics<T>& segment_stati
...
+}
 
 void generate_chunk_pruning_statistics(const std::shared_ptr<Chunk>& chunk) {
-  if (chunk->pruning_statistics()) {
-    // Pruning statistics should be stable no matter what encoding or sort order is used. Hence, when they are present
-    // they are up to date and we can skip the recreation.
-    return;
-  }
+  DebugAssert(pruning_statistics_should_be_generated_for_chunk(chunk),
+              "Method should only be called for chunks that need it.");
```
Bouncner, line 56: Well, no chunk really "needs" it. Maybe the optimizer "needs" it. I think "qualifying chunks" might be enough as the function above is very clear.

src/lib/statistics/generate_pruning_statistics.hpp
```diff
@@ -8,6 +8,11 @@ namespace hyrise {
 class Chunk;
 class Table;
 
+/**
+ * Check whether we need to generate pruning statistics for a chunk.
```
Bouncner, line 12: See above. We do not "need" to do that.

src/lib/statistics/generate_pruning_statistics.hpp
```diff
@@ -8,6 +8,11 @@ namespace hyrise {
...
+ */
+bool pruning_statistics_should_be_generated_for_chunk(const std::shared_ptr<Chunk>& chunk);
```
Bouncner, line 14: I don't like the naming.  `chunk_pruning_statistics_can_be_added` sucks as well. Can we find something better?

dey4ss: `finalized_chunk_lacks_(pruning_)statistics`? Or negate the condition and call it `finalized_chunk_has_(pruning_)statistics`?

Bouncner: Problem with `immutable_chunk_lacks_pruning_statistics` (switched to immutable) is that it is not really clear what the result for an mutable chunk should be.
"chunk qualifies for getting pruning stat" ish?

But we also shouldn't overthink this here. Hmpf!

dey4ss: Can we conduct a representative survey?

But what's against `immutable_chunk_has_pruning_statistics`? Or `chunk_(is)_immutable_and_has_pruning_statistics` (quite long though). To me the first one sounds okay, but I'd also say let's not make this a longer discussion than necessary.

Bouncner: `immutable_chunk_has_pruning_statistics` sounds strange as it is not clear what the result is when we pass a mutable chunk. Sure, with knowledge one knows that those chunks do not have any statistics, but I am curious if we can find a really accurate name. :(

> Or chunk_(is)_immutable_and_has_pruning_statistics (quite long though) [...]

`chunk_is_mutable_or_has_pruning_statistics`? Dann ist klar, dass man skippen kann. Ist ja kein wirkliches "and".

dey4ss: But yes, I'm five with your suggestion

Bouncner: Why are you talking in German?

Bouncner: Not sure if I really like `chunk_is_mutable_or_has_pruning_statistics`, but if we can agree on it, let's just take it?

My problem with `chunk_(is)_immutable_and_has_pruning_statistics` is, dass ich das ja gar nicht nutzen kann, oder? Wenn er immutable ist ohne Stats (return false), möchte ich die hinzufügen. Wenn er mutable ist (return false), kann ich nichts machen. Beide male ist das Ergebnis aber halt false.

j-hellenberg: I'd propose using `is_immutable_chunk_without_pruning_statistics`

Bouncner (approved):
LGTM.

PR #2665, opened 2024-11-29
Console: Restore default settings during `reset`

dey4ss:
Just a small PR to restore the settings (scheduler, binary caching) and to fix the segfault when exiting the Console with scheduler.

### DIFF
--- a/scripts/test/hyriseConsole_test.py
+++ b/scripts/test/hyriseConsole_test.py
@@ -129,10 +129,14 @@ def main():
     console.sendline("print nation")
     console.expect("=== Chunk 3 ===")
 
-    # Clear generated tables.
+    # Clear all generated tables and restore default configuration.
     console.sendline("reset")
     console.sendline("select * from meta_tables")
     console.expect("0 rows total")
+
+    # Make sure to turn off binary table caching again.
+    console.sendline("setting binary_caching off")
+    console.expect("Binary caching turned off")
 
     # Test TPC-C generation. We also tried different benchmarks here. SSB and JCC-H are not thread-safe due to the
     # external data generator, leading to problems since this test is executed multiple times. We consider TPC-DS data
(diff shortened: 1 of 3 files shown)


scripts/test/hyriseConsole_test.py
```diff
@@ -129,11 +129,15 @@ def main():
     console.sendline("print nation")
     console.expect("=== Chunk 3 ===")
 
-    # Clear generated tables.
+    # Clear all generated tables and restore default configuration.
     console.sendline("reset")
     console.sendline("select * from meta_tables")
     console.expect("0 rows total")
 
+    # Make sure to turn off binary table caching again.
```
Bouncner, line 137: Why?

src/bin/console/console.cpp
```diff
@@ -1037,12 +1036,14 @@ int Console::_unload_plugin(const std::string& input) {
 
 int Console::_reset() {
   _rollback();
-  _lqp_cache->clear();
-  _pqp_cache->clear();
 
   Hyrise::reset();
-  Hyrise::get().default_pqp_cache = _pqp_cache;
-  Hyrise::get().default_lqp_cache = _lqp_cache;
+  Hyrise::get().default_pqp_cache = std::make_shared<SQLPhysicalPlanCache>();
```
Bouncner, line 1041: What is the different between `clear()` and setting an entirely new one?

scripts/test/hyriseConsole_test.py
dey4ss: It's there to ensure that no binaries are written/read at the same time because we call the console tests in parallel from the project root for clang-release and clang-debug/gcc-debug.

Bouncner: Got it.

Bouncner (approved):

PR #2666, opened 2024-12-01
Parallelize loading of benchmark data

dey4ss:
Small PR that adds a job for each benchmark table that is loaded from a file (either cached binary or CSV file for file-based benchmarks). There should not be differences for the latter case (heavy parts of CSV parsing are already parallel).

However, parsing the binaries in single-threaded, so doing this in parallel for multiple tables is advantageous for benchmarks where not only a single table dominates the loading time. Thus, the loading times decrease significantly for TPC-DS, JOB, and TPC-C.

Though loading from binaries is by far not the slowest part of benchmark preparation, it might still be noticeable for larger SFs.

### DIFF
--- a/src/benchmarklib/abstract_table_generator.cpp
+++ b/src/benchmarklib/abstract_table_generator.cpp
@@ -456,20 +456,30 @@ bool AbstractTableGenerator::_all_chunks_sorted_by(const std::shared_ptr<Table>&
 std::unordered_map<std::string, BenchmarkTableInfo> AbstractTableGenerator::_load_binary_tables_from_path(
     const std::string& cache_directory) {
   auto table_info_by_name = std::unordered_map<std::string, BenchmarkTableInfo>{};
+  const auto table_files = list_directory(cache_directory);
+  auto jobs = std::vector<std::shared_ptr<AbstractTask>>{};
+  jobs.reserve(table_files.size());
 
-  for (const auto& table_file : list_directory(cache_directory)) {
+  for (const auto& table_file : table_files) {
     const auto table_name = table_file.stem();
-    std::cout << "-  Loading table '" << table_name.string() << "' from cached binary " << table_file.relative_path();
+    table_info_by_name[table_name] = BenchmarkTableInfo{};
 
-    auto timer = Timer{};
-    auto table_info = BenchmarkTableInfo{};
-    table_info.table = BinaryParser::parse(table_file);
-    table_info.loaded_from_binary = true;
-    table_info.binary_file_path = table_file;
-    table_info_by_name[table_name] = table_info;
+    jobs.emplace_back(std::make_shared<JobTask>([table_name, table_file, &table_info_by_name]() {
+      auto message = std::stringstream{};
+      message << "-  Loaded table '" << table_name.string() << "' from cached binary " << table_file.relative_path();
 
-    std::cout << " (" << timer.lap_formatted() << ")\n";
+      auto timer = Timer{};
+      auto& table_info = table_info_by_name[table_name];
+      table_info.table = BinaryParser::parse(table_file);
+      table_info.loaded_from_binary = true;
+      table_info.binary_file_path = table_file;
+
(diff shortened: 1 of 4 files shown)


Bouncner:
Do you have runtime numbers?

src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -456,21 +456,31 @@ bool AbstractTableGenerator::_all_chunks_sorted_by(const std::shared_ptr<Table>&
...
-    table_info.loaded_from_binary = true;
-    table_info.binary_file_path = table_file;
-    table_info_by_name[table_name] = table_info;
+    jobs.emplace_back(std::make_shared<JobTask>([table_name, table_file, &table_info_by_name]() {
+      auto message = std::stringstream{};
+      message << "-  Loaded table '" << table_name.string() << "' from cached binary " << table_file.relative_path();
 
-    std::cout << " (" << timer.lap_formatted() << ")\n";
+      auto timer = Timer{};
+      auto& table_info = table_info_by_name[table_name];
```
Bouncner, line 472: This can race, right?

Let's say line 465 causes an resizing and you are writing here to the reference which you got just immediately before the resize.

src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -456,21 +456,31 @@ bool AbstractTableGenerator::_all_chunks_sorted_by(const std::shared_ptr<Table>&
...
+      table_info.table = BinaryParser::parse(table_file);
+      table_info.loaded_from_binary = true;
+      table_info.binary_file_path = table_file;
+
+      message << " (" << timer.lap_formatted() << ")\n";
+      std::cout << message.str();
```
Bouncner, line 478: Can we flush the lines? I just checked SF 100 for the other data loading PR and often (if not all the time) all import message are coming altogether. Especially for large SFs, it would be helpful to get them immediately.

src/benchmarklib/abstract_table_generator.cpp
Bouncner: Maybe creating `table_info_by_name` later with the known number of tables might be sufficient?

src/benchmarklib/file_based_table_generator.cpp
```diff
@@ -82,33 +87,40 @@ std::unordered_map<std::string, BenchmarkTableInfo> FileBasedTableGenerator::gen
...
+          table_info.table = CsvParser::parse(*table_info.text_file_path, _benchmark_config->chunk_size);
+        } else {
+          Fail("Unknown textual file format. This should have been caught earlier.");
+        }
       }
-    }
 
-    std::cout << " (" << table_info.table->row_count() << " rows; " << timer.lap_formatted() << ")\n";
+      message << " (" << table_info.table->row_count() << " rows; " << timer.lap_formatted() << ")\n";
+      std::cout << message.str();
```
Bouncner, line 119: Flush?

And if you agree, maybe the same for encoding and storage-manager-adding? (have not checked if necessary)

PS: I think clang-tidy still prefers `'\n' << std::flush;` over `std::endl` (somehow stupid, but I somehow agree). What's your take on this?

src/benchmarklib/abstract_table_generator.cpp
dey4ss: Oh, true 😅 
I now create the map first out of the listed file paths and then iterate over the map.

dey4ss:
> Do you have runtime numbers?

nemea won't let me compile atm to try with larger SFs. TPC-H and SSB do not change a lot (lineitem/lineorder totally dominate).
Just for an intuition (SF 10 locally, with other stuff running):
TPC-H 6.2s --> 5.3s
TPC-DS 3.3s --> < 1.5s
TPC-C 660ms --> 413ms
SSB 1.1s --> 1.1s
JOB 2.1s --> 0.7s

Bouncner (approved):

Issue #2667, opened 2024-12-05
`hyriseBenchmarkJoinOrder` creates nested folder structure


tjeyy:
The folder name `imdb_data` is passed to the `setup_imdb.py` script of the Join-Order-Benchmark to define the extraction location for the downloaded `imdb_data.zip`. However, since the `.zip` file itself contains an `imdb_data` folder, the extracted data ends up in `project_root/imdb_data/imdb_data/`.  

This creates an unnecessary nested folder structure, which causes several problems:  
1. **Redundant Pathing**: The paths written to `table_info.text_file_path` by the `FileBasedTableGenerator` include the extra directory level.  
2. **Cache Misalignment**: The `AbstractTableGenerator`, which relies on these paths, cannot locate the cached table binaries during subsequent benchmark runs. As a result, the dataset is redownloaded and regenerated unnecessarily.

PR #2668, opened 2024-12-05
Exchange `imdb_data.zip` and change its host

tjeyy:
Fixes the problem described in #2667.
Alternatively, the `imdb_data` folder could be removed from the zip-file.
What are your opinions on this?

### DIFF
--- a/scripts/setup_imdb.py
+++ b/scripts/setup_imdb.py
@@ -6,6 +6,7 @@
 
 import hashlib
 import os
+import shutil
 import sys
 import urllib.request
 import zipfile
@@ -16,9 +17,14 @@ def clean_up(including_table_dir=False):
         os.remove(FILE_NAME)
 
     if including_table_dir and os.path.exists(table_dir):
-        for file in os.listdir(table_dir):
-            os.remove("./%s/%s" % (table_dir, file))
-        os.rmdir(table_dir)
+        try:
+            shutil.rmtree(table_dir)
+        except PermissionError as e:
+            print(f"PermissionError: Unable to remove {table_dir}. Please check file permissions.")
+            raise
+        except Exception as e:
+            print(f"An error occurred while cleaning up {table_dir}: {e}")
+            raise
 
 
 def is_setup():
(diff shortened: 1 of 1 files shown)


scripts/setup_imdb.py
```diff
@@ -115,7 +113,7 @@ def is_setup():
 
 hash_dl = hash_md5.hexdigest()
 
-if hash_dl != "24bb992f97dad7b83fd4009e312ddd40":
+if hash_dl != "7c2e84c64126897267d1cf745d47bc9b":
     print("  Aborting. MD5 checksum mismatch. Cleaning up.")
     clean_up()
     sys.exit(2)
```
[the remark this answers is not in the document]
dey4ss: True, we could call `is_setup()` after extracting to ensure that.

scripts/setup_imdb.py
```diff
@@ -132,6 +136,11 @@ def is_setup():
     clean_up(including_table_dir=True)
     sys.exit(3)
 
+if not is_setup():
+    print("  Aborting. Unzipping did not result in a correct imdb_files-setup. Cleaning up.")
+    clean_up(including_table_dir=True)
+    sys.exit(3)
```
dey4ss, line 142:
```suggestion
    sys.exit(4)
```

scripts/setup_imdb.py
```diff
@@ -16,9 +17,14 @@ def clean_up(including_table_dir=False):
...
-            os.remove("./%s/%s" % (table_dir, file))
-        os.rmdir(table_dir)
+        try:
+            shutil.rmtree(table_dir)
+        except PermissionError as e:
+            print(f"PermissionError: Unable to remove {table_dir}. Please check file permissions.")
+            raise
+        except Exception as e:
+            print(f"An error occurred while cleaning up {table_dir}: {e}")
+            raise
```
dey4ss, line 27: While I like condensing the deletion of the directory into a single call, I am not sure about the value added by catching. I think the errors are telling enough (esp. `PermissionError`). In the other places, we only catch to cleanup (so hopefully re-trying downloads the data again and that already helps), but IMO we can simply crash if the cleanup itself does not work.

dey4ss (commented):
Just two small things

scripts/setup_imdb.py
tjeyy: Agreed.

Bouncner:
Premature zündung.

tjeyy:
Auto-Merge :/
I'll put the suggested change in my next PR ;)

Issue #240, opened 2017-09-14
REPL: Pagination, History broken for OSX


mrks:
- [x] The history should be saved before anything is executed. If the DB crashes, the last command is not lost.
- [ ] The print operator should have an optional {from, to} pair that is used for pagination in the REPL console

PR #2675, opened 2025-01-06
Correctly parse named columns for joins

Rob2U:
The PR integrates the changes to the sql-parser (see [hyrise/sql-parser #240](https://github.com/hyrise/sql-parser/pull/240)) into the Hyrise join behavior.
In short, this allows a list of identifiers (a.k.a column names) when joins name columns with USING (...). The result should only contain the columns that were used for the join once.
This behavior is in line with the SQL standard and PostgreSQL. 

Specifically, the sql-parser parses a join into the updated struct:
```C++
struct JoinDefinition {
  JoinDefinition();
  virtual ~JoinDefinition();

  TableRef* left;
  TableRef* right;
  Expr* condition;

  JoinType type;
  std::vector<char*>* namedColumns;
};
```
We execute a `_join_using_named_columns` when only namedColumns is set and condition is not. We fail if both are set (due to ambiguity).

### DIFF
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -741,7 +741,7 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_table_ref(const hsql::
         return _translate_natural_join(*hsql_table_ref.join);
       } else {
         if (hsql_table_ref.join->namedColumns && !hsql_table_ref.join->namedColumns->empty()) {
-          return _translate_namedColumns_join(*hsql_table_ref.join);
+          return _translate_named_columns_join(*hsql_table_ref.join);
         }
         return _translate_predicated_join(*hsql_table_ref.join);
       }
@@ -1054,7 +1054,7 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_predicated_join(const 
   return result_state;
 }
 
-SQLTranslator::TableSourceState SQLTranslator::_translate_namedColumns_join(const hsql::JoinDefinition& join) {
+SQLTranslator::TableSourceState SQLTranslator::_translate_named_columns_join(const hsql::JoinDefinition& join) {
   const auto join_mode = translate_join_mode(join.type);
 
   auto left_state = _translate_table_ref(*join.left);
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -1069,8 +1069,8 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_named_columns_join(con
   const auto& left_sql_identifier_resolver = left_state.sql_identifier_resolver;
   const auto& right_sql_identifier_resolver = right_state.sql_identifier_resolver;
 
-  Assert(left_sql_identifier_resolver, "Could not find valid sql_identifier resolver for left input.");
-  Assert(right_sql_identifier_resolver, "Could not find valid sql_identifier resolver for right input.");
+  Assert(left_sql_identifier_resolver, "Expected SQLIdentifierResolver for left input.");
+  Assert(right_sql_identifier_resolver, "Expected SQLIdentifierResolver for right input.");
 
   auto join_predicates = std::vector<std::shared_ptr<AbstractExpression>>{};
   join_predicates.reserve(join.namedColumns->size());
--- a/src/lib/sql/sql_translator.cpp
+++ b/src/lib/sql/sql_translator.cpp
@@ -1063,8 +1063,8 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_named_columns_join(con
   auto left_input_lqp = left_state.lqp;
   auto right_input_lqp = right_state.lqp;
 
-  Assert(!join.condition, "Specify either a condition or namedColums for JOIN not both.");
-  Assert(!join.namedColumns->empty(), "No named column to use for JOIN.");
+  Assert(!join.namedColumns->empty(), "Expected at least one named column.");
+  Assert(!join.condition, "Did not expect join condition for join using named columns.");
 
   const auto& left_sql_identifier_resolver = left_state.sql_identifier_resolver;
   const auto& right_sql_identifier_resolver = right_state.sql_identifier_resolver;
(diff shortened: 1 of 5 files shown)


dey4ss:
Just some quick notes before I'll review in detail: The updated parser version now also allows more expressions for insert and execute statements (hyrise/sql-parser#242). Have you also tried what happens if you issue such queries to Hyrise? If something breaks, I don't think we shoould change something here, but we should add an assertion if necessary.

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -1586,6 +1586,107 @@ TEST_F(SQLTranslatorTest, JoinInnerComplexPredicateA) {
   EXPECT_LQP_EQ(actual_lqp, expected_lqp);
 }
 
+TEST_F(SQLTranslatorTest, JoinInnerUsingNamedColumnsSimple) {
```
dey4ss, line 1589: Haven't looked into the tests in detail, yet. However, the formatting looks a bit interesting. I saw that you draw inspiration from other tests (e.g., `JoinNaturalSimple`), but IMO using that much newlines is too bloaty (feel free to change it there, as well). Regarding formatting: we also usually format hand-crafted LQPs by hand and turn clang-format off.

For example, see the tests for `CastStatement`:
https://github.com/hyrise/hyrise/blob/5306d23c855bd24bc112295f582064f16a7a5a8c/src/test/lib/sql/sql_translator_test.cpp#L3225-L3245

You can also see how we test invalid user inputs (e.g., named column not present in one or both inputs) there.

dey4ss (changes requested):
First pass. Basically, I have one suggestion that helps giving better error messages if the named columns are not present in one/both inputs.

src/lib/sql/sql_translator.cpp
```diff
@@ -1051,6 +1054,80 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_predicated_join(const
...
+  Assert(!join.namedColumns->empty(), "Expected at least one named column.");
+  Assert(!join.condition, "Did not expect join condition for join using named columns.");
+
+  const auto& left_sql_identifier_resolver = left_state.sql_identifier_resolver;
+  const auto& right_sql_identifier_resolver = right_state.sql_identifier_resolver;
+
+  Assert(left_sql_identifier_resolver, "Expected SQLIdentifierResolver for left input.");
+  Assert(right_sql_identifier_resolver, "Expected SQLIdentifierResolver for right input.");
+
+  auto join_predicates = std::vector<std::shared_ptr<AbstractExpression>>{join.namedColumns->size()};
```
dey4ss, line 1075: Note for me: We have to write that down in the coding guidelines.

You're absolutely fine, we prefer curly braces for initialization. *BUT* some STL structures (such as `std::vector`) differ between the semantics: `std::vector<int>(2)` creates a vector of two default-initialized ints (`0`s), `std::vector<int>{2}` creates a vector with a single `2`. Thus, we use regular braces when it comes to the number of items to make that clear. Sure, it's no difference for a vector of pointers, but things tend to slip through.

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -1569,6 +1551,26 @@ TEST_F(SQLTranslatorTest, JoinNaturalColumnAlias) {
   EXPECT_LQP_EQ(actual_lqp, expected_lqp);
 }
 
+TEST_F(SQLTranslatorTest, JoinNaturalDerivedTable) {
+  // Test that the Natural join can work with derived tables.
```
dey4ss, line 1555:
```suggestion
  // Test that the natural join can work with selected constants.
```

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -1569,6 +1551,26 @@ TEST_F(SQLTranslatorTest, JoinNaturalColumnAlias) {
...
+
+  const auto [actual_lqp, translation_info] =
+      sql_to_lqp_helper("SELECT * FROM int_float NATURAL JOIN (SELECT 123 as a) bar;");
+
+  const auto derived_table_node = AliasNode::make(expression_vector(123), std::vector<std::string>({"a"}),
+                                                  ProjectionNode::make(expression_vector(123), DummyTableNode::make()));
```
dey4ss, line 1561:
```suggestion
  // clang-format off
  const auto derived_table_node = 
  AliasNode::make(expression_vector(123), std::vector<std::string>({"a"}),
    ProjectionNode::make(expression_vector(123), 
      DummyTableNode::make()));
  // clang-format on
```

dey4ss (commented):
Nice work! Can you also add the example of `SELECT ... FROM ... JOIN ... USING (...)` to `resources/test_data/sqlite_testrunner_queries.sql`?
I noticed that we failed with such queries there because we duplicated the columns before. You can find the tables available in these tests in `resources/test_data/sqlite_testrunner.tables`.

src/lib/sql/sql_translator.hpp
```diff
@@ -193,7 +194,6 @@ class SQLTranslator final {
   std::shared_ptr<AbstractExpression> _translate_hsql_case(
       const hsql::Expr& expr, const std::shared_ptr<SQLIdentifierResolver>& sql_identifier_resolver);
 
- private:
```
dey4ss, line 196: Wh is it not private anymore?

Rob2U: This entire section already is private from l.82 on. Therefore I though that this private is redundant.

src/lib/sql/sql_translator.cpp
```diff
@@ -1051,6 +1054,80 @@ SQLTranslator::TableSourceState SQLTranslator::_translate_predicated_join(const
   return result_state;
 }
 
+SQLTranslator::TableSourceState SQLTranslator::_translate_named_columns_join(const hsql::JoinDefinition& join) {
+  const auto join_mode = translate_join_mode(join.type);
+
+  auto left_state = _translate_table_ref(*join.left);
+  const auto& right_state = _translate_table_ref(*join.right);
+
+  const auto left_input_lqp = left_state.lqp;
```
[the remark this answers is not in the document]
Rob2U: In l. 1101 I move from `left_state` but I need the `left_input_lqp` later for the LQP construction in l. 1126.

src/lib/sql/sql_translator.hpp
dey4ss: Makes sense, nice catch

dey4ss (approved):
Nice work! :)
Feel free to merge – make sure to condense the commit description to a short form when squashing.

PR #2679, opened 2025-01-15
Encode tables from SQL import

Rob2U:
Update the SQL-Parser to allow the following syntax:  
`COPY a_table FROM 'a_file.tbl' WITH (FORMAT BIN, ENCODING 'Unencoded');`
This way it is possible to specify the encoding of a table that is loaded via `COPY`.

How the encoding of a Table is determined:
Binary Files keep their encoding.
If no encoding is specified, the imported table uses the default encoding of `SegmentEncodingSpec` (Dictionary).
If a column does not support the specified encoding, it is encoded using `Dictionary` encoding.


src/lib/logical_query_plan/export_node.hpp
```diff
@@ -13,12 +15,14 @@ namespace hyrise {
  */
 class ExportNode : public EnableMakeForLQPNode<ExportNode>, public AbstractNonQueryNode {
  public:
-  ExportNode(const std::string& init_file_name, const FileType init_file_type);
+  ExportNode(const std::string& init_file_name, const FileType init_file_type,
+             const std::optional<EncodingType> init_file_encoding = std::nullopt);
```
dey4ss, line 19: I don't think we need to maintain the encoding for `ExportNode`s as we won't implement the feature in the near future. I't basically dead code with the Assertion in SqlTranslator.

src/lib/logical_query_plan/import_node.cpp
```diff
@@ -2,46 +2,60 @@
...
 
 ImportNode::ImportNode(const std::string& init_table_name, const std::string& init_file_name,
-                       const FileType init_file_type)
+                       const FileType init_file_type, const std::optional<EncodingType> init_table_encoding)
     : AbstractNonQueryNode(LQPNodeType::Import),
       table_name(init_table_name),
       file_name(init_file_name),
-      file_type(init_file_type) {}
+      file_type(init_file_type),
+      table_encoding(init_table_encoding) {}
```
dey4ss, line 26: Mini: I guess the optional parameter can be a const ref. Also, we tend to prefer curly braces for initialization.

src/lib/logical_query_plan/import_node.cpp
```diff
@@ -2,46 +2,60 @@
...
+  boost::algorithm::to_lower(file_type_str);
+
+  if (table_encoding) {
+    auto file_encoding_str = std::string{magic_enum::enum_name(*table_encoding)};
+    boost::algorithm::to_lower(file_encoding_str);
+    return "[Import] Name: '" + table_name + "' from '" + file_name + "' (" + file_type_str + ") using encoding (" +
+           file_encoding_str + ")";
+  }
+
+  return "[Import] Name: '" + table_name + "' from '" + file_name + "' (" + file_type_str + ")";
```
dey4ss, line 39: I'm not sure if we require all of these details (they will be printed in the visualization as well and can arbitrarily bloat them). I guess that could be a reason not to include the path that will just be create a mess:
image

I guess I would prefer a message such as:
```
[Import] 'foo' (file type Auto, encoding Auto) 
[Import] 'foo' (file type Tbl, encoding Unencoded) 
```
(also without forced lowercase). The descriptions won't be printed on a hot path, so I'd prefer a single return. If it makes things more easy to read, there's also nothing to say against a stream to collect the message.

src/lib/logical_query_plan/import_node.hpp
```diff
@@ -13,13 +15,15 @@ namespace hyrise {
  */
 class ImportNode : public EnableMakeForLQPNode<ImportNode>, public AbstractNonQueryNode {
  public:
-  ImportNode(const std::string& init_table_name, const std::string& init_file_name, const FileType init_file_type);
+  ImportNode(const std::string& init_table_name, const std::string& init_file_name, const FileType init_file_type,
+             const std::optional<EncodingType> init_table_encoding = std::nullopt);
```
dey4ss, line 19: The ctor is not used _that_ often in the codebase, so I'd vote against a default argument and for adapting these cases. As we often use `::make()` with specifying the input(s), we have to pass an argument anyways.

src/lib/operators/import.cpp
```diff
@@ -66,15 +70,21 @@ std::shared_ptr<const Table> Import::_on_execute() {
 
   Hyrise::get().storage_manager.add_table(_tablename, table);
 
-  // must match ImportNode::output_expressions
+  // If a table encoding is specified, encode the table accordingly.
+  if (_table_encoding) {
+    const auto chunk_encoding_spec =
+        ChunkEncodingSpec(table->column_count(), SegmentEncodingSpec(_table_encoding.value()));
```
dey4ss, line 76: curly braces :)

src/lib/operators/import.cpp
```diff
@@ -66,15 +70,21 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
+    ChunkEncoder::encode_all_chunks(Hyrise::get().storage_manager.get_table(_tablename), chunk_encoding_spec);
+  }
```
dey4ss, line 78: @Bouncner what's your take on "default encoding"? As of now, `.tbl` and `.csv` files are stored `Unencoded`, `.bin`s in the encoding of the input.

While I would find it counterintuitive to re-encode an already encoded binary of no encoding is explicitly requested in the SQL statement (and the binary's encoding might be a lot more fine-grained), I think a default (e.g., `Dictionary`) for `.csv` and `.tbl` makes sense.

src/lib/operators/import.cpp
```diff
@@ -66,15 +70,21 @@ std::shared_ptr<const Table> Import::_on_execute() {
 
   Hyrise::get().storage_manager.add_table(_tablename, table);
 
-  // must match ImportNode::output_expressions
+  // If a table encoding is specified, encode the table accordingly.
+  if (_table_encoding) {
+    const auto chunk_encoding_spec =
+        ChunkEncodingSpec(table->column_count(), SegmentEncodingSpec(_table_encoding.value()));
+    ChunkEncoder::encode_all_chunks(Hyrise::get().storage_manager.get_table(_tablename), chunk_encoding_spec);
```
dey4ss, line 77: Is there a reason to obtain the table ptr again? It's already in the `table` variable.

src/lib/sql/sql_translator.cpp
```diff
@@ -1834,15 +1835,32 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_import(const hsql::Im
...
+    return ImportNode::make(import_statement.tableName, import_statement.filePath,
+                            import_type_to_file_type(import_statement.type));
+  }
+
+  auto file_encoding =
+      magic_enum::enum_cast<EncodingType>(std::string{import_statement.encoding}, magic_enum::case_insensitive);
+  AssertInput(file_encoding.has_value(), "Unknown encoding type '" + std::string{import_statement.encoding} + "'.");
   return ImportNode::make(import_statement.tableName, import_statement.filePath,
-                          import_type_to_file_type(import_statement.type));
+                          import_type_to_file_type(import_statement.type), file_encoding.value());
```
dey4ss, line 1848: sth like:
```c++
auto encoding = std::optional<EncodingType>{};
if (import_statement.encoding) {
  ...
}

return ImportNode::make(..., encoding);
```

dey4ss: `AssertInput(encoding, ...)`

src/lib/sql/sql_translator.cpp
```diff
@@ -1834,15 +1835,32 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_import(const hsql::Im
...
   auto lqp = std::shared_ptr<AbstractLQPNode>{};
 
+  if (export_statement.encoding) {
+    AssertInput(false, "Encoding '" + std::string{export_statement.encoding} +
+                           "' was specified"
+                           " but a set output encoding is not yet supported."
+                           "To achieve the same result, export the table once, "
+                           "then load using the wished encoding and export again.");
+  }
+
```
dey4ss, line 1863:
```c++
if (export_statement.encoding) {
  FailInput(...);
}
```
or just
```
AssertInput(!export_statement.encoding, ...);
```

I would also just be more brief here, sth like: "Encoding for table export is not supported. You can choose an encoding when importing tables."

dey4ss (commented):
Will have a look at the tests later

src/lib/operators/import.cpp
Bouncner: I was not aware we do not encudre CSV files to be honest. I would definitely favor our default encoding (dict) for imported files.

src/lib/operators/import.cpp
dey4ss: `SegmentEncodingSpec{*_table_encoding}`

src/lib/operators/import.cpp
```diff
@@ -66,15 +70,25 @@ std::shared_ptr<const Table> Import::_on_execute() {
 
   Hyrise::get().storage_manager.add_table(_tablename, table);
 
-  // must match ImportNode::output_expressions
+  // If a table encoding is specified, encode the table accordingly.
+  // The default encoding is `EncodingType::Dictionary` except for binary files.
+  // For binary files the default is the encoding of the file.
```
dey4ss, line 75: commas, and fill 120 chars.

src/lib/operators/import.cpp
```diff
@@ -66,15 +70,25 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
+  auto encoding_to_apply =
+      (_file_type != FileType::Binary && !_table_encoding) ? EncodingType::Dictionary : _table_encoding;
+
+  if (encoding_to_apply) {
+    const auto chunk_encoding_spec = ChunkEncodingSpec(table->column_count(), SegmentEncodingSpec(*encoding_to_apply));
+    ChunkEncoder::encode_all_chunks(table, chunk_encoding_spec);
+  }
```
dey4ss, line 82: Can't we just do the following (`_target_encoding` can even remain `const`)?
```suggestion
  if (_target_encoding || _file_type != FileType::Binary) {
    const auto encoding_spec = _target_encoding ? SegmentEncodingSpec{*_target_encoding} : {};
    ChunkEncoder::encode_all_chunks(table, encoding_spec);
  }
```

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+
+  // Apply the encoding to the table.
+  if (encoding) {
+    const auto encoding_spec = SegmentEncodingSpec{*encoding};
+    ChunkEncoder::encode_all_chunks(expected_table, encoding_spec);
+  } else if (file_type != FileType::Binary) {
+    // If no encoding is specified we want to check if the defaults are as expected.
+    // The binary file is unencoded so we do not have to change the encoding of `expected_table`.
+    // For `.csv` and `.tbl` the default encoding should be `EncodingType::Dictionary`.
+    const auto encoding_spec = SegmentEncodingSpec{EncodingType::Dictionary};
```
dey4ss, line 76: `SegmentEncodingSpec{}` even uses `EncosingType::Dictionary` as the default:
https://github.com/hyrise/hyrise/blob/1d7f8607ab004922e63d6ab55b4f8986116c6e5c/src/lib/storage/encoding_type.hpp#L61

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+    ChunkEncoder::encode_all_chunks(expected_table, encoding_spec);
+  }
+
+  const auto reference_filename =
+      reference_filepath + reference_filenames.at(file_type) + file_extensions.at(file_type);
+  auto importer = std::make_shared<Import>(reference_filename, "a", Chunk::DEFAULT_SIZE, file_type, encoding);
```
dey4ss, line 82: const

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
   importer->execute();
+  auto table = Hyrise::get().storage_manager.get_table("a");
```
dey4ss, line 84: const

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
 
-  EXPECT_TABLE_EQ_ORDERED(Hyrise::get().storage_manager.get_table("a"), expected_table);
+  EXPECT_TABLE_EQ_ORDERED(table, expected_table);
+
+  auto segment = table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+  auto expected_segment = expected_table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
```
dey4ss, line 89: const

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+
+  // Check if the imported segment has the same type as the manually encoded one.
+  resolve_segment_type<float>(*segment, [&](const auto& segment) {
+    EXPECT_TRUE(std::dynamic_pointer_cast<std::remove_reference_t<decltype(segment)>>(expected_segment));
```
dey4ss, line 93: Guess the following would be a bit nicer to read:
```suggestion
    using EncodedSegmentType = std::remove_reference_t<decltype(segment)>;
    EXPECT_TRUE(std::dynamic_pointer_cast<EncodedSegmentType>(expected_segment));
```

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+class OperatorsImportFileTypesTest : public OperatorsImportTest, public ::testing::WithParamInterface<FileType> {};
+
+std::string file_types_name_generator(const ::testing::TestParamInfo<FileType>& info) {
+  const auto& file_type = info.param;
+
+  return std::string{magic_enum::enum_name(file_type)};
+}
+
+INSTANTIATE_TEST_SUITE_P(FileTypes, OperatorsImportFileTypesTest,
+                         ::testing::Values(FileType::Csv, FileType::Tbl, FileType::Binary), file_types_name_generator);
```
dey4ss, line 106: have a look at `enum_formatter`:
https://github.com/hyrise/hyrise/blob/1d7f8607ab004922e63d6ab55b4f8986116c6e5c/src/test/base_test.hpp#L111-L114
```suggestion
INSTANTIATE_TEST_SUITE_P(FileTypes, OperatorsImportFileTypesTest,
                         ::testing::Values(FileType::Csv, FileType::Tbl, FileType::Binary), enum_formatter<FileType>);
```

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
       std::make_shared<Table>(TableColumnDefinitions{{"a", DataType::Float, false}}, TableType::Data, ChunkOffset{5});
   expected_table->append({1.1f});
   expected_table->append({2.2f});
   expected_table->append({3.3f});
   expected_table->append({4.4f});
 
+  // Make all chunks in the expected table immutable to allow encoding.
+  for (auto chunk_id = ChunkID{0}; chunk_id < expected_table->chunk_count(); ++chunk_id) {
+    expected_table->get_chunk(chunk_id)->set_immutable();
+  }
```
dey4ss, line 119: Just one chunk?

dey4ss: Also, `Table` sets the former last chunk immutable when appending a new one, so you always just need to call `set_immutable()` on the `table->last_chunk()`.

https://github.com/hyrise/hyrise/blob/1d7f8607ab004922e63d6ab55b4f8986116c6e5c/src/lib/storage/table.cpp#L203-L216

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+std::string file_types_name_generator(const ::testing::TestParamInfo<FileType>& info) {
+  const auto& file_type = info.param;
+
+  return std::string{magic_enum::enum_name(file_type)};
+}
+
+INSTANTIATE_TEST_SUITE_P(FileTypes, OperatorsImportFileTypesTest,
+                         ::testing::Values(FileType::Csv, FileType::Tbl, FileType::Binary), file_types_name_generator);
+
+TEST_P(OperatorsImportFileTypesTest, ImportWithoutFileTypeAndEncoding) {
```
dey4ss, line 108: The file type is given via the param, right?

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+  // For the `.csv` and `.tbl` files we expect a default encoding of `Dictionary`.
+  // The `.bin` file we load is stored and therefore expected to be `Unencoded`.
+  const auto encoding = (file_type == FileType::Binary) ? std::optional<EncodingType>{EncodingType::Unencoded}
+                                                        : std::optional<EncodingType>{EncodingType::Dictionary};
+
+  // Apply the encoding to the table.
+  if (encoding) {
+    const auto encoding_spec = SegmentEncodingSpec{*encoding};
+    ChunkEncoder::encode_all_chunks(expected_table, encoding_spec);
+  }
```
dey4ss, line 131: You always set the enoding – why is it an optional. Plus, the table is unencoded already. So we can simplify to:
```suggestion
  if (file_type != FileType::Binary) {
    ChunkEncoder::encode_all_chunks(expected_table);
  }
```

src/test/lib/operators/import_test.cpp
dey4ss: And just from a semantic point, I would switch `segment` and `expected_segment`.

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
   const auto reference_filename =
-      reference_filepath + reference_filenames.at(GetParam()) + file_extensions.at(GetParam());
-  auto importer = std::make_shared<Import>(reference_filename, "a");
+      reference_filepath + reference_filenames.at(file_type) + file_extensions.at(file_type);
+
+  // Instead of using the provided `file_type` for the Importer we use the default `FileType::Auto` here.
+  auto importer = std::make_shared<Import>(reference_filename, "a", Chunk::DEFAULT_SIZE, FileType::Auto,
+                                           std::optional<EncodingType>{});
   importer->execute();
+  auto table = Hyrise::get().storage_manager.get_table("a");
```
dey4ss, line 140: const

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
 
-  EXPECT_TABLE_EQ_ORDERED(Hyrise::get().storage_manager.get_table("a"), expected_table);
+  EXPECT_TABLE_EQ_ORDERED(table, expected_table);
+
+  auto segment = table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+  auto expected_segment = expected_table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
```
dey4ss, line 145: const

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+
+  // Check if the imported segment has the same type as the manually encoded one.
+  resolve_segment_type<float>(*segment, [&](const auto& segment) {
+    EXPECT_TRUE(std::dynamic_pointer_cast<std::remove_reference_t<decltype(segment)>>(expected_segment));
+  });
```
dey4ss, line 150: (see comment above)

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+  const auto table = Hyrise::get().storage_manager.get_table("a");
 
   EXPECT_EQ(table->uses_mvcc(), UseMvcc::Yes);
   EXPECT_TRUE(table->get_chunk(ChunkID{0})->has_mvcc_data());
   EXPECT_EQ(table->get_chunk(ChunkID{0})->mvcc_data()->max_begin_cid.load(), CommitID{0});
 }
 
+TEST_F(OperatorsImportTest, EncodeImportWithFrameOfReference) {
+  const auto& encoding_type = EncodingType::FrameOfReference;
+  auto expected_table =
```
dey4ss, line 170: const

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+      std::make_shared<Table>(TableColumnDefinitions{{"a", DataType::Int, false}}, TableType::Data, ChunkOffset{4});
+
+  expected_table->append({123});
+  expected_table->append({1234});
+  expected_table->append({12345});
+
+  // Make all chunks in the expected table immutable to allow encoding.
+  for (auto chunk_id = ChunkID{0}; chunk_id < expected_table->chunk_count(); ++chunk_id) {
+    expected_table->get_chunk(chunk_id)->set_immutable();
+  }
```
dey4ss, line 180: (see comment above)

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
-  auto table = Hyrise::get().storage_manager.get_table("a");
+  const auto table = Hyrise::get().storage_manager.get_table("a");
 
   EXPECT_EQ(table->uses_mvcc(), UseMvcc::Yes);
   EXPECT_TRUE(table->get_chunk(ChunkID{0})->has_mvcc_data());
   EXPECT_EQ(table->get_chunk(ChunkID{0})->mvcc_data()->max_begin_cid.load(), CommitID{0});
 }
 
+TEST_F(OperatorsImportTest, EncodeImportWithFrameOfReference) {
+  const auto& encoding_type = EncodingType::FrameOfReference;
```
dey4ss, line 169: no ref. Do we even need the variable?

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+
+  // Make all chunks in the expected table immutable to allow encoding.
+  for (auto chunk_id = ChunkID{0}; chunk_id < expected_table->chunk_count(); ++chunk_id) {
+    expected_table->get_chunk(chunk_id)->set_immutable();
+  }
+  // Encode the expected table with FrameOfReference.
+  const auto encoding_spec = SegmentEncodingSpec{encoding_type};
+  ChunkEncoder::encode_all_chunks(expected_table, encoding_spec);
+
+  const auto filename = std::string{"resources/test_data/tbl/int.tbl"};
```
dey4ss, line 185: const

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+  auto importer = std::make_shared<Import>(filename, "a", Chunk::DEFAULT_SIZE, FileType::Tbl, encoding_type);
+  importer->execute();
+
+  const auto table = Hyrise::get().storage_manager.get_table("a");
+
+  EXPECT_TABLE_EQ_ORDERED(table, expected_table);
+
+  auto segment = table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+  auto expected_segment = expected_table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
```
dey4ss, line 194: const

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+
+  // Check if the imported segment has the same type as the manually encoded one.
+  resolve_segment_type<float>(*segment, [&](const auto& segment) {
+    EXPECT_TRUE(std::dynamic_pointer_cast<std::remove_reference_t<decltype(segment)>>(expected_segment));
+  });
```
dey4ss, line 199: you know you want a `FrameOfReferenceSegment<int32_t>`:
```suggestion
  EXPECT_TRUE(std::dynamic_pointer_cast<const FrameOfReferenceSegment<int32_t>>(expected_segment));
```

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+}
+
+TEST_F(OperatorsImportTest, EncodeImportWithFixedStringDictionary) {
+  const auto encoding_type = EncodingType::FixedStringDictionary;
```
dey4ss, line 203: needed?

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+  auto expected_table =
```
dey4ss, line 204: const

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+  expected_table->append({"www"});
+  expected_table->append({"yyy"});
+  expected_table->append({"uuu"});
+  expected_table->append({"ttt"});
+  expected_table->append({"zzz"});
+
+  // Make all chunks in the expected table immutable to allow encoding.
+  for (auto chunk_id = ChunkID{0}; chunk_id < expected_table->chunk_count(); ++chunk_id) {
+    expected_table->get_chunk(chunk_id)->set_immutable();
+  }
```
dey4ss, line 217: (see above)

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+  // Encode the expected table with FrameOfReference.
+  const auto encoding_spec = SegmentEncodingSpec{encoding_type};
+  ChunkEncoder::encode_all_chunks(expected_table, encoding_spec);
+
+  const auto filename = std::string{"resources/test_data/tbl/string.tbl"};
+  auto importer = std::make_shared<Import>(filename, "a", Chunk::DEFAULT_SIZE, FileType::Tbl, encoding_type);
```
dey4ss, line 223: const

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +27,215 @@ class OperatorsImportTest : public BaseTest {
...
+
+  EXPECT_TABLE_EQ_ORDERED(table, expected_table);
+
+  auto segment = table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+  auto expected_segment = expected_table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+
+  // Check if the imported segment has the same type as the manually encoded one.
+  resolve_segment_type<std::string>(*segment, [&](const auto& segment) {
+    EXPECT_TRUE(std::dynamic_pointer_cast<std::remove_reference_t<decltype(segment)>>(expected_segment));
+  });
```
dey4ss, line 236: (see above)

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -3183,33 +3187,83 @@ TEST_F(SQLTranslatorTest, ComplexSetOperationQuery) {
 TEST_F(SQLTranslatorTest, CopyStatementImport) {
   {
     const auto [actual_lqp, translation_info] = sql_to_lqp_helper("COPY a_table FROM 'a_file.tbl';");
-    const auto expected_lqp = ImportNode::make("a_table", "a_file.tbl", FileType::Auto);
+    const auto expected_lqp = ImportNode::make("a_table", "a_file.tbl", FileType::Auto, std::optional<EncodingType>{});
```
dey4ss, line 3190: pick here: `std::nullopt`?

dey4ss (changes requested):
I have some comments, but mostly some simplification, style, etc. Overall, I really like it! :)

Bouncner:
Short question: can we now import into existing tables or not?
Not something I want to inject here, just curious as we will later need it for the benchmarks.

src/test/lib/operators/import_test.cpp
dey4ss: Got it. Maybe `AutoFileType` makes that a bit more clear.

Rob2U: I changed it to `ImportWithAutoFileType`

dey4ss:
Can you please add a short PR description?

src/lib/sql/sql_translator.cpp
```diff
@@ -1834,15 +1835,25 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_import(const hsql::Im
...
-                          import_type_to_file_type(import_statement.type));
+                          import_type_to_file_type(import_statement.type), encoding);
 }
 
 // NOLINTNEXTLINE: while this particular method could be made static, others cannot.
 std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_export(const hsql::ExportStatement& export_statement) {
   auto sql_identifier_resolver = std::make_shared<SQLIdentifierResolver>();
   auto lqp = std::shared_ptr<AbstractLQPNode>{};
 
+  AssertInput(!export_statement.encoding,
```
Bouncner, line 1854: I don't understand how the second sentence is relevant here?

src/test/lib/operators/import_test.cpp
```diff
@@ -18,55 +31,188 @@ class OperatorsImportTest : public BaseTest {
...
+  const auto table = Hyrise::get().storage_manager.get_table("a");
+
+  EXPECT_TABLE_EQ_ORDERED(table, expected_table);
+
+  const auto segment = table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+  // Check if the imported segment has the expected type (FrameOfReferenceSegment<int32_t>).
+  EXPECT_TRUE(std::dynamic_pointer_cast<FrameOfReferenceSegment<int32_t>>(segment));
+}
+
+TEST_F(OperatorsImportTest, EncodeImportWithFixedStringDictionary) {
```
Bouncner, line 186: Just skipping over the tests ...

What would happen if we chose frame of reference in the import statement but there are integer columns as well as string columns?
A warning? A fail? Is that case tested?

dey4ss: It should throw here:
https://github.com/hyrise/hyrise/blob/543beaadd396411286fdd5c0535adf59e8f06d2f/src/lib/storage/base_segment_encoder.hpp#L78-L92

Good catch anyways, we should add a test.

src/lib/sql/sql_translator.cpp
dey4ss: I mean, why should you want to export a table with a specific encoding? Only for the case that this encoding is applied when you import table again somewhere. And for this, you can specify the encoding directly during the import.

src/test/lib/operators/import_test.cpp
Bouncner: I cannot look it up right now, but the way we usually handle it is using Unencoded whenever the desired encoding cannot be used.
We should check what really happens.

I am not voting to implement that. Maybe we can use the methods that handle encoding in the benchmark runner?

dey4ss: I briefly looked into `BenchmarkTableEncoder`. It leaves tables unencoded, that's true. No strong opinion here.

If we use `BenchmarkTableEncoder` though, we could as well also pass a more detailed encoding spec as argument. Not sure if we want do this here, your call.

Bouncner: I think we should keep the command simple as it is. But calling "import with encoding frameofreference" should not lead to a crash.
Maybe default to dict whenever a column type is not supported by the requested encoding schema?

dey4ss: @Bouncner Sounds fair. Everything required for this is already part of `BenchmarkTableEncoder::encode()`: 
https://github.com/hyrise/hyrise/blob/543beaadd396411286fdd5c0535adf59e8f06d2f/src/benchmarklib/benchmark_table_encoder.cpp#L65

Only the default encoding is different there. However, we could create an `EncodingConfig` that cares for that:
https://github.com/hyrise/hyrise/blob/543beaadd396411286fdd5c0535adf59e8f06d2f/src/benchmarklib/encoding_config.hpp#L24-L25

If I get that right, setting the user-defined encoding as `default_encoding_spec` and dictionary encoding for data types that are not supported by that encoding in `type_encoding_mapping` should behave as you suggested?

@Rob2U you can find which encodings support which data types here: 
https://github.com/hyrise/hyrise/blob/543beaadd396411286fdd5c0535adf59e8f06d2f/src/lib/storage/encoding_type.hpp#L38-L43

src/lib/operators/import.cpp
```diff
@@ -73,8 +73,22 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
   if (_target_encoding || _file_type != FileType::Binary) {
-    const auto encoding_spec = _target_encoding ? SegmentEncodingSpec{*_target_encoding} : SegmentEncodingSpec{};
-    ChunkEncoder::encode_all_chunks(table, encoding_spec);
+    auto chunk_encoding_spec = ChunkEncodingSpec{};
+
+    for (auto column_id = ColumnID{0}; column_id < table->column_count(); ++column_id) {
+      // If a target encoding is specified, use it. Otherwise, use the default encoding: Dictionary.
+      const auto resolved_encoding = _target_encoding.value_or(EncodingType::Dictionary);
+
+      const auto& column_data_type = table->column_data_type(column_id);
```
dey4ss, line 82:
```suggestion
      const auto column_data_type = table->column_data_type(column_id);
```

src/test/lib/operators/import_test.cpp
```diff
@@ -50,16 +50,26 @@ INSTANTIATE_TEST_SUITE_P(FileTypesAndEncodings, OperatorsImportMultiFileTypeAndE
...
 TEST_P(OperatorsImportMultiFileTypeAndEncodingTest, ImportWithEncodingAndFileType) {
+  /*
+    int_string2.tbl (also for .bin and .csv)
+    a|b
+    int|string
+    123|A
+    1234|B
+    12345|C
+  */
+
```
dey4ss, line 66:
```suggestion
```

src/test/lib/operators/import_test.cpp
```diff
@@ -85,13 +100,24 @@ TEST_P(OperatorsImportMultiFileTypeAndEncodingTest, ImportWithEncodingAndFileTyp
...
 
-  const auto segment = table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
-  const auto expected_segment = expected_table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+  // Check Segment 1 / Column 1
+  const auto segment_1 = table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+  const auto expected_segment_1 = expected_table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
 
   // Check if the imported segment has the same type as the manually encoded one.
-  resolve_segment_type<float>(*expected_segment, [&](const auto& expected_segment) {
+  resolve_segment_type<int>(*expected_segment_1, [&](const auto& expected_segment) {
```
dey4ss, line 108:
```suggestion
  resolve_segment_type<int32_t>(*expected_segment_1, [&](const auto& expected_segment) {
```

src/test/lib/operators/import_test.cpp
```diff
@@ -85,13 +100,24 @@ TEST_P(OperatorsImportMultiFileTypeAndEncodingTest, ImportWithEncodingAndFileTyp
...
     using EncodedSegmentType = std::remove_reference_t<decltype(expected_segment)>;
-    EXPECT_TRUE(std::dynamic_pointer_cast<EncodedSegmentType>(segment));
+    EXPECT_TRUE(std::dynamic_pointer_cast<EncodedSegmentType>(segment_1));
+  });
+
+  // Check Segment 2 / Column 2
```
dey4ss, line 113:
```suggestion
  // Check Segment 2 / Column 2.
```

src/test/lib/operators/import_test.cpp
```diff
@@ -85,13 +100,24 @@ TEST_P(OperatorsImportMultiFileTypeAndEncodingTest, ImportWithEncodingAndFileTyp
 
   EXPECT_TABLE_EQ_ORDERED(table, expected_table);
 
-  const auto segment = table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
-  const auto expected_segment = expected_table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+  // Check Segment 1 / Column 1
```
dey4ss, line 103:
```suggestion
  // Check Segment 1 / Column 1.
```

src/test/lib/operators/import_test.cpp
```diff
@@ -131,13 +157,24 @@ TEST_P(OperatorsImportFileTypesTest, ImportWithAutoFileType) {
...
 
-  const auto segment = table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
-  const auto expected_segment = expected_table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+  // Check Segment 1.
+  const auto segment_1 = table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
+  const auto expected_segment_1 = expected_table->get_chunk(ChunkID{0})->get_segment(ColumnID{0});
 
   // Check if the imported segment has the same type as the manually encoded one.
-  resolve_segment_type<float>(*expected_segment, [&](const auto& expected_segment) {
+  resolve_segment_type<int>(*expected_segment_1, [&](const auto& expected_segment) {
```
dey4ss, line 165:
```suggestion
  resolve_segment_type<int32_t>(*expected_segment_1, [&](const auto& expected_segment) {
```

dey4ss (approved):
Just some minor suggestions. Feel free to merge once you applied them. Nice work!

src/test/lib/operators/import_test.cpp
Rob2U: @Bouncner Thanks for the good catch. I implemented the behavior that you proposed (if a column does not support the specified encoding, it is encoded using Dictionary encoding). For the implementation I used `ChunkEncoder::encode_all_chunks` similar to `BenchmarkTableEncoder::encode` (@dey4ss thanks for the references)

PR #2686, opened 2025-02-12
Remove llvm-symbolizer from Dockerfile

Bouncner:
Removing setting a soft link for the llvm symbolizer (added with #953) in the Dockerfile.


dey4ss:
> Removing setting a soft link for the llvm symbolizer (added with #953) in the Dockerfile.

The symbolizer was just bumped to 14 in #953, it's actually there since 1db8f29.

PR #2687, opened 2025-02-13
Fix `install_dependencies.sh`

tjeyy:
The `install_dependencies.sh` runs apt in the background, so the python requirements can be installed simultaneously.
This fails if `pip3` is called before it is installed via `apt-get`:
```
Processing triggers for dbus (1.14.10-4ubuntu4.1) ...
+ sudo apt-get install --no-install-recommends -y autoconf bash-completion bc clang-16 clang-17 clang-format-17 clang-tidy-17 cmake curl dos2unix g++-14 gcc-14 git graphviz libboost-all-dev libhwloc-dev libncurses5-dev libnuma-dev libnuma1 libpq-dev libreadline-dev libsqlite3-dev libtbb-dev lld-17 man parallel postgresql-server-dev-all python3 python3-pip valgrind
+ git submodule update --jobs 5 --init --recursive
+ pip3 install --break-system-packages -r requirements.txt
./install_dependencies.sh: line 64: pip3: command not found
+ echo 'Error during installation of python requirements.'
Error during installation of python requirements.
+ exit 1
Reading package lists... Done
```

### DIFF
--- a/install_dependencies.sh
+++ b/install_dependencies.sh
@@ -49,8 +49,8 @@ if echo $REPLY | grep -E '^[Yy]$' > /dev/null; then
         if [ -f /etc/lsb-release ] && cat /etc/lsb-release | grep DISTRIB_ID | grep Ubuntu >/dev/null; then
             echo "Installing dependencies (this may take a while)..."
             if sudo apt-get update >/dev/null; then
+            
                 # Packages added here should also be added to the Dockerfile
-
                 if ! sudo apt-get install --no-install-recommends -y software-properties-common lsb-release git python3 python3-pip autoconf bash-completion bc clang-16 clang-17 clang-format-17 clang-tidy-17 cmake curl dos2unix g++-14 gcc-14 graphviz libboost-all-dev libhwloc-dev libncurses5-dev libnuma-dev libnuma1 libpq-dev libreadline-dev libsqlite3-dev libtbb-dev lld-17 man parallel postgresql-server-dev-all valgrind; then
                     echo "Error during apt-get installations."
                     exit 1
--- a/install_dependencies.sh
+++ b/install_dependencies.sh
@@ -56,7 +56,7 @@ if echo $REPLY | grep -E '^[Yy]$' > /dev/null; then
 
                 # Packages added here should also be added to the Dockerfile
                 sudo apt-get install --no-install-recommends -y python3 python3-pip
-                sudo apt-get install --no-install-recommends -y autoconf bash-completion bc clang-16 clang-17 clang-format-17 clang-tidy-17 cmake curl dos2unix g++-14 gcc-14 git graphviz libboost-all-dev libhwloc-dev libncurses5-dev libnuma-dev libnuma1 libpq-dev libreadline-dev libsqlite3-dev libtbb-dev lld-17 man parallel postgresql-server-dev-all valgrind
+                sudo apt-get install --no-install-recommends -y autoconf bash-completion bc clang-16 clang-17 clang-format-17 clang-tidy-17 cmake curl dos2unix g++-14 gcc-14 git graphviz libboost-all-dev libhwloc-dev libncurses5-dev libnuma-dev libnuma1 libpq-dev libreadline-dev libsqlite3-dev libtbb-dev lld-17 man parallel postgresql-server-dev-all valgrind &
 
                 if ! git submodule update --jobs 5 --init --recursive; then
                     echo "Error during git fetching submodules."
--- a/install_dependencies.sh
+++ b/install_dependencies.sh
@@ -52,7 +52,7 @@ if echo $REPLY | grep -E '^[Yy]$' > /dev/null; then
                 sudo apt-get install --no-install-recommends -y software-properties-common lsb-release
 
                 # Packages added here should also be added to the Dockerfile
-                sudo apt-get install --no-install-recommends -y autoconf bash-completion bc clang-16 clang-17 clang-format-17 clang-tidy-17 cmake curl dos2unix g++-14 gcc-14 git graphviz libboost-all-dev libhwloc-dev libncurses5-dev libnuma-dev libnuma1 libpq-dev libreadline-dev libsqlite3-dev libtbb-dev lld-17 man parallel postgresql-server-dev-all python3 python3-pip valgrind &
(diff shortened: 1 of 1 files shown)


robert-richter-praxipal:
image

tjeyy:
What is your opinion on adding `set -x` to make the script a bit more verbose?

Bouncner:
As apt and other steps take quite some time, I think the parallelization is good.

What about installing `git` and `pip3` first (non-background) and then all the others in the background?

Bouncner:
> What is your opinion on adding `set -x` to make the script a bit more verbose?

Can you show an examplary output?

tjeyy:
I like your idea!
What is the point of installing git via the script?
Are there thinkable scenarios where Hyrise it not deployed via git?

Bouncner:
> I like your idea! What is the point of installing git via the script? Are there thinkable scenarios where Hyrise it not deployed via git?

No particular reason for it. It is just the next command that follows and several stripped down image (I think alpine) do not ship git. So I think it is reasonable to include it explicitly.

install_dependencies.sh
```diff
@@ -52,7 +55,8 @@ if echo $REPLY | grep -E '^[Yy]$' > /dev/null; then
                 sudo apt-get install --no-install-recommends -y software-properties-common lsb-release
 
                 # Packages added here should also be added to the Dockerfile
-                sudo apt-get install --no-install-recommends -y autoconf bash-completion bc clang-16 clang-17 clang-format-17 clang-tidy-17 cmake curl dos2unix g++-14 gcc-14 git graphviz libboost-all-dev libhwloc-dev libncurses5-dev libnuma-dev libnuma1 libpq-dev libreadline-dev libsqlite3-dev libtbb-dev lld-17 man parallel postgresql-server-dev-all python3 python3-pip valgrind &
+                sudo apt-get install --no-install-recommends -y python3 python3-pip
+                sudo apt-get install --no-install-recommends -y autoconf bash-completion bc clang-16 clang-17 clang-format-17 clang-tidy-17 cmake curl dos2unix g++-14 gcc-14 git graphviz libboost-all-dev libhwloc-dev libncurses5-dev libnuma-dev libnuma1 libpq-dev libreadline-dev libsqlite3-dev libtbb-dev lld-17 man parallel postgresql-server-dev-all valgrind
```
Bouncner, line 59: Re-add `&`

Bouncner:
> Example output for set -x
> ```
> root@nx05:~/hyrise# ./install_dependencies.sh
> + [[ -z '' ]]
> + read -p 'This script installs the dependencies of Hyrise. It might upgrade already installed packages. Continue? [y|n] ' -n 1 -r
> This script installs the dependencies of Hyrise. It might upgrade already installed packages. Continue? [y|n] y+ echo
> 
> + echo y
> + grep -E '^[Yy]$'
> ++ uname
> + unamestr=Linux
> ...
> ```

I wouldn't oppose it but I don't see how it really helps. At least for me, getting the last failing command is sufficient.

install_dependencies.sh
```diff
@@ -52,7 +55,8 @@ if echo $REPLY | grep -E '^[Yy]$' > /dev/null; then
                 sudo apt-get install --no-install-recommends -y software-properties-common lsb-release
 
                 # Packages added here should also be added to the Dockerfile
-                sudo apt-get install --no-install-recommends -y autoconf bash-completion bc clang-16 clang-17 clang-format-17 clang-tidy-17 cmake curl dos2unix g++-14 gcc-14 git graphviz libboost-all-dev libhwloc-dev libncurses5-dev libnuma-dev libnuma1 libpq-dev libreadline-dev libsqlite3-dev libtbb-dev lld-17 man parallel postgresql-server-dev-all python3 python3-pip valgrind &
+                sudo apt-get install --no-install-recommends -y git python3 python3-pip
```
Bouncner, line 58: Please explain in a comment why we add certain tools before others. Also merge with line 55?

tjeyy:
> I wouldn't oppose it but I don't see how it really helps. At least for me, getting the last failing command is sufficient.
Without `set -x`, you get an output, but you cannot see which command caused which output.
On the other hand, the association is no longer possible anyway due to the background process …
What is @dey4ss's stand on this?

A different topic: Currently, the script just goes on if `apt-get` fails. On the other hand, if something goes wrong with `pip`, the script aborts.
I would suggest harmonizing this.
We could add `set -e` to the script or put the `apt-get` call into an if statement, just like `pip`.
What do you think?

dey4ss:
Hm, I guess it makes sense to think of the scenarios the script is called
1. CI
2. Users

For 1.: If the script repeatedly fails in the CI, we need to lokk into it anyways. A bloaty output there (`-x`) does not hurt, but as we will have to look into it in that scenario anyways (re-run pipeline etc), we could as well just enable verbose putput on demand.

For 2.: I don't know if anyone would really run this script. If so, it must be inexperienced users and walll of output that is printed by default does not help either. On the other hand, it might help finding issues IF something goes wrong.

I guess it makes sense to fail if installs do not work to trigger a CI crash as early is possible. No strong opinion if `set -e` (only error is printed) of check on return value + dedicated error message is better.

Could we maybe add a `--verbose`/`-v` flag? By default, nothing is printed, with flag `set -x` is used? So we have a trade-off.

tjeyy:
I'm not sure if I like a `--verbose` flag, since setting it would still require changes to calling scripts.
In this case, `set -x` could just as well be set manually.

However, I would like to argue strongly against the background execution of apt-get.
If something goes wrong, it would be good to know exactly what went wrong and that everything that followed in the script was not carried out.
Executing the script in the background obscures the source of the error and makes it impossible to cancel in time.

Bouncner:
I am having no strong opinions here. Probably my only advice is: keep it simple. It is just a setup script.

Our docker image is quite recent, so there will not be much installed via apt. Doing it not in the background will be not issue. I wonder how it looks when we have a more outdated image. But then again: maybe we should rather keep our image up-to-date from time to time.

So I can life with non-background apt. And please: keep it simple.

tjeyy:
@dey4ss and @Bouncner what is your opinion on the current approach?
If you prefer parallel dependency installation, let's bring it back.
It made me some headaches debugging failing script executions, but the parallelism could, of course, be removed manually in this case.

Bouncner:
I think it more safe, cleaner and the added runtime does not bother me.

Fine with me. @dey4ss ?

Bouncner (approved):
Please wait for Daniel's approval.

dey4ss (approved):
Fine with what we discussed on Wendesday

PR #2694, opened 2025-04-02
Disable clang-tidy for the test folder

Greenscreen23:
Clangd highlights linter errors in VSCode (and other IDEs). It also does this for test files, which turn out very red and unreadable. This PR disables linting for the test folder, by placing a .clang-tidy file in the test folder that disables all rules.


tjeyy (approved):
Makes sense to me :)

PR #2695, opened 2025-04-03
Allow copy from table without JSON meta file

Greenscreen23:
This branch implements COPY FROM a .csv file into an existing table without requiring a .csv.json file. The csv file has to be separated by comma for this to work.

The csv parser now no longer searches for a .csv.json file itself. Instead, the caller to CsvReader::parse has to load the json file themselves. This was done because the import operator is now responsible for checking where to get the CsvMeta from, and I did not want to have duplicate logic in the import operator and csv parser. Since we want to move away from json files, it makes sense for the csv parser to not search for a json file automatically anymore.

Implementing a schema where one can specify that the csv file is separated by a different character will follow in a different PR.

(I'm waiting for a code review and approval before I add the fullci label)


src/lib/operators/import.cpp
```diff
@@ -51,9 +53,37 @@ std::shared_ptr<const Table> Import::_on_execute() {
   auto table = std::shared_ptr<Table>{};
 
   switch (_file_type) {
-    case FileType::Csv:
-      table = CsvParser::parse(filename, _chunk_size, _csv_meta);
+    case FileType::Csv: {
+      if (static_cast<int>(static_cast<bool>(_csv_meta)) +
```
tjeyy, line 57: I'm not sure how the others see this, but I'd prefer a priority ranking that chooses the first available definition source instead of failing if multiple are there.
I like the order determined by the if statement below.

src/lib/operators/import.cpp
```diff
@@ -51,9 +53,37 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
+              static_cast<int>(Hyrise::get().storage_manager.has_table(_tablename)) +
+              static_cast<int>(std::filesystem::exists(filename + CsvMeta::META_FILE_EXTENSION)) >=
+          2) {
+        Fail(
+            "Cannot load table from csv. Table definition source is ambiguous. At least two of the following sources "
+            "gave CsvMeta information: Import; existing table in DBMS; Meta file next to .csv file");
+      }
+
+      CsvMeta csv_meta;
```
tjeyy, line 66:
```suggestion
      auto csv_meta  = CsvMeta{};
```

tjeyy: According to the style guide, we should use auto-to-stick everywhere possible.

src/lib/operators/import.cpp
```diff
@@ -51,9 +53,37 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
+      if (_csv_meta) {
```
tjeyy, line 67: I just noticed that the `_csv_meta` optional of the `Import` operator is never set.
Since the `CsvParser` doesn't deal with such an optional anymore, let's get rid of it here, too.

src/lib/operators/import.cpp
```diff
@@ -51,9 +53,37 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
+        csv_meta = *_csv_meta;
+      } else if (Hyrise::get().storage_manager.has_table(_tablename)) {
+        TableColumnDefinitions column_definitions =
```
tjeyy, line 70: const & auto-to-stick

src/lib/operators/import.cpp
```diff
@@ -51,9 +53,37 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
+            Hyrise::get().storage_manager.get_table(_tablename)->column_definitions();
+        csv_meta.columns.resize(column_definitions.size());
+        for (TableColumnDefinitions::size_type i = 0; i < column_definitions.size(); i++) {
```
tjeyy, line 73:
```suggestion
        for (auto column_id = ColumnID{0}; column_id < column_count; ++column_id) {
```

tjeyy: The `ColumnID` type is the best fit to iterate over columns.
Auto-to-stick is preferred here, too.
We don't want to call `.size()` in loops since it is called each iteration, better save the size as a constant before.
Prefer the pre increment because the post increment returns an unnecessary rvalue to a copy of the old value of the index.

Probably the compiler would optimize these things anyway, but let's be sure.

src/lib/operators/import.cpp
```diff
@@ -51,9 +53,37 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
+      } else if (std::filesystem::exists(filename + CsvMeta::META_FILE_EXTENSION)) {
+        csv_meta = process_csv_meta_file(filename + CsvMeta::META_FILE_EXTENSION);
+      } else {
+        Fail("Cannot load table from csv. No table definition source found.");
+      }
+
+      table = CsvParser::parse(filename, csv_meta, _chunk_size);
       break;
+    }
     case FileType::Tbl:
```
tjeyy, line 87: Not your code, but please braces around the other cases for consistency.

src/lib/operators/import.cpp
tjeyy, line 57: I don't know how the others think about this, but I'd be fine just to prioritize the first definition source and not fail if there are multiple.

src/test/lib/import_export/csv/csv_parser_test.cpp
```diff
@@ -11,35 +11,42 @@ namespace hyrise {
 class CsvParserTest : public BaseTest {};
 
 TEST_F(CsvParserTest, SingleFloatColumn) {
-  auto table = CsvParser::parse("resources/test_data/csv/float.csv");
+  std::string csv_file = "resources/test_data/csv/float.csv";
```
tjeyy, line 14: const auto
Same for the other filename strings in this file.

tjeyy (changes requested):
Nice work!
I really like that the `CsvParser` doesn't need to deal with parsing the metadata JSON anymore.
Here are two suggested changes and some hints to the style guide :)

tjeyy:
> (I'm waiting for a code review and approval before I add the fullci label)

No need to wait :)

src/lib/operators/import.cpp
Greenscreen23: Failing if multiple are there was Martins proposal. His argument was that it is unintuitive if there is a hidden ranking somewhere, and that we should rather fail if two sources are given.

src/test/lib/import_export/csv/csv_parser_test.cpp
Greenscreen23: This was a bit more ticky, because c++ then turns this into a char array. I have found the operator""s, which fixes this. I'll commit this in a separate commit, so we can revert it if we want to.

https://en.cppreference.com/w/cpp/string/basic_string/operator%22%22s

src/lib/operators/import.cpp
Greenscreen23: It does not seem to be implemented, as the LQP ImportNode does not have a property for this, you are right. @Bouncner is it fine if we completely remove the _csv_meta property from the Import operator? Or should it be possible to load the csv_meta from SQL at some point?

src/lib/operators/import.cpp
tjeyy: Hm, sounds reasonable.
Let's keep it that way :)

tjeyy: It is!

src/test/lib/import_export/csv/csv_parser_test.cpp
tjeyy: std::string("hi") should do the work, too.

src/test/lib/import_export/csv/csv_parser_test.cpp
```diff
@@ -332,7 +330,7 @@ TEST_F(CsvParserTest, WithScheduler) {
 }
 
 TEST_F(CsvParserTest, ImmutableChunks) {
-  const auto csv_file = "resources/test_data/csv/float_int_large.csv"s;
+  const auto csv_file = std::string{"resources/test_data/csv/float_int_large.csv"};
```
Bouncner, line 333: I only skimmed over the files, but do we test cases where the table already exists?

Bouncner: And where, e.g., the table already stores data?

src/lib/operators/import.cpp
Bouncner: Having CSV options is definitely needed as this PR is meant to allow us loading from TBL files which are not separated by commas.
So yes, we need it. But as we also could have a meta file with all parameters, it is not required.

src/lib/operators/import.cpp
```diff
@@ -51,17 +53,49 @@ std::shared_ptr<const Table> Import::_on_execute() {
   auto table = std::shared_ptr<Table>{};
 
   switch (_file_type) {
-    case FileType::Csv:
-      table = CsvParser::parse(filename, _chunk_size, _csv_meta);
+    case FileType::Csv: {
+      if (static_cast<int>(static_cast<bool>(_csv_meta)) +
+              static_cast<int>(Hyrise::get().storage_manager.has_table(_tablename)) +
+              static_cast<int>(std::filesystem::exists(filename + CsvMeta::META_FILE_EXTENSION)) >=
```
Bouncner, line 59: Can we have much simple checks here? Performance does not matter.

If _csv_meta, check for table existing or fail.

If no _csv_meta, check that meta file exists and table does not. Be more explicit.

src/lib/operators/import.cpp
Bouncner: We might need clearer names here, as the csv meta file and `_csv_meta` are confusing.

src/benchmarklib/file_based_table_generator.cpp
```diff
@@ -109,7 +110,11 @@ std::unordered_map<std::string, BenchmarkTableInfo> FileBasedTableGenerator::gen
         if (extension == ".tbl") {
           table_info.table = load_table(*table_info.text_file_path, _benchmark_config->chunk_size);
         } else if (extension == ".csv") {
-          table_info.table = CsvParser::parse(*table_info.text_file_path, _benchmark_config->chunk_size);
+          table_info.table =
+              CsvParser::parse(*table_info.text_file_path,
+                               process_csv_meta_file(static_cast<std::string>(*table_info.text_file_path) +
```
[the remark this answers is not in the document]
Greenscreen23: It says `Invalid operands to binary expression ('std::filesystem::path' and 'const char *const')`, so I think its needed

src/test/lib/import_export/csv/csv_parser_test.cpp
Greenscreen23: This would be the wrong file to put it, but its also missing in `import_test.cpp`. I will add one.

src/lib/operators/import.cpp
Greenscreen23: I feel like this differs from what we discussed in person. Maybe it makes sense to talk about this in a call instead of here?

src/test/lib/operators/import_test.cpp
```diff
@@ -204,6 +207,31 @@ TEST_F(OperatorsImportTest, ReplaceExistingTable) {
...
+  Hyrise::get().storage_manager.add_table("a", existing_table);
+
+  auto expected_table = load_table("resources/test_data/tbl/float.tbl");
+  auto importer = std::make_shared<Import>("resources/test_data/csv/float_without_json_meta.csv", "a");
+  importer->execute();
+
+  EXPECT_TABLE_EQ_ORDERED(Hyrise::get().storage_manager.get_table("a"), expected_table);
+}
+
+TEST_F(OperatorsImportTest, RetrieveCsvMetaFromTable) {
```
Bouncner, line 222: Can you add a test where we append to an existing table? Or would that be out of scope for that PR?

Greenscreen23: I didn't think that we should be able to append to an existing table. And I don't think that the current code can do that.

src/lib/operators/import.cpp
Bouncner: As there are a couple of comments here referring to this topic: did we completely settle it last week? If not, let me know.

If so, feel free to resolve all the comments.

Greenscreen23:
In general, I think it is good practice if the person resolve the conversations they opened, as they can decide best whether its resolved or not. That's why I'm not resolving your conversations. But I think this PR might soon be ready. If you agree, then we can merge @Bouncner

Bouncner:
> In general, I think it is good practice if the person resolve the conversations they opened, as they can decide best whether its resolved or not. That's why I'm not resolving your conversations. But I think this PR might soon be ready. If you agree, then we can merge @Bouncner

I'd say it depends. There was one issue of asking you to add a comment. You reacted with a thumbs up, which is totally enough for me. I am too lazy to check out what you wrote in case you give a thumbs up. :)
If you are not sure, if you addressed it fully, simply ask "Is this comment okay?" and I know that I need to check the change.

src/lib/operators/import.hpp
```diff
@@ -18,7 +15,9 @@ namespace hyrise {
 /*
  * This operator reads a file, creates a table from that input and adds it to the storage manager.
  * Supported file types are .tbl, .csv and Hyrise .bin files.
- * For .csv files, a CSV config is additionally required, which is commonly located in the <filename>.json file.
+ * For .csv files, a CSV config is additionally required (if no table with the same name exists as a template).
+ * The operator looks for this content in the file <filename>.json.
+ * If a table with this filename already exists, it is dropped. This contradicts the COPY FROM semantic from SQL.
```
Bouncner, line 20: I think Daniel told me that COPY FROM is not SQL standard, but Postgres (@dey4ss, right?).
I would write ".. it is dropped (not appended as done in Postgres)".

src/test/lib/import_export/csv/csv_parser_test.cpp
Bouncner: Do we check that the assertation is raised when we have a json file but also the table exists?

Bouncner:
Looks good to me. Anything to add @dey4ss or @tjeyy ?

src/test/lib/import_export/csv/csv_parser_test.cpp
Greenscreen23: I added a test for that as well.

src/lib/operators/import.cpp
```diff
@@ -51,25 +50,51 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
   switch (_file_type) {
-    case FileType::Csv:
-      table = CsvParser::parse(filename, _chunk_size, _csv_meta);
+    case FileType::Csv: {
+      if (Hyrise::get().storage_manager.has_table(_tablename) &&
+          std::filesystem::exists(filename + CsvMeta::META_FILE_EXTENSION)) {
+        Fail(
+            "Cannot load table from csv, table definition source is ambiguous. The table already exists in DBMS and "
+            "there is a meta file next to .csv file");
+      }
```
dey4ss, line 59: Some comments here. First, I think storing if the table exists and if the file exists in variables with telling names makes sense. From a style POV, `if (...) { Fail(...); }` –> `Assert(...);`.

However, I do not think we should fail if we have both table _and_ JSON file. E.g., we automatically download the JOB data as CSVs with JSON files as part of our benchmark execution. If I now happen to use these CSVs to load data from, and I do the correct way Postgres defines `COPY` (with `CREATE TABLE` before), I get an error.

Same if I happen to load a CSV with JSON twice via `COPY` (which is perfectly fine) according to Postgres. The first call works fine, but the second one fails. I think that's counterintuitive.

I suggest to use the table definitions of an existing default as the first source of truth, and only look for a JSON file if there's no table. If there happens to be one and there already is a table, just ignore it.

src/lib/operators/import.cpp
```diff
@@ -91,6 +116,8 @@ std::shared_ptr<const Table> Import::_on_execute() {
     ChunkEncoder::encode_all_chunks(table, chunk_encoding_spec);
   }
 
+  Hyrise::get().storage_manager.add_table(_tablename, table);
```
dey4ss, line 119: Can you add a comment why the changed order has an impact?

src/lib/operators/import.hpp
```diff
@@ -18,7 +15,9 @@ namespace hyrise {
 /*
  * This operator reads a file, creates a table from that input and adds it to the storage manager.
  * Supported file types are .tbl, .csv and Hyrise .bin files.
- * For .csv files, a CSV config is additionally required, which is commonly located in the <filename>.json file.
+ * For .csv files, a CSV config is additionally required (if no table with the same name exists as a template).
+ * The operator looks for this content in the file <filename>.json.
+ * If a table with this filename already exists, it is dropped (not appended as done in Postgres).
```
dey4ss, line 20: I think it makes sense to match Postgres's behavior here (and add a test).

src/test/lib/import_export/csv/csv_parser_test.cpp
```diff
@@ -11,35 +11,42 @@ namespace hyrise {
 class CsvParserTest : public BaseTest {};
 
 TEST_F(CsvParserTest, SingleFloatColumn) {
-  auto table = CsvParser::parse("resources/test_data/csv/float.csv");
+  const auto csv_file = std::string{"resources/test_data/csv/float.csv"};
+  auto table = CsvParser::parse(csv_file, process_csv_meta_file(csv_file + CsvMeta::META_FILE_EXTENSION));
```
dey4ss, line 15: const (also below)

src/test/lib/operators/import_test.cpp
```diff
@@ -204,6 +207,42 @@ TEST_F(OperatorsImportTest, ReplaceExistingTable) {
...
+
+TEST_F(OperatorsImportTest, ThrowExceptionWithMultipleMetaSources) {
+  auto existing_table = CsvParser::create_table_from_meta_file("resources/test_data/csv/float.csv.json");
+
+  Hyrise::get().storage_manager.add_table("a", existing_table);
+
+  auto expected_table = load_table("resources/test_data/tbl/float.tbl");
+  auto importer = std::make_shared<Import>("resources/test_data/csv/float.csv", "a");
+
+  EXPECT_THROW(importer->execute(), std::exception);
```
dey4ss, line 243: `std::logic_error`. Also, I'd vote to rework this behavior (see comments above).

dey4ss (commented):
Some comments. I think we should stick to Postgres's behavior more and also just ignore meta files if a table already exists.

src/lib/operators/import.cpp
Bouncner: Not sure I agree. For JOB: Is there a reason for the json files if we have proper create tables? I think the JSONs are fine for testing, but they potentially cause confusion when both are supported.

I think following is Postgres is fine, but I did not see it in this PR. If it is a reasonable effort, let's do it here.

src/lib/operators/import.cpp
Bouncner: Since I requested the change, let me make a draft:
`// We create statistics when tables are added to the storage manager. As statistics can be expensive to create and their creation benefits from dictionary encoding, we add the tables after they are encoded.`

src/lib/operators/import.cpp
dey4ss: > For JOB: Is there a reason for the json files if we have proper create tables?

Guess it's convenience plus some functionality (NULL handling, quote, delimiter, ...) that `COPY` currently does not support and might never support. Just by comparing Postgres options (https://www.postgresql.org/docs/current/sql-copy.html) with `CsvMeta`, I see there could be things we could not express that easily:

```SQL
COPY table_name [ ( column_name [, ...] ) ]
    FROM { 'filename' | PROGRAM 'command' | STDIN }
    [ [ WITH ] ( option [, ...] ) ]
    [ WHERE condition ]

COPY { table_name [ ( column_name [, ...] ) ] | ( query ) }
    TO { 'filename' | PROGRAM 'command' | STDOUT }
    [ [ WITH ] ( option [, ...] ) ]

where option can be one of:

    FORMAT format_name
    FREEZE [ boolean ]
    DELIMITER 'delimiter_character'
    NULL 'null_string'
    DEFAULT 'default_string'
    HEADER [ boolean | MATCH ]
    QUOTE 'quote_character'
    ESCAPE 'escape_character'
    FORCE_QUOTE { ( column_name [, ...] ) | * }
    FORCE_NOT_NULL { ( column_name [, ...] ) | * }
    FORCE_NULL { ( column_name [, ...] ) | * }
    ON_ERROR error_action
    ENCODING 'encoding_name'
    LOG_VERBOSITY verbosity
```

https://github.com/hyrise/hyrise/blob/e9cbaab34bed765c2827c922c378a400b4f7c1c5/src/lib/import_export/csv/csv_meta.hpp#L26-L41

But yes, we could as well just use the `CREATE TABLE` statements JOB ships (or create our own with correct FKs at some point) and use Hyrise's default values implicitly. However, if we do that, I think we should get rid of the JSONs at all at some point (incl. new JOB zip etc).

> I think following is Postgres is fine, but I did not see it in this PR. If it is a reasonable effort, let's do it here.

It should be sth like that:
```c++
if (!table_exists) {
  Hyrise::get().storage_manager.add_table(_tablename, table);
} else {
  const auto existing_table = Hyrise::get().storage_manager.get_table(_tablename);
  const auto append_lock = existing_table->acquire_append_mutex();
  if (!existing_table->last_chunk()->is_immutable()) {
    existing_table->last_chunk()->set_immutable();
  }

    existing_table->append_chunk(table->get_chunk(chunk_id));
  }
}
```

We could even go a bit further, let `Import` inherit from `AbstractReadWriteOperator`,  add the new_table/chunks not in `_on_execute()` but in `_on_commit_records()` and set the current CID in the new table's/chunks MVCC data to make table import fully transaction-safe.

src/lib/import_export/csv/csv_parser.hpp
```diff
@@ -29,11 +27,11 @@ class CsvParser {
  public:
   /*
    * @param filename      Path to the input file.
-   * @param csv_meta      Custom csv meta information which will be used instead of the default "filename" + ".json" meta.
+   * @param csv_meta      Custom csv meta information (usually constained in the file "filename" + ".json")
```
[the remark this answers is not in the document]
Greenscreen23: I've rewritten the sentence a bit. Does this fit your needs?

src/lib/operators/import.cpp
Greenscreen23: I do not see why we can't just extend the postgres syntax with a few extra options to fully support our csv_meta set in the copy from statement. Which properties do you think are hard to implement? I am also fine with including the append functionality (whether it is with or without AbstractReadWriteOperator) in this PR. And if we completely stop supporting json meta files, we don't have to worry about the table exists and meta file exists problem.

src/lib/operators/import.hpp
Greenscreen23: I guess this is part of another comment, but I will leave this one open as a reminder to change this part as well.

src/lib/operators/import.cpp
dey4ss: > I do not see why we can't just extend the postgres syntax with a few extra options

That's not so much about Hyrise (where were do not aim to be fully compliant to Postgres) but the SQL parser. That component is used by other (also commercial) projects and we do not want to add stuff that's not Postgres-compatible (except for some legacy things that were done in the past). Let's discuss that in the roundtable meeting.

Bouncner:
Discussion from 2025-05-14:
- append to table, don't drop
- prefer table over json
- keep json support

Bouncner:
> Discussion from 2025-05-14:
> 
> * append to table, don't drop
> * prefer table over json
> * keep json support

This was all done, right?

src/lib/import_export/csv/csv_parser.cpp
```diff
@@ -92,11 +81,15 @@ std::shared_ptr<Table> CsvParser::parse(const std::string& filename, const Chunk
...
+    const auto escaped_linebreak =
+        std::string(1, csv_meta.config.delimiter_escape) + std::string(1, csv_meta.config.delimiter);
+
     // create and start parsing task to fill chunk
-    tasks.emplace_back(std::make_shared<JobTask>([relevant_content, field_ends, &table, &segments, &meta,
-                                                  &escaped_linebreak, &append_chunk_mutex]() {
-      _parse_into_chunk(relevant_content, field_ends, *table, segments, meta, escaped_linebreak, append_chunk_mutex);
-    }));
+    tasks.emplace_back(std::make_shared<JobTask>(
+        [relevant_content, field_ends, &table, &segments, &csv_meta, &escaped_linebreak, &append_chunk_mutex]() {
```
Bouncner: We copy `field_ends` every time to use it as a reference. Okay'ish I guess because we also write to `field_ends`, but I wonder if we could get around copying this structure every time.
E.g., by using a deque.

src/lib/operators/import.cpp
```diff
@@ -51,25 +51,45 @@ std::shared_ptr<const Table> Import::_on_execute() {
   auto table = std::shared_ptr<Table>{};
 
   switch (_file_type) {
-    case FileType::Csv:
-      table = CsvParser::parse(filename, _chunk_size, _csv_meta);
+    case FileType::Csv: {
+      auto csv_meta = CsvMeta{};
+      if (Hyrise::get().storage_manager.has_table(_tablename)) {
+        if (std::filesystem::exists(filename + CsvMeta::META_FILE_EXTENSION)) {
+          std::cerr << "Warning: Ignoring " << filename << CsvMeta::META_FILE_EXTENSION << " because table "
```
Bouncner: Are warnings cerr or cout?

src/lib/operators/import.cpp
```diff
@@ -91,6 +111,24 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
+    const auto existing_table = Hyrise::get().storage_manager.get_table(_tablename);
+    const auto append_lock = existing_table->acquire_append_mutex();
+    if (existing_table->chunk_count() > 0 && existing_table->last_chunk()->is_mutable()) {
+      existing_table->last_chunk()->set_immutable();
+    }
+
+    const auto chunk_count = table->chunk_count();
+    for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+      const auto chunk = table->get_chunk(chunk_id);
+      existing_table->append_chunk(chunk->get_segments(), chunk->mvcc_data());
```
Bouncner: We do not create statistics for the table (done when adding to the storage manager).

If we do  not add to the storage manager, but append to the table already registered in the storage manager, we should trigger a stats update.

src/test/lib/operators/import_test.cpp
```diff
@@ -193,12 +197,29 @@ TEST_F(OperatorsImportTest, UnknownFileExtension) {
...
+
+TEST_F(OperatorsImportTest, AppendToExistingTable) {
+  auto existing_table = load_table("resources/test_data/tbl/float.tbl");
+
+  Hyrise::get().storage_manager.add_table("a", existing_table);
+
+  auto expected_table = load_table("resources/test_data/tbl/float.tbl");
+  Assert(expected_table->chunk_count() == 1, "Testing code was only written to support one chunk tables");
+  const auto chunk = expected_table->get_chunk(ChunkID{0});
+  expected_table->append_chunk(chunk->get_segments(), chunk->mvcc_data());
```
Bouncner: What are you doing here?

src/test/lib/operators/import_test.cpp
```diff
@@ -193,12 +197,29 @@ TEST_F(OperatorsImportTest, UnknownFileExtension) {
...
+  EXPECT_TABLE_EQ_ORDERED(Hyrise::get().storage_manager.get_table("a"), expected_table);
+}
+
+TEST_F(OperatorsImportTest, AppendToExistingTable) {
+  auto existing_table = load_table("resources/test_data/tbl/float.tbl");
+
+  Hyrise::get().storage_manager.add_table("a", existing_table);
+
+  auto expected_table = load_table("resources/test_data/tbl/float.tbl");
+  Assert(expected_table->chunk_count() == 1, "Testing code was only written to support one chunk tables");
```
Bouncner: I don't get this comment.

src/test/lib/operators/import_test.cpp
Bouncner: Ah, got it. Can you move the first three lines before the importer?

Greenscreen23: I don't get this comment. All lines marked by your comment are already before the importer code.

src/test/lib/operators/import_test.cpp
Greenscreen23: The importer is supposed to append the table to itself. If the testing file would contain a big table with multiple chunks, our expected table would be wrong, as our code does not properly concatenate two tables. I can also rewrite it such that it supports multiple chunks, but that felt unnecessary.

src/lib/import_export/csv/csv_parser.cpp
Greenscreen23: There's a syntax to move objects into a lambda instead of copying. As the vector would be cleared afterwards anyway, it is fine to move it to the lambda and then reconstruct a new vector. I've added a commit.

src/lib/operators/import.cpp
Greenscreen23: From the codebase it seems like cerr is more common.

src/lib/operators/import.cpp
Greenscreen23: Do we have a better way to update statistics than to just regenerate all of them? I will do that for now.

Bouncner: Updating is kind of impossible (at least for histograms). So re-generating is the only viable solution I can see here.

src/test/lib/operators/import_test.cpp
Bouncner: `Assert(expected_table->chunk_count() == 1, "Expected single chunk.");` ?

src/lib/operators/import.cpp
Bouncner: Feel free to resolve.

src/lib/import_export/csv/csv_parser.cpp
```diff
@@ -92,12 +82,17 @@ std::shared_ptr<Table> CsvParser::parse(const std::string& filename, const Chunk
     // Remove processed part of the csv content
     content_view = content_view.substr(field_ends.back() + 1);
 
+    const auto escaped_linebreak =
+        std::string(1, csv_meta.config.delimiter_escape) + std::string(1, csv_meta.config.delimiter);
+
     // create and start parsing task to fill chunk
```
Bouncner:
```suggestion
    // Create and start parsing task to fill chunk.
```

Bouncner (approved):
Just a tiny thing in a test comment. @dey4ss : do you want to have another run?

src/lib/import_export/csv/csv_parser.cpp
```diff
@@ -81,7 +71,7 @@ std::shared_ptr<Table> CsvParser::parse(const std::string& filename, const Chunk
   auto tasks = std::vector<std::shared_ptr<AbstractTask>>{};
   auto field_ends = std::vector<size_t>{};
   auto append_chunk_mutex = std::mutex{};
-  while (_find_fields_in_chunk(content_view, *table, field_ends, meta)) {
+  while (_find_fields_in_chunk(content_view, *table, field_ends, csv_meta)) {
     // create empty chunk
```
dey4ss:
```suggestion
    // Create empty chunk.
```

src/lib/import_export/csv/csv_parser.cpp
```diff
@@ -92,12 +82,17 @@ std::shared_ptr<Table> CsvParser::parse(const std::string& filename, const Chunk
     // Remove processed part of the csv content
```
dey4ss:
```suggestion
    // Remove processed part of the CSV content.
```

src/lib/operators/import.cpp
```diff
@@ -51,25 +53,45 @@ std::shared_ptr<const Table> Import::_on_execute() {
   auto table = std::shared_ptr<Table>{};
 
   switch (_file_type) {
-    case FileType::Csv:
-      table = CsvParser::parse(filename, _chunk_size, _csv_meta);
+    case FileType::Csv: {
+      auto csv_meta = CsvMeta{};
+      if (Hyrise::get().storage_manager.has_table(_tablename)) {
+        if (std::filesystem::exists(filename + CsvMeta::META_FILE_EXTENSION)) {
```
dey4ss: (Nit-)Pick here: We _could_ cache the meta file name, the info if the table already exists, and the info if the meta file exists in variables. Could also be slightly easier to read with telling variable names.

src/lib/operators/import.cpp
```diff
@@ -91,6 +113,27 @@ std::shared_ptr<const Table> Import::_on_execute() {
...
+
+    const auto chunk_count = table->chunk_count();
+    for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+      const auto chunk = table->get_chunk(chunk_id);
+      existing_table->append_chunk(chunk->get_segments(), chunk->mvcc_data());
+    }
+
+    table->set_table_statistics(TableStatistics::from_table(*table));
+    generate_chunk_pruning_statistics(table);
+  } else {
```
dey4ss: Flip branches and use early return?

src/lib/storage/chunk.hpp
```diff
@@ -78,6 +78,7 @@ class Chunk : private Noncopyable {
    *       call `get_segment again`, be aware that the returned object might have changed.
    */
   std::shared_ptr<AbstractSegment> get_segment(ColumnID column_id) const;
+  const Segments& get_segments() const;
```
dey4ss: We're so inconsistent when naming getters with and without `get_` prefix. I would slightly prefer `segments()`, but not really a strong opinion.

dey4ss (commented):
Only small stuff

src/test/lib/operators/import_test.cpp
Greenscreen23: I believe that my version explains better why the assert is there. I can infer from the assertion that the code expects a single chunk.

Bouncner:
@dey4ss : Zündung?

dey4ss (approved):
> @dey4ss : Zündung?

Fel free to zünd as soon as the CI pipeline passes.

src/test/lib/operators/import_test.cpp
Bouncner: Well, imagine you get this error on the console. It conveys what the test was written for, but not what the actual issue is.
A simpler error with a brief comment might be more helpful.

PR #2698, opened 2025-04-30
Add cacheability information to optimization rules

Rob2U:
We only want Hyrise to cache query plans that will stay valid over time because using a (logical or physical) plan that becomes invalid could cause wrong results. Plans might turn invalid if they depend on temporary data patterns, like non-permanent UCCs found by the UccDiscoveryPlugin, or on related rules (FDs) based on those patterns. 

To avoid this, each optimization rule tells Hyrise whether the resulting plan is safe to store in the cache.

Note: This PR relies on #2599. We could also skip this and only depend on `master` but then #2600 would become messy because it relies on both cacheability information (this PR) and non-permanent UCCs (#2599).

### DIFF
--- a/src/lib/optimizer/optimizer.hpp
+++ b/src/lib/optimizer/optimizer.hpp
@@ -2,6 +2,7 @@
 
 #include <memory>
 #include <string>
+#include <utility>
 #include <vector>
 
 #include "cost_estimation/cost_estimator_logical.hpp"
@@ -42,7 +43,7 @@ class Optimizer final {
    * Returns optimized version of @param input.
    * @param rule_durations may be set in order to retrieve runtime information for each applied rule.
    */
-  std::shared_ptr<AbstractLQPNode> optimize(
+  std::pair<std::shared_ptr<AbstractLQPNode>, IsCacheable> optimize(
       std::shared_ptr<AbstractLQPNode> input,
       const std::shared_ptr<std::vector<OptimizerRuleMetrics>>& rule_durations = nullptr) const;
 
--- a/src/lib/optimizer/strategy/abstract_rule.hpp
+++ b/src/lib/optimizer/strategy/abstract_rule.hpp
@@ -11,10 +11,15 @@ class AbstractLQPNode;
 class LogicalPlanRootNode;
 class LQPSubqueryExpression;
 
+enum class IsCacheable : bool { Yes = true, No = false };
+
+constexpr IsCacheable operator&&(IsCacheable lhs, IsCacheable rhs) {
+  return lhs == IsCacheable::Yes && rhs == IsCacheable::Yes ? IsCacheable::Yes : IsCacheable::No;
+}
+
 class AbstractRule {
  public:
   virtual ~AbstractRule() = default;
-
   /**
    * This function applies the concrete Optimizer Rule to an LQP.
    * The default implementation
@@ -37,8 +42,12 @@ class AbstractRule {
    *
    * Rules can define their own strategy of optimizing subquery LQPs by overriding this function. See, for example, the
    * StoredTableColumnAlignmentRule.
+   *
+   * @return Whether the resulting optimized LQP can be cached by the optimizer. A plan may only be cached if the root
+   *    LQP and all subquery LQPs are cacheable. An optimized plan might be not cacheable for example if we used a UCC
+   *    for optimization of which we cannot be sure that it will still be valid the next time the same query occurs.
    */
-  virtual void apply_to_plan(const std::shared_ptr<LogicalPlanRootNode>& lqp_root) const;
+  virtual IsCacheable apply_to_plan(const std::shared_ptr<LogicalPlanRootNode>& lqp_root) const;
 
   virtual std::string name() const = 0;
 
@@ -53,8 +62,12 @@ class AbstractRule {
    *  left and the right side of the diamond. On both sides, we would reach the bottom of the diamond. From there, we
    *  would look at each node twice. visit_lqp prevents this by tracking which nodes have already been visited and
    *  avoiding visiting a node twice.
+   *
+   *  @return Whether the resulting optimized LQP can be cached by the optimizer. An optimized plan might be not
+   *    cacheable for example if we used a UCC for optimization of which we cannot be sure that it will still be valid
+   *    the next time the same query comes around.
    */
-  virtual void _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const = 0;
+  virtual IsCacheable _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const = 0;
 };
 
 }  // namespace hyrise
--- a/src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
+++ b/src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
@@ -26,7 +26,7 @@ using namespace hyrise::expression_functional;  // NOLINT(build/namespaces)
 void gather_rewrite_info(
     const std::shared_ptr<JoinNode>& join_node,
     std::vector<std::tuple<std::shared_ptr<JoinNode>, LQPInputSide, std::shared_ptr<PredicateNode>>>& rewritables,
-    IsCacheable& cacheable) { // NOLINT(misc-unused-parameters)
+    IsCacheable& cacheable) {  // NOLINT(misc-unused-parameters)
   const auto prunable_side = join_node->prunable_input_side();
   if (!prunable_side) {
     return;
--- a/src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
+++ b/src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
@@ -5,6 +5,7 @@
 #include <tuple>
 #include <vector>
 
+#include "abstract_rule.hpp"
 #include "expression/abstract_expression.hpp"
 #include "expression/binary_predicate_expression.hpp"
 #include "expression/expression_functional.hpp"
@@ -24,7 +25,8 @@ using namespace hyrise::expression_functional;  // NOLINT(build/namespaces)
 
 void gather_rewrite_info(
     const std::shared_ptr<JoinNode>& join_node,
-    std::vector<std::tuple<std::shared_ptr<JoinNode>, LQPInputSide, std::shared_ptr<PredicateNode>>>& rewritables) {
+    std::vector<std::tuple<std::shared_ptr<JoinNode>, LQPInputSide, std::shared_ptr<PredicateNode>>>& rewritables,
+    IsCacheable& cacheable) { // NOLINT(misc-unused-parameters)
   const auto prunable_side = join_node->prunable_input_side();
   if (!prunable_side) {
     return;
@@ -161,15 +163,16 @@ std::string JoinToPredicateRewriteRule::name() const {
   return name;
 }
 
-void JoinToPredicateRewriteRule::_apply_to_plan_without_subqueries(
+IsCacheable JoinToPredicateRewriteRule::_apply_to_plan_without_subqueries(
     const std::shared_ptr<AbstractLQPNode>& lqp_root) const {
   // `rewritables finally contains all rewritable join nodes, their unused input side, and the predicates to be used for
   // the rewrites.
   auto rewritables = std::vector<std::tuple<std::shared_ptr<JoinNode>, LQPInputSide, std::shared_ptr<PredicateNode>>>{};
+  auto cacheable = IsCacheable::Yes;
   visit_lqp(lqp_root, [&](const auto& node) {
     if (node->type == LQPNodeType::Join) {
       const auto join_node = std::static_pointer_cast<JoinNode>(node);
-      gather_rewrite_info(join_node, rewritables);
+      gather_rewrite_info(join_node, rewritables, cacheable);
     }
     return LQPVisitation::VisitInputs;
   });
@@ -177,6 +180,8 @@ void JoinToPredicateRewriteRule::_apply_to_plan_without_subqueries(
   for (const auto& [join_node, prunable_side, rewrite_predicate] : rewritables) {
     perform_rewrite(join_node, prunable_side, rewrite_predicate);
   }
+
+  return cacheable;
 }
 
 }  // namespace hyrise
--- a/src/lib/sql/sql_pipeline_statement.cpp
+++ b/src/lib/sql/sql_pipeline_statement.cpp
@@ -8,8 +8,6 @@
 #include <utility>
 #include <vector>
 
-#include <boost/algorithm/string.hpp>
-
 #include "SQLParser.h"
 #include "SQLParserResult.h"
 
@@ -26,9 +24,9 @@
 #include "operators/maintenance/drop_table.hpp"
 #include "operators/maintenance/drop_view.hpp"
 #include "optimizer/optimizer.hpp"
+#include "optimizer/strategy/abstract_rule.hpp"
 #include "scheduler/abstract_task.hpp"
 #include "scheduler/job_task.hpp"
-#include "sql/sql_pipeline_builder.hpp"
 #include "sql/sql_plan_cache.hpp"
 #include "sql/sql_translator.hpp"
 #include "types.hpp"
@@ -118,8 +116,8 @@ const SQLTranslationInfo& SQLPipelineStatement::get_sql_translation_info() {
   return _translation_info;
 }
 
-const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_optimized_logical_plan() {
-  if (_optimized_logical_plan) {
+std::pair<std::shared_ptr<AbstractLQPNode>, IsCacheable>& SQLPipelineStatement::get_optimized_logical_plan() {
+  if (_optimized_logical_plan.first) {
     return _optimized_logical_plan;
   }
 
@@ -132,7 +130,8 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_optimized_logi
       if (lqp_is_validated(plan) == (_use_mvcc == UseMvcc::Yes)) {
         // Copy the LQP for reuse as the LQPTranslator might modify mutable fields (e.g., cached output_expressions)
         // and concurrent translations might conflict.
-        _optimized_logical_plan = plan->deep_copy();
+        // Note that the plan we have received here was cached, so it is obviously cacheable.
+        _optimized_logical_plan = {plan->deep_copy(), IsCacheable::Yes};
         return _optimized_logical_plan;
       }
     }
@@ -156,8 +155,8 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_optimized_logi
   _metrics->optimizer_rule_durations = *optimizer_rule_durations;
 
   // Cache newly created plan for the according sql statement
-  if (lqp_cache && _translation_info.cacheable) {
-    lqp_cache->set(_sql_string, _optimized_logical_plan);
+  if (lqp_cache && _translation_info.cacheable && static_cast<bool>(_optimized_logical_plan.second)) {
+    lqp_cache->set(_sql_string, _optimized_logical_plan.first);
   }
 
   return _optimized_logical_plan;
@@ -193,8 +192,8 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
 
   if (!_physical_plan) {
     // "Normal" path in which the query plan is created instead of begin retrieved from cache
-    const auto& lqp = get_optimized_logical_plan();
-
+    const auto& optimization_result = get_optimized_logical_plan();
+    const auto& lqp = optimization_result.first;
     // Reset time to exclude previous pipeline steps
     started = std::chrono::steady_clock::now();
     _physical_plan = LQPTranslator{}.translate_node(lqp);
@@ -207,7 +206,8 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
   }
 
   // Cache newly created plan for the according sql statement (only if not already cached)
-  if (pqp_cache && !_metrics->query_plan_cache_hit && _translation_info.cacheable) {
+  if (pqp_cache && !_metrics->query_plan_cache_hit && _translation_info.cacheable &&
+      (_optimized_logical_plan.first && static_cast<bool>(_optimized_logical_plan.second))) {
     pqp_cache->set(_sql_string, _physical_plan);
   }
 
--- a/src/test/lib/optimizer/optimizer_test.cpp
+++ b/src/test/lib/optimizer/optimizer_test.cpp
@@ -166,13 +166,14 @@ TEST_F(OptimizerTest, VerifiesResults) {
     }
 
    protected:
-    void _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const override {
+    IsCacheable _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const override {
       // Change the `b` expression in the projection to `u`, which is not part of the input LQP.
       const auto projection_node = std::dynamic_pointer_cast<ProjectionNode>(lqp_root->left_input());
       if (!projection_node) {
-        return;
+        return IsCacheable::Yes;
       }
       projection_node->node_expressions[0] = _out_of_plan_expression;
+      return IsCacheable::Yes;
     }
 
     std::shared_ptr<AbstractExpression> _out_of_plan_expression;
@@ -200,11 +201,13 @@ TEST_F(OptimizerTest, OptimizesSubqueries) {
     }
 
    protected:
-    void _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const override {
+    IsCacheable _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const override {
       visit_lqp(lqp_root, [&](const auto& node) {
         nodes.emplace(node);
         return LQPVisitation::VisitInputs;
       });
+
+      return IsCacheable::Yes;
     }
 
     std::unordered_set<std::shared_ptr<AbstractLQPNode>>& nodes;
@@ -288,8 +291,9 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
     size_t& counter;
 
    protected:
-    void _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const override {
+    IsCacheable _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const override {
       ++counter;
+      return IsCacheable::Yes;
     }
   };
 
@@ -298,7 +302,7 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
   auto optimizer = Optimizer{};
   optimizer.add_rule(std::move(rule));
 
-  const auto optimized_lqp = optimizer.optimize(std::move(lqp));
+  const auto optimized_lqp = optimizer.optimize(std::move(lqp)).first;
   lqp = nullptr;
 
   /**
@@ -332,6 +336,16 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
   }
 }
 
+TEST_F(OptimizerTest, OptimizationWithoutKeyConstraintCacheable) {
+  // The resulting optimized LQP should be cacheable if no KeyConstraints or FunctionalDependencies are used.
+  auto optimizer = Optimizer::create_default_optimizer();
+
+  auto lqp = ProjectionNode::make(expression_vector(add_(b, subquery_a)),
+                                  PredicateNode::make(greater_than_(a, subquery_b), node_a));
+  const auto [_, cacheable] = optimizer->optimize(std::move(lqp));
+  EXPECT_TRUE(static_cast<bool>(cacheable));
+}
+
 TEST_F(OptimizerTest, PollutedCardinalityEstimationCache) {
   if constexpr (!HYRISE_DEBUG) {
     GTEST_SKIP();
@@ -348,7 +362,9 @@ TEST_F(OptimizerTest, PollutedCardinalityEstimationCache) {
     }
 
    protected:
-    void _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const override {}
+    IsCacheable _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const override {
+      return IsCacheable::Yes;
+    }
   };
 
   optimizer.add_rule(std::make_unique<MockRule>());
--- a/src/test/lib/optimizer/strategy/between_composition_rule_test.cpp
+++ b/src/test/lib/optimizer/strategy/between_composition_rule_test.cpp
@@ -2,6 +2,7 @@
 #include <string>
 #include <vector>
 
+#include "expression/abstract_expression.hpp"
 #include "expression/expression_functional.hpp"
 #include "logical_query_plan/aggregate_node.hpp"
 #include "logical_query_plan/join_node.hpp"
@@ -446,6 +447,13 @@ TEST_F(BetweenCompositionRuleTest, HandleMultipleEqualExpressions) {
   _apply_rule(_rule, _lqp);
 
   EXPECT_LQP_EQ(_lqp, expected_lqp);
+}
+
+TEST_F(BetweenCompositionRuleTest, CheckCacheability) {
+  auto input_lqp = std::dynamic_pointer_cast<AbstractLQPNode>(
+      PredicateNode::make(equals_(_a_a, 100), PredicateNode::make(equals_(_a_b, 100), _node_a)));
+  const auto is_cacheable = _apply_rule(_rule, input_lqp);
+  EXPECT_TRUE(static_cast<bool>(is_cacheable));
 }
 
 }  // namespace hyrise
--- a/src/test/lib/optimizer/strategy/chunk_pruning_rule_test.cpp
+++ b/src/test/lib/optimizer/strategy/chunk_pruning_rule_test.cpp
@@ -1,5 +1,8 @@
+#include <memory>
+
 #include "expression/expression_functional.hpp"
 #include "hyrise.hpp"
+#include "logical_query_plan/abstract_lqp_node.hpp"
 #include "logical_query_plan/aggregate_node.hpp"
 #include "logical_query_plan/dummy_table_node.hpp"
 #include "logical_query_plan/join_node.hpp"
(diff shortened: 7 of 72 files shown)


src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
```diff
@@ -24,7 +25,8 @@ using namespace hyrise::expression_functional;  // NOLINT(build/namespaces)
 
 void gather_rewrite_info(
     const std::shared_ptr<JoinNode>& join_node,
-    std::vector<std::tuple<std::shared_ptr<JoinNode>, LQPInputSide, std::shared_ptr<PredicateNode>>>& rewritables) {
+    std::vector<std::tuple<std::shared_ptr<JoinNode>, LQPInputSide, std::shared_ptr<PredicateNode>>>& rewritables,
+    IsCacheable& cacheable) {  // NOLINT(misc-unused-parameters)
   const auto prunable_side = join_node->prunable_input_side();
```
Rob2U: Used in #2600 ([here](https://github.com/hyrise/hyrise/pull/2600/files#diff-0a9271c5fb9d7a28b6776c6c2a384969a8fb01b7866309473af49a924dd32d93))

src/lib/optimizer/strategy/abstract_rule.hpp
```diff
@@ -37,8 +42,12 @@ class AbstractRule {
    *
    * Rules can define their own strategy of optimizing subquery LQPs by overriding this function. See, for example, the
    * StoredTableColumnAlignmentRule.
+   *
+   * @return Whether the resulting optimized LQP can be cached by the optimizer. A plan may only be cached if the root
+   *    LQP and all subquery LQPs are cacheable. An optimized plan might be not cacheable for example if we used a UCC
+   *    for optimization of which we cannot be sure that it will still be valid the next time the same query occurs.
    */
-  virtual void apply_to_plan(const std::shared_ptr<LogicalPlanRootNode>& lqp_root) const;
+  virtual IsCacheable apply_to_plan(const std::shared_ptr<LogicalPlanRootNode>& lqp_root) const;
```
dey4ss: We could also stick with the void return type and pass an `IsCacheable` reference as a parameter (which will be unused by most of the rules). Thus, most rules do not have to care about the return type. What do you think, @Rob2U?

dey4ss: (We might want to use this pattern later to pass an optimization context to the rules that holds state, such as data checks, information about which rules changed the plan, number of joins, etc. We could also introduce this context here with the cacheability information as the only member.)

dey4ss: Just had a brief discussion with @Bouncner. For the sake of extensibility, we agreed to pass an `OptimizationContext` struct. This struct should also be returned to the `SQLPipelineStatement` because it could also contain, e.g., conditions for plan reuse in the future.

Rob2U: Sure, that makes sense. Should I already include more than just the cacheability information in this `OptimizationContext`?

dey4ss: Do you have something in specific in mind?

Rob2U: I considered moving the `cost_estimator` that is currently a member of each rule into the context. This makes each rule entirely stateless.

src/test/lib/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -22,6 +24,8 @@ class BetweenCompositionRuleTest : public StrategyBaseTest {
  protected:
   void SetUp() override {
     _rule = std::make_shared<BetweenCompositionRule>();
+    _optimization_context =
+        OptimizationContext{std::make_shared<CostEstimatorLogical>(std::make_shared<CardinalityEstimator>())};
```
dey4ss: Pick here: that sounds like something that should be a member of `StrategyBaseTest`. Then, the context can also be accessed by `StrategyBaseTest::_apply_rule()` directly and does not need to be passed as a parameter. `_apply_rule()` can also always create a new context.

src/test/lib/optimizer/optimizer_test.cpp
```diff
@@ -332,6 +337,16 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
...
 
+TEST_F(OptimizerTest, OptimizationWithoutKeyConstraintCacheable) {
+  // The resulting optimized LQP should be cacheable if no KeyConstraints or FunctionalDependencies are used.
+  auto optimizer = Optimizer::create_default_optimizer();
+
+  auto lqp = ProjectionNode::make(expression_vector(add_(b, subquery_a)),
+                                  PredicateNode::make(greater_than_(a, subquery_b), node_a));
+  const auto [_, optimization_context] = optimizer->optimize_with_context(std::move(lqp));
+  EXPECT_TRUE(optimization_context->is_cacheable());
+}
```
dey4ss: Basically, not schema-given dependencies are a problem. In the end, I'm afraid this test is not really meaningful because we will not cover all cases where rules rewrite plans based on such dependencies here (this should be part of the rule tests).

What I could imagine is to have a test case here where we sneak in a mock rule that prohibits cacheability and ensure that information is still reflected in the returned context. We already do similar things in this test file:
https://github.com/hyrise/hyrise/blob/fbc8b1fd98fe5a0772162f2b4338f5a1755f7cb4/src/test/lib/optimizer/optimizer_test.cpp#L280-L299

src/lib/optimizer/strategy/join_to_predicate_rewrite_rule.cpp
dey4ss: I'd change that in the other PR. It's already requiring a `NOLINT` here.

src/lib/optimizer/optimization_context.hpp
```diff
@@ -0,0 +1,33 @@
...
+namespace hyrise {
+
+/**
+ * OptimizationContext is used to track metadata about the optimization process / resulting LQP. Currently, it only
+ * tracks whether the resulting LQP is cacheable or not. If it is not cacheable, the SQLPipeline will not cache the
+ * optimized LQP.
+ */
+struct OptimizationContext {
+  explicit OptimizationContext(std::shared_ptr<AbstractCostEstimator> init_cost_estimator)
+      : cost_estimator{std::move(init_cost_estimator)} {}
```
dey4ss: I guess we can remove the ctor and set the public member directly. I'm also not sure whether the shared ptr copy for the method call + move constructor is preferable over reference + copy ctor. We have to copy at some point anyways.

src/lib/optimizer/optimizer.hpp
```diff
@@ -39,13 +40,23 @@ class Optimizer final {
...
       std::shared_ptr<AbstractLQPNode> input,
       const std::shared_ptr<std::vector<OptimizerRuleMetrics>>& rule_durations = nullptr) const;
 
+  /**
+   * Returns optimized version of @param input LQP and the OptimizationContext used by all transformation rules.
+   * @param rule_durations may be set in order to retrieve runtime information for each applied rule.
+   * The OptimizationContext contains metadata about the optimization process, such as whether the resulting LQP
+   * is cacheable or not.
+   */
+  std::pair<std::shared_ptr<AbstractLQPNode>, std::shared_ptr<OptimizationContext>> optimize_with_context(
```
dey4ss: Can we actually return a unique ptr to the optimization context?

src/test/lib/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -443,9 +468,17 @@ TEST_F(BetweenCompositionRuleTest, HandleMultipleEqualExpressions) {
...
 
+  EXPECT_TRUE(_optimization_context.is_cacheable());
   EXPECT_LQP_EQ(_lqp, expected_lqp);
 }
 
+TEST_F(BetweenCompositionRuleTest, CheckCacheability) {
+  auto input_lqp = std::dynamic_pointer_cast<AbstractLQPNode>(
+      PredicateNode::make(equals_(_a_a, 100), PredicateNode::make(equals_(_a_b, 100), _node_a)));
+  _apply_rule(_rule, input_lqp, _optimization_context);
+  EXPECT_TRUE(_optimization_context.is_cacheable());
```
dey4ss: pick here: You can use `StrategyBaseTest`'s `_lqp` member. Please also have a look at the way we usually format LQPs we create in tests, e.g.: 
https://github.com/hyrise/hyrise/blob/fbc8b1fd98fe5a0772162f2b4338f5a1755f7cb4/src/test/lib/sql/sql_translator_test.cpp#L1398-L1406

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -206,8 +208,9 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
     _physical_plan->set_transaction_context_recursively(_transaction_context);
   }
 
-  // Cache newly created plan for the according sql statement (only if not already cached)
-  if (pqp_cache && !_metrics->query_plan_cache_hit && _translation_info.cacheable) {
+  // Cache newly created plan for the according sql statement (only if not already cached).
+  if (pqp_cache && !_metrics->query_plan_cache_hit && _translation_info.cacheable &&
+      _optimization_context->is_cacheable()) {
```
dey4ss: The problem here is that we _could_ have the case where we obtain the LQP from a cache, but have to create the PQP (e.g., because of different cache sizes or someone else issued the same query in between so no PQP was cached, but the LQP already is), right? In that case, the `_optimization_context` member is not set. However, the LQP was cached, so `is_cacheable()` must have been true.

Or is it something else?

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -192,12 +195,11 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
   }
 
   if (!_physical_plan) {
-    // "Normal" path in which the query plan is created instead of begin retrieved from cache
-    const auto& lqp = get_optimized_logical_plan();
-
+    // "Normal" path in which the query plan is created instead of begin retrieved from cache.
+    const auto& optimization_result = get_optimized_logical_plan();
```
dey4ss: I think I liked the old name better.

dey4ss (commented):
Some more comments

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -91,6 +91,8 @@ class SQLPipelineStatementTest : public BaseTest {
   std::shared_ptr<SQLPhysicalPlanCache> _pqp_cache;
 
   const std::string _select_query_a = "SELECT * FROM table_a";
+  const std::string _select_query_using_join_to_semi_join_optimization_a =
+      "SELECT table_b.a FROM table_a, table_b WHERE table_a.a = table_b.a";
```
dey4ss: That's not used (same for the keys constraint on `_table_a`), right? I think there should be tests for the caching behavior, but (similar to the optimizer test), these should mock rules where we explicitly prohibit caching.

src/lib/optimizer/optimization_context.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#include "cost_estimation/abstract_cost_estimator.hpp"
+
+namespace hyrise {
+
+/**
+ * OptimizationContext is used to track metadata about the optimization process / resulting LQP. Currently, it only
+ * tracks whether the resulting LQP is cacheable or not. If it is not cacheable, the SQLPipeline will not cache the
+ * optimized LQP.
+ */
```
dey4ss: I somehow miss a sentence that contains the words "OC holds state during the optimization"

src/lib/optimizer/strategy/abstract_rule.hpp
```diff
@@ -53,8 +53,13 @@ class AbstractRule {
    *  left and the right side of the diamond. On both sides, we would reach the bottom of the diamond. From there, we
    *  would look at each node twice. visit_lqp prevents this by tracking which nodes have already been visited and
    *  avoiding visiting a node twice.
+   *
+   *  @return Whether the resulting optimized LQP can be cached by the optimizer. An optimized plan might be not
+   *    cacheable for example if we used a UCC for optimization of which we cannot be sure that it will still be valid
+   *    the next time the same query comes around.
```
dey4ss: That's not true anymore. Remove?

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -149,14 +151,17 @@ const std::shared_ptr<AbstractLQPNode>& SQLPipelineStatement::get_optimized_logi
 
   auto optimizer_rule_durations = std::make_shared<std::vector<OptimizerRuleMetrics>>();
 
-  _optimized_logical_plan = _optimizer->optimize(std::move(unoptimized_lqp), optimizer_rule_durations);
+  auto [result_lqp, result_context] =
+      _optimizer->optimize_with_context(std::move(unoptimized_lqp), optimizer_rule_durations);
+  _optimized_logical_plan = result_lqp;
+  _optimization_context = std::move(result_context);
```
dey4ss: Can we just do the following?
```suggestion
  [_optimized_logical_plan, _optimization_context] =
      _optimizer->optimize_with_context(std::move(unoptimized_lqp), optimizer_rule_durations);
```

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -168,16 +173,16 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
     return _physical_plan;
   }
 
-  // If we need a transaction context but haven't passed one in, this is the last point where we can create it
+  // If we need a transaction context but haven't passed one in, this is the last point where we can create it.
```
dey4ss:
```suggestion
  // If we need a transaction context but have not passed one in, this is the last point where we can create it.
```

src/lib/sql/sql_pipeline_statement.cpp
```diff
@@ -206,8 +210,10 @@ const std::shared_ptr<AbstractOperator>& SQLPipelineStatement::get_physical_plan
     _physical_plan->set_transaction_context_recursively(_transaction_context);
   }
 
-  // Cache newly created plan for the according sql statement (only if not already cached)
-  if (pqp_cache && !_metrics->query_plan_cache_hit && _translation_info.cacheable) {
+  // Cache the newly created plan for the according sql statement (only if not already cached). If LQP was cached
+  // `_optimization_context` is set to `nullptr`. If the LQP was cached we can also safely cache the PQP.
```
dey4ss:
```suggestion
  // Cache the newly created PQP for the according SQL statement (only if not already cached). If the LQP was cached
  // (`_optimization_context` is set to `nullptr`), we can also safely cache the PQP.
```

src/test/lib/optimizer/optimizer_test.cpp
```diff
@@ -165,13 +166,15 @@ TEST_F(OptimizerTest, VerifiesResults) {
...
-    void _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root) const override {
+    void _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root,
+                                           OptimizationContext& optimization_context) const override {
       // Change the `b` expression in the projection to `u`, which is not part of the input LQP.
       const auto projection_node = std::dynamic_pointer_cast<ProjectionNode>(lqp_root->left_input());
       if (!projection_node) {
-        return;
+        optimization_context.set_not_cacheable();
+      } else {
+        projection_node->node_expressions[0] = _out_of_plan_expression;
```
dey4ss: This is not used at all in this test case, is it?

src/test/lib/optimizer/optimizer_test.cpp
```diff
@@ -331,6 +336,32 @@ TEST_F(OptimizerTest, OptimizesSubqueriesExactlyOnce) {
...
+                                           OptimizationContext& optimization_context) const override {
+      optimization_context.set_not_cacheable();
+    }
+  };
+
+  auto optimizer = Optimizer::create_default_optimizer();
+  optimizer->add_rule(std::make_unique<MockRule>());
+
+  auto lqp = ProjectionNode::make(expression_vector(add_(b, subquery_a)),
+                                  PredicateNode::make(greater_than_(a, subquery_b), node_a));
```
dey4ss:
```suggestion
  // clang-format off
  auto lqp =
  ProjectionNode::make(expression_vector(add_(b, subquery_a)),
    PredicateNode::make(greater_than_(a, subquery_b), 
      node_a));
  // clang-format on
```

src/test/lib/optimizer/strategy/between_composition_rule_test.cpp
```diff
@@ -445,7 +469,19 @@ TEST_F(BetweenCompositionRuleTest, HandleMultipleEqualExpressions) {
...
   EXPECT_LQP_EQ(_lqp, expected_lqp);
 }
 
+TEST_F(BetweenCompositionRuleTest, CheckCacheability) {
+  // clang-format off
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(
+      PredicateNode::make(equals_(_a_a, 100),
+          PredicateNode::make(equals_(_a_b, 100),
+            _node_a)));
```
dey4ss:
```suggestion
  _lqp =
  PredicateNode::make(equals_(_a_a, 100),
    PredicateNode::make(equals_(_a_b, 100),
      _node_a));
```

src/test/lib/optimizer/strategy/chunk_pruning_rule_test.cpp
```diff
@@ -455,10 +502,24 @@ TEST_F(ChunkPruningRuleTest, PredicateWithCorrelatedSubquery) {
...
   const auto expected_chunk_ids = std::vector<ChunkID>{ChunkID{0}};
   EXPECT_EQ(stored_table_node_1->pruned_chunk_ids(), expected_chunk_ids);
 }
 
+TEST_F(ChunkPruningRuleTest, CheckCacheability) {
+  const auto stored_table_node = StoredTableNode::make("fixed_string_compressed");
+
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(
+      PredicateNode::make(equals_(lqp_column_(stored_table_node, ColumnID{0}), "zzz")));
+  _lqp->set_left_input(stored_table_node);
```
dey4ss:
```suggestion
  _lqp = PredicateNode::make(equals_(stored_table_node->get_column("a"), "zzz"), stored_table_node);
```

src/test/lib/optimizer/strategy/column_pruning_rule_test.cpp
```diff
@@ -453,4 +473,16 @@ TEST_F(ColumnPruningRuleTest, AnnotatePrunableJoinInput) {
   }
 }
 
+TEST_F(ColumnPruningRuleTest, CheckCacheability) {
+  // clang-format off
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(
+      ExportNode::make("dummy.csv", FileType::Auto,
+          PredicateNode::make(greater_than_(a, 5),
+            node_abc)));
```
dey4ss:
```suggestion
  _lqp =
  ExportNode::make("dummy.csv", FileType::Auto,
    PredicateNode::make(greater_than_(a, 5),
      node_abc));
```

src/test/lib/optimizer/strategy/expression_reduction_rule_test.cpp
```diff
@@ -366,7 +384,18 @@ TEST_F(ExpressionReductionRuleTest, ApplyToLQP) {
...
 
+  EXPECT_TRUE(_optimization_context.is_cacheable());
   EXPECT_LQP_EQ(_lqp, expected_lqp);
 }
 
+TEST_F(ExpressionReductionRuleTest, CheckCacheability) {
+  // clang-format off
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(
+      AggregateNode::make(expression_vector(), expression_vector(sum_(b), count_star_(mock_node), avg_(b)),
+        mock_node));
```
dey4ss:
```suggestion
  _lqp =
  AggregateNode::make(expression_vector(), expression_vector(sum_(b), count_star_(mock_node), avg_(b)),
    mock_node));
```

src/test/lib/optimizer/strategy/in_expression_rewrite_rule_test.cpp
```diff
@@ -390,8 +424,16 @@ TEST_F(InExpressionRewriteRuleTest, AutoStrategy) {
...
 
+    EXPECT_TRUE(_optimization_context.is_cacheable());
     EXPECT_LQP_EQ(_lqp, expected_lqp);
   }
 }
 
+TEST_F(InExpressionRewriteRuleTest, CheckCacheability) {
+  auto rule = std::make_shared<InExpressionRewriteRule>();
+  rule->strategy = InExpressionRewriteRule::Strategy::ExpressionEvaluator;
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(PredicateNode::make(single_element_in_expression, node));
```
dey4ss:
```suggestion
  _lqp = PredicateNode::make(single_element_in_expression, node);
```

src/test/lib/optimizer/strategy/index_scan_rule_test.cpp
```diff
@@ -185,8 +199,18 @@ TEST_F(IndexScanRuleTest, NoIndexScanForSecondPredicate) {
...
   _apply_rule(rule, _lqp);
 
+  EXPECT_TRUE(_optimization_context.is_cacheable());
   EXPECT_EQ(static_cast<PredicateNode&>(*_lqp).scan_type, ScanType::TableScan);
   EXPECT_EQ(predicate_node->scan_type, ScanType::TableScan);
 }
 
+TEST_F(IndexScanRuleTest, CheckCacheability) {
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(PredicateNode::make(greater_than_(a, 10)));
+  _lqp->set_left_input(stored_table_node);
```
dey4ss:
```suggestion
  _lqp = PredicateNode::make(greater_than_(a, 10), stored_table_node);
```

src/test/lib/optimizer/strategy/index_scan_rule_test.cpp
```diff
@@ -185,8 +199,18 @@ TEST_F(IndexScanRuleTest, NoIndexScanForSecondPredicate) {
...
+
+  EXPECT_EQ(std::dynamic_pointer_cast<PredicateNode>(_lqp)->scan_type, ScanType::TableScan);
```
dey4ss: why?

src/test/lib/optimizer/strategy/join_ordering_rule_test.cpp
```diff
@@ -74,7 +76,25 @@ TEST_F(JoinOrderingRuleTest, MultipleJoinGraphs) {
...
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(AggregateNode::make(expression_vector(a_a), expression_vector(),
+      PredicateNode::make(equals_(a_a, b_b),
+        JoinNode::make(JoinMode::Cross,
+          node_a,
+          JoinNode::make(JoinMode::Left, equals_(b_b, d_d),
+            node_b,
+            PredicateNode::make(equals_(d_d, c_c),
+              JoinNode::make(JoinMode::Cross,
+                node_d,
+                node_c)))))));
```
dey4ss:
```suggestion
  _lqp =
  AggregateNode::make(expression_vector(a_a), expression_vector(),
    PredicateNode::make(equals_(a_a, b_b),
      JoinNode::make(JoinMode::Cross,
        node_a,
        JoinNode::make(JoinMode::Left, equals_(b_b, d_d),
          node_b,
          PredicateNode::make(equals_(d_d, c_c),
            JoinNode::make(JoinMode::Cross,
              node_d,
              node_c))))));
```

src/test/lib/optimizer/strategy/join_predicate_ordering_rule_test.cpp
```diff
@@ -121,7 +131,22 @@ TEST_F(JoinPredicateOrderingRuleTest, SemiGreaterAndEquiJoin) {
...
 
+TEST_F(JoinPredicateOrderingRuleTest, CheckCacheability) {
+  set_statistics_for_mock_node(node_b, 100,
+                               {GenericHistogram<int32_t>::with_single_bin(0, 40, 100, 5),
+                                GenericHistogram<int32_t>::with_single_bin(30, 70, 100, 5),
+                                GenericHistogram<int32_t>::with_single_bin(10, 50, 100, 5)});
+
+  const auto predicates = expression_vector(greater_than_(a_y, b_y), less_than_(a_z, b_z));
+
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(JoinNode::make(JoinMode::Inner, predicates, node_a, node_b));
```
dey4ss:
```suggestion
  _lqp = JoinNode::make(JoinMode::Inner, predicates, node_a, node_b);
```

src/test/lib/optimizer/strategy/null_scan_removal_rule_test.cpp
```diff
@@ -100,7 +110,14 @@ TEST_F(NullScanRemovalRuleTest, TableColumnDefinitionIsNotNullable) {
   const auto expected_lqp = table_node->deep_copy();
   _apply_rule(rule, _lqp);
 
+  EXPECT_TRUE(_optimization_context.is_cacheable());
   EXPECT_LQP_EQ(_lqp, expected_lqp);
 }
 
+TEST_F(NullScanRemovalRuleTest, CheckCacheability) {
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(PredicateNode::make(is_not_null_(table_node_column), table_node));
```
dey4ss:
```suggestion
  _lqp = PredicateNode::make(is_not_null_(table_node_column), table_node);
```

src/test/lib/optimizer/strategy/predicate_merge_rule_test.cpp
```diff
@@ -388,7 +406,14 @@ TEST_F(PredicateMergeRuleTest, NoRewriteDifferentSetOperationMode) {
 
   _apply_rule(rule, _lqp);
 
+  EXPECT_TRUE(_optimization_context.is_cacheable());
   EXPECT_LQP_EQ(_lqp, expected_lqp);
 }
 
+TEST_F(PredicateMergeRuleTest, CheckCacheability) {
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(PredicateNode::make(less_than_(a_a, 10), node_a));
```
dey4ss:
```suggestion
  _lqp = PredicateNode::make(less_than_(a_a, 10), node_a);
```

src/test/lib/optimizer/strategy/predicate_placement_rule_test.cpp
```diff
@@ -987,8 +1021,22 @@ TEST_F(PredicatePlacementRuleTest, DoNotMoveMultiPredicateSemiAndAntiJoins) {
...
   }
 }
 
+TEST_F(PredicatePlacementRuleTest, CheckCacheability) {
+  // clang-format off
+  _lqp = std::dynamic_pointer_cast<AbstractLQPNode>(PredicateNode::make(
+      or_(and_(equals_(_d_b, 1), equals_(_e_a, 10)), and_(equals_(_d_b, 2), equals_(_e_a, 1))),
+      JoinNode::make(JoinMode::Left, equals_(_d_a, _e_a),
+        _stored_table_d,
+        _stored_table_e)));
```
dey4ss: (format + superfluous cast just as before)

Also, is there a need for such a complex predicate just in a single predicate node that could be pushed down?

src/test/lib/optimizer/strategy/stored_table_column_alignment_rule_test.cpp
```diff
@@ -103,9 +111,16 @@ TEST_F(StoredTableColumnAlignmentRuleTest, CoverSubqueries) {
...
   _apply_rule(_rule, _lqp);
 
+  EXPECT_TRUE(_optimization_context.is_cacheable());
   EXPECT_EQ(_stored_table_node_left->pruned_column_ids(), pruned_column_set_a);
   EXPECT_EQ(_stored_table_node_right->pruned_column_ids(), pruned_column_set_a);
   EXPECT_EQ(stn_subquery->pruned_column_ids(), pruned_column_set_a);
 }
 
+TEST_F(StoredTableColumnAlignmentRuleTest, CheckCacheability) {
+  _lqp = _union_node->deep_copy();
```
dey4ss: why the copy?

src/test/lib/optimizer/strategy/strategy_base_test.cpp
```diff
@@ -5,11 +5,19 @@
...
 #include "logical_query_plan/abstract_lqp_node.hpp"
 #include "logical_query_plan/logical_plan_root_node.hpp"
+#include "optimizer/optimizer.hpp"
 #include "optimizer/strategy/abstract_rule.hpp"
 #include "statistics/cardinality_estimator.hpp"
 
 namespace hyrise {
 
+void StrategyBaseTest::SetUp() {
+  BaseTest::SetUp();
```
dey4ss: `BaseTest` has no own `SetUp` method.

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -713,6 +711,99 @@ TEST_F(SQLPipelineStatementTest, MetaTableNoCaching) {
...
+    }
+
+   protected:
+    void _apply_to_plan_without_subqueries(const std::shared_ptr<AbstractLQPNode>& lqp_root,
+                                           OptimizationContext& optimization_context) const override {
+      optimization_context.set_not_cacheable();
+    }
+  };
+
+  auto optimizer = Optimizer::create_default_optimizer();
```
dey4ss: const

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -713,6 +711,99 @@ TEST_F(SQLPipelineStatementTest, MetaTableNoCaching) {
...
+
+  auto sql_pipeline = SQLPipelineBuilder{query}
+                          .with_lqp_cache(_lqp_cache)
+                          .with_pqp_cache(_pqp_cache)
+                          .with_optimizer(optimizer)
+                          .create_pipeline();
+  auto statement = get_sql_pipeline_statements(sql_pipeline).at(0);
+  statement->get_result_table();
+
+  EXPECT_EQ(_lqp_cache->size(), 0u);
```
dey4ss:
```suggestion
  EXPECT_EQ(_lqp_cache->size(), 0);
```

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -713,6 +711,99 @@ TEST_F(SQLPipelineStatementTest, MetaTableNoCaching) {
...
+  EXPECT_FALSE(_lqp_cache->has(query));
+
+  EXPECT_EQ(_pqp_cache->size(), 0u);
```
dey4ss:
```suggestion
  EXPECT_EQ(_pqp_cache->size(), 0);
```

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -713,6 +711,99 @@ TEST_F(SQLPipelineStatementTest, MetaTableNoCaching) {
   EXPECT_FALSE(_pqp_cache->has(meta_table_query));
 }
 
+TEST_F(SQLPipelineStatementTest, NonCacheablePlanNotCached) {
+  const auto query = "SELECT * FROM table_a";
```
dey4ss: moce down to where it is actually used?

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -713,6 +711,99 @@ TEST_F(SQLPipelineStatementTest, MetaTableNoCaching) {
...
+
+  class MockRule : public AbstractRule {
```
dey4ss, line 717: Add a one-line comment what happens in the rule?

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -713,6 +711,99 @@ TEST_F(SQLPipelineStatementTest, MetaTableNoCaching) {
...
+  statement->get_result_table();
+
+  EXPECT_EQ(_lqp_cache->size(), 0u);
+  EXPECT_FALSE(_lqp_cache->has(query));
+
+  EXPECT_EQ(_pqp_cache->size(), 0u);
+  EXPECT_FALSE(_pqp_cache->has(query));
+}
+
+TEST_F(SQLPipelineStatementTest, CachedPlanStaysCached) {
```
dey4ss, line 749: (see comments for test above)

src/test/lib/sql/sql_pipeline_statement_test.cpp
```diff
@@ -713,6 +711,99 @@ TEST_F(SQLPipelineStatementTest, MetaTableNoCaching) {
...
+  optimizer->add_rule(std::make_unique<MockRule>());
+
+  auto sql_pipeline_non_cacheable = SQLPipelineBuilder{query}
+                                        .with_lqp_cache(_lqp_cache)
+                                        .with_pqp_cache(_pqp_cache)
+                                        .with_optimizer(optimizer)
+                                        .create_pipeline();
+
+  // Clear the PQPCache. Even though the new optimizer makes the plan non-cacheable, the previous plan should still be
+  // cached, because the optimizer is not invoked. Therefore the plan should now also appear in the PQP cache.
```
dey4ss:
```suggestion
  // Clear the PQP cache. Even though the new optimizer makes the plan non-cacheable, the previous plan should still be
  // cached because the optimizer is not invoked. Thus, the new PQP should appear in the PQP cache again.
```

dey4ss: Split the comment after the first sentence and move the second part down?

dey4ss (commented):
Nice work overall! Sorry to bother, but I still have some comments, mostly on the tests/formatting. I marked most (but not all) occurrences where LQPs are not formatted the way we usually do it and/or where there are superfluous pointer casts. Please also have a look at the formatting of comments: we prefer comments _above_ the code (in the same line only in rare cases, but I've seen quite some occurrences of that).

src/lib/optimizer/strategy/abstract_rule.hpp
```diff
@@ -52,11 +52,7 @@ class AbstractRule {
    *  The reason for this can be found in diamond LQPs: When using "trivial" recursion, we would go down both on the
    *  left and the right side of the diamond. On both sides, we would reach the bottom of the diamond. From there, we
    *  would look at each node twice. visit_lqp prevents this by tracking which nodes have already been visited and
-   *  avoiding visiting a node twice.
-   *
-   *  @return Whether the resulting optimized LQP can be cached by the optimizer. An optimized plan might be not
-   *    cacheable for example if we used a UCC for optimization of which we cannot be sure that it will still be valid
-   *    the next time the same query comes around.
+   *  avoiding visiting a node s.
```
dey4ss: something looks broken here

dey4ss (approved):
Looks pretty good except for one broken comment. Nice work! Feel free to merge once you fixed the comment and updated the PR description.

PR #2699, opened 2025-05-05
Add concurrency control and utilities for (key) constraints of a table

Rob2U:
This PR adds locking to the constraints stored for a table. Specifically a `_constraint_mutex` is added to a table. To acquire this mutex `acquire_constraints_modify_mutex` can be used.

To avoid redundant validation or invalidation of key constraints that are already known to be valid or invalid, two new timestamp attributes are introduced: `_last_validated_on` and `_last_invalidated_on`. These store the CommitID of the most recent validation or invalidation.

With this data, we can determine whether a `TableKeyConstraint` is still confidently valid or invalid by comparing the stored CommitIDs with those of the table’s chunks. This logic is implemented in two new methods: `is_constraint_confidently_valid` and `is_constraint_confidently_invalid`, located in `constraint_utils.cpp`.

As these methods are not yet used in the `UCCDiscoveryPlugin`, we can only add tests in the following locations:
- `src/test/lib/concurrency/stress_test.cpp`
- `src/test/lib/storage/constraints/constraint_utils_test.cpp`

### DIFF
--- a/src/lib/storage/constraints/constraint_utils.cpp
+++ b/src/lib/storage/constraints/constraint_utils.cpp
@@ -1,5 +1,6 @@
 #include "constraint_utils.hpp"
 
+#include <cstdint>
 #include <memory>
 #include <set>
 #include <string>
@@ -76,6 +77,56 @@ void order_constraint(const std::shared_ptr<Table>& table, const std::vector<std
   auto ordered_column_ids = column_ids_by_name(table, ordered_columns);
 
   table->add_soft_constraint(TableOrderConstraint{std::move(ordering_column_ids), std::move(ordered_column_ids)});
+}
+
+bool is_constraint_confidently_valid(const std::shared_ptr<Table>& table,
+                                     const TableKeyConstraint& table_key_constraint) {
+  if (!table_key_constraint.can_become_invalid()) {
+    return true;
+  }
+
+  if (!table_key_constraint.last_validated_on() ||
+      (table_key_constraint.last_invalidated_on() &&
+       table_key_constraint.last_validated_on() < *table_key_constraint.last_invalidated_on())) {
+    return false;
+  }
+
+  const auto chunk_count = table->chunk_count();
+  // Due to Hyrise being a append-only database the most recent chunks are the last ones added to the table. Therefore
+  // we iterate backwards through all chunks of the table to potentially return faster.
+  for (auto prev_chunk_id = static_cast<int32_t>(chunk_count - 1); prev_chunk_id >= 0; --prev_chunk_id) {
+    const auto source_chunk = table->get_chunk(static_cast<ChunkID>(prev_chunk_id));
+
+    // We use `max_begin_cid` here. This can lead to overly pessimistic results, but as of right now we don't have a
+    // better way to determine the last valid commit id here.
+    const auto max_begin_cid = source_chunk->mvcc_data()->max_begin_cid.load();
+    if (max_begin_cid != MAX_COMMIT_ID && max_begin_cid > table_key_constraint.last_validated_on()) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+bool is_constraint_confidently_invalid(const std::shared_ptr<Table>& table,
+                                       const TableKeyConstraint& table_key_constraint) {
+  if (!table_key_constraint.last_invalidated_on()) {
+    return false;
+  }
+
+  const auto chunk_count = table->chunk_count();
+  // Due to Hyrise being a append-only database the most recent chunks are the last ones added to the table. Therefore
+  // we iterate backwards through all chunks of the table to potentially return faster.
+  for (auto prev_chunk_id = static_cast<int32_t>(chunk_count - 1); prev_chunk_id >= 0; --prev_chunk_id) {
+    const auto source_chunk = table->get_chunk(static_cast<ChunkID>(prev_chunk_id));
+
+    const auto max_end_cid = source_chunk->mvcc_data()->max_end_cid.load();
+    if (max_end_cid != MAX_COMMIT_ID && max_end_cid > table_key_constraint.last_invalidated_on()) {
+      return false;
+    }
+  }
+
+  return true;
 }
 
 }  // namespace hyrise
--- a/src/lib/storage/constraints/table_key_constraint.cpp
+++ b/src/lib/storage/constraints/table_key_constraint.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <cstddef>
 #include <functional>
+#include <optional>
 #include <set>
 #include <utility>
 
@@ -14,17 +15,49 @@
 
 namespace hyrise {
 
-TableKeyConstraint::TableKeyConstraint(std::set<ColumnID>&& columns, const KeyConstraintType key_type)
-    : AbstractTableConstraint(TableConstraintType::Key), _key_type{key_type}, _columns{std::move(columns)} {
+TableKeyConstraint::TableKeyConstraint(std::set<ColumnID>&& columns, const KeyConstraintType key_type,
+                                       const std::optional<CommitID> last_validated_on,
+                                       const std::optional<CommitID> last_invalidated_on)
+    : AbstractTableConstraint(TableConstraintType::Key),
+      _columns{std::move(columns)},
+      _key_type{key_type},
+      _last_validated_on{last_validated_on},
+      _last_invalidated_on{last_invalidated_on} {
   Assert(!_columns.empty(), "Did not expect useless constraint.");
+}
+
+const std::set<ColumnID>& TableKeyConstraint::columns() const {
+  return _columns;
 }
 
 KeyConstraintType TableKeyConstraint::key_type() const {
   return _key_type;
 }
 
-const std::set<ColumnID>& TableKeyConstraint::columns() const {
-  return _columns;
+bool TableKeyConstraint::can_become_invalid() const {
+  return !_last_validated_on || _last_invalidated_on || _last_validated_on != MAX_COMMIT_ID;
+}
+
+const std::optional<CommitID>& TableKeyConstraint::last_validated_on() const {
+  return _last_validated_on;
+}
+
+const std::optional<CommitID>& TableKeyConstraint::last_invalidated_on() const {
+  return _last_invalidated_on;
+}
+
+void TableKeyConstraint::revalidated_on(const CommitID revalidation_commit_id) const {
+  Assert(!_last_validated_on || revalidation_commit_id >= _last_validated_on,
+         "Key constraint was already validated for larger commit id.");
+  _last_validated_on = revalidation_commit_id;
+}
+
+void TableKeyConstraint::invalidated_on(const CommitID invalidation_commit_id) const {
+  Assert(can_become_invalid(), "Cannot invalidate UCC that cannot become invalid.");
+
+  Assert(!_last_invalidated_on || invalidation_commit_id >= *_last_invalidated_on,
+         "Key constraint was already validated for larger commit id.");
+  _last_invalidated_on = invalidation_commit_id;
 }
 
 size_t TableKeyConstraint::hash() const {
--- a/src/lib/storage/constraints/table_key_constraint.hpp
+++ b/src/lib/storage/constraints/table_key_constraint.hpp
@@ -1,9 +1,11 @@
 #pragma once
 
+#include <optional>
 #include <set>
 #include <unordered_set>
 
 #include "abstract_table_constraint.hpp"
+#include "types.hpp"
 
 namespace hyrise {
 
@@ -18,14 +20,28 @@ class TableKeyConstraint final : public AbstractTableConstraint {
   /**
    * By requesting a std::set for the constructor, we ensure that the ColumnIDs have a well-defined order when creating
    * the vector (and equal constraints have equal columns). Thus, we can safely hash and compare key constraints without
-   * voilating the set semantics of the constraint.
+   * violating the set semantics of the constraint.
    */
-  TableKeyConstraint(std::set<ColumnID>&& columns, const KeyConstraintType key_type);
+  TableKeyConstraint(std::set<ColumnID>&& columns, const KeyConstraintType key_type,
+                     const std::optional<CommitID> last_validated_on = MAX_COMMIT_ID,
+                     const std::optional<CommitID> last_invalidated = {});
   TableKeyConstraint() = delete;
+
+  const std::set<ColumnID>& columns() const;
 
   KeyConstraintType key_type() const;
 
-  const std::set<ColumnID>& columns() const;
+  const std::optional<CommitID>& last_validated_on() const;
+  const std::optional<CommitID>& last_invalidated_on() const;
+
+  /**
+   * Returns whether or not this key constraint can become invalid if the table data changes. This is false for constraints specified
+   * by the table schema, but true for the "spurious" uniqueness of columns in any table state as adding duplicates
+   * would make them no longer unique.
+   */
+  bool can_become_invalid() const;
+  void revalidated_on(const CommitID revalidation_commit_id) const;
+  void invalidated_on(const CommitID invalidation_commit_id) const;
 
   size_t hash() const override;
 
@@ -39,13 +55,21 @@ class TableKeyConstraint final : public AbstractTableConstraint {
  protected:
   bool _on_equals(const AbstractTableConstraint& table_constraint) const override;
 
-  KeyConstraintType _key_type;
-
   /**
    * A std::set orders the columns ascending out of the box, which is desirable when printing them or comparing key
    * constraints.
    */
   std::set<ColumnID> _columns;
+
+  KeyConstraintType _key_type;
+
+  /**
+   * Commit ID of the snapshot this constraint was last validated on. Note that the constraint will still be valid
+   * during transactions with larger commit IDs if the table this constraint belongs to has not been modified since.
+   */
+  mutable std::optional<CommitID> _last_validated_on;
+  // The first element indicates if the constraint was invalid before, the second when it was.
+  mutable std::optional<CommitID> _last_invalidated_on;
 };
 
 using TableKeyConstraints = std::unordered_set<TableKeyConstraint>;
--- a/src/lib/storage/table.cpp
+++ b/src/lib/storage/table.cpp
@@ -228,7 +228,7 @@ void Table::append_mutable_chunk() {
 
   auto mvcc_data = std::shared_ptr<MvccData>{};
   if (_use_mvcc == UseMvcc::Yes) {
-    mvcc_data = std::make_shared<MvccData>(_target_chunk_size, MvccData::MAX_COMMIT_ID);
+    mvcc_data = std::make_shared<MvccData>(_target_chunk_size, MAX_COMMIT_ID);
   }
 
   append_chunk(segments, mvcc_data);
--- a/src/lib/storage/table.cpp
+++ b/src/lib/storage/table.cpp
@@ -9,6 +9,7 @@
 #include <mutex>
 #include <optional>
 #include <set>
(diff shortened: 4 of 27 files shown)


src/test/lib/logical_query_plan/static_table_node_test.cpp
```diff
@@ -57,7 +57,7 @@ TEST_F(StaticTableNodeTest, HashingAndEqualityCheck) {
 
   EXPECT_EQ(same_static_table_node->hash(), static_table_node->hash());
   EXPECT_NE(different_static_table_node_by_definitions->hash(), static_table_node->hash());
-  EXPECT_NE(different_static_table_node_by_constraints->hash(), static_table_node->hash());
+  EXPECT_EQ(different_static_table_node_by_constraints->hash(), static_table_node->hash());
```
dey4ss, line 60: Add a comment why the hash should be equal?

I think we more and more abandoned test cases that assert hashes are NOT equal (what do we know about hash collisions), so feel free to have a look at these cases.

src/test/lib/optimizer/strategy/join_to_predicate_rewrite_rule_test.cpp
```diff
@@ -54,7 +54,7 @@ INSTANTIATE_TEST_SUITE_P(JoinToPredicateRewriteRuleJoinModeTestInstance, JoinToP
 
 TEST_P(JoinToPredicateRewriteRuleJoinModeTest, PerformRewrite) {
   // The rule should only rewrite inner and semi joins.
-  auto key_constraints = TableKeyConstraints{};
+  auto key_constraints = std::unordered_set<TableKeyConstraint>{};
```
dey4ss, line 57: pick here: see comment in MockNode

dey4ss (commented):
I have a couple of comments. I was also surprised that the follwoing code remained unchanges, meaning that we do not re-validate a UCC that might have become invalid:
https://github.com/hyrise/hyrise/blob/2399fdccc98ae07727e2c7c6e1c4d5ca59ff073e/src/plugins/ucc_discovery_plugin.cpp#L133-L142

src/lib/storage/constraints/table_key_constraint.cpp
```diff
@@ -5,26 +5,62 @@
...
+  return last_invalidated == MAX_COMMIT_ID || (last_validated != MAX_COMMIT_ID && last_invalidated >= last_validated);
+}
+
+std::atomic<CommitID>& TableKeyConstraint::last_validated_on() const {
+  return _last_validated_on;
+}
+
+std::atomic<CommitID>& TableKeyConstraint::last_invalidated_on() const {
+  return _last_invalidated_on;
+}
```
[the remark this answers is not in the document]
Rob2U: No. After adapting `void TableKeyConstraint::revalidated_on(const CommitID revalidation_commit_id) const ` I forgot to reset `last_validated_on`.

Rob2U:
As pointed out above, we do not re-validate a UCC that might have become invalid yet. This change is introduced in #2599. 
([direct reference](https://github.com/hyrise/hyrise/pull/2599/files#r2097931861))

src/lib/storage/constraints/constraint_utils.cpp
```diff
@@ -78,4 +79,57 @@ void order_constraint(const std::shared_ptr<Table>& table, const std::vector<std
...
+    const auto max_begin_cid = source_chunk->mvcc_data()->max_begin_cid.load();
+    if (max_begin_cid != MAX_COMMIT_ID && max_begin_cid > last_validated_on) {
+      return false;
+    }
+  }
+
+  return true;
+}
+
+bool is_constraint_confidently_invalid(const std::shared_ptr<Table>& table,
```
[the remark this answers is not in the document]
Rob2U: Exactly this is used in #2599 (see comment below).

src/lib/logical_query_plan/mock_node.hpp
```diff
@@ -59,8 +60,8 @@ class MockNode : public EnableMakeForLQPNode<MockNode>, public AbstractLQPNode {
   void set_table_statistics(const std::shared_ptr<TableStatistics>& table_statistics);
 
   // Pure container functionality: MockNode does not use key constraints internally.
-  void set_key_constraints(const TableKeyConstraints& key_constraints);
-  const TableKeyConstraints& key_constraints() const;
+  void set_key_constraints(const std::unordered_set<TableKeyConstraint>& key_constraints);
```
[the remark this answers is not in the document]
Rob2U: I think you are right, the fact that we are not using `TableKeyConstraints` does not bring any benefits but probably causes confusion. Originally, I thought the other way around that using a concurrent data structure is unnecessary, but it should not impose measurable cost.

src/lib/storage/constraints/table_key_constraint.hpp
```diff
@@ -39,16 +98,26 @@ class TableKeyConstraint final : public AbstractTableConstraint {
...
   std::set<ColumnID> _columns;
+
+  KeyConstraintType _key_type;
+
+  /**
+   * Commit ID of the snapshot this constraint was last validated on. Note that the constraint will still be valid
+   * during transactions with larger commit IDs if the table this constraint belongs to has not been modified since.
+   */
+  mutable std::atomic<CommitID> _last_validated_on;
+  // The first element indicates if the constraint was invalid before, the second when it was.
```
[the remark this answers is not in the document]
Rob2U: I replaced the comment with an up-to-date one.

src/lib/storage/constraints/table_key_constraint.cpp
```diff
@@ -5,26 +5,62 @@
...
+
 KeyConstraintType TableKeyConstraint::key_type() const {
   return _key_type;
 }
 
-const std::set<ColumnID>& TableKeyConstraint::columns() const {
-  return _columns;
+bool TableKeyConstraint::can_become_invalid() const {
+  return _last_validated_on.load() != 0 || _last_invalidated_on.load() != MAX_COMMIT_ID ||
+         _last_validated_on == MAX_COMMIT_ID;
```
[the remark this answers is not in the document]
Rob2U: Good catch, that logic is unnecessary, as `_last_validated_on != 0` already implicates the rest of the expression.

src/lib/storage/constraints/table_key_constraint.hpp
```diff
@@ -39,16 +98,26 @@ class TableKeyConstraint final : public AbstractTableConstraint {
...
+  mutable std::atomic<CommitID> _last_invalidated_on;
 };
 
-using TableKeyConstraints = std::unordered_set<TableKeyConstraint>;
+using TableKeyConstraints = tbb::concurrent_unordered_set<TableKeyConstraint>;
+
+std::size_t hash_value(const hyrise::TableKeyConstraint& table_key_constraint);
```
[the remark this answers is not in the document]
Rob2U: TBB does not need it, so I will remove it. However, boost does ;)

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -695,4 +707,87 @@ TEST_F(StressTest, VisibilityOfInsertsBeingRolledBack) {
...
+  Hyrise::get().storage_manager.add_table("dummy_table", table);
+
+  /** Run multiple modifications and additions concurrently that could potentially lead to race conditions:
+   * - `UccDiscoveryPlugin::_validate_ucc_candidates`
+   * - `StaticTableNode::unique_column_combinations`
+   * - `Table::soft_key_constraints().clear()` to force a revalidation of the constraints. This does not happen in
+   * production, but we want to test the concurrent insertion of constraints.
+   * NOTE: We do not execute `Table::add_soft_constraint` because we do not have the knowledge whether or not the
+   * constraint already exists.
+   */
```
dey4ss, line 734: could you adjust the comment with the recent changes? Also, we're never in production :)

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -695,4 +707,87 @@ TEST_F(StressTest, VisibilityOfInsertsBeingRolledBack) {
...
+
+  Hyrise::get().default_pqp_cache = std::make_shared<SQLPhysicalPlanCache>();
+  Hyrise::get().default_lqp_cache = std::make_shared<SQLLogicalPlanCache>();
+  auto& pm = Hyrise::get().plugin_manager;
+  std::shared_mutex table_constraints_mutex;
```
dey4ss, line 739:
```suggestion
  auto table_constraints_mutex = std::shared_mutex{};
```

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -695,4 +707,87 @@ TEST_F(StressTest, VisibilityOfInsertsBeingRolledBack) {
...
+
+  pm.load_plugin(build_dylib_path("libhyriseUccDiscoveryPlugin"));
+  const auto validate_constraint = [&] {
+    std::shared_lock<std::shared_mutex> lock(table_constraints_mutex);
```
dey4ss, line 743:
```suggestion
    const auto lock = std::shared_lock{table_constraints_mutex};
```

dey4ss: (also below)

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -695,4 +707,87 @@ TEST_F(StressTest, VisibilityOfInsertsBeingRolledBack) {
...
+    // Populate the plan cache.
+    const std::string sql = "SELECT b,c FROM dummy_table GROUP BY b,c;";
```
dey4ss, line 745: auto to stick

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -695,4 +707,87 @@ TEST_F(StressTest, VisibilityOfInsertsBeingRolledBack) {
...
+      case 1:
+        threads.emplace_back(validate_constraint);
+        break;
+      case 2:
+        threads.emplace_back(clear_table_constraints);
+        break;
+      default:
+        break;
+    }
+  }
```
dey4ss, line 783: Without multiple loops and 100 threads, I'm not sure this will not just be executed quite on the fly. Could you have a look at other stress tests? I suggest that the threads (i) wait on an `atomic_flag` to start execution alltogether and (ii) run their code multiple times after.

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -695,4 +707,87 @@ TEST_F(StressTest, VisibilityOfInsertsBeingRolledBack) {
...
+  const auto clear_table_constraints = [&] {
+    std::unique_lock<std::shared_mutex> lock(table_constraints_mutex);
+    clear_soft_key_constraints(table);
+  };
+
+  const auto static_table_node_constraint_access = [&] {
+    std::shared_lock<std::shared_mutex> lock(table_constraints_mutex);
+    const auto static_table_node = StaticTableNode::make(table);
+    // Simply access the unique column combinations to require a `shared_lock`.
+    static_table_node->unique_column_combinations();
```
dey4ss, line 761: Can you use a `StoredTableNode` here (because this is the one that used in actual queries)? Furthermore, I would almost bet on this code to be optimized away in release builds. Maybe wrap this in some sort of assert? E.g., like the final check, `ASSERT_LE(stored_table_node->unique_column_combinations().size(), 3);`?

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -695,4 +707,87 @@ TEST_F(StressTest, VisibilityOfInsertsBeingRolledBack) {
...
+        threads.emplace_back(static_table_node_constraint_access);
+        break;
+      case 1:
+        threads.emplace_back(validate_constraint);
+        break;
+      case 2:
+        threads.emplace_back(clear_table_constraints);
+        break;
+      default:
+        break;
```
dey4ss, line 781: `Fail(...)`?
I guess two if statements would also be fine.

src/test/lib/logical_query_plan/mock_node_test.cpp
```diff
@@ -1,8 +1,11 @@
+#include <unordered_set>
```
dey4ss, line 1: superfluous?

src/test/lib/storage/constraints/constraint_utils_test.cpp
```diff
@@ -60,7 +63,8 @@ TEST_F(ConstraintUtilsTest, TableKeyConstraint) {
   EXPECT_THROW(primary_key_constraint(_table_a, {"a"}), std::logic_error);
   EXPECT_THROW(unique_constraint(_table_a, {"a"}), std::logic_error);
   EXPECT_THROW(primary_key_constraint(_table_a, {"c"}), std::logic_error);
-  EXPECT_THROW(unique_constraint(_table_a, {"c"}), std::logic_error);
+  EXPECT_THROW(unique_constraint(_table_a, {"c"}),
+               std::logic_error);  // Trying to add an already existing unique constraint is tolerated.
```
dey4ss, line 67: comment is off

src/test/lib/storage/constraints/constraint_utils_test.cpp
```diff
@@ -128,4 +132,44 @@ TEST_F(ConstraintUtilsTest, TableOrderConstraint) {
...
+  EXPECT_FALSE(key_constraint_is_confidently_valid(
+      _table_b, {{ColumnID{1}}, KeyConstraintType::UNIQUE, MAX_COMMIT_ID, CommitID{0}}));
+
+  // Manually modify the `max_begin_cid` of the chunk to simulate an insert to the table.
+  _table_b->append({0, 1});
+  _table_b->get_chunk(ChunkID{0})->mvcc_data()->max_begin_cid = CommitID{2};
+
+  // The first constraint is permanent and therefore valid, the second one was verified on a previous CommitID so we do
+  // not know whether or not it is still valid and the last constraint has a CommmitID that is up-to-date and is
+  // therefore certainly valid.
```
dey4ss, line 148: Could you state the individual parts directly before the asserts?

src/test/lib/storage/constraints/constraint_utils_test.cpp
```diff
@@ -128,4 +132,44 @@ TEST_F(ConstraintUtilsTest, TableOrderConstraint) {
...
+  EXPECT_TRUE(key_constraint_is_confidently_invalid(
+      _table_b, {{ColumnID{1}}, KeyConstraintType::UNIQUE, MAX_COMMIT_ID, CommitID{0}}));
+
+  // Manually modify the `max_end_cid` of the chunk to simulate a delete to the table.
+  _table_b->append({0, 1});
+  _table_b->get_chunk(ChunkID{0})->mvcc_data()->max_end_cid = CommitID{2};
+
+  // The first constraint is permanent and therefore NOT confidently invalid, the second was never verified but only
+  // invalidated on a CommitID prior to the deletion so it is also NOT confidently invalid and the last constraint was
+  // invalidated on a CommitID that is up-to-date and therefore certainly invalid.
```
dey4ss, line 167: (see above)

dey4ss (commented):
Some comments, mostly on the stress test

src/lib/storage/table.cpp
```diff
@@ -497,11 +495,12 @@ void Table::_add_soft_key_constraint(const TableKeyConstraint& table_key_constra
 
     // Ensure there is only one key constraint per column(s). Theoretically, there could be two unique constraints
     // {a, b} and {b, c}, but for now we prohibit these cases.
-    Assert(!columns_intersect(existing_constraint.columns(), table_key_constraint.columns()),
+    Assert((existing_constraint.columns().size() == 1 && table_key_constraint.columns().size() == 1) ||
+               !columns_intersect(existing_constraint.columns(), table_key_constraint.columns()),
```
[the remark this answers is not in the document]
Rob2U: This is deprecated.
(It was necessary for allowing calls of `add_soft_constraint` for a `TableKeyConstraint` that was already stored for the table)

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -695,4 +710,104 @@ TEST_F(StressTest, VisibilityOfInsertsBeingRolledBack) {
...
+  const auto stored_table_node_constraint_access = [&] {
+    start_flag.wait(false);
+    while (!stop_flag.test()) {
+      const auto stored_table_node = std::make_shared<StoredTableNode>("dummy_table");
+      // Access the unique column combinations. We need to lock here because `unique_column_combinations` uses a
+      // reference to iterate over the constraints. This reference is invalidated when the constraints are cleared.
+      while (writer_waiting) {
+        std::this_thread::sleep_for(SLEEP_TIME);
+      }
+      const auto lock = std::shared_lock{deletion_mutext};
```
dey4ss, line 780: do we need both the lock and the wait flag?

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -695,4 +710,104 @@ TEST_F(StressTest, VisibilityOfInsertsBeingRolledBack) {
...
+      ASSERT_LE(stored_table_node->unique_column_combinations().size(), 3);
+    }
+  };
+
+  // Start running the different modifications in parallel.
+  const auto thread_count = 10;
```
dey4ss, line 786: Guess you can easily do 50 or 100 here. Just double-check the (sanitizer) runtime in the pipeline.

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -695,4 +710,104 @@ TEST_F(StressTest, VisibilityOfInsertsBeingRolledBack) {
...
+
+  stop_flag.test_and_set();
+  stop_flag.notify_all();
+
+  for (auto& thread : threads) {
+    thread.join();
+  }
+
+  // Check that the set of TableKeyConstraints does not contain any duplicates.
+  ASSERT_LE(table->soft_key_constraints().size(), 3);
```
dey4ss, line 810: should be empty, right? the validation thread finally clears the constraints

dey4ss (dismissed):
Nice work! Feel free to merge once you addressed the comments.

src/test/lib/concurrency/stress_test.cpp
Rob2U: According to the tbb documentation for the `concurrent_unordered_set` [here](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/containers/concurrent_unordered_set_cls/unsafe_modifiers) `clear` "can only be performed serially".  Therefore, we need this lock to prevent `unique_column_combinations` from iterating over our set.

dey4ss:
`StressTest.AddModifyTableKeyConstraintsConcurrently` spins again in every build. Aborting for now. If you need access to a machine for testing, let me know.

Rob2U:
Now, I remember the exact reason for the `writer_waiting` atomic :) : The issue is that the unique lock in `validate_constraint` is waiting for all shared locks in `stored_table_node_constraint_access`. While this unique lock is waiting, new iterations in `stored_table_node_constraint_access` can require new shared locks. This blocks `validate_constraint`.
My solution is adding the `writer_waiting` atomic that, when set, stops the acquisition of new shared locks.

dey4ss (approved):

Bouncner:
> Now, I remember the exact reason for the `writer_waiting` atomic :) : The issue is that the unique lock in `validate_constraint` is waiting for all shared locks in `stored_table_node_constraint_access`. While this unique lock is waiting, new iterations in `stored_table_node_constraint_access` can require new shared locks. This blocks `validate_constraint`. My solution is adding the `writer_waiting` atomic that, when set, stops the acquisition of new shared locks.
> 
> *Apparently, this is not specified in the cpp standard ([stackoverflow](https://stackoverflow.com/questions/32243245/can-thread-trying-to-stdlock-unique-an-stdshared-mutex-be-starved-by-threa?utm_source=chatgpt.com)).

Can you add this link as a comment? I was sure libs would ensure no starvations.

PR #2700, opened 2025-05-12
Use specific header for `boost::sort::pdqsort`, update SQL parser, and fix out of bounds vector accesses

Greenscreen23:
I cannot compile the master branch anymore after a system upgrade due to the following errors:

```
third_party/sql-parser/src/util/sqlhelper.h error: unknown type name 'uintmax_t'
```

```
from /usr/include/boost/sort/spinsort/spinsort.hpp:17
/usr/include/c++/15.1.1/ciso646:46:4: error: <ciso646> is deprecated in C++17, use <version> to detect implementation-specific macros
```

The respective update is 
```
glibc (2.41+r47+g046b33800c3e -> 2.41+r48+g5cb575ca9a3d)
gcc-libs (14.2.1+r753+g1cd744a6828f -> 15.1.1+r7+gf36ec88aa85a)
gcc (14.2.1+r753+g1cd744a6828f -> 15.1.1+r7+gf36ec88aa85a)
```

There are also updates to these packages, but these seem to be unrelated (the error occurs at both boost versions)
```
libxml2 (2.13.8 -> 2.14.2)
boost-libs (1.87.0 -> 1.88.0)
boost (1.87.0 -> 1.88.0)
```

There are also updates to clang and llvm, but these did not change the version number. I suspect these are just rebuilds of the same code to match the package upgrade.

The first error was fixed by updating the sql-parser submodule to the newest commit. The second error could be fixed by including pdqsort instead of sort.hpp. We only use pdqsort, and thereby avoid including spinsort.


Greenscreen23:
I do compile with clang, but clang uses the gcc includes on linux, which are causing the error messages.

dey4ss:
Note that the parser introcuced foreign key constraints for `CREATE TABLE` statements in the meanwhile: https://github.com/hyrise/sql-parser/compare/57c763a99636c82f4ebc77439d762d39f263041b..main

We should at least add `Assert`s in the SQLTranslater if they are requested, or add them here as well.

Greenscreen23:
> Note that the parser introcuced foreign key constraints for `CREATE TABLE` statements in the meanwhile: https://github.com/hyrise/sql-parser/compare/57c763a99636c82f4ebc77439d762d39f263041b..main
> 
> We should at least add `Assert`s in the SQLTranslater if they are requested, or add them here as well.

Is there an issue with adding foreign key constraints to the table here?

https://github.com/hyrise/hyrise/blob/e9cbaab34bed765c2827c922c378a400b4f7c1c5/src/lib/sql/sql_translator.cpp#L1711-L1725

dey4ss:
> Is there an issue with adding foreign key constraints to the table here?

Currently, we would silently ignore them if they were set as column constraint, e.g. `CREATE TABLE foo (a int, b int REFERENCES bar)`:
https://github.com/hyrise/hyrise/blob/e9cbaab34bed765c2827c922c378a400b4f7c1c5/src/lib/sql/sql_translator.cpp#L1714-L1718

For table constraints, e.g., `CREATE TABLE foo (a int, b int, FOREIGN KEY (b) REFERENCES bar)`, we would at least throw an assertion:
https://github.com/hyrise/hyrise/blob/e9cbaab34bed765c2827c922c378a400b4f7c1c5/src/lib/sql/sql_translator.cpp#L1733-L1736

Your take if you want to add FKs in this PR or not.

Bouncner:
Discussion from 2025-05-14:
- add Assert for FKs

src/lib/sql/sql_translator.cpp
```diff
@@ -1712,10 +1712,9 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
       DebugAssert(parser_column_definition->column_constraints,
                   "Column " + column_definition.name + " is missing constraint information.");
       for (const auto& column_constraint : *parser_column_definition->column_constraints) {
-        if (column_constraint != hsql::ConstraintType::Unique &&
-            column_constraint != hsql::ConstraintType::PrimaryKey) {
-          continue;
-        }
+        Assert(
+            column_constraint == hsql::ConstraintType::Unique && column_constraint == hsql::ConstraintType::PrimaryKey,
+            "Only UNIQUE and PRIMARY KEY constraints are expected on a column level.");
```
dey4ss, line 1717: This lets the tests fail because explicitly requesting a column to contain `NULL` or `NOT NULL` is also saved as a constraint (this eases the parser to detect conflicting `PRIMARY KEY`/`NOT NULL` vs. `NULL` specifications).

Could you please add a test case (maybe in `SQLParserTest.InvalidConstraints`) that does sth like:
```sql
CREATE TABLE foo (a INT REFERENCES bar (x), b INT, FOREIGN KEY (b) REFERENCES bar (y));
```
and add a comment there that we currently do not support foreigns keys, and if we wanted to, we must make sure to store them as members of the `StaticTable` wrapped by the `StaticTableNode` that is the input of the `CreateTableNode`, and finally also add them to the newly created table in the `CreateTable` operator? This way, whoever thinks about adding foreign keys and relaxes the assert will be pointed to what to do.

dey4ss: **+** maybe think of a less cryptic PR name (e.g., more precise: "Use specific header for `boost::sort::pdqsort` and update SQL parser" or whatever you want).

Greenscreen23: Done. Sry for the early ping, I got distracted while waiting for the pipeline. It should be better now. I had to add some additional changes because the tests would not run on my local machine without them (gcc15 introduced bounds checks for `operator[]`, which we violate)

src/test/lib/storage/index/adaptive_radix_tree/adaptive_radix_tree_index_test.cpp
```diff
@@ -66,17 +66,17 @@ class AdaptiveRadixTreeIndexTest : public BaseTest {
...
-      const auto rbegin = distinct_values.rbegin();
+    const auto begin = distinct_values.begin();
+    const auto rbegin = distinct_values.rbegin();
 
-      Assert(*begin && *rbegin, "Optional has no value.");
-      int32_t min = **begin;
-      int32_t max = **rbegin;
+    Assert(*begin && *rbegin, "Optional has no value.");
+    int32_t min = **begin;
+    int32_t max = **rbegin;
```
dey4ss, line 76: auto?

src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp
```diff
@@ -20,10 +20,13 @@ FixedStringVector::FixedStringVector(const FixedStringVector& other, const Polym
...
 void FixedStringVector::push_back(const pmr_string& string) {
   Assert(string.size() <= _string_length, "Inserted string is too long to insert in FixedStringVector");
-  const auto pos = _chars.size();
-  // Default value of inserted elements using resize is null terminator ('\0')
-  _chars.resize(_chars.size() + _string_length);
-  string.copy(&_chars[pos], string.size());
+
+  if (_string_length > 0) {
+    const auto pos = _chars.size();
+    // Default value of inserted elements using resize is null terminator ('\0')
```
dey4ss, line 26:
```suggestion
    // Default value of inserted elements using resize is null terminator ('\0').
```

dey4ss (approved):
LG if you apply the suggestions + run format again.

src/lib/sql/sql_translator.cpp
```diff
@@ -1712,10 +1712,20 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
+        Assert(column_constraint == hsql::ConstraintType::Unique ||
+                   column_constraint == hsql::ConstraintType::PrimaryKey ||
+                   column_constraint == hsql::ConstraintType::NotNull ||
+                   column_constraint == hsql::ConstraintType::Null,
+               "Only UNIQUE, PRIMARY KEY, and (Not)Null constraints are expected on a column level.");
+
+        if (column_constraint == hsql::ConstraintType::NotNull || column_constraint == hsql::ConstraintType::Null) {
           continue;
         }
+
```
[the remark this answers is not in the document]
Greenscreen23: Done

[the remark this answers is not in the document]
Greenscreen23: I've also changed the second assert (for table constraints) to be an AssertInput, but that should not be a problem right?

Greenscreen23:
@Bouncner do you have something? Or is it fine if I merge this PR with one approval?

README.md
```diff
@@ -241,3 +241,4 @@ Contact: firstname.lastname@hpi.de
 -   Lukas     Wenzel
 -   Fabian    Wiebe
 -   Tim       Zimmermann
+-   Lukas     Hagen
```
Bouncner, line 244: This list is sorted alphabetically.

src/lib/sql/sql_translator.cpp
```diff
@@ -1712,10 +1712,16 @@ std::shared_ptr<AbstractLQPNode> SQLTranslator::_translate_create_table(const hs
...
                   "Column " + column_definition.name + " is missing constraint information.");
       for (const auto& column_constraint : *parser_column_definition->column_constraints) {
-        if (column_constraint != hsql::ConstraintType::Unique &&
-            column_constraint != hsql::ConstraintType::PrimaryKey) {
+        // (NOT) NULL was obtained above from `parser_column_definition->nullable`.
+        if (column_constraint == hsql::ConstraintType::NotNull || column_constraint == hsql::ConstraintType::Null) {
           continue;
         }
+
+        // We do not support e.g., FOREIGN KEYs yet.
```
Bouncner, line 1720:
```suggestion
        // We do not support, e.g., FOREIGN KEYs yet.
```

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -2638,6 +2638,14 @@ TEST_F(SQLTranslatorTest, InvalidConstraints) {
                InvalidInputException);
   // Constraints must not contain unknown columns.
   EXPECT_THROW(sql_to_lqp_helper("CREATE TABLE a_table (a_int INTEGER, UNIQUE(b_int))"), InvalidInputException);
+  // We currently do not support foreign keys. If we want to to so, we must make sure to store them as members of the
```
Bouncner, line 2641:
```suggestion
  // We currently do not support foreign keys. If we want to do so, we must ensure to store them as members of the
```

Bouncner (approved):
Tiny things.

README.md
Greenscreen23: oops, thanks for telling me.

PR #2701, opened 2025-05-14
Propagate INDs in the LQP

dey4ss:
This PR propagates INDs in the query plan. INDs are derived from ForeignKeyConstraints.


src/lib/logical_query_plan/abstract_non_query_node.hpp
```diff
@@ -10,21 +10,24 @@ namespace hyrise {
 /**
  * Base class for LQP nodes that do not query data (e.g, DML and DDL nodes) and therefore do not output columns.
  *
```
Bouncner: Is it a helper class or a base class?

src/test/lib/logical_query_plan/import_node_test.cpp
```diff
@@ -38,12 +38,10 @@ TEST_F(ImportNodeTest, Copy) {
   EXPECT_EQ(*_import_node, *_import_node->deep_copy());
 }
 
-TEST_F(ImportNodeTest, NoUniqueColumnCombinations) {
+TEST_F(ImportNodeTest, NoDataDependencies) {
```
Bouncner, line 41: Quite a lot of these tests.  `intersect_node_test.cpp` apparently missed one dependency. Given the number of dependencies, does it make sense to have a getter to obtain "all dependencies" or something else that avoids adding a single line every new dependency in almost all node tests?

src/lib/logical_query_plan/limit_node.cpp
```diff
@@ -31,6 +32,10 @@ OrderDependencies LimitNode::order_dependencies() const {
   return _forward_left_order_dependencies();
 }
 
+InclusionDependencies LimitNode::inclusion_dependencies() const {
+  return InclusionDependencies{};
```
Bouncner, line 37: Short comment?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -276,7 +281,7 @@ ColumnID AbstractLQPNode::get_column_id(const AbstractExpression& expression) co
 }
 
 bool AbstractLQPNode::has_output_expressions(const ExpressionUnorderedSet& expressions) const {
-  const auto& output_expressions = this->output_expressions();
+  const auto output_expressions = this->output_expressions();
```
Bouncner, line 286: Surprised to see that. Did we not discuss this recently because it was changed a couple of months ago already?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -292,7 +297,7 @@ bool AbstractLQPNode::has_matching_ucc(const ExpressionUnorderedSet& expressions
   DebugAssert(has_output_expressions(expressions),
               "The given expressions are not a subset of the LQP's output expressions.");
 
-  const auto& unique_column_combinations = this->unique_column_combinations();
+  const auto unique_column_combinations = this->unique_column_combinations();
```
Bouncner, line 302: Why no ref?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -310,7 +315,7 @@ bool AbstractLQPNode::has_matching_od(
   DebugAssert(has_output_expressions({ordered_expressions.cbegin(), ordered_expressions.cend()}),
               "The given ordered expressions are not a subset of the LQP's output expressions.");
 
-  const auto& order_dependencies = this->order_dependencies();
+  const auto order_dependencies = this->order_dependencies();
```
Bouncner, line 324: Why no ref?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -342,6 +347,87 @@ bool AbstractLQPNode::has_matching_od(
   return false;
 }
 
+bool AbstractLQPNode::has_matching_ind(const std::vector<std::shared_ptr<AbstractExpression>>& foreign_key_expressions,
+                                       const std::vector<std::shared_ptr<AbstractExpression>>& key_expressions) const {
+  const auto ind_size = key_expressions.size();
+  Assert(ind_size > 0, "Invalid input. Set of expressions should not be empty.");
+  DebugAssert(ind_size == foreign_key_expressions.size(), "Invlid IND requested.");
```
Bouncner: Invlid

Bouncner: Invalid IND passed? We don't really request an ind, right?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -342,6 +347,87 @@ bool AbstractLQPNode::has_matching_od(
...
+  }
+
+  // Translate required `foreign_key_expressions` to ColumnIDs of the actual stored table.
+  auto required_table = std::shared_ptr<Table>{};
+  auto required_column_id_for_expression = std::vector<std::pair<ColumnID, std::shared_ptr<AbstractExpression>>>{};
+  required_column_id_for_expression.reserve(ind_size);
+
+  for (auto expression_idx = ColumnID{0}; expression_idx < ind_size; ++expression_idx) {
+    const auto& expression = foreign_key_expressions[expression_idx];
+    DebugAssert(expression->type == ExpressionType::LQPColumn, "Expected column expression.");
```
Bouncner: Assert?
It's not a hot loop and a wrong cast can lead to very strange things.

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -342,6 +347,87 @@ bool AbstractLQPNode::has_matching_od(
...
+    } else {
+      // Columns must come from a single table.
+      Assert(required_table == table, "Invalid IND requested.");
+    }
+    required_column_id_for_expression.emplace_back(lqp_column_expression.original_column_id,
+                                                   key_expressions[expression_idx]);
+  }
+  // INDs can be equivalent, e.g., [A.a, A.b] in [B.x, B.y] is the same as [A.b, A.a] in [B.y, B.x]. Thus, we sort
+  // persisted ForeignKeyConstraints by the foreign key's ColumnIDs. We do the same here to compare with the valid INDs
+  // of the current node.
```
Bouncner, line 399: `we sort persisted ForeignKeyConstraints by the foreign key's ColumnIDs.` ... DebugAssert is_sorted?

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -232,6 +233,15 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
   bool has_matching_od(const std::vector<std::shared_ptr<AbstractExpression>>& ordering_expressions,
                        const std::vector<std::shared_ptr<AbstractExpression>>& ordered_expressions) const;
 
+  virtual InclusionDependencies inclusion_dependencies() const = 0;
+
+  /**
+   * @return True if there is an inclusion dependency (IND) matching the given list of output expressions and the given
```
Bouncner: Name which parameter is which list? The naming does not make it obvious.

Bouncner: A short example would be helpful.

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "expression/abstract_expression.hpp"
+#include "types.hpp"
+
+namespace hyrise {
+
+class Table;
+
+/**
+ * Container structure to define inclusion dependencies (INDs) for LQP nodes. An inclusion dependency on a given LQP
+ * node expresses that the resulting table's values for `expressions` (still) contain all values existing in the columns
```
Bouncner: I think a softer introduction to INDs would be helpful. At this point, I'd like to know what INDs are (one could google it, but 1-2 sentences would already help here). But instead I read `included_column_ids` without a clue what this means.

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
```diff
@@ -0,0 +1,55 @@
...
+ * specified by the `included_column_ids` of `included_table`. Inclusion dependencies can arise from foreign keys.
+ *
+ * Inclusion dependencies involve two relations. We only propagate them from the including side. For example, we
```
Bouncner: A bit more structure would be helpful. "Inclusion dependencies involve two relations" sounds like something that should be said earlier. "Inclusion dependencies can arise from foreign keys." is a bit lost where is it right now.

Clearly state (with examples) what the "including side" is. What is the connection to our FK naming scheme? What is the "life time" (lacking a better word right now) of SQL'ish FK syntax that defines FK constraints and LQPs that use INDs?
Might be something we should discuss on a whiteboard.

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
```diff
@@ -0,0 +1,55 @@
...
+ * consider the IND nation.n_regionkey in region.r_regionkey using a StoredTableNode for nation and a table for region.
+ * The IND of this looks like this:
+ *     InclusionDependency{{nation->get_column("n_regionkey")}, {region->get_column_id_by_name("r_regionkey")}, region}
+ *
+ * NOTE: Inclusion dependencies (INDs) are only valid for LQP nodes that contain no invalidated rows (i.e., where there
```
Bouncner: From what this class does, I don't get this comment. A short explanation would help.

Bouncner (commented):
I have a few questions but I think first making the explanation/terms clear in inclusion_dep.hpp would make most sense for me.

src/lib/logical_query_plan/abstract_non_query_node.hpp
dey4ss: `AbstractLQPNode` has no comment, `AbstractReadWriteOperator` and `AbstractReadOnlyOperator` say `superclass`. Anything you prefer (base/abstract/super...)? According to the naming scheme, it should probably not be "base".
https://github.com/hyrise/hyrise/blob/800a7e04117c32f34f6bae60353ecb05c3a9b968/src/lib/logical_query_plan/abstract_lqp_node.hpp#L60-L61
https://github.com/hyrise/hyrise/blob/800a7e04117c32f34f6bae60353ecb05c3a9b968/src/lib/operators/abstract_read_only_operator.hpp#L11
https://github.com/hyrise/hyrise/blob/800a7e04117c32f34f6bae60353ecb05c3a9b968/src/lib/operators/abstract_read_write_operator.hpp#L25

src/test/lib/logical_query_plan/import_node_test.cpp
dey4ss: There's currently no `AbstractDataDependency` that the individual dependnecies inherit from. Even if there was, we could not store them in a single data structure directly, but only via pointers (as we do it, e.g, for expressions). So I'm afraid that does not work that easy without changing a lot more code in this already huge PR.

dey4ss: Another option (for the future) would be to separate LQP nodes and dependencies again and have dependency propagation in own utility files.

src/lib/logical_query_plan/abstract_lqp_node.cpp
dey4ss: We did discuss this at some point. No using refs and let copy elision/RVO do the trick should be the right option here.

src/lib/logical_query_plan/abstract_lqp_node.cpp
dey4ss: (see above)

src/lib/logical_query_plan/abstract_lqp_node.cpp
dey4ss: (see above)

src/lib/logical_query_plan/abstract_lqp_node.cpp
dey4ss: Dou you just mean `requested` vs. `passed` here?

src/lib/logical_query_plan/abstract_lqp_node.hpp
dey4ss: Added an example and renamed the parameters

src/lib/logical_query_plan/abstract_non_query_node.hpp
dey4ss: (changed to "abstract super class", which we use quite often)

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
dey4ss: I re-wrote the comment and added an example

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
dey4ss: Basically only copypasta. I replaced the note with some sentences about dependencies becoming invalid, which I also copied to the other dependencies

src/lib/logical_query_plan/data_dependencies/functional_dependency.hpp
```diff
@@ -29,10 +29,14 @@ namespace hyrise {
  * Currently, the determinant expressions are required to be non-nullable to be involved in FDs. Combining null values
  * and FDs is not trivial. For more reference, see https://arxiv.org/abs/1404.4963.
  *
- * If the FD may become invalid in the future (because it is not based on a schema constraint, but on the data
- * incidentally fulfilling the constraint at the moment), the FD is marked as being not genuine.
- * This information is important because query plans that were optimized using a non-genuine FD are not safely
- * cacheable.
+ * NOTE: Because functional dependencies can be derived from soft key constraints, which are not verified to be valid
+ *       (especially for data changes), we cannot really safely assume that all FDs are valid. Handling that is future
```
Bouncner, line 33:
```suggestion
 *       (especially for data changes), we cannot safely assume that all FDs are valid. Handling that is future
```

src/lib/logical_query_plan/data_dependencies/functional_dependency.hpp
```diff
@@ -29,10 +29,14 @@ namespace hyrise {
...
+ *       work.
+ *       For FDs derived from discovered unique constraints, i.e., not defined by the DDL, we currently have means to
```
Bouncner, line 35: Reads a bit confusing to me. "we cannot really safely assume that all FDs are valid". Okay, so if we do not assume that, we will never cache them. Right?
But then, we say we do not cache non-genuine ones. Which means we cache genuine ones? But we cannot assume they are valid.

Can we split this added comment part here up? Maybe first talk about made assumptions when they hold. Then optimizer things (one sentence about this "we currently have means to guarantee correct query plans"; why not for genuine ones?) and caching.

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
```diff
@@ -0,0 +1,80 @@
...
+ *              3 |    CANADA |           1                          3 |      EUROPE
+ *              4 |     EGYPT |           4                          4 | MIDDLE EAST
+ *              5 |  ETHIOPIA |           0
+ *              6 |    FRANCE |           3
+ *              7 |   GERMANY |           3
+ *              8 |     INDIA |           2
+ *              9 | INDONESIA |           2
+ *             10 |      IRAN |           4
+ *
+ * We propagate INDs in LQPs for the referenced (primary key) side. Because an IND is (probably) ivalid whenever we
```
Bouncner, line 39:
```suggestion
 * We propagate INDs in LQPs for the referenced (primary key) side. Because an IND is (probably) invalid whenever we
```

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * remove a single tuple from the referenced side, we can thus reduce the recursion depth when fetching INDs because
```
Bouncner, line 40: Ufff ... that comment is too hard for me. Can you write a bit slower? :D

Bouncner: `JoinNode` MIGHT remove so we take the safe route and assume the IND is no longer given? Do I understand this correct?

Bouncner: Still, this does not really explain to me why we do not forward INDs for the including side, which is the beginning of this paragraph.

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * any LQP node that does not return all input tuples (PredicateNode, JoinNode, ...) can simply return an empty set of
+ * INDs in most cases.
+ *
+ * When retrieving an IND from a ForeignKeyConstraint, we do not know the StoredTableNode (and, thus, expressions) of
+ * the included side. For that reason, we store the table and ColumnIDs of the included side. If we want to check if an
+ * IND holds for a given node, we can then check if (i) the IND is propagated and (ii) all (LQP column) expressions for
+ * the included side are present. An IND for the example above looks like this:
+ *     InclusionDependency{{nation->get_column("n_regionkey")}, {region->get_column_id_by_name("r_regionkey")}, region};
```
Bouncner, line 48: Can you state above that `IND {nation} -> {region}` means `{referenced} -> {including}`?

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
```diff
@@ -0,0 +1,80 @@
...
+ *              8 |     INDIA |           2
+ *              9 | INDONESIA |           2
+ *             10 |      IRAN |           4
+ *
+ * We propagate INDs in LQPs for the referenced (primary key) side. Because an IND is (probably) ivalid whenever we
+ * remove a single tuple from the referenced side, we can thus reduce the recursion depth when fetching INDs because
+ * any LQP node that does not return all input tuples (PredicateNode, JoinNode, ...) can simply return an empty set of
+ * INDs in most cases.
+ *
+ * When retrieving an IND from a ForeignKeyConstraint, we do not know the StoredTableNode (and, thus, expressions) of
```
Bouncner, line 44: Short introduction sentence what you are talking about here? Why does it matter that we do not have the stored table node?

"For that reason, we store ..." ... where is we. The IND struct here?

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include "expression/abstract_expression.hpp"
+#include "types.hpp"
+
+namespace hyrise {
+
+class Table;
+
+/**
+ * Container structure to define inclusion dependencies (INDs) for LQP nodes. An inclusion dependency states that all
+ * values of a table's column (or value combinations of multiple columns) are also present in another column(s), usually
```
Bouncner, line 17: Q for me: Is something like `{A} -> {B,C}` possible? Basically B and C each contain all values of A?
Or do left and right lists have to have the same size?

I guess order matters for multi-column, so we are having lists, not sets?

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * any LQP node that does not return all input tuples (PredicateNode, JoinNode, ...) can simply return an empty set of
+ * INDs in most cases.
+ *
+ * When retrieving an IND from a ForeignKeyConstraint, we do not know the StoredTableNode (and, thus, expressions) of
+ * the included side. For that reason, we store the table and ColumnIDs of the included side. If we want to check if an
+ * IND holds for a given node, we can then check if (i) the IND is propagated and (ii) all (LQP column) expressions for
+ * the included side are present. An IND for the example above looks like this:
+ *     InclusionDependency{{nation->get_column("n_regionkey")}, {region->get_column_id_by_name("r_regionkey")}, region};
+ *
+ * NOTE: Because INDs are derived from soft constraints, which are not verified to be valid (especially for data
```
Bouncner, line 50: Repetition to above. Merge?

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.hpp
```diff
@@ -0,0 +1,80 @@
...
+ *       changes), we cannot really safely assume that INDs are valid. Handling that is future work.
+ */
+struct InclusionDependency final {
+  explicit InclusionDependency(std::vector<std::shared_ptr<AbstractExpression>>&& init_referenced_expressions,
```
Bouncner, line 54: Haven't read the cpp but that might be a helpful comment here: why do we have expressions on the one side and column ids on the other?
For order deps is both expressions.

src/lib/logical_query_plan/data_dependencies/inclusion_dependency.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+InclusionDependency::InclusionDependency(std::vector<std::shared_ptr<AbstractExpression>>&& init_referenced_expressions,
+                                         std::vector<ColumnID>&& init_included_column_ids,
+                                         const std::shared_ptr<Table>& init_included_table)
+    : referenced_expressions{std::move(init_referenced_expressions)},
+      included_column_ids{std::move(init_included_column_ids)},
+      included_table{init_included_table} {
+  Assert(!referenced_expressions.empty(), "InclusionDependency cannot be empty.");
+  Assert(referenced_expressions.size() == included_column_ids.size(),
+         "InclusionDependency expects same amount of included and referenced columns.");
```
Bouncner, line 25:
```suggestion
         "InclusionDependency expects same number of included and referenced columns.");
```

src/lib/logical_query_plan/data_dependencies/unique_column_combination.hpp
```diff
@@ -12,13 +12,14 @@ namespace hyrise {
...
  * PRIMARY KEY semantics, check if the expressions are nullable, cf. AbstractLQPNode::is_column_nullable.
  *
- * NOTE: Unique column combinations (UCCs) are only valid for LQP nodes that contain no invalidated rows (i.e., where
- *       there has been a ValidateNode before or where MVCC is disabled).
- *
- * If a UCC may become invalid in the future (because it is not based on a schema constraint, but on the data
- * incidentally being unique at the moment), the UCC is marked as being not genuine.
- * This information is important because query plans that were optimized using a non-genuine UCC are not safely
- * cacheable.
+ * NOTE: Because unique column combinations (UCCs) are derived from soft constraints, which are not verified to be valid
```
Bouncner, line 15: Pretty much the same as above. Can we discuss it once and reference it here?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -348,6 +353,89 @@ bool AbstractLQPNode::has_matching_od(
   return false;
 }
 
+bool AbstractLQPNode::has_matching_ind(
+    const std::vector<std::shared_ptr<AbstractExpression>>& included_expressions,
+    const std::vector<std::shared_ptr<AbstractExpression>>& referenced_expressions) const {
+  const auto ind_size = referenced_expressions.size();
+  Assert(ind_size > 0, "Invalid input. Set of expressions should not be empty.");
+  DebugAssert(ind_size == included_expressions.size(), "IND is invalid.");
```
Bouncner, line 361: Assert instead of DebugAssert?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -348,6 +353,89 @@ bool AbstractLQPNode::has_matching_od(
...
+  DebugAssert(has_output_expressions({referenced_expressions.cbegin(), referenced_expressions.cend()}),
```
Bouncner, line 362: Not this PR:

I think the name `has_output_expressions` should be rethought. How does "has" mean here? It's about a subset, which is not said by the function name.

Also the function should say "non-empty subset".

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -234,6 +235,22 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
   bool has_matching_od(const std::vector<std::shared_ptr<AbstractExpression>>& ordering_expressions,
                        const std::vector<std::shared_ptr<AbstractExpression>>& ordered_expressions) const;
 
+  virtual InclusionDependencies inclusion_dependencies() const = 0;
+
+  /**
```
Bouncner, line 240: How useful are INDs within the same relation?

Just curious.

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -234,6 +235,22 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
...
+   * @return True if there is an inclusion dependency (IND) matching the given lists of expressions. That means this
```
Bouncner, line 241: Now that we have soft contraints and non-geniune INDs, I guess the comment should be adapted?

src/lib/logical_query_plan/abstract_lqp_node.hpp
```diff
@@ -234,6 +235,22 @@ class AbstractLQPNode : public std::enable_shared_from_this<AbstractLQPNode> {
...
+   *         node provides the referenced expressions, no tuples were removed in this (sub-)plan, and there exists a
+   *         foreign key relationship on a table level where the passed referenced expressions are on the FK side and
+   *         the included expressions are on the PK side.
+   *
+   * Example: - Assume two StoredTableNodes nation and region, "n_regionkey" is a foreign key to "r_regionkey".
```
Bouncner, line 246: I think an even more helpful example would be some query that actually uses an IND. Is that possible in 1-2 sentences?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -348,6 +353,89 @@ bool AbstractLQPNode::has_matching_od(
...
+    return lhs.first < rhs.first;
+  });
+
+  // Look for an inclusion dependency that is based on the given expressions.
+  for (const auto& ind : inclusion_dependencies) {
+    if (required_table != ind.included_table) {
+      continue;
+    }
+    // IND references the same included table. Now, check that it also has the required columns in the correct order.
+    DebugAssert(std::ranges::is_sorted(ind.included_column_ids), "Included columns should be ordered.");
```
Bouncner, line 410: I am fine with the assertion, but that should be always given as `ind` instances should always have sorted column IDs, right? At least when they are properly constructed (should we delete the default constructor?)

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -348,6 +353,89 @@ bool AbstractLQPNode::has_matching_od(
...
+    auto required_columns_it = required_column_id_for_expression.cbegin();
+    const auto found_ind_size = ind.referenced_expressions.size();
+    const auto required_columns_end = required_column_id_for_expression.cend();
+    for (auto expression_idx = ColumnID{0};
+         expression_idx < found_ind_size && required_columns_it != required_columns_end; ++expression_idx) {
+      // Did not reach current required ColumnID yet.
+      if (required_columns_it->first > ind.included_column_ids[expression_idx]) {
```
Bouncner, line 417: Can you briefly add a comment that states why we are not simply matching? So there could be a lower column ID being checked (this the optimizer assume there is this IND) but that one might not be stored in the actual IND. Is that somewhat right?

What are cases where this happens? I am curious how that looks when being used.

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -348,6 +353,89 @@ bool AbstractLQPNode::has_matching_od(
   return false;
 }
 
+bool AbstractLQPNode::has_matching_ind(
```
Bouncner, line 356: Note to  myself: check usage.

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -483,12 +571,13 @@ void AbstractLQPNode::_add_output_pointer(const std::shared_ptr<AbstractLQPNode>
...
   Assert(left_input(), "Cannot forward unique column combinations without an input node.");
-  const auto& input_unique_column_combinations = left_input()->unique_column_combinations();
+  const auto input_unique_column_combinations = left_input()->unique_column_combinations();
 
   if constexpr (HYRISE_DEBUG) {
     // Check whether output expressions are missing.
+    const auto output_expressions = this->output_expressions();
     for (const auto& ucc : input_unique_column_combinations) {
-      Assert(has_output_expressions(ucc.expressions),
+      Assert(contains_all_expressions(ucc.expressions, output_expressions),
```
Bouncner, line 580: Why this change?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -511,6 +600,21 @@ OrderDependencies AbstractLQPNode::_forward_left_order_dependencies() const {
...
 
+InclusionDependencies AbstractLQPNode::_forward_left_inclusion_dependencies() const {
+  Assert(left_input(), "Cannot forward inclusion dependencies without an input node.");
+  const auto input_inclusion_dependencies = left_input()->inclusion_dependencies();
+
+  if constexpr (HYRISE_DEBUG) {
+    // Check whether output expressions are missing.
+    const auto output_expressions = this->output_expressions();
+    for (const auto& ind : input_inclusion_dependencies) {
+      Assert(contains_all_expressions(ind.referenced_expressions, output_expressions),
```
Bouncner, line 611: Again using `contains_all...` instead of `has_...`. Why?

src/lib/logical_query_plan/abstract_lqp_node.cpp
```diff
@@ -511,6 +600,21 @@ OrderDependencies AbstractLQPNode::_forward_left_order_dependencies() const {
   return input_order_dependencies;
 }
 
+InclusionDependencies AbstractLQPNode::_forward_left_inclusion_dependencies() const {
```
Bouncner, line 603: Probably a stupid question, but we do we only check the referenced columns, not the including ones?

src/lib/logical_query_plan/mock_node.cpp
```diff
@@ -171,6 +173,28 @@ OrderDependencies MockNode::order_dependencies() const {
...
+void MockNode::set_foreign_key_constraints(const ForeignKeyConstraints& foreign_key_constraints) {
+  _foreign_key_constraints = foreign_key_constraints;
+}
+
+InclusionDependencies MockNode::inclusion_dependencies() const {
+  auto inclusion_dependencies = InclusionDependencies{};
+  for (const auto& foreign_key_constraint : _foreign_key_constraints) {
+    const auto& table = foreign_key_constraint.foreign_key_table();
+    const auto& columns = foreign_key_constraint.primary_key_columns();
+    // As we only use this node in tests, we ignore that INDs with pruned columns could be adjusted. Fir instance,
```
Bouncner, line 185: Fir

src/lib/logical_query_plan/union_node.cpp
```diff
@@ -91,6 +92,26 @@ OrderDependencies UnionNode::order_dependencies() const {
...
+      return left_inclusion_dependencies;
+    }
+    case SetOperationMode::All: {
+      auto left_inclusion_dependencies = left_input()->inclusion_dependencies();
+      const auto right_inclusion_dependencies = right_input()->inclusion_dependencies();
+      left_inclusion_dependencies.insert(right_inclusion_dependencies.cbegin(), right_inclusion_dependencies.cend());
+      return left_inclusion_dependencies;
+    }
+    case SetOperationMode::Unique:
+      Fail("ToDo, see discussion https://github.com/hyrise/hyrise/pull/2156#discussion_r452803825");
```
Bouncner, line 110: Rather have a proper message instead of "Todo" and link the comment as a code comment?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -204,6 +205,113 @@ OrderDependencies JoinNode::order_dependencies() const {
...
+
+InclusionDependencies JoinNode::_output_inclusion_dependencies(
+    const InclusionDependencies& left_inclusion_dependencies,
+    const InclusionDependencies& right_inclusion_dependencies) const {
+  // Check if there are any INDs that might be forwarded.
+  if (left_inclusion_dependencies.empty() && (join_mode == JoinMode::Semi || right_inclusion_dependencies.empty())) {
+    return InclusionDependencies{};
+  }
+
+  // Left / right outer joins foward all tuples of the respective input. Its INDs remain valid.
```
Bouncner, line 251: foward

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -204,6 +205,113 @@ OrderDependencies JoinNode::order_dependencies() const {
...
+  const auto& join_predicates = this->join_predicates();
+  auto left_input_join_predicates = std::vector<std::shared_ptr<AbstractExpression>>{};
+  auto right_input_join_predicates = std::vector<std::shared_ptr<AbstractExpression>>{};
+  const auto left_expressions = left_input()->output_expressions();
+
+  for (const auto& expression : join_predicates) {
+    const auto& predicate = std::dynamic_pointer_cast<BinaryPredicateExpression>(expression);
+    if (!predicate || predicate->predicate_condition != PredicateCondition::Equals) {
+      if (join_mode == JoinMode::Left) {
+        return left_inclusion_dependencies;
```
Bouncner, line 269: Do I need lines 252 to 257 if I am going to return all left INDs here anyways?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -204,6 +205,113 @@ OrderDependencies JoinNode::order_dependencies() const {
...
+      }
+
+      if (join_mode == JoinMode::Right) {
+        return right_inclusion_dependencies;
+      }
+
+      return InclusionDependencies{};
+    }
+
+    if (find_expression_idx(*predicate->left_operand(), left_expressions)) {
```
Bouncner, line 279: What does this if do? `find...` sounds like an implicit int to bool conversion. Can we make that clearer?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -204,6 +205,113 @@ OrderDependencies JoinNode::order_dependencies() const {
...
+      DebugAssert(find_expression_idx(*predicate->right_operand(), right_input()->output_expressions()),
+                  "Expected to resolve right operand.");
+      left_input_join_predicates.emplace_back(predicate->left_operand());
+      right_input_join_predicates.emplace_back(predicate->right_operand());
+    } else {
+      DebugAssert(find_expression_idx(*predicate->left_operand(), right_input()->output_expressions()),
```
Bouncner, line 285: A comment about predicate switching would be helpful here.

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -204,6 +205,113 @@ OrderDependencies JoinNode::order_dependencies() const {
...
+                  "Expected to resolve left operand.");
+      DebugAssert(find_expression_idx(*predicate->right_operand(), left_expressions),
+                  "Expected to resolve right operand.");
+      left_input_join_predicates.emplace_back(predicate->right_operand());
```
Bouncner, line 289: What is the purpose of collecting all the predicates?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -204,6 +205,113 @@ OrderDependencies JoinNode::order_dependencies() const {
...
+    inclusion_dependencies.insert(right_inclusion_dependencies.cbegin(), right_inclusion_dependencies.cend());
+  }
+
+  // Check that all join predicates are equals_() predicates and map the join keys to the input nodes.
+  const auto& join_predicates = this->join_predicates();
+  auto left_input_join_predicates = std::vector<std::shared_ptr<AbstractExpression>>{};
+  auto right_input_join_predicates = std::vector<std::shared_ptr<AbstractExpression>>{};
+  const auto left_expressions = left_input()->output_expressions();
+
+  for (const auto& expression : join_predicates) {
```
Bouncner, line 265: Pretty long loop. Can you give an overview comment? What are short cuts (outer), why do we collect predicates, and when can we forward something.

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -204,6 +205,113 @@ OrderDependencies JoinNode::order_dependencies() const {
...
+      right_input_join_predicates.emplace_back(predicate->left_operand());
+    }
+  }
+
+  const auto predicate_count = join_predicates.size();
+  Assert(left_input_join_predicates.size() == predicate_count && right_input_join_predicates.size() == predicate_count,
+         "Could not resolve all join predicates.");
+
+  // Add INDs from left input if left join keys are contained in right join keys.
+  if (right_input()->has_matching_ind(left_input_join_predicates, right_input_join_predicates)) {
```
Bouncner, line 299: This is a nice example of ind usage. High level description near the implementation of `has_matching_ind` would be helpful.

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -204,6 +205,113 @@ OrderDependencies JoinNode::order_dependencies() const {
...
+                  "Expected to resolve right operand.");
+      left_input_join_predicates.emplace_back(predicate->left_operand());
+      right_input_join_predicates.emplace_back(predicate->right_operand());
+    } else {
+      DebugAssert(find_expression_idx(*predicate->left_operand(), right_input()->output_expressions()),
+                  "Expected to resolve left operand.");
+      DebugAssert(find_expression_idx(*predicate->right_operand(), left_expressions),
+                  "Expected to resolve right operand.");
+      left_input_join_predicates.emplace_back(predicate->right_operand());
+      right_input_join_predicates.emplace_back(predicate->left_operand());
```
Bouncner, line 290: Curious: could this ever fail when we have a constant as an operand? Or something that is not an LQPColumnE?

src/lib/logical_query_plan/join_node.cpp
```diff
@@ -204,6 +205,113 @@ OrderDependencies JoinNode::order_dependencies() const {
...
+  Assert(left_input_join_predicates.size() == predicate_count && right_input_join_predicates.size() == predicate_count,
+         "Could not resolve all join predicates.");
+
+  // Add INDs from left input if left join keys are contained in right join keys.
+  if (right_input()->has_matching_ind(left_input_join_predicates, right_input_join_predicates)) {
+    inclusion_dependencies.insert(left_inclusion_dependencies.cbegin(), left_inclusion_dependencies.cend());
+  }
+
+  if (join_mode == JoinMode::Semi) {
+    return inclusion_dependencies;
```
Bouncner, line 304: I am not sure I fully understand this part. Let's say we have an equal attribute so we end up here. Left remaining values are included in right side, yes. But the right side is discarded. And the IND is defined on the physical table. Why is this IND still valid?

src/lib/logical_query_plan/projection_node.cpp
```diff
@@ -82,6 +79,22 @@ OrderDependencies ProjectionNode::order_dependencies() const {
   return order_dependencies;
 }
 
+InclusionDependencies ProjectionNode::inclusion_dependencies() const {
+  auto inclusion_dependencies = InclusionDependencies{};
+  inclusion_dependencies.reserve(node_expressions.size());
+
+  // Forward inclusion dependencies, if applicable.
```
Bouncner, line 86: This comment does not really help. Remove or ", if applicable" >> " if all IND ref cols are oupput"?

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -187,6 +188,58 @@ OrderDependencies StoredTableNode::order_dependencies() const {
...
+    if (!referenced_table) {
+      // Referenced table was deleted, IND is useless.
+      continue;
+    }
+
+    // Remove pruned columns from the IND.
+    auto pruned_columns_it = _pruned_column_ids.cbegin();
+    auto pk_column_ids = std::vector<ColumnID>{};
+    auto fk_column_ids = std::vector<ColumnID>{};
+    for (auto pk_columns_it = foreign_key_constraint.primary_key_columns().cbegin(),
```
Bouncner, line 209: Wondering if we could do something simpler using a vector difference and conditional copying.

src/lib/logical_query_plan/stored_table_node.cpp
```diff
@@ -187,6 +188,58 @@ OrderDependencies StoredTableNode::order_dependencies() const {
...
+      ++pruned_columns_it;
+    }
+
+    DebugAssert(pk_column_ids.size() == fk_column_ids.size(), "IND's ColumnIDs were pruned incorrectly.");
+    if (pk_column_ids.empty()) {
+      continue;
+    }
+
+    // Search for expressions representing the inclusion constraint's ColumnIDs.
+    auto column_expressions = get_expressions_for_column_ids(*this, pk_column_ids);
```
Bouncner, line 234: Could this ever fail? Cannot think of something (unless concurrently we delete columns) that might be relevant, but want to be sure we don't oversee something.

src/test/lib/storage/table_test.cpp
```diff
@@ -288,7 +288,9 @@ TEST_F(StorageTableTest, StableChunks) {
   // The vector should have been resized / expanded by now
 
   EXPECT_EQ(first_chunk, &chunks_vector[0]);
-  EXPECT_EQ((*(*first_chunk)->get_segment(ColumnID{0}))[ChunkOffset{0}], AllTypeVariant{100});
+  EXPECT_EQ((*(*first_chunk)->get_segment(ColumnID{0})) [ChunkOffset{0}], AllTypeVariant {
```
Bouncner, line 291: Igitt.

Bouncner: Can we just change the line? Extract the chunk to a var or so, just to avoid this repeated uglyness.

Or fix this formatting, but I think we weren't sure where it came from, right?

src/test/lib/logical_query_plan/window_node_test.cpp
```diff
@@ -141,9 +141,24 @@ TEST_F(WindowNodeTest, ForwardOrderDependencies) {
...
+  EXPECT_TRUE(_mock_node->inclusion_dependencies().empty());
+  EXPECT_TRUE(_window_node->inclusion_dependencies().empty());
+
+  const auto dummy_table = Table::create_dummy_table({{"a", DataType::Int, false}});
+  const auto ind = InclusionDependency{{_a}, {ColumnID{0}}, dummy_table};
+  const auto foreign_key_constraint = ForeignKeyConstraint{{ColumnID{0}}, dummy_table, {ColumnID{0}}, nullptr};
+  _mock_node->set_foreign_key_constraints({foreign_key_constraint});
+  EXPECT_EQ(_mock_node->inclusion_dependencies().size(), 1);
+
+  const auto inclusion_dependencies = _window_node->inclusion_dependencies();
```
Bouncner, line 159: Are windows that simple? Surprised as I am very scared of windows due to all the crazy stuff you can do with them.

src/lib/logical_query_plan/abstract_lqp_node.cpp
dey4ss: To avoid computing the output expressions multiple times.

src/lib/logical_query_plan/abstract_lqp_node.cpp
dey4ss: (see above)

src/test/lib/storage/table_test.cpp
dey4ss: Gone by master merge, where it's wrapped by `clang-format off`. This cannot be a finite solution, though. It must be a somewhat recent clang-format version that does this.

PR #2702, opened 2025-05-16
Reduce CI pipeline runtime

dey4ss:
Our CI pipeline now runs around 3 to 3.5 hours, this PR tries to push it more to ~2.5 h.
Decrease the CI pipeline runtime by:
- Combining tests for row counts and PKs/FKs of TPCDSTableGenerator
- Omitting SQLite tests with encoded data in shuffled system tests
- Giving more resources to clang release build with TSan
- Skipping SQLite tests in valgrind stage
- Reducing the load of some stress tests (at least in sanitizer builds), skip tests for unencoded data in _encoded_ SQLite tests, cache SQLite results there

If we want to further decrease the runtime, the following ideas might be taken into account:
- skip release build on x64 Mac
- run TPC-DS verification in parallel to the pipeline, skip Nix release build 
- further look into decreasing system test runtime, e.g., TPCC tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -66,350 +66,350 @@ try {
     //   def hyriseCI = docker.image('hyrise/hyrise-ci:24.04');
     //   hyriseCI.pull()
 
-      // LSAN (executed as part of ASAN) requires elevated privileges. Therefore, we had to add --cap-add SYS_PTRACE.
-      // Even if the CI run sometimes succeeds without SYS_PTRACE, you should not remove it until you know what you are
-      // doing. See also: https://github.com/google/sanitizers/issues/764
-      hyriseCI.inside("--cap-add SYS_PTRACE -u 0:0") {
-        try {
-          stage("Setup") {
-            checkout scm
+    //   // LSAN (executed as part of ASAN) requires elevated privileges. Therefore, we had to add --cap-add SYS_PTRACE.
+    //   // Even if the CI run sometimes succeeds without SYS_PTRACE, you should not remove it until you know what you are
+    //   // doing. See also: https://github.com/google/sanitizers/issues/764
+    //   hyriseCI.inside("--cap-add SYS_PTRACE -u 0:0") {
+    //     try {
+    //       stage("Setup") {
+    //         checkout scm
 
-            // Check if ninja-build is installed. As make is sufficient to work with Hyrise, ninja-build is not
-            // installed via install_dependencies.sh but is part of the hyrise-ci docker image.
-            sh "ninja --version > /dev/null"
+    //         // Check if ninja-build is installed. As make is sufficient to work with Hyrise, ninja-build is not
+    //         // installed via install_dependencies.sh but is part of the hyrise-ci docker image.
+    //         sh "ninja --version > /dev/null"
 
-            // During CI runs, the user is different from the owner of the directories, which blocks the execution of
-            // git commands since the fix of the git vulnerability CVE-2022-24765. git commands can then only be
-            // executed if the corresponding directories are added as safe directories.
-            sh '''
-            git config --global --add safe.directory $WORKSPACE
-            # Get the paths of the submodules; for each path, add it as a git safe.directory
-            grep path .gitmodules | sed 's/.*=//' | xargs -n 1 -I '{}' git config --global --add safe.directory $WORKSPACE/'{}'
-            '''
+    //         // During CI runs, the user is different from the owner of the directories, which blocks the execution of
+    //         // git commands since the fix of the git vulnerability CVE-2022-24765. git commands can then only be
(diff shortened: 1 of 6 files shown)


Jenkinsfile
```diff
@@ -222,7 +222,7 @@ try {
             stage("clang-debug-unity-odr") {
               if (env.BRANCH_NAME == 'master' || full_ci) {
                 // Check if unity builds work even if everything is batched into a single compilation unit. This helps prevent ODR (one definition rule) issues.
-                sh "cd clang-debug-unity-odr && ninja all -j \$(( \$(nproc) / 20))"
+                sh "cd clang-debug-unity-odr && ninja all -j 2"
```
Bouncner, line 225: Can this lead to issues when we use a smaller VM again as we did a year ago? Let's say one with 8 cores?

src/test/lib/concurrency/stress_test.cpp
```diff
@@ -32,7 +32,7 @@ class StressTest : public BaseTest {
 
   static constexpr auto DEFAULT_LOAD_FACTOR = uint32_t{10};
 
-  const uint32_t CPU_COUNT = std::thread::hardware_concurrency();
+  const uint32_t CPU_COUNT = std::min(std::thread::hardware_concurrency(), uint32_t{32});
```
Bouncner, line 35: auto?

src/test/lib/sql/sqlite_testrunner/sqlite_testrunner.cpp
```diff
@@ -2,7 +2,7 @@
 
 namespace hyrise {
 
-void SQLiteTestRunner::SetUpTestCase() {
+void SQLiteTestRunner::SetUpTestSuite() {
```
Bouncner, line 5: SetUpTestSuite runs one, SetUpTestCase runs once. Are we using a "test suite" here? Is that a thing?

src/test/lib/sql/sqlite_testrunner/sqlite_testrunner_encodings.cpp
```diff
@@ -3,9 +3,27 @@
 
 namespace hyrise {
 
+namespace {
+
+std::vector<EncodingType> encoding_types_without_unencoded() {
+  static auto encodings = std::vector<EncodingType>{};
+
+  if (encodings.empty()) {
```
Bouncner, line 11: When is `encodings` not empty?

src/test/lib/sql/sqlite_testrunner/sqlite_testrunner.cpp
dey4ss: It was just renamed, `SetUpTestSuite` is the new name: https://google.github.io/googletest/reference/testing.html

src/test/lib/sql/sqlite_testrunner/sqlite_testrunner_encodings.cpp
dey4ss: On the first call I guess? I was not entirely sure is the test params are called only once or for each test case, and I saw we have a similar pattern for that elsewhere in the code.

Jenkinsfile
dey4ss: `$(nproc) / 20` would be `0` then, but `2` stays the same. So I guess it would be an improvement in that case?

Bouncner: `($(nproc) / 20) + 1`? I don't think it matters much though.

src/test/lib/concurrency/stress_test.cpp
dey4ss: this is a non-constexpr class member

Jenkinsfile
dey4ss: That still means we allocate a fair amount of parallel workers for a compilation task that does not need many resources on our large CI servers.

dey4ss: I don't really get what's the issue with using a small, fixed number of compilation threads

Bouncner: Well, it's fixed. If we use a larger machine, we use not enough cores. If we use a smaller machine, we use too many cores.

dey4ss: I mean, we'll always be upper-bounded by the (system) tests (esp. in sanitizer builds etc). Not really a chance that this build would be slower than that even when assigning olny a single core. Saving some cores here enables us to give more resources to the costlier ones on big machines. I don't really buy that 2 is "too many cores" on smaller machines when literally 1 is the lowest we can go.

dey4ss: if you insist, I'll change it back though

Bouncner (approved):
Not that I fully agree with the thread counts, but apart from that: feel free to merge.

PR #2707, opened 2025-06-02
[DYOD] Group 1: increase test coverage

luisegarberding:
We improved the coverage of:

1. `variable_length_key`
2.  `variable_length_key_store`
3.  `fixed_string_vector`

**Coverage before changes:**
image
pre
image

**Coverage after:**
image
post
image

### DIFF
--- a/src/test/lib/storage/index/group_key/variable_length_key_test.cpp
+++ b/src/test/lib/storage/index/group_key/variable_length_key_test.cpp
@@ -123,7 +123,7 @@ TEST_F(VariableLengthKeyTest, OStreamOperator) {
   auto oss = std::ostringstream{};
   oss << key;
   const auto str = oss.str();
-  EXPECT_FALSE(str.empty());
+  EXPECT_EQ(str, "ab cd");
 }
 
 }  // namespace hyrise


src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
```diff
@@ -84,14 +84,28 @@ TEST_F(FixedStringVectorTest, IteratorConst) {
...
     EXPECT_EQ(fixed_string.size(), 3u);
   }
 }
 
+TEST_F(FixedStringVectorTest, ReverseIteratorsExplicit) {
+  auto v = std::vector<pmr_string>{"abc", "def", "ghi"};
+  auto fs_vector = FixedStringVector{v.cbegin(), v.cend(), 3};
+
+  auto rbegin = fs_vector.rbegin();
+  auto rend = fs_vector.rend();
```
dey4ss: You could make these const or just pull (at least the `rbegin`) in the loop below.

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
```diff
@@ -84,14 +84,28 @@ TEST_F(FixedStringVectorTest, IteratorConst) {
...
+
+  auto expected = std::vector<pmr_string>{"ghi", "def", "abc"};
+  auto i = 0;
+  for (auto it = rbegin; it != rend; ++it, ++i) {
+    EXPECT_EQ(*it, expected[i]);
+  }
```
dey4ss: Could you also use an iterator on `expected` just alike? Then, you could get rid of the index variable `i` and have sth. like: 
```suggestion
  for (auto reverse_it = fs_vector.rbegin(), expected_it = expected.begin(); reverse_it != rend; ++reverse_it, ++expected_it) {
    EXPECT_EQ(*reverse_it, *expected_it);
  }
```

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
```diff
@@ -106,6 +120,16 @@ TEST_F(FixedStringVectorTest, Erase) {
   EXPECT_EQ((*fixed_string_vector)[0u], "foo");
 }
 
+TEST_F(FixedStringVectorTest, EraseWithZeroStringLength) {
+  auto strings = std::vector<pmr_string>{"", ""};
+  auto vec = FixedStringVector(strings.begin(), strings.end(), 0u);
```
dey4ss: If you can, get rid of the unsigned literal
```suggestion
  auto vec = FixedStringVector(strings.begin(), strings.end(), 0);
```

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
```diff
@@ -106,6 +120,16 @@ TEST_F(FixedStringVectorTest, Erase) {
...
+
+  auto it = vec.begin();
+  vec.erase(it, vec.end());
```
dey4ss: There's no need for the extra variable, is it?

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
```diff
@@ -106,6 +120,16 @@ TEST_F(FixedStringVectorTest, Erase) {
...
+
+  EXPECT_EQ(vec.size(), 2u);
```
dey4ss:
```suggestion
  EXPECT_EQ(vec.size(), 2);
```

dey4ss: Totally not yours, I think you found a bug here. Erasing the entire contents and still preserving the number of elements is a bit counterintuitive – it should be empty.

It's also handled differently in `fixex_strin_vector.cpp`: https://github.com/hyrise/hyrise/blob/2f0f50fcb6c484f356399ae1585aa8f2f196804d/src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp#L107-L113

However, the iterators cannot really handle a vector with more than one empty elements, and we initialize the `end` iterator in a way I think is incorrect (it's literally the same as `begin` no matter how many empty elements the vector has): https://github.com/hyrise/hyrise/blob/2f0f50fcb6c484f356399ae1585aa8f2f196804d/src/lib/storage/fixed_string_dictionary_segment/fixed_string_vector.cpp#L42-L48 

Could you please open an issue with this test's code and some short notes and reference it here?

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
```diff
@@ -208,4 +232,9 @@ TEST_F(FixedStringVectorTest, MemoryLayout) {
   EXPECT_EQ(*(fixed_string_vector->data() + 17), '\0');
 }
 
+TEST_F(FixedStringVectorTest, GetAllocator) {
+  (void)fixed_string_vector->get_allocator();
+  SUCCEED();
```
dey4ss: IMO this test is not really meaningful yet. It would be better if you create a FixedStringVector with a default-created allocator and check that `get_allocator` returns the same allocator.
```suggestion
  const auto alloc = PolymorphicAllocator<char>{};
  const auto strings = {"", ""};
  auto fs_vector = FixedStringVector(strings.begin(), strings.end(), 0, alloc);
  EXPECT_EQ(&alloc, &fixed_string_vector->get_allocator());
```

src/test/lib/storage/index/group_key/variable_length_key_test.cpp
```diff
@@ -49,4 +49,62 @@ TEST_F(VariableLengthKeyTest, CreateKeysWithOrAndShift) {
   EXPECT_TRUE(key == _key_reference);
 }
 
+
+/**
+ * Test that copy‐assignment correctly duplicates the bit pattern
+ * (this also covers operator=(const VariableLengthKeyBase&)).
+ */
```
dey4ss, line 55: Try to fit 120 characters, i.e., break later.

src/test/lib/storage/index/group_key/variable_length_key_test.cpp
```diff
@@ -49,4 +49,62 @@ TEST_F(VariableLengthKeyTest, CreateKeysWithOrAndShift) {
...
+TEST_F(VariableLengthKeyTest, CopyAssignment) {
+  const auto bytes = CompositeKeyLength{8};
+
+  auto original = VariableLengthKey{bytes};
+  original <<= CompositeKeyLength{4};
+  original |= uint64_t{0xFFFFFFFFu};
```
dey4ss: Could you please explain why this is necessary (an maybe find a way to ensure that `original` differs from the initialized `copy` below that I can still understand)?

src/test/lib/storage/index/group_key/variable_length_key_test.cpp
```diff
@@ -49,4 +49,62 @@ TEST_F(VariableLengthKeyTest, CreateKeysWithOrAndShift) {
...
+}
+
+/**
+ * Test operator!= between two distinct keys and that identical keys compare equal.
+ */
+TEST_F(VariableLengthKeyTest, InequalityOperator) {
+  const auto bytes = CompositeKeyLength{4};
+
+  auto a = VariableLengthKey{bytes};
+  a.shift_and_set(uint64_t{1}, uint8_t{8});
```
dey4ss: Can the compiler deduce the parameter type here (also below)?

src/test/lib/storage/index/group_key/variable_length_key_test.cpp
```diff
@@ -49,4 +49,62 @@ TEST_F(VariableLengthKeyTest, CreateKeysWithOrAndShift) {
...
+
+  auto b = VariableLengthKey{bytes};
+  b.shift_and_set(uint64_t{2}, uint8_t{8});
+
+  EXPECT_TRUE(a != b);
+  EXPECT_FALSE(a != a);
```
dey4ss, line 101: Can you also test `==` and `<` here (and change the test name)?

src/test/lib/storage/index/group_key/variable_length_key_test.cpp
```diff
@@ -49,4 +49,62 @@ TEST_F(VariableLengthKeyTest, CreateKeysWithOrAndShift) {
...
+
+/**
+ * Test that streaming a key to std::ostream produces a non‐empty string.
+ */
+TEST_F(VariableLengthKeyTest, OStreamOperator) {
+  const auto bytes = CompositeKeyLength{2};
+  auto key = VariableLengthKey{bytes};
+  key.shift_and_set(uint64_t{0xABCD}, uint8_t{16});
+
+  std::ostringstream oss;
```
dey4ss: auto to stick

src/test/lib/storage/index/group_key/variable_length_key_test.cpp
```diff
@@ -49,4 +49,62 @@ TEST_F(VariableLengthKeyTest, CreateKeysWithOrAndShift) {
...
+  oss << key;
+  const auto str = oss.str();
+  EXPECT_FALSE(str.empty());
```
dey4ss: I guess it would make sense to actually test the output and not only ensure it's non-empty.

src/test/lib/storage/index/group_key/variable_length_key_test.cpp
```diff
@@ -49,4 +49,62 @@ TEST_F(VariableLengthKeyTest, CreateKeysWithOrAndShift) {
...
+}
+
+
```
dey4ss: extra blank line

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
```diff
@@ -84,14 +84,28 @@ TEST_F(FixedStringVectorTest, IteratorConst) {
...
+  auto v = std::vector<pmr_string>{"abc", "def", "ghi"};
   auto fs_vector = FixedStringVector{v.cbegin(), v.cend(), 3};
 
   for (const auto& fixed_string : fs_vector) {
     EXPECT_EQ(fixed_string.size(), 3u);
   }
 }
 
+TEST_F(FixedStringVectorTest, ReverseIteratorsExplicit) {
+  auto v = std::vector<pmr_string>{"abc", "def", "ghi"};
```
dey4ss: According to the [guidelines,](https://github.com/hyrise/hyrise/blob/master/CONTRIBUTING.md#formatting-and-naming) that's not really a good name (e.g., it's too short).

dey4ss (commented):
Nice tests overall! I just have a couple of comments.

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
dey4ss: (And it could be `const`.)

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
luisegarberding: That's nice!

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
luisegarberding: I created an issue: https://github.com/hyrise/hyrise/issues/2713

luisegarberding: It would be more correct to change this line to:
`EXPECT_EQ(vec.size(), 0);`
Because this is what we expect and want, right?
But I guess we shouldn't merge a failing test. So what do we do in this case?

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
luisegarberding: Good point

src/test/lib/storage/fixed_string_dictionary_segment/fixed_string_vector_test.cpp
dey4ss: The test should pass, so please leave it at `2` for now. Please add a comment that `0` would be correct though and mention the issue you opened.

src/test/lib/storage/index/group_key/variable_length_key_test.cpp
EnnioEnnio: yeah, I didn't want to break in between `operator=(const VariableLengthKeyBase&)` but I guess I can break after `const`.

src/test/lib/storage/index/group_key/variable_length_key_test.cpp
EnnioEnnio: I noticed that `<<=` and `|=` is not tested since `shift_and_set` directly goes into `VariableLengthKeyBase`. 
But you're right - in this form it's pretty much useless and just tests if the operator executes something.

To keep things clean and concise I separated this mess into two tests. One to test copy assignment and the other to test the shifting and setting, what do you say?

src/test/lib/storage/index/group_key/variable_length_key_test.cpp
EnnioEnnio: Check! I did it in the most basic way, maybe that's not enough, let me know

dey4ss (approved):

Bouncner:
Hi group 1, feel free to merge! :)  (if you cannot merge, let us know; not sure what permissions are required)

If you want to, you can also add yourself to the contributor list.

PR #2708, opened 2025-06-04
[DYOD] Group 3: Improve test coverage

Stastez:


src/test/lib/operators/aggregate_test.cpp
```diff
@@ -217,6 +217,17 @@ TYPED_TEST(OperatorsAggregateTest, OperatorName) {
   }
 }
 
+TYPED_TEST(OperatorsAggregateTest, OperatorDescription) {
+  const auto table = this->_table_wrapper_1_1->get_output();
+  const auto aggregate_expressions = std::vector<std::shared_ptr<WindowFunctionExpression>>{
+    max_(pqp_column_(ColumnID{1}, table->column_data_type(ColumnID{1}), table->column_is_nullable(ColumnID{1}),
+                     table->column_name(ColumnID{1})))};
+  auto aggregate =
```
dey4ss: `const`?

src/test/lib/operators/update_test.cpp
```diff
@@ -65,6 +65,26 @@ class OperatorsUpdateTest : public BaseTest {
...
+  const auto projection = std::make_shared<Projection>(
+      scan, std::vector<std::shared_ptr<AbstractExpression>>{column_a});
+
+  get_table->execute();
+  scan->execute();
+  projection->execute();
+
+  const auto transaction_context = Hyrise::get().transaction_manager.new_transaction_context(AutoCommit::No);
+  const auto update = std::make_shared<Update>(table_to_update_name, scan, projection);
+  update->set_transaction_context(transaction_context);
```
dey4ss: None of this is required for obtaining the name, right?

src/test/lib/storage/index/partial_hash/partial_hash_index_test.cpp
```diff
@@ -514,6 +514,17 @@ TEST_F(PartialHashIndexTest, NotEqualsValueNotFound) {
   EXPECT_EQ(end2, cend(index));
 }
 
+TEST_F(PartialHashIndexTest, AssignFlatMapIterator) {
+  auto iterator{cbegin(index)};
+
+  iterator = cbegin(index);
```
dey4ss:
```suggestion
  auto iterator = cbegin(index);
```
?

src/test/lib/storage/index/partial_hash/partial_hash_index_test.cpp
```diff
@@ -514,6 +514,17 @@ TEST_F(PartialHashIndexTest, NotEqualsValueNotFound) {
...
+  EXPECT_EQ(iterator, cbegin(index));
```
dey4ss, line 523: Does it make sense to also dereference and check the value here?

src/test/lib/storage/index/partial_hash/partial_hash_index_test.cpp
```diff
@@ -514,6 +514,17 @@ TEST_F(PartialHashIndexTest, NotEqualsValueNotFound) {
...
+
+  iterator = cend(index);
+  EXPECT_EQ(iterator, cend(index));
+  EXPECT_NE(iterator, cbegin(index));
```
dey4ss: (same here)

src/test/lib/storage/value_segment_test.cpp
```diff
@@ -86,27 +87,76 @@ TEST_F(StorageValueSegmentTest, MemoryUsageEstimation) {
...
    */
 
+  ValueSegment<int64_t> vs_long{false, ChunkOffset{100}};
+  ValueSegment<float> vs_float{false, ChunkOffset{100}};
+
+  ValueSegment<int32_t> nullable_vs_int{true, ChunkOffset{100}};
+  ValueSegment<int64_t> nullable_vs_long{true, ChunkOffset{100}};
+  ValueSegment<float> nullable_vs_float{true, ChunkOffset{100}};
+  ValueSegment<double> nullable_vs_double{true, ChunkOffset{100}};
+  ValueSegment<pmr_string> nullable_vs_str{true, ChunkOffset{100}};
```
dey4ss: Does the deleted copy assignment of `Noncopyable` prevent auto to stick here? If so, feel free to use shared ptrs. To me, your additions pretty much call for a typed test suite instantiated for all data types (maybe an extra `TEST_F` for strings if that reuires too many `if constexpr`s in the test body).

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+TEST_F(PausableLoopThreadTest, PauseAndResume) {
+  PausableLoopThread loop(std::chrono::milliseconds(5), [&](size_t) {
+    ++call_count;
+  });
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(20));
+  loop.pause();
+
+  auto paused_count = call_count.load();
```
dey4ss: const?

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  std::this_thread::sleep_for(std::chrono::milliseconds(30));
+  EXPECT_EQ(paused_count, call_count.load());
+
+  loop.resume();
+  std::this_thread::sleep_for(std::chrono::milliseconds(20));
+  EXPECT_GT(call_count.load(), paused_count);
+}
+
+TEST_F(PausableLoopThreadTest, DestructorStopsThread) {
+  std::atomic<size_t> temp{0};
```
dey4ss: Please use auto to stick.

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  {
+    PausableLoopThread loop(std::chrono::milliseconds(10), [&](size_t) {
```
dey4ss: auto to stick

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+      ++temp;
+      ++call_count;
+    });
+    std::this_thread::sleep_for(std::chrono::milliseconds(30));
+  }
+
+  auto call_count_after_destruction = call_count.load();
+  auto temp_count_after_destruction = temp.load();
```
dey4ss: const?

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(std::chrono::milliseconds(50)));
+
+  EXPECT_EQ(call_count.load(), call_count_after_destruction);
+  EXPECT_EQ(temp.load(), temp_count_after_destruction);
+
+  SUCCEED();
```
dey4ss: Is there a reason to add this macro call?

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+TEST_F(PausableLoopThreadTest, SetLoopSleepTimeChangesInterval) {
+  auto timestamps = std::vector<std::chrono::steady_clock::time_point>();
+  std::mutex timestamps_mutex;
```
dey4ss: auto

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  PausableLoopThread loop(std::chrono::milliseconds(50), [&](size_t) {
```
dey4ss: auto

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+    std::lock_guard lock(timestamps_mutex);
```
dey4ss: auto

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+    timestamps.push_back(std::chrono::steady_clock::now());
+  });
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(125));
+
+  loop.pause();
+
+  size_t initial_timestamp_count;
```
dey4ss: auto

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  {
+    std::lock_guard lock(timestamps_mutex);
```
dey4ss: auto

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+    initial_timestamp_count = timestamps.size();
+  }
+
+  EXPECT_GE(initial_timestamp_count, 2);
+
+  loop.set_loop_sleep_time(std::chrono::milliseconds(10));
+  {
+    std::lock_guard lock(timestamps_mutex);
```
dey4ss: auto

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+    EXPECT_EQ(initial_timestamp_count, timestamps.size());
+  }
+
+  loop.resume();
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(55));
+  {
+    std::lock_guard lock(timestamps_mutex);
```
dey4ss: auto

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+    // We expect at least 4-5 new timestamps in this period.
+    // Check for more than initial + 3 to be safe.
```
dey4ss: Please break after 120 characters.

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+    ++call_count;
+  });
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(30));
+
+  EXPECT_GE(call_count.load(), 1);
+}
+
+TEST_F(PausableLoopThreadTest, PauseAndResume) {
+  PausableLoopThread loop(std::chrono::milliseconds(5), [&](size_t) {
```
dey4ss: auto

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+ protected:
+  void SetUp() override {
+    call_count = 0;
+  }
+
+  std::atomic<size_t> call_count{0};
+};
+
+TEST_F(PausableLoopThreadTest, BasicExecution) {
+  PausableLoopThread loop(std::chrono::milliseconds(10), [&](size_t) {
```
dey4ss: auto

dey4ss (commented):
Nice catch that changing `PausableLoopThread`'s pause time was not safe yet! Plase, _stick_ to the code guidelines. I think testing the memory usage estimation for different data types should be a typed test to reduce redundancy.

src/test/lib/utils/pausable_loop_thread_test.cpp
beerjamin: Made const with 7a1e8d4311377d725126a0142882f5f2a0c836db

src/test/lib/utils/pausable_loop_thread_test.cpp
beerjamin: auto to stick in 7a1e8d4311377d725126a0142882f5f2a0c836db.

src/test/lib/utils/pausable_loop_thread_test.cpp
beerjamin: Changed to auto in 7a1e8d4311377d725126a0142882f5f2a0c836db.

I think it could make sense to extend the clang-tidy and linter rules to enforce some of the coding guidelines. Stick to auto might be difficult to enforce (at least I am not aware of a rule that does it) but something like [modernize-use-auto](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-auto.html) would save some review time and CI resources.

src/test/lib/utils/pausable_loop_thread_test.cpp
beerjamin: Maybe I am missing something, but the annotated lines are both under 120 characters.

src/test/lib/storage/index/partial_hash/partial_hash_index_test.cpp
Stastez: As far as I can tell, this would call the constructor, not the assignment operator. I have however changed the value with which the iterator is constructed to make sure that it actually is the assignment operator being tested.

src/test/lib/utils/pausable_loop_thread_test.cpp
Bouncner: We usually try to fill each line as much as possible (meaning, the second sentence should start on the first line) because over time people add sentence after sentence and we end up with 10 lines each using only half of the possible space.

src/test/lib/utils/pausable_loop_thread_test.cpp
Bouncner: Can you please resolve everything that is easy to resolve (like here when adding `auto` or `const`)?

Makes reviewing easier for us. :)

src/test/lib/utils/pausable_loop_thread_test.cpp
beerjamin: Ok that makes sense. Thank you for the explanation. Now the comment is a single line with 871dc80b306fefde2ab1a1579e428c6c6e4eb6fd

src/lib/storage/value_segment.cpp
```diff
@@ -160,9 +160,9 @@ size_t ValueSegment<T>::memory_usage(const MemoryUsageCalculationMode mode) cons
 
   if constexpr (std::is_same_v<T, pmr_string>) {
     return common_elements_size + string_vector_memory_usage(_values, mode);
+  } else {
+    return common_elements_size + _values.capacity() * sizeof(T);
```
Bouncner: Why the change?

src/lib/utils/pausable_loop_thread.hpp
```diff
@@ -10,8 +10,9 @@
 namespace hyrise {
 
 // This class spawns a thread that executes a procedure, defined by @param loop_func, in a loop. Between each iteration
-// there is a user-definable sleep period. The loop can be paused, resumed and finished. The loop starts in paused
-// state. Whenever the @param loop_func is called, an execution counter is passed to it.
+// there is a user-definable sleep period. The loop can be paused, resumed and finished. The loop starts executing
```
Bouncner:
```suggestion
// there is a user-definable sleep period. The loop can be paused, resumed, and finished. The loop starts executing
```

src/test/lib/storage/index/partial_hash/partial_hash_index_test.cpp
```diff
@@ -514,6 +514,23 @@ TEST_F(PartialHashIndexTest, NotEqualsValueNotFound) {
   EXPECT_EQ(end2, cend(index));
 }
 
+TEST_F(PartialHashIndexTest, AssignFlatMapIterator) {
+  // auto iterator = something; would not call the assignment operator, but the constructor.
+  auto iterator = FlatMapIterator{cend(index)};
+
+  // Explicitely invoke assign operator
```
Bouncner:
```suggestion
  // Explicitely invoke assign operator.
```

src/test/lib/storage/index/partial_hash/partial_hash_index_test.cpp
```diff
@@ -514,6 +514,23 @@ TEST_F(PartialHashIndexTest, NotEqualsValueNotFound) {
...
+  iterator = cbegin(index);
+  EXPECT_EQ(iterator, cbegin(index));
+  EXPECT_EQ(*iterator, *cbegin(index));
+  EXPECT_NE(iterator, cend(index));
+
+  // Test re-assignment of the same value
```
Bouncner:
```suggestion
  // Test re-assignment of the same value.
```

src/test/lib/storage/index/partial_hash/partial_hash_index_test.cpp
```diff
@@ -514,6 +514,23 @@ TEST_F(PartialHashIndexTest, NotEqualsValueNotFound) {
...
+  iterator = cbegin(index);
```
Bouncner, line 528: Is there a reason to expect that something unexpected could happen when we re-assign?

src/test/lib/storage/value_segment_test.cpp
```diff
@@ -80,33 +81,51 @@ TEST_F(StorageValueSegmentTest, ArraySubscriptOperatorReturnsNullValue) {
...
-  EXPECT_GE(vs_str.memory_usage(MemoryUsageCalculationMode::Sampled), empty_usage_str);
-  // The short string will fit within the SSO capacity of a string and the long string will be placed on the heap.
-  EXPECT_EQ(vs_str.memory_usage(MemoryUsageCalculationMode::Full), empty_usage_str + longer_str.capacity() + 1);
+TYPED_TEST(TypedStorageValueSegmentTest, MemoryUsageEstimation) {
+  /**
+    * As ValueSegments are pre-allocated, their size should not change when inserting data, except for strings placed
+    * on the heap.
+    */
+
+  // this is necessary here because we are in a derived class template
```
Bouncner:
```suggestion
  // `this` is necessary here because we are in a derived class template.
```

src/test/lib/storage/value_segment_test.cpp
```diff
@@ -80,33 +81,51 @@ TEST_F(StorageValueSegmentTest, ArraySubscriptOperatorReturnsNullValue) {
...
+  const auto empty_usage = this->vs.memory_usage(MemoryUsageCalculationMode::Sampled);
+  const auto empty_usage_nullable = this->nullable_vs.memory_usage(MemoryUsageCalculationMode::Sampled);
+
+  if constexpr (std::is_same_v<TypeParam, pmr_string>) {
+    const auto short_str = pmr_string{"Hello"};
+    const auto longer_str = pmr_string{"HelloWorldHaveANiceDayWithSunshineAndGoodCofefe"};
```
Bouncner, line 106: <3

src/test/lib/storage/value_segment_test.cpp
```diff
@@ -80,33 +81,51 @@ TEST_F(StorageValueSegmentTest, ArraySubscriptOperatorReturnsNullValue) {
...
+    this->vs.append(longer_str);
+
+    this->nullable_vs.append(short_str);
+    this->nullable_vs.append(longer_str);
+    this->nullable_vs.append(NULL_VALUE);
+
+    EXPECT_GE(this->vs.memory_usage(MemoryUsageCalculationMode::Sampled), empty_usage);
+    EXPECT_GE(this->nullable_vs.memory_usage(MemoryUsageCalculationMode::Sampled), empty_usage_nullable);
+    // The short string will fit within the SSO capacity of a string and the long string will be placed on the heap.
+    EXPECT_EQ(this->vs.memory_usage(MemoryUsageCalculationMode::Full), empty_usage + longer_str.capacity() + 1);
```
Bouncner, line 118: + 1 for the null byte?

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+  auto loop = PausableLoopThread(std::chrono::milliseconds(50), [&](size_t) {
+    auto lock = std::lock_guard(timestamps_mutex);
+    timestamps.push_back(std::chrono::steady_clock::now());
+  });
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(125));
+
+  loop.pause();
+
+  auto initial_timestamp_count = 0;
```
Bouncner: Please be specific with the integer type (`(u)int32/64_t` ...).

src/lib/utils/pausable_loop_thread.cpp
```diff
@@ -15,7 +15,7 @@ PausableLoopThread::PausableLoopThread(std::chrono::milliseconds loop_sleep_time
   _loop_thread = std::thread([&, loop_func] {
     size_t counter = 0;
     while (!_shutdown_flag) {
-      auto lock = std::unique_lock<std::mutex>{_mutex};
+      auto lock = std::unique_lock{_mutex};
```
Bouncner, line 18: Why does this single thread need a lock? I am not sure I can follow.

I know this is not your code, but you also added it to `set_loop_sleep_time` and I was wondering which scenario we are covering here.

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+      ++temp;
+      ++call_count;
+    });
+    std::this_thread::sleep_for(std::chrono::milliseconds(30));
+  }
+
+  const auto call_count_after_destruction = call_count.load();
+  const auto temp_count_after_destruction = temp.load();
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(std::chrono::milliseconds(50)));
```
Bouncner: std::chrono::milliseconds(std::chrono::milliseconds(50)

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,98 @@
+#include "base_test.hpp"
+#include "utils/pausable_loop_thread.hpp"
+
+namespace hyrise {
+
+class PausableLoopThreadTest : public BaseTest {
```
Bouncner: Not that I have an immediate idea how to do that better, but I am wondering if we (i) can get into trouble when the CI system is under very high load (will the tests fail?) and (ii) if we can speed up the tests (this will probably add a second).

src/lib/utils/pausable_loop_thread.cpp
Bouncner: Is it because of setting the sleep time while reading it?

Could we rather use `std::atomic<std::chrono::milliseconds::rep> _loop_sleep_time_ms`?

Bouncner (commented):
Mostly some comments on the pausable loop thread. Most of it is actually not your code.

src/lib/storage/value_segment.cpp
Stastez: The semantic is the same, and I believe that this will actually yield the same machine code when compiled in release mode.
However,  I 1) think this is cleaner as it better communicates the intent, and 2) the test coverage report complained about the "return for non-string" line not being covered for the `pmr_string` instantiation of the test. It of course should never be hit in the `pmr_string` instantiation, but I believe that a debug build actually contains this unreachable code.

src/test/lib/storage/index/partial_hash/partial_hash_index_test.cpp
Stastez: The idea was that assignment operators could theoretically have two branches: 1) assigning a new value, and 2) re-assigning the same value. An assignment operator could theoretically implement these two cases differently (which `FlatMapIterator::operator=` technically actually does), so the idea was to test that both cases yield the desired result.
But if this seems unnecessary I would of course change this.

src/test/lib/storage/value_segment_test.cpp
Stastez: Actually, I have no idea. The +1 is just copied from the previous version of this test.

src/test/lib/utils/pausable_loop_thread_test.cpp
beerjamin: Changed in 0f2d79d52a55a7b687b6fe2eeb6a7b0ea46d82e9

src/lib/utils/pausable_loop_thread.cpp
beerjamin: You are correct, one issue was setting the sleep time while reading it. I tried `std::atomic<std::chrono::milliseconds::rep> _loop_sleep_time_ms` but for some reason, the macos tests were failing and I dont know why. Debugging using the CI is really cumbersome because the FullCI was taking around 2 hours and the macos builds were the very last stage.

Bouncner: In case you don't have a mac at hand, can you give me the commit id? I can test it.

Otherwise, I would vote for an `std::atomic<size_t> sleep_time_ms` or so. Solve the problem without locking and should still be easy to maintain/read.

src/test/lib/utils/pausable_loop_thread_test.cpp
Bouncner: Sorry, my comment wasn't clear enough. You shouldn't need to do this.  A single `chrono::milliseconds` should suffice.

src/lib/storage/value_segment.cpp
Stastez: Is this change alright with you, @Bouncner ?

Bouncner: I am surprise that clang-tidy did not complain: https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html

But today, I learned that removing the else can actually lead to problems (see https://reviews.llvm.org/D53372).
Can you link this issue and name `clang-tidy` and `else-after-return` in the comment?
It might help people.

Bouncner: So yes, please leave the code as you have it and add a comment. Nice catch!

src/lib/utils/pausable_loop_thread.cpp
```diff
@@ -65,9 +65,8 @@ void PausableLoopThread::resume() {
   _cv.notify_one();
 }
 
-void PausableLoopThread::set_loop_sleep_time(std::chrono::milliseconds loop_sleep_time) {
-  auto lock = std::unique_lock{_mutex};
-  _loop_sleep_time = loop_sleep_time;
+void PausableLoopThread::set_loop_sleep_time(size_t loop_sleep_time_ms) {
```
Bouncner: I’d keep the interface simpler (chrono milliseconds) and just use internally the atomic. Cleaner interface.


What do you think?

beerjamin: Yes thats much better. This way I dont have to change other files and PausableLoopThread has more explicit units in its constructor and methods. Fixed in 635277205618a18866dab9880cac9ee2c4cc38c4

src/lib/utils/pausable_loop_thread.cpp
beerjamin: Done in 635277205618a18866dab9880cac9ee2c4cc38c4

src/test/lib/utils/pausable_loop_thread_test.cpp
beerjamin: I think its possible that the tests fail if the CI system is under very high load. I also dont know how to improve this. But I have to say that its difficult to test the behavior of this class since the busy waiting makes the thread susceptible to the unpredictability of the OS thread scheduler. I also think that a failing test would actually point out that the implementation needs some reworking.

```
void PausableLoopThread::pause() {
  if (!_loop_thread.joinable()) {
    return;
  }
  _pause_requested = true;
  while (!_is_paused) {}
}
```

src/lib/utils/pausable_loop_thread.cpp
Bouncner: (removed question about semaphores)

Bouncner: Any reason not to move the lock into the conditional?

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+ protected:
+  void SetUp() override {
+    call_count = 0;
+  }
+
+  std::atomic<size_t> call_count{0};
+};
+
+TEST_F(PausableLoopThreadTest, BasicExecution) {
+  auto loop = PausableLoopThread(std::chrono::milliseconds(std::chrono::milliseconds(10)), [&](size_t) {
```
Bouncner: `std::chrono::milliseconds(std::chrono::milliseconds` > `std::chrono::milliseconds` ... several places.

src/test/lib/utils/pausable_loop_thread_test.cpp
```diff
@@ -0,0 +1,98 @@
...
+  EXPECT_EQ(call_count.load(), call_count_after_destruction);
+  EXPECT_EQ(temp.load(), temp_count_after_destruction);
+}
+
+TEST_F(PausableLoopThreadTest, SetLoopSleepTimeChangesInterval) {
+  auto timestamps = std::vector<std::chrono::steady_clock::time_point>();
+  auto timestamps_mutex = std::mutex();
+
+  auto loop = PausableLoopThread(std::chrono::milliseconds(std::chrono::milliseconds(50)), [&](size_t) {
+    auto lock = std::lock_guard(timestamps_mutex);
```
Bouncner: Why do we need the mutex when only a single thread is writing `timestamps`?

src/lib/utils/pausable_loop_thread.hpp
```diff
@@ -28,6 +29,6 @@ struct PausableLoopThread {
   std::mutex _mutex;
   std::condition_variable _cv;
   std::thread _loop_thread;
-  std::chrono::milliseconds _loop_sleep_time;
```
Bouncner, line 31: Can you add a short comment that `std::atomic<std::chrono::milliseconds>` did not compile with all CI nodes at the time of writing?

Stastez:
@Bouncner I have merged master. I am unsure why the CI run fails, as all tests in the run marked as failed actually passed, apart from five that were skipped outright.

Bouncner:
I restarted the CI run. Now, I failed with the following test:
`[  FAILED  ] PausableLoopThreadTest.SetLoopSleepTimeChangesInterval`

PR #2710, opened 2025-06-05
[DYOD] Group 2: Improve Code Coverage

robert-oleynik:


src/test/lib/import_export/csv/csv_parser_test.cpp
```diff
@@ -16,6 +16,12 @@ TEST_F(CsvParserTest, SingleFloatColumn) {
   EXPECT_TABLE_EQ_ORDERED(table, expected_table);
 }
 
+TEST_F(CsvParserTest, SingleLongWithNullColumn) {
+  auto table = CsvParser::parse("resources/test_data/csv/long_with_null.csv");
+  std::shared_ptr<Table> expected_table = load_table("resources/test_data/tbl/long_with_null.tbl", ChunkOffset{5});
```
dey4ss: `const auto` * 2

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
   EXPECT_THROW(chunk->mark_as_full(), std::logic_error);
 }
 
+TEST_F(StorageChunkTest, ReferencesExactlyOneTable) {
+  // Create a test table with two columns
```
dey4ss: Pick here: Pleasse end comments with a full stop.

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+  TableColumnDefinitions column_definitions;
```
dey4ss: Please use auto to stick.

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+  column_definitions.emplace_back("a", DataType::Int, true);
+  column_definitions.emplace_back("b", DataType::String, true);
+  auto test_table = std::make_shared<Table>(column_definitions, TableType::Data, ChunkOffset{3});
```
dey4ss: const

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+  test_table->append_chunk({vs_int, vs_str});
+
+  // Create a second test table
+  TableColumnDefinitions column_definitions2;
+  column_definitions2.emplace_back("c", DataType::Int, false);
+  column_definitions2.emplace_back("d", DataType::String, false);
+  auto test_table2 = std::make_shared<Table>(column_definitions2, TableType::Data, ChunkOffset{3});
```
dey4ss: (see above)

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+  test_table2->append_chunk({ds_int, ds_str});
+
+  // Chunk with non-reference segments
+  { EXPECT_FALSE(chunk->references_exactly_one_table()); }
```
dey4ss: Why the extra brackets?

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+
+  // Chunk with reference segments to the same table with the same pos_list
+  {
+    auto pos_list = std::make_shared<RowIDPosList>();
```
dey4ss: const

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+    pos_list->emplace_back(ChunkID{0}, ChunkOffset{0});
+    pos_list->emplace_back(ChunkID{0}, ChunkOffset{1});
+
+    auto ref_segment1 = std::make_shared<ReferenceSegment>(test_table, ColumnID{0}, pos_list);
+    auto ref_segment2 = std::make_shared<ReferenceSegment>(test_table, ColumnID{1}, pos_list);
```
dey4ss: const

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+
+    Segments reference_segments;
```
dey4ss: auto to stick

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+    reference_segments.push_back(ref_segment1);
+    reference_segments.push_back(ref_segment2);
+
+    auto reference_chunk = std::make_shared<Chunk>(reference_segments);
```
dey4ss: const

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+    EXPECT_TRUE(reference_chunk->references_exactly_one_table());
+  }
+
+  // Chunk with reference segments to different tables
+  {
+    auto pos_list = std::make_shared<RowIDPosList>();
```
dey4ss: const

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+    pos_list->emplace_back(ChunkID{0}, ChunkOffset{0});
+
+    auto ref_segment1 = std::make_shared<ReferenceSegment>(test_table, ColumnID{0}, pos_list);
+    auto ref_segment2 = std::make_shared<ReferenceSegment>(test_table2, ColumnID{0}, pos_list);
```
dey4ss: const

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+
+    Segments mixed_reference_segments;
```
dey4ss: auto to stick

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+    mixed_reference_segments.push_back(ref_segment1);
+    mixed_reference_segments.push_back(ref_segment2);
+
+    auto mixed_reference_chunk = std::make_shared<Chunk>(mixed_reference_segments);
```
dey4ss: const

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+    pos_list2->emplace_back(ChunkID{0}, ChunkOffset{1});
+
+    auto ref_segment1 = std::make_shared<ReferenceSegment>(test_table, ColumnID{0}, pos_list1);
+    auto ref_segment2 = std::make_shared<ReferenceSegment>(test_table, ColumnID{1}, pos_list2);
+
+    Segments different_pos_list_segments;
+    different_pos_list_segments.push_back(ref_segment1);
+    different_pos_list_segments.push_back(ref_segment2);
+
+    auto different_pos_list_chunk = std::make_shared<Chunk>(different_pos_list_segments);
```
dey4ss: (see above)

src/test/lib/storage/segment_access_counter_test.cpp
```diff
@@ -196,4 +196,46 @@ TEST_F(SegmentAccessCounterTest, AccessPattern8) {
   EXPECT_EQ(_access_pattern(positions), AccessPattern::Random);
 }
 
+TEST_F(SegmentAccessCounterTest, SelfAssignment) {
+  SegmentAccessCounter counter;
```
dey4ss: auto to stick

src/test/lib/storage/segment_access_counter_test.cpp
```diff
@@ -196,4 +196,46 @@ TEST_F(SegmentAccessCounterTest, AccessPattern8) {
...
+  counter[SegmentAccessCounter::AccessType::Point] = 10;
+  counter[SegmentAccessCounter::AccessType::Sequential] = 20;
+
+  // Store original values to check after self-assignment
```
dey4ss: Pick here: End with full stop.

src/test/lib/storage/segment_access_counter_test.cpp
```diff
@@ -196,4 +196,46 @@ TEST_F(SegmentAccessCounterTest, AccessPattern8) {
...
+  const auto original_point = counter[SegmentAccessCounter::AccessType::Point].load();
+  const auto original_sequential = counter[SegmentAccessCounter::AccessType::Sequential].load();
+
+  // Assign via a reference to avoid -Wself-assign-overloaded
+  SegmentAccessCounter& ref = counter;
```
dey4ss: auto

src/test/lib/storage/segment_access_counter_test.cpp
```diff
@@ -196,4 +196,46 @@ TEST_F(SegmentAccessCounterTest, AccessPattern8) {
...
+  counter = ref;
+
+  EXPECT_EQ(counter[SegmentAccessCounter::AccessType::Point], original_point);
+  EXPECT_EQ(counter[SegmentAccessCounter::AccessType::Sequential], original_sequential);
+}
+
+TEST_F(SegmentAccessCounterTest, InequalityOperator) {
+  SegmentAccessCounter counter1;
+  SegmentAccessCounter counter2;
```
dey4ss: auto

dey4ss (commented):
Nice work! I have quite some comments on `const` variables and the use of auto to stick. Please also merge `SegmentAccessCounterTest.EqualityOperator` and `SegmentAccessCounterTest.InequalityOperator` (name it, e.g., `EqualityAndInequality`).

src/test/lib/storage/segment_access_counter_test.cpp
dey4ss: Maybe change a all counters to not have the majority zero-initialized in this test for equality and compare to all possible counters differing in one access type for inequality? Thus, we would actually notice if the comparison operators miss on some comparisons.

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
   EXPECT_THROW(chunk->mark_as_full(), std::logic_error);
 }
 
+TEST_F(StorageChunkTest, ReferencesExactlyOneTable) {
+  // Create a test table with two columns.
+  auto column_definitions = TableColumnDefinitions();
```
Bouncner: You could do something like 
```
const auto column_definitions = TableColumnDefinitions{{"a", DataType::Int, true}, {"b", DataType::String, true}};
```

src/test/lib/storage/segment_access_counter_test.cpp
```diff
@@ -196,4 +196,53 @@ TEST_F(SegmentAccessCounterTest, AccessPattern8) {
   EXPECT_EQ(_access_pattern(positions), AccessPattern::Random);
 }
 
+TEST_F(SegmentAccessCounterTest, SelfAssignment) {
```
Bouncner, line 201: Do you want to test assignment or really self-assignment?

Appear a bit "edge casey".

src/test/lib/storage/segment_access_counter_test.cpp
```diff
@@ -196,4 +196,53 @@ TEST_F(SegmentAccessCounterTest, AccessPattern8) {
...
+  const auto access_types = {AccessType::Point, AccessType::Sequential, AccessType::Monotonic, AccessType::Random,
+                             AccessType::Dictionary};
+  for (const auto access_type : access_types) {
+    auto modified_counter = zero_counter;
+
+    EXPECT_TRUE(zero_counter == modified_counter);  // Should be equal.
+
+    modified_counter[access_type] = 42;
+
+    // Should be different.
```
Bouncner: Should be equal, should be different. :D

src/test/lib/storage/segment_access_counter_test.cpp
```diff
@@ -196,4 +196,53 @@ TEST_F(SegmentAccessCounterTest, AccessPattern8) {
...
+  const auto access_types = {AccessType::Point, AccessType::Sequential, AccessType::Monotonic, AccessType::Random,
+                             AccessType::Dictionary};
+  for (const auto access_type : access_types) {
+    auto modified_counter = zero_counter;
+
+    EXPECT_FALSE(zero_counter != modified_counter);  // Should be equal.
+
+    modified_counter[access_type] = 42;
+
+    // Should be different.
```
Bouncner: See below.

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+
+    const auto ref_segment1 = std::make_shared<ReferenceSegment>(test_table, ColumnID{0}, pos_list1);
+    const auto ref_segment2 = std::make_shared<ReferenceSegment>(test_table, ColumnID{1}, pos_list2);
+
+    auto different_pos_list_segments = Segments();
+    different_pos_list_segments.push_back(ref_segment1);
+    different_pos_list_segments.push_back(ref_segment2);
+
+    const auto different_pos_list_chunk = std::make_shared<Chunk>(different_pos_list_segments);
+    EXPECT_FALSE(different_pos_list_chunk->references_exactly_one_table());
```
Bouncner: That's surprising to see. But checking the code, you are right. But then again: that's absolutely not what I would expect given the function's name.

src/test/lib/storage/chunk_test.cpp
```diff
@@ -366,4 +366,75 @@ TEST_F(StorageChunkTest, TrySetImmutable) {
...
+    const auto pos_list = std::make_shared<RowIDPosList>();
+    pos_list->emplace_back(ChunkID{0}, ChunkOffset{0});
+    pos_list->emplace_back(ChunkID{0}, ChunkOffset{1});
+
+    const auto ref_segment1 = std::make_shared<ReferenceSegment>(test_table, ColumnID{0}, pos_list);
+    const auto ref_segment2 = std::make_shared<ReferenceSegment>(test_table, ColumnID{1}, pos_list);
+
+    auto reference_segments = Segments();
+    reference_segments.push_back(ref_segment1);
+    reference_segments.push_back(ref_segment2);
```
dey4ss: (also below):
```c++
const auto reference_segments = Segments{ref_segment1, ref_segment2};
```
?

src/test/lib/storage/chunk_test.cpp
dey4ss: I don't really get what surprises you here. Just scrolled over it: We create a chunk with two reference segments. These reference segments point to the same table (though they don't share the same position list). So it's literally what the function name says, isn't it?

Bouncner: > These reference segments point to the same table [..]

And `references_exactly_one_table()` is false?

dey4ss: _That_ was my mistake, I skipped the `EXPECT_FALSE` and just assumed it was true. 🙄
Should be something like `references_same_positions`, `shares_single_position_list`, or alike, right?

Bouncner: Need to check how it’s used (right usage, just imperfect name; or right name and we don’t run some code because the check is wrong?). Maybe we need both?
But it’s definitely something we can improve.

dey4ss: We use it in two assertions: once in the Delete operator and once in the Validate operator. In both cases, one PosList is required because we only use the PosList of the first segment to determine which rows should be deleted/validated.

Bouncner: So `shares_single_position_list`?

dey4ss: I think that's better than the current name. It does not tell that also the referenced table must be the same across the reference segments though.

Bouncner: But as a single ref segment can only reference one table that should be rather clear, right?

As the chunk does not share, maybe `ref_segments_share_single_position_list` (`reference_segments_share...` is pretty long). Or we could use `references_multiple_positions_lists` which is only true when multipe reference segments do not share.

dey4ss: > But as a single ref segment can only reference one table that should be rather clear, right?

At least I could not easily come up with a case where the PosList is be the same, but the tables are not. That's why I suggested the name in the first place; just wanted to double-check with you.

`references_multiple_positions_lists` sounds even better to me. Or `segments_reference_different_positions`/`references_different_positions`? The same positions in different tables are not equal, neither are different positions in the same table.

dey4ss: @zainab-rehan @hieutdle @robert-oleynik We just discussed the topic and came to the following conclusion:
1. The method should be renamed to `segments_share_table_and_positions`.
2. After line 179, add a debug assert to ensure that different position lists indeed reference different positions (you can use `if constexpr (HYRISE_DEBUG) {...}` for that). https://github.com/hyrise/hyrise/blob/cb35cf306059ec4a498b1a0c2c2f40b4f1af4fa1/src/lib/storage/chunk.cpp#L179-L181
4. Please add a case in your test for 2. that expects that an error is thrown.
5. Please add the following comment to the renamed method in the header:
   `Some operators only check the first segment to obtain positions (e.g., Delete and Validate). This method checks that all reference segments point to the same positions in the same table.`

Bouncner:
Dear @zainab-rehan @hieutdle @robert-oleynik, can you please briefly update us here?

Bouncner:
> Dear @zainab-rehan @hieutdle @robert-oleynik, can you please briefly update us here?

Ping @zainab-rehan @hieutdle @robert-oleynik.

PR #2714, opened 2025-06-18
[Readme] Make bibtex reference more obvious

Bouncner:
As people appear to be still struggling with properly referencing Hyrise (cf. https://dl.acm.org/doi/pdf/10.1145/3677141), this PR attempts to make the bibtex reference more obvious.

Further, it removes some clutter from the entry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,39 +12,18 @@ You can still find the (archived) previous version of Hyrise on [Github](https:/
 
 ## Citation
 
-When referencing this version of Hyrise, please use the following bibtex entry:
-<details><summary>(click to expand)</summary>
-  
+When referencing this version of Hyrise, please use the following bibtex entry: 
 ```bibtex
-@inproceedings{DBLP:conf/edbt/DreselerK0KUP19,
-  author    = {Markus Dreseler and
-               Jan Kossmann and
-               Martin Boissier and
-               Stefan Klauck and
-               Matthias Uflacker and
-               Hasso Plattner},
-  editor    = {Melanie Herschel and
-               Helena Galhardas and
-               Berthold Reinwald and
-               Irini Fundulaki and
-               Carsten Binnig and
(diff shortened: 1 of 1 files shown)


dey4ss (approved):

README.md
```diff
@@ -12,39 +12,18 @@ You can still find the (archived) previous version of Hyrise on [Github](https:/
...
-               Berthold Reinwald and
-               Irini Fundulaki and
-               Carsten Binnig and
-               Zoi Kaoudi},
-  title     = {Hyrise Re-engineered: An Extensible Database System for Research in
-               Relational In-Memory Data Management},
-  booktitle = {Advances in Database Technology - 22nd International Conference on
-               Extending Database Technology, {EDBT} 2019, Lisbon, Portugal, March
-               26-29, 2019},
+@inproceedings{DBLPlike:conf/edbt/DreselerK0KUP19,
```
dey4ss: Is there a need to alter the DBLP key?

Bouncner: Not really, I just thought the original one is too cluttered and I did not want to keep the key when it’s no longer “the original”.

Maybe key `hyrise_reengineered`?

dey4ss: Hm, or just keep the source tag? I mean getting directly to the DBLP entry is actually a quite nice service. We could as well just remove the `DBLP:` prefix.

Bouncner: > We could as well just remove the DBLP: prefix

I like that.

Zündung?

tjeyy:
How do you like the idea of printing the citation to the console when Hyrise is started?
Maybe alongside an appeal to cite it if Hyrise is used in one's research?
I saw another project doing this and liked the idea.

README.md
Bouncner: @dey4ss ?

dey4ss: sure

PR #2718, opened 2025-08-01
Expand sort definitions to support `NULLS LAST`

dey4ss:


src/benchmarklib/abstract_table_generator.cpp
```diff
@@ -114,7 +114,7 @@ void AbstractTableGenerator::generate_and_store() {
 
         const auto sort_table = [&]() {
           auto& table = table_info_by_name[table_name].table;
-          const auto sort_mode = SortMode::Ascending;  // currently fixed to ascending
+          const auto sort_mode = SortMode::AscendingNullsFirst;  // Currently fixed to ascending.
```
Bouncner: "ascending and NULLs first"?

src/test/lib/logical_query_plan/sort_node_test.cpp
```diff
@@ -62,9 +69,11 @@ TEST_F(SortNodeTest, HashingAndEqualityCheck) {
 TEST_F(SortNodeTest, Copy) {
   EXPECT_EQ(*_sort_node->deep_copy(), *_sort_node);
 
-  const auto sort_b = SortNode::make(
-      expression_vector(_a_d, _a_f, _a_i),
-      std::vector<SortMode>{SortMode::Descending, SortMode::Ascending, SortMode::Descending}, _table_node);
+  const auto sort_b =
```
Bouncner, line 72: For the LQP tests, we should easily be able to test NULLS LAST as well, right?

src/test/lib/operators/table_scan_test.cpp
```diff
@@ -92,15 +92,15 @@ class OperatorsTableScanTest : public BaseTest, public ::testing::WithParamInter
...
+                                 std::make_optional(std::vector<SortColumnDefinition>{
+                                     SortColumnDefinition(ColumnID(0), SortMode::AscendingNullsFirst)}));
   }
 
   std::shared_ptr<TableWrapper> get_int_only_null_op() {
-    return load_and_encode_table(
-        "resources/test_data/tbl/int_only_null.tbl", ChunkOffset{4},
-        std::make_optional(std::vector<SortColumnDefinition>{SortColumnDefinition(ColumnID(0), SortMode::Ascending)}));
+    return load_and_encode_table("resources/test_data/tbl/int_only_null.tbl", ChunkOffset{4},
+                                 std::make_optional(std::vector<SortColumnDefinition>{
```
Bouncner: Do we need the `make_optional`?

src/test/lib/sql/sql_translator_test.cpp
```diff
@@ -378,7 +378,7 @@ TEST_F(SQLTranslatorTest, SelectListAliasUsedInOrderBy) {
       sql_to_lqp_helper("SELECT a AS x, b AS y FROM int_float ORDER BY x, y");
 
   const auto aliases = std::vector<std::string>({"x", "y"});
-  const auto sort_modes = std::vector<SortMode>({SortMode::Ascending, SortMode::Ascending});
+  const auto sort_modes = std::vector<SortMode>({SortMode::AscendingNullsFirst, SortMode::AscendingNullsFirst});
```
Bouncner, line 387: General question: I wonder where we need to add tests for `NULLS LAST`? Getting this PR landed helps a lot, but I don't know when we will revisit important parts where we do not properly test?

dey4ss: That's a good point. Usually, we do not consider sort orders, and I added assertions in `SortedSegmentSearch` (one of the few places where we make use of sorting). LQP tests, for sure.

We could add some queries in the `SQLTranslatorTest` that request `NULLS LAST` and are expected not to be valid. Thus, we definitely notice if the parser allows `NULLS LAST`.

I looked at the spots where we actually use the sorting information (basically some `TableScans` and `AggregateSort`). I think, there is something we did not really consider yet:
https://github.com/hyrise/hyrise/blob/1b7b592d158b852dab9b8750533e326053410bee/src/lib/operators/table_scan/column_vs_value_table_scan_impl.cpp#L49-L54
https://github.com/hyrise/hyrise/blob/1b7b592d158b852dab9b8750533e326053410bee/src/lib/operators/table_scan/column_between_table_scan_impl.cpp#L54-L59
https://github.com/hyrise/hyrise/blob/1b7b592d158b852dab9b8750533e326053410bee/src/lib/operators/table_scan/column_is_null_table_scan_impl.cpp#L58-L63

These `TableScan` implementations assume that a chunk is ordered by all columns in `individually_sorted_by` (which kind of explains the name and also the current assertion). E.g., if the vector contains columns `a`, `b`, `c`, it means that the current chunk has the same sort order for each of the statements `ORDER BY a`, `ORDER by b`, `ORDER by c`, not `ORDER BY a, b, c`. The `IsNullScan` even misses a return I guess.

For `AggregateSort`, I'm not even sure if it can handle different sort orders:
https://github.com/hyrise/hyrise/blob/1b7b592d158b852dab9b8750533e326053410bee/src/lib/operators/aggregate_sort.cpp#L335-L343

src/test/lib/logical_query_plan/sort_node_test.cpp
dey4ss: Added occurrences of NULLs last for description and equality

src/test/lib/sql/sql_translator_test.cpp
dey4ss: Checked on `AggregateSort` (which seems to be fine) and added some SQLTranslator tests. Also added a NULLs last sort mode in tests for LQPTranslator. I guess that's now pretty much everywhere where we currently have to take care.

The individually sorted by thing: What do you think? We either have to change the operators or the method description.

dey4ss: Or we would have just read the comment before:
https://github.com/hyrise/hyrise/blob/1b7b592d158b852dab9b8750533e326053410bee/src/lib/storage/chunk.hpp#L142-L145

dey4ss: Added a test for that

src/lib/storage/chunk.cpp
```diff
@@ -266,16 +266,20 @@ void Chunk::set_individually_sorted_by(const SortColumnDefinition& sorted_by) {
...
+  // Currently, we assume that `set_individually_sorted_by` is called only once at most. Thus, there should be no
+  // existing sorting and the new sorting should contain at least one column. Feel free to remove this assertion if
+  // necessary.
   Assert(!sorted_by.empty() && _sorted_by.empty(), "Sorting information cannot be empty or reset.");
 
   if constexpr (HYRISE_DEBUG) {
-    const auto& sorted_by_column = sorted_by[0];
-    const auto sorted_segment = get_segment(sorted_by_column.column);
-    if (sorted_segment->size() > 1) {
+    // "Individually sorted by" means that the chunk is sorted by ALL passed columns individually, NOT only by their
```
Bouncner, line 275: This individually sorted is pretty strange.

Bouncner: Who came up with this?

(don't blame here please)

Bouncner: Maybe we should add a comment here.

This thing came with the sort merge join. We wanted to keep the information, that the table is really sorted individually (both join columns). But this thing is something very different from typical sort information (as we use it here for DYOD).

src/test/lib/logical_query_plan/sort_node_test.cpp
```diff
@@ -56,13 +55,14 @@ TEST_F(SortNodeTest, HashingAndEqualityCheck) {
...
       SortNode::make(expression_vector(_a_i), std::vector<SortMode>{SortMode::AscendingNullsFirst}, _table_node);
+  const auto sort_d =
+      SortNode::make(expression_vector(_a_i), std::vector<SortMode>{SortMode::AscendingNullsLast}, _table_node);
 
   EXPECT_NE(*_sort_node, *sort_a);
   EXPECT_NE(*_sort_node, *sort_b);
   EXPECT_EQ(*_sort_node, *sort_c);
+  EXPECT_NE(*_sort_node, *sort_d);
 
-  EXPECT_NE(_sort_node->hash(), sort_a->hash());
```
Bouncner, line 57: Why removed?

src/lib/storage/chunk.cpp
Bouncner: In the long run, would we need something like a vector of vectors of column IDs?

src/test/lib/logical_query_plan/sort_node_test.cpp
dey4ss: We agreed some time ago that expecting there are no hash collisions can break too easily. Hashes of equal nodes and equality checks should of course return the correct results.

src/lib/storage/chunk.cpp
dey4ss: In the long run, we might want to use `sorted_by` instead of `individually_sorted_by` and keep track of order dependencies (e.g., the TPC-DS `d_date` key orders each of the day, month, quarter, year columns). I would not consider anything of this for now though.

Bouncner (approved):
I’d add a comment to the chunk marking. Apart from that, ready to go.

dey4ss:
> I’d add a comment to the chunk marking. Apart from that, ready to go.

How should that look like? Something in the chunk cpp or hpp that says: ``TODO(anyone): In the future, we might want to change the semantics to `sorted_by` instead of `individually_sorted_by` and keep track of order dependencies.``?

PR #2719, opened 2025-08-02
[DYOD 2025] Group 1

gwauge:
This is our implementation of the sort operator for Hyrise as part of the course DYOD in the summer term 2025.


luisegarberding:
Dear Reviewers, 
please only review our changes in `src/lib/operators/sort.cpp` and `src/lib/operators/sort.hpp`.
We need the other changes at the moment for perfetto profiling, but these changes will not be merged into Hyrise. 
Thank you!

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+bool is_descending(const SortMode& mode) {
+  return mode == SortMode::DescendingNullsFirst || mode == SortMode::DescendingNullsLast;
+}
+
+bool is_nulls_first(const SortMode& mode) {
+  return mode == SortMode::AscendingNullsFirst || mode == SortMode::DescendingNullsFirst;
+}
+
+inline void encode_string(uint8_t* dest, const size_t data_length, const pmr_string& value) {
+  auto string_len = value.size();
```
Stastez:
```suggestion
  const auto string_len = value.size();
```

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+  memset(dest + 1, 0, data_length);  // set all bytes to 0
+  // copy the string data into the key buffer
+  memcpy(dest + 1, value.data(), string_len);                                //NOLINT
+  memset(dest + 1 + data_length - 2, static_cast<uint16_t>(string_len), 2);  // store actual string length
```
Stastez: Casting to uint16_t is misleading. According to the standard (https://en.cppreference.com/w/c/string/byte/memset.html), the value given here will internally be cast to unsigned char

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+}
+
+inline void encode_double(uint8_t* dest, const double value) {
+  // Encode double value; reinterpret double as raw 64-bit bits
+  auto bits = uint64_t{0};
+  memcpy(&bits, &value, sizeof(bits));
```
Stastez:
```suggestion
  auto bits = std::bit_cast<uint64_t>(value);
```

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+  if (std::signbit(value)) {
+    bits = ~bits;  // Negative values are bitwise inverted
+  } else {
+    bits ^= 0x8000000000000000ULL;  // Flip the sign bit for positive values
+  }
+
+  // Write to buffer in big-endian order (MSB first)
+  for (auto byte_idx = uint32_t{0}; byte_idx < 8; ++byte_idx) {
+    dest[1 + byte_idx] = static_cast<uint8_t>(bits >> ((7 - byte_idx) * 8));
+  }
```
Stastez: This logic is repeated and could be extracted into a separate function

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+}
+
+inline void encode_float(uint8_t* dest, const float value) {
+  auto bits = uint32_t{0};
+  memcpy(&bits, &value, sizeof(bits));
```
Stastez:
```suggestion
  auto bits = std::bit_cast<uint32_t>(value);
```

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+      high = cut_left;
+    }
+  }
+  return {low, diag - low};
+}
+
+// Parallel merge of two consecutive runs using Merge Path.
+template <typename Compare>
+void merge_path_parallel(RowIDPosList& rows, const size_t start, const size_t mid, const size_t end, Compare comp,
+                         size_t max_workers) {
```
Stastez:
```suggestion
                         const size_t max_workers) {
```

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+  const auto len_left = mid - start;
+  const auto len_right = end - mid;
+  const auto total_len = len_left + len_right;
+
+  const auto workers = static_cast<size_t>(std::min(max_workers, total_len));
+
+  auto dest = RowIDPosList{};
+  dest.resize(total_len);
```
Stastez:
```suggestion
  auto dest = RowIDPosList(total_len);
```

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+    }));
+  }
+  Hyrise::get().scheduler()->schedule_and_wait_for_tasks(jobs);
+  const auto start_offset = static_cast<std::ptrdiff_t>(start);
+  std::move(dest.begin(), dest.end(), rows.begin() + start_offset);
+}
+
+template <typename Compare>
+void parallel_sort_rowids(RowIDPosList& rows, Compare comp) {
+  auto row_count = rows.size();
```
Stastez:
```suggestion
  const auto row_count = rows.size();
```

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+  auto is_multithreaded = Hyrise::get().is_multi_threaded();
```
Stastez:
```suggestion
  const auto is_multithreaded = Hyrise::get().is_multi_threaded();
```

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+  if (!HYRISE_DEBUG && (!is_multithreaded || row_count < SMALL_ARRAY_THRESHOLD)) {
+    boost::sort::pdqsort(rows.begin(), rows.end(), comp);
+    return;
+  }
+
+  // 1) Get number of workers and block size.
+  // Default to single-threaded execution
+  auto num_workers = size_t{1};
+
+  auto nq_scheduler = std::dynamic_pointer_cast<NodeQueueScheduler>(Hyrise::get().scheduler());
```
Stastez:
```suggestion
  const auto nq_scheduler = std::dynamic_pointer_cast<NodeQueueScheduler>(Hyrise::get().scheduler());
```

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+  field_width.reserve(sort_definitions_size);
+
+  for (const auto& def : _sort_definitions) {
+    const auto sort_col = def.column;
+    resolve_data_type(input_table->column_data_type(sort_col), [&](auto type) {
       using ColumnDataType = typename decltype(type)::type;
+      if constexpr (std::is_same_v<ColumnDataType, pmr_string>) {
+        // Iterate over all chunks to find the longest string in the column.
+        auto max_string_length = size_t{0};
+        for (ChunkID chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
```
Stastez:
```suggestion
        for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
```

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+            if (!val.is_null()) {
+              const auto string_length = val.value().size();
+              if (string_length > max_string_length) {
+                max_string_length = string_length;
+              }
+            }
+          });
+        }
+        // Store size of the string + 2 for string length.
+        field_width.emplace_back(max_string_length + 2);
```
Stastez: Why +2?

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+    });
+  }
 
-      auto sort_impl = SortImpl<ColumnDataType>(input_table, sort_definition.column, sort_definition.sort_mode);
-      previously_sorted_pos_list = sort_impl.sort(previously_sorted_pos_list);
+  // Total width of each normalized key is the width of all columns to be sorted by plus null bytes.
+  auto key_width = size_t{0};
+  for (const auto& column : field_width) {
+    key_width += column + 1;
+  }
```
Stastez:
```suggestion
  auto key_width = std::reduce(field_width.begin(), field_width.end(), size_t{0});
  key_width += field_width.size();
```

Extracting the n additions into one addition should be faster. Given this, it would be possible to change the for-loop into `std::reduce`, depending on preference.

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
-      total_sort_time += sort_impl.sort_time;
-    });
+  /**
+   * Offsets for each column in the key, i.e. `key_offsets[i]` is the offset of the i-th column in the key.
+   * This means, that `buffer[key_offsets[i]]` is the location of the i-th column's value in the key.
+  */
+  auto key_offsets = std::vector<size_t>(sort_definitions_size);
+  key_offsets[0] = 0;
+  for (auto index = size_t{1}; index < sort_definitions_size; ++index) {
+    key_offsets[index] = key_offsets[index - 1] + field_width[index - 1] + 1;
```
Stastez: It could be helpful to remind the reader that `+1` is because of the null byte.

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+  }
```
Stastez: Also, since this is a prefix sum, you could use:
```suggestion
  std::transform_exclusive_scan(field_width.begin(), field_width.end(), key_offsets.begin(), size_t{0}, std::plus<>{}, [](const auto value) { return value + 1; });
```

I believe that would communicate the intent better.

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+    auto chunk = input_table->get_chunk(chunk_id);
+    auto row_count_for_chunk = chunk->size();
+
+    chunk_sizes[chunk_id] = row_count_for_chunk;
+    // Total size of the keys for all chunks.
+    total_buffer_size += row_count_for_chunk * key_width;
+
+    // Offset for the next chunk.
+    if (chunk_id > 0) {
+      row_id_offsets.emplace_back(row_id_offsets[chunk_id - 1] + chunk_sizes[chunk_id - 1]);
```
Stastez: This could also be extracted into a `std::exclusive_scan`.

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+  auto preparation_time = timer.lap();
+
+  /**************************************************************************************************************
+   *********************************************** Key-Generation ***********************************************
+   **************************************************************************************************************/
+
+  // Job queue for generating keys in parallel.
+  auto keygen_jobs = std::vector<std::shared_ptr<AbstractTask>>{};
+  keygen_jobs.reserve(static_cast<size_t>(chunk_count) * sort_definitions_size);
+  for (ChunkID chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
```
Stastez:
```suggestion
  for (auto chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
```

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+    auto chunk = input_table->get_chunk(chunk_id);
+
+    for (auto index = size_t{0}; index < sort_definitions_size; ++index) {
+      keygen_jobs.emplace_back(std::make_shared<JobTask>([&, chunk_id, index, chunk = chunk]() {
```
Stastez:
```suggestion
      keygen_jobs.emplace_back(std::make_shared<JobTask>([&, chunk_id, index, chunk]() {
```

src/lib/operators/sort.cpp
```diff
@@ -54,95 +260,77 @@ std::shared_ptr<Table> write_materialized_output_table(const std::shared_ptr<con
...
-          value_segment = std::make_shared<ValueSegment<ColumnDataType>>(std::move(value_segment_value_vector),
-                                                                         std::move(value_segment_null_vector));
-        } else {
-          value_segment = std::make_shared<ValueSegment<ColumnDataType>>(std::move(value_segment_value_vector));
-        }
-        chunk_it->push_back(value_segment);
+      for (auto output_chunk_id = ChunkID{0}; output_chunk_id < output_chunk_count; ++output_chunk_id) {
+        jobs.emplace_back(
+            std::make_shared<JobTask>([&, output_chunk_id, column_id, column_is_nullable = column_is_nullable,
+                                       accessor_by_chunk_id = accessor_by_chunk_id]() {
```
Stastez:
```suggestion
            std::make_shared<JobTask>([&, output_chunk_id, column_id, column_is_nullable,
                                       accessor_by_chunk_id]() {
```

src/lib/operators/sort.cpp
```diff
@@ -202,46 +387,49 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
-                      "Input column references more than one column");
-          const auto& input_reference_pos_list = input_reference_segment.pos_list();
-          output_pos_list->emplace_back((*input_reference_pos_list)[row_id.chunk_offset]);
-        } else {
-          output_pos_list->emplace_back(row_id);
-        }
+      for (auto output_chunk_id = ChunkID{0}; output_chunk_id < output_chunk_count; ++output_chunk_id) {
+        jobs.emplace_back(std::make_shared<JobTask>([&, output_chunk_id, column_id, input_segments = input_segments,
+                                                     referenced_table = referenced_table,
+                                                     referenced_column_id = referenced_column_id]() {
```
Stastez:
```suggestion
        jobs.emplace_back(std::make_shared<JobTask>([&, output_chunk_id, column_id, input_segments,
                                                     referenced_table,
                                                     referenced_column_id]() {
```

src/lib/operators/sort.cpp
```diff
@@ -31,6 +38,7 @@
 
 namespace {
 
+constexpr size_t SMALL_ARRAY_THRESHOLD = 10'000;
```
beerjamin: Out of curiosity, how did you come up with this number?

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+
+bool is_nulls_first(const SortMode& mode) {
+  return mode == SortMode::AscendingNullsFirst || mode == SortMode::DescendingNullsFirst;
+}
+
+inline void encode_string(uint8_t* dest, const size_t data_length, const pmr_string& value) {
+  auto string_len = value.size();
+  memset(dest + 1, 0, data_length);  // set all bytes to 0
+  // copy the string data into the key buffer
+  memcpy(dest + 1, value.data(), string_len);                                //NOLINT
```
beerjamin: What linter rule is this line actually breaking? Is it necessary to resolve it?

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+      static_cast<UnsignedT>(value) ^ (static_cast<UnsignedT>(1) << ((data_length * 8) - 1));  // flip sign bit
+
+  // Store bytes in big-endian order starting at dest[1]
+  for (auto byte_idx = size_t{0}; byte_idx < data_length; ++byte_idx) {
+    dest[1 + byte_idx] = static_cast<uint8_t>(biased >> ((data_length - 1 - byte_idx) * 8));
+  }
+}
+
+template void encode_integer<int32_t>(uint8_t* dest, const int32_t value, const size_t data_length);
+template void encode_integer<int64_t>(uint8_t* dest, const int64_t value, const size_t data_length);
```
beerjamin: I really like how you have implemented the normalization, but I would suggest to put this in a separate utility class. This will make it much easier to test and maintain. Since these are not part of a class, they are difficult to test in isolation without setting up the entire Sort operator context.  Testing normalization should not depend on sorting.

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+      std::merge(l_begin, l_end, r_begin, r_end, out, comp);
+    }));
+  }
+  Hyrise::get().scheduler()->schedule_and_wait_for_tasks(jobs);
+  const auto start_offset = static_cast<std::ptrdiff_t>(start);
+  std::move(dest.begin(), dest.end(), rows.begin() + start_offset);
+}
+
+template <typename Compare>
+void parallel_sort_rowids(RowIDPosList& rows, Compare comp) {
```
beerjamin: I think this is also really nicely done. My suggestion would be to abstract the entire parallel sorting process into a separate class or a set of functions that are not tied to the Sort operator's internal state. This would allow for easy replacement of the sorting algorithm (e.g., swapping out the current merge sort with a different implementation like a parallel radix sort) without modifying the Sort operator itself. This is only a suggestion so please resolve if you dont agree.

src/lib/operators/sort.cpp
beerjamin: It may be that strings longer than data_length are silently truncated. string_len is derived from value.size(), which can be larger than data_length - 2. If string_len is greater than data_length - 2, this memcpy will write past the end of the destination buffer, leading to a buffer overflow.

src/lib/operators/sort.cpp
beerjamin: For very large tables this could be O(n log n) extra memory usage. I did not measure and this is just an assumption.

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+  auto row_count = rows.size();
+  auto is_multithreaded = Hyrise::get().is_multi_threaded();
+
+  if (!HYRISE_DEBUG && (!is_multithreaded || row_count < SMALL_ARRAY_THRESHOLD)) {
+    boost::sort::pdqsort(rows.begin(), rows.end(), comp);
+    return;
+  }
```
beerjamin: I think this code block can be removed. The block creates a divergence between debug and release builds. I dont think its worth it to have a divergence for small tasks. This also goes with the comment above, where did the SMALL_ARRAY_THRESHOLD come from.

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+
+  auto key_buffer = std::vector<uint8_t>();
+  auto total_buffer_size = size_t{0};
+
+  // Number of rows per chunk in the input table.
+  auto chunk_sizes = std::vector<size_t>(chunk_count);
+
+  auto row_ids = RowIDPosList{};
+  row_ids.reserve(row_count);
```
beerjamin: The reserve() method in only allocates memory, it does not change the size of the container. The size() of the key_buffer remains 0. When the parallel jobs attempt to write data to &key_buffer[(...)], they are writing to uninitialized memory outside the valid bounds of the vector, this will likely be undefined behavior. You should resize(total_buffer_size) before parallel keygen.

src/lib/operators/sort.cpp
Bouncner: Can you please add a comment explaining why you this value and how you determined it?

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
 
+bool is_descending(const SortMode& mode) {
+  return mode == SortMode::DescendingNullsFirst || mode == SortMode::DescendingNullsLast;
+}
+
+bool is_nulls_first(const SortMode& mode) {
+  return mode == SortMode::AscendingNullsFirst || mode == SortMode::DescendingNullsFirst;
+}
+
+inline void encode_string(uint8_t* dest, const size_t data_length, const pmr_string& value) {
```
Bouncner: Instead of using `dest + 1` everywhere, why not pass `dest + 1` to the function?

Bouncner: `uint8` type is misleading.

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+    chunk_sizes[chunk_id] = row_count_for_chunk;
+    // Total size of the keys for all chunks.
+    total_buffer_size += row_count_for_chunk * key_width;
+
+    // Offset for the next chunk.
+    if (chunk_id > 0) {
+      row_id_offsets.emplace_back(row_id_offsets[chunk_id - 1] + chunk_sizes[chunk_id - 1]);
+    }
+
+    for (ChunkOffset row = ChunkOffset{0}; row < row_count_for_chunk; ++row) {
```
Bouncner: auto

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
   return (lhs + rhs - 1u) / rhs;
 }
 
+bool is_descending(const SortMode& mode) {
+  return mode == SortMode::DescendingNullsFirst || mode == SortMode::DescendingNullsLast;
+}
+
+bool is_nulls_first(const SortMode& mode) {
```
Bouncner: The function is okay, but it's a one liner that exactly used once.

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+  return mode == SortMode::AscendingNullsFirst || mode == SortMode::DescendingNullsFirst;
+}
+
+inline void encode_string(uint8_t* dest, const size_t data_length, const pmr_string& value) {
+  auto string_len = value.size();
+  memset(dest + 1, 0, data_length);  // set all bytes to 0
```
Bouncner: Style guide: Start comments with upper case, end with period.

src/lib/operators/sort.cpp
Bouncner: I think a `DebugAssert` would be helpful here as it checks your assumption/code that the field length is sized to the max string length.
If you change something to the string handling, this function might silently cause very strange issues as @beerjamin  noted.

src/lib/operators/sort.cpp
Bouncner: Is the string length used somewhere? You're storing the entire string, do you need the length?

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+  for (auto byte_idx = uint32_t{0}; byte_idx < 4; ++byte_idx) {
+    dest[1 + byte_idx] = static_cast<uint8_t>(bits >> ((3 - byte_idx) * 8));
+  }
+}
+
+template <typename T>
+inline void encode_integer(uint8_t* dest, const T value, const size_t data_length) {
+  // Bias the value to get a lexicographically sortable encoding
+  using UnsignedT = std::make_unsigned_t<T>;
+  UnsignedT const biased =
```
Bouncner: `const auto`?

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+      static_cast<UnsignedT>(value) ^ (static_cast<UnsignedT>(1) << ((data_length * 8) - 1));  // flip sign bit
+
+  // Store bytes in big-endian order starting at dest[1]
```
Bouncner: Please briefly mention once why we care about big/little endianess here? It is the only part in Hyrise, where we care.

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+    dest[1 + byte_idx] = static_cast<uint8_t>(biased >> ((data_length - 1 - byte_idx) * 8));
+  }
+}
+
+template void encode_integer<int32_t>(uint8_t* dest, const int32_t value, const size_t data_length);
+template void encode_integer<int64_t>(uint8_t* dest, const int64_t value, const size_t data_length);
+
+// Finds the cut point for the merge path diagonal inspired by https://arxiv.org/pdf/1406.2628.
+template <typename T, typename Compare>
+inline std::pair<size_t, size_t> find_cut_point(const T* const left, const size_t len_left, const T* const right,
```
Bouncner: Do you use `inline` for performance purposes? That should usually be done by the compiler.

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+  // Store bytes in big-endian order starting at dest[1]
+  for (auto byte_idx = size_t{0}; byte_idx < data_length; ++byte_idx) {
+    dest[1 + byte_idx] = static_cast<uint8_t>(biased >> ((data_length - 1 - byte_idx) * 8));
+  }
+}
+
+template void encode_integer<int32_t>(uint8_t* dest, const int32_t value, const size_t data_length);
+template void encode_integer<int64_t>(uint8_t* dest, const int64_t value, const size_t data_length);
+
+// Finds the cut point for the merge path diagonal inspired by https://arxiv.org/pdf/1406.2628.
```
Bouncner: Please add a brief (3-5 sentences) explanation why you use merge path (binary merge, multiple threads, ...).

src/lib/operators/sort.cpp
Bouncner: The struct `Cut` is nothing else than `std::pair<size_t, size_t>`. Why not use Cut here? Or the pair everywhere else?

src/lib/operators/sort.cpp
```diff
@@ -39,6 +47,204 @@ size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
...
+
+  // 3) Bottom-up merge sorted runs, doubling the run size each pass:
+  auto run = block;
+  while (run < row_count) {
+    jobs.clear();
+    for (auto left = size_t{0}; left + run < row_count; left += 2 * run) {
+      auto mid = left + run;
+      auto right = std::min(left + (2 * run), row_count);
+
+      merge_path_parallel(rows, left, mid, right, comp, num_workers);
```
Bouncner: Performance-wise, spawning 32 jobs (let's say we have 32 workers) for parallel merging of the first levels is probably not very efficient. For the last phases, it makes a lot of sense though.
You could check the level or the size and only use parallel mergen for large merges, std::merge otherwise.

src/lib/operators/sort.cpp
```diff
@@ -54,95 +260,77 @@ std::shared_ptr<Table> write_materialized_output_table(const std::shared_ptr<con
...
+              if (column_is_nullable) {
+                value_segment = std::make_shared<ValueSegment<ColumnDataType>>(std::move(value_segment_value_vector),
+                                                                               std::move(value_segment_null_vector));
+              } else {
+                value_segment = std::make_shared<ValueSegment<ColumnDataType>>(std::move(value_segment_value_vector));
+              }
+
+              output_segments_by_chunk[output_chunk_id][column_id] = value_segment;
+            }));
+        jobs.back()->schedule();  // schedule immediately because job creation is somewhat expensive
```
Bouncner: `schedule_and_wait_for_tasks` is preferable here as task creation is cheap and the scheduler cannot group already-scheduled tasks.

src/lib/operators/sort.cpp
```diff
@@ -54,95 +260,77 @@ std::shared_ptr<Table> write_materialized_output_table(const std::shared_ptr<con
...
+                  static_cast<ChunkOffset>(row_count - (static_cast<uint64_t>(output_chunk_id * output_chunk_size))));
+
+              value_segment_value_vector.reserve(chunk_size);
+              if (column_is_nullable) {
+                value_segment_null_vector.reserve(chunk_size);
+              }
+
+              for (auto row_index = size_t{0}; row_index < chunk_size; ++row_index) {
+                const auto [chunk_id, chunk_offset] =
+                    pos_list[(static_cast<size_t>(output_chunk_size * output_chunk_id)) + row_index];
```
Bouncner: Pull `static_cast<size_t>(output_chunk_size * output_chunk_id)` out of the loop?

src/lib/operators/sort.cpp
```diff
@@ -202,46 +387,49 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
-          write_output_pos_list();
-        }
-      }
-      if (!output_pos_list->empty()) {
-        write_output_pos_list();
+          DebugAssert(!output_pos_list->empty(), "Asked to write empty output_pos_list");
+          output_segments_by_chunk.at(output_chunk_id)[column_id] =
+              std::make_shared<ReferenceSegment>(referenced_table, referenced_column_id, output_pos_list);
+        }));
+        jobs.back()->schedule();  // schedule job immediately
```
Bouncner: See above.

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
-  for (auto sort_step = static_cast<int64_t>(_sort_definitions.size() - 1); sort_step >= 0; --sort_step) {
-    const auto& sort_definition = _sort_definitions[sort_step];
-    const auto data_type = input_table->column_data_type(sort_definition.column);
+  /**************************************************************************************************************
+   ***************************************** Pre-compute offsets and other info *********************************
+   **************************************************************************************************************/
 
-    resolve_data_type(data_type, [&](auto type) {
+  // The key length is calculated based on the sizes of the columns to be sorted by,
+  // e.g. if sorting by int, string it should be [4, 8].
```
Bouncner: 8 is not the size of a string (assuming you mean `sizeof(pmr::string)`). Probably an outdated comment given your string handling?

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
       using ColumnDataType = typename decltype(type)::type;
+      if constexpr (std::is_same_v<ColumnDataType, pmr_string>) {
+        // Iterate over all chunks to find the longest string in the column.
+        auto max_string_length = size_t{0};
+        for (ChunkID chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+          const auto abstract_segment = input_table->get_chunk(chunk_id)->get_segment(sort_col);
+          segment_iterate<ColumnDataType>(*abstract_segment, [&](const auto& val) {
+            if (!val.is_null()) {
+              const auto string_length = val.value().size();
+              if (string_length > max_string_length) {
```
Bouncner: std::max?

src/lib/operators/sort.cpp
Bouncner: As you materialize string, you should definitely consider parallelizing this loop. Should be trivial.

What would have been even better: keep the strings around. Materializing strings is expensive, but you currently throw them away.

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+
+            // Pointer to the destination in the key buffer for this column.
+            auto* dest = &key_buffer[(row_id_offsets[chunk_id] + row) * key_width + key_offsets[index]];
+
+            const auto is_not_null = !val.is_null();
+            const ColumnDataType value = is_not_null ? val.value() : ColumnDataType{};
+            const auto data_length = field_width[index];
+
+            // Set the first byte to indicate if the value is null or not.
+            auto null_byte = is_not_null ? 0x00U : 0xFFU;
```
Bouncner: Unneccessary verbose for my taste.

Bouncner: Why not directly create `uint8_t`'s?

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+            }
+            dest[0] = static_cast<uint8_t>(null_byte);
+
+            // Encode the value into the key based on the data type.
+            if constexpr (std::is_same_v<ColumnDataType, pmr_string>) {
+              encode_string(dest, data_length, value);
+            } else if constexpr (std::is_same_v<ColumnDataType, double>) {
+              encode_double(dest, value);
+            } else if constexpr (std::is_same_v<ColumnDataType, float>) {
+              encode_float(dest, value);
```
Bouncner: As said above, but `dest` (destination) is not really the destination here as the function all use `dest + 1`. Why not pass `dest + 1`?

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+            } else if constexpr (std::is_integral<ColumnDataType>::value && std::is_signed<ColumnDataType>::value) {
+              encode_integer<ColumnDataType>(dest, value, data_length);
+            } else {
+              const auto type_name = std::string{typeid(ColumnDataType).name()};
+              throw std::logic_error("Unsupported data type for sorting: " + type_name);
```
Bouncner: Assert!

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+
+  /**************************************************************************************************************
+   *********************************************** Key-Generation ***********************************************
+   **************************************************************************************************************/
+
+  // Job queue for generating keys in parallel.
+  auto keygen_jobs = std::vector<std::shared_ptr<AbstractTask>>{};
+  keygen_jobs.reserve(static_cast<size_t>(chunk_count) * sort_definitions_size);
+  for (ChunkID chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
+    auto chunk = input_table->get_chunk(chunk_id);
```
Bouncner: Pull chunk into job.

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+  auto chunk_sizes = std::vector<size_t>(chunk_count);
+
+  auto row_ids = RowIDPosList{};
+  row_ids.reserve(row_count);
+
+  auto row_id_offsets = std::vector<size_t>();
+  row_id_offsets.reserve(chunk_count);
+  row_id_offsets.emplace_back(0);
+
+  for (ChunkID chunk_id = ChunkID{0}; chunk_id < chunk_count; ++chunk_id) {
```
Bouncner: auto

Bouncner: What about parallelizing this loop?

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+              encode_float(dest, value);
+            } else if constexpr (std::is_integral<ColumnDataType>::value && std::is_signed<ColumnDataType>::value) {
+              encode_integer<ColumnDataType>(dest, value, data_length);
+            } else {
+              const auto type_name = std::string{typeid(ColumnDataType).name()};
+              throw std::logic_error("Unsupported data type for sorting: " + type_name);
+            }
+
+            // Invert for descending order (excluding the null byte).
+            if (descending) {
```
Bouncner, line 450: Not need to implement, but instead of writing the values and inverting them, one could also do the invertion right away.

src/lib/operators/sort.cpp
```diff
@@ -301,44 +491,197 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+      row_id_offsets.emplace_back(row_id_offsets[chunk_id - 1] + chunk_sizes[chunk_id - 1]);
+    }
+
+    for (ChunkOffset row = ChunkOffset{0}; row < row_count_for_chunk; ++row) {
+      row_ids.emplace_back(chunk_id, row);
+    }
+  }
+
+  key_buffer.reserve(total_buffer_size);
+  auto preparation_time = timer.lap();
```
Bouncner: const

src/lib/operators/sort.cpp
EnnioEnnio: `+ 1` changed, type unchanged, WIP

src/lib/operators/sort.cpp
EnnioEnnio: Afaik it's just a hint and the compiler does its own work anyways. I used it to transport the intend to the reader that it's a helper that should be inlined to achieve best performance. 
Could remove it - up to you

src/lib/operators/sort.cpp
EnnioEnnio: `auto` done - rest to be discussed with group

src/lib/operators/sort.cpp
gwauge: It is still required in case a null byte is included at the end of the string. In that case, we cannot differentiate from padding (both are null bytes). I added a comment for explanation.

src/lib/operators/sort.cpp
gwauge: We can't parallelize here, because the calculation of `row_id_offsets` depends on previous values, so sequential iteration is required. 
Since the number of chunks tends to be manageable, we think this is fine being left unparallelized. Heat maps from perfetto supported this hypothesis.

src/lib/operators/sort.cpp
gwauge: `row_ids` holds `RowID` elements, which are added to the vector below using `emplace_back` in the typical C++ manner. since `total_buffer_size` holds the sum of the number of bytes for all keys, this wouldn't apply to `row_ids` anyway. Maybe I am misunderstanding something but I think this comment can be resolved as it is incorrect.

src/lib/operators/sort.cpp
Bouncner: + auto

src/lib/operators/sort.cpp
Bouncner: Got it.

src/lib/operators/sort.cpp
Bouncner: Unresolving here.

I think @beerjamin refers to line 587 in the current sort.cpp. The current code still looks like UB to me.
You could use the `uninitialized_vector` that we use elsewhere or simply `resize()` (which is not for free).

src/lib/operators/sort.cpp
gwauge: I see your point. However, the teaching team seems to not mind diverging code paths (see [here](https://github.com/hyrise/hyrise/pull/2719#discussion_r2324802760)), so we will leave it in. I admit that the value for `SMALL_ARRAY_THRESHOLD` was chosen arbitrarily and should probably be determined empirically.

src/lib/operators/sort.cpp
gwauge: Good idea! Since we would have to restructure the entire loop and likely split it up, we decided against using it here.

src/lib/operators/sort.cpp
gwauge: I replaced the magic number with a variable. Hopefully, this clears up confusion.

Bouncner:
Thanks for participating in DYOD! We hope you've enjoyed the seminar. :)

PR #2720, opened 2025-08-03
Improve Sort Operator (DYOD 2025)

robert-oleynik:


src/lib/operators/sort.cpp
```diff
@@ -33,10 +53,54 @@ namespace {
...
+  return (left + right - 1) / right;
+}
+
+// Returns the next multiple. If value is already a multiple, then return value.
+auto next_multiple(const auto value, const auto multiple) {
+  return div_ceil(value, multiple) * multiple;
+}
+
+size_t opt_batch_size(size_t count, size_t min_batch_size, size_t max_tasks_per_thread) {
+  const auto hardware_concurrency = std::thread::hardware_concurrency();
```
Noel-Bastubbe: I think this should be based on the number of workers the hyrise scheduler has (https://moodle.hpi.de/mod/forum/discuss.php?d=9586)

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+
+// Append segement's values as byte array to the normalized keys. Expects that the normalized_key_iter is valid for
+// each segment's values.
+//
+// Parameters:
+// @param segment              Segment to encode values from
+// @param offset               Offset to the start of the normalized key row's byte array.
+// @param expected_width       The number of bytes all values must be encoded to.
+// @param ascending            Sort normalized keys in ascending order.
+// @param nullable             Put exrta byte in front to encode null values.
```
Noel-Bastubbe: These comments are very helpful. Here is a typo in "extra".

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+        unsigned_value = byteswap(unsigned_value);
+        unsigned_value ^= static_cast<UnsignedColumnDataType>(full_modifier);
+        memcpy(normalized_key_start, &unsigned_value, sizeof(UnsignedColumnDataType));
+        normalized_key_start += sizeof(UnsignedColumnDataType);
+      } else {
+        Fail(std::format("Cannot encode values of type '{}'", typeid(ColumnDataType).name()));
       }
-      if (!output_pos_list->empty()) {
-        write_output_pos_list();
+      if constexpr (HYRISE_DEBUG) {
```
Noel-Bastubbe: This if condition is not needed for DebugAssert, right?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+/*
+ * Classify all elements of stripe_range into buckets. Each bucket holds at most block size many elements. A full
+ * bucket is written back to the start of the stripe range.
+ */
+StripeClassificationResult classify_stripe(NormalizedKeyRange auto& stripe_range,
+                                           const NormalizedKeyRange auto& classifiers, const size_t block_size,
+                                           const NormalizedKeyComparator auto& comp) {
+  const auto num_buckets = std::ranges::size(classifiers) + 1;
+  DebugAssert(num_buckets > 0, "At least one bucket is required");
+  DebugAssert(block_size > 0, "Invalid bock size");
```
Noel-Bastubbe: Type block size

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+
+  // Find last stripe a bucket is in.
+  auto last_stripe = std::ranges::lower_bound(  // NOLINT
+      stripe_ranges, last_bucket_end, std::less<int64_t>(), [&](const auto& stripe_range) {
+        return std::distance(sort_range.begin(), stripe_range.end());
+      });
+  if (last_stripe == stripe_ranges.end()) {
+    last_stripe = --stripe_ranges.end();
+  }
+  DebugAssert(last_stripe != stripe_ranges.end(), "Bucket should end in last stripe");
```
Noel-Bastubbe: This DebugAssert in combination with the if-condition before looks unintuitive. How can last_strip and stripe_ranges.end() be equal at that point?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+
+  // Move blocks into the correct bucket (based on the aligned bucket delimiter).
+  const auto permute_blocks_tasks = run_parallel_batched(num_stripes, 1, [&](auto stripe) {
+    permute_blocks(classifiers, stripe, bucket_iterators, comp, num_buckets, block_size, overflow_bucket,
+                   overflow_bucket_begin);
+  });
+  Hyrise::get().scheduler()->wait_for_tasks(permute_blocks_tasks);
+
+  // Until this point all buckets are aligned to the block_size and there still some classified elements in the stripe
+  // local buffer left. Because of that, we know move elements overflowing the bucket to the start of that bucket and
```
Noel-Bastubbe: typo "we now move"

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+
+  /// Checks if some reads are pending.
+  bool pending_reads() {
+    return _pending_reads.load(std::memory_order_relaxed) != 0;
+  }
+
+  /// Decrement the number of pending reads.
+  void complete_read() {
+    const auto pending_reads = _pending_reads.fetch_sub(1, std::memory_order_relaxed);
+    if constexpr (HYRISE_DEBUG) {
```
Noel-Bastubbe: Why not use DebugAssert here?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+      if constexpr (HYRISE_DEBUG) {
+        DebugAssert(normalized_key_start == normalized_key_iter->key_head + offset + column_info.width(),
+                    "Encoded unexpected number of bytes");
       }
+      ++it, ++normalized_key_iter;
+    }
+  });
+}
+
+namespace ips4o {
```
Noel-Bastubbe: The ips4o implementation is very long. Maybe it should get its own file.

Noel-Bastubbe (commented):
The IPS4O algorithm is hard to follow, but you did a good job of structuring the code and adding useful comments. You followed the hyrise coding conventions very well.

CMakeLists.txt
```diff
@@ -20,6 +20,8 @@ function(compiler_not_supported message)
     endif()
 endfunction(compiler_not_supported)
 
+option(ENABLE_PERFETTO "Add perfetto to the dependencies. Default: ON" OFF)
```
Bouncner: If you have a clean way of integrating perfetto, we would love to see a separate PR for it. :)

src/benchmark/benchmark_playground.cpp
```diff
@@ -2,99 +2,65 @@
 
 #include "benchmark/benchmark.h"
 
+#include "benchmark_config.hpp"
```
Bouncner: Revert file?

src/lib/operators/sort.cpp
Bouncner: For large servers (see the IBM Power8 machine in the lab), the number of CPU threads you obtain here and the REASONABLE number of threads to use in Hyrise, can differ significantly (we usually limit Hyrise to a single CPU on NUMA machines).
You might be overloading the scheduler here.

src/lib/operators/sort.cpp
```diff
@@ -33,10 +53,54 @@ namespace {
...
+auto div_ceil(const auto left, const auto right) {
+  return (left + right - 1) / right;
+}
+
+// Returns the next multiple. If value is already a multiple, then return value.
+auto next_multiple(const auto value, const auto multiple) {
+  return div_ceil(value, multiple) * multiple;
+}
+
+size_t opt_batch_size(size_t count, size_t min_batch_size, size_t max_tasks_per_thread) {
```
Bouncner: Please comment what this function does and why you need it.

Bouncner: And even though the values are passed by value, we still prefer to explicitly mark them as `const`.

src/lib/operators/sort.cpp
```diff
@@ -33,10 +53,54 @@ namespace {
...
+  const auto hardware_concurrency = std::thread::hardware_concurrency();
+  const auto max_count_per_thread = (count + hardware_concurrency - 1) / hardware_concurrency;
+  const auto max_count_per_batch = (max_count_per_thread + max_tasks_per_thread - 1) / max_tasks_per_thread;
+  return std::max(max_count_per_batch, min_batch_size);
+}
+
+std::vector<std::shared_ptr<AbstractTask>> run_parallel_batched(size_t count, size_t batch_size, auto handler) {
```
Bouncner: Looking at the parameter name and the code, I assume `function`, `functor` or something like that is a more suitable name than `handler`.

src/lib/operators/sort.cpp
```diff
@@ -33,10 +53,54 @@ namespace {
...
+  const auto task_count = (count + batch_size - 1) / batch_size;
```
Bouncner: You reimplement task grouping which is done by the scheduler and also should be done by the scheduler (it depends on the current load of the system).

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+}
+
+/// Allocate an large chunk of uninitialized memory. This class is faster than C++ std::vector with resize, as it does
+/// not initialize each entry.
+template <typename T>
+class UVector {
+ public:
+  explicit UVector(const size_t size)
+      : _allocator(),
+        _ptr(_allocator.allocate(size)),  // NOLINT
```
Bouncner: Please add what lint rule you are disabling here: `  // NOLINT(rule_xzy)`

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+    return _size;
+  }
+
+ private:
+  std::allocator<T> _allocator;
+  T* _ptr;  // NOLINT
+  size_t _size;
+};
+
+template <size_t start, size_t end>
```
Bouncner, line 256: Please add a comment what you are doing here.

1-2 sentences (+ maybe naming the DuckDB paper that talks a bit more about it)

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+  constexpr NormalizedKeyRow(std::byte* key_head, RowID row_id) noexcept : key_head(key_head), row_id(row_id) {}
+
+  std::byte* key_head = nullptr;
+  RowID row_id;
+
+  bool less_than(const NormalizedKeyRow& other, size_t expected_size) const {
+    if (expected_size == 0) {
+      return false;
+    }
+    DebugAssert(key_head, "NormalizedKey is not valid");
```
Bouncner: End the error message with a period.

src/lib/operators/sort.cpp
```diff
@@ -297,48 +1368,170 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+  auto sorted_table = std::shared_ptr<Table>{};
+
+  auto column_materializer =
+      pmr_vector<std::function<void(const AbstractSegment&, size_t, bool, ScanResult, bool, NormalizedKeyIter)>>();
+  column_materializer.reserve(_sort_definitions.size());
+  for (const auto& column_sort_definition : _sort_definitions) {
+    const auto column_data_type = input_table->column_data_type(column_sort_definition.column);
+    resolve_data_type(column_data_type, [&](auto type) {
+      using ColumnDataType = typename decltype(type)::type;
+      column_materializer.emplace_back(materialize_segment_as_normalized_keys<ColumnDataType>);
```
Bouncner: Why create functions upfront?

src/lib/operators/sort.cpp
```diff
@@ -297,48 +1368,170 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+    }
+  }
+
+  auto column_stats = std::vector<ScanResult>(search_column_count, ScanResult{});
+  auto normalized_key_size = size_t{0};
+  for (auto column_index = size_t{0}; column_index < search_column_count; ++column_index) {
+    const auto column_id = _sort_definitions[column_index].column;
+    auto aggregated_stats = std::accumulate(chunk_stats[column_index].begin(), chunk_stats[column_index].end(),
+                                            ScanResult{}, [](ScanResult result, ScanResult& chunk) {
+                                              return result.merge(chunk);
```
Bouncner, line 679: If the local string vectors would be sorted, you could linearly merge.

src/lib/operators/sort.cpp
```diff
@@ -297,48 +1368,170 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+    normalized_key_size += aggregated_stats.width();
+    column_stats[column_index] = std::move(aggregated_stats);
+  }
+
+  const auto sort_long_strings = run_parallel_batched(search_column_count, 1, [&](const auto column) {
+    boost::sort::pdqsort(column_stats[column].long_strings.begin(), column_stats[column].long_strings.end());
+  });
+  Hyrise::get().scheduler()->wait_for_tasks(sort_long_strings);
+
+  const auto padded_key_size = ((normalized_key_size + 3) / 4) * 4;
```
Bouncner, line 699: Comment why pad.

Bouncner: Have you measured the impact?
Curious as you are quite space-efficient when writing NULLs, but here you round up quite a bit?

src/lib/operators/sort.cpp
```diff
@@ -297,48 +1368,170 @@ std::shared_ptr<const Table> Sort::_on_execute() {
...
+    chunk_offsets[chunk_id] = total_offset;
+    total_offset += chunk_size;
+  }
+
+  auto chunk_allocations = std::vector<pmr_vector<std::byte>>(chunk_count);
+  const auto materialization_tasks = run_parallel_batched(chunk_count, 1, [&](const auto chunk_index) {
+    const auto chunk_id = static_cast<ChunkID>(chunk_index);
+    const auto chunk = input_table->get_chunk(chunk_id);
+    const auto chunk_size = chunk->size();
+    chunk_allocations[chunk_id] = pmr_vector<std::byte>(chunk_size * padded_key_size);
```
Bouncner: Couldn't you use UVector here instead of pmr_vector?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+        // the order is consistent after the conversion. Because of that, we will set the null value to half of the
+        // numeric space. While this can be done by setting the highest bit to 1 for all positive values, negative
+        // values are encoded as Two's complement and therefore need to be inverted. Because we use var int encoding,
+        // the highest bit is determined by expected encoding length.
+        //
+        // Examples for int32_t:
+        //
+        // Number | Hex Representation | Unsinged (full) Representation | Unsigned (variable) Representation
+        // -------|--------------------|--------------------------------|-----------------------------------
+        //   -42  | 0xFFFFFFd6         | 0x7FFFFFd6                     | 0x56
```
Bouncner, line 544: Nice visualization. It would be very helpful to get the same above for your string handling.

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+      DebugAssert(std::ranges::empty(bucket_empty_tail[bucket]),
+                  std::format("Bucket {} is not fully filled (tail)", bucket));
+    }
+  }
+
+  return aggregated_bucket_sizes;
+}
+
+/**
+ * This sorting algorithm consists of two sorting strategies:
```
Bouncner: Can you give a short overview of IPSO? Is it inplace? Is it sample-based? Concurrent?
Maybe reference to paper where this entire thing comes from?

Bouncner: strategies > passes ?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+ * 2. pdqsort to sort small enough buckets.
+ *
+ * A bucket is small enough, if it contains not enough elements to create at least two stripes (See config.min_batch_size).
+ */
+void sort(NormalizedKeyRange auto& sort_range, const Sort::Config& config, const NormalizedKeyComparator auto& comp) {
+  using RangeIterator = decltype(std::ranges::begin(sort_range));
+
+  const auto total_size = std::ranges::size(sort_range);
+  const auto total_num_blocks = div_ceil(total_size, config.block_size);
+  if (total_num_blocks < 2 * config.min_blocks_per_stripe) {
```
Bouncner: Above in the comment you say "buckets", here you use blocks and stripes.
Can you add a simple explanation (simple Ascii art?) to tell what those are?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+  auto classification_results = std::vector<StripeClassificationResult>(num_stripes);
+  const auto classification_tasks = run_parallel_batched(num_stripes, 1, [&](auto stripe) {
+    classification_results[stripe] = classify_stripe(stripes[stripe], classifiers, block_size, comp);
+  });
+  Hyrise::get().scheduler()->wait_for_tasks(classification_tasks);
+
+  // Prepare permuting classified blocks into the correct position. This preparation is done in the following:
+  // 1. We create the prefix sum of all buckets. As a result, we receive a list of bucket end indices.
+  // 2. We align these indices to the next block size.
+  // 3. We move all blocks so that the following conditions is met for all buckets: Full blocks are followed by empty
```
Bouncner: Are you talking about the "written back" blocks or the ones stored in classification results?

Why align? Can you please provide high level description, please?

src/test/lib/operators/sort_test.cpp
```diff
@@ -234,4 +250,131 @@ TEST_F(SortTest, InputReferencesDifferentColumns) {
...
+
+  auto last = 10000.0f;
+  for (const auto& row : output_table->get_rows()) {
+    EXPECT_EQ(row.size(), 1);
+
+    const auto current = boost::get<float>(row[0]);
+    EXPECT_GE(last, current);
+    last = current;
+  }
+  EXPECT_GT(last, -1000.0f);
```
Bouncner, line 310: Why not float-compare against 8466.5?

src/test/lib/operators/sort_test.cpp
```diff
@@ -234,4 +250,131 @@ TEST_F(SortTest, InputReferencesDifferentColumns) {
...
+
+TEST_F(SortTest, Ips4oFullConfig) {
+  auto config = Sort::Config();
+  config.block_size = 128;
+  config.max_parallelism = 8;
+  config.bucket_count = config.max_parallelism * 2;
+  config.samples_per_classifier = 4;
+  config.min_blocks_per_stripe = 16;
+
+  const auto count = 1000000;
```
Bouncner:
```suggestion
  const auto count = 1'000'000;
```

src/test/lib/operators/sort_test.cpp
```diff
@@ -234,4 +250,131 @@ TEST_F(SortTest, InputReferencesDifferentColumns) {
...
+    table->append({42});
+  }
+
+  const auto table_wrapper = std::make_shared<TableWrapper>(table);
+  table_wrapper->execute();
+  auto sort = Sort{
+      table_wrapper,
+      {SortColumnDefinition{ColumnID{0}, SortMode::AscendingNullsFirst}},
+  };
+  sort.execute();
```
Bouncner, line 327: You only test that it does not crash right now.

For this test and the one above, maybe add to table a second vector and have a helper function that compare the sort output with the second vector content?

src/lib/operators/sort.cpp
```diff
@@ -1,27 +1,47 @@
...
+#include <ranges>
 #include <string>
+#include <thread>
+#include <type_traits>
 #include <unordered_map>
 #include <utility>
 #include <vector>
 
+#include <boost/accumulators/statistics_fwd.hpp>
+#include <boost/algorithm/cxx11/iota.hpp>
```
Bouncner: Needed?

src/lib/operators/sort.cpp
```diff
@@ -1,27 +1,47 @@
...
+#include <numeric>
+#include <ranges>
 #include <string>
+#include <thread>
+#include <type_traits>
 #include <unordered_map>
 #include <utility>
 #include <vector>
 
+#include <boost/accumulators/statistics_fwd.hpp>
```
Bouncner: Needed?

src/lib/operators/sort.cpp
```diff
@@ -1,27 +1,47 @@
...
+#include <boost/algorithm/cxx11/iota.hpp>
+#include <boost/range/numeric.hpp>
```
Bouncner: Needed?

src/lib/operators/sort.cpp
```diff
@@ -33,10 +53,54 @@ namespace {
 
 using namespace hyrise;  // NOLINT
 
-// Ceiling of integer division
-size_t div_ceil(const size_t lhs, const ChunkOffset rhs) {
-  DebugAssert(rhs > 0, "Divisor must be larger than 0.");
-  return (lhs + rhs - 1u) / rhs;
+constexpr size_t STRING_CUTOFF = 64;
```
Bouncner, line 53: Please add a comment (you don't need to say how to you handle longer strings, but the general idea of cutting off strings for materialization). It's a pretty important number for this operator.

src/lib/operators/sort.cpp
```diff
@@ -54,94 +118,68 @@ std::shared_ptr<Table> write_materialized_output_table(const std::shared_ptr<con
...
-          value_segment_null_vector.reserve(next_chunk_size);
-        }
-      }
-
       auto accessor_by_chunk_id =
-          std::vector<std::unique_ptr<AbstractSegmentAccessor<ColumnDataType>>>(unsorted_table->chunk_count());
+          std::vector<std::unique_ptr<AbstractSegmentAccessor<ColumnDataType>>>(input_chunk_count);
       for (auto input_chunk_id = ChunkID{0}; input_chunk_id < input_chunk_count; ++input_chunk_id) {
         const auto& abstract_segment = unsorted_table->get_chunk(input_chunk_id)->get_segment(column_id);
+        DebugAssert(abstract_segment, "Segment should exist");
```
Bouncner:
```suggestion
        Assert(chunk, "Physically deleted chunk should not reach this point, see get_chunk / #1686.");
        DebugAssert(abstract_segment, "Expected segment.");
```
Assert the chunk as well. You'll find this snippet (first line) all over Hyrise.

src/lib/operators/sort.cpp
```diff
@@ -54,94 +118,68 @@ std::shared_ptr<Table> write_materialized_output_table(const std::shared_ptr<con
...
         accessor_by_chunk_id[input_chunk_id] = create_segment_accessor<ColumnDataType>(abstract_segment);
+        DebugAssert(accessor_by_chunk_id[input_chunk_id], "Accessor must be initialized");
```
Bouncner:
```suggestion
        DebugAssert(accessor_by_chunk_id[input_chunk_id], "Accessor must be initialized.");
```

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+  }
+
+  for (auto& segments : output_segments_by_chunk) {
+    output_table->append_chunk(segments);
+  }
+
+  return output_table;
+}
+
+/// Allocate an large chunk of uninitialized memory. This class is faster than C++ std::vector with resize, as it does
```
Bouncner:
```suggestion
// Allocate a large chunk of uninitialized memory. This class is faster than C++ std::vector with resize, as it does
```

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+/// not initialize each entry.
+template <typename T>
+class UVector {
```
Bouncner: Please check if you can use the `uninitialized_vector` we have in Hyrise (it's a third party module).

If you need yours: that's not a vector. Maybe `UnitializedBlob`/`UninitializedMemory`?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+template <typename T>
+auto to_unsigned(T value) {
+  if constexpr (std::is_same_v<decltype(value), int32_t> || std::is_same_v<decltype(value), float>) {
+    return std::bit_cast<uint32_t>(value);
+  } else if constexpr (std::is_same_v<decltype(value), int64_t> || std::is_same_v<decltype(value), double>) {
+    return std::bit_cast<uint64_t>(value);
+  }
+}
+
+struct ScanResult {
```
Bouncner, line 303: Okay, that's the result of a "scan". But what is a "scan"?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+};
+
+// Map signed to unsigned data types with same number of bytes (e.g., int32_t to uint32_t).
+template <typename T>
+auto to_unsigned(T value) {
+  if constexpr (std::is_same_v<decltype(value), int32_t> || std::is_same_v<decltype(value), float>) {
+    return std::bit_cast<uint32_t>(value);
+  } else if constexpr (std::is_same_v<decltype(value), int64_t> || std::is_same_v<decltype(value), double>) {
+    return std::bit_cast<uint64_t>(value);
+  }
```
Bouncner, line 293: Fail for other data types?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+}
+
+struct ScanResult {
+  // Maximum number of bytes required to encode all values.
```
Bouncner: All values of a tuple? Row? Or all values of the input table?

src/lib/operators/sort.cpp
Bouncner: Result for table, row, column, chunk?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+  size_t encoding_width = 0;
+  // Number of extra bytes required to encode the length of an string.
+  size_t extra_width = 0;
+  // Extra bytes required to encode long strings.
```
Bouncner: Is that on heap?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+
+// Copy an integer to the byte array. To ensure the encoded integer is comparable with memcmp the endianess of the
+// encoded integer is changed to big-endian. In addition, it will take the number of bytes necessary to encode this
+// integer, because of the variable length encoding of int32_t and uint32_t (see materialization and scanning).
+void copy_uint_to_byte_array(std::byte* byte_array, std::unsigned_integral auto uint,
+                             size_t len = sizeof(decltype(uint))) {
+  if constexpr (std::endian::native == std::endian::little) {
+    uint = byteswap(uint);
+  } else if constexpr (std::endian::native != std::endian::big) {
+    Fail("mixed-endian is unsupported");
```
Bouncner:
```suggestion
    Fail("Mixed-endian is unsupported.");
```

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+  bool nullable = false;
+  // List of long sorted strings.
+  std::vector<pmr_string> long_strings;
+
+  // Returns number of bytes required to encode all scanned values. Includes the extra byte for null values.
+  size_t width() const {
+    return encoding_width + ((long_strings.empty()) ? 0 : long_width) + extra_width + ((nullable) ? 1 : 0);
+  }
+
+  ScanResult merge(ScanResult& other) {
```
Bouncner, line 321: Is there a good reason to not sort the string lists in the thread jobs and use a linear merge here (probably `std::set_union`)?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+        .encoding_width = encoded_width,
+        .extra_width = std::max(extra_width, other.extra_width),
+        .long_width = long_width,
+        .nullable = nullable || other.nullable,
+        .long_strings = std::move(long_strings),
+    };
+  }
+};
+
+/// Scans column for maximum bytes necessary to encode all segment values and null values.
```
Bouncner: Are you using `///` on intention? We usually use `//` for short comments or `/** */` for longer comments.

Can you also elaborate a littttle bit (1-2 sentences) on what this scan does? You scan what and what do you mean by "encode"?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+template <typename ColumnDataType>
+ScanResult scan_column(const AbstractSegment& segment) {
+  auto result = ScanResult();
+  segment_with_iterators<ColumnDataType>(segment, [&](auto it, const auto end) {
+    while (it != end) {
+      const auto& segment_position = *it;
+      if (segment_position.is_null()) {
+        result.nullable = true;
```
Bouncner, line 352: Guard clause? `continue;`?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+      } else if constexpr (std::is_same_v<ColumnDataType, pmr_string>) {
+        const auto& value = segment_position.value();
+        const auto size = std::min(value.size(), STRING_CUTOFF);
+        result.encoding_width = std::max(result.encoding_width, size);
+        result.extra_width = sizeof(uint64_t) - (std::countl_zero(result.encoding_width) / size_t{8});
```
Bouncner: Please add a comment.

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+
+        if (value.size() >= STRING_CUTOFF) {
+          result.long_strings.push_back(value);
+        }
+      } else {
+        const auto& value = segment_position.value();
+        if constexpr (std::is_same_v<ColumnDataType, int32_t> || std::is_same_v<ColumnDataType, int64_t>) {
+          using UnsignedColumnDataType = decltype(to_unsigned(ColumnDataType{}));
+          static_assert(std::is_unsigned_v<UnsignedColumnDataType>, "Can not convert column data type to unsigned");
+          // We often encounter small number, but we still 32-bit numbers while 8 or 16 bit would be sufficient.
```
Bouncner:
```suggestion
          // We often encounter small numbers, but we still (USE?!?) 32-bit numbers while 8 or 16 bit would be sufficient.
```

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+          // Because of that we want to reduce the bit-width of our integer numbers to the bare minimum. The idea
+          // is to drop all leading 0 for positive integers and all 1 leading 1 for negative numbers. We only keep
```
Bouncner:
```suggestion
          // is to drop all leading 0 for positive integers and all leading ones for negative numbers. We only keep
```

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+          // 1-bit to store the signess for our integers. To keep the reduction simple we only drop full bytes.
```
Bouncner: Above you say "reduce the bit-width of our integer numbers to the bare minimum." and then here "only drop full bytes".

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+          //
+          // Example for 32-bit:
+          //
+          // Number | Hex Representation | Number of dropped bytes | Shortened Int
+          // -------|--------------------|-------------------------|--------------
+          //   -42  | 0xFFFFFFd6         | 6                       | 0xd6
```
Bouncner: You drop 6 bytes from a 4-byte value?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+// Swap byte order. Should be replaced with C++23's std::byteswap if possible.
+auto byteswap(std::unsigned_integral auto value) {
+  auto* uint_bytes = reinterpret_cast<std::byte*>(&value);
+  for (auto start = uint_bytes, end = uint_bytes + sizeof(decltype(value)) - 1; start < end; ++start, --end) {
+    std::swap(*start, *end);
+  }
+  return value;
+}
+
+// Copy an integer to the byte array. To ensure the encoded integer is comparable with memcmp the endianess of the
```
Bouncner:
```suggestion
// Copy an integer to the byte array. To ensure the encoded integer is comparable with memcmp, the endianess of the
```

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+// encoded integer is changed to big-endian. In addition, it will take the number of bytes necessary to encode this
+// integer, because of the variable length encoding of int32_t and uint32_t (see materialization and scanning).
```
Bouncner, line 411: We don't have a `uint32_t` type in Hyrise. Integers are always signed.

.gitignore
```diff
@@ -56,3 +56,5 @@ sftp*
 # Clangd files
 compile_commands.json
 .cache/
+
+file:*
```
Bouncner, line 60: ?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+  } else if constexpr (std::endian::native != std::endian::big) {
+    Fail("mixed-endian is unsupported");
+  }
+  const auto* uint_byte_array = reinterpret_cast<std::byte*>(&uint);
+  memcpy(byte_array, uint_byte_array + (sizeof(decltype(uint)) - len), len);
+}
+
+using NormalizedKeyIter = NormalizedKeyRow*;
+
+// Append segement's values as byte array to the normalized keys. Expects that the normalized_key_iter is valid for
```
Bouncner:
```suggestion
// Append segment values as byte array to the normalized keys. Expects that the normalized_key_iter is valid for
```

Bouncner: "Expects that the normalized_key_iter is valid for .."

What does valid mean? That the underlying data structure in properly sized? Who moves the iterator? Can we accidentally make it invalid?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+// each segment's values.
+//
+// Parameters:
+// @param segment              Segment to encode values from
+// @param offset               Offset to the start of the normalized key row's byte array.
+// @param expected_width       The number of bytes all values must be encoded to.
```
Bouncner: All values of the segment? Of the sortable table? Of the row?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+// @param ascending            Sort normalized keys in ascending order.
+// @param nullable             Put exrta byte in front to encode null values.
+// @param nulls_first          Put null values to the start of the normalized key order else to the back.
```
Bouncner, line 434: Switch `nullable` and `null_first` params?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+// @param normalized_key_iter  Iterator to the normalized keys.
```
Bouncner, line 435: How is `offset` related to this iterator?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+template <typename ColumnDataType>
+void materialize_segment_as_normalized_keys(const AbstractSegment& segment, const size_t offset, const bool ascending,
+                                            const ScanResult column_info, const bool nulls_first,
+                                            NormalizedKeyIter normalized_key_iter) {
+  // Normalized keys are ordered in ascendingly by default. The modifier is applied to invert this order.
+  const auto modifier = (ascending) ? std::byte{0x00} : std::byte{0xFF};
+  const auto full_modifier = (ascending) ? uint64_t{0x00} : std::numeric_limits<uint64_t>::max();
```
Bouncner: From the name, I wouldn't know what it does.

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+  const auto normalized_null_value = ((nulls_first) ? std::byte{0x00} : std::byte{0xFF});
+  const auto normalized_non_null_value = ((nulls_first) ? std::byte{0xFF} : std::byte{0x00});
+  segment_with_iterators<ColumnDataType>(segment, [&](auto it, const auto end) {
+    while (it != end) {
+      const auto segment_position = *it;
+      DebugAssert(column_info.nullable || !segment_position.is_null(),
+                  "Segment must be nullable or contains no null values");
```
Bouncner:
```suggestion
                  "Segment must be nullable or contain no null values`");
```

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+      auto* normalized_key_start = normalized_key_iter->key_head + offset;
+
+      // Encode the null byte for nullable segments. This byte is used to order null vs. non-null values (e.g.
+      // put them to the front or end depending on the specified null order)
```
Bouncner:
```suggestion
      // put them to the front or end depending on the specified null order).
```

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+      if (column_info.nullable && segment_position.is_null()) {
+        *(normalized_key_start++) = normalized_null_value;
+        // Initialize actual key by setting all bytes to 0.
+        for (auto counter = size_t{0}; counter < column_info.width() - 1; ++counter) {
```
Bouncner, line 458: Why `width() - 1`?

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+    while (it != end) {
+      const auto segment_position = *it;
+      DebugAssert(column_info.nullable || !segment_position.is_null(),
+                  "Segment must be nullable or contains no null values");
+      auto* normalized_key_start = normalized_key_iter->key_head + offset;
+
+      // Encode the null byte for nullable segments. This byte is used to order null vs. non-null values (e.g.
+      // put them to the front or end depending on the specified null order)
+      if (column_info.nullable && segment_position.is_null()) {
+        *(normalized_key_start++) = normalized_null_value;
```
Bouncner, line 456: We prefer to put increments on the next line when possible.

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
+        for (auto counter = segment_position.value().size(); counter < column_info.encoding_width; ++counter) {
+          *(normalized_key_start++) = std::byte{0} ^ modifier;
+        }
+
+        // Encode an identifier for long strings. Use 0 if string is small.
+        if (!column_info.long_strings.empty()) {
+          if (value.size() < STRING_CUTOFF) {
+            copy_uint_to_byte_array(normalized_key_start, uint64_t{0}, column_info.long_width);
+          } else {
+            const auto iter = std::ranges::lower_bound(column_info.long_strings, value);  // NOLINT
```
Bouncner: Add the name of the silenced rule to NOLINT.

src/lib/operators/sort.cpp
```diff
@@ -181,92 +219,1123 @@ std::shared_ptr<Table> write_reference_output_table(const std::shared_ptr<const
...
-                      "Input column references more than one column");
-          const auto& input_reference_pos_list = input_reference_segment.pos_list();
-          output_pos_list->emplace_back((*input_reference_pos_list)[row_id.chunk_offset]);
+
+        // Add null bytes to pad values to correct size. (All keys must have the same size)
+        for (auto counter = segment_position.value().size(); counter < column_info.encoding_width; ++counter) {
+          *(normalized_key_start++) = std::byte{0} ^ modifier;
+        }
+
+        // Encode an identifier for long strings. Use 0 if string is small.
```
Bouncner: "identifier" does not tell us much about this very nice trick that allows you to skip most (all?) resolving afterwards.
Maybe mention we are binary searching for string and store offset?

Bouncner (commented):
Still haven't read all the code.

What would be help is a very brief high level description of IPSO. The paper has  a very readable  introduction of "buckets" and other terms. 
Don't put too much effort in it, but a parapraph of the high level concept wiould be very helpful.

CMakeLists.txt
robert-oleynik: No, we just hat a flag to enable/disable perfetto on demand (mainly because it caused build issues on some systems). This is just an artifact and I will remove this.

src/lib/operators/sort.cpp
robert-oleynik: I've removed that function as it is only needed for `run_parallel_batched`.

src/lib/operators/sort.cpp
robert-oleynik: I've removed that function and replaced it with `schedule_and_wait_for_tasks`, but this maybe comes at a performance cost.

src/lib/operators/sort.cpp
robert-oleynik: `run_parallel_batched` is replaced with `schedule_and_wait_for_task`.

src/lib/operators/sort.cpp
robert-oleynik: The time we introduced that, there was no measurable performance benefit. Since then a lot of things changed, so a new evaluation may be necessary.

src/lib/operators/sort.cpp
robert-oleynik: Not used anymore

src/lib/operators/sort.cpp
robert-oleynik: Merge moves the values into the result array.

src/lib/operators/sort.cpp
robert-oleynik: Passes may be overloaded a bit. I have rewritten the comment for ips4o to be a bit more detailed.

src/lib/operators/sort.cpp
robert-oleynik: Because a debug assert would remove the call to _pending_reads if inlined, or it would cause lint issues for the release build.

src/lib/operators/sort.cpp
robert-oleynik: There is a short explanation for blocks and stripes in the documentation of ips4o. I hope this is sufficient.

src/lib/operators/sort.cpp
robert-oleynik: See documentation of `hyrise::ips4o::sort` in the utils.

.gitignore
robert-oleynik: There are a lot of `file:sqlitewrapper_<number>?mode=memory&chache=shared` files generated. This just excludes them from commit view.

src/lib/operators/sort.cpp
robert-oleynik: `uninitialized_vector` seems to be working fine.

src/lib/operators/sort.cpp
robert-oleynik: Variable length encoding is also used for the offset into the long string array and the length of a string. These are unsigned integers.

src/lib/operators/sort.cpp
robert-oleynik: I've updated the comment

src/lib/operators/sort.cpp
robert-oleynik: I've added a little bit of documentation to this struct

src/lib/operators/sort.cpp
robert-oleynik: No, this is just the number of bytes required to encode the offset into the long string array.

src/lib/operators/sort.cpp
robert-oleynik: I've updated the docs.

src/lib/operators/sort.cpp
robert-oleynik: No, because it is a while loop, but I changed the loop to a for-loop and it is a guard clause now.

src/lib/operators/sort.cpp
robert-oleynik: I modified the description

src/lib/operators/sort.cpp
robert-oleynik: No, the offset is into the bytes of the normalized key and is dependent on the column.

src/lib/operators/sort.cpp
robert-oleynik: `nullable` is not passed as parameter. I changed the documentation to `column_info`

src/lib/operators/sort.cpp
robert-oleynik: I updated the docs

src/lib/operators/sort.cpp
Bouncner: Yes, I know. But why one large sort instead of several small local ones (also allows removing duplicates early on) with linear merging?

src/lib/operators/sort.cpp
```diff
@@ -337,26 +302,36 @@ auto to_unsigned(T value) {
     return std::bit_cast<uint32_t>(value);
   } else if constexpr (std::is_same_v<decltype(value), int64_t> || std::is_same_v<decltype(value), double>) {
     return std::bit_cast<uint64_t>(value);
+  } else {
+    static_assert(false, "Unsupported data type");
```
Bouncner: Using some like `requires` would be a cleaner solution here.

src/lib/operators/sort.cpp
robert-oleynik: It depends on the amount of large strings. Using a few sorted small arrays and merge them seems to be fine, but than it is also not to costly sorting all of them at once. If we want to merge large sorted arrays, we will into the same performance problem as for merging long normalized key. Instead, we could improve the string sorting, by using IPS4o which should yield better performance. A bit of testing would be necessary to decide on that.

